PR #8, opened 2019-08-22
ADD: cmd flushall

karelrooted:


src/redis_db.cc
```diff
@@ -248,7 +248,43 @@ rocksdb::Status Database::FlushDB() {
...
+    delete iter;
+    return rocksdb::Status::OK();
+  }
+  auto last_key = iter->key().ToString();
+  auto s = db_->DeleteRange(rocksdb::WriteOptions(), metadata_cf_handle_, first_key, last_key);
+  if (!s.ok()) {
+    delete iter;
+    return s;
+  }
+  s = db_->Delete(rocksdb::WriteOptions(), metadata_cf_handle_, last_key);
```
git-hulk: // 添加一下为什么删除 last key 注释: we want to flush all keys but the  end_key was excluded in `DeleteRange ` so using single delete here

PR #16, opened 2019-08-30
Feature/add sortedint data type

karelrooted:


src/redis_cmd.cc
```diff
@@ -2432,6 +2433,129 @@ class CommandZInterStore : public CommandZUnionStore {
   }
 };
 
+class CommandSortintAdd : public Commander {
```
git-hulk: rename the Sortint to SortedInt because the Sort is a verb, not adj.

src/redis_cmd.cc
```diff
@@ -2432,6 +2433,129 @@ class CommandZInterStore : public CommandZUnionStore {
...
+ public:
+  CommandSortintAdd() : Commander("siadd", -3, true) {}
+
+  Status Parse(const std::vector<std::string> &args) override {
+    try {
+      for (unsigned i = 2; i < args.size(); i++) {
+        auto id = std::stoi(args[i]);
```
git-hulk: `std::stoi()` would return int type but the uint64 wants, use `std::stoull` here

src/redis_cmd.cc
```diff
@@ -2432,6 +2433,129 @@ class CommandZInterStore : public CommandZUnionStore {
...
+};
+
+class CommandSortintRem : public Commander {
+ public:
+  CommandSortintRem() : Commander("sirem", -3, true) {}
+
+  Status Parse(const std::vector<std::string> &args) override {
+    try {
+      for (unsigned i = 2; i < args.size(); i++) {
+        auto id = std::stoi(args[i]);
```
git-hulk: ditto

src/redis_sortint.cc
```diff
@@ -0,0 +1,137 @@
...
+  for (const auto id : ids) {
+    std::string id_buf;
+    PutFixed64(&id_buf, id);
+    InternalKey(ns_key, id_buf, metadata.version).Encode(&sub_key);
+    s = db_->Get(rocksdb::ReadOptions(), sub_key, &value);
+    if (!s.ok()) continue;
+    batch.Delete(sub_key);
+    *ret += 1;
+  }
+  if (*ret > 0) {
```
git-hulk: do nothing when didn't delete any member
```
if (*ret == 0) return Status::OK()
metadata.size -= *ret;
std::string bytes;
metadata.Encode(&bytes);
batch.Put(metadata_cf_handle_, ns_key, bytes);
return storage_->Write(rocksdb::WriteOptions(), &batch);
```

src/redis_sortint.cc
```diff
@@ -0,0 +1,137 @@
...
+  for (!reversed ? iter->Seek(start_key) : iter->SeekForPrev(start_key);
+       iter->Valid() && iter->key().starts_with(prefix);
+       !reversed ? iter->Next() : iter->Prev()) {
+    InternalKey ikey(iter->key());
+    Slice sub_key = ikey.GetSubKey();
+    GetFixed64(&sub_key, &id);
+    if (id == cursor_id) {
+      continue;
+    }
+    if (offset >= 0 && pos++ < offset) continue;
```
git-hulk: simplify:

```
if ( id == cursor_id || pos++ < offset ) continue
```

git-hulk:
LGTM

PR #80, opened 2020-04-29
FIX: if fullsync and restorefrombackup failed, we should keep retry f…

karelrooted:
…ullsync, stop any other action, any use of db or cf_handler pointer will cause core dump

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -579,8 +579,7 @@ ReplicationThread::CBState ReplicationThread::fullSyncReadCB(bufferevent *bev,
       self->pre_fullsync_cb_();
       s = self->storage_->RestoreFromBackup();
       if (!s.IsOK()) {
-        LOG(ERROR) << "[replication] Failed to restore backup while " + s.Msg();
-        self->post_fullsync_cb_();
+        LOG(ERROR) << "[replication] Failed to restore backup while " + s.Msg() + ", restart fullsync";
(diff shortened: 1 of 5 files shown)


src/task_runner.cc
```diff
@@ -28,15 +28,14 @@ void TaskRunner::Start() {
   }
 }
 
-void TaskRunner::Restart() {
+void TaskRunner::StopAndClear() {
```
git-hulk, line 31: 可以增加一个 Purge 函数就好了, 然后外部通过调用:

task_runner.Stop();
task_runner.Purge()

会更加清晰一些

src/replication.cc
```diff
@@ -579,8 +579,7 @@ ReplicationThread::CBState ReplicationThread::fullSyncReadCB(bufferevent *bev,
       self->pre_fullsync_cb_();
       s = self->storage_->RestoreFromBackup();
       if (!s.IsOK()) {
-        LOG(ERROR) << "[replication] Failed to restore backup while " + s.Msg();
-        self->post_fullsync_cb_();
```
git-hulk, line 583: post_fullsunc_cb 为啥去掉了

karelrooted: restore失败后，这是已经调用了closedb, db和cf指针是空的，一旦调用了post_fullsync_cb_, 会有新请求进来，使用到db_和cf指针

PR #98, opened 2020-07-02
improve the compaction

git-hulk:
the main goal of the compaction checker was to avoid unnecessary full compaction every day. How it works:
* step 1: check the current hour whether in compaction checker range or not, if yes go to step 2
* step 2: fetch all table properties and then pick the file was created
more than 1 week or which has highest delete ratio every minute
* step 3: compact keys in the selected file

### TODO

- [x]  fix pubsub/slot compaction

- [x]  test compaction checker was worked as expected

- [ ] smarter compaction rule

### DIFF
--- a/src/storage.cc
+++ b/src/storage.cc
@@ -18,15 +18,19 @@
 #include "redis_slot.h"
 #include "event_listener.h"
 #include "compact_filter.h"
+#include "table_properties_collector.h"
 
 namespace Engine {
 
 const char *kPubSubColumnFamilyName = "pubsub";
 const char *kZSetScoreColumnFamilyName = "zset_score";
 const char *kMetadataColumnFamilyName = "metadata";
+const char *kSubkeyColumnFamilyName = "default";
 const char *kSlotMetadataColumnFamilyName = "slot_metadata";
 const char *kSlotColumnFamilyName = "slot";
+
 const uint64_t kIORateLimitMaxMb = 1024000;
+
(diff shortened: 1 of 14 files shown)


src/storage.h
```diff
@@ -70,6 +70,7 @@ class Storage {
   void IncrFlushCount(uint64_t n) { flush_count_.fetch_add(n); }
   uint64_t GetCompactionCount() { return compaction_count_; }
   void IncrCompactionCount(uint64_t n) { compaction_count_.fetch_add(n); }
+  bool CondisEnabled() { return config_->codis_enabled; }
```
karelrooted, line 73: typo

src/storage.cc
```diff
@@ -174,6 +178,8 @@ Status Storage::Open(bool read_only) {
   subkey_opts.table_factory.reset(rocksdb::NewBlockBasedTableFactory(subkey_table_opts));
   subkey_opts.compaction_filter_factory = std::make_shared<SubKeyFilterFactory>(this);
   subkey_opts.disable_auto_compactions = config_->RocksDB.disable_auto_compactions;
+  metadata_opts.table_properties_collector_factories.emplace_back(
```
karelrooted, line 181: typo, metadata_opts => subkey_opts

git-hulk: good catch

karelrooted:
LGTM

PR #100, opened 2020-07-08
FIX: GetUpdatesSince should not get data that is older than seq (seem…

karelrooted:
…s like a bug in rocksdb...)

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -81,7 +81,8 @@ void FeedSlaveThread::loop() {
     // iter_ would be always valid here
     auto batch = iter_->GetBatch();
     if (batch.sequence != next_repl_seq_) {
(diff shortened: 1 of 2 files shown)


src/replication.cc
```diff
@@ -81,7 +81,9 @@ void FeedSlaveThread::loop() {
     // iter_ would be always valid here
     auto batch = iter_->GetBatch();
     if (batch.sequence != next_repl_seq_) {
-      LOG(ERROR) << "Fatal error encountered, WAL iterator is discrete, some seq might be lost";
+      LOG(ERROR) << "Fatal error encountered, WAL iterator is discrete, some seq might be lost"
+                 << ", expect sequence: " << next_repl_seq_
```
git-hulk, line 85: expect => expected

get => got

PR #107, opened 2020-08-04
FIX: get command would return nil string when the value was empty

git-hulk:


src/redis_cmd.cc
```diff
@@ -355,8 +363,9 @@ class CommandMGet : public Commander {
     }
     std::vector<std::string> values;
     // always return OK
-    string_db.MGet(keys, &values);
-    *output = Redis::MultiBulkString(values);
+    std::vector<std::string> result;
```
karelrooted, line 366: unused variable?

PR #115, opened 2020-08-18
Update redis_geo.cc

byronhe:


src/redis_geo.cc
```diff
@@ -174,7 +174,7 @@ rocksdb::Status Geo::MGet(const Slice &user_key,
 }
 
 std::string Geo::EncodeGeoHash(double longitude, double latitude) {
-  std::string geoalphabet = "0123456789bcdefghjkmnpqrstuvwxyz";
+  const static std::string geoalphabet = "0123456789bcdefghjkmnpqrstuvwxyz";
```
git-hulk, line 177: CI failed, `static` identifier is unnecessary

git-hulk (changes requested):

src/redis_geo.cc
ShooterIT: > src/redis_geo.cc:177:`  Storage-class specifier (static, extern, typedef, etc) should be at the beginning of the declaration.

@git-hulk From cpplint detail, it seems we should change `const static` into `static const`?

git-hulk: yeah, or remove the `static` should be OK

ShooterIT: Hi @git-hulk @byronhe Since we haven't handle with it for a long time, i just remove `static` to resolve this patch

ShooterIT (approved):

git-hulk (approved):

PR #120, opened 2020-08-24
dynamic resize block and sst

### LINKED ISSUE #118: find the suitable size for SST
the key-value size may so be quite different in many scenes, and use 256MiB as SST file size may cause data loading(large index/filter block) ineffective when the key-value was too small. kvrocks supports user-defined SST file in config(`rocksdb.target_file_size_base`), but it still too trivial and inconvenient to adjust the different sizes for different instances.  so we want to periodic auto-adjust the SST size in-flight with user avg key-value size. for example:

```
case 1: 0 ~ 128 bytes，Block Size = 2k, SST size = 16 M, block size = 

case 2:  128 ~ 1024 bytes, Block Size = 8k, SST size = 64 M

case 3: 1k ~ 32k, Block Size = 32k,   SST size = 128 M

case 4: 32k ~ 256k, Block Size = 256k, SST size = 256 M

case 5: 256k ~ 512k, Block Size = 512k , SST size = 512 M

case 6: > 512 k, Block Size = 1M, SST size = 1 G
```

TODO:

- [x] develop the feature

- [x] code review

- [ ] benchmark between fixed and dynamic SST size(32bytes/128bytes/512bytes/1k/16k/64k/256k/1M)

- [ ] how about dynamic block size? which may affect the performance of the block cache hit rate

karelrooted:
resolved #118

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -279,8 +279,8 @@ rocksdb.write_buffer_size 64
 # Target file size for compaction, target file size for Leve N can be caculated
 # by target_file_size_base * (target_file_size_multiplier ^ (L-1))
 #
-# Default: 256MB
-rocksdb.target_file_size_base 256
+# Default: 128MB
+rocksdb.target_file_size_base 128
 
 # The maximum number of write buffers that are built up in memory.
 # The default and the minimum number is 2, so that when 1 write buffer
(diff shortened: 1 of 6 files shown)


git-hulk (approved):
LGTM

git-hulk (approved):

git-hulk (approved):
LGTM

PR #136, opened 2020-11-20
FIX:  replication dead loop  (some seq might be lost  in the middle of wal …

karelrooted:
…, seems like a bug in rocksdb)

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -65,6 +65,10 @@ void FeedSlaveThread::checkLivenessIfNeed() {
 }
 
 void FeedSlaveThread::loop() {
+  // is_first_repl_batch was used to fix that replication may be stuck in a dead loop
+  // when some seqs might be lost in the middle of the WAL log, so forced to replicate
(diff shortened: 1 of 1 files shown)


src/replication.cc
```diff
@@ -65,6 +65,7 @@ void FeedSlaveThread::checkLivenessIfNeed() {
 }
 
 void FeedSlaveThread::loop() {
+  bool send_by_batch = false;
```
git-hulk, line 68: how about renaming the `send_by_batch ` to `is_first_repl_batch` with a comment on why we need it

```
// is_first_repl_batch was used to fix that replication may be stuck in a dead loop 
// when some seqs might be lost in the middle of the WAL log, so forced to replicate
// first batch here to work around this issue instead of waiting for enough batch size. 
bool is_first_repl_batch = false;
```

git-hulk (approved):

git-hulk:
LGTM, nice~

PR #138, opened 2020-12-10
Add gTest related pkg to README

clyang:
When the user want to run `make test` for testing, the process will fail if gTest is not installed. Hence, it would be a good idea to add gTest to README.


git-hulk:
LGTM, thanks @clyang

PR #155, opened 2021-01-28
Support to set separate log directory

ShooterIT:
If logs of server are stored in working directory, there may be a little messy, I think we  support to set separate log directory, so i add a configuration to support that.


git-hulk (approved):
LGTM, thanks @ShooterIT

PR #156, opened 2021-01-28
Implement psetex command and handle negative expire time when set

ShooterIT:
We said we support `psetex` command but actually we don't support, and no body finds that 🤣

For `setex` `psetex`  `set ex/pex` commands, we should check expire time is negative or not. 
There is redis code https://github.com/redis/redis/blob/unstable/src/t_string.c#L81

### DIFF
--- a/tests/functional/string_test.py
+++ b/tests/functional/string_test.py
@@ -101,13 +101,13 @@ def test_setex():
 def test_psetex():
     key = "test_psetex"
     conn = get_redis_conn()
-    ret = conn.psetex(key, "bar", 1000*1024)
+    ret = conn.psetex(key, 1000*1024, "bar")
     assert(ret == True)
     ret = conn.ttl(key)
     assert(ret >= 1023 and ret <= 1025)
     ret = conn.delete(key)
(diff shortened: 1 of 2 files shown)


tests/functional/string_test.py
```diff
@@ -98,6 +98,17 @@ def test_setex():
     assert(ret == 1)
     assert_raise(redis.RedisError, conn.setex, "foo", "bar", "invalid")
 
+def test_psetex():
+    key = "test_psetex"
+    conn = get_redis_conn()
+    ret = conn.psetex(key, "bar", 1000*1024)
```
karelrooted, line 104: need to change the parameter order: psetex {time_ms} {key}

ShooterIT: Thanks @karelrooted Tbn, i didn't run this test, initially, i copied test form `test_setex`, it seems `psetex` is different with `setex`, `setex` syntax is `setex key value time`, but `psetex` still respect redis syntax.

karelrooted (approved):
LGTM

PR #162, opened 2021-02-02
Compatible with redis protocol error handlers

ShooterIT:
- Make some errors compatible with redis
- Add max multibulk size limit

Hi @git-hulk I am planting redis tcl regression tests to kvrocks, now, find some protocol errors are different with redis, to make this job easy, we should change this errors, plead have a look.

### DIFF
--- a/src/redis_request.cc
+++ b/src/redis_request.cc
@@ -14,6 +14,7 @@
 namespace Redis {
 const size_t PROTO_INLINE_MAX_SIZE = 16 * 1024L;
 const size_t PROTO_BULK_MAX_SIZE = 512 * 1024L * 1024L;
+const size_t PROTO_MULTI_MAX_SIZE = 1024 * 1024L;
 
 Status Request::Tokenize(evbuffer *input) {
   char *line;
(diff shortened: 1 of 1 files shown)


git-hulk (approved):
LGTM, cool. Thanks for your great work!

src/redis_request.cc
```diff
@@ -36,13 +37,17 @@ Status Request::Tokenize(evbuffer *input) {
...
+          if (multi_bulk_len_ > PROTO_MULTI_MAX_SIZE) {
+            free(line);
+            return Status(Status::NotOK, "Protocol error: invalid multibulk length");
           }
           state_ = BulkLen;
         } else {
           if (len > PROTO_INLINE_MAX_SIZE) {
             free(line);
-            return Status(Status::NotOK, "Protocol error: too big inline request");
+            return Status(Status::NotOK, "Protocol error: invalid bulk length");
```
enjoy-binbin, line 50: so is this change intentional? i see the arg name saying INLINE, just trying to confirm

PR #164, opened 2021-02-02
Fix setbit/getbit bit offset out of range

ShooterIT:
For `std::stoul`, if arg is negative number, there is no exception, it will be very big number.
Why I limit the range into [0, INT_MAX] is that i found we get offset by `std::stoi` in `msetbit/bitcount`, do i miss somethings?

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -620,15 +620,26 @@ class CommandDel : public Commander {
   }
 };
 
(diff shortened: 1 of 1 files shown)


git-hulk:
yeah, you're right that we should limit the range of the bit offset👍 . can you help to fix the lint error.

git-hulk (approved):

PR #165, opened 2021-02-02
Support msetnx command

ShooterIT:
For msetnex, we should try best effort to guarantee atomicity of command,
so we should check if all keys exist, return 0 directly if one of them
already existed.

For test cases, I will plant redis tcl test ASAP.


karelrooted (approved):
LGTM

PR #167, opened 2021-02-02
Fix incrby/incrbyfloat on wrong value and float precision check

ShooterIT:
For example
```
set a " 10" // note there is a blank space
incrby a 1
```
before this commit, we would rewrite a into 11.

It seems `incr` and `decr` also have this problem before this commit.

if the arg of incrby is 'inf', the sum of `-1` and `inf` also is `inf`, so i changed float precision check. actually for float/double, overflow is `inf` or `-inf`

### DIFF
--- a/src/redis_string.cc
+++ b/src/redis_string.cc
@@ -233,7 +233,7 @@ rocksdb::Status String::IncrByFloat(const std::string &user_key, double incremen
     metadata.Encode(&raw_value);
   }
   value = raw_value.substr(STRING_HDR_SIZE, raw_value.size()-STRING_HDR_SIZE);
-  float n = 0;
+  double n = 0;
   std::size_t idx;
   if (!value.empty()) {
     try {
--- a/src/redis_string.cc
+++ b/src/redis_string.cc
@@ -2,6 +2,7 @@
 #include <utility>
 #include <string>
 #include <limits>
+#include <cmath>
 
 namespace Redis {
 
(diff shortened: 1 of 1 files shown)


src/redis_string.cc
```diff
@@ -228,21 +233,23 @@ rocksdb::Status String::IncrByFloat(const std::string &user_key, double incremen
     metadata.Encode(&raw_value);
   }
   value = raw_value.substr(STRING_HDR_SIZE, raw_value.size()-STRING_HDR_SIZE);
-  double n = 0;
+  float n = 0;
```
ShooterIT, line 236:
```suggestion
  double n = 0;
```

ShooterIT:
Hi @git-hulk other problem. Thes function std::stoi/std::stoll/std::stof/std::stod don't seem to satisfy to get `int/long long/float/double` from arguments. Incrby command just is happened to find by tcl test.

git-hulk:
Hi @ShooterIT, it's ok to migrate `string2xx` functions from Redis if necessary.

git-hulk (approved):
LGTM

ShooterIT:
> it's ok to migrate `string2xx` functions from Redis if necessary.

Yes, this is a solution. We also can wrap `std::stoi/std::stoll/std::stof/std::stod` and check the string totally represents integer or float number, just like https://github.com/bitleak/kvrocks/pull/167/files#diff-cae8084ef37038efad2d99da8eb5761ffc0dd8d8354eb136fbca1b522cddf61dR244

I think it is not urgent, we can fix it later.

PR #168, opened 2021-02-02
Fix hincrby/hincrbyfloat on wrong value and float precision check

ShooterIT:
Similar to #167

### DIFF
--- a/src/redis_hash.cc
+++ b/src/redis_hash.cc
@@ -1,5 +1,6 @@
 #include "redis_hash.h"
 #include <limits>
+#include <cmath>
 #include <iostream>
 #include <rocksdb/status.h>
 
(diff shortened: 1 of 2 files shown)


src/redis_hash.cc
```diff
@@ -96,20 +100,24 @@ rocksdb::Status Hash::IncrByFloat(const Slice &user_key, const Slice &field, flo
...
+        return rocksdb::Status::InvalidArgument("value is not an float");
+      }
       exists = true;
     }
   }
-  if ((increment < 0 && old_value < 0 && increment < (std::numeric_limits<float>::lowest()-old_value))
-      || (increment > 0 && old_value > 0 && increment > (std::numeric_limits<float>::max()-old_value))) {
-    return rocksdb::Status::InvalidArgument("increment or decrement would overflow");
+  if (increment+old_value < std::numeric_limits<float>::lowest()
+      || increment+old_value > std::numeric_limits<float>::max()) {
```
git-hulk, line 119: we should use `increment > (std::numeric_limits<float>::max()-old_value)` to check whether it's overflow or not, since the `increment+old_value` may overflow and it would smaller than std::numeric_limits<float>::max().

git-hulk (approved):

src/redis_hash.cc
ShooterIT: Thanks @git-hulk 
I think the overflow behavior of float/double is different with integer.
`std::numeric_limits<float>::max() -/+ 1` still equal `std::numeric_limits<float>::max() `
`std::numeric_limits<float>::min() -/+ 1` still equal `std::numeric_limits<float>::min() `
It seems the overflowed value of float/double is Inf, so we can only check the result is Inf or Nan
https://github.com/redis/redis/blob/unstable/src/t_string.c#L643

git-hulk (dismissed):
LGTM

git-hulk (approved):
LGTM

PR #170, opened 2021-02-03
Fix some list bug

ShooterIT:
- For rpoplpush command, don't pop keys if dst key is not list
- For linsert command, return -1 if don't find index
- Fix ltrim command range calculation

For ltrim command, bad cases are following
```
lpush key 1 2 3 4 5
ltrim key 0 -5 => should return 1, instead of  1 2 3 4 5
ltrim key 0 -6 => should return empty, instead of 1 2 3 4 5
```


git-hulk (approved):
LGTM 👍

PR #174, opened 2021-02-04
Fix subscribe and unsubscribe command and pattern mode

ShooterIT:
- Fix subscribe and psubscribe command reply, the return number
  is the sum of subscription and psubscription
- Fix unpsubscribe command operates on subscription data instead of
  psubscription
- The unsubscribe and punsubscribe commands can operates on more than
  one channel and pattern
- The unsubscribe and punsubscribe commands should have reply

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -3270,16 +3270,21 @@ class CommandPublish : public Commander {
   }
 };
 
+void SubscribeCommmandReply(std::string *output, std::string name, std::string sub_name, int num) {
+  output->append(Redis::MultiLen(3));
+  output->append(Redis::BulkString(name));
+  output->append(sub_name.empty() ? Redis::NilString() : Redis::BulkString(sub_name));
+  output->append(Redis::Integer(num));
+}
+
 class CommandSubscribe : public Commander {
  public:
-  CommandSubscribe() : Commander("subcribe", -2, false) {}
+  CommandSubscribe() : Commander("subscribe", -2, false) {}
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     for (unsigned i = 1; i < args_.size(); i++) {
       conn->SubscribeChannel(args_[i]);
-      output->append(Redis::MultiLen(3));
-      output->append(Redis::BulkString("subscribe"));
-      output->append(Redis::BulkString(args_[i]));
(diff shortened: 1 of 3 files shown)


src/redis_cmd.cc
```diff
@@ -3270,57 +3270,72 @@ class CommandPublish : public Commander {
...
-  CommandUnSubscribe() : Commander("unsubcribe", -1, false) {}
+  CommandUnSubscribe() : Commander("unsubscribe", -1, false) {}
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    if (args_.size() > 1) {
-      conn->UnSubscribeChannel(args_[1]);
-    } else {
+    if (args_.size() == 1) {
       conn->UnSubscribeAll();
+      SubscribeCommmandReply(output, "unsubscribe", "",
+        conn->SubscriptionsCount() + conn->PSubscriptionsCount());
```
karelrooted, line 3300: in unsubscribe all, a message for every unsubscribed channel will be sent to the client in stock redis (verified on redis v5.0.4)， is this response on purpose？

src/redis_cmd.cc
```diff
@@ -3270,57 +3270,72 @@ class CommandPublish : public Commander {
...
  public:
-  CommandPUnSubscribe() : Commander("punsubcribe", -1, false) {}
+  CommandPUnSubscribe() : Commander("punsubscribe", -1, false) {}
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    if (args_.size() > 1) {
-      conn->PUnSubscribeChannel(args_[1]);
+    if (args_.size() == 1) {
+      conn->PUnSubscribeAll();;
+      SubscribeCommmandReply(output, "punsubscribe", "",
+        conn->SubscriptionsCount() + conn->PSubscriptionsCount());
```
karelrooted, line 3332: in punsubscribe all, a message for every punsubscribed channel_patten will be sent to the client in stock redis (verified on redis v5.0.4)， is this response on purpose？

karelrooted (changes requested):

src/redis_cmd.cc
ShooterIT: You are right, i miss that, please have a look again.

karelrooted (approved):
LGTM

PR #175, opened 2021-02-04
Fix reply content for pattern subscribers when publish message

ShooterIT:
For pattern subscribers, we should publish` pmessage, pattern, channel and message` for connections when clients publish message.


karelrooted (approved):
LGTM

ErnestoLara (approved):

PR #179, opened 2021-02-07
Implement DEBUG SLEEP command

ShooterIT:
For redis, it needs DEBUG command to debug server. Actually for kvrocks, we also need, currently, I only implement sleep <second> subcommand, it is useful for tests.

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -3729,6 +3729,37 @@ class CommandQuit : public Commander {
   }
 };
 
+class CommandDebug : public Commander {
+ public:
+  CommandDebug() : Commander("debug", -2, false) {}
+  Status Parse(const std::vector<std::string> &args) override {
+    subcommand_ = Util::ToLower(args[1]);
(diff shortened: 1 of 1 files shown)


src/worker.cc
```diff
@@ -142,6 +142,12 @@ void Worker::Stop() {
   }
 }
 
+void Worker::Sleep(uint64_t microsecond) {
+  // For other threads, don't allow to handle new connections during sleeping
```
karelrooted, line 146: look like the lock is not necessarily, when in sleep, current worker won't accept new connection (what's this debug sleep mainly use for ?  online instance usually hava 8 workers, pause one worker and then check something about the current worker thread?)

ShooterIT: Actually, I want to block all threads to stop accepting new connections. In some redis tests, we may hope server don't process any request. As we know, In kvrocks, every thread is separate, but now i realize my way is wrong, i can set only one work thread to implement that. So it is unnecessary, thanks @karelrooted

git-hulk (approved):
LGTM

PR #181, opened 2021-02-07
Plant redis tcl tests into kvrocks

ShooterIT:
As we know, redis tcl tests are very complete and contain a lot of boundary testing, so we carry them into kvrocks to help us find problems and improve quality of our project.

Currently redis tcl covers tests
- [X] string
- [X] list
- [X] hash
- [X] set
- [X] zset
- [X] expire
- [X] geo
- [X] auth
- [X] pubsub
- [X] protocol
- [X] slowlog
- [X] keyspace
- [X] scan
- [X] maxclients
- [X] some introspection commands: monitor, client, quit, info, etc..

Some problems I find, but I think they are not urgent, we can implement them gradually if needed.
**string type**
 - bitmap and string uses different encoding, so in redis string command can operate bitmap but kvrocks can’t
 - set command supports more options in redis, especially from redis 6.2

**zset type**
- zadd command can’t support many advanced option
- zlexcount can’t support many advanced option
- zunionstore/zinterstore command can’t operate set type key but redis supports

**geo**
- can’t support box range that introduced from redis 6.2
- of course, also can’t some advanced options

**slowlog**
 - In lastet redis version, slowlog get command also return client id and name info

In addition, some tests currently can’t be run, but I don’t remove just comment them, since I think maybe we also support such feature later and we can use these test cases directly.

For replication tests, because kvrocks uses `listen port + 1` as replication port, so redis tcl tests can’t support. Maybe we don’t use another port for replication, I will open an issue to discuss that later.

Cheers!


git-hulk:
awesome!!!

tests/tcl/tests/assets/default.conf
```diff
@@ -0,0 +1,14 @@
+bind 0.0.0.0
+timeout 0
+workers 1
```
ShooterIT, line 3: hi @karelrooted here, only one worker thread for tests

karelrooted (approved):
LGTM

git-hulk (approved):

PR #188, opened 2021-02-19
Handle the query starting with \r\n

ShooterIT:
I found tcl test sometimes may be time out
https://github.com/bitleak/kvrocks/runs/1932218223?check_suite_focus=true#step:8:569
because kvrocks may not continue to process clients' request if last query starting with \r\n.

```tcl
test "Handle an empty query" {
    reconnect
    r write "\r\n"
    r flush
    assert_equal "PONG" [r ping]
}
```
Why does this test sometimes timeout? i guess that kvrocks doesn't process the query after `r flush`, or the `ping` query will start with  \r\n

### DIFF
--- a/src/redis_request.cc
+++ b/src/redis_request.cc
@@ -28,7 +28,10 @@ Status Request::Tokenize(evbuffer *input) {
           if (pipeline_size > 128) {
             LOG(INFO) << "Large pipeline detected: " << pipeline_size;
           }
-          if (line) continue;
+          if (line) {
+            free(line);
+            continue;
+          }
           return Status::OK();
         }
         pipeline_size++;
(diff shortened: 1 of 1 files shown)


src/redis_request.cc
```diff
@@ -28,6 +28,7 @@ Status Request::Tokenize(evbuffer *input) {
           if (pipeline_size > 128) {
             LOG(INFO) << "Large pipeline detected: " << pipeline_size;
           }
+          if (line) continue;
```
ShooterIT, line 31:
```suggestion
          if (line) {
            free(line);
            continue;
          }
```

ShooterIT (commented):
Hi @karelrooted @git-hulk From libevnet code, if `input` only include `\r\n`, `evbuffer_readln` also malloc 1 byte memory to store `\0`, so we need to free `line`.
https://github.com/libevent/libevent/blob/852af060af7c7e439f0db97829ec74bedeba64d5/buffer.c#L1720-L1731

karelrooted (dismissed):
LGTM

src/redis_request.cc
git-hulk: yeah, we should free the line here if not null.

git-hulk (approved):
LGTM

PR #189, opened 2021-02-20
Enable the socket option tcp keepalive and no delay by default

git-hulk:
TCP keepalive was used to detect whether a TCP connection died or not. TCP no-delay option was used to disable the Nagle algorithm which occurs delay ACK(default in Linux was 40ms) may cause high latency.

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -70,6 +70,20 @@ const std::string Float2String(double d) {
   return buf;
 }
 
+Status SockSetTcpNoDelay(int fd, int val) {
+  if (setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, &val, sizeof(val)) == -1) {
+    return Status(Status::NotOK, strerror(errno));
+  }
+  return Status::OK();
+}
+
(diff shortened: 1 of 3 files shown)


ShooterIT:
Hi @git-hulk For TCP keepalive option, i think it is not enough, because 'keepalive' is set to 7200000ms (2 hours) in linux by default, that is too long to find problem, so we also need to set `tcp_keepalive_intvl`, `tcp_keepalive_probes`. FYI,  how to do in redis https://github.com/redis/redis/blob/unstable/src/anet.c#L120.

We can add a config, but if you want to solve it easily, you can set a default value, in my experience, it is a good choice to limit total detection time into 120s.

Actually, i also want to submit a PR to optimize it 😃

git-hulk:
@ShooterIT yup, you're right that we should set probe interval as well, my original intent was to enable the TCP no-delay by default. Can you help to submit a PR to set the tcp-keepalive interval and probes?

karelrooted (approved):
LGTM

ShooterIT:
@git-hulk OK, i will optimize it later

PR #191, opened 2021-02-20
Support to set max total wal size for rocksdb

ShooterIT:
In kvrocks, we use multiple column families to store metadata, subkeys, etc.
If users always use string type, but use list, hash and other complex data types
infrequently, there will be a lot of old WALs if we don't set size limit
(0 by default in rocksdb), because rocksdb will dynamically choose the WAL size
limit to be [sum of all write_buffer_size * max_write_buffer_number] * 4 if set to 0.

For max_total_wal_size usage
> // Once write-ahead logs exceed this size, we will start forcing the flush of
>   // column families whose memtables are backed by the oldest live WAL file
>   // (i.e. the ones that are causing all the space amplification). If set to 0
>   // (default), we will dynamically choose the WAL size limit to be
>   // [sum of all write_buffer_size * max_write_buffer_number] * 4
>   // This option takes effect only when there are more than one column family as
>   // otherwise the wal size is dictated by the write_buffer_size.
>   //
>   // Default: 0
>   //
>   // Dynamically changeable through SetDBOptions() API.

Related
https://github.com/facebook/rocksdb/issues/5789
https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log#dboptionsmax_total_wal_size

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -320,6 +320,27 @@ rocksdb.max_sub_compactions 2
 # Defalut: 1 Hours
 rocksdb.wal_ttl_seconds 3600
 
+# In order to limit the size of WALs, RocksDB uses DBOptions::max_total_wal_size
+# as the trigger of column family flush. Once WALs exceed this size, RocksDB
+# will start forcing the flush of column families to allow deletion of some
+# oldest WALs. This config can be useful when column families are updated at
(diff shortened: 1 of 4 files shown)


git-hulk (approved):
LGTM, really cool! 💯

PR #193, opened 2021-02-23
Fix using the line as error msg which has freed

git-hulk:


src/replication.cc
```diff
@@ -750,13 +751,15 @@ Status ReplicationThread::fetchFile(int sock_fd, std::string path,
...
         evbuffer_free(evbuf);
-        return Status(Status::NotOK, std::string("read sst file data, err: ")+strerror(errno));
+        return Status(Status::NotOK, std::string("read sst file: ")+strerror(errno));
       }
     }
   }
   if (crc != tmp_crc) {
     evbuffer_free(evbuf);
-    return Status(Status::NotOK, "CRC mismatch");
+    char err_buf[64];
```
ShooterIT, line 760: 64 is enough, but just enough 😃

ShooterIT (approved):

src/replication.cc
git-hulk: yeah, the max error msg should 58 chars

karelrooted (approved):
LGTM

PR #194, opened 2021-02-24
Feature/add flushbackup command

karelrooted:


src/storage.cc
```diff
@@ -315,22 +315,29 @@ Status Storage::RestoreFromBackup() {
...
 void Storage::PurgeOldBackups(uint32_t num_backups_to_keep, uint32_t backup_max_keep_hours) {
+  time_t now = time(nullptr);
   std::vector<rocksdb::BackupInfo> backup_infos;
   backup_mu_.lock();
   backup_->GetBackupInfo(&backup_infos);
   if (backup_infos.size() > num_backups_to_keep) {
     uint32_t num_backups_to_purge = static_cast<uint32_t>(backup_infos.size()) - num_backups_to_keep;
-    LOG(INFO) << "[storage] Going to purge " << num_backups_to_purge << " old backups";
     for (uint32_t i = 0; i < num_backups_to_purge; i++) {
+      // we should't purge the backup if the backup is not expired (for replication stability)
```
git-hulk, line 325:
```suggestion
      // we shouldn't purge the backup if the backup is not expired,
      // while multi slaves may create more replications than the `num_backups_to_keep`
```

git-hulk (dismissed):
we should change the configuration description as well since the behavior has changed

ShooterIT (approved):

kvrocks.conf
```diff
@@ -164,6 +164,7 @@ max-db-size 0
 # The maximum backup to keep, server cron would run every minutes to check the num of current
 # backup, and purge the old backup if exceed the max backup num to keep. If max-backup-to-keep
 # is 0, no backup would be keep.
+# exception: The backup will not be purged if it's not yet expired (when max-backup-keep-hours is not 0)
```
git-hulk, line 167:
```suggestion
# exception: the backup will not be purged if it's not yet expired (when max-backup-keep-hours is not 0)
```

git-hulk (approved):
LGTM

PR #200, opened 2021-03-09
Don't use extra port for replication

ShooterIT:
As we know, in order to implement replication, kvrocks has specified replication threads  that use 'port + 1' as listening port, that may make somethings easy, but it is not convenient to deploy and maintain.

In this PR, kvrocks doesn't use extra port for replication instead of processing replication requests in general work threads. 

Some problems we need to care
**Blocking**
We can't block work threads for processing replication requests. Currently, for replication commands, `_fetch_meta` and `_fetch_file` are blocked, in this commit, we process both commands in new threads.

**Compatibility**
We provide a configuration item `master-use-repl-port` to make new version able to replicate old version master. If the masters are old version, you should set `master-use-repl-port` to yes, otherwise you should set it to no.

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -4106,13 +4106,13 @@ class CommandFetchFile : public Commander {
   CommandFetchFile() : Commander("_fetch_file", 2, false) {}
 
   Status Parse(const std::vector<std::string> &args) override {
-    paths_sts_ = args[1];
+    files_str_ = args[1];
     return Status::OK();
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    std::vector<std::string> paths;
-    Util::Split(paths_sts_, ",", &paths);
+    std::vector<std::string> files;
+    Util::Split(files_str_, ",", &files);
 
     int repl_fd = conn->GetFD();
     std::string ip = conn->GetIP();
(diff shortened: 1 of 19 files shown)


src/redis_cmd.cc
```diff
@@ -4058,22 +4058,42 @@ class CommandFetchMeta : public Commander {
...
+    // Feed-replica-meta thread
+    std::thread t = std::thread([svr, repl_fd, ip]() {
+      Util::ThreadSetName("feed-replica-meta");
+      int fd;
+      uint64_t file_size;
+      rocksdb::BackupID meta_id;
+      auto s = Engine::Storage::BackupManager::OpenLatestMeta(
+          svr->storage_, &fd, &meta_id, &file_size);
+      if (!s.IsOK()) {
+        close(repl_fd);
```
karelrooted: if we close without sending error msg to replica, the replica can't log the reason why master close the connection, seems like with a error msg will be better

ShooterIT: Yes, thanks

karelrooted (dismissed):
LGTM

src/redis_cmd.cc
```diff
@@ -4713,19 +4780,11 @@ std::map<std::string, CommanderFactory> repl_command_table = {
 Status LookupCommand(const std::string &cmd_name,
```
git-hulk, line 4780: remove the `is_repl` while it's used?

src/redis_cmd.cc
```diff
@@ -4083,22 +4106,70 @@ class CommandFetchFile : public Commander {
...
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    uint64_t file_size = 0;
-    auto fd = Engine::Storage::BackupManager::OpenDataFile(svr->storage_, path_,
-                                                           &file_size);
-    if (fd < 0) return Status(Status::DBBackupFileErr);
-    conn->Reply(std::to_string(file_size) + CRLF);
-    conn->SendFile(fd);
+    std::vector<std::string> paths;
```
git-hulk: seems `files` was more precious than `paths`, as well as in replication.  But `paths` also ok to me. : )

git-hulk (commented):
Awesome!

git-hulk (approved):

ShooterIT:
hi @git-hulk @karelrooted  Currently i just implement basic function. we still need to test its performance although it doesn't have damage.
We still have some other jobs for this change, such as clean up jobs, replication connection identifier, ... But i don't want to make one PR too complicated to review, it is hard to find problems. We can fix these in other PRs.

git-hulk:
It should be fine to merge the PR since the master was defined as unstable now. Also, those problems weren't imported by this PR so makes a lot of sense to file a new PR.

PR #204, opened 2021-03-15
Fix incorrect used_db_size value 

karelrooted:
Fix incorrect used_db_size value which always uses the same cf to find the key range

GetApproximateSizes needs the start/end key of the corresponding column family.
But we always use the metadata column family to find the start/end key range.

### DIFF
--- a/src/redis_db.h
+++ b/src/redis_db.h
@@ -31,7 +31,10 @@ class Database {
                        std::vector<std::string> *keys);
   rocksdb::Status RandomKey(const std::string &cursor, std::string *key);
   void AppendNamespacePrefix(const Slice &user_key, std::string *output);
-  rocksdb::Status FindKeyRangeWithPrefix(const std::string &prefix, std::string *begin, std::string *end);
+  rocksdb::Status FindKeyRangeWithPrefix(const std::string &prefix,
+                                         std::string *begin,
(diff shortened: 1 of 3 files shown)


src/redis_db.h
```diff
@@ -31,7 +31,10 @@ class Database {
                        std::vector<std::string> *keys);
   rocksdb::Status RandomKey(const std::string &cursor, std::string *key);
   void AppendNamespacePrefix(const Slice &user_key, std::string *output);
-  rocksdb::Status FindKeyRangeWithPrefix(const std::string &prefix, std::string *begin, std::string *end);
+  rocksdb::Status FindKeyRangeWithPrefix(const std::string &prefix,
```
git-hulk, line 34: How about moving the `rocksdb::ColumnFamilyHandle` as the first param to keep the consistency with others

git-hulk:
Seems the commit message didn't match the PR changes.  It should be:

```
Fix incorrect used_db_size value which always uses the same cf to find the key range

GetApproximateSizes needs the start/end key of the corresponding column family.
But we always use the metadata column family to find the start/end key range.
```

git-hulk (approved):

ShooterIT (approved):

git-hulk (approved):

PR #205, opened 2021-03-18
Use checkpoint to implement data snapshot when full replication

ShooterIT:
# Motivation
Currently, master creates a new snapshot by `backup` function of rocksdb when replicas ask for full replication. Although, rocksdb `backup` function can support incremental backup, it still cost much time, disk bandwidth,  and disk space if there is no based backup or current backup exists for long time, and `backup` function really needs copy real files. If kvrocks holds a huge amount of data, copying data files would influence the writing or reading of rocksdb that results in latency of user request.

As we know, the function `checkpoint` of rocksdb would hard link data files instead of copying files, that has minor resource consumption, so we wish using checkpoint to implement data snapshot when full replication to avoid influencing user requests.

# Solution
### Master
- As similar with full replication by using backup, master will try to create a new checkpoint if there is no checkpoint when replicas ask full replication, and tell replicas all files name list of checkpoint.
- We can create a new checkpoint for every replica, but there will be many checkpoints that may occupy disk space because hard link could make deleted files existing during copying file period. So we hope to use one shared checkpoint, but replicas still wait for next checkpoint if current checkpoint is too old to satisfy feeding new coming replicas.
-  How to clean checkpoint, it is much difficult because replicas fetch data info and data files by different connections and threads. So i use current fetching files threads number as checkpoint reference count, master will delete checkpoint if there no thread to fetch files during several seconds. To avoid that a slow replica always keep the checkpoint alive too long time, master still delete the checkpoint if it is very long time (1 day) from creating checkpoint.

### Replica
- Replicas still fetch checkpoint files like before, after fetching all files, replicas will destroy old db, rename synced checkpoint dir to db dir and reload this checkpoint. To avoid loading wrong checkpoint causes db broken, we rename old db dir, and restore it if the synced checkpoint is corrupt to fail to open.
- How to resume broken transfer based files. We store synced files into a specified dir (synced_checkpoint). Firstly, because we don't get file CRC by checkpoint to identify unique files, so we must clean the synced_checkpoint dir when changing master. Secondly, we need to clean up invalid files of master old checkpoint when we check existing files of master new checkpoint for skipping some transferred files.

# Compatibility
As #200, kvrocks still support to replicate data from old version master if you set `master-use-repl-port` to yes. For detail, we still use old implementation to coping from old version master.

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -606,12 +606,12 @@ ReplicationThread::CBState ReplicationThread::fullSyncReadCB(bufferevent *bev,
         auto s = Engine::Storage::ReplDataManager::CleanInvalidFiles(
             self->storage_, target_dir, need_files);
         if (!s.IsOK()) {
-          LOG(WARNING) << "[replication] Fail to clean up invalid files of old checkpoint,"
+          LOG(WARNING) << "[replication] Failed to clean up invalid files of the old checkpoint,"
                        << " error: " << s.Msg();
           LOG(WARNING) << "[replication] Try to clean all checkpoint files";
           auto s = rocksdb::DestroyDB(target_dir, rocksdb::Options());
           if (!s.ok()) {
-            LOG(WARNING) << "[replication] Fail to clean all checkpoint files, error: "
+            LOG(WARNING) << "[replication] Failed to clean all checkpoint files, error: "
                          << s.ToString();
           }
         }
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -277,6 +277,12 @@ Status ReplicationThread::Start(std::function<void()> &&pre_fullsync_cb,
   pre_fullsync_cb_ = std::move(pre_fullsync_cb);
   post_fullsync_cb_ = std::move(post_fullsync_cb);
 
+  // Clean synced checkpoint fronm old master because replica starts to flow new master
+  auto s = rocksdb::DestroyDB(srv_->GetConfig()->sync_checkpoint_dir, rocksdb::Options());
+  if (!s.ok()) {
+    LOG(WARNING) << "Can't clean synced checkpoint from master, error: " << s.ToString();
+  }
+
   // cleanup the old backups, so we can start replication in a clean state
   storage_->PurgeOldBackups(0, 0);
 
@@ -532,6 +538,11 @@ ReplicationThread::CBState ReplicationThread::fullSyncReadCB(bufferevent *bev,
   auto input = bufferevent_get_input(bev);
   switch (self->fullsync_state_) {
     case kFetchMetaID:
+      // New version master only sends meta file content
+      if (!self->srv_->GetConfig()->master_use_repl_port) {
+        self->fullsync_state_ = kFetchMetaContent;
+        return CBState::AGAIN;
+      }
       line = evbuffer_readln(input, &line_len, EVBUFFER_EOL_CRLF_STRICT);
       if (!line) return CBState::AGAIN;
       if (line[0] == '-') {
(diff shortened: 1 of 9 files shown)


src/replication.cc
```diff
@@ -277,6 +277,12 @@ Status ReplicationThread::Start(std::function<void()> &&pre_fullsync_cb,
   pre_fullsync_cb_ = std::move(pre_fullsync_cb);
   post_fullsync_cb_ = std::move(post_fullsync_cb);
 
+  // Clean synced checkpoint fronm old master because replica starts to follow new master
```
karelrooted: typo: fronm  => from

src/storage.cc
```diff
@@ -314,6 +318,62 @@ Status Storage::RestoreFromBackup() {
   return s.ok() ? Status::OK() : Status(Status::DBBackupErr, s.ToString());
 }
 
+Status Storage::Reopen() {
+  auto s = Open();
+  if (!s.IsOK()) {
+    LOG(ERROR) << "[storage] Fail to reopen db, error: " << s.Msg();
+    LOG(ERROR) << "[storage] Exiting...";
+    exit(1);
```
karelrooted: the old backup method will restart fullsync in this situation,  restart seems  better?

karelrooted (dismissed):
LGTM

src/storage.cc
ShooterIT: yes, I will fix

src/replication.cc
```diff
@@ -566,17 +577,51 @@ ReplicationThread::CBState ReplicationThread::fullSyncReadCB(bufferevent *bev,
...
+        Util::Split(std::string(line), ",", &need_files);
+        for (auto f : need_files) {
+          meta.files.emplace_back(f, 0);
+        }
+        target_dir = self->srv_->GetConfig()->sync_checkpoint_dir;
+        // Clean invaild files of checkpoint
+        auto s = Engine::Storage::ReplDataManager::CleanInvalidFiles(
+            self->storage_, target_dir, need_files);
+        if (!s.IsOK()) {
+          LOG(WARNING) << "[replication] Fail to clean up invalid files of old checkpoint,"
```
git-hulk:
```suggestion
          LOG(WARNING) << "[replication] Failed to clean up invalid files of the old checkpoint,"
```

src/replication.cc
```diff
@@ -566,17 +577,51 @@ ReplicationThread::CBState ReplicationThread::fullSyncReadCB(bufferevent *bev,
...
+                       << " error: " << s.Msg();
+          LOG(WARNING) << "[replication] Try to clean all checkpoint files";
+          auto s = rocksdb::DestroyDB(target_dir, rocksdb::Options());
+          if (!s.ok()) {
+            LOG(WARNING) << "[replication] Fail to clean all checkpoint files, error: "
```
git-hulk:
```suggestion
            LOG(WARNING) << "[replication] Failed to clean all checkpoint files, error: "
```

git-hulk (dismissed):
Really good implementation! LGTM except `Fail` => `Failed` in error message.

Another potential optimization was in `GetFullReplDataInfo` when multi-slaves were in full sync at the same time. 

For example:

A slave comes at first and creating the checkpoint, then B slave follows closely, it would get an error if the new checkpoint was creating. The way to fix was by waiting for the new checkpoint to finish. 

```
lock();
if (!sync_dir.IsExists() || checkpoint.Expired()) {
    CreateCheckpoint();
}
unlock();

if (checkpoint.Expired()) {
   return error;
}
Get Replication Data Info
```
But the current implementation was fine to me that B slave would restart full sync again.

git-hulk (dismissed):

tests/tcl/tests/integration/replication.tcl
```diff
@@ -94,3 +92,36 @@ start_server {tags {"repl"}} {
...
+                    [string match {*connected*} [$A role]] &&
+                    [string match {*connected*} [$B role]]
+                } else {
+                     fail "Slaves can't sync with master"
+                }
+                # Only 2 full sync
+                assert_equal 2 [s sync_full]
+            }
+        }
+
```
ShooterIT:
```suggestion

```

git-hulk (approved):
LGTM, cheers!!!

karelrooted (approved):

PR #207, opened 2021-03-22
Fix compression algorithm typo

ShooterIT:
Another related question
For compression algorithm, currently, we only support `snappy`, as we know, `zstd` has great performance，high speed and compression ratio. I think we should support multiple compression algorithms, WDYT?


git-hulk (approved):

git-hulk:
@ShooterIT You're right that we should support multiple compression algorithms

PR #211, opened 2021-03-25
Fix the SST file creation time maybe 0 in some unknown conditions

git-hulk:
we use the file creation time to force compact the SST file which
created over 2 days, so fallback to file modification time here to
prevent repeatedly compacting the same SST file if the file creation
time was zero in some unknown conditions.


karelrooted (approved):
LGTM

ShooterIT (approved):

ShooterIT:
@git-hulk Is it necessary to back port to 1.3 version? it looks like just a rare case, right?

git-hulk:
> @git-hulk Is it necessary to back port to 1.3 version? it looks like just a rare case, right?

yes, I can't find the place where file_creation_time would zero while it's always assigned on the building table. The rocksdb comment says 0 means unknown, but I can't reproduce it on my side.

I'll backport into v1.3 after merging

PR #212, opened 2021-03-27
 Decrease the default value of the max total wal size to 16GiB

git-hulk:
Currently, the default max total WAL size 100GiB which may keep too many
    WAL files for most scenes, so reduce it to 16GiB to make the disk space
    happy. Also increase the WAL TTL from 1 hour to 3 hours to keep those
    WAL files longer for making the psync possible.


ShooterIT (approved):

karelrooted (approved):
LGTM

PR #221, opened 2021-04-15
Fix get corruption writebatch data  after psync restart

karelrooted:
Fix get corruption writebatch data (wrongly include writebatch size at the header of the writebatch data) after a restart cause by a failed storage->writebatch() , should always set incr_state_ to Incr_batch_size whether storage_->writebatch success or not , so the next incrementBatchLoopCB will start from the right step

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -215,6 +215,7 @@ void ReplicationThread::CallbacksStateMachine::Start() {
     LOG(ERROR) << "[replication] Failed to start state machine, err: " << strerror(errno);
   }
   handler_idx_ = 0;
+  repl_->incr_state_ = Incr_batch_size;
   if (getHandlerEventType(0) == WRITE) {
     SetWriteCB(bev, EvCallback, this);
   } else {


src/replication.cc
```diff
@@ -499,6 +499,7 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(
         // Read bulk data (batch data)
         if (self->incr_bulk_len_+2 <= evbuffer_get_length(input)) {  // We got enough data
           bulk_data = reinterpret_cast<char *>(evbuffer_pullup(input, self->incr_bulk_len_ + 2));
+          self->incr_state_ = Incr_batch_size;
```
git-hulk: maybe we should reset this state in the replication start function? or it may confuse why MUST reset the state before write batch

git-hulk (dismissed):

git-hulk:
Many thanks to @karelrooted try hard to fix this hidden deep issue.

git-hulk (approved):

ShooterIT (approved):

ShooterIT:
@karelrooted I think we need to port back to 1.3 branch

karelrooted:
> @karelrooted I think we need to port back to 1.3 branch

👌

PR #222, opened 2021-04-17
Build and push docker image when tagging

git-hulk:
Docker was an easy way for users to play the kvrocks without compiling the source code,
so let's built and release the docker image when the new tag was published.

It's hard to review the workflow modification for reviewers, but take it easy that I have tested it on my side.

### DIFF
--- a/.github/workflows/docker_image.yaml
+++ b/.github/workflows/docker_image.yaml
@@ -1,0 +1,42 @@
+name: build docker image
+
+on:
+  push:
+    tags:
+      - v* 
+
+jobs:
+  build-docker-image:
+    name: Build And Push Docker Image
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


karelrooted (approved):
LGTM

.github/workflows/docker_image.yaml
```diff
@@ -0,0 +1,42 @@
+name: build docker image
+
+on:
+  push:
```
ShooterIT, line 4: Why not `pull_request`? Now, we hope that every change should be a PR

ShooterIT (approved):
For action, now we should add 

- daily action like https://github.com/redis/redis/blob/unstable/.github/workflows/daily.yml
- support macOS in action

.github/workflows/docker_image.yaml
git-hulk: It seems that tag can't be a PR? Github would auto push a new version when publishing the new release.

PR #232, opened 2021-04-23
Use checkpoint to implement kvrocks backup

### LINKED ISSUE #220: [QUESTION] Should we use backup or checkpoint of rocksdb for kvrocks backup
Currently, we use checkpoint of rocksdb to implement kvrocks full synchronization, but kvrocks backup still uses backup of rocksdb. As we know, rocksdb backup costs much bandwidth and space of disk, but rocksdb backup can implement incremental backup and support to store into HDFS, these are good features. So i think we should make a decision to use  backup or checkpoint of rocksdb for kvrocks backup. WDYT? @git-hulk @karelrooted @Alfejik

ShooterIT:
resolves #220

For incremental backup, please look https://github.com/bitleak/kvrocks/wiki/How-to-backup

### DIFF
--- a/src/storage.cc
+++ b/src/storage.cc
@@ -42,6 +42,7 @@ Storage::Storage(Config *config)
   SetCreatingCheckpoint(false);
   SetCheckpointCreateTime(0);
   SetCheckpointAccessTime(0);
+  backup_creating_time_ = std::time(nullptr);
 }
 
 Storage::~Storage() {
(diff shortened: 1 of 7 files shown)


git-hulk (approved):

src/storage.cc
```diff
@@ -276,6 +297,12 @@ Status Storage::RestoreFromBackup() {
     LOG(ERROR) << "[storage] Failed to reopen db: " << s2.Msg();
     return Status(Status::DBOpenErr, s2.Msg());
   }
+
+  // Clean up backup engine
+  backup_->PurgeOldBackups(0);
```
karelrooted, line 302: the backup_ is not being  initialized under the new code

ShooterIT: In line 284, we open it
that just is compatible with full synchronization with old version kvrocks. Compatibility makes me headache 😭

karelrooted (approved):
LGTM

PR #233, opened 2021-04-23
Replicas can empty db before full synchronization

ShooterIT:
To guarantee slave's data safe and serve when it is in full synchronization
state, slave still keep itself data. But this way needs to occupy much disk
space, so we provide a way to reduce disk occupation, slave will delete itself
entire database before fetching files from master during full synchronization.
If you want to enable this way, you can set 'slave-delete-db-before-fullsync'
to yes, but you must know that database will be lost if master is down during
full synchronization, unless you have a backup of database.

This option is similar redis replicas RDB diskless load option: `repl-diskless-load on-empty-db`

Other changes:
 - change `is_loading_` and `stop_` to atomic variable
 - sleep 1 ms when waiting for finishing executing commands

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -151,6 +151,20 @@ master-use-repl-port no
 #
 slave-serve-stale-data yes
 
+# To guarantee slave's data safe and serve when it is in full synchronization
+# state, slave still keep itself data. But this way needs to occupy much disk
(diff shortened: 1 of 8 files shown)


karelrooted (approved):
LGTM

git-hulk (approved):

PR #236, opened 2021-04-27
Fix uninitialized Rocksdb.level0_stop_writes_trigger

ShooterIT:


karelrooted (dismissed):

git-hulk (dismissed):

ShooterIT:
Hi, i add the support to config level0_stop_writes_trigger, because we have this variable in `config.h`, moreover, we shouldn't  bind them together, they are changed respectively for tuning.

src/config.cc
```diff
@@ -288,6 +282,8 @@ void Config::initFieldCallback() {
       {"rocksdb.max_background_compactions", set_db_option_cb},
       {"rocksdb.max_background_flushes", set_db_option_cb},
       {"rocksdb.compaction_readahead_size", set_db_option_cb},
+      {"rocksdb.level0_slowdown_writes_trigger", set_db_option_cb},
```
karelrooted: level0_slowdown_writes_trigger and level0_stop_writes_trigger is a column family option, can not use setdboption

karelrooted (changes requested):

src/config.cc
ShooterIT: oh, thanks, i found unit test failed

karelrooted (approved):

git-hulk (approved):

ShooterIT:
@Mergifyio backport 1.3

PR #238, opened 2021-04-30
Fix slave can't execute publish command

karelrooted:


karelrooted:
do we need a new way to implement pub/sub without rely on wal replication ?   the current solution write to db  cause  wal seq to increase (have some feedback from user that it sometimes cause a fullsync when a redis-sentinel failover is trigged, cause by "publish __sentinel:hello" message in new slave)

git-hulk:
It seems fine that don’t write the WAL when the role was slave. But it would be better to propagate it without writing to WAL.

git-hulk (approved):

ShooterIT (approved):

ShooterIT:
I feel ok for current solution. But i have a question, why we persist publish messages？

git-hulk:
We want to use the WAL to help us propagating the publish command like others.

karelrooted:
@Mergifyio backport 1.3

karelrooted:
@Mergifyio backport 2.0

PR #246, opened 2021-05-05
Fix condition race for ReclaimOldDBPtr

ShooterIT:
I change `excuting_command_num_` memory order to `std::memory_order_seq_cst`, even `loading` can establish inter-thread sync, but i think it is not error-prone if atomic variable has sync semantic cross multi threads.

### DIFF
--- a/src/redis_request.cc
+++ b/src/redis_request.cc
@@ -200,6 +200,14 @@ void Request::ExecuteCommands(Connection *conn) {
     auto start = std::chrono::high_resolution_clock::now();
     bool is_profiling = isProfilingEnabled(cmd_name);
     svr_->IncrExecutingCommandNum();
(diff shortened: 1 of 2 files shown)


git-hulk (approved):
LGTM

ShooterIT:
Actually, i don't satisfied with my PR, so it is a draft, for `ReclaimOldDBPtr` implementation, maybe `std::shared_mutex` in c++17 is our wanted way, but C++17 is too new, and i also don't want to implement read write lock

git-hulk:
yeah, another alternative was the boost `shared mutex` for C++ 11, which we have considered before and thought too heavy for us to import the boost library. I'm not sure whether have holes if we implement a simple one.

ShooterIT:
@Mergifyio backport 1.3

PR #250, opened 2021-05-07
Move command attributes into redisCommand

git-hulk:


src/redis_cmd.cc
```diff
@@ -4194,184 +4047,184 @@ struct redisCommand {
...
-    ADD_CMD("replconf", 0, 0, 0, CommandReplConf),
-    ADD_CMD("psync", 0, 0, 0, CommandPSync),
-    ADD_CMD("_fetch_meta", 0, 0, 0, CommandFetchMeta),
-    ADD_CMD("_fetch_file", 0, 0, 0, CommandFetchFile),
-    ADD_CMD("_db_name", 0, 0, 0, CommandDBName),
+    ADD_CMD("replconf", -3, false, 0, 0, 0, CommandReplConf),
+    ADD_CMD("psync", 2, false, 0, 0, 0, CommandPSync),
+    ADD_CMD("_fetch_meta", 1, false, 0, 0, 0, CommandFetchMeta),
+    ADD_CMD("_fetch_file", 2, false, 0, 0, 0, CommandFetchFile),
+    ADD_CMD("_db_name", 1, false, 0, 0, 0, CommandDBName),
```
ShooterIT, line 4201: for `is_write`, we can use one description，such as "write, ok-loading, access-db, lock-db", and we parse it generate its flags, it is much more readable.

src/redis_cmd.cc
```diff
@@ -4388,7 +4241,11 @@ Status LookupCommand(const std::string &cmd_name,
   if (cmd_factory == commands.end()) {
     return Status(Status::RedisUnknownCmd);
   }
-  *cmd = cmd_factory->second.factory();
+  auto redisCmd = cmd_factory->second;
```
ShooterIT: `cmd_factory` is not a fine variable name currently.

src/redis_cmd.cc
```diff
@@ -4388,7 +4241,11 @@ Status LookupCommand(const std::string &cmd_name,
...
+  *cmd = redisCmd.factory();
+  (*cmd)->SetArity(redisCmd.arity);
+  (*cmd)->SetIsWrite(redisCmd.is_write);
+  (*cmd)->SetName(redisCmd.name);
```
ShooterIT: for `arity, is_write, or others`, i think they are attributes of `RedisCommand` instead of `Commander`, i think commander just like a executor. so could we set one member to refer `RedisCommand`, the member function `IsWrite` can access `RedisCommand` to its attributes. we needn't to always add members into commander if we add new flags

src/redis_cmd.cc
git-hulk: good catch! any advice? how about cmd_iter

ShooterIT (commented):
So quickly !!! 👍

src/redis_cmd.cc
ShooterIT: `cmd_iter` is ok to me
How about changing `LookupCommand` to `LookupAndCreateCommander`

src/redis_cmd.cc
git-hulk: nice advice, but seems `RedisCommand` wasn't a good name if we want to make it as the `Commader` member. We need a better name like `CommandAttributes`, how do you think.

ShooterIT: +10086

src/redis_cmd.cc
git-hulk: yeah, it's more readable with literal but I think can improve it at the next PR, to keep this PR more simple and clear by just move command attributes together.

PR #254, opened 2021-05-08
Fix returning wrong string range in getrange command

karelrooted:


git-hulk (approved):

karelrooted:
@Mergifyio backport 1.3

ShooterIT:
I changed the title since it is truly a bug not only typo

PR #260, opened 2021-05-12
Fix skipping wrong CURRENT file when full synchronization

ShooterIT:
Currently, resuming broken transfer based files when full synchronization may skip CURRENT file of RocksDB, but CURRENT doesn't have number, so we don't distinguish CURRENT is valid or not. If we keep old CURRENT file, slave will fail to restore checkpoint and full sync again, i.e. resuming broken transfer based files is not available in this case. In this commit, we never skip CURRENT file.

New test cases 
- resume broken transfer based files
- share one checkpoint when two slaves full sync with master at the same time.

I always want to add some test cases for replication, eventually i did, actually, i found this bug just because i write these test cases.

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -6,6 +6,7 @@
 #include <future>
 #include <string>
 #include <thread>
+#include <algorithm>
 #include <event2/buffer.h>
 #include <event2/bufferevent.h>
 #include <event2/event.h>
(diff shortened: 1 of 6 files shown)


git-hulk:
cool, do you mean that got a stale manifest number when adding test cases?

ShooterIT:
Yes, CURRENT file points out a wrong MANIFEST file

> W0512 16:00:19.617869 77352 storage.cc:388] [storage] Fail to open master checkpoint, error: IO error: No such file or directoryWhile opening a file for sequentially reading: ./tests/tmp/server.76661.3/db/MANIFEST-000008: No such file or directory

git-hulk (approved):
Great! 🎊

karelrooted (approved):

PR #271, opened 2021-05-19
Allow to use the dynamic libraries to make the building faster

git-hulk:


karelrooted (approved):

ShooterIT:
So now should we install dynamic libraries firstly and then make with 'NO_BUILD_DEPENDENCIES'? @git-hulk

git-hulk:
> So now should we install dynamic libraries firstly and then make with 'NO_BUILD_DEPENDENCIES'? @git-hulk

Right, 'NO_BUILD_DEPENDENCIES' means we should install all dependencies first before building.

ShooterIT:
Maybe this way should be default way in CI, or one job

ShooterIT:
Because now we couldn't  know it works well

git-hulk:
Good advice, but we should build from source at Linux and macOS once, so maybe it's a good way to build multi-version Linux, and only keep the latest version was built from source.

src/worker.cc
```diff
@@ -77,7 +77,11 @@ void Worker::newConnection(evconnlistener *listener, evutil_socket_t fd,
     return;
   }
   event_base *base = evconnlistener_get_base(listener);
+#ifndef DISABLE_USE_PTHREADS
   auto evThreadSafeFlags = BEV_OPT_THREADSAFE | BEV_OPT_DEFER_CALLBACKS | BEV_OPT_UNLOCK_CALLBACKS;
```
ShooterIT, line 81: @git-hulk I am not  familiar with `BEV_OPT_THREADSAFE`, AFAIK, one thread may operate another thread's event buffer, so, is it ok?

karelrooted: https://github.com/bitleak/kvrocks/pull/94 (FIX: add BEV_OPT_THREADSAFE option (monitor cmd will coredump when evbuffer_add and evbuffer_drain concurrently in multi thread))

PR #272, opened 2021-05-20
Add rename-command config

karelrooted:


src/config.cc
```diff
@@ -266,6 +266,27 @@ void Config::initFieldCallback() {
...
         return Status::OK();
       }},
+      {"rename-command", [this](Server *srv, const std::string &k, const std::string &v) -> Status {
+        if (v.empty()) {
+          return Status::OK();
+        }
+        std::vector<std::string> args;
+        Util::Split(v, " \t", &args);
+        if (args.size() != 2 || args[0] == args[1]) {
+          return Status(Status::NotOK, "invalid rename-command format");
```
ShooterIT:
```suggestion
          return Status(Status::NotOK, "Invalid rename-command format");
```

src/config.h
```diff
@@ -83,6 +83,8 @@ struct Config{
   Cron bgsave_cron;
   CompactionCheckerRange compaction_checker_range{-1, -1};
   std::map<std::string, std::string> tokens;
+  std::string rename_command_;
+  std::string rename_command_new_name_;
```
git-hulk: Should we allow renaming more than one command?

karelrooted: redis only allow rename one command ,  kvrocks allow rename multi command ?

git-hulk: Redis allows renaming multi commands indeed

git-hulk (dismissed):

src/config.cc
```diff
@@ -175,6 +176,27 @@ void Config::initFieldValidator() {
         compaction_checker_range.Stop = stop;
         return Status::OK();
       }},
+      {"rename-command", [](const std::string &k, const std::string &v) -> Status {
+        std::vector<std::string> args;
+        Util::Split(v, " \t", &args);
+        if (args.size() != 2 || args[0] == args[1]) {
```
ShooterIT: I don't think it is an error if renamed name is the same as origin name

kvrocks.conf
```diff
@@ -270,6 +270,22 @@ compaction-checker-range 0-7
 # e.g. bgsave-cron 0 3 * * * 0 4 * * *
 # would bgsave the db at 3am and 4am everyday
 
+# Command renaming.
+#
+# It is possible to change the name of dangerous commands in a shared
+# environment. For instance the Keys command may be renamed into something
```
ShooterIT:
```suggestion
# environment. For instance the KEYS command may be renamed into something
```
just the same as the following

src/config.cc
```diff
@@ -240,7 +260,8 @@ void Config::initFieldCallback() {
           if (cmd == "*") {
             profiling_sample_all_commands = true;
             profiling_sample_commands.clear();
-          } else if (srv->IsCommandExists(cmd)) {
+          } else if (Redis::IsCommandExists(cmd)) {
```
ShooterIT, line 263: So we must ask users to use origin name, kvrocks won't profile renamed command?

ShooterIT:
@karelrooted And could you supplement some tcl tests

src/config.cc
karelrooted: redis_request.cc use isProfilingEnabled(original_cmd_name) (line 205), so profile_sample_commands need to use original cmd name

ShooterIT (dismissed):

git-hulk (dismissed):

karelrooted:
> @karelrooted And could you supplement some tcl tests

add a rename one command testcase,  the current code base will not allow a multi rename-command directive,  so rename-two-command  testcase is not included

git-hulk (approved):

ShooterIT (approved):

PR #278, opened 2021-05-26
Refactor db mutex

ShooterIT:
1.  In `SubKeyFilter`, remove lock to access db, because we call `CancelAllBackgroundWork(wait=true)` when closing db.
2.  Use read write lock to replace incr/decr refs for db, and in IncrDBRefs, we also check closing flag, that doesn't its name.
  now, i think, it is much easy to understand:
     * acquire db read lock firstly and check flag when reading db
     * acquire db write lock, then close db and  set `closing` flag when writing db
     
I think all accesses of db are protected when closing db.

@bitleak/kvrocks-core-team I know this PR is dangerous, even i don't test for that :). You(@git-hulk and @karelrooted ) already costed much energy to find many bugs before, but I still open that, we should push this project forward, so it is much important to keep code easy to understand and modify, actually i have plan to refactor  `ReclaimOldPtr` . So ask you to help me review, maybe you can reproduce old bugs conditions to check it.

from a saboteur

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -50,6 +50,18 @@ Server::~Server() {
   }
 }
 
+// Kvrocks threads list:
+// - Work-thread: process client's connections and requests
+// - Task-runner: one thread pool, handle some jobs that may freeze server if run directly
+// - Cron-thread: server's crontab, clean backups, resize sst and memtable size
+// - Compaction-checker: active compaction according to collected statistics
+// - Replication-thread: replicate incremental stream from master if in slave role, there
+//   are some dynamic threads to fetch files when full sync.
+//     - fetch-file-thread: fetch SST files from master
+// - Feed-slave-thread: feed data to slaves if having slaves, but there also are some dynamic
+//   threads when full sync, TODO(@shooterit) we should manage this threads uniformly.
+//     - feed-replica-data-info: generate checkpoint and send files list when full sync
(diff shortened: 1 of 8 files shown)


src/server.cc
```diff
@@ -71,6 +83,13 @@ Status Server::Start() {
     Util::ThreadSetName("compaction-checker");
     CompactionChecker compaction_checker(this->storage_);
     while (!stop_) {
+      // Sleep first
+      std::this_thread::sleep_for(std::chrono::milliseconds(100));
+
+      // To guarantee accessing DB safely
+      auto guard = storage_->ReadLockGuard();
```
ShooterIT, line 90: @karelrooted lock here

src/server.cc
```diff
@@ -50,6 +50,18 @@ Server::~Server() {
   }
 }
 
+// Kvrocks threads list:
```
ShooterIT, line 53: @karelrooted all threads, please review

karelrooted: Confirmed:  all thread is included in the current thread list

ShooterIT: @karelrooted thanks

karelrooted (dismissed):

karelrooted (approved):

git-hulk (approved):

PR #280, opened 2021-05-28
Remove kvrocksrestore tool

karelrooted:
backup use checkpoint now , kvrocksrestore tool is not needed anymore


ShooterIT:
For me, it is ok, actually, I also want to remove it, but just one concern, old version maybe need this tool?

git-hulk:
It's ok to remove this tool from the kvrocks while few users depend on it, we can offer it again in another repo instead of kvrocks.

git-hulk (approved):

ShooterIT (approved):

karelrooted:
> For me, it is ok, actually, I also want to remove it, but just one concern, old version maybe need this tool?

1.3 branch has this tool

PR #281, opened 2021-05-28
Refactor workers concurrency

ShooterIT:
If the command need to process exclusively, we need to get `WriteLock`
that can guarantee other threads can't come into critical zone, such as DEBUG,
MULTI, LUA (in the immediate future). Otherwise, we just use  `ReadLock`
to allow all workers to execute commands at the same time.

This commit may influence performance a bit.

### DIFF
--- a/src/redis_request.cc
+++ b/src/redis_request.cc
@@ -1,5 +1,6 @@
 #include <chrono>
 #include <utility>
+#include <memory>
 #include <glog/logging.h>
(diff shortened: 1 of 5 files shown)


git-hulk:
It may affect the performance since we use the read lock when executing commands, right? We can have a simple benchmark before merging.

karelrooted (approved):

git-hulk (approved):

PR #282, opened 2021-05-28
Improve robustness for replication tests

ShooterIT:
@karelrooted If last compaction was not finished, the new checkpoint may be different with old checkpoint, all tranferred files are not invalid, so there is no skipped file, the test is failed.
@git-hulk Also fix error on executing `r config get dir` when replica is restoring.

### DIFF
--- a/tests/tcl/tests/integration/replication.tcl
+++ b/tests/tcl/tests/integration/replication.tcl
@@ -135,8 +135,13 @@ start_server {tags {"repl"} overrides {max-replication-mb 1 rocksdb.compression 
     populate 1024 "" 10240
     r set a b
     r compact
+    after 1000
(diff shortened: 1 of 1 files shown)


git-hulk:
LGTM, info field `is_compacting` can tell us whether the compact is progressing or not, but it's ok if sleep longer and we can have an eye on this case.

git-hulk (dismissed):

ShooterIT:
@git-hulk oh, thanks, i missed this, let me fix

ShooterIT:
And I realize I can't use `wait_for_sync` because kvrocks doesn't support to run `info` when restoring

karelrooted (approved):

git-hulk (approved):

PR #284, opened 2021-06-01
Set TCP_KEEPALIVE for replication connection

shangxiaoxiong:
Before this commit, socket packet loss caused by network failure may result in
unrecoverable replication interrupt. When master failed to send data to replicas,
master would close repl-fd and terminate FeedSlaveThread. If the network loses
packets, replicas could not sense the replication interruption any more.
Now, repl-fd is set TCP_KEEPALIVE in replica-side, bad case above mentioned
can be handled.

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -214,6 +214,12 @@ void ReplicationThread::CallbacksStateMachine::Start() {
     // in ConnEventCB. the error here is something fatal.
     LOG(ERROR) << "[replication] Failed to start state machine, err: " << strerror(errno);
   }
+
+  auto s = Util::SockSetTcpKeepalive(bufferevent_getfd(bev), 120);
+  if (!s.IsOK()) {
+    LOG(ERROR) << "[replication] Failed to set tcp-keepalive, err:" << s.Msg();
+  }
+
   handler_idx_ = 0;
   repl_->incr_state_ = Incr_batch_size;
   if (getHandlerEventType(0) == WRITE) {


ShooterIT (approved):

git-hulk:
Very cool for finding out such a corner case in replication. I have a question that the master should transmit the TCP `fin` packet to the slave after closing the connection, and the slave would get an `EOF` if it reads the fd(write operation would get the broken pipe error), right?

shangxiaoxiong:
> Very cool for finding out such a corner case in replication. I have a question that the master should transmit the TCP `fin` packet to the slave after closing the connection, and the slave would get an `EOF` if it reads the fd(write operation would get the broken pipe error), right?
l think you are right !

shangxiaoxiong:
> Very cool for finding out such a corner case in replication. I have a question that the master should transmit the TCP `fin` packet to the slave after closing the connection, and the slave would get an `EOF` if it reads the fd(write operation would get the broken pipe error), right?

I think you are right !

ShooterIT:
This commit is a esay way to mitigate this problem. 
But there also is a bad case: master tcp still is alive but master hang much time, replicas should disconnect with master, let us know that. I remember, for redis, replicas will disconnect if it is too long time without recieving master replication stream.

git-hulk (approved):

git-hulk:
I think this PR would fix the issue when the master server went away suddenly before the kernel sent out the fin packet. EOF and Error should have checked now.

karelrooted (approved):

ShooterIT:
@git-hulk Yes, what you said is one of cases. if master is alive, it also can't send replication data to slaves such as swither and router failuere

PR #289, opened 2021-06-02
Support the auto-resize-block-and-sst config directive

### LINKED ISSUE #286: [QUESTION]when will rocksdb.write_buffer_size be changed to 16
image

Version: 2.0.1 @5f41145a

I set target_file_size_base and rocksdb.write_buffer_size to 128 in kvrocks.conf then use config get to double check the value was set to 128.

But, some minutes later, the value was changed to 16, log in kvrocks.INFO

```
I0602 11:43:43.800396 11653 server.cc:1022] [server] Resize rocksdb.target_file_size_base from 128 to 16, average_kv_size: 15, total_size: 104606264039, total_keys: 6620970652, result: ok
I0602 11:43:45.627482 11653 server.cc:1038] [server] Resize rocksdb.write_buffer_size from 16 to 16, average_kv_size: 15, total_size: 104606264039, total_keys: 6620970652, result: ok
```

We can find that `rocksdb.target_file_size_base` was changed to 16, then `rocksdb.write_buffer_size` was changed to the same as `rocksdb.target_file_size_base`.
And the most interesting is `from 16 to 16` but not `from 128 to 16`, I'm quite sure the value for `rocksdb.write_buffer_size` in kvrocks.conf is 128 before kvrocks started.

Questions:

1. when will this happen or this is a bug?
2. what also can I do for your tracing

### LINKED ISSUE #287: [BUG] Auto-adjust write buffer size didn't update the config item
As issue: https://github.com/bitleak/kvrocks/issues/286 mentioned that the write buffer size was resized but didn't update the config value as well.

BTW: as @ShooterIT mentioned, we should become this behavior with a config value that allows users to enable/disable this feture.

karelrooted:
resolves #287 
resolves #286

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -286,6 +286,21 @@ compaction-checker-range 0-7
 #
 # rename-command KEYS ""
 
+# The key-value size may so be quite different in many scenes, and use 256MiB as SST file size
+# may cause data loading(large index/filter block) ineffective when the key-value was too small.
+# kvrocks supports user-defined SST file in config(rocksdb.target_file_size_base),
+# but it still too trivial and inconvenient to adjust the different sizes for different instances.
+# so we want to periodic auto-adjust the SST size in-flight with user avg key-value size.
+#
+# If enabled, kvrocks will auto resize rocksdb.target_file_size_base
+# and rocksdb.write_buffer_size in-flight with user avg key-value size.
+# Please see https://github.com/bitleak/kvrocks/issues/118.
+#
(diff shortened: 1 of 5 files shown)


karelrooted:
auto resize block_size will be added in another PR (rocksdb 6.18.0 add this feature https://github.com/facebook/rocksdb/pull/7936)

git-hulk (approved):

ShooterIT (approved):

karelrooted:
@Mergifyio backport 1.3

PR #291, opened 2021-06-02
Encode key with slot id

shangxiaoxiong:
Encode key with slot id, so it will be efficient to migrate keys by slot when scale

Also support CLUSTER KEYSLOT command


kvrocks.conf
```diff
@@ -200,6 +200,12 @@ max-backup-to-keep 1
 # default: 1 day
 max-backup-keep-hours 24
 
+# Enable the kvrocks to support cluster mode that encoding key with slot id, if the db enabled
```
ShooterIT: these lines are too long

src/redis_db.cc
```diff
@@ -355,7 +433,8 @@ rocksdb::Status Database::Type(const Slice &user_key, RedisType *type) {
 }
 
 void Database::AppendNamespacePrefix(const Slice &user_key, std::string *output) {
-  ComposeNamespaceKey(namespace_, user_key, output);
+
```
ShooterIT:
```suggestion
```
remove empty line

tests/tcl/tests/unit/cluster.tcl
```diff
@@ -0,0 +1,14 @@
...
+
+    test {CLUSTER KEYSLOT} {
+
+        set slot_table_len [llength $::CRC16_SLOT_TABLE]
+
+        for {set i 0} {$i < $slot_table_len} {incr i} {
+            assert_equal $i [r cluster keyslot [lindex $::CRC16_SLOT_TABLE $i]]
+        }
+    }
+}
```
ShooterIT:
```suggestion
}

```
add one more empty line

src/version.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
```
ShooterIT: remove this file

src/redis_db.cc
```diff
@@ -206,17 +240,60 @@ rocksdb::Status Database::Scan(const std::string &cursor,
...
+    }
+
+    if (++slot_id >= HASH_SLOTS_SIZE) {
+      break;
+    }
+
+    if (slot_id > slot_start + HASH_SLOTS_MAX_ITERATIONS) {
+      if (keys->empty()) {
+        if (iter->Valid()) {
+
```
ShooterIT:
```suggestion
```

kvrocks.conf
```diff
@@ -200,6 +200,12 @@ max-backup-to-keep 1
...
+# the cluster mode at first open, this option must not be disabled after restarted, and vice versa
+#
+# Defalut: yes
+cluster-enabled yes
```
ShooterIT: by default, it should be `no`
```suggestion
# Defalut: no
cluster-enabled no
```

src/redis_cmd.cc
```diff
@@ -32,6 +32,8 @@
 
 namespace Redis {
 
+const std::string kCursorPrefix = "_";
```
ShooterIT: I think we should add comments for `scan` in doc/commands to say our cursor is not a muber, but 0 always is the end of cursor

ShooterIT:
@shangxiaoxiong Maybe you should add a new rocksdb CF to identify cluster mode and check if it is safe to enable/disable cluster mode in config file. WDYT?

shangxiaoxiong:
Good suggestion！l will do it right away. @ShooterIT

ShooterIT (dismissed):

ShooterIT:
We don't have `sortedint` type tcl tests, so currently some code is not tested, i hope you can supplement some basic tcl test cases.

ShooterIT:
After discussion, i prefer to encode slot id into key **by default**, so it is easy to change from master-replica mode to cluster mode for new version, even encoding slot id doesn't support `namespace`.

src/redis_db.cc
```diff
@@ -355,7 +433,7 @@ rocksdb::Status Database::Type(const Slice &user_key, RedisType *type) {
 }
 
 void Database::AppendNamespacePrefix(const Slice &user_key, std::string *output) {
-  ComposeNamespaceKey(namespace_, user_key, output);
+  ComposeNamespaceKey(namespace_, user_key, output, storage_->IsClusterEnabled());
```
git-hulk: How about moving the `is_cluster_enabled` as the first parameter that keeps the input params before outputs

src/redis_db.cc
```diff
@@ -150,31 +159,43 @@ void Database::Keys(std::string prefix, std::vector<std::string> *keys, KeyNumSt
...
-      int32_t ttl = metadata.TTL();
-      stats->n_key++;
-      if (ttl != -1) {
-        stats->n_expires++;
-        if (ttl > 0) ttl_sum += ttl;
+
+  while (true) {
+    ns_prefix.empty() ? iter->SeekToFirst() : iter->Seek(ns_prefix);
+    for (; iter->Valid(); iter->Next()) {
+      if (!prefix.empty() && !iter->key().starts_with(ns_prefix)) {
```
git-hulk: Should be `ns_prefix`?

src/redis_db.cc
```diff
@@ -150,31 +159,43 @@ void Database::Keys(std::string prefix, std::vector<std::string> *keys, KeyNumSt
...
+        keys->emplace_back(user_key);
       }
     }
-    if (keys) {
-      ExtractNamespaceKey(iter->key(), &ns, &user_key);
-      keys->emplace_back(user_key);
-    }
+
+    if (!storage_->IsClusterEnabled()) break;
+    if (prefix.empty()) break;
```
git-hulk, line 191: Why we need to break the loop when the prefix was empty(which means `keys *` pattern)

src/redis_db.cc
```diff
@@ -150,31 +159,43 @@ void Database::Keys(std::string prefix, std::vector<std::string> *keys, KeyNumSt
...
+    if (++slot_id >= HASH_SLOTS_SIZE) break;
+
+    ComposeNamespaceKey(namespace_, "", &ns_prefix);
+    PutFixed16(&ns_prefix, slot_id);
+    ns_prefix.append(prefix);
   }
+
```
git-hulk, line 198: It Seems uses do...while sentence would be more clear, like below:
```
do {
} while(storage_->IsClusterEnabled() && ++slot_id < HASH_SLOTS_SIZE)
```

git-hulk (commented):
cool, overall was good!

src/redis_db.cc
shangxiaoxiong: Seems a bug. Should be ns_prefix.

src/redis_db.cc
shangxiaoxiong: when prefix was empty, keys * is executed. 
Internal cycle `for {} ` can traverse all the keys in a namespace.

shangxiaoxiong:
Comparing with multi-db of Redis,  I prefer that multi-namespace is allowed whether slot-id-encoded is enabled or not but not allowed when cluster mode is enabled.

tests/tcl/tests/unit/type/sint.tcl
```diff
@@ -0,0 +1,17 @@
...
+        assert_equal 7 [r SICARD mysi]
+        assert_equal {245 123 5} [r SIREVRANGE mysi 0 3]
+        assert_equal {4 3 2} [r SIREVRANGE mysi 0 3 cursor 5]
+        assert_equal {245} [r SIRANGE mysi 0 3 cursor 123]
+        assert_equal {1 2 3 4} [r SIRANGEBYVALUE mysi 1 (5]
+        assert_equal {5 4 3 2} [r SIREVRANGEBYVALUE mysi 5 (1]
+        assert_equal {1 0 1} [r SIEXISTS mysi 1 88 2]
+        assert_equal {1} [r SIREM mysi 2]
+    }
+}
```
ShooterIT:
```suggestion
}

```

ShooterIT:
I think maybe w can add a special column family named `server`, and special information about server, such as `slot-id-encoded`, `cluster-enabled`, in the future, cluster topology also could be persistent in this CF.

src/redis_db.cc
shangxiaoxiong: it is not easy to use this way, because we may need to init some variables.

src/redis_db.cc
shangxiaoxiong: `ExtractNamespaceKey` also has the same problem. I won't deal with it this time beacuse tuning will affect kvrocks2redis tool. Kvrocks2redis tool does not support slot-id-encoded currently and l will make it support next time and deal with the problem incidentally.

src/redis_zset.cc
```diff
@@ -31,7 +31,8 @@ rocksdb::Status ZSet::Add(const Slice &user_key, uint8_t flags, std::vector<Memb
   std::string member_key;
   std::set<std::string> added_member_keys;
   for (int i = static_cast<int>(mscores->size()-1); i >= 0; i--) {
-    InternalKey(ns_key, (*mscores)[i].member, metadata.version).Encode(&member_key);
+    InternalKey(ns_key, (*mscores)[i].member, metadata.version,
+                storage_->IsSlotIdEncoded()).Encode(&member_key)).Encode(&member_key);
```
ShooterIT:
```suggestion
                storage_->IsSlotIdEncoded()).Encode(&member_key);
```

src/storage.cc
```diff
@@ -199,7 +204,16 @@ Status Storage::Open(bool read_only) {
   std::vector<std::string> old_column_families;
   auto s = rocksdb::DB::ListColumnFamilies(options, config_->db_dir, &old_column_families);
   if (!s.ok()) return Status(Status::NotOK, s.ToString());
-
+  if (!config_->slot_id_encoded && column_families.size() != old_column_families.size()) {
```
ShooterIT: please revert this mechanism

tests/tcl/tests/unit/scan.tcl
```diff
@@ -207,4 +207,63 @@ start_server {tags {"scan network"}} {
...
+                set cur [lindex $res 0]
+                set k [lindex $res 1]
+                lappend keys {*}$k
+                if {$cur == 0} break
+            }
+            assert_equal 1000 [llength $keys]
+
+            incr index
+        }
+    } {}
```
ShooterIT: `{}` is necessary?

tests/tcl/tests/unit/keyspace.tcl
```diff
@@ -107,4 +107,39 @@ start_server {tags {"keyspace"}} {
...
+        # Check KEYS and KEYS MATCH in different namespace
+        set index 0
+        foreach token $tokens {
+            r auth $token
+            assert_equal [lsort [r keys *]]  [lindex $ns_keyspaces $index]
+            assert_equal [lsort [r keys foo*]]  [lindex $foo_prefix_keyspaces $index]
+
+            incr index
+        }
+    } {}
```
ShooterIT: {} is necessary?

ShooterIT (dismissed):

ShooterIT:
@shangxiaoxiong The remaining work
- kvrocks2redis support `slot-id-encoded`, and you also should make sure all keys belonging to one slot are stored together.
- `slot-id-encoded` information are stored in a special CF after we support cluster mode

shangxiaoxiong:
> @shangxiaoxiong The remaining work
> 1、kvrocks2redis support `slot-id-encoded`, and you also should make sure all keys belonging to one slot are stored together.
> 2、`slot-id-encoded` information are stored in a special CF after we support cluster mode

ShooterIT (dismissed):

shangxiaoxiong:
> @shangxiaoxiong The remaining work
> 
> * kvrocks2redis support `slot-id-encoded`, and you also should make sure all keys belonging to one slot are stored together.
> * `slot-id-encoded` information are stored in a special CF after we support cluster mode

Strictly speaking, make sure all keys belonging to one slot of **a namespace** are stored together.

kvrocks.conf
```diff
@@ -200,6 +200,21 @@ max-backup-to-keep 1
...
+#
+# PLEASE NOTE:
+# If you just start to use kvrocks firstly, please don't change it optionally,
+# because cluster mode is dependent with enabling that.
+#
+# If you want to upgrade your old kvrocks version(without encoding slot it into key)
+# whatever by master-replicas replication or restarting using new binary, you must
+# disable this option, otherwise, kvrocks will make data corrupt.
+#
+# Defalut: yes
```
```suggestion
# Default: yes
```

git-hulk (dismissed):

ShooterIT (approved):

git-hulk (approved):

karelrooted (approved):

ShooterIT:
@shangxiaoxiong Finally, merged, another thing i find is that we should tell users our cursor of scan is not a number instead of a string that has a special prefix `-`.

PR #296, opened 2021-06-03
Fix testcases "PUBLISH/SUBSCRIBE after PUNSUBSCRIBE/UNSUBSCRIBE without arguments"

karelrooted:
read PUNSUBSCRIBE/UNSUBSCRIBE all's response before publish (publish may exec before PUNSUBSCRIBE/UNSUBSCRIBE all command finished in multi worker thread envoriment when PUNSUBSCRIBE/UNSUBSCRIBE all command's response is not being readed)


ShooterIT:
You may aslo need to fix `PUBLISH/PSUBSCRIBE after PUNSUBSCRIBE without arguments` test

> `punsubscribe $rd1`

ShooterIT (approved):
It is luckly that the subscribe patterns are add into a list, so the return of unsubscribe is ordered

git-hulk (approved):

PR #298, opened 2021-06-05
Fix zset can't add the same member with different scores

git-hulk:
Fix the corner case that adds the same member which may add the score
column family many times and cause problems in the ZRANGE command.

For example, we add members with `ZADD mykey 1 a 2 a` and `ZRANGE mykey 0 1`
return only one member(`a`) was expected but got the member `a` twice now.

The root cause of this issue was the score key  was composed by member and score,
so the last one can't overwrite the previous when the score was different.
A simple workaround was add those members with reversed order and skip the member if has added.

### DIFF
--- a/src/redis_zset.cc
+++ b/src/redis_zset.cc
@@ -2,6 +2,8 @@
 
 #include <math.h>
 #include <map>
+#include <limits>
+#include <cmath>
 #include <memory>
 #include <set>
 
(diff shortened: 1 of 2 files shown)


git-hulk:
Many thanks for @smartlee finding this issue

ShooterIT (dismissed):

karelrooted (dismissed):

PR #299, opened 2021-06-06
Allow to use the hostname in the replication

### LINKED ISSUE #293: [BUG] Allow to use host in the replication
I have configured kvrocks master and slave on one server.
master kvrocks port : 9188
slave kvrocks port: 9288

In the slave's configuration file.
If I put hostname instead of ip in "slaveof field", an error occurs.

but Redis works fine with both ip and hostname in slaveof field.

#####################################################
kvrocks.conf of slave
...............
slavof localhost 9188
...............

SLAVE$ ./kvrocks -c kvrocks.conf
Version: 999.999.999 @0daf3b5e
E0603 09:06:30.111851 5138 replication.cc:214] [replication] Failed to start state machine, err: Network is unreachable <====== ERROR!

####################################################
kvrocks.conf of slave
...............
slavof 127.0.0.1 9188
...............
SLAVE$ ./kvrocks -c kvrocks.conf
Version: 999.999.999 @0daf3b5e <====== works good!

cf.
STAGE$ ping localhost
PING localhost (127.0.0.1) 56(84) bytes of data.
64 bytes from localhost (127.0.0.1): icmp_seq=1 ttl=64 time=0.031 ms
64 bytes from localhost (127.0.0.1): icmp_seq=2 ttl=64 time=0.041 ms
64 bytes from localhost (127.0.0.1): icmp_seq=3 ttl=64 time=0.052 ms

git-hulk:
resolved #293

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -5,8 +5,10 @@
 #include <sys/uio.h>
 #include <netinet/tcp.h>
 #include <sys/socket.h>
+#include <arpa/inet.h>
 #include <poll.h>
 #include <errno.h>
+#include <netdb.h>
 #include <pthread.h>
 #include <fcntl.h>
 #include <math.h>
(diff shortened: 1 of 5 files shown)


src/util.cc
```diff
@@ -38,18 +40,54 @@
...
   return sin;
 }
 
 Status SockConnect(std::string host, uint32_t port, int *fd) {
+  char ip[16];
   sockaddr_in sin{};
+
   sin.sin_family = AF_INET;
-  sin.sin_addr.s_addr = inet_addr(host.c_str());
+  Status s = Host2IP(host, ip, sizeof(ip)/sizeof(ip[0]));
```
ShooterIT, line 85: here, we maybe should iterate `getaddrinfo` result, and try to connect, because some IPs may be not connected.

PR #301, opened 2021-06-07
Allow to use host in the master-slave replication

git-hulk:


src/replication.cc
```diff
@@ -201,22 +201,18 @@ void ReplicationThread::CallbacksStateMachine::EvCallback(bufferevent *bev,
...
-    // NOTE: Connection error will not appear here, network err will be reported
-    // in ConnEventCB. the error here is something fatal.
-    LOG(ERROR) << "[replication] Failed to start state machine, err: " << strerror(errno);
-  }
-
-  auto s = Util::SockSetTcpKeepalive(bufferevent_getfd(bev), 120);
+  Status s = Util::SockConnect(repl_->host_, repl_->port_, &cfd);
   if (!s.IsOK()) {
-    LOG(ERROR) << "[replication] Failed to set tcp-keepalive, err:" << s.Msg();
+    LOG(ERROR) << "[replication] Failed to connect the master, err:" << s.Msg();
```
ShooterIT, line 211: Just one log is enough?

git-hulk: no, should return here. it's missing before, fix it now.

shangxiaoxiong: If return here, stateMachine terminates and will not run again. It seems unreasonable.

ShooterIT:
Just as we discussed, i record one card https://github.com/bitleak/kvrocks/projects/1#card-62672516, we should do that ASAP @bitleak/kvrocks-contributors

ShooterIT (approved):

karelrooted (approved):

PR #302, opened 2021-06-08
Support cluster mode

### LINKED ISSUE #219: [NEW] Support redis cluster mode
# 1 Background

Currently, we discard the support for `codis` mode,  if we want to deploy kvrocks cluster, we need to pre-shard for cluster and deploy a proxy like `twemproxy` to route requests. This cluster solution also is adopted when redis doesn't support cluster mode, but it is not convenient for failover, linear scale, management, moreover, we must route requests by a proxy, that must cause performance loss because of extra network communication. So we hope there a way we can implement a feature that is similar with redis cluster mode.

But if we totally support redis cluster mode in kvrocks, it doesn't seem a good idea, because it is hard to implement redis cluster management and gossip protocol without bugs. We must upgrade server if we find bugs, after all, it costs many years to make redis cluster work well. In order to get benefits from redis cluster mode, we can only support partial redis cluster commands that make a chance to let `redis cluster sdk(such as jedis)`  or `redis cluster proxy(such as redis-cluster-proxy, predixy)` work well.

# 2 Implementation
## 2.1 Basic version
Firstly, every node needs to know the entire cluster topology, itself identifier. When clients(`redis-cli`, SDK or proxy) connect to  arbitrary one node of kvrocks cluster, they could know slot distribution of cluster by sending `CLUSTER NODES` command, then route requests.  And the kvrocks should reply `MOVED` when clients send requests to wrong server. To do that, we should implement `CLUSTER SETNODES` command to set the entire cluster topology for every nodes, and provide `CLUSTER NODES` command to make kvrocks node similar with redis.

- CLUSTER NODES
   The return of this command is the same format of redis cluster. 
    Example: 
    ```
    07c37dfeb235213a872192d90877d0cd55635b91 127.0.0.1:30004 slave e7d1eecce10fd6bb5eb35b9f99a514335d9ba9ca 0 1426238317239 4 connected
    67ed2db8d677e59ec4a4cefb06858cf2a1a89fa1 127.0.0.1:30002 master - 0 1426238316232 2 connected 5461-10922
   ```

- CLUSTER SETNODES `$ALL_NODES_INFO` `$VERSION` `FORCE`
   `$ALL_NODES_INFO` format: `$node_id` `$ip $port` `$role` `$master_node_id` `$slot_range`. `$role`: master or slave,  `$master_node_id`: master_node_id if it is slave, otherwise, this value is `-`. slave needn't set `$slot_range`.
    Example: 
    ```
    07c37dfeb235213a872192d90877d0cd55635b91 127.0.0.1 30004 slave e7d1eecce10fd6bb5eb35b9f99a514335d9ba9ca
    67ed2db8d677e59ec4a4cefb06858cf2a1a89fa1 127.0.0.1 30002 master - 5461-10922
    ```
    `$VERSION` means current cluster topology version that is stored or generated by management component such as config server or meta-server.
      Kvrocks can know itself node id by compared with itself ip and port in received `all nodes info`

From now, it works well, we just tell every node of kvrocks cluster by `CLUSTER SETNODES` when scale(slot data migration) or failover. Clients(redis-cli, sdk, proxy) will reacquire cluster topology when they finds requests are rejected or `MOVED` by using old topology to route requests. To avoid missing notify nodes new topology, we can set new cluster topology every longer interval.

  _*Please notice that: the cluster solution for kvrocks doesn't handle scale or failover._

## 2.1 Advanced version
We find we always tell every node entire cluster topology even just changing one node info. If the cluster is very large, the message will be big, so we also fall into the damage of network communication, and cluster management component may can't support that. So we may need to support incremental changing notification. When changing topology, we only send one light command to change the topology in kvrocks server. To implement incremental changing notification, we must use `version` to identify current topology version, otherwise, kvrocks may keep a wrong cluster topology if missing one changing topology message. Kvrocks only supports to apply one command content when its version is valid by compared with the `$version` in command, if not, kvrocks must return an error, and we will set new entire cluster topology by `CLUSTER SETNODES` command.

There are some commands we may need to support.
-  (optional) CLUSTER VERSION
    Get version of node, so that we can check its' version is right or not.
-  (optional) CLUSTER SETNODEID `$NODE_ID`
    Explicitly tell the node its node id.
-  (optional) CLUSTER SETNODE `$NODE_ID` `$NODE_INFO` `$VERSION`
    Only update one node info
-  (optional) CLUSTER DELNODE `$NODE_ID` `$VERSION`
    Delete one node of cluster
-  (optional) CLUSTER REPLICATE `$NODE_ID`
    Just like `slaveof`, similar with redis command.
-  (optional) CLUSTER MIGRATE `$SLOT_ID` `$NODE_ID`
    Migrate all keys belonging the slot to another node
-  (optional) CLUSTER SETSLOT `$SLOT_ID` `$NODE_ID` `$VERSION`
    Only update one slot distribution, that is useful to avoid too many big messages when we scale cluster.
-  (optional) CLUSTER SLOTS/INFO/KEYSLOT/GETKEYSINSLOT/COUNTKEYSINSLOT
    To make kvrocks cluster similar with redis cluster

# 3 Advantages and disadvantages
## 3.1 Advantages
- It is easy to implement and make less bugs.
- Cluster logic are removed from server, so that we needn't change kvrocks code when optimizing fault check or failover for kvrocks instances. 
## 3.2 Disadvantages
- Topology changing notifications are sent by other component, clients may feel latency when cluster topology change, even although that is low-frequency.
- It is not perfect, we still need a cluster management component to manage cluster such as failover and scale.

# 4 Extra works
- More redis cluster commands to manage kvrocks cluster conveniently just like redis cluster.
- Cluster management component. To make kvrocks cluster easy, we may provide a cluster management component to deal with linear scale, failover, or management.

ShooterIT:
resolves #219 

# Deploy Kvrocks Cluster
In our cluster solution, you should firstly deploy some kvrocks instances and set replication relationship, and
Kvrocks doesn't support to communicate with each others, so we need to tell every node the cluster topology.
The command is `CLUSTER SETNODES`
- CLUSTERX SETNODES `$ALL_NODES_INFO` `$VERSION` `$FORCE`
   `$ALL_NODES_INFO` format: `$node_id` `$ip $port` `$role` `$master_node_id` `$slot_range`. `$role`: master or slave,  `$master_node_id`: master_node_id if it is slave, otherwise, this value is `-`. slave needn't set `$slot_range`.
    Example: 
    ```
    07c37dfeb235213a872192d90877d0cd55635b91 127.0.0.1 30004 slave e7d1eecce10fd6bb5eb35b9f99a514335d9ba9ca
    67ed2db8d677e59ec4a4cefb06858cf2a1a89fa1 127.0.0.1 30002 master - 5461-10922
    ```
    `$VERSION` means current cluster topology version that is stored or generated by management component such as config server or meta-server.
 
Kvrocks can know itself node id by compared with itself ip and port in received `all nodes info`, but you maybe bind kvrorks to `0.0.0.0`, so it is necessary to set `NODEID` explicitly to let it know itself. As the same with redis cluster, node id is 40 characters length.
- CLUSTERX SETNODEID `$NODE_ID`

Moreover, cluster doesn't handle failover and migration, you should set cluster topology again after failover, of course, you should do that if you add some new nodes into cluster.

# Access Kvrocks Cluster
After setting cluster topology, we have tested, it adopt `redis-cli`, `redis-benchmark`, `python redis cluster sdk`, and `redis-cluster-proxy`. So you may doesn't need to change your redis cluster request code to access kvrocks cluster.

# Have a try
In `tool/try_cluster`, you can change `BIN_PATH` of kvrocks bin path, and then
`./try_cluster.sh start` to create a small cluster that has 6 nodes(3 maters and 3 replicas)

then run `./try_cluster.sh watch` to observe cluster nodes,  you also can do some tests by
`redis-cli`, `redis-benchmark`, `redis cluster sdk` and `redis-cluster-proxy` for your fun.

finally you can run `./try_cluster.sh stop` to stop all nodes and clean up directory.

# Supplements
Use `CLUSTERX` command to set cluster topology instead of `CLUSTER` #324


git-hulk (approved):
LGTM, the cluster topology and version would store on local storage?

ShooterIT:
@git-hulk That is a good question, I still consider that, we must should know the nodes may keep old cluster topology if restarted. That may cause terrible things  if the IP and port are reused.

git-hulk:
yup, use ip:port as the unique identifier would cause problems, the node id would be better if we also persist it.

karelrooted (approved):

PR #304, opened 2021-06-10
 Remove the default value of is_slotid_encoded

git-hulk:

### DIFF
--- a/tools/kvrocks2redis/parser.cc
+++ b/tools/kvrocks2redis/parser.cc
@@ -38,7 +38,7 @@ Status Parser::ParseFullDB() {
 
 Status Parser::parseSimpleKV(const Slice &ns_key, const Slice &value, int expire) {
   std::string op, ns, user_key;
-  ExtractNamespaceKey(ns_key, &ns, &user_key);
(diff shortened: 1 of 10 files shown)


tools/kvrocks2redis/parser.cc
```diff
@@ -155,7 +155,7 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
   std::string ns, user_key, sub_key;
   std::vector<std::string> command_args;
   if (column_family_id == kColumnFamilyIDMetadata) {
-    ExtractNamespaceKey(key, &ns, &user_key);
+    ExtractNamespaceKey(key, &ns, &user_key, this->is_slotid_encoded_);
```
karelrooted: other private parameter (first_seen_, aof_strings_, etc..) did not use this->,  is this on purpose ?

git-hulk: Good, you're right that we shouldn't use `this` keyword

karelrooted (approved):

ShooterIT (approved):
lgtm but don't read

PR #314, opened 2021-06-18
Fix LTRIM only delete first item, LREM cannot correctly remove repeated items

ChrisZMF:
Fix LTRIM command only can delete the first index item, the metadata is changed,
but items are not delete from RocksDB. 
Cases:
rpush listkey 0 1 2 3 4 3 6 7 3 9 -> listkey: 0 1 2 3 4 3 6 7 3 9
ltrim listkey 2 -3 -> listkey: 2 3 4 3 6 7
lrem listkey 3 3 -> listkey should be: 2 4 6 7, but it is: 2 4 6

Fix LREM cannot remove item correctly from sequence with repeated items,
the reason is that the number of deleted elements is not judged, which
will cause the repeated items that need to be retained are not copied.
Case:
rpush listkey a b c d e f a f -> listkey: a b c d e f a f
lrem listkey 1 f -> listkey should be: a b c d e a f, but it is: a b c d e a a

Other fixes:
Fix wrong read/write attributes of some commands, such as lrem, linsert, flushdb


git-hulk (approved):

git-hulk (dismissed):

ShooterIT (dismissed):

ShooterIT (commented):
please give more details for `ltrim` bug @ChrisZMF It is a not small bug

ShooterIT:
After second look, I think `ltrim` only can rightly delete the first left index item, other indexes maybe  don't really be deleted, but what's worse, we may delete some items that we don't want to delete.
if a list has 150 items, we execute ltrim 3 -3, so we will try to delete the index 1, 12, 123, 123149, 123149148, 123149148147, actually, we deleted 1, 12, 123, it is terrible to delete 12 and 123. right? @ChrisZMF @git-hulk

ShooterIT:
so i may ask you to add tcl test cases, the list must have not less than 20 items to `trim` some range in test cases

git-hulk:
yeah, we should clear the buf before using it

git-hulk:
> if a list has 150 items, we execute ltrim 3 -3, so we will try to delete the index 1, 12, 123, 123149, 123149148, 123149148147, actually, we deleted 1, 12, 123, it is terrible to delete 12 and 123. right?

I think this wouldn't happen since it would always write 8 bytes everytime

git-hulk:
@ChrisZMF  Can you help to fix the trim test case as well:

https://github.com/KvrocksLabs/kvrocks/blob/11ea1f6b4862d6ab05bce9abfda2576900ea1b5d/tests/t_list_test.cc#L104

ShooterIT:
oh, you are right, one index is 8 bit, there is no possibility to have so many items in one list. i changed title and comments.

ShooterIT:
@ChrisZMF For the title, you should write the most important things, it is not a small bug, you should tell users the concrete bug firstly, now by this title, I don't clearly understand the bug.  For some other not much important things, you can say in comments.

ChrisZMF:
> @ChrisZMF For the title, you should write the most important things, it is not a small bug, you should tell users the concrete bug firstly, now by this title, I don't clearly understand the bug. For some other not much important things, you can say in comments.

git-hulk (dismissed):

tests/t_list_test.cc
```diff
@@ -12,13 +12,26 @@ class RedisListTest : public TestBase {
   }
   void SetUp() override {
     key_ = "test-list-key";
-    fields_ = {"list-test-key-1", "list-test-key-2", "list-test-key-3", "list-test-key-4", "list-test-key-5"};
+    fields_ = {
+                "list-test-key-1", "list-test-key-2", "list-test-key-3", "list-test-key-4", "list-test-key-5",
+                "list-test-key-1", "list-test-key-2", "list-test-key-3", "list-test-key-4", "list-test-key-5",
```
ShooterIT, line 17: why so many other items except `list-test-key-1`, as the following test cases, i find you only test `lrem list-test-key-1`

tests/t_list_test.cc
```diff
@@ -101,6 +114,117 @@ TEST_F(RedisListTest, Range) {
...
+  list->Size(key_, &len);
+  EXPECT_EQ(fields_.size()-1, len);
+  int cnt = 0;
+  for (size_t i = 0; i < fields_.size(); i++) {
+    if (fields_[i] == del_elem) {
+      if (++cnt <= 1) continue;
+    }
+    std::string elem;
+    list->Pop(key_, &elem, true);
+    EXPECT_EQ(elem, fields_[i].ToString());
```
ShooterIT, line 208: Actually I prefer we can also add some tcl test cases, you can show the new list members for comparison, that may be easier and clearer

ShooterIT:
@ChrisZMF any thought for above comments

tests/t_list_test.cc
ChrisZMF: Other items are used to test `lrem`  same items which are separated by others, like case `rem key_ -5 list-test-key-1`.

tests/t_list_test.cc
ChrisZMF: Yes, tcl test cases can be easier and clearer. I will add tcl cases for `lrem` and `ltrim` later.

ShooterIT (approved):

git-hulk (approved):

ShooterIT:
@ChrisZMF I update top comments, please have a look, correct if it's wrong

PR #321, opened 2021-06-29
Fix unable to link the libunwind in old version debian(<4.9)

xujianhai666:


git-hulk:
Many thanks to @xujianhai666 

We should add dynamic libraries at `target_link_libraries` not `target_compile_options` since it's linking behavior.

git-hulk (approved):

git-hulk (commented):
Many thanks to @xujianhai666

We should add dynamic libraries at target_link_libraries not target_compile_options since it's linking behavior.

ShooterIT:
thanks, merged

PR #323, opened 2021-07-09
Set slot-id-encoded to no by default for compatibility

ShooterIT:
Before, i may make mistakes, after many thoughts, i think we should set slot-id-encoded to no by default.
- Easy to compatible with old version
- From standalone mode to cluster mode, i don't think it is usual, client SDK will change for enabling cluster.
   so migrating data is reasonable.

### DIFF
--- a/tests/tcl/tests/assets/default.conf
+++ b/tests/tcl/tests/assets/default.conf
@@ -14,4 +14,4 @@ max-replication-mb 0
 max-io-mb 500
 max-db-size 0
 cluster-enable no
-slot-id-encoded no
+slot-id-encoded yes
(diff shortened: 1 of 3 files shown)


tests/tcl/tests/assets/default.conf
```diff
@@ -14,3 +14,4 @@ max-replication-mb 0
 max-io-mb 500
 max-db-size 0
 cluster-enable no
+slot-id-encoded no
```
ShooterIT:
```suggestion
slot-id-encoded yes
```

git-hulk (approved):
Cool, I think it should be a good idea to disable the slot id encoded to make users happy.

PR #274, opened 2021-05-24
HSET supports to set multiple fields

ShooterIT:
Fixes #273

PR #328, opened 2021-07-15
Release 2.0.2

ShooterIT:
# Version 2.0.2
We design new cluster solution for kvrocks, but it is experimental, don't recommend to
deploy in production environments. And because in cluster mode, to be efficient to migrate
keys based on slot when scale, we encode key with slot id, but to be compatible with
the old version, we don't do that in standalone mode. So data is not compatible between
standalone mode with cluster mode, you must migrate data if you want to change mode,
otherwise, kvrocks  will make data corrupt.
In order to better support the cluster mode, we are  developing a  meta server to manage
kvrocks cluster, coming soon!

Here is a comprehensive list of changes in this release compared to 2.0.1.

## New features
  - Support rename command in config file (#272)
  - Support transaction (#285)
    Support MULTI, EXEC, DISCARD command, but don't support
    WATCH and UNWATCH commands.
  - Support the auto-resize-block-and-sst config directive (#289)
  - Support to resize rocksdb.block_size automatically (#294)
  - Support cluster mode (#302)
    Similar with redis cluster mode, use CLUSTERX command to set
    cluster topology (#324).

## Data encoding changes
  - Encode key with slot id (#291) (#330)
    In cluster mode, to be efficient to migrate keys by slot when scale,
    we encode key with slot id, but in standalone mode, we don't. That is
    to say, data is not compatible between standalone mode with cluster mode,
    you must migrate data if you want to change mode, otherwise, kvrocks
    will make data corrupt.

## Bug fixes
  - Set TCP_KEEPALIVE for replication connection (#284)
    Replicas can handle network failure with master, before this commit,
    network failure may result in unrecoverable replication interrupt.
  - Fix ZSET can't add the same member with different scores (#298)
  - Make CONFIG SET command thread safe (#310)
  - Fix LTRIM only delete first item, LREM can't correctly remove
    repeated items (#314)

## Improvements
  - HSET command supports to set multiple fields (#274)
  - Allow to use host in the master-slave replication (#301)
  - Adapt redis sentinel ping check (#316)


ShooterIT:
For `Encode key with slot id` #291 , if you don't use cluster mode, please don't change that, just use default config, ignore that.

git-hulk (approved):

git-hulk (approved):

karelrooted (approved):

PR #331, opened 2021-07-17
Use the rocksdb MultiGet in the MGet command

git-hulk:


ShooterIT:
It is great not to link `unwind` if it is not existed, i also find this problem, cool job 💯 , from this point, i want to merge this commit into 2.0.2. But in this commit, you also changed some others. so should we open another pr to resolve `unwind` warnings?

src/redis_string.cc
```diff
@@ -6,6 +6,34 @@
...
+  for (size_t i = 0; i < keys.size(); i++) {
+    if (!statuses[i].ok()) continue;
+    Metadata metadata(kRedisNone, false);
+    metadata.Decode((*raw_values)[i]);
+    if (metadata.Expired()) {
+      (*raw_values)[i].clear();
+      statuses[i] = rocksdb::Status::NotFound(kErrMsgKeyExpired);
+      continue;
+    }
+    if (metadata.Type() != kRedisString && metadata.size > 0) {
```
ShooterIT, line 28: in redis, mget don't return an error, just return `$-1`, i think we should return an error.

src/redis_string.cc
```diff
@@ -6,6 +6,34 @@
...
+  read_options.snapshot = ss.GetSnapShot();
+  std::vector<rocksdb::ColumnFamilyHandle*> cfs(keys.size(), metadata_cf_handle_);
+  auto statuses = db_->MultiGet(read_options, cfs, keys, raw_values);
+  for (size_t i = 0; i < keys.size(); i++) {
+    if (!statuses[i].ok()) continue;
+    Metadata metadata(kRedisNone, false);
+    metadata.Decode((*raw_values)[i]);
+    if (metadata.Expired()) {
+      (*raw_values)[i].clear();
+      statuses[i] = rocksdb::Status::NotFound(kErrMsgKeyExpired);
```
ShooterIT, line 25: if key is expired, in redis, also return `$-1`

src/redis_string.cc
```diff
@@ -19,23 +47,33 @@ rocksdb::Status String::getRawValue(const std::string &ns_key, std::string *raw_
   metadata.Decode(*raw_value);
   if (metadata.Expired()) {
     raw_value->clear();
-    return rocksdb::Status::NotFound("the key was expired");
+    return rocksdb::Status::NotFound(kErrMsgKeyExpired);
```
ShooterIT, line 50: As `mget`, redis also return `$-1\r\n`

src/redis_string.cc
ShooterIT: i am wrong, this error doesn't return to clients

git-hulk:
> It is great not to link `unwind` if it is not existed, i also find this problem, cool job 💯 , from this point, i want to merge this commit into 2.0.2. But in this commit, you also changed some others. so should we open another pr to resolve `unwind` warnings?

Okay, it's better to file another PR if want to merge into 2.0.2, I'll do it soon.

src/redis_string.cc
git-hulk: Good catch, we only transfer the not found error to nil string which may behave the same with Redis.

src/redis_string.cc
git-hulk: same with the previous comment, the NotFound error wouldn't throw to the user.

src/redis_string.cc
```diff
@@ -63,15 +101,22 @@ rocksdb::Status String::Append(const std::string &user_key, const std::string &v
...
+    ns_keys.emplace_back(ns_key);
   }
-  return statuses;
+  std::vector<Slice> slice_keys;
+  slice_keys.reserve(ns_keys.size());
+  // don't use range-based for loop here, coz the slice member
+  // would refer the address instead of copy the value, and use
+  // range-based for loop may cause all members refer to the same addr
+  for (size_t i = 0; i < ns_keys.size(); i++) {
+    slice_keys.emplace_back(ns_keys[i]);
```
ShooterIT, line 117: why we use `slice_keys`, `ns_keys` is not enough?

git-hulk: the RocksDB needs the slice vector instead of the string vector, so we need to convert the ns_keys to the slice vector here.

ShooterIT (approved):

karelrooted (approved):

PR #333, opened 2021-07-18
Only link the library when it's existed

git-hulk:


ShooterIT (approved):
IRRC, there only are warnings if unwind is not existed, not a bug, right?

git-hulk:
Seems the reply from my iPad didn't send out. Yes, U are right that is not a bug for the kvrocks, removed the label before merging.

PR #334, opened 2021-07-18
Supplements of guideline of compilation on macOS with make and clean up some code

LykxSassinator:
I have tried to compile the KVRocks on MacOS, with 11.4 Big Sur, finding that there have missed some tools for compilation on the first time. So i have make the supplements in the README.md.
Unfortunately, it seems that there still exists some limitations and errs while compiling KVRocks on MacOS by make with scripts automatically, which should be manually interfered with. Therefore, currently, it's more recommended to use cmake on MacOS or compiling it on other Linux systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ sudo yum install -y epel-release && sudo yum install -y git gcc gcc-c++ make sna
 sudo apt-get install gcc g++ make libsnappy-dev autoconf automake libtool which libgtest-dev
 
 # MACOSX
-brew install snappy googletest
+brew install autoconf automake libtool snappy googletest
 ```
 
 It is as simple as:
--- a/src/cron.cc
+++ b/src/cron.cc
@@ -29,7 +29,7 @@ Status Cron::SetScheduleTime(const std::vector<std::string> &args) {
     if (!s.IsOK()) {
       return Status(Status::NotOK, "time expression format error : " + s.Msg());
     }
-    new_schedulers.emplace_back(std::move(st));
+    new_schedulers.push_back(st);
   }
   schedulers_ = std::move(new_schedulers);
   return Status::OK();
--- a/src/cron.cc
+++ b/src/cron.cc
@@ -29,7 +29,7 @@ Status Cron::SetScheduleTime(const std::vector<std::string> &args) {
     if (!s.IsOK()) {
       return Status(Status::NotOK, "time expression format error : " + s.Msg());
     }
-    new_schedulers.push_back(st);
+    new_schedulers.emplace_back(std::move(st));
   }
   schedulers_ = std::move(new_schedulers);
   return Status::OK();
--- a/src/redis_db.cc
+++ b/src/redis_db.cc
@@ -365,6 +365,9 @@ rocksdb::Status Database::FindKeyRangeWithPrefix(const std::string &prefix,
   if (cf_handle == nullptr) {
     cf_handle = metadata_cf_handle_;
   }
+  if (prefix.empty()) {
+    return rocksdb::Status::NotFound();
+  }
   begin->clear();
   end->clear();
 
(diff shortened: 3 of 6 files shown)


LykxSassinator:
It's my unpleasable push of two separatable commits. The second commit is used for making several tiny improvements by converting unnecessary usages of Copy Constructor with the appropriate movement ops.

git-hulk:
cool, thanks for your contribution. Can u help to fix issues found by cpplint?

```
src/redis_db.cc:468:  Extra space for operator  ++;  [whitespace/operators] [4]
src/redis_hash.cc:178:  Add #include <utility> for move  [build/include_what_you_use] [4]
```

src/redis_db.cc
```diff
@@ -383,10 +386,8 @@ rocksdb::Status Database::FindKeyRangeWithPrefix(const std::string &prefix,
   // it's ok to increase the last char in prefix as the boundary of the prefix
   // while we limit the namespace last char shouldn't be larger than 128.
   std::string next_prefix = prefix;
-  char last_char = next_prefix.back();
-  last_char++;
-  next_prefix.pop_back();
-  next_prefix.push_back(last_char);
+  auto prefix_size = prefix.size();
+  next_prefix[prefix_size - 1] ++;
```
ShooterIT:
```suggestion
  next_prefix[prefix_size - 1]++;
```

src/cron.cc
```diff
@@ -29,7 +29,7 @@ Status Cron::SetScheduleTime(const std::vector<std::string> &args) {
     if (!s.IsOK()) {
       return Status(Status::NotOK, "time expression format error : " + s.Msg());
     }
-    new_schedulers.push_back(st);
+    new_schedulers.emplace_back(std::move(st));
```
ShooterIT: i think it is not a good idea to call move, since `st` may be access again

ShooterIT: actually, I also did that in parsing requests, but i find c++ standard doesn't recommend that, i just think we don't call that frequently.

ShooterIT: And i find we don't define its `move constructor`, so this change doesn't take any improvements. And all members of schedule just are `int` type.

src/redis_db.cc
LykxSassinator: It's my pleasure. Thx for the contributes of @ShooterIT .

src/cron.cc
LykxSassinator: Yep, at this point, std::move was not a recommendable operation for the user-defined `st` object. I would revoke it here.

LykxSassinator (commented):
All codes have been reviewed and checked again.
And It's my pleasure to make contributions to KVRocks.

README.md
```diff
@@ -47,7 +47,7 @@ sudo yum install -y epel-release && sudo yum install -y git gcc gcc-c++ make sna
 sudo apt-get install gcc g++ make libsnappy-dev autoconf automake libtool which libgtest-dev
 
 # MACOSX
-brew install snappy googletest
+brew install autoconf automake libtool snappy googletest
```
ShooterIT, line 50: do we need to brew install gcc? I find we install gcc in CI for macOS.
https://github.com/KvrocksLabs/kvrocks/blob/unstable/.github/workflows/kvrocks.yaml#L60

ShooterIT:
cpplint error https://github.com/KvrocksLabs/kvrocks/runs/3101400846#step:4:721 please add necessary header files.

README.md
LykxSassinator: The original version of GCC in MacOS is `4.2.1`. Although the compilation requirement requires a higher version(gcc >= 4.8), the compilation works around in fact. So, should we recommend the MacOS users upgrade the GCC version?

LykxSassinator (commented):
Done.

LykxSassinator (commented):
The error checked by cpplint has been fixed.

README.md
ShooterIT: it seems that gcc is clang on macOS platform, it is fine for compilation.

ShooterIT (approved):

git-hulk (approved):
LGTM

PR #335, opened 2021-07-20
Fix:SUBSCRIBE can't update last interaction time

smartlee:
Act like redis,SUBSCRIBE must update last interaction time while a publish message received, otherwise this channel may be closed for timeout reason by default.

### DIFF
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -234,6 +234,7 @@ Status Worker::Reply(int fd, const std::string &reply) {
   std::unique_lock<std::mutex> lock(conns_mu_);
   auto iter = conns_.find(fd);
   if (iter != conns_.end()) {
+    iter->second->SetLastInteraction();
     Redis::Reply(iter->second->Output(), reply);
     return Status::OK();
   }


git-hulk:
good catch @smartlee

git-hulk (approved):

git-hulk:
@Mergifyio backport 1.3

git-hulk:
@Mergifyio backport 1.3

ShooterIT:
@git-hulk Although there is a problem, i don't think it is right way to fix, because we just add reply to libevent buffer instead of writing to clients successfully, clients may not read anythings after sending subscribe some channels, so we never kill these clients since of timeout. This commit may plant a new bug.

git-hulk:
so Redis updates the last interaction time after writing to the client socket buffer?

ShooterIT:
yep

git-hulk:
seems we didn't have a fast way to do that, any advice to workaround?

ShooterIT:
Yes, there is no a easy way, maybe enable libevent write event and set appropriate water level

git-hulk:
okay

PR #338, opened 2021-07-22
Avoid useless space allocation in SetBit

smartlee:
when exec command `setbit aaa 8180 1` `setbit aaa 8191 1` in order ,SetBit allocates 2046 bytes, but index 1024-2046 is useless and will casue waste of disk space.


git-hulk (approved):

ShooterIT:
LGTM, but don't read.

ShooterIT (approved):

PR #343, opened 2021-07-24
Upgrade version of cpplint to calibrate the code-style of kvrocks.

LykxSassinator:
As discussed in https://github.com/KvrocksLabs/kvrocks/issues/342, i've upgrade the version of cpplint and calibrate the code-style of kvrocks, making consistent with the standard of Google code-style.


ShooterIT:
What is the order of the header files?

1. implemented header file
2. C system files
3. C++ system files
4. Other libraries' .h files
5. Your project's .h files

right?

LykxSassinator:
> What is the order of the header files?
> 
> 1. implemented header file
> 2. C system files
> 3. C++ system files
> 4. Other libraries' .h files
> 5. Your project's .h files
> 
> right?

Yep, those are recommended standard of cpplint.

LykxSassinator (commented):
Double-checked.

src/redis_cmd.cc
```diff
@@ -1,36 +1,38 @@
+#include "redis_cmd.h"
+
 #include <arpa/inet.h>
 #include <fcntl.h>
+#include <glog/logging.h>
```
ShooterIT, line 5: glog libevent rocksdb are third party libraries, should put after C++ std lib?

LykxSassinator: Yep,u r right. It seems that there still exists some mysterious judgements in the `cpplint`, making the incorrect alignments.

Truely, `glog/logging.h` is the thirdparty lib including in `external`. However, the `cpplint` has made a judgement, as a C system lib, if the `glog/logging.h` was included with `<>`. And If we moved it after the c++ stdlib as the thirdparty lib, the checked result was showed as `src/redis_cmd.cc:15:  Found C system header after C++ system header. Should be: redis_cmd.h, c system, c++ system, other.`  But, if we could include `glog/logging.h`, it showed that it was `valid`. What a mysterious result!

The automatic alignments in this PR were automatically generated by the code tool in `VSCODE`. The results showed that the automatic alignments were valid through `cpplint - v1.5.5`, though there still existed some errs, such as the arrangment of `glog`、`libevent` and `rocksdb`.

LykxSassinator: It seems that there still exists errs in `cpplint`, i will check it again.

LykxSassinator: Should we still involved `cpplint` as the default CI code checker?

karelrooted (approved):

git-hulk (approved):

src/redis_cmd.cc
ShooterIT: so could we put the header files of `glog、libevent and rocksdb` together, and after c system header files
change from
```c
#include <arpa/inet.h>
#include <fcntl.h>
#include <glog/logging.h>
#include <sys/socket.h>
```
to
```c
#include <fcntl.h>
#include <sys/socket.h>
#include <arpa/inet.h>
#include <glog/logging.h>
```

LykxSassinator (commented):
Yep, i think this alignment could be an proposal of the code-style standard in KVRocks, based on the code check of `cpplint-1.5.5`. Currently, i would check all alignment again by this standard. It's more recommended that we put an issue to `cpplint`.

ShooterIT:
We still need it, because it can check indent, tab width, max line length...
As you said, `#include <glog/logging.h>` instead of `#include "glog/logging.h"` make it hard to distinguish third party lib with c system lib, maybe it is the problem of kvrocks.

ShooterIT:
Actually, it is not strict in kvrocks, i don't want this PR to take much your precious time, let us just keep one style.

LykxSassinator:
> We still need it, because it can check indent, tab width, max line length...
> As you said, `#include <glog/logging.h>` instead of `#include "glog/logging.h"` make it hard to distinguish third party lib with c system lib, maybe it is the problem of kvrocks.

Yes, it's necessary for us to standardize the code of KVRocks. But `cpplint` is a simple and convinient tool to make code-style check, existes some false positive cases, which should be promoted to be fixed up. What i meant before is that, should we get a better pratice of code-style check tool, recommended in other issue: `https://github.com/PaddlePaddle/Paddle/issues/1184`.

Currently, we could use `cpplint` as the default checker, assited with the standard proposed as ***put the header files of `glog`、`libevent` and `rocksdb` together, and after c system header files***.

By the way, it' ok to do such a job, making my personal contributes to this proj, and it's interesting to find such a false positive phenomenon in `cpplint`, introduced by `cpp-1.5.0`.

PR #344, opened 2021-07-25
Fix the slave won't reconnect when failing to connect with master

### LINKED ISSUE #341: 有时候从服务器无法进行同步
我的主服务器在国内，从服务器在国外，  
从服务器每隔几天就不再进行同步了，只能进行重起才能恢复正常，  
    
主从服务器之间的网络可能不是非常稳定，导致从服务器的同步功能卡住，希望可以解决这个问题

git-hulk:
Fix #341 
Before this commit, replicas can't reconnect when failing to connect with master and stop to sync with master, on master side, feeding replica thread was be terminated. But now, replicas always try to reconnect with master when failing to connect with master.

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -223,7 +223,7 @@ void ReplicationThread::CallbacksStateMachine::Start() {
       continue;
     }
   }
-  if (bev == nullptr) { // failed to connect the master and received the stop signal
+  if (bev == nullptr) {  // failed to connect the master and received the stop signal
     return;
   }
 
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -202,18 +202,28 @@ LOOP_LABEL:
 
 void ReplicationThread::CallbacksStateMachine::Start() {
   int cfd;
+  struct bufferevent *bev = nullptr;
 
   if (handlers_.empty()) {
     return;
   }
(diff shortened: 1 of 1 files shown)


karelrooted (approved):

ShooterIT:
It seems this is an improvement, it is not only `the master was down` but also `can't connect with master because of network`. Before this commit, replicas can't reconnect when failing to connect with master, and stop to sync with master, on master side, feeding replica thread was be terminated. But now, replicas always try to reconnect with master when failing to connect with master.
```
Replicas always try to reconnect with master when failing to connect with master.
```
@git-hulk WDYT?

ShooterIT:
We also can say it is a bug that we don't catch the case.

ShooterIT:
After this PR, maybe this card is not urgent, https://github.com/KvrocksLabs/kvrocks/projects/1#card-62672516

ShooterIT (approved):

git-hulk:
> We also can say it is a bug that we don't catch the case.

aha, from the behavior expected side, I prefer to regard it as a bug since we want to reconnect when failing to connect with the master.

git-hulk:
> After this PR, maybe this card is not urgent, https://github.com/KvrocksLabs/kvrocks/projects/1#card-62672516

yes

PR #347, opened 2021-07-30
Fix master pauses synchronizing with replicas

### LINKED ISSUE #346: Data inconsistency in master-slave replication on certain situation
**Describe the bug**
If we run kvrocks in master-slave structure, the data consistency may not be guaranteed on certain situation. Consider this situation, after all master data have been replicate to slave, we write a new key which will increase the sequence of rocksdb more than 20, then if we stop writing, this new key will not be copied to slave. The key will not be sent，util new data with sequence increment less than 20 or the length of master-slave replication queue exceeds 20. This situation will cause data inconsistency.

 - OS: Linux 3.10.0_3-0-0-36 x86_64
 - Version: all

**To Reproduce**
1. Start two kvrocks servers and establish master-slave replication relationship.
2. After replication, write a hash key with more than 20 f-v pairs, like `hmset hkey f1 v1 f2 v2 ....... f21 v21`, then stop writing.
3. Check replication info on master and slave, then the following result will be gotten:
image
In above picture, master info is on left side, slave is on right side. We can get that `offset=69,lag=0` on master and `slave_repl_offset:47` on slave. `offset` represents the slave's sequence recorded on master which is 69, and `lag=0` represents there is no lag between master and slave. In other words, from the perspective of master, slave and master are synchronized. But, on slave‘s perspective, `slave_repl_offset` is 47 which means slave lags behind master, it is not synchronized with master.
4. Check keys on both master and slave, we will get this:
image
Also, left side is mater and right side is slave in above picture. We can see that the new key `hkey` is not copied to slave.

**Expected behavior**
On slave, `slave_repl_offset` should be 69 too, and key `hkey` should exist.

**Reason analysis**
In `FeedSlaveThread::loop()`, `if (... || latest_seq - batch.sequence <= 20 || batch_list.size() >= 20) ` is adopted to improve data transmission efficiency in incremental replication phase. But, after all data being synchronized, ` latest_seq - batch.sequence <= 20` will cause the bug described above

ShooterIT:
Fixes #346

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -92,7 +92,7 @@ void FeedSlaveThread::loop() {
     }
     updates_in_batches += batch.writeBatchPtr->Count();
     batches_bulk += Redis::BulkString(batch.writeBatchPtr->Data());
-    // 1. we must send the first repl batch, as said above.
+    // 1. We must send the first replication batch, as said above.
     // 2. To avoid frequently calling 'write' system call to send replication stream,
     //    we pack multiple bacthes into one big bulk if possible, and only send once.
     //    But we should send the bulk of batches if its size exceed kMaxDelayBytes,
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -115,6 +115,7 @@ void FeedSlaveThread::loop() {
       }
       is_first_repl_batch = false;
       batches_bulk.clear();
+      if (batches_bulk.capacity() > kMaxDelayBytes * 2) batches_bulk.shrink_to_fit();
       updates_in_batches = 0;
     }
     next_repl_seq_ = batch.sequence + batch.writeBatchPtr->Count();
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -68,15 +68,16 @@ void FeedSlaveThread::loop() {
   // when some seqs might be lost in the middle of the WAL log, so forced to replicate
   // first batch here to work around this issue instead of waiting for enough batch size.
   bool is_first_repl_batch = true;
-  uint32_t yield_milliseconds = 2000;
-  std::vector<std::string> batch_list;
+  uint32_t yield_microseconds = 2 * 1000;
+  std::string batches_bulk;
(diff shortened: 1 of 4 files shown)


src/replication.cc
```diff
@@ -89,26 +90,36 @@ void FeedSlaveThread::loop() {
...
+      auto s = Util::SockSend(conn_->GetFD(), batches_bulk);
+      if (!s.IsOK()) {
+        LOG(ERROR) << "Write error while sending batch to slave: " << s.Msg()
+                   << ". batches: 0x" << Util::StringToHex(batches_bulk);
+        Stop();
+        return;
       }
       is_first_repl_batch = false;
-      batch_list.clear();
+      batches_bulk.clear();
```
ShooterIT, line 117:
```suggestion
      batches_bulk.clear();
      if (batches_bulk.capacity() > kMaxDelayBytes * 2) batches_bulk.shrink_to_fit();
```

src/replication.cc
```diff
@@ -89,26 +90,37 @@ void FeedSlaveThread::loop() {
...
-        auto s = Util::SockSend(conn_->GetFD(), bulk_str);
-        if (!s.IsOK()) {
-          LOG(ERROR) << "Write error while sending batch to slave: " << s.Msg() << ". batch: 0x"
-                     << Util::StringToHex(data);
-          Stop();
-          return;
-        }
+    updates_in_batches += batch.writeBatchPtr->Count();
+    batches_bulk += Redis::BulkString(batch.writeBatchPtr->Data());
+    // 1. we must send the first repl batch, as said above.
```
git-hulk: Upper case: we => We

git-hulk (dismissed):

src/replication.cc
ShooterIT:
```suggestion
    // 1. We must send the first replication batch, as said above.
```

git-hulk (approved):

PR #352, opened 2021-08-10
Support client kill type subcommands

ShooterIT:
When redis-sentinel executes failover, it will send such commands to redis
```
r multi
r slaveof  $ip $port
r config rewrite
r client kill type normal
r client kill type pubsub
r exec
```
From 2.0.2, we support MULTI-EXEC, but don't support `client kill type $type`, so total MULTI-EXEC will be rollbacked when failing to parse unknown `kill type` in client command, and cause redis-sentinel fails to failover. To make it available, we should support client kill type subcommands

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -3365,7 +3365,7 @@ class CommandSlaveOf : public Commander {
         LOG(WARNING) << "MASTER MODE enabled (user request from '" << conn->GetAddr() << "')";
       }
     } else {
-      s = svr->AddMaster(host_, port_);
+      s = svr->AddMaster(host_, port_, false);
       if (s.IsOK()) {
         *output = Redis::SimpleString("OK");
         LOG(WARNING) << "SLAVE OF " << host_ << ":" << port_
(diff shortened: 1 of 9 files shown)


src/redis_cmd.cc
```diff
@@ -4442,7 +4455,7 @@ CommandAttributes redisCommandTable[] = {
     ADD_CMD("compact", 1, "read-only", 0, 0, 0, CommandCompact),
     ADD_CMD("bgsave", 1, "read-only", 0, 0, 0, CommandBGSave),
     ADD_CMD("flushbackup", 1, "read-only", 0, 0, 0, CommandFlushBackup),
-    ADD_CMD("slaveof", 3, "read-only", 0, 0, 0, CommandSlaveOf),
+    ADD_CMD("slaveof", 3, "read-only exclusive", 0, 0, 0, CommandSlaveOf),
```
ShooterIT, line 4458: I think `slaveof` command should have `exclusive` attribute, it is safer, @git-hulk WDYT?

tests/tcl/tests/unit/multi.tcl
```diff
@@ -78,4 +78,17 @@ start_server {tags {"multi"}} {
...
+    test {MULTI-EXEC used in redis-sentinel for failover} {
+        start_server {} {
+            r multi
+            r slaveof [srv -1 host] [srv -1 port]
+            r config rewrite
+            r client kill type normal
+            r client kill type pubsub
+            r exec
+            reconnect
+            assert_equal "slave" [s role]
```
ShooterIT, line 91:
```suggestion
            assert_equal "slave" [s role]

            r multi
            r slaveof no one
            r config rewrite
            r client kill type normal
            r client kill type pubsub
            r exec
            reconnect
            assert_equal "master" [s role]
```

src/redis_cmd.cc
git-hulk: It's fine since the `slaveof` command was a low-frequency operation.

git-hulk (approved):
Nice!

src/redis_cmd.cc
ShooterIT: moreover, we master_host_ without mutex lock

git-hulk: we had the `slaveof_mutex` to protect the master_host_ indeed

ShooterIT: but there is no mutex lock when accessing it

ShooterIT: such as info command

git-hulk: oh, I misunderstand your meaning, it's better to lock that as well.

PR #356, opened 2021-08-25
Auto set master/slave replication by cluster topology

ShooterIT:
Now we set cluster topology by `clusterx setnodes` commands, but replication relationship still need to set manually, actually, we could learn node role from cluster topology, so we will remove its master for master node, and set its master for replica node, users don't need to extra set replication relationship.


git-hulk (approved):
cool!

PR #360, opened 2021-09-10
Fix wrong master port if enabling master-use-repl-port

ShooterIT:
If enabling master-use-repl-port, in replica, its master port is wrong,
it should be master listen port instead of listen port + 1.


git-hulk (approved):

git-hulk:
So the replica would be failed after restarting the server if we rewrite the config file, right?

ShooterIT:
> So the replica would be failed after restarting the server if we rewrite the config file, right?

Yes, and other clients would see wrong master port, for sentinel, it will always try to fix replication config but fails

PR #365, opened 2021-09-23
Metadata and subkey column family share a singleblock cache

caipengbo:
By default, `metadata` Column Family and `subkey` Column Family have their own block cache, so block cache must be allocated in advance for the two CFs. 
It is not possible to predict the usage of the two CFs and allocate the optimal capacity for the two block caches, so we can provide a configuration item that specifies whether the two CFs should share a single block cache (see rocksdb [Wiki](https://github.com/facebook/rocksdb/wiki/Memory-usage-in-RocksDB#indexes-and-filter-blocks)).

Regarding this configuration item, although it is not an item for Rocksdb Option, it is related to Rocksdb, so I gave it a name prefixed with `rocksdb.`.

### DIFF
--- a/src/storage.cc
+++ b/src/storage.cc
@@ -153,15 +153,26 @@ Status Storage::Open(bool read_only) {
   size_t block_size = static_cast<size_t>(config_->RocksDB.block_size);
   size_t metadata_block_cache_size = config_->RocksDB.metadata_block_cache_size*MiB;
   size_t subkey_block_cache_size = config_->RocksDB.subkey_block_cache_size*MiB;
+  bool share_metadata_and_subkey_block_cache = config_->RocksDB.share_metadata_and_subkey_block_cache;
+
   rocksdb::Options options;
   InitOptions(&options);
   CreateColumnFamilies(options);
+
+  std::shared_ptr<rocksdb::Cache> shared_block_cache;
+  if (share_metadata_and_subkey_block_cache) {
+    size_t shared_block_cache_size = metadata_block_cache_size + subkey_block_cache_size;
+    shared_block_cache = rocksdb::NewLRUCache(shared_block_cache_size, -1, false, 0.75);
+  }
+
   rocksdb::BlockBasedTableOptions metadata_table_opts;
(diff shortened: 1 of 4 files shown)


git-hulk:
wow, really a good feature.

src/storage.cc
```diff
@@ -775,4 +787,4 @@ bool Storage::ReplDataManager::FileExists(Storage *storage, const std::string &d
   return crc == tmp_crc;
 }
 
-}  // namespace Engine
+}  // namespace Engine
```
ShooterIT:
```suggestion
}  // namespace Engine

```

src/storage.cc
```diff
@@ -153,15 +153,26 @@ Status Storage::Open(bool read_only) {
...
   size_t metadata_block_cache_size = config_->RocksDB.metadata_block_cache_size*MiB;
   size_t subkey_block_cache_size = config_->RocksDB.subkey_block_cache_size*MiB;
+  bool share_metadata_and_subkey_block_cache = config_->RocksDB.share_metadata_and_subkey_block_cache;
+
   rocksdb::Options options;
   InitOptions(&options);
   CreateColumnFamilies(options);
+
+  std::shared_ptr<rocksdb::Cache> shared_block_cache;
+  if (share_metadata_and_subkey_block_cache) {
```
ShooterIT:
```suggestion
  if (config_->RocksDB.share_metadata_and_subkey_block_cache) {
```

src/storage.cc
```diff
@@ -153,15 +153,26 @@ Status Storage::Open(bool read_only) {
   size_t block_size = static_cast<size_t>(config_->RocksDB.block_size);
   size_t metadata_block_cache_size = config_->RocksDB.metadata_block_cache_size*MiB;
   size_t subkey_block_cache_size = config_->RocksDB.subkey_block_cache_size*MiB;
+  bool share_metadata_and_subkey_block_cache = config_->RocksDB.share_metadata_and_subkey_block_cache;
+
```
ShooterIT:
```suggestion
```

ShooterIT (changes requested):

ShooterIT (approved):

git-hulk (approved):
LGTM

PR #369, opened 2021-10-11
Implement the evalsha and script command

git-hulk:
This PR implements the `evalsha` and `script`(except debug and kill subcommand) command, `script kill` and Lua time limit would be supported in the next one to keep the PR simple.


src/server.h
```diff
@@ -158,6 +170,9 @@ class Server {
   std::mutex last_random_key_cursor_mu_;
 
   lua_State *lua_;
+  std::mutex lua_scripts_mu_;
+  std::unordered_map<std::string, std::string> lua_scripts_;
```
ShooterIT: Why don't we store the lua scripts into rocksdb instead of memory? it may cost much more memory.

And we also regard lua scripts as key value pairs in propagate CF, so we don't handle it in replication.

git-hulk: As you say, it's a good idea to persist those scripts on rocksdb.

src/scripting.cc
```diff
@@ -676,7 +784,7 @@ Status createFunction(lua_State *lua, const std::string &body, std::string *sha)
     return Status(Status::NotOK,
                   "Error running script (new function): " + errMsg + "\n");
   }
-  *sha = funcdef.substr(2);
+  srv->ScriptSet(*sha, body);
```
ShooterIT, line 795: please add comments here, such as `store lua script into xx cf of rocksdb and will auto propagate it`

src/server.cc
```diff
@@ -1222,3 +1222,76 @@ Status Server::LookupAndCreateCommand(const std::string &cmd_name,
...
+}
+
+void Server::ScriptReset() {
+  Lua::DestroyState(lua_);
+  lua_ = Lua::CreateState();
+}
+
+void Server::ScriptFlush() {
+  auto cf = storage_->GetCFHandle(Engine::kPropagateColumnFamilyName);
+  storage_->DeleteAll(rocksdb::WriteOptions(), cf);
```
ShooterIT: we shouldn't delete all, we just want to delete keys that have `f_` prefix

src/server.cc
```diff
@@ -1222,3 +1222,76 @@ Status Server::LookupAndCreateCommand(const std::string &cmd_name,
...
+  auto s = storage_->GetDB()->Get(rocksdb::ReadOptions(), cf, funcname, body);
+  if (!s.ok()) {
+    if (s.IsNotFound()) return Status(Status::NotFound);
+    return Status(Status::NotOK, s.ToString());
+  }
+  return Status::OK();
+}
+
+void Server::ScriptSet(const std::string &sha, const std::string &body) {
+  std::string funcname = "f_" + sha;
```
ShooterIT: maybe we create one const variable(`lua_script_prefix`) instead of `f_` here.

i think `lua_f_` is better than `f_`, WDYT?

src/storage.cc
```diff
@@ -30,6 +30,9 @@ const char *kPubSubColumnFamilyName = "pubsub";
 const char *kZSetScoreColumnFamilyName = "zset_score";
 const char *kMetadataColumnFamilyName = "metadata";
 const char *kSubkeyColumnFamilyName = "default";
+const char *kPropagateColumnFamilyName = "propagate";
+
+const char *kPropagateTypeKey = "lua";
```
ShooterIT: kPropagateTypeKey -> kPropagateLua？

src/replication.cc
```diff
@@ -917,11 +917,20 @@ rocksdb::Status ReplicationThread::ParseWriteBatch(const std::string &batch_stri
...
   if (!status.ok()) return status;
-  if (write_batch_handler.IsPublish()) {
-    srv_->PublishMessage(write_batch_handler.GetPublishChannel().ToString(),
-                         write_batch_handler.GetPublishValue().ToString());
+  switch (write_batch_handler.Type()) {
+    case kBatchTypePublish:
+      srv_->PublishMessage(write_batch_handler.Key(), write_batch_handler.Value());
+      break;
+    case kBatchTypePropagate:
+      if (write_batch_handler.Key() == "command") {
```
ShooterIT: still `command`？

tests/tcl/tests/unit/scripting.tcl
```diff
@@ -42,21 +42,40 @@ start_server {tags {"scripting"}} {
...
+
+    test {EVALSHA - Do we get an error on invalid SHA1?} {
+        catch {r evalsha NotValidShaSUM 0} e
+        set _ $e
+    } {ERR NOSCRIPT*}
+
+    test {EVALSHA - Do we get an error on non defined SHA1?} {
+        catch {r evalsha ffd632c7d33e571e9f24556ebed26c3479a87130 0} e
+        set _ $e
+    } {ERR NOSCRIPT*}
```
ShooterIT, line 70: i think, in lua test, you should add mater-slaves replication test case.

src/server.cc
git-hulk: good catch, it's my carelessness

src/server.cc
git-hulk: both were ok to me, so pick the `lua_f`

tests/tcl/tests/unit/scripting.tcl
git-hulk: There's a little tricky on the script command but at least we have the test case.

tests/tcl/tests/support/util.tcl
```diff
@@ -95,7 +95,7 @@ proc wait_for_sync r {
 
 proc wait_for_ofs_sync {r1 r2} {
     wait_for_condition 50 100 {
-        [status $r1 master_repl_offset] eq [status $r2 master_repl_offset]
```
ShooterIT, line 98: i think it is time we should support `master_repl_offset` in INFO command output

tests/tcl/tests/unit/scripting.tcl
```diff
@@ -252,4 +271,104 @@ start_server {tags {"scripting"}} {
...
+        set master_port [srv -1 port]
+        set slave [srv 0 client]
+        test {SCRIPTING: setup slave} {
+            $slave slaveof $master_host $master_port
+            wait_for_condition 50 100 {
+                [s 0 master_link_status] eq {up}
+            } else {
+                fail "Replication not started."
+            }
+        }
```
ShooterIT:
```suggestion
        $slave slaveof $master_host $master_port
        wait_for_sync $slave
```

docs/support-commands.md
```diff
@@ -148,6 +148,13 @@
 **NOTE : String and Bitmap is different type in kvrocks, so you can't do bit with string, vice versa.**
 
 
+## Script Commands
+
+| Command   | Supported OR Not | Desc |
+| --------- | ---------------- | ---- |
+| eval      | √                |      |
+| evalsha   | √                |      |
+| script    | √                |      |
```
ShooterIT: script kill and debug subcommand are not supported

src/compact_filter.h
```diff
@@ -71,6 +71,26 @@ class SubKeyFilterFactory : public rocksdb::CompactionFilterFactory {
   Engine::Storage *stor_ = nullptr;
 };
 
+class TransitFilter : public rocksdb::CompactionFilter {
```
ShooterIT: i accept 'propagate' name, so i think here we should also can use PropagateFilter, WDYT?
especially, you use `CompactTransitFiles` for `pubsub` and `propagate` CF, it is confused.

src/compact_filter.h
```diff
@@ -71,6 +71,26 @@ class SubKeyFilterFactory : public rocksdb::CompactionFilterFactory {
...
+ public:
+  const char *Name() const override { return "TransitFilter"; }
+  bool Filter(int level, const Slice &key, const Slice &value,
+              std::string *new_value, bool *modified) const override {
+    // only filter the propagating command here
```
ShooterIT:
```suggestion
    // We propagate Lua commands which don't store data,
    // just in order to implement updating Lua state.
```
How about above comments

src/redis_cmd.cc
```diff
@@ -4473,6 +4524,8 @@ CommandAttributes redisCommandTable[] = {
     ADD_CMD("clusterx", -2, "cluster no-script", 0, 0, 0, CommandClusterX),
 
     ADD_CMD("eval", -3, "exclusive write no-script", 0, 0, 0, CommandEval),
+    ADD_CMD("evalsha", -3, "exclusive write no-script", 0, 0, 0, CommandEvalSHA),
+    ADD_CMD("script", -2, "exclusive read-only no-script", 0, 0, 0, CommandScript),
```
ShooterIT:
```suggestion
    ADD_CMD("script", -2, "exclusive no-script", 0, 0, 0, CommandScript),
```
remove `read-only` attribute ?

tests/tcl/tests/tmp/.gitignore
```diff
@@ -1 +0,0 @@
-*
```
ShooterIT, line 1: Why do you delete this file

src/storage.h
```diff
@@ -21,13 +21,19 @@ enum ColumnFamilyID{
...
 };
 
 namespace Engine {
 extern const char *kPubSubColumnFamilyName;
 extern const char *kZSetScoreColumnFamilyName;
 extern const char *kMetadataColumnFamilyName;
 extern const char *kSubkeyColumnFamilyName;
+extern const char *kPropagateColumnFamilyName;
+
+extern const char *kPropagateTypeLua;
```
ShooterIT:
```suggestion
extern const char *kPropagateLuaCommand;
```
this is a key name, i think `type` word is not clear.

`key` | `value`
lua -> script load
lua -> script flush

src/server.cc
```diff
@@ -1222,3 +1222,76 @@ Status Server::LookupAndCreateCommand(const std::string &cmd_name,
...
+  auto propagateCf = storage_->GetCFHandle(Engine::kPropagateColumnFamilyName);
+  batch.Put(propagateCf, key, value);
+  auto s = storage_->Write(rocksdb::WriteOptions(), &batch);
+  if (!s.ok()) {
+    return Status(Status::NotOK, s.ToString());
+  }
+  return Status::OK();
+}
+
+Status Server::Propagate(const std::string &type, const std::vector<std::string> &tokens) {
```
```suggestion
// Generally, we store data into rocksdb and just replicate WAL instead of propagating
// commands. But sometimes, we need to update inner states or do special operations
// for specific commands, such as `script flush`.
// channel: we put the same function commands into one channel to handle uniformly 
// tokens: the serialized commands
Status Server::Propagate(const std::string &channel, const std::vector<std::string> &tokens) {
```

src/server.cc
```diff
@@ -1222,3 +1222,76 @@ Status Server::LookupAndCreateCommand(const std::string &cmd_name,
...
+  return WriteToPropagateCF(type, value);
+}
+
+Status Server::replayScriptCommand(const std::vector<std::string> &tokens) {
+  auto subcommand = Util::ToLower(tokens[1]);
+  if (subcommand == "flush") {
+    ScriptReset();
+    return Status::OK();
+  } else if (subcommand == "load" && tokens.size() == 3) {
+    // no need to replay the script load command, eval/evalsha would load the script from
```
ShooterIT: so i think we don't need to propagate this subcommand.

src/server.cc
```diff
@@ -1222,3 +1222,76 @@ Status Server::LookupAndCreateCommand(const std::string &cmd_name,
...
+    // db if the Lua stack was not exists.
+    return Status::OK();
+  }
+  return Status(Status::NotOK, "Unknown SCRIPT subcommand or wrong # of args");
+}
+
+Status Server::ReplayCommand(const std::vector<std::string> &tokens) {
```
ShooterIT: i don't think `ReplayCommand` is a good name. it seems loading data, we aim to execute propagated commands on slaves. how about `ExecPropagatedCommand`, `ExecPropagatedLuaCommand`

ShooterIT (commented):
A lot of comments, maybe it makes no sense, please check
i think we should add more comments for some special operations, otherwise, it is hard to understand.

tests/tcl/tests/support/util.tcl
git-hulk: You're right, how about improving this in the next PR?

tests/tcl/tests/tmp/.gitignore
git-hulk: my mistake

src/storage.h
git-hulk: agree

src/server.cc
git-hulk: `ExecPropagatedCommand ` is better than ReplayCommand, but how about `ExecPropagatedLuaCommand` to ExecPropagatedScriptCommand` should be more preciously?

src/server.cc
git-hulk: this command wasn't propagated now, removed this condition

src/redis_cmd.cc
git-hulk: we allow executing `script exists` command on slave, so it should perform as `read-only`

src/compact_filter.h
git-hulk: good catch.

Initially, I use the same compact filter(pubsub compact filter) between propagate and pubsub, so renamed it as `transit` compact filter, but didn't revert this name after splitting.

git-hulk:
> A lot of comments, maybe it makes no sense, please check i think we should add more comments for some special operations, otherwise, it is hard to understand.

most advice totally made sense and was great, I have fixed them but some comments. You can have a look again.

src/server.cc
ShooterIT: yes, `ExecPropagatedScriptCommand` is better

tests/tcl/tests/support/util.tcl
ShooterIT: OK

src/redis_cmd.cc
ShooterIT: if no `read-only` attribute, we don't consider it as write command, right?

src/storage.h
```diff
@@ -21,13 +21,19 @@ enum ColumnFamilyID{
...
 };
 
 namespace Engine {
 extern const char *kPubSubColumnFamilyName;
 extern const char *kZSetScoreColumnFamilyName;
 extern const char *kMetadataColumnFamilyName;
 extern const char *kSubkeyColumnFamilyName;
+extern const char *kPropagateColumnFamilyName;
+
+extern const char *kPropagateLuaCommand;
```
ShooterIT: my bad, `kPropagateScriptCommand` seems better?

src/server.cc
ShooterIT: i changed parameter `const std::string &type` into `const std::string &channel`, and in comments, i prefer channel, WDYT?

src/storage.h
git-hulk: oh no, it's my fault, it should be `kPropagateScriptCommand `

src/redis_cmd.cc
git-hulk: actually, some subcommands weren't, but it's ok that we only allow the `exists` subcommand when the role was a slave.

src/server.cc
git-hulk: Fixed

ShooterIT (approved):
Cool job,  cheers!

PR #373, opened 2021-10-17
Set rocksdb.share_metadata_and_subkey_block_cache to yes by default

### LINKED ISSUE #372: [NEW] Enable share block cache between  the metadata and subkey column family by default
Kvrocks supported the share block cache between the metadata and subkey column family, it's really a good point to speed up the lookup query. So it should be enabled by default? since it hurts nothing, also most users should NOT care about how to allocate the ratio of block cache between the metadata and subeky column family.

caipengbo:
Resolved #372


git-hulk:
cool,  really speedy.

ShooterIT (approved):

PR #375, opened 2021-10-19
Support executing INFO command during restoring database

ShooterIT:
Currently, we can't execute INFO command during restoring database on kvrocks,
but redis can. Kvrocks returns error messages directly, it is not easy for some
SDK/proxy/tools to check loading or other status, so this commit implements.

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -799,6 +799,11 @@ void Server::GetCommandsStatsInfo(std::string *info) {
   *info = string_stream.str();
 }
 
+// WARNING: we must not access DB(i.e.RocksDB) when server is loading since
+// DB is closed and the pointer is invalid. Server may crash if we access DB
+// during loading.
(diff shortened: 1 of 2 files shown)


git-hulk (dismissed):

src/server.cc
```diff
@@ -873,7 +882,9 @@ void Server::GetInfo(const std::string &ns, const std::string &section, std::str
       string_stream << "used_disk_percent: " << used_disk_percent << "%\r\n";
     }
   }
-  if (all || section == "rocksdb") {
+
+  // In keyspace section, we access DB, so we can't do that when loading
```
```suggestion
  // In rocksdb section, we access DB, so we can't do that when loading
```

ShooterIT:
Tests will be added in the next PR.

git-hulk (approved):

PR #378, opened 2021-10-22
Improve the description of Perflog in default config file

caipengbo:
I added `profiling-sample-commands` config item and some comments about `PERFLOG` in the default config file. 

This will help users understand how to use `PERFLOG` to troubleshoot their performance problems.

I also checked the `profiling- samp-commands` in code logic, which will returns an error if a command is not supported by Kvrocks. And fixed a small bug about `stats_dump_period_sec` config item.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -215,7 +215,7 @@ max-backup-keep-hours 24
 
 ################################## SLOW LOG ###################################
 
-# The Kvrocks Slow Log is a system to log queries that exceeded a specified
+# The Kvrocks Slow Log is a mechanism to log queries that exceeded a specified
 # execution time. The execution time does not include the I/O operations
 # like talking with the client, sending the reply and so forth,
 # but just the time needed to actually execute the command (this is the only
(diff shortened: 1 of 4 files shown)


git-hulk:
@caipengbo really good to make the perflog description more clearly. The unit test failure should be caused by the empty perf command string which we didn't ignore.

caipengbo:
The reason about UT failed is that I forgot to call `PopulateCommands` before calling `Redis::IsCommandExists()`, causing `original_commands` to be empty.

git-hulk (dismissed):

tests/config_test.cc
```diff
@@ -9,6 +9,8 @@ TEST(Config, GetAndSet) {
   Config config;
 
   config.Load(path);
+  Redis::InitCommandsTable();
```
git-hulk, line 14:
```suggestion
  // Config set need accessing commands, so we should init and populate
  // the command table here.
  Redis::InitCommandsTable();
```

git-hulk (dismissed):

kvrocks.conf
```diff
@@ -265,6 +248,46 @@ slowlog-max-len 128
 #       They do not enable continuous liveness pings back to your supervisor.
 supervised no
 
+################################## PERF LOG ###################################
+
+# The Kvrocks Perf Log is a system to log queries' performance context that
```
ShooterIT:
```suggestion
# The Kvrocks Perf Log is a mechanism to log queries' performance context that
```
how about system -> mechanism?

caipengbo: Cool, `SLOWLOG` could also be called that.

ShooterIT (approved):

git-hulk (approved):

PR #213, opened 2021-03-29
Add qps and input/output kbps for INFO command

ChrisZMF:
Add QPS and  instantaneous network flow statistics.

PR #387, opened 2021-11-04
Add rocksdb ops statistics in info cmd

caipengbo:
Kvrocks will encode the Redis command as key-value and write it to Rocksdb. and actually Kvrocks has read and write amplification. We cannot get the actual load of Rocksdb from the `instantaneous_ops_per_sec` information introduced by #213. Although Kvrocks provides the `stats` command to obtain Rocksdb statistics, it is not very intuitive to see the instantaneous load of Rocksdb.

Therefore, I added the ops statistics about Rocksdb, including read, write and iterator seek/prev/ next, so that users or op can intuitively see the stress of the storage engine to help them optimize their workload.

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -478,15 +478,17 @@ std::atomic<uint64_t> *Server::GetClientID() {
   return &client_id_;
 }
 
-void Server::recordInstantaneousMetric() {
+void Server::recordInstantaneousMetrics() {
   auto rocksdb_stats = storage_->GetDB()->GetDBOptions().statistics;
   stats_.TrackInstantaneousMetric(STATS_METRIC_COMMAND, stats_.total_calls);
   stats_.TrackInstantaneousMetric(STATS_METRIC_NET_INPUT, stats_.in_bytes);
   stats_.TrackInstantaneousMetric(STATS_METRIC_NET_OUTPUT, stats_.out_bytes);
-  stats_.TrackInstantaneousMetric(STATS_METRIC_ROCKSDB_WRITTEN,
+  stats_.TrackInstantaneousMetric(STATS_METRIC_ROCKSDB_PUT,
     rocksdb_stats->getTickerCount(rocksdb::Tickers::NUMBER_KEYS_WRITTEN));
-  stats_.TrackInstantaneousMetric(STATS_METRIC_ROCKSDB_READ,
+  stats_.TrackInstantaneousMetric(STATS_METRIC_ROCKSDB_GET,
     rocksdb_stats->getTickerCount(rocksdb::Tickers::NUMBER_KEYS_READ));
+  stats_.TrackInstantaneousMetric(STATS_METRIC_ROCKSDB_MULTIGET,
+    rocksdb_stats->getTickerCount(rocksdb::Tickers::NUMBER_MULTIGET_KEYS_READ));
   stats_.TrackInstantaneousMetric(STATS_METRIC_ROCKSDB_SEEK,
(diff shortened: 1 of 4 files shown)


ShooterIT:
great idea!
could you add some tcl tests?

src/server.cc
```diff
@@ -478,6 +478,23 @@ std::atomic<uint64_t> *Server::GetClientID() {
   return &client_id_;
 }
 
+void Server::recordInstantaneousMetric() {
```
git-hulk: Should be `recordInstantaneousMetrics`?  since we collect multi-metrics here.

caipengbo: Yes

caipengbo:
I wanted to add some TCL test cases, but found that in Rocksdb, the statistic `NUMBER_KEYS_READ` is always 0.

git-hulk:
> I wanted to add some TCL test cases, but found that in Rocksdb, the statistic `NUMBER_KEYS_READ` is always 0.

NUMBER_KEYS_READ would increase after getting a value from rocksdb, see the [db_impl.cc](https://github.com/facebook/rocksdb/blob/29102641ddf4632bddf34bbd35de9637ea526dee/db/db_impl/db_impl.cc#L1896) for more details.

caipengbo:
> NUMBER_KEYS_READ would increase after getting a value from rocksdb, see the db_impl.cc for more details.
Yeah, I read the details, but test it still 0.

caipengbo:
Because `MultiGet` is used in Kvrocks, its corresponding statistic is `NUMBER_MULTIGET_KEYS_READ`.
I add it together with `NUMBER_KEYS_READ` to measure the read load on Rocksdb.

In addition, I add some tcl tests.

git-hulk (dismissed):

src/server.cc
```diff
@@ -605,6 +622,12 @@ void Server::GetRocksDBInfo(std::string *info) {
   string_stream << "num_background_errors:" << num_backgroud_errors << "\r\n";
   string_stream << "flush_count:" << storage_->GetFlushCount()<< "\r\n";
   string_stream << "compaction_count:" << storage_->GetCompactionCount()<< "\r\n";
+  string_stream << "written_per_sec:" << stats_.GetInstantaneousMetric(STATS_METRIC_ROCKSDB_PUT) << "\r\n";
```
ShooterIT: why did you use **written** word?

how about
`put_per_sec` `get_per_sec`
or 
`put_ops_per_sec` `get_ops_per_sec`
especially, for `next_per_sec` and `pre_per_sec`, i prefer to use `next_ops_per_sec`, `prev_ops_per_sec`, just using `next` or `pre` is not clear i think.

caipengbo: `ops` means `operation per second`, same thing as `per_sec`.
The words of `put` or `get` is great. In fact, the intention of using `written` is to be consistent with rocksdb statistics

ShooterIT: `ops` just is the short of operations
the reason why I want to add `ops` is that `pre` or `next` is not clear, the `ops` can know they are operations.
Currently, both are ok for me, @git-hulk WDYT?

also see `instantaneous_ops_per_sec`

caipengbo: Is the abbr `ops` confusing? The `ops` in `instantaneous_ops_per_sec` confused me at start.

git-hulk: I think `put_per_sec` should be better than `get_ops_per_sec` since the `ops` may be confused with the instantaneous_ops_per_sec.

ShooterIT (approved):

git-hulk (approved):

PR #395, opened 2021-11-17
Add some rocksdb new feature

### LINKED ISSUE #209: RocksDB optimization
## Compression type
- Support multiple compression types

## Parameters
- enable enable_pipelined_write if having more CPUs
- use index_type: kTwoLevelIndexSearch to make index and filter blocks smaller
- enable pin_top_level_index_and_filter, pin_l0_filter_and_index_blocks_in_cache
- for cache_index_and_filter_blocks_with_high_priority, to avoid too high
- NewBloomFilterPolicy(bits_per_key, false) i.e. BloomFilterPolicy::kAutoBloom
- data_block_index_type: BlockBasedTableOptions::kDataBlockBinaryAndHash
- Set BlockBasedTableOptions::optimize_filters_for_memory to true, jemalloc_usable_size is safe if no bugs, redis adopts it
-  memtable_prefix_bloom_size_ratio and memtable_whole_key_filtering, to enable memtable filter.
- CompressionOptions::max_dict_bytes for lz4 and zstd

## Data partition  by ColumnFamily
- ~~One type one ColumnFamily~~

### LINKED ISSUE #366: Does kvrocks support ROCKSDB's native KEY/VALUE separation, BLOB STORE?
The KVROCKS I use has 100 billion records ， VALUE IS 4KB BYTES ， Reading speed ( GET )  is very slow

### LINKED ISSUE #384: [BUG] The configuration item rocksdb.compression cannot take effect
Kvrocks has a configuration item called `rocksdb.compression`, which is assigned to `rocksdb::Options`:
https://github.com/KvrocksLabs/kvrocks/blob/2e43741ddedc06f0376d71dc385e9d53a24b3204/src/storage.cc#L87

But Kvrocks calls rocksdb's built-in optimization function:
https://github.com/KvrocksLabs/kvrocks/blob/2e43741ddedc06f0376d71dc385e9d53a24b3204/src/storage.cc#L80

This function will set `compression_per_level`. Like the [Rocksdb Wiki](https://github.com/facebook/rocksdb/wiki/Compression#configuration) says, when `options.compression_per_level` is used, `options.compression` is ignored.

Rocksdb's built-in optimization function selects the best compression algorithm. **So do we need to remove `rocksdb.compression` configuration item in Kvrocks?**

caipengbo:
#209 proposed some optimization points for RocksDB, and I tried to apply them to Kvrocks through some research.

Here are my optimizations:

1. Dynamic resize block and SST introduced by #120  and #289  to deal with the problem of large index/filter blocks.
    
    This mechanism resize `target_file_size_base` and `block_size` by calculating the average key-value size:

     https://github.com/KvrocksLabs/kvrocks/blob/782c106d4513eb79a94431f0750a626c26b1fdd2/src/server.cc#L1050

      It uses the compressed total size to calculate the average key-value, which can easily be less than 128, resulting in `target_file_size_base` always be 16. When there is a large amount of data, a large number of SST files are generated. Too many files may cause problems.

    Therefore, I use [Partitioned Index/Filter](https://github.com/facebook/rocksdb/wiki/Partitioned-Index-Filters) to deal with large Index/Filter blocks and remove `autoResizeBlockAndSST()` . Through some test, I found that when our files were large(128M), we did not generate large index/filter blocks to increase the long tail latency.

2. Add Rocksdb's new feature integrated [BlobDB](https://github.com/facebook/rocksdb/wiki/BlobDB)(Since 6.18.0) for large value scenarios. Although the integrated BlobDB has not been announced for production yet, I think it can be added for users to test.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -350,7 +350,7 @@ void Config::initFieldCallback() {
         if (!RocksDB.enable_blob_files) {
           return Status(Status::NotOK, "Must set rocksdb.enable_blob_files to yes first.");
         }
-        std::string enable_blob_garbage_collection = v == "yes" ? "true" : "false";                               
+        std::string enable_blob_garbage_collection = v == "yes" ? "true" : "false";
         return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), enable_blob_garbage_collection);
       }},
       {"rocksdb.blob_garbage_collection_age_cutoff", [this](Server* srv, const std::string &k,
--- a/src/config.cc
+++ b/src/config.cc
@@ -18,6 +18,9 @@
 #include "log_collector.h"
 
 const char *kDefaultNamespace = "__namespace";
+
+const char *errNotEnableBlobDB = "Must set rocksdb.enable_blob_files to yes first.";
+
 configEnum compression_type_enum[] = {
     {"no", rocksdb::CompressionType::kNoCompression},
     {"snappy", rocksdb::CompressionType::kSnappyCompression},
@@ -227,6 +230,7 @@ void Config::initFieldCallback() {
     if (!srv) return Status::OK();  // srv is nullptr when load config from file
     return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), v);
   };
+
   std::map<std::string, callback_fn> callbacks = {
       {"dir", [this](Server* srv,  const std::string &k, const std::string& v)->Status {
         db_dir = dir + "/db";
@@ -326,20 +330,20 @@ void Config::initFieldCallback() {
       }},
       {"rocksdb.enable_blob_files", [](Server* srv, const std::string &k, const std::string& v)->Status {
         if (!srv) return Status::OK();
-        std::string enable_blob_files = v == "yes" ? "true" : "false";
+        std::string enable_blob_files = RocksDB.enable_blob_files ? "true" : "false";
         return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), enable_blob_files);
       }},
       {"rocksdb.min_blob_size", [this](Server* srv, const std::string &k, const std::string& v)->Status {
         if (!srv) return Status::OK();
         if (!RocksDB.enable_blob_files) {
-          return Status(Status::NotOK, "Must set rocksdb.enable_blob_files to yes first.");
+          return Status(Status::NotOK, errNotEnableBlobDB);
         }
         return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), v);
       }},
       {"rocksdb.blob_file_size", [this](Server* srv, const std::string &k, const std::string& v)->Status {
         if (!srv) return Status::OK();
         if (!RocksDB.enable_blob_files) {
-          return Status(Status::NotOK, "Must set rocksdb.enable_blob_files to yes first.");
+          return Status(Status::NotOK, errNotEnableBlobDB);
         }
         return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), v);
       }},
@@ -347,7 +351,7 @@ void Config::initFieldCallback() {
                                                         const std::string& v)->Status {
         if (!srv) return Status::OK();
         if (!RocksDB.enable_blob_files) {
-          return Status(Status::NotOK, "Must set rocksdb.enable_blob_files to yes first.");
+          return Status(Status::NotOK, errNotEnableBlobDB);
         }
         std::string enable_blob_garbage_collection = v == "yes" ? "true" : "false";
         return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), enable_blob_garbage_collection);
@@ -356,7 +360,7 @@ void Config::initFieldCallback() {
                                                             const std::string& v)->Status {
         if (!srv) return Status::OK();
         if (!RocksDB.enable_blob_files) {
-          return Status(Status::NotOK, "Must set rocksdb.enable_blob_files to yes first.");
+          return Status(Status::NotOK, errNotEnableBlobDB);
         }
         int val;
         try {
--- a/src/config.cc
+++ b/src/config.cc
@@ -328,7 +328,7 @@ void Config::initFieldCallback() {
         return srv->storage_->SetDBOption(trimRocksDBPrefix(k),
                                           std::to_string(RocksDB.max_total_wal_size * MiB));
       }},
-      {"rocksdb.enable_blob_files", [](Server* srv, const std::string &k, const std::string& v)->Status {
+      {"rocksdb.enable_blob_files", [this](Server* srv, const std::string &k, const std::string& v)->Status {
         if (!srv) return Status::OK();
         std::string enable_blob_files = RocksDB.enable_blob_files ? "true" : "false";
         return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), enable_blob_files);
--- a/src/config.cc
+++ b/src/config.cc
@@ -128,6 +128,11 @@ Config::Config() {
        false, new IntField(&RocksDB.level0_slowdown_writes_trigger, 20, 1, 1024)},
       {"rocksdb.level0_stop_writes_trigger",
        false, new IntField(&RocksDB.level0_stop_writes_trigger, 40, 1, 1024)},
+      {"rocksdb.enable_blob_files", false, new YesNoField(&RocksDB.enable_blob_files, false)},
+      {"rocksdb.min_blob_size", false, new IntField(&RocksDB.min_blob_size, 4096, 0, INT_MAX)},
+      {"rocksdb.blob_file_size", false, new IntField(&RocksDB.blob_file_size, 128, 0, INT_MAX)},
+      {"rocksdb.enable_blob_garbage_collection", false, new YesNoField(&RocksDB.enable_blob_garbage_collection, true)},
+      {"rocksdb.blob_garbage_collection_age_cutoff", false, new IntField(&RocksDB.blob_garbage_collection_age_cutoff, 128, 0, 100)}
   };
   for (const auto &wrapper : fields) {
     auto field = wrapper.field;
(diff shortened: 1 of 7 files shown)


src/config.cc
```diff
@@ -318,6 +324,67 @@ void Config::initFieldCallback() {
         return srv->storage_->SetDBOption(trimRocksDBPrefix(k),
                                           std::to_string(RocksDB.max_total_wal_size * MiB));
       }},
+      {"rocksdb.enable_blob_files", [](Server* srv, const std::string &k, const std::string& v)->Status {
+        if (!srv) return Status::OK();
+        std::string enable_blob_files = v == "yes" ? "true" : "false";
```
ShooterIT: @git-hulk currently, config parser could ignore case, i.e. YES will be changed into yes?

src/server.cc
```diff
@@ -537,12 +537,6 @@ void Server::cron() {
       }
     }
 
-    // check every 30 minutes
```
ShooterIT, line 540: I agree, we should discard that

src/storage.cc
```diff
@@ -115,6 +115,16 @@ Status Storage::SetColumnFamilyOption(const std::string &key, const std::string
   return Status::OK();
 }
 
+Status Storage::SetColumnFamilyOption(const std::string &cf_name, const std::string &key,
```
ShooterIT: how about
`Status Storage::SetColumnFamilyOption(const std::vector<std::string> &cf_names, const std::string &key,`
to handle multi CF setting options

src/storage.cc
```diff
@@ -171,7 +181,12 @@ Status Storage::Open(bool read_only) {
   }
 
   rocksdb::BlockBasedTableOptions metadata_table_opts;
-  metadata_table_opts.filter_policy.reset(rocksdb::NewBloomFilterPolicy(10, true));
+  metadata_table_opts.format_version = 5;
```
ShooterIT: Although rocksdb will do some optimization if format_version is 5, i still want to set those options explicitly.

Otherwise, i think we should enable `pin_l0_filter_and_index_blocks_in_cache`, WDYT @git-hulk

src/storage.cc
```diff
@@ -184,9 +199,21 @@ Status Storage::Open(bool read_only) {
...
+  metadata_opts.enable_blob_files = config_->RocksDB.enable_blob_files;
+  metadata_opts.min_blob_size = config_->RocksDB.min_blob_size;
+  metadata_opts.blob_file_size = config_->RocksDB.blob_file_size;
+  metadata_opts.blob_compression_type = static_cast<rocksdb::CompressionType>(config_->RocksDB.compression);
+  metadata_opts.enable_blob_garbage_collection = config_->RocksDB.enable_blob_garbage_collection;
+  metadata_opts.blob_garbage_collection_age_cutoff = config_->RocksDB.blob_garbage_collection_age_cutoff / 100;
 
   rocksdb::BlockBasedTableOptions subkey_table_opts;
-  subkey_table_opts.filter_policy.reset(rocksdb::NewBloomFilterPolicy(10, true));
+  subkey_table_opts.format_version = 5;
```
ShooterIT: I find metadata CF's some options are the same as and subkey CF, could we create a new function to set both of them options to make code clean

src/config.cc
git-hulk: The config parser only ignores the case when using YesNoField::Set. We can use the config's member `enable_blob_files` here since the value has been parsed.

src/storage.cc
git-hulk: I also agree that should enable `pin_l0_filter_and_index_blocks_in_cache` by default which would bring better performance with a little pin memory cost.

src/config.cc
```diff
@@ -318,6 +324,67 @@ void Config::initFieldCallback() {
...
+        Status s = srv->storage_->SetColumnFamilyOption(Engine::kMetadataColumnFamilyName, trimRocksDBPrefix(k),
```
git-hulk: How about using `set_cf_option_cb` to apply the column family option?

git-hulk:
Really good feature, some users also look forward to the blob store. https://github.com/KvrocksLabs/kvrocks/issues/366

src/storage.cc
caipengbo: Good idea

src/config.cc
caipengbo: `set_cf_option_cb` will set ALL cf

git-hulk: yes, it would be confusing if some options apply to all column families and some not, To be more clear, what I mean is that maybe we should keep consistent. How do you think?

caipengbo: Do you mean to apply this option to all CF?

git-hulk: Yes, it should make sense since it wouldn't separate the key-value if the size didn't exceed the blob threshold, right?

git-hulk: Another question, those options should NOT take effect if we didn't enable the blob store?

caipengbo: Indeed, I will add a judgment here.

caipengbo:
By the way, I will solve #384

src/storage.cc
```diff
@@ -77,14 +95,24 @@ void Storage::InitOptions(rocksdb::Options *options) {
   // See: https://github.com/facebook/rocksdb/wiki/Statistics
   options->statistics = rocksdb::CreateDBStatistics();
   options->stats_dump_period_sec = config_->RocksDB.stats_dump_period_sec;
-  options->OptimizeLevelStyleCompaction();
```
caipengbo, line 80: Remove `OptimizeLevelStyleCompaction` and put some options to `Storage`

src/config.cc
```diff
@@ -318,6 +324,53 @@ void Config::initFieldCallback() {
...
       }},
+      {"rocksdb.enable_blob_files", [](Server* srv, const std::string &k, const std::string& v)->Status {
+        if (!srv) return Status::OK();
+        std::string enable_blob_files = v == "yes" ? "true" : "false";
+        return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), enable_blob_files);
+      }},
+      {"rocksdb.min_blob_size", [this](Server* srv, const std::string &k, const std::string& v)->Status {
+        if (!srv) return Status::OK();
+        if (!RocksDB.enable_blob_files) {
+          return Status(Status::NotOK, "Must set rocksdb.enable_blob_files to yes first.");
```
git-hulk: Can use a const string to avoid copying this error message

src/config.cc
git-hulk: @caipengbo Need lowercase the value or use 

std::string opt_value = RocksDB.enable_blob_files ? "true":"false";

git-hulk (dismissed):

src/config.cc
```diff
@@ -318,6 +328,53 @@ void Config::initFieldCallback() {
         return srv->storage_->SetDBOption(trimRocksDBPrefix(k),
                                           std::to_string(RocksDB.max_total_wal_size * MiB));
       }},
+      {"rocksdb.enable_blob_files", [](Server* srv, const std::string &k, const std::string& v)->Status {
```
```suggestion
      {"rocksdb.enable_blob_files", [this](Server* srv, const std::string &k, const std::string& v)->Status {
```

git-hulk: my bad, should capture `this` context here to access the member RocksDB.

git-hulk (dismissed):

src/server.cc
ShooterIT: If we remove this, we also should remove the implementation of `autoResizeBlockAndSST`
we also can do this in the future if  there is no performance loss about `kTwoLevelIndexSearch`.

src/storage.cc
```diff
@@ -69,6 +68,25 @@ void Storage::CloseDB() {
   db_ = nullptr;
 }
 
+void Storage::InitTableOptions(rocksdb::BlockBasedTableOptions *table_options) {
+  table_options->format_version = 5;
+  table_options->index_type = rocksdb::BlockBasedTableOptions::IndexType::kTwoLevelIndexSearch;
```
ShooterIT, line 73: > Although rocksdb will do some optimization if format_version is 5, i still want to set those options explicitly.

As i said, i think we should set `data_block_index_type` to BlockBasedTableOptions::kDataBlockBinaryAndHash and `data_block_hash_table_util_ratio` explicitly. Some users may are interested in optimizations and learn those.

ShooterIT (commented):
LGTM, just minor comments

src/storage.cc
caipengbo: Oh Sorry, I forgot about this option!

ShooterIT:
Please also do some surveys on` memtable_prefix_bloom_size_ratio` and `memtable_whole_key_filtering`.
i think for metadata CF, there are most point lookup, it is a good idea to enable memtable filter, for subkey CF, it may not, sine users may execute `lrange` `hgetall` `zrangeXX` commands that use range lookup.

caipengbo:
> Please also do some surveys on memtable_prefix_bloom_size_ratio and memtable_whole_key_filtering.
i think for metadata CF, there are most point lookup, it is a good idea to enable memtable filter, for subkey CF, it may not, sine users may execute lrange hgetall zrangeXX commands that use range lookup.

Yes, I haven't noticed this optimization yet, I will investigate it later.

src/storage.cc
caipengbo: I don't think we need to expose it explicitly because it's a more advanced configuration that users generally don't need to set. 
And this configuration is at the same level as the other newly added configurations.

src/server.cc
caipengbo: I think it's better to remove this function in the future.

src/storage.cc
```diff
@@ -209,6 +212,9 @@ Status Storage::Open(bool read_only) {
   metadata_opts.table_factory.reset(rocksdb::NewBlockBasedTableFactory(metadata_table_opts));
   metadata_opts.compaction_filter_factory = std::make_shared<MetadataFilterFactory>(this);
   metadata_opts.disable_auto_compactions = config_->RocksDB.disable_auto_compactions;
+  // Enable whole key bloom filter in memtable
+  metadata_opts.memtable_whole_key_filtering = true;
+  metadata_opts.memtable_prefix_bloom_size_ratio = 0.1;
```
ShooterIT, line 217: cool, does it cause write performance loss? is there a read improvement?

caipengbo: I did a simple test and the performance didn't drop, but there was no significant improvement either; 
In theory, this operation can reduce CPU utilization, which is also beneficial.

ShooterIT (approved):

git-hulk (approved):

ShooterIT:
Summary of this PR
- Support blob db (key-value separation) which requires RocksDB version greater than 6.18
- Enable partitioned index/filter for SST and hash index for data block
- Enable whole key filter for memtable
- Remove dynamic resize block and SST
- Only apply compression to SST which level is more than 2

@caipengbo do i miss something?

caipengbo:
> Summary of this PR
> - Support blob db (key-value separation) which requires RocksDB version greater than 6.18
> - Enable partitioned index/filter for SST and hash index for data block
> - Enable whole key filter for memtable
> - Remove dynamic resize block and SST
> - Only apply compression to SST which level is more than 2

@ShooterIT It's a complete summary!

PR #396, opened 2021-11-19
Add write stall stats to info command output

caipengbo:
Kvrocks uses RocksDB as the storage engine. When RocksDB can't flush and compact data promptly, it uses a feature called "[ Write Stalls](https://github.com/facebook/rocksdb/wiki/Write-Stalls)" to try and slow the amount of data coming into the engine. 

This is information user should know. Currently, Kvrocks uses RocksDB's [EventListener](https://github.com/facebook/rocksdb/wiki/EventListener) mechanism to capture stall information. The user must grep the log to get this information.

So I add it to the output of the `INFO` command so that users can get information from the `INFO` command directly. It can also be easily applied to their monitor systems.

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -599,6 +599,7 @@ void Server::GetRocksDBInfo(std::string *info) {
   string_stream << "# RocksDB\r\n";
   for (const auto &cf_handle : *storage_->GetCFHandles()) {
     uint64_t estimate_keys, block_cache_usage, block_cache_pinned_usage, index_and_filter_cache_usage;
+    std::map<std::string, std::string> cf_stats_map;
     db->GetIntProperty(cf_handle, "rocksdb.estimate-num-keys", &estimate_keys);
     string_stream << "estimate_keys[" << cf_handle->GetName() << "]:" << estimate_keys << "\r\n";
     db->GetIntProperty(cf_handle, "rocksdb.block-cache-usage", &block_cache_usage);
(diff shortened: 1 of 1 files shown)


git-hulk (dismissed):
LGTM, so speedy.

src/server.cc
```diff
@@ -608,6 +609,19 @@ void Server::GetRocksDBInfo(std::string *info) {
...
+    string_stream << "level0_file_limit_stop:[" << cf_handle->GetName() << "]:"
+                  << cf_stats_map["io_stalls.level0_numfiles"] << "\r\n";
+    string_stream << "pending_compaction_bytes_slowdown:[" << cf_handle->GetName() << "]:"
+                  << cf_stats_map["io_stalls.slowdown_for_pending_compaction_bytes"] << "\r\n";
+    string_stream << "pending_compaction_bytes_stop:[" << cf_handle->GetName() << "]:"
+                  << cf_stats_map["io_stalls.stop_for_pending_compaction_bytes"] << "\r\n";
+    string_stream << "memtable_count_limit_slowdown:[" << cf_handle->GetName() << "]:"
+                  << cf_stats_map["io_stalls.memtable_slowdown"] << "\r\n";
+    string_stream << "memtable_count_limit_slowdown:[" << cf_handle->GetName() << "]:"
+                  << cf_stats_map["io_stalls.memtable_compaction"] << "\r\n";
```
```suggestion
    string_stream << "memtable_count_limit_stop:[" << cf_handle->GetName() << "]:"
                  << cf_stats_map["io_stalls.memtable_compaction"] << "\r\n";
```

ShooterIT (commented):
should we also show "io_stalls.total_slowdown" and "io_stalls.total_stop"

src/server.cc
caipengbo: I don't think it's necessary to add `io_stalls.total_slowdown` and `io_stalls.total_stop`. Since this value is equal to the sum of the listed three, the user can simply add them through the monitoring system.

If we all add it, we'll have too much information.

caipengbo:
> should we also show "io_stalls.total_slowdown" and "io_stalls.total_stop"

@ShooterIT I don't think it's necessary to add `io_stalls.total_slowdown` and `io_stalls.total_stop`. Since this value is equal to the sum of the listed three, the user can simply add them through the monitoring system.

If we all add it, we'll have too much information.

ShooterIT:
fine, both are ok to me

ShooterIT (approved):

git-hulk (approved):

PR #404, opened 2021-11-25
Make kvrocks2redis work

caipengbo:
I debugged kvRocks2Redis to make it work and added some content to make it easier to use.

### DIFF
--- a/tools/kvrocks2redis/config.cc
+++ b/tools/kvrocks2redis/config.cc
@@ -10,9 +10,9 @@
 #include <sstream>
 #include <vector>
 
-#include "../../src/util.h"
-#include "../../src/status.h"
-#include "../../src/config.h"
+#include "src/util.h"
+#include "src/config.h"
+#include "status.h"
 
 namespace Kvrocks2redis {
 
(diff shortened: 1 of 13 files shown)


tools/kvrocks2redis/config.h
```diff
@@ -19,17 +19,19 @@ struct Config {
...
   std::string aof_file_name = "appendonly.aof";
   std::string next_offset_file_name = "last_next_offset.txt";
-  std::string next_seq_file_path = dir + "/last_next_seq.txt";
+  std::string next_seq_file_path = output_dir + "/last_next_seq.txt";
 
   std::string kvrocks_auth;
   std::string kvrocks_host;
   int kvrocks_port = 0;
   std::map<std::string, redis_server> tokens;
+  bool cluster_enable = true;
```
git-hulk: The default value should be false?

tools/kvrocks2redis/config.cc
```diff
@@ -41,11 +41,25 @@ Status Config::parseConfigFromString(std::string input) {
       return Status(Status::NotOK, "argument must be 'yes' or 'no'");
     }
     daemonize = (i == 1);
-  } else if (size == 2 && args[0] == "dir") {
-    dir = args[1];
-    db_dir = dir + "/db";
-    pidfile = dir + "/kvrocks2redis.pid";
-    next_seq_file_path = dir + "/last_next_seq.txt";
+  } else if (size == 2 && args[0] == "data-dir") {
```
git-hulk, line 44: this would have a breaking change for the old version?

tools/kvrocks2redis/config.h
caipengbo: yeah

tools/kvrocks2redis/config.cc
caipengbo: I think it's a tool that won't be run by users for a long time, so making these changes doesn't affect users much, right?

git-hulk: Okay, it's fine to me.

git-hulk (dismissed):
LGTM, except the `cluster_enable` default value should be false.

tools/kvrocks2redis/config.h
ShooterIT: `cluster_enable` should be false by default.

ShooterIT (changes requested):

git-hulk (approved):

ShooterIT (approved):
LGTM, don't read

Issue #283, opened 2021-05-31
[NEW] TTL of member


hellower:
```
redis> SADD mybasket goods-id
redis> expire mybasket goods-id 10
redis>	1
redis>	expire mybasket goods-id1 goods-id2 goods-id3 10
redis>	3
```

```
redis>	sadd myset mem10
redis>	1
redis>	expire myset mem10 5
redis>	1
redis>	ttl myset mem10
redis>	3
redis>	ttl myset mem10
redis>	-2
redis>	smembers myset
redis>	(empty list or set)
```

This is a feature of redis-enterprise only.   (supported data type: SET, ZSET, HASH)

git-hulk:
This feature seems really useful in many scenes and supports it shouldn't be hard except having the breaking change. Maybe we can support it in V3, how do you think? @bitleak/kvrocks-core-team 

BTW:  @hellower welcome to join [the slack channel](https://join.slack.com/t/kvrockscommunity/shared_invite/zt-p5928e3r-OUAK8SUgC8GOceGM6dAz6w) if wants to exchange any thoughts about the kvrocks.

ShooterIT:
OK. Let's add this feature request into 3.0 project todo.

PR #408, opened 2021-12-01
Add ExHash type, support field expire

caipengbo:
## Background
I opened a Discussion  #399 in which we discussed support TTL for member and I proposed a draft design. Based on this design, I extended the `Hash` type to support field expiration.

##  Design
1. Add a new type, called `ExHash`
2. Based on the coding design of `Hash` Type, we encode expire time in the value of subkey:

metadata:
```
        +----------+------------+-----------+-----------+
key =>  |  flags   |  expire    |  version  |  size     |
        | (1byte)  | (4byte)    |  (8byte)  | (8byte)   |
        +----------+------------+-----------+-----------+
```
subkey:
```
                     +---------------+
key|version|field => |     value     |
                     |     (Nbyte)   |
                     +---------------+
```
Encode expire in subkey's value:
```
                     +----------+---------------+
key|version|field => |  expire  |   value       |
                     |  (4byte) |  (Nbyte)      |
                     +----------+---------------+
```
3. In addition to commands that have the same semantics as `Hash`, `ExHash` has the following **special** commands:

| Command         | Syntax                                   | Description  |
| ----------------- | --------------------------------------------- | ------------ |
| EXHSETEX        | EXHSETEX key field value seconds | Adds a field with a second timeout to a specified ExHash. If the key does not exist, a key for the ExHash is created. If the field already existed, this command overwrites the value of the field.  <br>Returns OK on success. When the seconds parameter is invalid, the command returns an error.|
| EXHEXPIRE        | EXHEXPIRE key field seconds   | Specifies the timeout of a field in a specified ExHash. If the field timeout already existed, this command overwrites the timeout of the field. <br> Returns 1 on success or 0 if key or field does not exist. When the seconds parameter is invalid, the command returns an error.|
| EXHEXPIREAT  | EXHEXPIREAT  key field timestamp   | Like EXHEXPIRE, just using unix timestamp. <br> Returns 1 on success or 0 if key or field does not exist. The command returns an error when the timestamp parameter is invalid. <br>**Note:** If timestamp is less than the current timestamp, the field will not set the lifetime.|
| EXHTTL       | EXHTTL key field                     | Get the remaining expiration time of a field in a specified ExHash. Unit: Seconds. |
| EXHPERSIST   | EXHPERSIST key field                 | Remove the expire of a field in a specified ExHash. Returns 1 when the expire is successfully removed. <br> Returns 0 if field has no expiration set or if ExHash or field does not exist.|
| EXHLEN | EXHLEN key [EXACT]| Get the number of fields about a specified ExHash. If there is no EXACT flag, the length returned is inaccurate (it will contain expired fields that have not been reclaimed). If EXACT flag is set, the exact length is obtained, but this is time-consuming because it will check the entire ExHash. <br> Return the number of fields in a ExHash. Returns 0 if key does not exist. |
| EXHGETALL | EXHGETALL key [WITHTTL] | Returns all fields and values in the ExHash table key without the optional WITHTTL parameter. When WITHTTL is added, all fields, values, and lifetime in the ExHash key are returned. <br> Returns the ExHash's fields and their values as a list, along with the ttl (-1 if no expire is set). If the key does not exist, the empty list is returned. |

The other commands are the same as `Hash`. In particular, I didn't implement EXHMSET (synonymous with HMSET) because this command can be used instead of EXHSET (synonymous with HSET) to set multiple fields and values.

## Implement
1. Add a new `RedisType`, called `kRedisExHash`
2. Modify the `Hash` class and add member variables `support_field_expire_ `to determine whether field expiration is supported. When get the value of the field, we need to check expire.
3. In `SubKeyFilter::IsKeyExpired`, judge whether the Value expire. Note that we do not update the size in metadata because this accesses the `Storage` object and causes data race, which will affect performance. **So the size in `ExHash` metadata is inaccurate, and it contains fields that have expired.**

### DIFF
--- a/src/compact_filter.cc
+++ b/src/compact_filter.cc
@@ -79,6 +79,21 @@ bool SubKeyFilter::IsKeyExpired(const InternalKey &ikey, const Slice &value) con
       || ikey.GetVersion() != metadata.version) {
     return true;
   }
+
+  // field expired
+  if (metadata.Type() == kRedisExHash) {
+    Slice input(value);
+    uint32_t field_expire = 0;
+    GetFixed32(&input, &field_expire);
+    int64_t now;
+    rocksdb::Env::Default()->GetCurrentTime(&now);
+    if (field_expire > 0 && field_expire < now) {
+      // Note: We do not update the size in metadata because this accesses the storage and
+      // causes data race, which affects performance.
+      return true;
+    }
+  }
+
   return metadata.Type() == kRedisBitmap && Redis::Bitmap::IsEmptySegment(value);
 }
 
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -1176,7 +1176,7 @@ class CommandHMSet : public Commander {
     Redis::Hash hash_db(svr->storage_, conn->GetNamespace());
     std::vector<FieldValue> field_values;
     for (unsigned int i = 2; i < args_.size(); i += 2) {
-      field_values.emplace_back(FieldValue{args_[i], args_[i + 1]});
+      field_values.emplace_back(FieldValue{args_[i], args_[i + 1], 0});
     }
     rocksdb::Status s = hash_db.MSet(args_[1], field_values, false, &ret);
     if (!s.ok()) {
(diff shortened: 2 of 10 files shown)


caipengbo:
Can deal with part of the #283

ColinChamber:
Cool. I have some suggestions, maybe it would be better to be more similar to the interface of [Tairhash](https://help.aliyun.com/document_detail/145970.html). Many people have already used it, which already been provided to a [client](https://github.com/aliyun/alibabacloud-tairjedis-sdk).

caipengbo:
Thank you very much. I will take your advice seriously. @ColinChamber

tisonkun:
Closed as stale.

Codebase evolves a lot since the last activities on this patch. There're a few of conflicts. @caipengbo you're welcome to propose the changes again based on the nightly unstable branch.

PR #415, opened 2021-12-03
Implement CAS and CAD commands for string type

### LINKED ISSUE #413: [NEW] Support CAS/CAD commands
CAS (Compare And Swap) and CAD (Compare And Delete) are frequently used in distributed systems, but are not currently supported by Kvrocks.

When we need to use this function, we usually need to use MULTI or Lua script implementation, which is quite troublesome.

caipengbo:
## CAS
This command can be used to  atomic change the value of a specified key to a new value if the current value of the key matches a specified value. The value is not changed if the current value of the key does not match the specified value.

Syntax:
```
CAS key old_value new_value [EX seconds | PX milliseconds]
```

Parameters:
| Parameters         | Description                                   |
| ----------------- | --------------------------------------------- |
| key | The key of the string that you want to manage with CAS command |
| old_value | The value that you compare with the current value of the specified key |
| new_value | Changes the value of the specified key to the value of this parameter if the current value of the key matches the specified value |
| EX or PX | Expiration time, it has the same meaning as the SET command  |

Returned values:
- 1: the operation is successful
- -1: the specified key does not exist
- 0: the operation fails
- Otherwise, an error message is returned.

## CAD
This command can be used to atomic delete a specified key if the current value of the key matches a specified value. The key is not deleted if the current value of the key does not match the specified value.

Syntax:
```
CAD key value
```
Parameters:
| Parameters         | Description                                   |
| ----------------- | --------------------------------------------- |
| key | The key of the Redis string that you want to manage with CAD command |
| value | The value that you compare with the current value of the specified key |

Returned values:
- 1: the operation is successful
- -1: the specified key does not exist
- 0: the operation fails
- Otherwise, an error message is returned

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -640,6 +640,68 @@ class CommandDecrBy : public Commander {
   int64_t increment_ = 0;
 };
 
+class CommandCAS : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    bool last_arg;
+    for (size_t i = 4; i < args.size(); i++) {
+      last_arg = (i == args.size()-1);
+      std::string opt = Util::ToLower(args[i]);
+      if (opt == "ex") {
+        if (last_arg) return Status(Status::NotOK, errWrongNumOfArguments);
+        ttl_ = atoi(args_[++i].c_str());
+        if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "px") {
+        if (last_arg) return Status(Status::NotOK, errWrongNumOfArguments);
+        auto ttl_ms = atol(args[++i].c_str());
+        if (ttl_ms <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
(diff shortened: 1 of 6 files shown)


ShooterIT:
lgtm, cool job!

After merging this PR, i think we should support EXAT PXAT options for SET and CAS, since these options are frequently-used and friendly, WDYT? @caipengbo @git-hulk

caipengbo:
> After merging this PR, i think we should support EXAT PXAT options for SET and CAS, since these options are frequently-used and friendly, WDYT? 

Yeah, we should be as compliant with Redis commands as possible.

ShooterIT (approved):

ShooterIT:
@caipengbo Could update our wiki https://github.com/KvrocksLabs/kvrocks/wiki/Support-Commands for these new commands

caipengbo:
I want to update the wiki, but I don't have permission @ShooterIT

caipengbo:
Solved #413

git-hulk (approved):

ShooterIT:
oh, let me update

PR #423, opened 2021-12-11
Auto Rewrite the config after the namespace was updated

### LINKED ISSUE #422: [NEW] Auto persist namespaces after updating the namespace or token
Currently, we didn't auto rewrite namespaces and tokens after users updated them, which may cause configuration loss after restarting the Kvrocks without executing the config rewrite command. But the data is still in Kvrocks. All we need to do was re-add those namespaces again.

git-hulk:
resolved #422 

Currently, we didn't rewrite the config when the namespace was updated,
which may cause the namespace loss after restarting without manually
executing the config rewrite command. So we fix the issue by rewriting the
config here to make users happy.

This PR also fixes some small bugs like `blob_garbage_collection_age_cutoff` default value was out of range and didn't force covert it to double when dividing.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -538,8 +538,10 @@ Status Config::Rewrite() {
   for (const auto &iter : fields_) {
     new_config[iter.first] = iter.second->ToString();
   }
+
+  std::string namespacePrefix = "namespace.";
   for (const auto &iter : tokens) {
-    new_config["namespace." + iter.second] = iter.first;
+    new_config[namespacePrefix + iter.second] = iter.first;
   }
 
   std::ifstream file(path_);
@@ -556,6 +558,10 @@ Status Config::Rewrite() {
       Util::Split2KV(trim_line, " \t", &kv);
       if (kv.size() != 2) {
         lines.emplace_back(raw_line);
+        continue;
+      }
+      if (Util::HasPrefix(kv[0], namespacePrefix)) {
+        // Ignore namespace fields here since we would always rewrite them
         continue;
       }
       auto iter = new_config.find(Util::ToLower(kv[0]));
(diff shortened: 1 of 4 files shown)


caipengbo (dismissed):
LGTM

ShooterIT:
> This PR also fixes some small bugs like blob_garbage_collection_age_cutoff default value was out of range and didn't force covert it to double when dividing.

I think we should separate this fix to another PR, and link the PR #395 which introduce this bug, so users could  trace the change history of new features and their bug fixes.

ShooterIT:
I think rewriting config file just is a temporary solution since  config changes wouldn't be propagated into replicas, uses need to set namespaces on all the instance when instances are added into. Moreover, config file also may be lost.

I thinks we could persist them and propagate namespaces information into replicas.

git-hulk:
> > This PR also fixes some small bugs like blob_garbage_collection_age_cutoff default value was out of range and didn't force covert it to double when dividing.
> 
> I think we should separate this fix to another PR, and link the PR #395 which introduce this bug, so users could trace the change history of new features and their bug fixes.

Yes, you're right, my bad for being lazy, would split later

git-hulk:
> I think rewriting config file just is a temporary solution since config changes wouldn't be propagated into replicas, uses need to set namespaces on all the instance when instances are added into. Moreover, config file also may be lost.
> 
> I thinks we could persist them and propagate namespaces information into replicas.

yup, propagating to replicas would be better but it would have a bit of behavior change if users use the different passwords now. Maybe we can do that in 3.0.

git-hulk:
Fix the `blob_garbage_collection_age_cutoff` was always zero issue has separated into another PR to make the context more clear. But the invalid default value of `blob_garbage_collection_age_cutoff` was still in this PR since it would rewrite and reload the config, which would fail the unittest.

ShooterIT (approved):

caipengbo (approved):

PR #427, opened 2021-12-16
fix:parseComplexKV with unknown metadatatype:7

fishery86:
kRedisSortedint metadata type 7 out of range

### DIFF
--- a/tools/kvrocks2redis/parser.cc
+++ b/tools/kvrocks2redis/parser.cc
@@ -54,7 +54,7 @@ Status Parser::parseSimpleKV(const Slice &ns_key, const Slice &value, int expire
 
 Status Parser::parseComplexKV(const Slice &ns_key, const Metadata &metadata) {
   RedisType type = metadata.Type();
-  if (type < kRedisHash || type > kRedisBitmap) {
+  if (type < kRedisHash || type > kRedisSortedint) {
     return Status(Status::NotOK, "unknown metadata type: " + std::to_string(type));
   }
 


tools/kvrocks2redis/parser.cc
```diff
@@ -54,7 +54,7 @@ Status Parser::parseSimpleKV(const Slice &ns_key, const Slice &value, int expire
 
 Status Parser::parseComplexKV(const Slice &ns_key, const Metadata &metadata) {
   RedisType type = metadata.Type();
-  if (type < kRedisHash || type > kRedisBitmap) {
+  if (type < kRedisHash || type > kRedisSortedint) {
```
caipengbo, line 57: Since `Sortedint` is a new addition to Kvrocks and is not supported by Redis, we don't need to deal with it here.

git-hulk: @caipengbo Seems we have converted the sorted int to the Redis set type, so it should make sense here?

fishery86: a configuration should be added to support such compatiable convert?

ShooterIT: If it is possible, we should support to convert the sorted int to the Redis set type, we shouldn't let user to handle this problem.

ShooterIT: > If it is possible

this also means user could read sorted int by redis commands after migrating from kvrocks to redis

git-hulk: yeah, @karelrooted should have converted to the Redis set if my memory serves. Can u help to confirm? @caipengbo

caipengbo: I didn't find it converted to Redis set.

caipengbo:
Currently, sorted int is not converted to Redis set type.

@fishery86 Are you interested in supporting this compatibility?  If not, I'll support it in the future.

fishery86:
> Currently, sorted int is not converted to Redis set type.
> 
> @fishery86 Are you interested in supporting this compatibility? If not, I'll support it in the future.

ok,I will do the job

ShooterIT:
resolved in #431

PR #429, opened 2021-12-16
Fix config rewrite overwrite the first rename-command with the last one

### LINKED ISSUE #428: [BUG] It is wrong to rewrite config `rename-command` and lead kvrocks to fail to restart
### Describe
If your kvrocks config file has multiple `rename-command`, kvrocks would generate duplicate configuration items after executing `config rewrite`, whatever automatically or manually, and leads kvrocks to fail to restart.

It is introduced in #272

one example
```
rename-command debug                               290033_dont_run_debug_rudely
rename-command bgsave                              290033_dont_run_bgsave_rudely
rename-command shutdown                            290033_dont_run_shutdown_rudely
rename-command keys                                290033_dont_run_keys_rudely
rename-command debug                               290033_dont_run_debug_rudely
```
there are two debug rename-commands, so there will be the following error logs.
```
* Failed to load config, err: at line: #L19, err: No such command in rename-command
```
### How to fix
- you can find the duplicate configuration items, and remove duplicate items, only keep one remame-command for one command.
- of course, you can cherry-pick this commit https://github.com/KvrocksLabs/kvrocks/commit/6c8534a17f056f06faa09bfeada3f92df686975a (#429 ) 
- use 2.0.6 version (coming soon)

git-hulk:
For example, we have below rename-commands in the config file:
```
rename-command KEYS KEYS_NEW
rename-command GET  GET_NEW
rename-command SET  SET_NEW
```
then `rename-command KEYS KEYS_NEW` would be overwriten to
`rename-command SET  SET_NEW` since we didn't skip the `rename-command`
correctly in config rewrite.

Closes #428

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -536,6 +536,11 @@ Status Config::Rewrite() {
   std::vector<std::string> lines;
   std::map<std::string, std::string> new_config;
   for (const auto &iter : fields_) {
+    if (iter.first == "rename-command") {
+      // We should NOT overwrite the rename command since it cannot be rewritten in-flight,
+      // so skip it here to avoid rewriting it as new item.
+      continue;
+    }
     new_config[iter.first] = iter.second->ToString();
   }
 
--- a/tests/config_test.cc
+++ b/tests/config_test.cc
@@ -2,6 +2,8 @@
 #include "server.h"
 #include <map>
 #include <vector>
+#include <fstream>
+#include <iostream>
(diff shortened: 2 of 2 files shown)


git-hulk:
tcl tests config `overvides` didn't support multi items with the same key name like `rename-command`, so let's test it with the googletest.

ShooterIT (approved):

PR #430, opened 2021-12-17
Support slot-based data migration

### LINKED ISSUE #412: [NEW] Support slot-based data migration
# 1 Background
Data online migration is an essential feature for database servers if they are deployed in the cluster. Since kvrocks already supports redis cluster mode #219, it is more necessary to support data online migration. Different from the key-based data migration method of the Redis community, we proposed a slot-based data migration method for kvrocks. 

# 2 Implementation
## 2.1 Data encoding format
To support slot-based migration, we need to encode slotid onto every key to improve the efficiency of iterating data.  The string key and hash key are adopted to explain the slotid encoding for simple key and complex key respectively as follows.

**String key**:
```
    +--------+----+---------+----------+
    | ns_len | ns | slot_id | user_key |
    +--------+----+---------+----------+
```
**Hash key**:
```
    hash metakey
    +--------+----+---------+----------+
    | ns_len | ns | slot_id | user_key |
    +--------+----+---------+----------+
    hash subkey
    +--------+----+---------+--------------+----------+---------+-------+
    | ns_len | ns | slot_id | user_key_len | user_key | Version | field |
    +--------+----+---------+--------------+----------+---------+-------+
```
As shown above, slotid is encoded onto the prefix of every key. Keys of the same slot will have the same prefix, they will be stored in an adjacent location in rocksdb which will improve data iterating efficiency.  Encoding slotid onto keys has been supported at #291 for kvrocks. 

## 2.2 Slot-based migration brief design
Slot-based migration process mainly includes the following stages.

1. Start migrating
2. Migrating existing data
3. Migrating incremental data
4. End migrating

The main process of slot-based migration can be described in the following diagram.
kvrocks data migration
Figure 1. Slot-based migration process diagram

### 2.3.1 Details of migrating process
As shown in the process diagram (Figure 1), the data migration will be triggered by sending a request to the source server. The source server will create a migration task after it got the data migration request. The main processes of slot-based migration are processed by this migration task. The details can be described as the following stages.
#### 1) Start migrating stage
At this stage, the source server will notify the destination server to prepare to import data. If the destination server is ready, the source server will go to the next stage to migrate data. Otherwise, the source server will stop the migration task.
#### 2) Migrating existing data stage
At this stage, the existing data will be migrated. Existing data of kvrocks is described by rocksdb snapshot at the migration beginning moment. Then, the source server will iterate all data of the snapshot, and construct data into Redis commands to send to the destination server. Constructed Redis commands will be sent by pipeline to improve efficiency.
#### 3) Migrating incremental data stage
While migrating existing data, the migrating slot can keep writing. In other words, new data will be written during migrating existing data. These new data must be migrated to the destination server too.  Before migrating incremental data, the migrating slot will be forbidden to write to maintain consistency. New data of the target slot cannot be written to the source server again. 
The amount of the incremental data may be very large, because it may take a long time to migrate the existing data. It will cause the slot to forbid writing for a long time.  To reduce the forbidden writing time duration, the incremental data migration will be processed in the two-step. 
- **First step**: Slot will not be forbidden writing while migrating incremental data. This step will be repeated until the amount of new data is less than a threshold, or the repetition times reach a threshold.
- **Second step**: Slot will be forbidden from writing before migrating the rest new data.
The incremental data will be gotten via iterating WAL of rocksdb.
#### 4) End migrating stage
The previous stages may succeed or fail. In this stage, the source server will notify the destination server that the migration task succeeded or failed. If the migration is successful, both source server and the destination server will change the cluster topology maintained by themslves, and source server will clear data belonging to migreted slot. If the migration fails, only destination server will clear imported data of the target slot.

### 2.3.2 Support commands
- CLUSTERX MIGRATE `$slot` `$dst_nodeid`
     `$dst_nodeid` is the node id of destination server in the cluster. See #219 for more details.
- CLUSTER IMPORT `$slot` `$state`
It is an internal command which will be sent by the source server to notify the destination server to prepare for data importing. **This command cannot be used directly by clients**.

# 3 Advantages and Disadvantages
## 3.1 Advantages
- Efficiency
  Compare with the key-based data migration method, the slot-based method supports data transmission with the pipeline, it is more efficient. 
- More convenient failure rollback
  Data in the source server can be deleted only when all data is migrated successfully. If any failure happens during data migrating, the migration task will be stopped, data won't be deleted.
- Consistency
  Writing to the migrating slot is forbidden while transmitting the last piece of incremental data. It can guarantee data consistency.
- asynchronous migration 
  Data migrating is processed in an independent thread, and does not affect the main threads to process requests.

## 3.2 Disadvantages
- Currently, it only supports to migrate slot one by one.

# 4 Extra work
- Support concurrent data migration.

ChrisZMF:
resolves #412 

## Support Commands
- CLUSTERX MIGRATE `$slot` `$dst_nodeid` 
`$dst_nodeid` is the node id of destination server in the cluster. See [NEW] Support redis cluster mode #219 for more details.
- CLUSTER IMPORT `$slot` `$state`
It is an internal command which will be sent by the source server to notify the destination server to prepare for data importing. This command cannot be used by clients.
- Migrating status and importing status are merged into the output of `CLUSTER INFO`.
Example:
Source node info:
```
127.0.0.1:30001> cluster info
cluster_state:ok
cluster_slots_assigned:16384
......
migrating_slot: 0
destination_node: kvrockskvrockskvrockskvrockskvrocksnode3
migrating_state: success
  ```
Destination node info:
```
127.0.0.1:30003> cluster info
cluster_state:ok
cluster_slots_assigned:16384
......
importing_slot: 0
import_state: success
```

After migration slot, you should use CLUSTERX SETSLOT #463  to change cluster slot distribution.

### CODE REVIEW

**Comment by ShooterIT** [comment]
@ChrisZMF  Please resolve conflicts （related to kvrocks2redis）

File: src/batch_parser.h [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+#include "redis_metadata.h"
+
+
+// An extractor to extract update from raw writebatch
+class WriteBatchExtractor : public rocksdb::WriteBatch::Handler {
+ public:
+  explicit WriteBatchExtractor(bool is_slotid_encoded, int16_t slot = -1, bool to_redis = false)
+  : is_slotid_encoded_(is_slotid_encoded), slot_(slot), to_redis_(to_redis) {}
+  void LogData(const rocksdb::Slice &blob) override;
+  rocksdb::Status PutCF(uint32_t column_family_id, const Slice &key,
```
ShooterIT: How about `Put` instead of `PutCF`?
Maybe you get ideas from old code, but i think `PutCF` means put a CF into rocksdb, and the first parameter is the cf id, so CF in function name is not necessary. And here we want to parse the `Put` and `Delete` operations (that is consistent with rocksdb) into redis commands

Also `DeleteCF` -> `Delete`and `DeleteRangeCF` -> `DeleteRange`?

src/batch_parser.h
```diff
@@ -0,0 +1,32 @@
...
+#include <map>
+
+#include "redis_db.h"
+#include "status.h"
+#include "storage.h"
+#include "redis_metadata.h"
+
+
+// An extractor to extract update from raw writebatch
+class WriteBatchExtractor : public rocksdb::WriteBatch::Handler {
```
ShooterIT: `WriteBatchExtractor` is not consistent with  the file name `batch_parser`, i think we should keep consistent.
How about just `BatchExtractor` and file name also is `batch_extractor`?

src/batch_parser.h
```diff
@@ -0,0 +1,32 @@
...
+  void LogData(const rocksdb::Slice &blob) override;
+  rocksdb::Status PutCF(uint32_t column_family_id, const Slice &key,
+                        const Slice &value) override;
+
+  rocksdb::Status DeleteCF(uint32_t column_family_id, const Slice &key) override;
+  rocksdb::Status DeleteRangeCF(uint32_t column_family_id,
+                                const Slice& begin_key, const Slice& end_key) override;
+  std::map<std::string, std::vector<std::string>> *GetAofStrings() { return &aof_strings_; }
+ private:
+  std::map<std::string, std::vector<std::string>> aof_strings_;
```
ShooterIT: I think is not a good idea to use `aof` word, even we usually say aof comamnd, but actually AOF is the short of `Append Only File`.
How about `resp_commands`?

src/batch_parser.cc
```diff
@@ -0,0 +1,207 @@
...
+  std::vector<std::string> command_args;
+  if (column_family_id == kColumnFamilyIDMetadata) {
+    ExtractNamespaceKey(key, &ns, &user_key, is_slotid_encoded_);
+    if (slot_ >= 0) {
+      if (static_cast<uint16_t>(slot_) != GetSlotNumFromKey(user_key)) return rocksdb::Status::OK();
+    }
+    Metadata metadata(kRedisNone);
+    metadata.Decode(value.ToString());
+    if (metadata.Type() == kRedisString) {
+      command_args = {"SET", user_key, value.ToString().substr(5, value.size() - 5)};
```
ShooterIT: prefer to use `set Key Value [EXAT unix-time-seconds]` when having expire time. But now we support this flag.

src/batch_parser.cc
```diff
@@ -0,0 +1,207 @@
...
+      aof_strings_[ns].emplace_back(Util::Command2RESP(command_args));
+      if (metadata.expire > 0) {
+        command_args = {"EXPIREAT", user_key, std::to_string(metadata.expire)};
+        aof_strings_[ns].emplace_back(Util::Command2RESP(command_args));
+      }
+    } else if (metadata.expire > 0) {
```
ShooterIT: should use `else`? whatever key type, you also need to try to set expire time

src/cluster.cc
```diff
@@ -53,6 +54,26 @@ Status Cluster::SetNodeId(std::string node_id) {
...
 
+Status Cluster::SetSlot(int slot, std::string node_id) {
+  if (slot < 0 || slot >= HASH_SLOTS_SIZE) {
+    return Status(Status::NotOK, "Slot is out of range");
+  }
+  if (node_id.empty() || nodes_.find(node_id) == nodes_.end()) {
+    return Status(Status::NotOK, "Invalid node id");
+  }
+
+  std::lock_guard<std::mutex> guard(cluster_mutex_);
```
ShooterIT: > // We access cluster without lock, acutally we guarantte data-safe by work theads
> // ReadWriteLockGuard, CLUSTER command doesn't have 'execlusive' attribute, i.e.
> // CLUSTER command can be executed concurrently, but some subcommand may change
> // cluster data, so these commands should be executed exclusively, and ReadWriteLock
> // also can guarantte accessing data is safe.

please see comments above

src/cluster.cc
```diff
@@ -412,10 +562,13 @@ Status Cluster::CanExecByMySelf(const Redis::CommandAttributes *attributes,
   }
   if (slot == -1) return Status::OK();
 
+  std::lock_guard<std::mutex> guard(cluster_mutex_);
```
ShooterIT: so i think it is not necessary

src/cluster.cc
```diff
@@ -53,6 +54,26 @@ Status Cluster::SetNodeId(std::string node_id) {
...
+
+  std::lock_guard<std::mutex> guard(cluster_mutex_);
+  if (slots_nodes_[slot]) {
+    slots_nodes_[slot]->slots_.set(slot, 0);
+  }
+  nodes_[node_id]->slots_.set(slot, 1);
+  slots_nodes_[slot] = nodes_[node_id];
+
+  // Mark topology changed
+  version_++;
```
ShooterIT: maybe it is not a good idea to change cluster topology and version_ directly, since we use controler to manage everything,
An idea: we set the slot to `imported` state on the dst node which can serve requests on the imported slot in this status, and we set the slot to `migrated` state on the src node which will `MOVE` the requests on the migrated slot,

src/redis_cmd.cc
```diff
@@ -3435,13 +3457,26 @@ class CommandSlaveOf : public Commander {
       if (s.IsOK()) {
         *output = Redis::SimpleString("OK");
         LOG(WARNING) << "MASTER MODE enabled (user request from '" << conn->GetAddr() << "')";
+        if (svr->GetConfig()->slot_id_encoded) {
+          svr->slot_migrate_->StartMigrateTask();
+          LOG(INFO) << "Change server role to master, restart migration task";
+        }
       }
     } else {
+      std::unique_lock<std::mutex> ul(svr->migrate_mutex, std::defer_lock);
```
ShooterIT: why not stop migrating?

src/redis_connection.cc
```diff
@@ -363,14 +397,24 @@ void Connection::ExecuteCommands(const std::vector<Redis::CommandTokens> &to_pro
...
+      s = svr_->cluster_->CanExecByMySelf(attributes, cmd_tokens, this);
       if (!s.IsOK()) {
         if (IsFlagEnabled(Connection::kMultiExec)) multi_error_ = true;
         Reply(Redis::Error(s.Msg()));
         continue;;
       }
     }
 
+    // Check slot keys of forbidden slot for migration in cluster mode
+    if (svr_->GetConfig()->slot_id_encoded) {
```
ShooterIT: maybe `cluster_enabled` is better

src/redis_db.cc
```diff
@@ -482,6 +490,68 @@ rocksdb::Status Database::FindKeyRangeWithPrefix(const std::string &prefix,
...
+
+  std::string prefix, prefix_end;
+  ComposeSlotKeyPrefix(ns, slot, &prefix);
+  ComposeSlotKeyPrefix(ns, slot + 1, &prefix_end);
+  std::string begin_key, end_key;
+  auto s = FindKeyRangeWithPrefix(prefix, prefix_end, &begin_key, &end_key, storage_->GetCFHandle("metadata"));
+  if (s.IsNotFound()) {
+    return rocksdb::Status::OK();
+  }
+  s = storage_->DeleteRange(begin_key, end_key);
```
ShooterIT: call `DeleteRange(slot, slot+1)` directly

src/redis_metadata.cc
```diff
@@ -10,7 +10,6 @@
 
 // 52 bit for microseconds and 11 bit for counter
 const int VersionCounterBits = 11;
```
ShooterIT: don't delete empty line here

src/redis_request.h
```diff
@@ -5,6 +5,7 @@
 #include <event2/buffer.h>
 
 #include "status.h"
+#include "redis_cmd.h"
```
ShooterIT: why do this change. i.e add here and delete it in `redis_request.cc`

src/server.h
```diff
@@ -109,6 +113,9 @@ class Server {
   ReplState GetReplicationState();
 
   void PrepareRestoreDB();
+  void ReclaimOldDBPtr();
```
ShooterIT: ?

src/slot_import.h
```diff
@@ -0,0 +1,38 @@
...
+#include <string>
+#include <vector>
+#include <mutex>
+#include <glog/logging.h>
+
+#include "redis_db.h"
+#include "config.h"
+#include "server.h"
+
+// To adapt PegaDB1.0, cannot modify the order of import statuses here
```
ShooterIT: delete it

ShooterIT (commented):
Amazing work, let's review ASAP @git-hulk

src/batch_parser.h
ChrisZMF: `PutCF` is a virtual member function of class `WriteBatch::Handler`, it is overrided here. It is called by function `WriteBatch::Iterate`, we have to keep the same function name as it is declared. Also `DeleteCF` and `DeleteRangeCF`.

src/batch_parser.h
ChrisZMF: It is a good idea to replace it by `resp_commands`, also to rename the file as 'batch_extractor'.

src/batch_parser.cc
ChrisZMF: I think it should be `else if` here. No matter what the type of key is, the expiration time of the key needs to be set only when `meatdata.expire > 0` .

src/batch_parser.h
ShooterIT: ok, fine, i miss that

src/batch_parser.cc
ShooterIT: i mean use `if`, don't need `else`

src/cluster.cc
```diff
@@ -140,6 +161,134 @@ void Cluster::SetMasterSlaveRepl() {
   }
 }
 
+Status Cluster::MigrateSlot(std::string dst_node, int slot) {
+  if (nodes_.find(dst_node) == nodes_.end()) {
+    return Status(Status::NotOK, "Destination node id is wrong");
```
git-hulk:
```suggestion
    return Status(Status::NotOK, "Cannot find the destination node id");
```

src/cluster.cc
```diff
@@ -140,6 +161,134 @@ void Cluster::SetMasterSlaveRepl() {
   }
 }
 
+Status Cluster::MigrateSlot(std::string dst_node, int slot) {
```
git-hulk: Can change the `dst_node` to `const std::string &dst_node_id`

src/cluster.cc
```diff
@@ -140,6 +161,134 @@ void Cluster::SetMasterSlaveRepl() {
...
+  if (myself_->role_ != kClusterMaster || svr_->IsSlave()) {
+    return Status(Status::NotOK, "Slave can't import slot");
+  }
+  if (slot < 0 || slot >= HASH_SLOTS_SIZE) {
+    return Status(Status::NotOK, "Slot is out of range");
+  }
+
+  switch (state) {
+    case kImportStart:
+      if (svr_->slot_import_->Start(conn->GetFD(), slot) == false) {
```
git-hulk:
```suggestion
      if (!svr_->slot_import_->Start(conn->GetFD(), slot)) {
```

src/cluster.cc
```diff
@@ -140,6 +161,134 @@ void Cluster::SetMasterSlaveRepl() {
...
+        return Status(Status::NotOK, "Can't start importing slot " + std::to_string(slot));
+      } else {
```
git-hulk: I think we can remove this `else` keyword

src/cluster.cc
```diff
@@ -140,6 +161,134 @@ void Cluster::SetMasterSlaveRepl() {
...
+    case kImportSuccess:
+      if (svr_->slot_import_->Success(slot) == false) {
+        LOG(ERROR) << "[import] Failed to set slot importing success, maybe slot is wrong"
+                  << ", recieved slot: " << slot
+                  << ", current slot: " << svr_->slot_import_->GetSlot();
+        return Status(Status::NotOK, "Failed to set slot " + std::to_string(slot) + " importing success");
+      }
+      LOG(INFO) << "[import] Succeed to import slot " << slot;
+      break;
+    case kImportFail:
```
git-hulk: kImportFail should be `kImportFailed`

src/cluster.cc
```diff
@@ -140,6 +161,134 @@ void Cluster::SetMasterSlaveRepl() {
...
+        myself_->importing_slot_ = slot;
+        // Set link error callback
+        conn->close_cb_ = std::bind(&SlotImport::StopForLinkError, svr_->slot_import_, conn->GetFD());
+        // Stop forbiding writing slot to accept write commands
+        svr_->slot_migrate_->ReleaseForbiddenSlot();
+        LOG(INFO) << "[import] Start importing slot " << slot;
+      }
+      break;
+    case kImportSuccess:
+      if (svr_->slot_import_->Success(slot) == false) {
```
git-hulk:
```suggestion
      if (!svr_->slot_import_->Success(slot)) {
```

src/cluster.cc
```diff
@@ -140,6 +161,134 @@ void Cluster::SetMasterSlaveRepl() {
...
+      break;
+    default:
+      return Status(Status::NotOK, "Invalid import state");
+      break;
+  }
+  return Status::OK();
+}
+
+Status Cluster::GetMigrateInfo(int slot, std::vector<std::string> *info) {
+  if (myself_->role_ != kClusterMaster || svr_->IsSlave()) {
```
git-hulk: I saw the role and slot check in many places, we can become it to the member function.

src/cluster.cc
```diff
@@ -412,10 +562,13 @@ Status Cluster::CanExecByMySelf(const Redis::CommandAttributes *attributes,
   }
   if (slot == -1) return Status::OK();
 
+  std::lock_guard<std::mutex> guard(cluster_mutex_);
   if (slots_nodes_[slot] == nullptr) {
     return Status(Status::ClusterDown, "CLUSTERDOWN Hash slot not served");
   } else if (myself_ && myself_ == slots_nodes_[slot]) {
     return Status::OK();  // I serve this slot
+  } else if (myself_ && myself_->importing_slot_ == slot && conn->IsImporting()) {
+    return Status::OK();  // I server importing connection
```
git-hulk:
```suggestion
    return Status::OK();  // I'm serving the importing connection
```

src/config.cc
```diff
@@ -455,6 +474,9 @@ Status Config::finish() {
   if (requirepass.empty() && !tokens.empty()) {
     return Status(Status::NotOK, "requirepass empty wasn't allowed while the namespace exists");
   }
+  if ((cluster_enabled) && !tokens.empty()) {
+    return Status(Status::NotOK, "enabled cluster mode wasn't allowed while the namespace exists");
```
git-hulk, line 483:
```suggestion
    return Status(Status::NotOK, "Enabled cluster mode wasn't allowed while the namespace exists");
```

src/config.cc
```diff
@@ -637,6 +659,9 @@ Status Config::AddNamespace(const std::string &ns, const std::string &token) {
   if (requirepass.empty()) {
     return Status(Status::NotOK, "forbidden to add namespace when requirepass was empty");
   }
+  if (slot_id_encoded) {
+    return Status(Status::NotOK, "forbidden to add namespace when cluster mode was enabled");
```
git-hulk, line 668:
```suggestion
    return Status(Status::NotOK, "Forbidden to add namespace when cluster mode was enabled");
```

src/redis_cmd.cc
```diff
@@ -4354,6 +4476,9 @@ class CommandClusterX : public Commander {
   std::string nodes_str_;
   uint64_t set_version_ = 0;
   bool force_ = false;
+  std::string dst_node_;
```
git-hulk: dst_node_ => dst_node_id_

src/redis_db.cc
```diff
@@ -482,6 +490,68 @@ rocksdb::Status Database::FindKeyRangeWithPrefix(const std::string &prefix,
...
+                                          std::vector<std::string> *keys,
+                                          int count) {
+  const rocksdb::Snapshot *snapshot;
+  snapshot = storage_->GetDB()->GetSnapshot();
+  rocksdb::ReadOptions read_options;
+  read_options.snapshot = snapshot;
+  read_options.fill_cache = false;
+  auto iter = db_->NewIterator(read_options, metadata_cf_handle_);
+  bool end = false;
+  for (int i = 0; i < 16384; i++) {
```
git-hulk: 16384 should use const

ShooterIT:
Hi @ChrisZMF I have merge `CLUSTER SETSLOT` PR, please resolve the conflicts and implement what we have reached an agreement!

src/config.cc
```diff
@@ -642,6 +664,9 @@ Status Config::AddNamespace(const std::string &ns, const std::string &token) {
   if (requirepass.empty()) {
     return Status(Status::NotOK, "forbidden to add namespace when requirepass was empty");
   }
+  if (slot_id_encoded) {
```
ShooterIT: cluster_enabled

src/redis_cmd.cc
```diff
@@ -3450,13 +3464,21 @@ class CommandSlaveOf : public Commander {
       if (s.IsOK()) {
         *output = Redis::SimpleString("OK");
         LOG(WARNING) << "MASTER MODE enabled (user request from '" << conn->GetAddr() << "')";
+        if (svr->GetConfig()->slot_id_encoded) {
```
ShooterIT: cluster_enabled

src/server.cc
```diff
@@ -976,6 +989,8 @@ void Server::PrepareRestoreDB() {
   task_runner_.Join();
   task_runner_.Purge();
```
ShooterIT, line 991: add comments

src/server.h
```diff
@@ -161,6 +166,7 @@ class Server {
 
   std::atomic<bool> stop_;
   std::atomic<bool> is_loading_;
+  std::atomic<bool> is_blocking_{false};
```
ShooterIT: need to delete

src/util.cc
```diff
@@ -23,6 +23,7 @@
 
 #include "util.h"
 #include "status.h"
+#include "redis_reply.h"
```
ShooterIT: delete?

src/slot_import.cc
```diff
@@ -0,0 +1,151 @@
...
+  std::lock_guard<std::mutex> guard(mutex_);
+  return import_slot_;
+}
+
+int SlotImport::GetStatus() {
+  std::lock_guard<std::mutex> guard(mutex_);
+  return import_status_;
+}
+
+Status SlotImport::GetImportInfo(std::vector<std::string> *info, int slot) {
```
ShooterIT: std::lock_guard<std::mutex> guard(mutex_);
also is needed

src/slot_import.cc
```diff
@@ -0,0 +1,151 @@
+#include "slot_import.h"
+
+SlotImport::SlotImport(Server *svr) : Database(svr->storage_, kDefaultNamespace), svr_(svr) {
+  std::lock_guard<std::mutex> guard(mutex_);
+  // Let db_ and metadata_cf_handle_ be nullptr, then get them in real time while use them.
+  // See comments in SlotMigrate::SlotMigrate for detailed reason.
+  db_ = nullptr;
```
ShooterIT, line 7: is it necessary in SlotImport?

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+  // 4. It will occur segment fault when using metadata_cf_handle_ to create iterator of rocksdb.
+  // [Reason]:
+  // After full sychronization, DB will be reopened, db_ and metadata_cf_handle_ will be released.
+  // Then, if we create rocksdb iterator with metadata_cf_handle_, it will go wrong.
+  // [Solution]:
+  // db_ and metadata_cf_handle_ will be replaced by storage_->GetDB() and storage_->GetCFHandle("metadata")
+  // in all functions used in migration process.
+  // [Note]:
+  // This problem may exist in all functions of Database called in slot migration process.
+  db_ = nullptr;
```
ShooterIT, line 34: is it necessary?

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+          LOG(INFO) << "[migrate] Succeed to send snapshot";
+          state_machine_ = kSlotMigrateWal;
+        } else {
+          LOG(ERROR) << "[migrate] Failed to send snapshot";
+          state_machine_ = kSlotMigrateFailed;
+        }
+        break;
+      }
+      case kSlotMigrateWal: {
+        auto s = SyncWal();
```
ShooterIT, line 161: how about SendWal(), correspond with SendSnapshot

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+        break;
+      }
+      case kSlotMigrateSuccess: {
+        auto s = Success();
+        if (s.IsOK()) {
+          LOG(INFO) << "[migrate] Succeed to handle state 'SUCCESS'";
+          state_machine_ = kSlotMigrateClean;
+          migrate_state_ = kMigrateSuccess;
+        } else {
+          LOG(ERROR) << "[migrate] Failed to handle state 'SUCCESS'";
```
ShooterIT: how about `Importing server failed to ...`

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+}
+
+Status SlotMigrate::Success(void) {
+  if (stop_) {
+    LOG(ERROR) << "[migrate] Stop migrating slot " << migrate_slot_;
+    return Status(Status::NotOK);
+  }
+  // Set destination status SUCCESS
+  if (!SetDstImportStatus(slot_job_->slot_fd_, kImportSuccess)) {
+    LOG(ERROR) << "[migrate] Fail to set destination import status " << kImportSuccess;
```
ShooterIT: kImportSuccess just is a number, we should use text
[migrate] Fail to set destination server SUCCESS import status

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+    return Status(Status::NotOK);
+  }
+  std::string dst_ip_port = dst_ip_ + ":" + std::to_string(dst_port_);
+  Status st = svr_->cluster_->SetSlotMigrated(migrate_slot_, dst_ip_port);
+  if (!st.IsOK()) {
+    LOG(ERROR) << "[migrate] Failed to set slot, Err:" << st.Msg();
```
ShooterIT, line 325: "Failed to set slot migrated state,..."

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+    return Status(Status::NotOK);
+  }
+  return Status::OK();
+}
+
+Status SlotMigrate::Fail(void) {
+  // Set destination status
+  if (!SetDstImportStatus(slot_job_->slot_fd_, kImportFailed)) {
+    LOG(INFO) << "[migrate] Fail to set importing status " << kImportFailed;
```
ShooterIT: "Fail to set destination server FAIL import state"

src/slot_migrate.h
```diff
@@ -0,0 +1,163 @@
...
+
+  Status CreateMigrateHandleThread(void);
+  void *Loop(void *arg);
+  Status MigrateStart(Server *svr, const std::string &node_id, const std::string dst_ip,
+                      int dst_port, int slot, int speed, int pipeline_size, int seq_gap);
+  void ReleaseForbiddenSlot();
+  void SetMigrateSpeedLimit(int speed) { if (speed >= 0) migrate_speed_ = speed; }
+  void SetPipelineSize(uint32_t size) { if (size > 0) pipeline_size_limit_ = size; }
+  void SetSequenceGapSize(int size) { if (size > 0) seq_gap_limit_ = size; }
+  void StartMigrateTask() { stop_ = false; }
```
ShooterIT: don't think this name is a good idea

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+
+  state_machine_ = kSlotMigrateNone;
+  current_pipeline_size_ = 0;
+  wal_begin_seq_ = 0;
+  wal_incremet_seq_ = 0;
+  std::lock_guard<std::mutex> guard(job_mutex_);
+  delete slot_job_;
+  slot_job_ = nullptr;
+  migrate_slot_ = -1;
+  StartMigrateTask();
```
ShooterIT: just set stop_ to 0?
this function name is not clear

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+  std::string cmd = Redis::MultiBulkString({"cluster", "import", std::to_string(slot), std::to_string(status)});
+  auto s = Util::SockSend(sock_fd, cmd);
+  if (!s.IsOK()) {
+    LOG(ERROR) << "[migrate] Failed to send import command to destination, slot: " << slot << ", error: " << s.Msg();
+    return false;
+  }
+
+  bool st = CheckResponseOnce(sock_fd);
+  if (!st) {
+    LOG(ERROR) << "[migrate] Failed to set destnation status " << status;
```
ShooterIT: should point specific state

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+
+Status SlotMigrate::MigrateOneKey(rocksdb::Slice key, std::string *restore_cmds) {
+  std::string prefix_key;
+  AppendNamespacePrefix(key, &prefix_key);
+  std::string bytes;
+  bool st = GetSlotKeyMetadata(prefix_key, &bytes);
+  if (!st) return Status(Status::NotOK, "[migrate] Failed to get key's metadata");
+  Metadata metadata(kRedisNone, false);
+  metadata.Decode(bytes);
+  if (metadata.Expired()) {
```
ShooterIT: for correctness, we should also migreate expired key, since we may increame its ttl, but it is hard to support currently, please know that.

src/slot_migrate.cc
```diff
@@ -0,0 +1,967 @@
...
+
+Status SlotMigrate::GetMigrateInfo(std::vector<std::string> *info, int16_t slot) {
+  info->clear();
+  if (migrate_slot_ < 0 && forbidden_slot_ < 0) {
+    info->push_back("There is no migrating slot");
+    return Status::OK();
+  }
+
+  if (slot != migrate_slot_ && slot != forbidden_slot_) {
+    return Status(Status::NotOK, "Input slot is not in migrating");
```
ShooterIT: input -> the

src/batch_extractor.h
```diff
@@ -0,0 +1,32 @@
...
+  explicit WriteBatchExtractor(bool is_slotid_encoded, int16_t slot = -1, bool to_redis = false)
+  : is_slotid_encoded_(is_slotid_encoded), slot_(slot), to_redis_(to_redis) {}
+  void LogData(const rocksdb::Slice &blob) override;
+  rocksdb::Status PutCF(uint32_t column_family_id, const Slice &key,
+                        const Slice &value) override;
+
+  rocksdb::Status DeleteCF(uint32_t column_family_id, const Slice &key) override;
+  rocksdb::Status DeleteRangeCF(uint32_t column_family_id,
+                                const Slice& begin_key, const Slice& end_key) override;
+  std::map<std::string, std::vector<std::string>> *GetAofStrings() { return &resp_commands_; }
```
ShooterIT: GetAofStrings -> GetRESPCommands ?

src/cluster.cc
```diff
@@ -660,6 +649,6 @@ Status Cluster::CanExecByMySelf(const Redis::CommandAttributes *attributes,
   } else {
     std::string ip_port = slots_nodes_[slot]->host_ + ":" +
                           std::to_string(slots_nodes_[slot]->port_);
-    return Status(Status::RedisExecErr, "MOVED " + std::to_string(slot) + " " + ip_port);
+    return Status(Status::RedisExecErr, "Serve MOVED " + std::to_string(slot) + " " + ip_port);
```
ShooterIT: can't change this, this error MUST be the same with redis

src/server.cc
```diff
@@ -992,6 +1010,16 @@ void Server::PrepareRestoreDB() {
   storage_->CloseDB();
 }
 
+void Server::WaitNoMigrateProcessing() {
+  if (config_->slot_id_encoded) {
```
```suggestion
  if (config_->cluster_enabled_) {
```

src/server.cc
```diff
@@ -992,6 +1010,16 @@ void Server::PrepareRestoreDB() {
   storage_->CloseDB();
 }
 
+void Server::WaitNoMigrateProcessing() {
+  if (config_->cluster_enabled_) {
```
```suggestion
  if (config_->cluster_enabled) {
```

src/batch_extractor.cc
```diff
@@ -0,0 +1,207 @@
...
+    if (metadata.Type() == kRedisString) {
+      command_args = {"SET", user_key, value.ToString().substr(5, value.size() - 5)};
+      resp_commands_[ns].emplace_back(Redis::Command2RESP(command_args));
+      if (metadata.expire > 0) {
+        command_args = {"EXPIREAT", user_key, std::to_string(metadata.expire)};
+        resp_commands_[ns].emplace_back(Redis::Command2RESP(command_args));
+      }
+    } else if (metadata.expire > 0) {
+      auto args = log_data_.GetArguments();
+      if (args->size() > 0) {
```
ShooterIT, line 38:
```suggestion
      if (args->size() >= 0) {
```

ShooterIT (approved):
LGTM, Cheers!
@ChrisZMF Thanks for your great PR which authentically makes kvrocks scalable , also thanks for your patience of keeping following CRs and tolerating my preferences.

git-hulk:
Cheers! @ShooterIT I didn't have other suggestions, can you help to summary this PR.

git-hulk (approved):

ChrisZMF:
I am really glad this PR is accepted. Thank you for the suggestions you given for this PR to make it better. It really makes a lot of sense to me. @ShooterIT @git-hulk

src/redis_cmd.cc
```diff
@@ -3450,13 +3467,21 @@ class CommandSlaveOf : public Commander {
       if (s.IsOK()) {
         *output = Redis::SimpleString("OK");
         LOG(WARNING) << "MASTER MODE enabled (user request from '" << conn->GetAddr() << "')";
+        if (svr->GetConfig()->cluster_enabled) {
+          svr->slot_migrate_->SetMigrateStopFlag(false);
+          LOG(INFO) << "Change server role to master, restart migration task";
+        }
```
enjoy-binbin, line 3473: isn’t this block of code are dead code? since in the beginning we will throw an error if we are in the cluster mode
```
    if (svr->GetConfig()->cluster_enabled) {
      return Status(Status::RedisExecErr, "can't change to slave in cluster mode");
    }
```

sorry for commenting in this very old PR... i tracked the changes, afraid i am missing something.

PR #431, opened 2021-12-17
Support to convert sortedint type to zset type in kvrocks2redis

fishery86:
covert sortint to sorted set

### DIFF
--- a/tools/kvrocks2redis/parser.cc
+++ b/tools/kvrocks2redis/parser.cc
@@ -252,7 +252,7 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
         break;
       }
       case kRedisSortedint: {
-        std::string val=std::to_string(DecodeFixed64(sub_key.data()));
+        std::string val = std::to_string(DecodeFixed64(sub_key.data()));
         if (!to_redis_) {
           command_args = {"SIADD", user_key, val};
         }else{
--- a/tools/kvrocks2redis/parser.cc
+++ b/tools/kvrocks2redis/parser.cc
@@ -255,7 +255,7 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
         std::string val = std::to_string(DecodeFixed64(sub_key.data()));
         if (!to_redis_) {
           command_args = {"SIADD", user_key, val};
-        }else{
+        } else {
           command_args = {"ZADD", user_key, val, val};
         }
         break;
--- a/tools/kvrocks2redis/parser.cc
+++ b/tools/kvrocks2redis/parser.cc
@@ -96,7 +96,7 @@ Status Parser::parseComplexKV(const Slice &ns_key, const Metadata &metadata) {
         break;
       }
       case kRedisSortedint: {
-        std::string val= std::to_string(DecodeFixed64(ikey.GetSubKey().data()));
+        std::string val = std::to_string(DecodeFixed64(ikey.GetSubKey().data()));
         output = Rocksdb2Redis::Command2RESP({"ZADD", user_key, val, val});
         break;
       }
@@ -256,7 +256,7 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
         if (!to_redis_) {
           command_args = {"SIADD", user_key, val};
         }else{
-          command_args={"ZADD", user_key, val, val};
+          command_args = {"ZADD", user_key, val, val};
         }
         break;
       }
@@ -331,10 +331,9 @@ rocksdb::Status WriteBatchExtractor::DeleteCF(uint32_t column_family_id, const S
         break;
       }
       case kRedisSortedint: {
-        std::string sub_key_str=std::to_string(DecodeFixed64(sub_key.data()));
-        cmd_str = to_redis_ ? "SREM" : "SIREM";
-        }
-        command_args={cmd_str, user_key, sub_key_str};
+        std::string sub_key_str = std::to_string(DecodeFixed64(sub_key.data()));
+        std::string cmd_str = to_redis_ ? "ZREM" : "SIREM";
+        command_args = {cmd_str, user_key, sub_key_str};
         break;
       }
       default: break;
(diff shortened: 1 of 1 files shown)


tools/kvrocks2redis/parser.cc
```diff
@@ -95,6 +95,11 @@ Status Parser::parseComplexKV(const Slice &ns_key, const Metadata &metadata) {
         s = Parser::parseBitmapSegment(ns, user_key, index, value);
         break;
       }
+      case kRedisSortedint: {
+        std::string val= std::to_string(DecodeFixed64(ikey.GetSubKey().data()));
+        output = Rocksdb2Redis::Command2RESP({"ZADD", user_key, val, val});
```
caipengbo, line 100: I think we should convert sorted int to Redis set type instead of sorted set type.

tools/kvrocks2redis/parser.cc
```diff
@@ -323,9 +331,12 @@ rocksdb::Status WriteBatchExtractor::DeleteCF(uint32_t column_family_id, const S
         break;
       }
       case kRedisSortedint: {
-        if (!to_redis_) {
-          command_args = {"SIREM", user_key, std::to_string(DecodeFixed64(sub_key.data()))};
+        std::string sub_key_str=std::to_string(DecodeFixed64(sub_key.data()));
+        std::string cmd_str="SIREM";
+        if (to_redis_) {
+          cmd_str="ZREM";
```
caipengbo: Using `? :` may be more concise.
```suggestion
        cmd_str = to_redis_ ? "SREM" : "SIREM";
```

tools/kvrocks2redis/parser.cc
fishery86: an equivalent  command in redis by type sortedin can be found by any command described  in docs/custom-api-sortedint.md. if using Redis set will losing some features,such as zrangebyvalue which an equivalent  command is  zrangebyscore?

ShooterIT: yes, we should use ZSET for "SORT"

caipengbo: I think sorted int set is just a special set type. Of course there are some special features added to Kvrocks, but these only work in Kvrocks and I think they can be discarded in Redis. 

WDYT @git-hulk @ShooterIT ?

fishery86: > I think sorted int set is just a special set type. Of course there are some special features added to Kvrocks, but these only work in Kvrocks and I think they can be discarded in Redis.
> 
> WDYT @git-hulk @ShooterIT ?

the first commit should applied  if we just want to discard sortedint type. if not,kvrocks2redis will not working ,it stops when tring to parse sortedint

caipengbo: I mean, discard some exclusive features, not the sortedint type.

ShooterIT: IIRC, after migrating form kvrocks to redis, if we want to use `range`-like commands, we must put sorted int keys into ZSET, so i think the idea of this PR is fine to me.

> sirangebyvalue -> zrange
> sirevrangebyvalue -> zrerange

caipengbo: OK, you convinced me, but I think we should explain it in kvrocks2redis/README.md.

tools/kvrocks2redis/parser.cc
```diff
@@ -323,9 +331,10 @@ rocksdb::Status WriteBatchExtractor::DeleteCF(uint32_t column_family_id, const S
         break;
       }
       case kRedisSortedint: {
-        if (!to_redis_) {
-          command_args = {"SIREM", user_key, std::to_string(DecodeFixed64(sub_key.data()))};
+        std::string sub_key_str=std::to_string(DecodeFixed64(sub_key.data()));
+        cmd_str = to_redis_ ? "SREM" : "SIREM";
```
ShooterIT: there is a compiling error, without definition. Please fix it @fishery86

tools/kvrocks2redis/parser.cc
```diff
@@ -323,9 +331,10 @@ rocksdb::Status WriteBatchExtractor::DeleteCF(uint32_t column_family_id, const S
...
         }
+        command_args={cmd_str, user_key, sub_key_str};
```
ShooterIT: and also please fix some code styles, a blank space is required before and after the operator.

ShooterIT (changes requested):

tools/kvrocks2redis/parser.cc
caipengbo: And change `SREM` to `ZREM`

tools/kvrocks2redis/parser.cc
```diff
@@ -247,8 +252,11 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
         break;
       }
       case kRedisSortedint: {
+        std::string val=std::to_string(DecodeFixed64(sub_key.data()));
```
caipengbo:
```suggestion
        std::string val = std::to_string(DecodeFixed64(sub_key.data()));
```

tools/kvrocks2redis/parser.cc
```diff
@@ -247,8 +252,11 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
         break;
       }
       case kRedisSortedint: {
+        std::string val = std::to_string(DecodeFixed64(sub_key.data()));
         if (!to_redis_) {
-          command_args = {"SIADD", user_key, std::to_string(DecodeFixed64(sub_key.data()))};
+          command_args = {"SIADD", user_key, val};
+        }else{
```
caipengbo:
```suggestion
        } else {
```

caipengbo (approved):
LGTM

ShooterIT:
thanks @fishery86 currently code also is fine to me.

But there are remaining jobs after merging

- [ ] Test kvrocks2redis (include  converting sortedint to zset) , and add tests into CI jobs
- [ ] Update kvrocks2redis readme.md for converting sortedint to zset implicitly

 could you do these jobs? @fishery86  or @caipengbo

ShooterIT (approved):

caipengbo:
I'm going to do that in the near future.

PR #436, opened 2021-12-26
Allow to  disable jemalloc when compiling

git-hulk:


git-hulk:
Allow to run the CI with jemalloc, but I think we should only run it manually labelled with `run ci without jemalloc` like the current PR. HDYT @ShooterIT

ShooterIT:
I think we can add this job into `kvrocks.yaml` since all jobs can run in parallel, this can't bring extra much time.

git-hulk:
@ShooterIT We take the job into the single file since we want to run it when the PR was labelled, and other jobs should be always running.

ShooterIT (approved):

PR #444, opened 2022-01-06
Support flushall/flushdb option

shangxiaoxiong:
When flushall or flushdb is executed, the disk is still occupied until a compaction occurs. this may bring confusion to users because to free disk space quickly is expected in most case. this commit makes kvrocks support flushall/flushdb with option which decides whether to compact immediately. Note that compaction caused by flushall/flushdb will spread to slave.

supported options: 
+ `sync:`   delete data and compact immediately
+ `async:` delete data
+ `none:`   the same with sync

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -143,21 +143,59 @@ class CommandKeys : public Commander {
 
 class CommandFlushDB : public Commander {
  public:
+  Status Parse(const std::vector<std::string> &args) override {
+     if (args.size() == 2 && !strcasecmp(args[1].data(), "async")) {
+       sync_ = false;
+     } else if (args.size() == 2 && !strcasecmp(args[1].data(), "sync")) {
+       sync_ = true;
+     } else if (args.size() == 1) {
+       sync_ = true;
+     } else {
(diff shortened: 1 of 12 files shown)


src/redis_cmd.cc
```diff
@@ -143,21 +143,59 @@ class CommandKeys : public Commander {
 
 class CommandFlushDB : public Commander {
  public:
+  Status Parse(const std::vector<std::string> &args) override {
+     if (args.size() == 2 && !strcasecmp(args[1].data(), "async")) {
+       sync_ = false;
```
git-hulk, line 148: It confused me that sync was determine whether to emit `compact` or not, instead of sync or async compact. Does it better to use other word like `with_compact` or others?

shangxiaoxiong:
l found a bug through test and l am locating

ShooterIT:
@shangxiaoxiong solved?

tisonkun:
Closed as stale.

Codebase evolves a lot since the last activities on this patch. @shangxiaoxiong you're welcome to propose the changes again based on the nightly unstable branch.

PR #445, opened 2022-01-06
Update grammar in metadata design doc

calvinxiao:
1. Change some grammar
2. Update the math in bitmap section


git-hulk:
@calvinxiao cool, can also help to fix the size in the metadata should be 4 bytes instead of 8 bytes.

ShooterIT (approved):

PR #446, opened 2022-01-07
Fix failing full replication when enabling slave-empty-db-before-fullsync

ShooterIT:
if enabling `slave-empty-db-before-fullsync`, replica will core dump since we operate against opened RocksDB, and renaming a non-exist directory never succeed.

Introduced in #233

**WARNING**: Please don't enable `slave-empty-db-before-fullsync`

### DIFF
--- a/src/storage.cc
+++ b/src/storage.cc
@@ -373,7 +373,7 @@ Status Storage::RestoreFromCheckpoint() {
   rocksdb::DestroyDB(config_->checkpoint_dir, rocksdb::Options());
 
   // Maybe there is no db dir
-  auto s = backup_env_->CreateDirIfMissing(config_->db_dir);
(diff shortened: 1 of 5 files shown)


src/storage.cc
```diff
@@ -415,17 +421,6 @@ void Storage::EmptyDB() {
   if (!s.ok()) {
     LOG(ERROR) << "[storage] Failed to destroy db, error: " << s.ToString();
   }
-
-  // Reopen db, it is empty if succeeded destroying db
-  auto s2 = Open();
```
git-hulk, line 420: The expected behavior of the request should be failed if we didn't re-open the db. Another question is wether the DestroyDB can be reenter or not since we also need to destroy the db before restoring.

git-hulk (approved):

PR #447, opened 2022-01-09
Support GET command on bitmap type

ColinChamber:
# Background
In Kvrocks, `bitmap` and `string` are two completely different types. Although `setbit`, `getbit`, `bitcount`, and `bitpos` operations on the string type are supported. @smartlee once mentioned that when using `setbit` on the string type, master-slave replication will have a tremendous amount of write amplification because Kvrocks replication does not use command propagation like Redis. In addition, using type `bitmap` conserves storage space. So we recommend using type bitmap. To make it easy to use, we want to get the whole string after `setbit` operations, so we implement this interface.

# Implementation
We covert bitmap to string when executing GET command and return string reply, otherwise, we provide one configuration item  `max-bitmap-to-string-mb` which uses to limit the max size of bitmap to string transformation(MB).

# Plan
We have implemented the **`bitop`** command. If this commit is approved, the "Redis bitops.tcl"  can be wholly converted to "Kvrocks bitmap.tcl" when submitting bitop.
# QA
* Why we need to swap the bit when converting bitmap type to string.
If you `setbit bit 0 1`, the value is stored as `0x01` in Kvocks but `0x80` in Redis. So  swapping bits is to get the same return value as Redis.

### DIFF
--- a/src/redis_bitmap.cc
+++ b/src/redis_bitmap.cc
@@ -1,5 +1,6 @@
 #include "redis_bitmap.h"
 #include <vector>
+#include <algorithm>
 
 #include "redis_bitmap_string.h"
 
@@ -7,6 +8,9 @@ namespace Redis {
 
 const uint32_t kBitmapSegmentBits = 1024 * 8;
 const uint32_t kBitmapSegmentBytes = 1024;
+
+const char kErrBitmapStringOutOfRange[] = "The size of the bitmap string exceeds the "
+                                          "configuration item max-bitmap-to-string-mb";
 
 uint32_t kNum2Bits[256] = {
     0, 1, 1, 2, 1, 2, 2, 3, 1, 2, 2, 3, 2, 3, 3, 4,
@@ -79,6 +83,61 @@ rocksdb::Status Bitmap::GetBit(const Slice &user_key, uint32_t offset, bool *bit
   return rocksdb::Status::OK();
 }
 
+// Use this function after careful estimation, and reserve enough memory
+// according to the max size of the bitmap string to prevent OOM.
+rocksdb::Status Bitmap::GetString(const Slice &user_key, const uint32_t &max_btos_size, std::string *value) {
+  value->clear();
+  std::string ns_key, raw_value;
+  AppendNamespacePrefix(user_key, &ns_key);
+
+  BitmapMetadata metadata(false);
+  rocksdb::Status s = GetMetadata(ns_key, &metadata, &raw_value);
+  if (!s.ok()) return s;
+  if (metadata.size > max_btos_size) {
+    return rocksdb::Status::Aborted(kErrBitmapStringOutOfRange);
+  }
+  value->assign(metadata.size, 0);
+
+  std::string fragment, prefix_key;
+  fragment.reserve(kBitmapSegmentBytes * 2);
+  InternalKey(ns_key, "", metadata.version, storage_->IsSlotIdEncoded()).Encode(&prefix_key);
+
+  rocksdb::ReadOptions read_options;
+  LatestSnapShot ss(db_);
+  read_options.snapshot = ss.GetSnapShot();
+  read_options.fill_cache = false;
+  uint32_t frag_index, valid_size;
+
+  auto iter = db_->NewIterator(read_options);
+  for (iter->Seek(prefix_key);
+       iter->Valid() && iter->key().starts_with(prefix_key);
+       iter->Next()) {
+    InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+    frag_index = std::stoul(ikey.GetSubKey().ToString());
+    fragment = iter->value().ToString();
+    // To be compatible with data written before the commit d603b0e(#338)
+    // and avoid returning extra null char after expansion.
+    valid_size = std::min({fragment.size(),
+                          static_cast<size_t>(kBitmapSegmentBytes),
+                          static_cast<size_t>(metadata.size - frag_index)});
+
+    for (uint32_t i = 0; i < valid_size; i++) {
+        if (!fragment[i]) continue;
+        fragment[i] = ((fragment[i] & 0x80) >> 7)|\
+                      ((fragment[i] & 0x40) >> 5)|\
+                      ((fragment[i] & 0x20) >> 3)|\
+                      ((fragment[i] & 0x10) >> 1)|\
+                      ((fragment[i] & 0x08) << 1)|\
+                      ((fragment[i] & 0x04) << 3)|\
+                      ((fragment[i] & 0x02) << 5)|\
+                      ((fragment[i] & 0x01) << 7);
+    }
+    value->replace(frag_index, valid_size, fragment.data(), valid_size);
+  }
+  delete iter;
+  return rocksdb::Status::OK();
+}
+
 rocksdb::Status Bitmap::SetBit(const Slice &user_key, uint32_t offset, bool new_bit, bool *old_bit) {
   std::string ns_key, raw_value;
   AppendNamespacePrefix(user_key, &ns_key);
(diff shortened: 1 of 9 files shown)


git-hulk:
cool, many thanks for @ColinChamber doing this improvement.

src/redis_bitmap.cc
```diff
@@ -79,6 +83,61 @@ rocksdb::Status Bitmap::GetBit(const Slice &user_key, uint32_t offset, bool *bit
...
+    fragment = iter->value().ToString();
+    // To be compatible with data written before the commit d603b0e(#338)
+    // and avoid returning extra null char after expansion.
+    valid_size = std::min({fragment.size(),
+                          static_cast<size_t>(kBitmapSegmentBytes),
+                          static_cast<size_t>(metadata.size - frag_index)});
+
+    for (uint32_t i = 0; i < valid_size; i++) {
+        if (!fragment[i]) continue;
+        fragment[i] = ((fragment[i] & 0x80) >> 7)|\
```
git-hulk, line 128: can help to explain why we need to swap the bit.

src/redis_cmd.cc
```diff
@@ -234,6 +234,12 @@ class CommandGet : public Commander {
     std::string value;
     Redis::String string_db(svr->storage_, conn->GetNamespace());
     rocksdb::Status s = string_db.Get(args_[1], &value);
+    if (s.IsNotSupported()) {
```
git-hulk:
```suggestion
    // The IsNotSupported error means the key type was a bitmap
    // which we need to fall back to the bitmap's GetString according
    // to the `max-bitmap-to-string-mb` configuration.
    if (s.IsNotSupported()) {
```

src/redis_bitmap.cc
ColinChamber: Sorry for my late reply. Swapping bits is to get the same return value as Redis. This is like the endian conversion because Kvrocks setbit storage differs from Redis.

ColinChamber:
> cool, many thanks for @ColinChamber doing this improvement.

Thanks a lot, nothing can be done for this improvement without @smartlee

src/redis_bitmap.cc
git-hulk: Cool, how about adding a comment to explain the reason? it would be helpful if we can't recall the context.

ColinChamber: OK, I'll update some details tomorrow.

src/redis_string.cc
```diff
@@ -54,6 +54,9 @@ rocksdb::Status String::getRawValue(const std::string &ns_key, std::string *raw_
     return rocksdb::Status::NotFound(kErrMsgKeyExpired);
   }
   if (metadata.Type() != kRedisString && metadata.size > 0) {
+    if (metadata.Type() == kRedisBitmap) {
```
git-hulk: There's a bit tricky here to use the NotSupported error as implicitly the metadata type was bitmap. But it's a bit ugly if we also return type on this function, do you have any idea? @ShooterIT @ColinChamber

ColinChamber: For now, I actually don't have any perfect idea as well.

ColinChamber:
> # Background
> In Kvrocks, `bitmap` and `string` are two completely different types. Although `setbit`, `getbit`, `bitcount`, and `bitpos` operations on the string type are supported. @smartlee once mentioned that when using `setbit` on the string type, master-slave replication will have a tremendous amount of write amplification because Kvrocks replication does not use command propagation like Redis. In addition, using type `bitmap` conserves storage space. So we recommend using type bitmap. But some of our colleagues want to get the whole string after `setbit`, so we implement this interface.
> # Plan
> We have implemented the **`bitop`** command. If this commit is approved, the "Redis bitops.tcl"  can be wholly converted to "Kvrocks bitmap.tcl" when submitting bitop.
> # QA
> * Why we need to swap the bit.
> If you `setbit bit 0 1`, the value is stored as `0x01` in Kvocks but `0x80` in Redis. So  swapping bits is to get the same return value as Redis.

@git-hulk

git-hulk:
Got it, thanks @ColinChamber

src/redis_bitmap.cc
```diff
@@ -79,6 +83,61 @@ rocksdb::Status Bitmap::GetBit(const Slice &user_key, uint32_t offset, bool *bit
...
+    InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+    frag_index = std::stoul(ikey.GetSubKey().ToString());
+    fragment = iter->value().ToString();
+    // To be compatible with data written before the commit d603b0e(#338)
+    // and avoid returning extra null char after expansion.
+    valid_size = std::min({fragment.size(),
+                          static_cast<size_t>(kBitmapSegmentBytes),
+                          static_cast<size_t>(metadata.size - frag_index)});
+
+    for (uint32_t i = 0; i < valid_size; i++) {
```
git-hulk, line 126:
```suggestion
    //  If you setbit bit 0 1, the value is stored as 0x01 in Kvocks but 0x80 in Redis.
    // So we need to swap bits is to keep the same return value as Redis.
    for (uint32_t i = 0; i < valid_size; i++) {
```

git-hulk (dismissed):
LGTM

src/redis_string.cc
git-hulk: It's fine to me before we have a better solution, @ShooterIT can take a look.

src/redis_string.cc
```diff
@@ -124,11 +127,10 @@ std::vector<rocksdb::Status> String::MGet(const std::vector<Slice> &keys, std::v
...
 
 rocksdb::Status String::Get(const std::string &user_key, std::string *value) {
-  std::vector<Slice> keys{user_key};
-  std::vector<std::string> values;
-  std::vector<rocksdb::Status> statuses = MGet(keys, &values);
-  *value = std::move(values[0]);
-  return statuses[0];
+  std::string ns_key;
+  AppendNamespacePrefix(user_key, &ns_key);
+  LockGuard guard(storage_->GetLockManager(), ns_key);
```
ShooterIT: is it necessary for this `LockGuard`?

src/redis_string.cc
ShooterIT: To be honest, i don't like this design, there is my idea

- don't have any change here
- if returning an `InvalidArgument` error when getting from string db, then try to get from bitmap db. I think we don't usually operate against wrong types in production environments, so don't worry performance loss

ShooterIT (commented):
Thanks @ColinChamber I like this idea, we should make kvrocks compatible with redis for well-known behaviors.

src/redis_string.cc
ColinChamber: Sorry, it is unnecessary.

src/redis_string.cc
ColinChamber: Thanks.I will modify it according to your suggestion.

ShooterIT:
@ColinChamber  anything updated? did you forget to push commits?

src/redis_bitmap.h
```diff
@@ -12,6 +12,7 @@ class Bitmap : public Database {
  public:
   Bitmap(Engine::Storage *storage, const std::string &ns): Database(storage, ns) {}
   rocksdb::Status GetBit(const Slice &user_key, uint32_t offset, bool *bit);
+  rocksdb::Status GetString(const Slice &user_key, const uint32_t &max_btos_size, std::string *value);
```
ShooterIT:
```suggestion
  rocksdb::Status GetString(const Slice &user_key, const uint32_t max_btos_size, std::string *value);
```
for plain types, i think we don't need to use `const reference`, right?

ColinChamber: You are right.

ShooterIT (approved):
LGTM, now, thanks for your patience @ColinChamber

ColinChamber:
Thanks for your guidance, too.

git-hulk (approved):
Cool, great work!

ShooterIT:
merged, truly like this idea, so now in most regular cases, users should not feel 

> bitmap and string are two completely different types.

PR #452, opened 2022-01-12
Add tcl tests about redis-cli

shangxiaoxiong:
Add tcl tests about redis-cli to make sure that redis-cli can work well with kvrocks.


git-hulk:
cool, it's too heavy to import the Redis repository to install the redis-cli, how about using apt/brew to install on CI steps and required to install the redis-cli if wants to run the tcl tests on local.

ShooterIT:
yes, only installing redis-cli in CI is a good idea.
in default redis tcl tests, you don't include this test, we can test it by `--single`.

PS:
I wanted to say we only do that in CI when i left your desk yesterday 😜, actually, redis is not a dependency for kvrocks

git-hulk:
Thanks hard work from @shangxiaoxiong, I see the last modified, my doubt is yum/brew didn't work? so we need to compile the Redis ourself?

shangxiaoxiong:
Redis with the right version should be installed to ensure that tcl tests pass. I failed on my mac. @git-hulk 
There may be no corresponding installation package in some env.

git-hulk:
@shangxiaoxiong I'm sorry for didn't your point, do you mean the older or newer redis-cli was failed on you side? Can we find out those test cases?

shangxiaoxiong:
Older redis-cli was failed.

ShooterIT (approved):

git-hulk (approved):

Issue #448, opened 2022-01-09
[BUG] Maybe there's a bug in the publish/subscribe logic


git-hulk:
**Describe the bug**
The github actions occasionally failed at Publish/Subscribe tcl test cases like below: 

https://github.com/KvrocksLabs/kvrocks/runs/4751562466?check_suite_focus=true#step:7:725

After reviewing the test case many times and it always happened on the ubuntu which was multi-threads, so i think it should be caused by the data race condition in the publish/subscribe logical, but still didn't find the right way to reproduce this issue.

git-hulk:
Maybe we can have a try at Ubuntu with enabling the thread sanitizer.

tisonkun:
Closed as stale. This issue seems vanished.

PR #460, opened 2022-01-17
Fix a cmake string subcommand error.

mzygQAQ:
env : MacosX
cmake_version: 3.18.4

When i execute cmake to compile, a error reported as the following:
[cmake] CMake Error at CMakeLists.txt:52 (string):
[cmake]   string sub-command STRIP requires two arguments.

While CMakeLists.txt file does syntactically contain two arguments, ${GIT_SHA} can be empty. If so, it is not an argument semantically and never reaches string(), which then sees just one. If it's possible for a string to be empty (and you want to treat it as an empty string in such case instead of skipping it), quote it:
_string(STRIP "${GIT_SHA}" GIT_SHA)_

I found an same error in stackoverflow, link: [https://stackoverflow.com/questions/26585513/cmake-error-with-string-sub-command-strip-requires-two-arguments](url)

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -44,8 +44,13 @@ jobs:
 
       - name: Redis Tcl Test
         run: |
+          wget https://github.com/redis/redis/archive/refs/tags/6.2.6.tar.gz
+          tar -zxvf 6.2.6.tar.gz
+          cd redis-6.2.6/ && make redis-cli && cd -
+          cp redis-6.2.6/src/redis-cli tests/tcl/redis-cli
           sudo apt-get install tcl8.5
-          cd tests/tcl && sh runtest && cd -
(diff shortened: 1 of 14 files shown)


mzygQAQ:
TCL test script failed:  tests/unit/pubsub.tcl

git-hulk:
Thanks @guoxiangCN, for the pubsub tcl test case, we are still investigating the root cause. Can see the issue: https://github.com/KvrocksLabs/kvrocks/issues/448, no worry about this case.

mzygQAQ:
> Thanks @guoxiangCN, for the pubsub tcl test case, we are still investigating the root cause. Can see the issue: #448, no worry about this case.

Thanks for your reply, I'll pay attention, too.

git-hulk (approved):

ShooterIT (approved):

PR #463, opened 2022-01-19
Support CLUSTERX SETSLOT command

ShooterIT:
**Command formart**

CLUSTERX SETSLOT $slot_id NODE $node_id $new_version

assign the slot to the node if new version is current version+1.

**Parameters**
>  $slot_id: the slot which we want to assign
>  $node_id: the node which we want to assign the slot into
>  NODE: keep the same with redis CLUSTER SETSLOT command, maybe we want to expand this command in the future
>  $new_version: the new version MUST be +1 of current version so that kvrocks could execute this command, this is to guarantee this change is based on a special cluster topology 

**Purpose**
- users can use this command to set slot distribution
- slot migration needs this command to update slot distribution

**New version constraint**
The reason why the new version MUST be +1 of current version is that, the command changes topology based on specific topology (also means specific version), we must guarantee current topology is exactly expected, otherwise this update may make topology corrupt, so base topology version is very important. This is different with CLUSTERX SETNODES commands because it uses new version topology to cover current version, it allows kvrocks nodes lost some topology updates since of network failure, it is state instead of operation.

### DIFF
--- a/src/cluster.cc
+++ b/src/cluster.cc
@@ -30,6 +30,8 @@ bool Cluster::SubCommandIsExecExclusive(const std::string &subcommand) {
     return true;
   } else if (strcasecmp("setnodeid", subcommand.c_str()) == 0) {
     return true;
+  } else if (strcasecmp("setslot", subcommand.c_str()) == 0) {
+    return true;
   }
   return false;
 }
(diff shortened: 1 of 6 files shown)


src/redis_cmd.cc
```diff
@@ -4328,8 +4330,25 @@ class CommandClusterX : public Commander {
       }
       return Status(Status::RedisParseErr, "Invalid setnodes options");
     }
+
+    // CLUSTERX SETSLOT $SLOT_ID NODE $NODE_ID $VERSION
+    if (subcommand_ == "setslot" && args_.size() == 6) {
+      slot_id_ = atoi(args_[2].c_str());
+      if (slot_id_ < 0 || slot_id_ >= kClusterSlots) {
```
ChrisZMF: consider replacing by `IsValidSlot()`

ShooterIT: cool, thanks, i forgot this API

src/cluster.cc
```diff
@@ -53,6 +55,52 @@ Status Cluster::SetNodeId(std::string node_id) {
...
+  }
+
+  // Update topology
+  //  1. Remove the slot from old node if existing
+  //  2. Add the slot into to-assign node
+  //  3. Update the map of slots to nodes.
+  std::shared_ptr<ClusterNode> old_node = slots_nodes_[slot];
+  if (old_node != nullptr) {
+    old_node->slots_[slot] = 0;
+  }
```
ChrisZMF, line 100: Is it necessary to handle `old_node == to_assign_node` ?

ShooterIT: I ever thought, but i found we don't need since it don't change anything in current code, anything i miss?

src/cluster.cc
```diff
@@ -53,6 +55,55 @@ Status Cluster::SetNodeId(std::string node_id) {
...
+  std::shared_ptr<ClusterNode> to_assign_node = nodes_[node_id];
+  if (to_assign_node == nullptr) {
+    return Status(Status::NotOK, "No this node in the cluster");
+  }
+  if (to_assign_node->role_ != kClusterMaster) {
+    return Status(Status::NotOK, "The node is not the master");
+  }
+
+  // Update version
+  version_ = new_version;
```
git-hulk, line 91: We need to mark the clusterx command as exclusive to prevent data race?

ShooterIT: yes, for `clusterx` command, some subcommands already have `exclusive` attribute, `setslot` is one of them, please see `Cluster::SubCommandIsExecExclusive`

git-hulk: ok, I forgot this point, only have a look at the command table.

git-hulk (approved):

src/cluster.cc
ChrisZMF: It's OK.

PR #465, opened 2022-01-19
Fix the bitmap range is smaller than that of Redis

### LINKED ISSUE #454: [QUESTION]Setbit max offset limit
Kvrocks' limit on setbit offset seems narrow, less than the maximum value in Redis (512M*8-1). According to the design of `metadata size(4byte)`, the max offset can even be larger than Redis. Why is the limit `INT_MAX` #164? Should `std::stoi` not be used in `bitcount`? Is there some context I don't comprehend @ShooterIT ?

screenshot_358
screenshot_359

ColinChamber:
Background in issue #454, I originally planned to expand the permitted range of `start end` parameters of `bitcount bitop` to be the same as in Redis. But, I found that there were too many changes, and it was unnecessary to do that because the offset of setbit is at most `2^32-1`. So it remains `std::stoi` doesn't seem to be a problem after the fix because start/end is different from offset,  one is byte, and the other is bit.

closes #454

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -729,7 +729,7 @@ Status getBitOffsetFromArgument(std::string arg, uint32_t *offset) {
   } catch (std::exception &e) {
     return Status(Status::RedisParseErr, errValueNotInterger);
   }
-  if (offset_arg < 0 || offset_arg > INT_MAX) {
+  if (offset_arg < 0 || offset_arg > UINT_MAX) {
     return Status(Status::RedisParseErr, "bit offset is out of range");
   }
   *offset = static_cast<uint32_t>(offset_arg);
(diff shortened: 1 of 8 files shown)


src/redis_cmd.cc
```diff
@@ -788,11 +788,16 @@ class CommandSetBit : public Commander {
...
-      if (args.size() >= 3) start_ = std::stoi(args[2]);
-      if (args.size() >= 4) stop_ = std::stoi(args[3]);
-    } catch (std::exception &e) {
-      return Status(Status::RedisParseErr, errValueNotInterger);
+    if (args.size() == 3) {
+      return Status(Status::RedisParseErr, errInvalidSyntax);
+    }
+    if (args.size() == 4) {
+      try {
+        start_ = std::stoi(args[2]);
```
git-hulk: How about changing the start and stop to int64 as well? since the newer Redis version can support byte|bit mode.

ColinChamber: What is byte|bit mode? I didn't find it, can you give me some references?

git-hulk: Can see the doc: https://redis.io/commands/bitcount

the last parameter can be bit or byte.

ColinChamber: > Redis version >= 7.0.0: Added the `BYTE|BIT` option.

Thank you, I just learned that.

ColinChamber: Did you just want to replace it with int64, or did we implement this option too? Redis 7.0.0 doesn't seem to be released.

git-hulk: yes, just replace them with int64, to prevent suffering the same problem when we want to implement the option.

ColinChamber: > BITPOS key bit [start [end [BYTE|BIT]]]

OK, this part as well.

git-hulk: cool~

git-hulk (approved):

ShooterIT (approved):

git-hulk:
Merged, thanks @ColinChamber

PR #442, opened 2021-12-31
Fix the misuse of multi-get API

ColinChamber:
Perhaps the MulitGet interface with seven parameters is more suitable for getting multiple values ​​from the metadata column family at once. The [original interface used](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/db.h#L507) is to get many keys **across column families** consistently. There are more details in the rocksdb source code [comments](https://github.com/facebook/rocksdb/blob/main/include/rocksdb/db.h#L554).

PR #467, opened 2022-01-25
Revert "Fix the misuse of multi-get API"

shangxiaoxiong:
Reverts KvrocksLabs/kvrocks#442
l find that new multi-get api can get keys which have been deleted before. the case could be reproduced with my PR [KvrocksLabs/kvrocks/pull/444](https://github.com/KvrocksLabs/kvrocks/pull/444)  . It seems a bug of RocksDB and I would submit a relevant PR to RocksDB in the near future.

### DIFF
--- a/src/redis_string.cc
+++ b/src/redis_string.cc
@@ -14,14 +14,10 @@ std::vector<rocksdb::Status> String::getRawValues(
   rocksdb::ReadOptions read_options;
   LatestSnapShot ss(db_);
   read_options.snapshot = ss.GetSnapShot();
-  raw_values->resize(keys.size());
-  std::vector<rocksdb::Status> statuses(keys.size());
-  std::vector<rocksdb::PinnableSlice> pin_values(keys.size());
-  db_->MultiGet(read_options, metadata_cf_handle_, keys.size(),
-                keys.data(), pin_values.data(), statuses.data(), false);
+  std::vector<rocksdb::ColumnFamilyHandle*> cfs(keys.size(), metadata_cf_handle_);
+  auto statuses = db_->MultiGet(read_options, cfs, keys, raw_values);
   for (size_t i = 0; i < keys.size(); i++) {
     if (!statuses[i].ok()) continue;
-    (*raw_values)[i].assign(pin_values[i].data(), pin_values[i].size());
     Metadata metadata(kRedisNone, false);
     metadata.Decode((*raw_values)[i]);
     if (metadata.Expired()) {


ShooterIT:
FYI @ColinChamber

ShooterIT (approved):

git-hulk (approved):

ColinChamber:
I did not reproduce by executing some simple commands in Redis-CLI. You just provided code for a new feature. Please give the execution steps. @shangxiaoxiong

shangxiaoxiong:
The new multi-get api works well in most scenarios which include your mentioned tests by redis-cli. But in some corner cases, it works not as expected. Tests in flush.tcl of my PR [444](https://github.com/KvrocksLabs/kvrocks/pull/444) may be one of the corner cases. The test case shows that the key deleted in master can be read out in slave. It should be noted that the test case has specific context. There may be a problem with my implementation of the new feature but possibility is very small in my view. The stronger evidence is that the new feature [444](https://github.com/KvrocksLabs/kvrocks/pull/444) works well with old multi-get api. @ColinChamber

git-hulk:
@shangxiaoxiong Did you find the root cause in `MultiGet` API?

ColinChamber:
Thanks @shangxiaoxiong . Just give me a second to reproduce it in `flush.tcl`

ColinChamber:
> @shangxiaoxiong Did you find the root cause in `MultiGet` API?

Indeed, `MultiGet` API doesn't seem to be the root cause. I reproduced but didn't merge #444.
Here are my steps.
I comment out [here](https://github.com/shangxiaoxiong/kvrocks/blob/support_flush_async/tests/tcl/tests/unit/flush.tcl#L31)
```tcl
              # check log file of slave
              # if {$flag eq "sync"} {
              #     wait_for_condition 50 1000 {
              #         [log_file_matches $dir/kvrocks.INFO "*exec propagate compact*"]
              #     } else {
              #         fail "Fail to propagate compact to salve when flushall/db sync"
              #     }  
              # }
```
changes [here](https://github.com/KvrocksLabs/kvrocks/blob/unstable/src/redis_cmd.cc#L4492) 
```cpp
ADD_CMD("flushdb", -1, "write", 0, 0, 0, CommandFlushDB),
ADD_CMD("flushall", -1, "write", 0, 0, 0, CommandFlushAll),
```
no problem found
image

Maybe we should change the direction of suspicion @git-hulk  @shangxiaoxiong .

git-hulk:
Cool

shangxiaoxiong:
The new feature of my PR [444](https://github.com/KvrocksLabs/kvrocks/pull/444) is part of the corner case and you just miss it. @ColinChamber

shangxiaoxiong:
I have a general idea of the suspected bug and l try to fix it in RocksDB. The new feature works well with fixed new multi-get api. But I don't fully understand in some points, that is why l do not submit fix PR to RocksDB now. Just give me some time. @git-hulk

shangxiaoxiong:
image

ColinChamber:
> The new feature works well with fixed new multi-get api. 

Maybe this is just a superficial phenomenon and not the root cause.
    I've seen the ci bug record for #444, and since it's an unmerged feature, we can't rule out that the flush modification introduced the bug. So I reproduced it using the `flush.tcl` you provided **without introducing #444**. My reproduction shows that writing the master, flush, and then reading the masters and slaves gives the expected results. Of course, I can't rule out that there is no problem with this interface of RocksDB. There may be three reasons for this problem: **the modified flush; the replication logic has problems, the RocksDB mget is unreliable**. So, we can't quickly draw conclusions based on a single case.Today, I will take time to check out the details of #444. We should have an in-depth discussion on #442 #444 #467 and merge with caution. HDYT @shangxiaoxiong.

shangxiaoxiong:
Thank you for your serious reply. Your analysis is rigorous. There are many factors affecting the results. If separate RocksDB test case about the new multi get api is provided, it is easier to explain this problem. I will try to do that. @ColinChamber

ColinChamber:
I have read your code, and it seems to be ok. Thank you for this interesting case. I am looking forward to your separated RocksDB test case. @shangxiaoxiong

ShooterIT:
@shangxiaoxiong @ColinChamber is there any conclusion?

ShooterIT:
I want to release a new ASAP

ColinChamber:
I can't conclude now. Wait for shangxiaoxiong to provide a separated rocksdb test case. If you want to release a new ASAP,  I think merging this PR would work too.

ColinChamber:
Maybe you can wait a moment. I will submit `bitop` tomorrow. @ShooterIT

ShooterIT:
@ColinChamber don't worry about new feature, for current release, we should guarantee there is no bug, we will keep optimize kvrocks 2.0 branch.

I wish you could have a conclusion @shangxiaoxiong so i can decide we could merge it or not

ShooterIT:
@shangxiaoxiong @ColinChamber have a look https://github.com/facebook/rocksdb/pull/9453

ShooterIT:
we already update rocksdb to 6.9.25, this bug https://github.com/facebook/rocksdb/pull/9453 is fixed, i think we don't need to revert, and i can release this PR #442 in next version, and #444 can be merged.

cc @shangxiaoxiong @ColinChamber

ShooterIT:
closed, as describe above. reopen again if there are problems i missed.

PR #450, opened 2022-01-10
Support to compile on macOS 11

ShooterIT:
- [ ] leak reports many errors

PR #440, opened 2021-12-30
Fix data race in `SetCurrentConnection`

calvinxiao:
I found one data race using Thread Sanitizer.

`SetCurrentConnection` was introduced by #363 , and it can only be used with exclusive commands.

Fixed by moving `svr_->SetCurrentConnection(this);` after the `ExclusivityGuard` is acquired.

Verified with these benchmark commands:

- `redis-benchmark -p 6666 -r 10000 -n 10000 eval 'return redis.call("ping")' 0`
- `redis-benchmark -p 6666`

Here is the Thread Sanitizer output:

```
./src/kvrocks -c ./kvrocks.conf
Version: 999.999.999 @fc00bf69
==================
WARNING: ThreadSanitizer: data race (pid=299262)
  Write of size 8 at 0x7b6800000140 by thread T26:
    #0 Server::SetCurrentConnection(Redis::Connection*) /home/calvin/github/kvrocks/src/./server.h:140:73 (kvrocks+0x71a801)
    #1 Redis::Connection::ExecuteCommands(std::vector<std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, std::allocator<std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > > > const&) /home/calvin/github/kvrocks/src/redis_connection.cc:295:9 (kvrocks+0x71a801)
    #2 Redis::Connection::OnRead(bufferevent*, void*) /home/calvin/github/kvrocks/src/redis_connection.cc:68:9 (kvrocks+0x71a1cf)
    #3 bufferevent_run_deferred_callbacks_unlocked /home/calvin/github/kvrocks/external/libevent/bufferevent.c:208:3 (kvrocks+0x7d4aed)
    #4 WorkerThread::Start()::$_0::operator()() const /home/calvin/github/kvrocks/src/worker.cc:332:22 (kvrocks+0x7aba65)
    #5 void std::__invoke_impl<void, WorkerThread::Start()::$_0>(std::__invoke_other, WorkerThread::Start()::$_0&&) /usr/lib/gcc/x86_64-redhat-linux/11/../../../../include/c++/11/bits/invoke.h:61:14 (kvrocks+0x7aba65)
[...]

PR #434, opened 2021-12-24
Fix zrevrangebyscore for multi members with the same max score

shangxiaoxiong:
The reply of "zrevrangebyscore zset_key  min_score  max_score" just include one member for max_score when the zset_key has multi members with the same max_score.

PR #468, opened 2022-01-27
Release 2.0.6

ShooterIT:
# Version 2.0.6

## New features
  - Support slot-based data migration (#430)
    Biggest feature of this version, authentically makes kvrocks scalable,
    user can use CLUSTERX MIGRATE command to migrate slot and use CLUSTERX
    SETSLOT to change cluster slot distribution after finishing migration.
  - Make level0_file_num_compaction_trigger of RocksDB configurable (#455)
  - Support CLUSTERX SETSLOT command (#463)

## Improvements
  - Support executing INFO command during restoring database (#375)
  - Auto rewrite the config after the namespace was updated (#423)
  - Support to convert sortedint type to zset type in kvrocks2redis (#431)
  - Optimize the seek of RocksDB iterators (#438)
  - Make the bitmap range the same with that of Redis (#465)
  - Support GET command on bitmap type (#447)
  - Support to compile on macOS 11 M1 inside (#450)

## Bug fixes
  - Fix the parameter blob_garbage_collection_age_cutoff would be always zero (#425)
  - Fix config rewrite overwrite the first rename-command with the last one (#429)
  - Fix zrevrangebyscore for multi members with the same max score (#434)
  - Fix data race in SetCurrentConnection (#440)
  - Fix failing full replication when enabling slave-empty-db-before-fullsync (#446)
  - Fix memory leak when full replication (#461)
    Maybe leak dozen kilobytes per full replication


git-hulk (approved):

ShooterIT:
@ColinChamber @shangxiaoxiong  Since the RocksDB multi get API which is used in #442 has dispute, i don't merge it into 2.0 branch, but i also don't revert it before we have a clear conclusion, so it still is in unstable branch, please you know that.

git-hulk (approved):

git-hulk (approved):

PR #471, opened 2022-01-31
Support BITOP command

ColinChamber:


git-hulk (dismissed):
cool, thanks for your hard work! @ColinChamber 

This implementation may cause performance issue when the bitmap was sparsely, but it's more clear and simple than using the seek operator. So it looks good to me. HDYT @ShooterIT

git-hulk:
> cool, thanks for your hard work! @ColinChamber
> 
> This implementation may cause performance issue when the bitmap was too sparsely, but it's more clear and simple than using the seek operator. So it looks good to me. HDYT @ShooterIT

ColinChamber:
>This implementation may cause performance

I think so.
Currently, the bitop scene can't simply use the seek operator. Kvrocks use the default ordering function for key, which orders bytes lexicographically. For example, the order of the bitmap subkey after iteration is  `key|version|1024, key|version|10240, key|version|2048`, not `key|version|1024, key|version|2048, key|version|10240`, which will make the operation of the same offset fragment between different keys unattainable. I think this is also the reason why [bitcount](https://github.com/KvrocksLabs/kvrocks/blob/unstable/src/redis_bitmap.cc#L227) and [bitpos](https://github.com/KvrocksLabs/kvrocks/blob/unstable/src/redis_bitmap.cc#L285) can't use the seek operator. In the future, we can do some refactoring, modify the encode and decode of the bitmap, and fill in the same number of digits. @git-hulk

git-hulk:
@ColinChamber Yeah, the key order was easy to change, what i think was to seek out all fragments and apply the operator on those fragments without depending on the key order.

ColinChamber:
Do you mean `GetString` and then do the operation? That seems not a suitable method that will use too much memory and is inefficient when operating on **multiple keys**.

git-hulk:
@ColinChamber RocksDB write batch would also cache those fragments on memory before writing indeed, so it should be not different.

ColinChamber:
> RocksDB write batch would also cache those fragments on memory before writing indeed.

That is the memory consumed by destkey. 
Seeking out all  fragments will increase the memory consumed by srckey.

>* BITOP AND destkey srckey1 srckey2 srckey3 ... srckeyN
>* BITOP OR destkey srckey1 srckey2 srckey3 ... srckeyN
>* BITOP XOR destkey srckey1 srckey2 srckey3 ... srckeyN
>* BITOP NOT destkey srckey

git-hulk:
Yeah, we can apply those fragments one by one, so the number of memory was same as current solution.

ColinChamber:
It will depend on the subkey order. If we want to seek, then iterate one by one.

git-hulk:
Sorry that I didn't explain clearly. What I mean was below:

1. alloc the dest bitmap with the max size
2. seek key fragments then apply the operator and fragment into the dest bitmap fragment with the subkey(it's ok that subkey was lexical order)
3. repeat step 2 if we have many keys

ColinChamber:
Sorry for my late understanding. Your method will improve performance, but the number of calculations will increase. I think the best way is to iterate and calculate all srckeys simultaneously, which relies on getting subkeys in index order. 

Getting subkeys in index order can also improve the performance of bitcount. What is the reason we don't do that? Is it to maintain compatibility with the previous version?

git-hulk:
For why we didn't use seek operator at bitcount command, it's a design mistake because we only focus on hot commands like getbit and setbit, and didn't carefully design the key format before implementing the bitcount.

A compatible way is to use the extra bits to flag the new key format, but it's too tricky which I don't think we should do it now. So current implementation was good to me and can refactor it if those commands performance issue was the key point for us.

src/redis_bitmap.cc
```diff
@@ -313,6 +315,185 @@ rocksdb::Status Bitmap::BitPos(const Slice &user_key, bool bit, int64_t start,
...
+      return s;
+    }
+    if (metadata.Type() == kRedisString) {
+      return rocksdb::Status::InvalidArgument(kErrMsgWrongType);
+    }
+    if (metadata.size > max_size) max_size = metadata.size;
+    meta_pairs.emplace_back(std::make_pair(ns_op_key, metadata));
+  }
+
+  rocksdb::WriteBatch batch;
```
ShooterIT, line 346: @ChrisZMF @git-hulk here we should call PutLogData for record type or somethings, WDYT? and for slot_migration, should we need to adapt?

git-hulk: It makes sense for me since we have done that for parsing batch.

src/redis_bitmap.cc
```diff
@@ -313,6 +315,191 @@ rocksdb::Status Bitmap::BitPos(const Slice &user_key, bool bit, int64_t start,
...
+    AppendNamespacePrefix(op_key, &ns_op_key);
+    s = GetMetadata(ns_op_key, &metadata, &raw_value);
+    if (!s.ok()) {
+      if (s.IsNotFound()) {
+        num_keys--;
+        continue;
+      }
+      return s;
+    }
+    if (metadata.Type() == kRedisString) {
```
ChrisZMF, line 339: Why not consider the type `BitmapString`?

ColinChamber: When using setbit on the string type, master-slave replication will have a tremendous amount of write amplification because Kvrocks replication does not use command propagation like Redis. In addition, using type bitmap conserves storage space. So we recommend using type bitmap.

ColinChamber: Maybe in the future we can disable `BitmapString` by default in kvrocks unless the user modifies the kvrocks.conf. HDYT @ChrisZMF  @git-hulk

ColinChamber: Maybe in the future, operations between `BitmapString` can be supported separately, which is not difficult. If you want to support the operation between `bitmap` and `string`, the logic will become very complicated. It seems unnecessary.

git-hulk:
I have no advice on this PR. @ShooterIT @ChrisZMF

src/redis_bitmap.cc
ChrisZMF: It's OK, because if `max_size == 0` key will be deleted which is equivalent to executing command `DEL`. The logdata is unnecessary for parsing `DEL`.

src/redis_bitmap.cc
ChrisZMF: Got it, thx. It can be considered in the future according to user needs.

ChrisZMF:
> 

LGTM @ShooterIT

ShooterIT (approved):

git-hulk (approved):

PR #472, opened 2022-02-07
Avoid accessing slot_migrate_ before it is created

ChrisZMF:
Make sure to create slot_migrate_ object before starting the worker threads to fix potential bug.
Case:
Currently, slot_migrate_ object is created after worker threads. There exists potential risks. Because slot_migrate_ object may be accessed before it is created.
Detail reason:
- If worker threads are started, kvrocks will accept requests. During handling request, slot_migrate_ will be accessed to check whether the key of request belongs to the write forbidden slot. See #412 for more about write forbidden slot. Also, see code `Cluster::IsWriteForbiddenSlot`.
- Now, the worker threads are started before the slot_migrate_ object is created. If there is a request is accepted before slot_migrate_ is created. The accessing of slot_migrate_ object will cause coredump.

### DIFF
--- a/src/cluster.cc
+++ b/src/cluster.cc
@@ -246,7 +246,7 @@ Status Cluster::SetSlotImported(int slot) {
 
 Status Cluster::MigrateSlot(int slot, const std::string &dst_node_id) {
   if (nodes_.find(dst_node_id) == nodes_.end()) {
-    return Status(Status::NotOK, "Cannot find the destination node id");
+    return Status(Status::NotOK, "Can't find the destination node id");
   }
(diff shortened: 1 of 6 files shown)


git-hulk (approved):
cool, LGTM with a bit advice.

This PR resolves below issues:
* Fix slot_migrate_ was initialed after starting workers which may cause coredump
* Fix typo
* Fix wrong initial current_pipeline_size_ value
* Ignore some cluster config items if the cluster mode was disabled

but we squashed those into a single commit would be NOT easy to review, maybe we can split into multi commits in one PR.

ShooterIT (approved):

PR #495, opened 2022-02-23
fix set thread name fail with string more than 16 byte.

patpatbear:
according to `man pthread_setname_np`, thread name should be  ` restricted to 16 characters, including the terminating null byte ('\0'). `

```
 By  default,  all the threads created using pthread_create() inherit the program name.  The pthread_setname_np() function can be used to set a unique name for a thread, which can be useful for debugging multithreaded applica‐
       tions.  The thread name is a meaningful C language string, whose length is restricted to 16 characters, including the terminating null byte ('\0').  The thread argument specifies the thread whose name is to be  changed;  name
       specifies the new name.
```

### DIFF
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -25,7 +25,7 @@ FeedSlaveThread::~FeedSlaveThread() {
 Status FeedSlaveThread::Start() {
   try {
     t_ = std::thread([this]() {
-      Util::ThreadSetName("feed-slave-thread");
+      Util::ThreadSetName("feed-replica");
       sigset_t mask, omask;
       sigemptyset(&mask);
       sigemptyset(&omask);
--- a/src/server.cc
+++ b/src/server.cc
@@ -86,7 +86,7 @@ Status Server::Start() {
   compaction_checker_thread_ = std::thread([this]() {
     uint64_t counter = 0;
     int32_t last_compact_date = 0;
-    Util::ThreadSetName("compaction-checker");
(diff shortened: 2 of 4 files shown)


src/server.cc
```diff
@@ -86,7 +86,7 @@ Status Server::Start() {
   compaction_checker_thread_ = std::thread([this]() {
     uint64_t counter = 0;
     int32_t last_compact_date = 0;
-    Util::ThreadSetName("compaction-checker");
+    Util::ThreadSetName("comp-checker");
```
ShooterIT: `comp` is not clear, how about
```suggestion
    Util::ThreadSetName("compact-check");
```

src/replication.cc
```diff
@@ -25,7 +25,7 @@ FeedSlaveThread::~FeedSlaveThread() {
 Status FeedSlaveThread::Start() {
   try {
     t_ = std::thread([this]() {
-      Util::ThreadSetName("feed-slave-thread");
+      Util::ThreadSetName("feed-slave");
```
ShooterIT: Let's avoid `slave` from outside observation
```suggestion
      Util::ThreadSetName("feed-replica");
```

ShooterIT (commented):
LGTM, thanks
But please submit PR to unstable branch, i will merge it into 2.0 in the next release.

patpatbear:
opened new PR to unstable branch https://github.com/KvrocksLabs/kvrocks/pull/496, closing this one.

PR #502, opened 2022-02-25
Add lz4 and zstd compression algorithms options support.

WyattJia:


WyattJia:
I think I should append some documentation about how to install compression libraries.

git-hulk:
Yes, agreed. We should make sure it works after installing depend libraries.

ShooterIT:
And you should also set these flags when compiling rocksdb. It doesn't work now i think

WyattJia:
> And you should also set these flags when compiling rocksdb. It doesn't work now i think

Yep, I'll add documentation, compilation flags and test cases later.

git-hulk:
https://github.com/KvrocksLabs/kvrocks/blob/unstable/build_detect_platform#L127

Maybe it should work since we added the detection, but didn't have a try on my side.

WyattJia:
> https://github.com/KvrocksLabs/kvrocks/blob/unstable/build_detect_platform#L127
> 
> Maybe it should work since we added the detection, but didn't have a try on my side.

Copy that, I'll fill and check the flags.Thanks.

git-hulk:
Thanks for your hard work. @WyattJia

PR #503, opened 2022-02-25
Fix blob file size unit error and set default value to 256 MB

WyattJia:
* Unit error: `blob_file_size` should be Bytes number , not MiB.
* Default value: 
   * `min_blob_size` default value is `0` Bytes, see:
   https://github.com/facebook/rocksdb/blob/main/include/rocksdb/advanced_options.h#L848
   * `blob_file_size` default value is 256 MiB, 268435456 Bytes ,see: https://github.com/facebook/rocksdb/blob/main/include/rocksdb/advanced_options.h#L857

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -571,8 +571,8 @@ rocksdb.enable_blob_files no
 # which have an uncompressed size smaller than this threshold are stored alongside
 # the keys in SST files in the usual fashion.
 #
-# Default: 0 MiB, 0 means that all values are stored in blob files
-rocksdb.min_blob_size 0
+# Default: 4096 Bytes
+rocksdb.min_blob_size 4096
 
 # The size limit for blob files. When writing blob files, a new file is
 # opened once this limit is reached.
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -571,14 +571,14 @@ rocksdb.enable_blob_files no
 # which have an uncompressed size smaller than this threshold are stored alongside
 # the keys in SST files in the usual fashion.
 #
-# Default: 4096 byte, 0 means that all values are stored in blob files
-rocksdb.min_blob_size 4096
+# Default: 0 bytes, 0 means that all values are stored in blob files
+rocksdb.min_blob_size 0
 
 # The size limit for blob files. When writing blob files, a new file is
 # opened once this limit is reached.
 #
-# Default: 128 M
-rocksdb.blob_file_size 128
+# Default: 256 M
+rocksdb.blob_file_size 268435456
 
 # Enables garbage collection of blobs. Valid blobs residing in blob files
 # older than a cutoff get relocated to new files as they are encountered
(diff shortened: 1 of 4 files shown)


git-hulk (dismissed):

ShooterIT:
CC @caipengbo

ShooterIT:
I also think we should keep MB unit. and fix not multiplying `MB` when setting rocksdb config.

WyattJia:
@ShooterIT @caipengbo I've update unit to MiB, please check.

caipengbo (dismissed):

git-hulk (dismissed):

ShooterIT:
oh, I wanted to say `rocksdb.blob_file_size` uses `MB` unit, but we don't multiple `MiB` when setting. `min_blob_size` still should use `byte` since we need to control more precisely, and i think it is a good idea to set 4k by default for key value separately stored, IIRC, Tikv also use 4k.

kvrocks.conf
```diff
@@ -571,14 +571,14 @@ rocksdb.enable_blob_files no
 # which have an uncompressed size smaller than this threshold are stored alongside
 # the keys in SST files in the usual fashion.
 #
-# Default: 4096 byte, 0 means that all values are stored in blob files
+# Default: 4096 Bytes
```
ShooterIT: we shouldn't delete this comment
```suggestion
# Default: 4096 byte, 0 means that all values are stored in blob files
```

src/config.cc
```diff
@@ -367,14 +367,16 @@ void Config::initFieldCallback() {
         if (!RocksDB.enable_blob_files) {
           return Status(Status::NotOK, errNotEnableBlobDB);
         }
-        return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), v);
+        return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k),
+                                                    std::to_string(RocksDB.min_blob_size));
```
ShooterIT: don't need to change, right?
If not necessary, we should not dirty blame log which provide the clue to kvrocks evolution.

```suggestion
        return srv->storage_->SetColumnFamilyOption(trimRocksDBPrefix(k), v);
```

src/config.cc
```diff
@@ -141,8 +141,8 @@ Config::Config() {
       {"rocksdb.level0_file_num_compaction_trigger",
        false, new IntField(&RocksDB.level0_file_num_compaction_trigger, 4, 1, 1024)},
       {"rocksdb.enable_blob_files", false, new YesNoField(&RocksDB.enable_blob_files, false)},
-      {"rocksdb.min_blob_size", false, new IntField(&RocksDB.min_blob_size, 4096, 0, INT_MAX)},
-      {"rocksdb.blob_file_size", false, new IntField(&RocksDB.blob_file_size, 128, 0, INT_MAX)},
+      {"rocksdb.min_blob_size", false, new IntField(&RocksDB.min_blob_size, 0, 0, INT_MAX)},
```
ShooterIT: keep old configuration
```suggestion
      {"rocksdb.min_blob_size", false, new IntField(&RocksDB.min_blob_size, 4096, 0, INT_MAX)},
```

ShooterIT (approved):

git-hulk (approved):

PR #507, opened 2022-03-01
Avoid requirepass and masterauth conflicts with the namespace tokens

caipengbo:
If `requirepass` and `masterauth` are duplicated with namespace tokens, the `AUTH` command cannot be executed correctly. So we should avoid conflicts between them.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -524,6 +524,8 @@ Status Config::Load(const std::string &path) {
   }
 
   for (const auto &iter : fields_) {
+    // line_number = 0 means the user didn't specify the field value
+    // on config file and would use default value, so won't validate here.
     if (iter.second->line_number != 0 && iter.second->validate) {
       auto s = iter.second->validate(iter.first, iter.second->ToString());
       if (!s.IsOK()) {
--- a/src/config.cc
+++ b/src/config.cc
@@ -164,6 +164,15 @@ void Config::initFieldValidator() {
         if (v.empty() && !tokens.empty()) {
           return Status(Status::NotOK, "requirepass empty not allowed while the namespace exists");
         }
+        if (tokens.find(v) != tokens.end()) {
+          return Status(Status::NotOK, "requirepass is duplicated with namespace tokens");
+        }
+        return Status::OK();
+      }},
+      {"masterauth", [this](const std::string& k, const std::string& v)->Status {
+        if (tokens.find(v) != tokens.end()) {
+          return Status(Status::NotOK, "masterauth is duplicated with namespace tokens");
+        }
         return Status::OK();
       }},
       {"compact-cron", [this](const std::string& k, const std::string& v)->Status {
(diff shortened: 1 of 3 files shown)


src/config.cc
```diff
@@ -164,6 +164,15 @@ void Config::initFieldValidator() {
         if (v.empty() && !tokens.empty()) {
           return Status(Status::NotOK, "requirepass empty not allowed while the namespace exists");
         }
+        if (tokens.find(v) != tokens.end()) {
```
git-hulk, line 167: This validation should move to the callback since the validator would be executed before parsing the entire config file. So tokens array would be empty when requirepass was parsed before those token items.

caipengbo: Indeed, but I think this behavior is a validate behavior, and it's not elegant to put it in a callback.

Can we modify the `Config::parseConfigFromString()`, remove the inside of the validate behavior, And then validate in the `Config::Load`(just like the callback behavior).

git-hulk: Yes, it would be better, go ahead.

caipengbo: New PR or in this PR?

ShooterIT: you can do it in this PR since it is internal refactor instead of influencing user's usage.

src/config.cc
```diff
@@ -516,6 +525,17 @@ Status Config::Load(const std::string &path) {
     std::cout << "Warn: no config file specified, using the default config. "
                     "In order to specify a config file use kvrocks -c /path/to/kvrocks.conf" << std::endl;
   }
+
+  // Validate again when read all the config items from the file.
+  for (const auto &iter : fields_) {
```
git-hulk, line 526: I think it would confuse others why to validate again here. Also, this validation was needed when dynamic config the `masterauth` and `requirepass`.

After reconsidering this issue, maybe we can rename the `callback` function to postValidate, if you think the `callback` name was NOT suitable to add this validation.

caipengbo: I think `postValidate` is inferior to `callback`, so I'll put this check in `callback`

git-hulk: yeah, what i mean was we also need to validate it when use `config set`. So put it on the callback also makes sense for me.

caipengbo: The behavior of `CONFIG SET` is validate -> set -> callback. When `callback` is called, the configuration has been modified; If placed in `callback`, the configuration needs to be rollback.

git-hulk: hmm... Yes

caipengbo: So perhaps it would be better to save the line number when reading the configuration file and then validate it after reading all the configuration items?

git-hulk: Yes, it should be better.

src/config_type.h
```diff
@@ -29,6 +29,7 @@ class ConfigField {
   virtual Status ToBool(bool *b) { return Status(Status::NotOK, "not supported"); }
 
  public:
+  int line_number;
```
git-hulk:
```suggestion
  int line_number = 0;
```

src/config.cc
```diff
@@ -516,6 +522,17 @@ Status Config::Load(const std::string &path) {
     std::cout << "Warn: no config file specified, using the default config. "
                     "In order to specify a config file use kvrocks -c /path/to/kvrocks.conf" << std::endl;
   }
+
+  for (const auto &iter : fields_) {
+    if (iter.second->line_number != 0 && iter.second->validate) {
```
git-hulk, line 529:
```suggestion
    // line_number = 0 means the user didn't specify the field value
    // on config file and would use default value, so won't validate here.
    if (iter.second->line_number != 0 && iter.second->validate) {
```

git-hulk (approved):
LGTM

PR #514, opened 2022-03-06
Support auth before migrating slot

### LINKED ISSUE #509: [BUG]  kvrocks migrate does not support pasword
**Describe the bug**
When kvrocks migrate, if the cluster has a password, migrate fails。

 - OS: [e.g. Centos 7]
 - Version [2.0.6]

**To Reproduce**
1. 
451646291138_ pic

2.  redis-cli -h $ip -p $port -a $password clusterx migrate $slot $dst_nodeid

**Expected behavior**
migrate should successfully!

441646291115_ pic

ChrisZMF:
Closes #509. If `reqirepass` is set, auth the destination server by the setting password before migrating slot. It is based on the following assumption.
- Password should be set for all nodes in the cluster at the same time
- All nodes have same password

### DIFF
--- a/src/slot_migrate.cc
+++ b/src/slot_migrate.cc
@@ -217,6 +217,15 @@ Status SlotMigrate::Start(void) {
     return Status(Status::NotOK);
   }
 
+  // Auth first
+  std::string pass = svr_->GetConfig()->requirepass;
+  if (!pass.empty()) {
+    bool st = AuthDstServer(slot_job_->slot_fd_, pass);
+    if (!st) {
+      return Status(Status::NotOK, "Failed to auth destination server");
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


ShooterIT (approved):
LGTM, but don't read carefully 😜

git-hulk (approved):
LGTM, for `CheckResponseOnce`, I think we should return status instead of boolean since we need to propagate the status to callers. But it should not refactor in this PR, just for reminding.

ChrisZMF:
> LGTM, for `CheckResponseOnce`, I think we should return status instead of boolean since we need to propagate the status to callers. But it should not refactor in this PR, just for reminding.

Got it, it should be refactored in the future PR.

PR #515, opened 2022-03-08
fix: return bulk string (instead of array) when spop without count.

patpatbear:
kvrocks spop reply array.
redis spop reply:
- bulk string(or nil) when spop without count
- array when spop with count

this patch fix kvrocks spop reply without count.


ShooterIT (approved):

git-hulk (approved):

git-hulk:
@patpatbear Thanks for your contribution, cheers!

PR #517, opened 2022-03-10
Support unlink command

ColinChamber:
This is to avoid errors when users use `redis unlink`. It's a bit tricky, but RocksDB also frees space asynchronously.


git-hulk (approved):
LGTM,cool.

ShooterIT (approved):
i also wanted to support it, but i don't know why i forgot 😂

PR #520, opened 2022-03-13
Fix slave can't resync with master after password change

### LINKED ISSUE #513: [BUG] Slave can't re-sync with master again after changing password
**Describe the bug**
If the connection between master and slave is broken (reasons may be network error, restarting master... ) after changing the password of master, the slave can't resynchronize with master again automatically. The resynchronization will fail because slave can't auth master successfully though the `masterauth` of slave is set rightly. It can be resolved by restarting slave with new `masterauth` or execute `slaveof no one` to reset the master-slave relationship.
image

 - OS: all
 - Version: all

**To Reproduce**
1. Sync server B(slave) with server A(master)
2. Set or change `requirepass` of A and `masterauth` of B at the same time
3. Restart master with new password, then slave can't resync with master.

**Expected behavior**
The B(slave) can re-sync with A(master) automatically.

**Reason Analysis**
When slave sync with master, a `ReplicationThread` object will be created and the password of master will be assigned to its member variable `auth_`.  Although the value of `masterauth` is re-assigned, `auth_` will still hold the old value because it can't be re-assigned dynamically.

ChrisZMF:
Closes #513

### DIFF
--- a/src/config_type.h
+++ b/src/config_type.h
@@ -153,7 +153,7 @@ class EnumField : public ConfigField {
   Status Set(const std::string &v) override {
     int e = configEnumGetValue(enums_, v.c_str());
     if (e == INT_MIN) {
-      return Status(Status::NotOK, "invaild enum option");
+      return Status(Status::NotOK, "invalid enum option");
     }
     *receiver_ = e;
     return Status::OK();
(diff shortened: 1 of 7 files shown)


git-hulk (approved):
Good catch!

ShooterIT (approved):

PR #523, opened 2022-03-18
Optimize expiration to make sure eventual consistency between replicas

### LINKED ISSUE #522: [BUG] The behavior that slave can delete expired key in compactionfilter may lead to Inconsistency with master forever
Pika has the same problem.

shangxiaoxiong:
- Only master is allowed to delete expired meta key (except STRING) in compaction
- Master spreads expire-deletion operation to slaves
- Expire-deletion speed can be limited by max-expire-delete-mb
- Subkeyfilter can be more efficient, subkey will be simply dropped instead of output until the bottom level. Refer to FilterV2 API https://github.com/facebook/rocksdb/wiki/Compaction-Filter

Fix https://github.com/KvrocksLabs/kvrocks/issues/522

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -198,6 +198,12 @@ max-replication-mb 0
 # Default: 500
 max-io-mb 500
 
+# The maximum allowed rate (in MB/s) that should be used by MetadataFilter for deleting expired complex key.
+# If the rate exceeds max-expire-delete-io-mb, deletion of expired key will slow down.
+# 0 is no limit
+# Default: 5
+max-expire-delete-io-mb 5
+
 # The maximum allowed space (in GB) that should be used by RocksDB.
 # If the total size of the SST files exceeds max_allowed_space, writes to RocksDB will fail.
 # Please see: https://github.com/facebook/rocksdb/wiki/Managing-Disk-Space-Utilization
--- a/src/compact_filter.cc
+++ b/src/compact_filter.cc
@@ -23,11 +23,56 @@ bool MetadataFilter::Filter(int level,
                  << ", err: " << s.ToString();
     return false;
   }
+
+  bool is_filter = metadata.Expired();
+
+  if (is_filter && metadata.Type() != kRedisString) {
+    is_filter = stor_->IsDBNoExpire() ? false : IfRemoveExpiredRecord(key, metadata.Type());
+  }
+
   DLOG(INFO) << "[compact_filter/metadata] "
              << "namespace: " << ns
              << ", key: " << user_key
-             << ", result: " << (metadata.Expired() ? "deleted" : "reserved");
-  return metadata.Expired();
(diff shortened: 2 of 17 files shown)


git-hulk:
Cool, can u help to add more information about the design of this fix.

shangxiaoxiong:
On the way.

shangxiaoxiong:
### How Kvrocks expire keys
Keys are expired in only way: compaction.
RocksDB provides a way to delete or modify key/value pairs based on custom logic in background by using custom compaction filter. Kvrocks implement MetadataFilter and SubkeyFilter to delete expired metadata and subkey in compaction.

**MetadataFilter**:  input record considered expired by judging timestamp in value will be marked deleted
**SubkeyFilter**:  input record are marked deleted under four conditions. 1) meta key is not found 2)  meta key is expired 3) type unmatch with meta key 4) version unmatch with meta key

### How expires are handled in the replicas
 Unlike redis, the replicas expire keys independently.

### Challenges and opportunities in existing mechanisms
**Problem1**: The expration operation of subkey is not atomic so that some subkey may be lost while clients change expiration time during the period of compaction.

**Problem2**: The design that replicas can expire keys independently by using compaction fliter is concise, But consistency is sacrificed. Compared with master, slaves may lose some keys forever.

Above all, master may lose data and eventual consistency can not be guaranted between replicas.

Problem1 will be solved if SubkeyFilter is not dependent upon expiration time of meta key.

For problem2, what is expected ? In order to obtain a correct behavior without sacrificing consistency, when a key expires, a DEL operation is synthesized in both the WAL(AOF in Redis) file and gains all the attached replicas nodes. This way the expiration process is centralized in the master instance, and there is no chance of consistency errors. That is what Redis did.

### Getting it to work
For problem1, a slight modification to SubkeyFilter is necessary. 
**new SubkeyFilter**: input record are marked deleted under three conditions. 1) meta key is not found 2) type unmatch with meta key 3) version unmatch with meta key

For problem2,  critical modification of MetadataFilter is that, when timestamp of input record shows it is timed out, lookup the latest value of the key from DB and append a delete operation to DB if the latest value shows the key is timed out. To prevent expiration time is modified during query, the key should be locked before query and be unlocked after appending deletion.

### Make it better
Subkeyfilter can be more efficient, subkey under three condition above will be simply dropped instead of output until the bottom level. It also reduces disk write. Refer to FilterV2 API https://github.com/facebook/rocksdb/wiki/Compaction-Filter.

Otherwise, expire-deletion speed can be limited by max-expire-delete-mb. 

@git-hulk

git-hulk:
@shangxiaoxiong Cool, thanks for your detail explain.

tisonkun:
Closed as stale.

Codebase evolves a lot since the last activities on this patch. There're a few conflicts. @shangxiaoxiong you're welcome to reopen and fix the conflicts or propose the changes again based on the nightly unstable branch.

Issue #510, opened 2022-03-04
[QUESTION]  Does the kvrocks migrate command and setslot command support batch slot operations?


tianshimoyi:
Are there any plans for the kvrocks migrate command and setslot to develop functionality to support batch slot migration？setslot needs to reset the version for each slot, which will make the version more and more difficult to control in the future.

ShooterIT:
Yes, we have the plan but now we don't have much spare time, may can't support it soon. And current implementation is generally available. Of course, PR is welcome 😜

ChrisZMF:
I will support batch slot operation for command setslot first.

tisonkun:
Closed as stale. You can ask questions at the [Discussions forum](https://github.com/apache/incubator-kvrocks/discussions/categories/q-a).

PR #529, opened 2022-03-24
Support slot batch operation for CLUSTERX SETSLOT

ChrisZMF:
Support for #510 

It is an enhancement to the original implement of command `CLUSTERX SETSLOT`  #463 

**Command formart**

`CLUSTERX SETSLOT "$slot_batch" NODE $node_id $new_version`

assign batch slots to the node if new version is current version+1. 
It is compatible with the original command syntax.

**Parameter Changes**

> $slot_batch: batch slots which we want to assign. It can be given by single slotid or slot batch.  Slot batch can be given by continuous slot range or discrete slot list with spaces as separators, e.g. "3 5 9 10-20".

**Example**
`CLUSTERX SETSLOT 1 NODE $nodeid1 1`
`CLUSTERX SETSLOT "3 5 9 20-30" NODE $nodeid1 2`

### DIFF
--- a/src/cluster.cc
+++ b/src/cluster.cc
@@ -57,9 +57,8 @@ Status Cluster::SetNodeId(std::string node_id) {
   return Status::OK();
 }
 
-// Set the slot to the node if new version is current version +1. It is useful
-// when we scale cluster avoid too many big messages, since we only update one
-// slot distribution and there are 16384 slot in our design.
+// Set slots to the node if new version is current version +1. It is useful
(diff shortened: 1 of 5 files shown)


tisonkun:
@ChrisZMF you may separate the fix typo part into a dedicated patch and resolve the conflict so that we can continue reviewing this patch :)

tisonkun:
Closed as inactive. You may resubmit the patch base on the latest unstable branch.

Issue #530, opened 2022-03-27
[QUESTION] Change the license to Apache License Version 2.0


ShooterIT:
Hi, all Kvrocks contributors, we want to donate Kvrocks to Apache Software Foundation, the first step for us is changing the license to Apache License Version 2.0. As we know, Kvrocks is an open source software under the BSD-3-Clause license, we truly thank every contributor and user, you guys make Kvrocks great and have an active community, before we could submit PR to change license, we must to get your approval.

@git-hulk @karelrooted @ShooterIT  @Alfejik @caipengbo @ChrisZMF @shangxiaoxiong @ColinChamber @guoxiangCN @calvinxiao @patpatbear @WyattJia @smartlee @popunit @clyang @byronhe @QQxiaoyuyu @nioshield @singgel @liangjingyang @gaoyuanzong @rongzeng54 @unext-ops @fishery86 @xujianhai666 @LykxSassinator 

Please reply "Approved" if you agree.

ShooterIT:
Approved

caipengbo:
Approved

WyattJia:
Approved🎉

git-hulk:
Approved

shangxiaoxiong:
Approved

karelrooted:
Approved

ChrisZMF:
Approved

Alfejik:
Approved

xujianhai666:
Approved

ColinChamber:
Approved

LykxSassinator:
Approved！
LGTM！！！

patpatbear:
Approved

liangjingyang:
Approved

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

unext-ops:
Approved

calvinxiao:
Approved

nioshield:
Approved🎉🎉🎉

singgel:
Approved

mzygQAQ:
Approved

smartlee:
Approved

QQxiaoyuyu:
Good ,Approved

popunit:
Approved

rongzeng54:
Approved

clyang:
Approved

byronhe:
Approved

fishery86:
Approved

gaoyuanzong:
Approved

[...]

PR #531, opened 2022-03-30
Add support for listening to unix socket

### LINKED ISSUE #148: Please support unix domain socket
As the title

torwig:
Two (or more) threads/processes cannot bind (and listen) to the same Unix socket: 
https://unix.stackexchange.com/questions/615330/what-happens-when-two-processes-listen-on-the-same-berkeley-unix-file-socket.
That's why only one Worker binds the socket.
Introduce a separate ConfigField and a function to parse Unix socket permissions from the config file. This number is defined
in octal form (e.g. 770 or 0777).

### DIFF
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -2,8 +2,11 @@
 
 #include <sys/types.h>
 #include <sys/socket.h>
+#include <sys/stat.h>
+#include <sys/un.h>
 #include <list>
 #include <cctype>
+#include <cstring>
 #include <utility>
 #include <algorithm>
 #include <glog/logging.h>
@@ -24,7 +27,7 @@ Worker::Worker(Server *svr, Config *config, bool repl) : svr_(svr) {
   int port = config->port;
   auto binds = config->binds;
   for (const auto &bind : binds) {
-    Status s = listen(bind, port, config->backlog);
+    Status s = listenTCP(bind, port, config->backlog);
     if (!s.IsOK()) {
       LOG(ERROR) << "[worker] Failed to listen on: "<< bind << ":" << port
                  << ", encounter error: " << s.Msg();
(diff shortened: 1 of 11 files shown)


git-hulk:
Cool, thanks for your great contribution!

src/worker.cc
```diff
@@ -86,23 +89,49 @@ void Worker::newConnection(evconnlistener *listener, evutil_socket_t fd,
                     Redis::Connection::OnEvent, conn);
   bufferevent_enable(bev, EV_READ);
   Status status = worker->AddConnection(conn);
+  if (!status.IsOK()) {
+    std::string err_msg = Redis::Error("ERR " + status.Msg());
+    write(fd, err_msg.data(), err_msg.size());
+    conn->Close();
```
git-hulk, line 95: Need to return here after closing the connection

torwig: Should I do it in a separate commit or I can change this commit + make a force push to branch?

git-hulk: It's fine to add a new commit to fix it. 👍

torwig: Done. And in line 128 too.

git-hulk (approved):
LGTM

git-hulk:
@torwig Thanks for your great contribution again, we are planning to donate Kvrocks to ASF recently, so this PR would be pending before that. We also had filed the issue to let every contributor know and agree. For more detail can see: #530

ShooterIT:
Thanks @torwig LGTM.
Are you familiar with TCL? please add test cases if yes, if no, i will do that in the next PR.

torwig:
Hi @ShooterIT , I'm not familiar with TCL right now.

ShooterIT (approved):

ShooterIT:
@torwig fine, i merged, cheers!

PR #538, opened 2022-04-22
PSYNC based on Unique Replication Sequence ID

### LINKED ISSUE #462: [BUG] Incorrect data after SLAVEOF
**Describe the bug**
After SLAVEOF, 2 instances can contain different values for a given key.

**To Reproduce**
Start 2 kvrocks instances and execute the following commands in the respective instances.

- **SERVER2**: SLAVEOF SERVER1
- **SERVER2**: SLAVEOF NO ONE
- **SERVER2**: SET test 222
- **SERVER1**: SET test 111
- **SERVER2**: SLAVEOF SERVER1
- **SERVER2**: GET test
  - Returns "222". However, it should return "111".

If a new SET command is used on SERVER1, SERVER2 immediately picks up the latest value.
This issue doesn't occur if, on the "test" key, SERVER1 executed multiple SET commands while SERVER2 was not a slave.

**Expected behavior**
After SLAVEOF command, SERVER2 should always contain identical data to SERVER1.

ShooterIT:
### Background
Currently, master only checks sequence number when replica asks for PSYNC, that is not enough since they may have different replication history even the replica asking sequence is in the range of the master current WAL.

Bug report in https://github.com/KvrocksLabs/kvrocks/issues/462

PS: Master check also DB name before checking sequence number, but it is also not enough since it is setting in conf, and don't change when role is changed.
 
### Solution

We design 'PSYNC based on Unique Replication Sequence ID', we add unique replication id for every write batch (the operation of each command on the storage engine), so the combination of replication id and sequence is unique for write batch. The master can identify whether the replica has the same replication history by checking replication id and sequence. 
Like Redis, replicas can partially resynchronize with new master if old master is failed and new selected master has largest offset, and replicas can also partially resynchronize with master after restarting if the replicas' conf file has the its origin master.

Unique replication id will be changed when replicas become master.

By default, it is not enabled since this stricter check may easily lead to full synchronization. You should enable it if you want to data correctness, maybe we enable it by default in kvrocks 3.0.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -325,9 +325,6 @@ int main(int argc, char* argv[]) {
     }
   };
 
-  // Generate new replication id if not a replica
-  if (!srv->IsSlave()) storage.ShiftReplId();
-
   s = srv->Start();
   if (!s.IsOK()) {
     removePidFile(config.pidfile);
--- a/src/main.cc
+++ b/src/main.cc
@@ -324,7 +324,6 @@ int main(int argc, char* argv[]) {
       srv->Stop();
     }
   };
-
   s = srv->Start();
   if (!s.IsOK()) {
     removePidFile(config.pidfile);
--- a/src/main.cc
+++ b/src/main.cc
@@ -324,6 +324,10 @@ int main(int argc, char* argv[]) {
       srv->Stop();
     }
   };
+
+  // Generate new replication id if not a replica
+  if (!srv->IsSlave()) storage.ShiftReplId();
+
   s = srv->Start();
   if (!s.IsOK()) {
     removePidFile(config.pidfile);
--- a/src/replication.cc
+++ b/src/replication.cc
@@ -187,6 +187,8 @@ LOOP_LABEL:
   switch (st) {
     case CBState::NEXT:
       ++self->handler_idx_;
+    case CBState::PREV:
+      if (st == CBState::PREV) --self->handler_idx_;
       if (self->getHandlerEventType(self->handler_idx_) == WRITE) {
         SetWriteCB(bev, EvCallback, ctx);
       } else {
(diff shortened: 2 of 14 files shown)


src/server.cc
```diff
@@ -1422,3 +1412,24 @@ void Server::AdjustOpenFilesLimit() {
...
+Status ServerLogData::Decode(const rocksdb::Slice &blob) {
+  if (blob.size() == 0) {
+    return Status(Status::NotOK);
+  }
+
+  const char *header = blob.data();
+  if (*header == kReplIdTag) {
+    type_ = kReplIdLog;
+  }
+  content_ = std::string(blob.data()+2, blob.size()-2);
```
git-hulk: should check the blob data length before using it?

src/replication.cc
```diff
@@ -466,10 +468,39 @@ ReplicationThread::CBState ReplicationThread::replConfReadCB(
...
+  // Get replication id
+  std::string replid_in_wal = self->storage_->GetReplIdFromWalBySeq(cur_seq);
+  // Set if valid replication id
+  if (replid_in_wal.length() == kReplIdLength) {
+    replid = replid_in_wal;
+  } else {
+    // Maybe there is no WAL, we can get replication id from db since master
+    // always write replication id into db before any operation when starting
+    // new "replication history".
+    std::string replid_in_db = self->storage_->GetReplIdFromDbEngine();
```
git-hulk, line 484: The slave also can't get replid from DB as well if there's no WAL(empty slave db), so propagating the replication id would never be used?

src/main.cc
```diff
@@ -324,6 +324,10 @@ int main(int argc, char* argv[]) {
       srv->Stop();
     }
   };
+
+  // Generate new replication id if not a replica
+  if (!srv->IsSlave()) storage.ShiftReplId();
+
```
git-hulk: Master was added after starting the server, so it should always be true here.

ShooterIT: you are right, i missed this, we set master info in `Server::Start ` instead of loading config file.

src/replication.cc
ShooterIT: Maybe empty WAL, as we know, wal may be not persistent in disk after restarting server.

src/server.cc
ShooterIT: ok, it is more strict

src/replication.cc
ShooterIT: And one checkpoint from master, WAL may be flushed when creating checkpoint

git-hulk: Cool, got it.

git-hulk: Cool, got it.

git-hulk:
@ShooterIT Need to add the License Header for new files

git-hulk (dismissed):
LGTM

git-hulk (approved):

PR #539, opened 2022-04-22
Stop sending files if failed in full synchronization

ShooterIT:
Before this commit, master may still send file even if failed in full synchronization. It is not necessary, and cause next full synchronization failed until all files totally are sent.

Thanks for report from @ChrisZMF

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -4274,10 +4274,12 @@ class CommandFetchFile : public Commander {
             Util::SockSendFile(repl_fd, fd, file_size).IsOK()) {
           LOG(INFO) << "[replication] Succeed sending file " << file << " to "
                     << ip;
-        } else {
-          LOG(WARNING) << "[replication] Fail to send file " << file << " to "
-                       << ip << ", error: " << strerror(errno);
+          close(fd);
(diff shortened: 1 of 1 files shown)


src/redis_cmd.cc
```diff
@@ -4274,10 +4274,12 @@ class CommandFetchFile : public Commander {
             Util::SockSendFile(repl_fd, fd, file_size).IsOK()) {
           LOG(INFO) << "[replication] Succeed sending file " << file << " to "
                     << ip;
-        } else {
-          LOG(WARNING) << "[replication] Fail to send file " << file << " to "
-                       << ip << ", error: " << strerror(errno);
+          close(fd);
+          break;
```
git-hulk, line 4281: We should break the loop when sending the file with an error? or next files will not be sent here.

ShooterIT: yes

git-hulk: So we should NOT break the loop since this condition means sending file was ok.

ShooterIT: my bad 🤦

git-hulk (approved):

PR #554, opened 2022-05-08
Support to listen on ipv6 address

### LINKED ISSUE #491: [BUG] It doesn't work with ipv6?
I'm trying to make kvrocks works with an ipv6 address on https://fly.io

**To Reproduce**
Here is my Dockerfile :
```dockerfile
FROM kvrocks/kvrocks:v2.0.6

RUN sed -i "s|bind 0.0.0.0|bind ::|g" ./conf/kvrocks.conf
```

```bash
# lsof -i -P -n | grep LISTEN
kvrocks  523 root   16u  IPv4   4450      0t0  TCP *:6666 (LISTEN)
kvrocks  523 root   21u  IPv4   4452      0t0  TCP *:6666 (LISTEN)
kvrocks  523 root   26u  IPv4   4454      0t0  TCP *:6666 (LISTEN)
kvrocks  523 root   31u  IPv4   4456      0t0  TCP *:6666 (LISTEN)
kvrocks  523 root   36u  IPv4   4458      0t0  TCP *:6666 (LISTEN)
kvrocks  523 root   41u  IPv4   4460      0t0  TCP *:6666 (LISTEN)
kvrocks  523 root   46u  IPv4   4462      0t0  TCP *:6666 (LISTEN)
kvrocks  523 root   51u  IPv4   4464      0t0  TCP *:6666 (LISTEN)
```

**Expected behavior**
Kvrocks should listen on `IPv6 TCP *:6666`

git-hulk:
I didn't add test cases for this PR since the ipv6 was supported in TCL8.6, and the default TCL version on MacOSX was 8.5. So would force developers to upgrade TCL version manually if we want to add the ipv6 test cases, I think it's not worth to do that. I have tested the client and replication on my side. 

Closes #491

### DIFF
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -48,11 +48,7 @@ Worker::Worker(Server *svr, Config *config, bool repl) : svr_(svr) {
   int port = config->port;
   auto binds = config->binds;
   for (const auto &bind : binds) {
-    if (strchr(bind.data(), ':')) {
-      s = listenTCP(bind, port, AF_INET6, config->backlog);
-    } else {
-      s = listenTCP(bind, port, AF_INET, config->backlog);
-    }
+    s = listenTCP(bind, port, config->backlog);
     if (!s.IsOK()) {
       LOG(ERROR) << "[worker] Failed to listen on: "<< bind << ":" << port
                  << ", encounter error: " << s.Msg();
@@ -158,10 +154,16 @@ void Worker::newUnixSocketConnection(evconnlistener *listener, evutil_socket_t f
   }
 }
 
-Status Worker::listenTCP(const std::string &host, int port, int af, int backlog) {
+Status Worker::listenTCP(const std::string &host, int port, int backlog) {
   char _port[6];
+  int af;
   int rv, fd, sock_opt = 1;
 
+  if (strchr(host.data(), ':')) {
+    af = AF_INET6;
+  } else {
+    af = AF_INET;
+  }
   snprintf(_port, sizeof(_port), "%d", port);
   struct addrinfo hints, *srv_info, *p;
   memset(&hints, 0, sizeof(hints));
--- a/src/worker.cc
+++ b/src/worker.cc
@@ -156,8 +156,7 @@ void Worker::newUnixSocketConnection(evconnlistener *listener, evutil_socket_t f
 
 Status Worker::listenTCP(const std::string &host, int port, int backlog) {
   char _port[6];
-  int af;
-  int rv, fd, sock_opt = 1;
+  int af, rv, fd, sock_opt = 1;
 
   if (strchr(host.data(), ':')) {
     af = AF_INET6;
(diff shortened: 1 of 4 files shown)


src/worker.cc
```diff
@@ -153,27 +158,42 @@ void Worker::newUnixSocketConnection(evconnlistener *listener, evutil_socket_t f
...
-  sin.sin_port = htons(port);
-  int fd = socket(AF_INET, SOCK_STREAM, 0);
-  int sock_opt = 1;
-  if (setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &sock_opt, sizeof(sock_opt)) < 0) {
-    return Status(Status::NotOK, evutil_socket_error_to_string(EVUTIL_SOCKET_ERROR()));
-  }
-  // to support multi-thread binding on macOS
-  if (setsockopt(fd, SOL_SOCKET, SO_REUSEPORT, &sock_opt, sizeof(sock_opt)) < 0) {
-    return Status(Status::NotOK, evutil_socket_error_to_string(EVUTIL_SOCKET_ERROR()));
+Status Worker::listenTCP(const std::string &host, int port, int af, int backlog) {
```
ShooterIT: i don't think `af` should be a argument. you can judge host inner to check ipv4 or ipv6. so when calling, we don't need two statements

git-hulk: you mean move the host check inside listenTCP? If yes, I'm ok to do that.

ShooterIT: yes, i think it will be more generic.

git-hulk: OK, nice advice.

src/worker.cc
```diff
@@ -154,26 +155,47 @@ void Worker::newUnixSocketConnection(evconnlistener *listener, evutil_socket_t f
...
-  int sock_opt = 1;
-  if (setsockopt(fd, SOL_SOCKET, SO_REUSEADDR, &sock_opt, sizeof(sock_opt)) < 0) {
-    return Status(Status::NotOK, evutil_socket_error_to_string(EVUTIL_SOCKET_ERROR()));
-  }
-  // to support multi-thread binding on macOS
-  if (setsockopt(fd, SOL_SOCKET, SO_REUSEPORT, &sock_opt, sizeof(sock_opt)) < 0) {
-    return Status(Status::NotOK, evutil_socket_error_to_string(EVUTIL_SOCKET_ERROR()));
+  char _port[6];
+  int af;
+  int rv, fd, sock_opt = 1;
```
ShooterIT:
```suggestion
  int rv, fd, af, sock_opt = 1;
```

src/worker.cc
```diff
@@ -154,26 +155,47 @@ void Worker::newUnixSocketConnection(evconnlistener *listener, evutil_socket_t f
...
+    af = AF_INET;
+  }
+  snprintf(_port, sizeof(_port), "%d", port);
+  struct addrinfo hints, *srv_info, *p;
+  memset(&hints, 0, sizeof(hints));
+  hints.ai_family = af;
+  hints.ai_socktype = SOCK_STREAM;
+  hints.ai_flags = AI_PASSIVE;
+
+  if ((rv = getaddrinfo(host.data(), _port, &hints, &srv_info)) != 0) {
```
ShooterIT, line 173: forget to call `freeaddrinfo` to free `addrinfo`

ShooterIT (commented):
yes, we should use `getaddrinfo`

src/worker.cc
git-hulk: good catch, will fix later.

ShooterIT (approved):

PR #564, opened 2022-05-09
Use FetchContent instead of ExternalProject and git submodules in CMake

PragmaTwice:
Check https://github.com/apache/incubator-kvrocks/issues/562 and  https://github.com/apache/incubator-kvrocks/issues/561 for more details.

@tisonkun


PragmaTwice:
After this PR, the cmake build process will not depend on any files in the `externel` directory. However, in order to make Makefiles still work, I will restore the `externel` directory (which is currently deleted in this pr).

git-hulk:
Thanks @PragmaTwice. I think we can do that first if it's too complex.

cmake/utils.cmake
```diff
@@ -0,0 +1,26 @@
+include_guard()
```
tisonkun, line 18: You should add a license head as other files to pass the checker.

tisonkun: I'll do you a favor :)

cmake/rocksdb.cmake
```diff
@@ -15,67 +15,34 @@
...
-  set(rocksdb_LIBRARIES ${rocksdb_INSTALL}/${CMAKE_INSTALL_LIBDIR}/librocksdb.a)
-endif()
+FetchContent_Declare(rocksdb
+  GIT_REPOSITORY https://github.com/facebook/rocksdb
+  GIT_TAG v6.29.5
+)
+
+include(cmake/utils.cmake)
+
+FetchContent_MakeAvailableWithArgs(rocksdb
```
git-hulk, line 38: rocksdb depends on jemalloc and snappy, so we should also do that like the old CMake file.

CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.16)
```
git-hulk: Do we use any instructions that required 3.16? I saw fetch content instruction required 3.11+.

cmake/rocksdb.cmake
PragmaTwice: `FetchContent` commands will be executed on configuration stage, not the build stage, so there is no `DEPENDS` parameter (and the FetchContent declared names are not cmake targets actually, so no `add_dependencies`). The execution order plays the role of `DEPENDS`, you can check [here](https://cmake.org/cmake/help/latest/module/FetchContent.html#examples) to find details.
I will figure out why the CI failed :)

CMakeLists.txt
PragmaTwice: Yeah, use 3.16 just because it is the default version in apt repo of ubuntu 20.04. But I highly recommend a newer version of cmake. [A good post](https://www.scivision.dev/cmake-changelog) described its advantages.

git-hulk: OK, thanks for your explain.

cmake/rocksdb.cmake
git-hulk: We specified the include and library path in jemalloc cmake
and referenced those paths in rocksdb cmake when enabling the jemalloc.
That's what I commented here.

tisonkun:
@PragmaTwice please fix CI failure and ping me when this PR ready. I'll do a manual test on macos env.

tisonkun:
@git-hulk it seems that `lint-build-test-on-ubuntu` fails on test phase, not on build phase, please take a look whether it's about unstable test or stateful test.

@PragmaTwice it seems `build-on-ubuntu-with-cmake` complains with jemalloc not found.

git-hulk:
> @git-hulk it seems that lint-build-test-on-ubuntu fails on test phase, not on build phase, please take a look whether it's about unstable test or stateful test.

OK, I had a glance at the failure test case. The root cause should be that case depends on
time sleep, would improve it later.

cmake/utils.cmake
PragmaTwice: thx!

git-hulk:
@PragmaTwice Last failure caused by didn't link `-pthread`. Can have a test on local with running `sh build.sh _build`, so you are NOT need to wait for the CI feedback.

PragmaTwice:
> @PragmaTwice Last failure caused by didn't link `-pthread`. Can have a test on local with running `sh build.sh _build`, so you are NOT need to wait for the CI feedback.

Sorry, fixed. I didn't encounter this problem when I tested the build locally, probably due to some differences between CI and my local environment.

PragmaTwice:
> @PragmaTwice please fix CI failure and ping me when this PR ready. I'll do a manual test on macos env.

@tisonkun CI has been fixed now.

tisonkun:
@PragmaTwice so, how can I build kvrocks with cmake now? You may add a section on README also.

tisonkun:
I try to run with:

```bash
mkdir cmake-build-debug ; cd cmake-build-debug
cmake -DCMAKE_BUILD_TYPE=Release ..
```

which failed and complained:

```
CMake Error at /usr/local/lib/cmake/GTest/GTestTargets.cmake:37 (message):

  /usr/local/lib/cmake/GTest/GTestConfig.cmake:32 (include)
  /usr/local/Cellar/cmake/3.23.1/share/cmake/Modules/FindGTest.cmake:194 (find_package)
  cmake-build-debug/_deps/snappy-src/CMakeLists.txt:40 (find_package)
```

It seems that current script make conflict with system-wise library?

PragmaTwice:
> It seems that current script make conflict with system-wise library?

Snappy use `find_package(gtest)` mechanism to locate gtest libs (in https://github.com/google/snappy/blob/b02bfa754ebf27921d8da3bd2517eab445b84ff9/CMakeLists.txt#L40).
The newer version of Snappy do not use `find_package` any more (it use `add_subdirectory` and submodules now), but for the current version, we probably should add `FindGTest.cmake` to tell Snappy we have already build our own gtest.
BTW, we actually defined `SNAPPY_BUILD_TESTS=OFF`, so gtest is useless for Snappy (but in the current Snappy version we depend on, they just do not check the option before `find_package`, haha).

PragmaTwice:
I feel that updating snappy from 1.1.7 to 1.1.9 is another good solution to this problem, which removes the need for us to add `FindGTest.cmake`. What do you think is better to do? @git-hulk

git-hulk:
> I feel that updating snappy from 1.1.7 to 1.1.9 is another good solution to this problem, which removes the need for us to add `FindGTest.cmake`. What do you think is better to do? @git-hulk

it's good to me

tisonkun (dismissed):
LGTM. Also manually test on macosx.

Tests failure looks unrelated with this change.

We can later remove submodules and improve build efficiency.

ShooterIT:
I am not familiar with this topic, but it seems a better way.

BTW, i ever wanted to remove Makefile, only keep cmake, since we need to install some libs firstly if we use make.
@git-hulk told me we need a high version cmake which may be not convenient for users. I don't know 3.16 is high? If not, maybe in the future, we can only support cmake.

git-hulk:
> I am not familiar with this topic, but it seems a better way.
> 
> BTW, i ever wanted to remove Makefile, only keep cmake, since we need to install some libs firstly if we use make. @git-hulk told me we need a high version cmake which may be not convenient for users. I don't know 3.16 is high? If not, maybe in the future, we can only support cmake.

Yes, especially on centos, but I'm agree to remove the Makefile since it will cause much extra
work to maintain, as well as can't compatible with FetchContent.

tisonkun:
@git-hulk @ShooterIT as stated in https://github.com/apache/incubator-kvrocks/pull/564#issuecomment-1121972637, we may proceed this patch to introduce a `FetchContent` based cmake build system. And remove Makefile build system in a separated PR.

cc @PragmaTwice do I get it right?

git-hulk:
Overall looks good, will take a new pass again and have a try on my side tonight.

PragmaTwice:
> @git-hulk @ShooterIT as stated in [#564 (comment)](https://github.com/apache/incubator-kvrocks/pull/564#issuecomment-1121972637), we may proceed this patch to introduce a `FetchContent` based cmake build system. And remove Makefile build system in a separated PR.
> 
> cc @PragmaTwice do I get it right?

Yeah, I also think it would be good to remove makefiles after this PR.

> BTW, i ever wanted to remove Makefile, only keep cmake, since we need to install some libs firstly if we use make. @git-hulk told me we need a high version cmake which may be not convenient for users. I don't know 3.16 is high? If not, maybe in the future, we can only support cmake.

You can easily get cmake binaries from https://github.com/Kitware/CMake/releases (I do not try it in an old distro environment like centos or debian stable, but I do know some glibc linking issue will appear while the glibc version in build environment is higher than it in user's environment). I think the most compatible way to use cmake is to build it by your own (as in `build.sh`) or use a container.

PragmaTwice:
I can lower the required cmake version to 3.11 (if it is confirmed that this version contains all the features used), if that helps with compatibility.

ShooterIT:
> Yeah, I also think it would be good to remove makefiles after this PR.

OK, thanks

> I can lower the required cmake version to 3.11 (if it is confirmed that this version contains all the features used), if that helps with compatibility.

If cmake version 3.11 can work well, i vote 3.11

tisonkun:
@ShooterIT for choosing a newer version, @PragmaTwice explain in this https://github.com/apache/incubator-kvrocks/pull/564#discussion_r869138940.

CMakeLists.txt
ShooterIT: @PragmaTwice what features we use for 3.16?

ShooterIT: i think 20.04 also is newer, we still use ubuntu 18.04 in CI. For some enterprise user, they even still use CentOS 6.

ShooterIT:
CMake 3.16.0 is released at Nov 26, 2019
CMake 3.11.0 is released at Dec 01, 2018

no much difference 😂

ShooterIT (dismissed):

PragmaTwice:
I just tested the build process in cmake 3.11, and an error occurred because a feature [here](https://cmake.org/cmake/help/latest/command/install.html) in cmake 3.13 is used.
> New in version 3.13: [install(TARGETS)](https://cmake.org/cmake/help/latest/command/install.html#install-targets) can install targets that were created in other directories. When using such cross-directory install rules, running make install (or similar) from a subdirectory will not guarantee that targets from other directories are up-to-date. You can use [target_link_libraries()](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#command:target_link_libraries) or [add_dependencies()](https://cmake.org/cmake/help/latest/command/add_dependencies.html#command:add_dependencies) to ensure that such out-of-directory targets are built before the subdirectory-specific install rules are run.

And cmake 3.13 works well. I will change the required cmake version from 3.16 to 3.13.

cmake/gtest.cmake
```diff
@@ -15,38 +15,18 @@
...
-    # gtest will use pthreads if it's available in the system, so we must link with it
-    find_package(Threads)
+include(FetchContent)
 
-    # build directory
-    set(gtest_PREFIX ${CMAKE_BUILD_DIRECTORY}/external/gtest-prefix)
-    # install directory
-    set(gtest_INSTALL ${CMAKE_BUILD_DIRECTORY}/external/gtest-install)
+FetchContent_Declare(gtest
+  GIT_REPOSITORY https://github.com/google/googletest/
```
git-hulk: Can remove extra `/`

git-hulk:
@PragmaTwice Cool, Looks good to me. Also tested on my side on MacOSX and Debian.

cmake/gtest.cmake
PragmaTwice: done

PragmaTwice:
@ShooterIT The required cmake version has been updated to 3.13. (refer to https://github.com/apache/incubator-kvrocks/pull/564#issuecomment-1123562228)

build.sh
```diff
@@ -24,7 +24,7 @@ fi
 BUILD_DIR=$1
 WORKING_DIR=$(pwd)
 CMAKE_INSTALL_DIR=$WORKING_DIR/$BUILD_DIR/cmake
-CMAKE_REQUIRE_VERSION="3.13.0"
+CMAKE_REQUIRE_VERSION="3.16.0"
```
git-hulk:
```suggestion
CMAKE_REQUIRE_VERSION="3.13.0"
```

git-hulk (approved):

ShooterIT (approved):
LGTM, thanks for your work

tisonkun (approved):
May @git-hulk or @ShooterIT take the honor of merging this patch :)

git-hulk:
OK, many thanks for @tisonkun help, I'm very glad to merge and summary.

git-hulk:
Thanks @PragmaTwice great work again.

PR #566, opened 2022-05-09
CI also check the build with CMake status on ubuntu

### LINKED ISSUE #565: CI should also check the build status with CMake on ubuntu
Currently, Kvrocks supported make and cmake ways to build the source,
but CI workflow only check the make command. To be more rigorous, we
need to also check the cmake way.

I will file a PR to do this.

git-hulk:
Closes #565 

I disabled the `unittest` target on build.sh, can take it back after
reintroducing the gtest in the CMake.

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -34,6 +34,7 @@ jobs:
       matrix:
         os: [ubuntu-18.04]
     runs-on: ${{ matrix.os }}
+    needs: license # wait for license job
     steps:
       - name: Checkout Code Base
         uses: actions/checkout@v2.3.4
@@ -83,6 +84,7 @@ jobs:
       matrix:
         os: [ubuntu-18.04]
     runs-on: ${{ matrix.os }}
+    needs: license # wait for license job
     steps:
       - name: Checkout Code Base
         uses: actions/checkout@v2.3.4
@@ -101,6 +103,7 @@ jobs:
 
   build-on-macos-latest:
     runs-on: macos-11
+    needs: license # wait for license job
     steps:
       - name: Checkout Code Base
         uses: actions/checkout@v2.3.4
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -78,6 +78,27 @@ jobs:
           cd tests/tcl && sh runtest
           sh runtest --single integration/redis-cli && cd -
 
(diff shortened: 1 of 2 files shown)


.github/workflows/kvrocks.yaml
```diff
@@ -34,6 +34,7 @@ jobs:
       matrix:
         os: [ubuntu-18.04]
     runs-on: ${{ matrix.os }}
+    needs: license # wait for license job
```
tisonkun:
```suggestion
    needs: [license]
```

Redundant comment.

.github/workflows/kvrocks.yaml
```diff
@@ -78,8 +79,31 @@ jobs:
           cd tests/tcl && sh runtest
           sh runtest --single integration/redis-cli && cd -
 
+  build-on-ubuntu-with-cmake:
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+    runs-on: ${{ matrix.os }}
+    needs: license # wait for license job
```
tisonkun:
```suggestion
    needs: [license]
```

Redundant comment.

.github/workflows/kvrocks.yaml
```diff
@@ -78,8 +79,31 @@ jobs:
...
+          sudo apt-get install -y gcc g++ cmake
+          mkdir _build
+
+      - name: Build
+        run: |
+          sh build.sh _build
+
   build-on-macos-latest:
     runs-on: macos-11
+    needs: license # wait for license job
```
tisonkun:
```suggestion
    needs: [license]
```

Redundant comment.

.github/workflows/kvrocks.yaml
```diff
@@ -78,8 +79,31 @@ jobs:
           cd tests/tcl && sh runtest
           sh runtest --single integration/redis-cli && cd -
 
+  build-on-ubuntu-with-cmake:
```
tisonkun, line 82: I'm considering whether we can build only with cmake so that we should maintain only one build system. It's inspired by an ongoing effort #564. But we can discuss in details there later.

tisonkun (commented):

.github/workflows/kvrocks.yaml
git-hulk: OK. We also discussed about this before, I'm to maintain only one build system.

tisonkun (approved):

ShooterIT (approved):

PR #571, opened 2022-05-11
Fix make uninstall file deletion path

yancz2000:
make uninstall did not remove the binaries.


git-hulk:
Thanks @yancz2000. As mentioned in https://github.com/apache/incubator-kvrocks/pull/564#issuecomment-1121972637, we're planning to remove the Makefile to keep the build system unify and easy to maintain. @PragmaTwice will help to create a new issue to track this change later, welcome to discuss in that thread if you have any thoughts.

yancz2000:
Thanks for the timely review, I will put forward any ideas in time.

PR #576, opened 2022-05-12
Remove handwritten makefiles to keep only one build system

### LINKED ISSUE #574: Proposal: Remove handwritten makefiles to keep only one build system
### Search before asking

### Motivation

As already discussed in https://github.com/apache/incubator-kvrocks/issues/562, https://github.com/apache/incubator-kvrocks/pull/564 and https://github.com/apache/incubator-kvrocks/pull/566, remove handwritten makfiles allow us to focus on improving CMake builds and reduce maintenance costs. CMake is a build system (generator) with a higher level of abstraction than makefiles, which can describe the build structure and hierarchy of a project in a more convenient form.

### Solution

_No response_

PragmaTwice:
This resolves #574.


tisonkun:
@PragmaTwice we may still keep some targets in Makefile such as lints to running checker.

In a separated pass we can consider whether adopt Makefile to do such job or port them in scripts or other approaches.

NOTICE: CI failed because of `make lint` missing a Makefile.

PragmaTwice:
> @PragmaTwice we may still keep some targets in Makefile such as lints to running checker.
> 
> In a separated pass we can consider whether adopt Makefile to do such job or port them in scripts or other approaches.
> 
> NOTICE: CI failed because of `make lint` missing a Makefile.

I think `lint` is the only target which is out of the responsibility of cmake (not sure, becuase actually it can be written into cmake), and it is quite simple, only call two scripts, `cppcheck.sh` and `cpplint.sh`.

Hence I consider to remove the target to delete all Makefiles to make sure these handwritten makefiles will not conflict with cmake-generated makefiles (i.e. call `cmake .` directly on the root directory of the project, which will overwrite current makefiles).

tisonkun:
@PragmaTwice thanks for your investigation! SGTM do the replacement since it's trivial.

git-hulk (approved):
LGTM, I like this PR.

tisonkun (approved):

ShooterIT (approved):
Cool job, it is much clear, thanks!

There are many bug reports about linking snappy when using `make`, i think this PR can also resolve this problem.

tisonkun:
Merging...

tisonkun:
@PragmaTwice thanks for your contribution!

PR #577, opened 2022-05-13
Add LMove command

torwig:
- LMove is able to operate on a single list if the source and destination are the same
- In the case of two lists LMove acquires two locks and performs Pop and Push atomically

I wrote only unit tests because I'm not currently familiar with TCL.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,8 +1,5 @@
 # Version 999.999.999
 
-New features
-  - Add LMove command
-
 # Version 1.3.1
 
 New features
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,8 @@
 # Version 999.999.999
 
+New features
+  - Add LMove command
+
 # Version 1.3.1
 
 New features
--- a/src/lock_manager.cc
+++ b/src/lock_manager.cc
@@ -22,6 +22,7 @@
 
 #include <thread>
 #include <string>
+#include <set>
 
 LockManager::LockManager(int hash_power): hash_power_(hash_power) {
   hash_mask_ = (1U << hash_power) - 1;
@@ -52,12 +53,23 @@ void LockManager::UnLock(const rocksdb::Slice &key) {
   mutex_pool_[hash(key)]->unlock();
 }
 
-void LockManager::LockTwo(const rocksdb::Slice &first_key, const rocksdb::Slice &second_key) {
-  auto first_idx = hash(first_key);
-  auto second_idx = hash(second_key);
-  if (first_idx == second_idx) {
-    // to avoid deadlock on the same mutex: get another mutex for the second key
-    second_idx = (second_idx + 1) & hash_mask_;
+std::vector<std::mutex *> LockManager::MultiGet(const std::vector<std::string> &keys) {
+  std::vector<std::mutex *> locks;
+  std::set<unsigned, std::greater<unsigned>> to_acquire_indexes;
+  // We are using the `set` to avoid retrieving the mutex, as well as guarantee to retrieve
+  // the order of locks.
+  //
+  // For example, we need lock the key `A` and `B` and they have the same lock hash
+  // index, it will be deadlock if lock the same mutex twice. Besides, we also need
+  // to order the mutex before acquiring locks since different threads may acquire
+  // same keys with different order.
+  for (const auto &key : keys) {
+    to_acquire_indexes.insert(hash(key));
   }
-  std::lock(*mutex_pool_[first_idx], *mutex_pool_[second_idx]);
+
+  locks.reserve(to_acquire_indexes.size());
+  for (const auto &index : to_acquire_indexes) {
+    locks.emplace_back(mutex_pool_[index]);
+  }
+  return locks;
 }
--- a/src/lock_manager.cc
+++ b/src/lock_manager.cc
@@ -51,3 +51,7 @@ void LockManager::Lock(const rocksdb::Slice &key) {
 void LockManager::UnLock(const rocksdb::Slice &key) {
   mutex_pool_[hash(key)]->unlock();
 }
+
+void LockManager::LockTwo(const rocksdb::Slice &first_key, const rocksdb::Slice &second_key) {
+  std::lock(*mutex_pool_[hash(first_key)], *mutex_pool_[hash(second_key)]);
+}
--- a/src/lock_manager.cc
+++ b/src/lock_manager.cc
@@ -53,5 +53,11 @@ void LockManager::UnLock(const rocksdb::Slice &key) {
 }
 
 void LockManager::LockTwo(const rocksdb::Slice &first_key, const rocksdb::Slice &second_key) {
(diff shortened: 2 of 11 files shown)


git-hulk:
Thanks @torwig.

No worry, I can add test cases before merging.

torwig:
@git-hulk Could you please explain a little, why did the job `[lint-build-test-on-ubuntu (ubuntu-18.04)]` fail? There are some issues with TCL tests. However, I didn't add anything to them, except increase the number of commands.

git-hulk:
> @git-hulk Could you please explain a little, why did the job `[lint-build-test-on-ubuntu (ubuntu-18.04)]` fail? There are some issues with TCL tests. However, I didn't add anything to them, except increase the number of commands.

@torwig Yes, the failure was NOT related with this PR. The root cause some test cases would fail occasionally when the CI environment become slowly. We have expected to find out those and fix them.

git-hulk:
@torwig I submitted the test case PR to your repository, you can have a look and merge into your branch if it's ok. https://github.com/torwig/kvrocks/pull/1

torwig:
@git-hulk Thank you. I merged your commit into my branch.

git-hulk:
OK, thanks. Will have a look soon.

src/lock_manager.cc
```diff
@@ -51,3 +51,7 @@ void LockManager::Lock(const rocksdb::Slice &key) {
 void LockManager::UnLock(const rocksdb::Slice &key) {
   mutex_pool_[hash(key)]->unlock();
 }
+
+void LockManager::LockTwo(const rocksdb::Slice &first_key, const rocksdb::Slice &second_key) {
+  std::lock(*mutex_pool_[hash(first_key)], *mutex_pool_[hash(second_key)]);
```
git-hulk: It will be deadlock when them have the same hash value. BTW, we can extend to multi locks instead of only two, we can reuse it later.

torwig: In my implementation, if the source and destination are the same LockManager::LockTwo won't be called. Instead, only one mutex will be locked via LockGuard. Additionally, `TwoLockGuard` or `LockTwo()` can be modified to check whether two keys are equal and make a proper lock. 
Yes, I was thinking about multi-lock, but there are 2 points:
- is it possible that more than 2 locks should be locked?
- multi-version should deal with templates :)

git-hulk: Aha, we are using `crc(key)%buckets` to calculate the hash value, so different keys can't promise the value always be different.

>  Is it possible that more than 2 locks should be locked?

To be honest, I'm not sure. What I think is that `MultiLocks` contains 2 locks and more generic.

> multi-version should deal with templates :)

All is ok to me

git-hulk:
Good unit test coverage, thanks @torwig 👍

src/lock_manager.cc
torwig: Unfortunately, I can't implement `MultiLockGuard` because of a lack of experience working with templates and variable argument functions. 

> Aha, we are using crc(key)%buckets to calculate the hash value, so different keys can't promise the value always be different.

In this case, I think, even in the case of implementing `MultiLockGuard` the issue of a possible deadlock will be present.

git-hulk: My bad. They are two points:

1. Lock two keys without checking the hash value may cause deadlock
2. `MultiLockGuard` may be better than `TwoLockGuard` since it can be used in other scenarios

The current implementation also makes sense to me. Except we should check whether the hash value is same or not, only lock once if yes.

torwig: I added a check that will track the case when two indexes from the mutex pool are the same. In that case, the second index will be increased to point to reference different mutex. Does it have any drawbacks?

git-hulk: This way may unlock the wrong mutex? coz the unlock method didn't recalculate the mutex index. After reconsidering, to avoid deadlock that we need to do below things:

1. Do NOT acquire the same lock in the single thread
2. Acquire locks with the hash index order, this may also cause deadlock if one thread acquired the `keys A B` and another acquired `B A`

Maybe I can try to implement multi locks.

torwig: > This way may unlock the wrong mutex?

You are right. My bad.

git-hulk: 👍 It's a bit complex, I'm very happy to have a try and submit the PR to your branch.

git-hulk: @torwig You can have a look at: https://github.com/torwig/kvrocks/pull/2

torwig: @git-hulk Thanks! It looks like a good solution against a deadlock.

git-hulk (dismissed):

src/lock_manager.h
```diff
@@ -55,3 +58,25 @@ class LockGuard {
...
+  explicit MultiLockGuard(LockManager *lock_mgr, const std::vector<std::string> &keys):
+      lock_mgr_(lock_mgr) {
+    locks_ = lock_mgr_->MultiGet(keys);
+    for (const auto &iter : locks_) {
+      iter->lock();
+    }
+  }
+
+  ~MultiLockGuard() {
+    // Lock with order `A B C` and unlock should be `C B A`
```
ShooterIT, line 73: why we need to unlock reversely, AFAIK, they are separate.

Changelog
```diff
@@ -1,5 +1,8 @@
 # Version 999.999.999
 
+New features
+  - Add LMove command
+
```
ShooterIT:
```suggestion
```
you don't need to modify this file, i will do this when we release new version.

ShooterIT (commented):
cool, i ever wanted to add multi lock, but no action, thanks

Changelog
torwig: @ShooterIT I removed the record from Changelog. Sorry, didn't see that your suggestion can be committed just by clicking one button.

ShooterIT: no sorry, i think "commit suggestion" just provide a alternative easy way to apply CRs without changing and pushing on the local machine, but everyone has their own preferences, it is not a problem.

src/lock_manager.h
git-hulk: Yes, for the `MultiLockGuary`, it's NOT different, but I think it's a good habit the lock and unlock with the reverse order.

git-hulk (approved):

ShooterIT:
one problem i need to confirm is that this command may not adapt slot migration, we should handle it in `batch_extractor.cc`

cc @ChrisZMF

git-hulk:
> one problem i need to confirm is that this command may not adapt slot migration, we should handle it in `batch_extractor.cc`
> 
> cc @ChrisZMF

@ShooterIT Do you think is it a good idea to separate commands between Kvrocks and Kvrocks2Redis? Or we can do this in the next PR.

ShooterIT:
> @ShooterIT Do you think is it a good idea to separate commands between Kvrocks and Kvrocks2Redis? Or we can do this in the next PR.

fine, let's complete in the next PR.

ShooterIT (approved):
Thanks for impressive work @torwig

PR #584, opened 2022-05-17
Support to use lz4 compression in RocksDB

### LINKED ISSUE #601: [NEW] add lz4 lib
### Search before asking

### Motivation

KVrocks not included lz4 lib for lz4 compresssion. Only snappy is not enough，lz4 is a  efficient algorithm.

https://user-images.githubusercontent.com/52393536/168716155-82a51edf-74df-48e5-b047-f4e05a192719.png

### Solution

1. Add lz4 lib into build file.
2. Add conf for rocksdb.compression "lz4"

xiaobiaozhao:
## Description
Why need the PR?
Provides more Rocksdb.compression configuration parameters for customization

image

This closes #601


ShooterIT:
The biggest problem we need to solve is how to compile these compression algorithms.

git-hulk:
Do guys think this change makes sense for Kvrocks?

cmake/lz4.cmake
```diff
@@ -0,0 +1,82 @@
...
+endif()
+
+if (EXISTS ${LZ4_ZIP_ABSOLUT_PATH})
+	file(MD5 ${LZ4_ZIP_ABSOLUT_PATH} LZ4_MD5SUM_)
+	if (NOT ${LZ4_MD5SUM} STREQUAL ${LZ4_MD5SUM_})
+		message(FATAL_ERROR "${LZ4_ZIP_ABSOLUT_PATH} seems be something wrong, please check")
+	endif()
+else()
+	message(FATAL_ERROR "${LZ4_ZIP_ABSOLUT_PATH} seems be something wrong, please check")
+endif()
```
PragmaTwice: Could you use [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html) instead of these manual downloading commands? 
Just like [lua.cmake](https://github.com/apache/incubator-kvrocks/blob/unstable/cmake/lua.cmake#L22), it is easy to maintain and more straightforward.

cmake/lz4.cmake
```diff
@@ -0,0 +1,82 @@
...
+
+message(STATUS "remove ${LZ4_SOURCE_DIR}")
+execute_process(COMMAND sh -c "rm -rf ${LZ4_SOURCE_DIR}")
```
PragmaTwice: I think these commands are not friendly to incremental build, could you consider to remove them? 
We can use FetchContent to manipulate these file system stuff automatically.

cmake/rocksdb.cmake
```diff
@@ -38,10 +38,14 @@ FetchContent_GetProperties(snappy)
 FetchContent_MakeAvailableWithArgs(rocksdb
   CMAKE_MODULE_PATH=${PROJECT_SOURCE_DIR}/cmake/modules # to locate FindJeMalloc.cmake
   Snappy_DIR=${PROJECT_SOURCE_DIR}/cmake/modules # to locate SnappyConfig.cmake
+  lz4_ROOT_DIR=${LZ4_LIB_SOURCE_DIR}
+  lz4_LIBRARIES=${LZ4_LIB_SOURCE_DIR}/liblz4.a
+  lz4_INCLUDE_DIRS=${LZ4_LIB_SOURCE_DIR}
```
PragmaTwice: Like cmake files in [cmake/modules](https://github.com/apache/incubator-kvrocks/tree/unstable/cmake/modules), injecting the lz4 target via hooking find package mechanism can propagate dependency relation into the rocksdb target, in addition this makes the dependencies more modular, we can easily do some switches to selectively enable them.
Of course, if you are not familiar with them, I can follow up to help you implement them : )

PragmaTwice (commented):
Great job! Thanks for your hard work. Comments inline.

cmake/lz4.cmake
git-hulk: Lz4 dirs are under GPL license except the lib which we are needed, so NOT sure whether is it ok to use `FetchContent` like others.

PragmaTwice: We can just fetch files and do not call `add_subdirectory` (do not use these cmake files in lz4), like in `lua.cmake`, so I think it has few difference on the license issue than the manual way.

PragmaTwice: To facilitate the modification, I show a concrete example below:
```cmake
FetchContent_Declare(lz4
  URL https://github.com/lz4/lz4/archive/v1.9.3.tar.gz
)

FetchContent_GetProperties(lz4)
if(NOT lz4_POPULATED)
  FetchContent_Populate(lz4)

  add_custom_target(make_lz4 COMMAND make ...)
  ...
endif()
```

git-hulk: cool, thanks for Twice's explain.

cmake/lz4.cmake
datavisorxiaobiaozhao: Cool

.gitignore
```diff
@@ -44,4 +44,10 @@ Makefile.dep
 make_config.mk
 .vscode
 kvrocks2redis
+<<<<<<< HEAD
 build
+=======
+
+build/
+external/
+>>>>>>>  feat: add lz4.cmake
```
PragmaTwice: could you try to resolve the conflict?

cmake/rocksdb.cmake
PragmaTwice: maybe you should replace these to dir in `${lz4_SOURCE_DIR}/lib`

.gitignore
PragmaTwice: I think just keeping the HEAD is fine, `external` is useless.

cmake/rocksdb.cmake
xiaobiaozhao: OK, I'll do these optimizations tomorrow

.gitignore
```diff
@@ -44,4 +44,3 @@ Makefile.dep
 make_config.mk
 .vscode
 kvrocks2redis
-build
```
PragmaTwice: I think ignoring the `build` dir is useful, so could you keep it?

PragmaTwice:
The unit test failed in https://github.com/apache/incubator-kvrocks/runs/7003199495?check_suite_focus=true for 1ee66beb15407846cce3c9952dffd64ded2a3b98.
Just to record for more investigate, I will try to rerun the CI.

PragmaTwice (approved):
LGTM. cc @git-hulk @tisonkun @ShooterIT

tisonkun (commented):
Cool! Is there any test coverage for lz4 compression option? Or do you verify it locally?

git-hulk (approved):

git-hulk:
> Cool! Is there any test coverage for lz4 compression option? Or do you verify it locally?

I tested it on my side, run with LZ4 is ok and check the symbols:

```shell
000000000072a890 t LZ4HC_compress_generic_dictCtx
0000000000725ec0 t LZ4HC_compress_generic_noDictCtx.part.0
00000000007216f0 t LZ4HC_compress_optimal
0000000000730790 T LZ4_attach_HC_dictionary
0000000000714790 T LZ4_attach_dictionary
0000000000721590 T LZ4_compress
000000000070b9a0 T LZ4_compressBound
0000000000730be0 T LZ4_compressHC
0000000000730ec0 T LZ4_compressHC2
00000000007319f0 T LZ4_compressHC2_continue
...
```

tisonkun (approved):
@git-hulk thanks for your input. +1 to merge.

BTW do you confirm that introduce lz4 support in this way won't cause dependency issue? IIRC lz4 can contain GPLv2 code, but if we use `lib` only, it should be fine.

git-hulk:
> @git-hulk thanks for your input. +1 to merge.
> 
> BTW do you confirm that introduce lz4 support in this way won't cause dependency issue? IIRC lz4 can contain GPLv2 code, but if we use `lib` only, it should be fine.

Yes that we use `lib` only, but NOT sure about whether we need to involve more people to discuss or not([previous discussion thread](https://lists.apache.org/thread/0krk4nwwvzv9hgc10h6f3owo00qccsxo)). I saw some apache projects also used the `LZ4` with downloading the repo: https://github.com/search?l=CMake&q=org%3Aapache+lz4%2Flz4&type=Code.

PragmaTwice:
> @git-hulk thanks for your input. +1 to merge.
> 
> BTW do you confirm that introduce lz4 support in this way won't cause dependency issue? IIRC lz4 can contain GPLv2 code, but if we use `lib` only, it should be fine.

Yeah, I think this way only uses files in the `lib` directory, so there will be no GPL issues involved.

git-hulk:
> > @git-hulk thanks for your input. +1 to merge.
> > BTW do you confirm that introduce lz4 support in this way won't cause dependency issue? IIRC lz4 can contain GPLv2 code, but if we use `lib` only, it should be fine.
> 
> Yeah, I think this way only uses files in the `lib` directory, so there will be no GPL issues involved.

So, let's merge this PR if this's crystal clear. cc @tisonkun @PragmaTwice @ShooterIT

git-hulk:
Thanks for @xiaobiaozhao contribution again!

PR #586, opened 2022-05-17
Refactor CMakeLists: introduce target `kvrocks_objs` to avoid redundance

### LINKED ISSUE #563: [QUESTION] About compile option `-O0` in the cmake target `kvrocks`
The `-O0` option currently added to target `kvrocks` in [CMakeLists.txt](https://github.com/apache/incubator-kvrocks/blob/unstable/CMakeLists.txt#L85) is confusing, as it may prevent compilation optimizations in Release mode.

The performance impact of compile optimizations can be huge, especially with heavy use of the c++ standard library, where optimized versions can be tens of times faster than unoptimized versions.

Were there any special considerations for adding `-O0` at that time?

### LINKED ISSUE #569: Proposal: Consider introducing target `libkvrocks` and using `file(GLOB ...)` to avoid redundance in CMake
Currently there are three targets in cmakelists: `kvrocks`, `kvrocks2redis` and `unittest`. these three targets have many common properties which are redundant, we can introduce a new target `libkvrocks`, define common properties in this target (e.g. source files in `src/` except `src/main.cc`), header directories to include, compilation options `-Wall ...`, standard versions `cxx_std_11`, etc.), and link this target to `kvrocks`, `kvrocks2redis`, and `unittest`.

We can also use `file(GLOB ...) ` and list operations in CMake to search for specific files in a directory to avoid a handwritten list of source files.

PragmaTwice:
Closes #563 and closes #569.

`-O0` and `-g` is removed:
- `-O0` is automatically applied while `-DCMAKE_BUILD_TYPE=Debug` is passed to cmake
- `-g` is automatically applied while `-DCMAKE_BUILD_TYPE=Debug` (or `RelWithDebInfo`) is passed to cmake

(in most generators of CMake, like Makefiles, the default `CMAKE_BUILD_TYPE` is `Debug`)

`file(GLOB ...)` is used to avoid handwritten source file lists.

Advantages:
- Do not need to modify CMakeLists.txt while you add/remove some source files.
- Any common properties e.g. compile options, include directories etc. can be added only to `kvrocks_objs` target, not every target including `kvrocks`, `kvrocks2redis` and `unittest`.


PragmaTwice:
I cannot reproduce the test error of github actions on my local, could you check it for whether it is about the PR or not? Thanks! @git-hulk

git-hulk:
@PragmaTwice It's NOT related with this PR, I will rerun the CI. I didn't investigate why this test case frequently failed in CI environment. @ChrisZMF Can you have a look?

git-hulk (dismissed):
Cool, many thanks for simplifying the CMake, cheers!

tisonkun (dismissed):
LGTM. Thanks for your contribution!

ShooterIT (dismissed):

ShooterIT:
Thanks @PragmaTwice are you familiar with `address sanitizer` and `thread sanitize`. Kvrocks lacks these check on CI currently. We try to do that, but find it doesn't work well. @git-hulk do you remember？

git-hulk:
> Thanks @PragmaTwice are you familiar with `address sanitizer` and `thread sanitize`. Kvrocks lacks these check on CI currently. We try to do that, but find it doesn't work well. @git-hulk do you remember？

Yes, I found the address sanitizer can't detect the explicitly memory leak that mallocs memory without releasing them.


@ShooterIT Can you help to summary and merge this PR? We can file another issue to discuss about the sanitizer.

PragmaTwice:
> Thanks @PragmaTwice are you familiar with `address sanitizer` and `thread sanitize`. Kvrocks lacks these check on CI currently. We try to do that, but find it doesn't work well. @git-hulk do you remember？

Maybe a higher version of GCC (>= 5) and `-fsanitize=leak` is needed for detecting memory leaks.
BTW I highly recommand Clang while sanitizer or fuzzer instrumentation is required, LeakSanitizer is a part of AddressSanitizer in Clang.

PragmaTwice:
Try it here: https://godbolt.org/z/8r48Mvz3W : )

I will improve these sanitizer options in another PR.

git-hulk:
> > Thanks @PragmaTwice are you familiar with `address sanitizer` and `thread sanitize`. Kvrocks lacks these check on CI currently. We try to do that, but find it doesn't work well. @git-hulk do you remember？
> 
> Maybe a higher version of GCC (>= 5) and `-fsanitize=leak` is needed for detecting memory leaks. BTW I highly recommand Clang while sanitizer or fuzzer instrumentation is required, LeakSanitizer is a part of AddressSanitizer in Clang.

Yes, we are running the sanitizer in Clang and can work when memory leak directly in starting the server, but after that. For example, leaks in `HSET` command, the sanitizer can't find out this issue even we ran this command.

PragmaTwice:
That's weird. I'll investigate it later.

CMakeLists.txt
```diff
@@ -50,351 +46,60 @@ include(cmake/rocksdb.cmake)
...
-target_compile_features(kvrocks2redis PRIVATE cxx_std_11)
-target_compile_options(kvrocks2redis PRIVATE -Wall -Wpedantic -g -Wsign-compare -Wreturn-type)
-option(ENABLE_ASAN "enable ASAN santinizer" OFF)
+add_library(kvrocks_objs OBJECT ${KVROCKS_SRCS})
+
+target_include_directories(kvrocks_objs PUBLIC src ${PROJECT_BINARY_DIR})
+target_compile_features(kvrocks_objs PUBLIC cxx_std_11)
+target_compile_options(kvrocks_objs PUBLIC ${WARNING_FLAGS} -fno-omit-frame-pointer)
+target_link_libraries(kvrocks_objs PUBLIC -fno-omit-frame-pointer)
 if(ENBALE_ASAN)
```
PragmaTwice: Oh, there is a typo in the original CMakeLists, `ENBALE -> ENABLE`, which is found while I look into the build behaviors while ASAN is enabled.
Is it a reason for this issue? @git-hulk @ShooterIT

PragmaTwice:
> Yes, we are running the sanitizer in Clang and can work when memory leak directly in starting the server, but after that. For example, leaks in HSET command, the sanitizer can't find out this issue even we ran this command.

It works on my side (archlinux, kernel 5.10, gcc 11.2) for a simple leak in `HSET`:

Screenshot
Screenshot

I think if there is some leak allocated in rocksdb or other deps, it may cannot be detected in current build configurations.

CMakeLists.txt
ShooterIT: maybe😂

CMakeLists.txt
```diff
@@ -50,351 +46,60 @@ include(cmake/rocksdb.cmake)
...
 
-add_executable(kvrocks)
-target_compile_features(kvrocks PRIVATE cxx_std_11)
-target_compile_options(kvrocks PRIVATE -Wall -Wpedantic -g -Wsign-compare -Wreturn-type -fno-omit-frame-pointer -O0)
-option(ENABLE_ASAN "enable ASAN santinizer" OFF)
-if(ENBALE_ASAN)
-    target_compile_options(kvrocks PRIVATE -fsanitize=address)
-    target_link_libraries(kvrocks PRIVATE -fsanitize=address)
-endif()
+set(WARNING_FLAGS -Wall -Wpedantic -Wsign-compare -Wreturn-type)
```
ShooterIT, line 70: we want to use `-O2 -g -DNDEBUG` in release version.
what are the default values

PragmaTwice: CMake will add `-O3 -DNDEBUG ...` for `CMAKE_BUILD_TYPE=Release`, `-O2 -DNDEBUG -g ...` for `RelWithDebInfo` and `-O0 -g ...` for `Debug` in the makefile generator.
You can use `RelWithDebInfo` directly without changing the semantics of `Release` mode if you want `-O2 -g -DNDEBUG`.

ShooterIT: @git-hulk should we us `CMAKE_BUILD_TYPE= RelWithDebInfo` instead of `Release` in `build.sh`?
AFAIK, `-g` also is important in production environment.

git-hulk: I think we should keep the debug info that will help us a lot when crashing, even it will sacrifice a bit optimize feature. How do you think? @PragmaTwice @ShooterIT

PragmaTwice: I think it is ok to use `RelWithDebInfo` (`-g -O2 -DNDEBUG`) mode in production environments.

ShooterIT: yes, exactly

CMakeLists.txt
git-hulk: My bad for this stupid typo.

CMakeLists.txt
git-hulk: Cool, @PragmaTwice can help to do this change, then can merge this improvement.

PragmaTwice:
Changed build type to `RelWithDebInfo`, and also fixed the typo. @git-hulk @ShooterIT

git-hulk (approved):

ShooterIT (approved):

PR #588, opened 2022-05-18
Fix release CI: use build.sh & use fresh directory & remove useless deps

PragmaTwice:
- use build.sh to enable `RelWithDebInfo` mode
- use a fresh directory for packaging
- remove libsnappy package since we link snappy static library built by ourselves
- other small fix on shell commands

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -42,17 +42,16 @@ jobs:
       - name: Install Dependencies
         run: |
           sudo apt-get install -y tar cmake
-          mkdir -p build/bin
-          mkdir -p build/conf
+          mkdir -p build/release/bin
+          mkdir -p build/release/conf
 
       - name: Build
         run: |
+          ./build.sh build
           cd build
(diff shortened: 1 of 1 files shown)


ShooterIT:
Other CI, i think we also should use `RelWithDebInfo` mode.

> use a fresh directory for packaging

Could explain the reason. The old way can't work well?

PragmaTwice:
> Other CI, i think we also should use `RelWithDebInfo` mode.
> 
> > use a fresh directory for packaging
> 
> Could explain the reason. The old way can't work well?

While we use Makefiles, the `build` directory is empty.
But now we use `cd build && cmake ..`, so there is lots of CMake-generated files in the `build` dir, which is useless for packaging.  
Hence I move some binary and kvrocks.conf to `build/release` to package.

ShooterIT (approved):

git-hulk (approved):

tisonkun:
FWIW we may not use GitHub Actions to do the release work. At least the artifacts should go to Apache distribution file server.

PR #592, opened 2022-05-19
Search library first before linking `libstdc++.a` in CMake

### LINKED ISSUE #580: Proposal: Search library first before linking `libstdc++.a` in CMake
### Search before asking

### Motivation

This idea comes from @git-hulk, and the following is the discussion at that time:
> > > The problem is caused by that `-static-libstdc++` is added as a flag to the compiler in [CMakeLists.txt#L72](https://github.com/apache/incubator-kvrocks/blob/unstable/CMakeLists.txt#L72), which is not added in makefiles, so a static library of libstdc++ is needed.
> > 
> > 
> > Maybe some tips should be added to README to let users know that kvrocks need libstdc++.a for build. @git-hulk
> 
> Yes, you are right. Do you think it's a good idea to search library first before linking?

I think it is a good idea, there are two solutions for searching library:
1. find library, and give an error or warning if not found; give an option to user of whether `-static-libstdc++` is enabled
2. find library, do not enable `-static-libstdc++` if not found

I prefer the first solution, because I think the second is more implicit, users may not know whether `-static-libstdc++` is enabled while they are building the project (it is totally depend on the environment).

_Originally posted by @PragmaTwice in https://github.com/apache/incubator-kvrocks/issues/575#issuecomment-1125220948_

### Solution

_No response_

PragmaTwice:
Close #580.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,8 +66,10 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static-libgcc")
     
     if(ENABLE_STATIC_LIBSTDCXX)
-        find_library(FOUND_STATIC_LIBSTDCXX libstdc++.a)
-        if(NOT (FOUND_STATIC_LIBSTDCXX OR FORCE_STATIC_LIBSTDCXX))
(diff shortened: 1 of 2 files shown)


PragmaTwice:
I first try to use `find_library` to find `libstdc++.a`, but in some environment (like github actions), cmake cannot find it via `find_library` but `gcc -static-libstdc++` actually works.
So I change the method to `try_compile`, the command will try to compile source files with some special compile&link options to test whether these options exist and work well.

git-hulk:
cool, thanks @PragmaTwice contribution.

ShooterIT (approved):

git-hulk (approved):
LGTM

PR #594, opened 2022-05-19
Add optional parameter to LPOP/RPOP to pop more than 1 element

torwig:
Related TCL-test were copied from the Redis repository.
Method `List::Pop` refactored: output param `std::string *elem` became the last one for convenience purposes.

Please let me know if something should be checked additionally.

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -1484,7 +1484,7 @@ class CommandPop : public Commander {
       }
     } else {
       std::string elem;
-      rocksdb::Status s = list_db.Pop(args_[1], &elem, left_);
+      rocksdb::Status s = list_db.Pop(args_[1], left_, &elem);
       if (!s.ok() && !s.IsNotFound()) {
         return Status(Status::RedisExecErr, s.ToString());
       }
@@ -1570,7 +1570,7 @@ class CommandBPop : public Commander {
     rocksdb::Status s;
     for (const auto &key : keys_) {
       last_key = key;
-      s = list_db.Pop(key, &elem, left_);
+      s = list_db.Pop(key, left_, &elem);
       if (s.ok() || !s.IsNotFound()) {
         break;
       }
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -65,6 +65,7 @@ const char *errInvalidExpireTime = "invalid expire time";
 const char *errWrongNumOfArguments = "wrong number of arguments";
 const char *errValueNotInterger = "value is not an integer or out of range";
 const char *errAdministorPermissionRequired = "administor permission required to perform the command";
+const char *errValueMustBePositive = "value is out of range, must be positive";
 
 class CommandAuth : public Commander {
  public:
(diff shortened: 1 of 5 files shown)


git-hulk:
Thanks @torwig, we can use `popMulti` to implement the `pop` as well? since I think pop is the special condition for popMulti.

torwig:
@git-hulk I had a similar idea, but then I read the Redis docs:

> When called without the count argument:
> Bulk string reply: the value of the first element, or nil (**$-1**) when key does not exist.
> When called with the count argument:
> Array reply: list of popped elements, or nil (***-1**) when key does not exist.

and I saw that BLPOP/BRPOP and RPopLPush use just `Pop` without `count` option, so I decided to leave `Pop` for popping a single element and introduce `PopMulti` for popping more than one element.
Additionally, it's more explicit to return multiple elements from `PopMulti` inside a vector than to return just a single element inside a vector (if it's a special case for pop).

git-hulk:
@torwig Oooop, my bad that didn't have a check at where's using `pop`. Current implementation is good to me.

git-hulk (dismissed):

ShooterIT:
thanks @torwig Whatever i don't think it is a good idea to have `PopMulti` and `Pop`, we should implement `Pop` as a special `PopMulti`

torwig:
@ShooterIT I replaced Pop implementation with multi-pop's one and pushed.
Or do you think I should implement Pop like this:

```
rocksdb::Status List::Pop(const Slice &user_key, bool left, std::string *elem) {
  std::vector<std::string> elems;
  auto s = PopMulti(user_key, left, 1, &elems);
  if (!s.ok()) return s;
  *elem = elems[0];
  return rocksdb::Status::OK();
}
```
?
And in this case, the code will be much cleaner.

ShooterIT:
```
rocksdb::Status List::Pop(const Slice &user_key, bool left, std::string *elem) {
  std::vector<std::string> elems;
  auto s = PopMulti(user_key, left, 1, &elems);
  if (!s.ok()) return s;
  *elem = elems[0];
  return rocksdb::Status::OK();
}
```

I think this idea is better. WDYT @git-hulk

torwig:
@git-hulk I've just pushed the version with the idea that I mentioned (Pop as PopMulti).
Here you can see the previous version with just a single `PopMulti` method in all the places: https://github.com/torwig/kvrocks/commit/48ca0d911db02dc1f6276e3b4e1ce5f106cc39e7

git-hulk:
cool, thanks @torwig

git-hulk:
> ```
> rocksdb::Status List::Pop(const Slice &user_key, bool left, std::string *elem) {
>   std::vector<std::string> elems;
>   auto s = PopMulti(user_key, left, 1, &elems);
>   if (!s.ok()) return s;
>   *elem = elems[0];
>   return rocksdb::Status::OK();
> }
> ```
> 
> I think this idea is better. WDYT @git-hulk

Yes, I think that retrieves multi elements in `pop` is a bit weird, it would be better to use `popMulti` to implement `pop`.

ShooterIT (approved):

git-hulk (approved):

git-hulk:
@ShooterIT Can you help to merge this PR?

PR #599, opened 2022-05-20
Improve using of ASan and TSan in CMake build

PragmaTwice:
Close #590.

PR #605, opened 2022-05-24
unittests: fix leaks reported by ASan

PragmaTwice:
use `std::unique_ptr` to avoid missing delete and dtor call

this fix can remove a large amount of leak reports from ASan (~10000+ lines in stderr), so we can check easier when other important leaks is reported.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -27,6 +27,7 @@
 #include <cctype>
 #include <string>
 #include <vector>
+#include <utility>
 #include <memory>
 
 #include "status.h"
--- a/src/util.h
+++ b/src/util.h
@@ -27,6 +27,7 @@
 #include <cctype>
 #include <string>
 #include <vector>
+#include <memory>
 
 #include "status.h"
 #include "chrono"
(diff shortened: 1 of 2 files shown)


tisonkun (approved):
@PragmaTwice could you share the ASan report before and after this change?

PragmaTwice:
> @PragmaTwice could you share the ASan report before and after this change?

The ASan report before this is about 6MiB, I will soon find a way to upload. Seems reach the size limit in GitHub?

PragmaTwice:
[kvrocks-unittest-asan.log](https://github.com/apache/incubator-kvrocks/files/8760302/kvrocks-unittest-asan.log)
[kvrocks-unittest-asan-after.log](https://github.com/apache/incubator-kvrocks/files/8760306/kvrocks-unittest-asan-after.log)

Seems just some error due to my poor network, not limit exceeding. Done @tisonkun.

tisonkun:
@PragmaTwice looks great! It seems that this PR fixes all leaks currently reported. We should fail if there are new leaks in #599.

git-hulk (approved):
LGTM, cool!

PragmaTwice:
> @PragmaTwice looks great! It seems that this PR fixes all leaks currently reported. We should fail if there are new leaks in #599.

Seems there are also some leaks in TCL tests, I do not know the current status, maybe @git-hulk know that.
In #599, ASan is enabled in `daily-ci.yaml`, dont know whether it should be added to `kvrocks.yaml`.

ASan will slow down the program execution about x2 times, and TSan about x10 times.
Refer to https://clang.llvm.org/docs/AddressSanitizer.html and https://clang.llvm.org/docs/ThreadSanitizer.html.

git-hulk:
Yes, I can reproduce the memory leak on my side, but didn't investigate root causes now.

tisonkun:
I'm going to merge this PR.

Further discussion about #599 can go to that thread. If you'd like to investigate leak issues, it'd be better to create another issue.

PR #606, opened 2022-05-24
Fix the slave may crash when restoring the db

### LINKED ISSUE #603: The slave would crash when restoring the db from checkpoint error
### Search before asking

### Version

2.0.6

### Minimal reproduce step

1. slaveof `x.x.x.x`
2. remove some files to fail the restore db process

The slave should serve with the old database

### What did you see instead?

image

### Anything Else?

_No response_

git-hulk:
The db pointer was forbidden to use when restoring the db, but we
didn't check the loading status in `recordInstantaneousMetrics`
in the server's cron thread. We can skip all the cron operations
when restoring the db instead of checking status one by one.

This closes #603

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -558,8 +558,15 @@ void Server::cron() {
     if (storage_->IsClosing()) continue;
 
     updateCachedTime();
+    counter++;
+    if (is_loading_) {
+      // We need to skip the cron operations since `is_loading_` means the db is restoring,
+      // and the db pointer will be modified after that. It will panic if we use the db pointer
+      // before the new db was reopened.
+      continue;
+    }
     // check every 20s (use 20s instead of 60s so that cron will execute in critical condition)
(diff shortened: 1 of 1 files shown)


ShooterIT (approved):

tisonkun (approved):
LGTM

src/server.cc
```diff
@@ -558,8 +558,15 @@ void Server::cron() {
     if (storage_->IsClosing()) continue;
```
ShooterIT, line 558: @git-hulk here we already check db is closing, are you sure kvrock would execute the next code

git-hulk: Yes, there still have the window between `closing` and the db pointer is ready, since we set it to `true` at the begin of the open db.

ShooterIT: ok, thanks

PR #610, opened 2022-05-28
Update max_bytes_for_level_base unit to bytes and remove unuseful error checker.

WyattJia:
This refers to #612

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -174,7 +174,7 @@ Config::Config() {
       {"rocksdb.blob_garbage_collection_age_cutoff",
        false, new IntField(&RocksDB.blob_garbage_collection_age_cutoff, 25, 0, 100)},
       {"rocksdb.max_bytes_for_level_base",
-        false, new IntField(&RocksDB.max_bytes_for_level_base, 256, 0, INT_MAX)},
+        false, new IntField(&RocksDB.max_bytes_for_level_base, 268435456, 0, INT_MAX)},
       {"rocksdb.max_bytes_for_level_multiplier",
         false, new IntField(&RocksDB.max_bytes_for_level_multiplier, 10, 1, 100)},
       {"rocksdb.level_compaction_dynamic_level_bytes",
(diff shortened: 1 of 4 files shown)


src/config.cc
```diff
@@ -449,26 +449,17 @@ void Config::initFieldCallback() {
       }},
       {"rocksdb.max_bytes_for_level_base", [this](Server* srv, const std::string &k, const std::string& v)->Status {
         if (!srv) return Status::OK();
-        if (!RocksDB.level_compaction_dynamic_level_bytes) {
```
tisonkun, line 452: Copy comment from https://github.com/apache/incubator-kvrocks/pull/611#discussion_r884190774:

Why do you make these changes?

tisonkun: It seems another option and with this patch you remove all usages of this config. There must be a motivation and if you remove all usages, perhaps it's necessary to remove the option also.

WyattJia: > Copy comment from [#611 (comment)](https://github.com/apache/incubator-kvrocks/pull/611#discussion_r884190774):
> 
> Why do you make these changes?

After my discussion with @ShooterIT , we believe that this configuration item may not be required. If the configuration is not filled in, an error should not be returned, and the default value should be set to false.

ShooterIT:
@WyattJia please also check again
```
  options->max_bytes_for_level_base = config_->RocksDB.max_bytes_for_level_base * MiB;
```
need to remove `* MiB`

WyattJia:
> @WyattJia please also check again
> 
> ```
>   options->max_bytes_for_level_base = config_->RocksDB.max_bytes_for_level_base * MiB;
> ```
> 
> need to remove `* MiB`

@ShooterIT @tisonkun I have updated changes about unit and error config checker.Please check and review. Very thankful.

ShooterIT (approved):

git-hulk (approved):

git-hulk:
Thanks @WyattJia, merged.

PR #613, opened 2022-05-29
Allow to send logs to stdout/stderr

### LINKED ISSUE #478: Is it possible to send log output to system out instead of file?
### Discussed in https://github.com/KvrocksLabs/kvrocks/discussions/477

<sup>Originally posted by **djotanov** February 10, 2022</sup>
Is it possible to send kvrocks logging output to system out instead of file?
This is needed in docker environment. </div>

git-hulk:
This closes #478 

Currently, we only supported sending logs to the log file,
but we expected to collect logs through stdout/stderr instead of
the log file when deploying the Kvrocks in the container. We also
upgrade the glog to v0.6.0 since logstostdout was supported after that.

### DIFF
--- a/cmake/glog.cmake
+++ b/cmake/glog.cmake
@@ -21,7 +21,7 @@ include(FetchContent)
 
 FetchContent_Declare(glog
   GIT_REPOSITORY https://github.com/google/glog
-  GIT_TAG v0.4.0
+  GIT_TAG v0.6.0
 )
 
 include(cmake/utils.cmake)
--- a/cmake/snappy.cmake
+++ b/cmake/snappy.cmake
@@ -29,4 +29,5 @@ include(cmake/utils.cmake)
 FetchContent_MakeAvailableWithArgs(snappy
   SNAPPY_BUILD_TESTS=OFF
   SNAPPY_BUILD_BENCHMARKS=OFF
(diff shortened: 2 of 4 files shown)


git-hulk:
This CI failure seems was related to issue: https://github.com/google/snappy/pull/128, will investigate why it only happens in current PR(maybe the GCC version was updated).

PragmaTwice:
> This CI failure seems was related to issue: [google/snappy#128](https://github.com/google/snappy/pull/128), will investigate why it only happens in current PR(maybe the GCC version was updated).

Try add `BUILD_SHARED_LIBS=OFF` to https://github.com/apache/incubator-kvrocks/blob/unstable/cmake/glog.cmake#L32, or set a global variable `set(BUILD_SHARED_LIBS OFF CACHE ...)`.
Some options are changed in glog v0.6.0, which may affect the global configurations.
This error occurred only when shared libraries of snappy are built.

git-hulk:
cool, thanks @PragmaTwice.

ShooterIT (approved):

tisonkun (approved):

git-hulk:
Thanks @ShooterIT @tisonkun @PragmaTwice, will merge this PR.

PR #614, opened 2022-05-31
Compile lua by C++ compilers to avoid memory leaks

### LINKED ISSUE #609: Memory leaks in Lua::redisGenericCommand
### Search before asking

### Version

`redisGenericCommand` use `raiseError` to process errors, but `raiseError` will perform a longjmp in the `lua_error` api call, so `raiseError` never returns.

longjmp does not compatible with RAII in C++ (C++ exceptions may use longjmp in implementation but the compiler will insert some dtor call autometically and carefully, i.e. stack unwinding), so every object in the stack of `redisGenericCommand` will never be destroyed, i.e. no destructor will be call, hence all resource allocated in ctor like `std::vector`, `std::unique_ptr` etc. cannot be released.

@git-hulk

### Minimal reproduce step

`./runtest --dont-clean` on a kvrocks binary with Address Sanitizer enabled. (try to build with #599)
You can only run tests in unit/scripting to save time : )

No memory leaks

### What did you see instead?

Memory leaks are reported by Address Sanitizer

### Anything Else?

_No response_

PragmaTwice:
The clause `by C++ compilers` just means `in C++ mode` since most mainstream C compilers are also a C++ compiler. 😆 
Close #609.


PragmaTwice:
There are some weird problem in the default xcode toolchains (i.e. apple clang):
```
$ echo "#include <assert.h>" | /Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -c -
clang: warning: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated [-Wdeprecated]
a.c:1:10: fatal error: 'assert.h' file not found
#include <assert.h>
         ^~~~~~~~~~
1 error generated.
```

So I try to use `clang++` in PATH as cxx compiler instead.

PragmaTwice:
Blocked by https://github.com/KvrocksLabs/lua/pull/3.

git-hulk:
@PragmaTwice Seems ${CMAKE_CXX_COMPILER} will use the apple-clang and $CXX will use clang++, clang++ works good after updating the Lua. So we can replace `${CMAKE_CXX_COMPILER}` with `${CXX}` in lua.cmake?

PragmaTwice:
> @PragmaTwice Seems ${CMAKE_CXX_COMPILER} will use the apple-clang and $CXX will use clang++, clang++ works good after updating the Lua. So we can replace `${CMAKE_CXX_COMPILER}` with `${CXX}` in lua.cmake?

I think the concern here is that we'd better to keep the same c++ compiler during the whole build process. So I pass the c++ compiler specified by cmake to make.

git-hulk:
> > @PragmaTwice Seems ${CMAKE_CXX_COMPILER} will use the apple-clang and $CXX will use clang++, clang++ works good after updating the Lua. So we can replace `${CMAKE_CXX_COMPILER}` with `${CXX}` in lua.cmake?
> 
> I think the concern here is that we'd better to keep the same compiler during the whole build process. So I pass the c++ compiler specified by cmake to make.

I got your point, but it will fail to build if we didn’t specify them explicitly. I’m not sure if it’s ok to other users.

PragmaTwice:
> > > @PragmaTwice Seems ${CMAKE_CXX_COMPILER} will use the apple-clang and $CXX will use clang++, clang++ works good after updating the Lua. So we can replace `${CMAKE_CXX_COMPILER}` with `${CXX}` in lua.cmake?
> > 
> > 
> > I think the concern here is that we'd better to keep the same compiler during the whole build process. So I pass the c++ compiler specified by cmake to make.
> 
> I got your point, but it will fail to build if we didn’t specify them explicitly. I’m not sure if it’s ok to other users.

It is a common usage to specify `CMAKE_CXX_COMPILER` before build e.g. `cmake .. -DCMAKE_CXX_COMPILER=/some/path/some-g++`, so I think using `CXX` may lead to different compilers in one build.
I think we can use `try_compile` to workaround it: if it is in osx and `try_compile(a code snippet with <assert.h>)` fail, we will use `CXX`, otherwise we use `CMAKE_CXX_COMPILER` to keep C++ compiler unchanged.
These logics are easy to implement in cmake : )

PragmaTwice:
> It is a common usage to specify `CMAKE_CXX_COMPILER` before build e.g. `cmake .. -DCMAKE_CXX_COMPILER=/some/path/some-g++`, so I think using `CXX` may lead to different compilers in one build. I think we can use `try_compile` to workaround it: if it is in osx and `try_compile(a code snippet with <assert.h>)` fail, we will use `CXX`, otherwise we use `CMAKE_CXX_COMPILER` to keep C++ compiler unchanged. These logics are easy to implement in cmake : )

done :tada:

git-hulk:
cool, thanks to @PragmaTwice great work!

.github/workflows/kvrocks.yaml
```diff
@@ -88,11 +88,11 @@ jobs:
...
         run: |
-          brew install gcc autoconf automake libtool cmake
+          brew install autoconf automake libtool cmake
           mkdir build
 
       - name: Build
         run: |
           cd build
-          cmake ..
+          cmake .. -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
```
git-hulk: So can we remove CMAKE_CXX_COMPILER/CMAKE_C_COMPILER? since we can workaround it when failing to build.

tisonkun:
@PragmaTwice as https://github.com/KvrocksLabs/lua/pull/3 merged, is it still necessary to introduce such workaround? IIUC now it should be happy to build with `CMAKE_CXX_COMPILER`.

src/scripting.cc
```diff
@@ -74,12 +74,10 @@ enum {
   LL_WARNING,
 };
 
-extern "C" {
```
git-hulk, line 77: The tcl test cases failure seems was caused by this change.

PragmaTwice: Sorry, the error is caused by a typo in cmake, I have fixed it in the latest commit.
If `extern "C"` is added, the linker will fail, since we compile lua by c++ compiler and so these lua symbols is mangled by Itanium C++ ABI.

.github/workflows/kvrocks.yaml
PragmaTwice: done

src/scripting.cc
PragmaTwice: `extern "C"` in C++ is basically aimed to prevent function symbols from mangling by c++ abi.

.github/workflows/kvrocks.yaml
PragmaTwice: seems some problem here, I will check it

PragmaTwice:
I think I figure out this problem.
CMake deals with apple clang well because it automatically adds `-isysroot ${CMAKE_OSX_SYSROOT}` to compile options.
But Makefiles in lua do not do that, so `assert.h` cannot be found while using apple `clang++`.
We can just add it to `LUA_CFLAGS`!
@git-hulk @tisonkun

PragmaTwice:
Sorry for many time to solve it, I do not have an Mac osx device so I use github actions in a temp repo to debug the build process.

git-hulk (approved):

tisonkun (approved):
Great work! Thanks for your contribution.

git-hulk:
Thanks for @PragmaTwice great investigation and contribution again. 👍

PR #618, opened 2022-06-03
Build static lib of glog in CMake

PragmaTwice:
Fix #615.

After this PR (liblzma.so is brought in by libunwind.so):
```
$ ldd kvrocks
        linux-vdso.so.1 (0x00007fff165d7000)
        libunwind.so.8 => /usr/lib/libunwind.so.8 (0x00007fa86a7ae000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007fa86a6c6000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007fa86a4bc000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007fa86b8d8000)
        liblzma.so.5 => /usr/lib/liblzma.so.5 (0x00007fa86a493000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fa86a48e000)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,8 @@ if(NOT DISABLE_JEMALLOC)
     list(APPEND EXTERNAL_LIBS PRIVATE jemalloc)
 endif()
 
+set(BUILD_SHARED_LIBS OFF CACHE BOOL "do not build shared libs by default")
+
 include(cmake/gtest.cmake)
 include(cmake/glog.cmake)
 include(cmake/snappy.cmake)
--- a/cmake/glog.cmake
+++ b/cmake/glog.cmake
@@ -29,4 +29,5 @@ include(cmake/utils.cmake)
 FetchContent_MakeAvailableWithArgs(glog
   WITH_GFLAGS=OFF
   WITH_GTEST=OFF
+  BUILD_SHARED_LIBS=OFF
 )


git-hulk (approved):

CMakeLists.txt
```diff
@@ -45,6 +45,8 @@ if(NOT DISABLE_JEMALLOC)
     list(APPEND EXTERNAL_LIBS PRIVATE jemalloc)
 endif()
 
+set(BUILD_SHARED_LIBS OFF CACHE BOOL "do not build shared libs by default")
```
tisonkun, line 48: Why do we set this flag on root CMakeLists also?

BTW, the comment seems magic without background (why forbid shared lib?).

tisonkun (commented):
Thanks for your contribution! Comments inline.

CMakeLists.txt
PragmaTwice: I think we just want to build static libraries of dependences and link them statically, so I added it to CMakeLists to avoid some dependencies set this cache variable to true in their cmake to pollute the global configuration. (cache variables can only be assigned once, and the next assignment needs to specify `FORCE`, otherwise it is invalid)

PragmaTwice: Static libraries allow users to not worry about library versioning issues and linking issues, and are more portable (just like go / rust, statically link almost everything), which I think is probably secondary to the increase in size of these binaries.
And `BUILD_SHARED_LIBS` only controls the default behavior, if we want to build shared libraries then we can still specify `SHARED` manually.

tisonkun: Thanks for your explanation. Sounds good to me.

tisonkun (approved):

tisonkun:
Merging...

FWIW a discussion (#615) cannot be resolved. You may create an issue from the discussion the next time.

PragmaTwice:
> Merging...
> 
> FWIW a discussion (#615) cannot be resolved. You may create an issue from the discussion the next time.

My mistake, will open an issue next time :)

PR #620, opened 2022-06-04
Fix don't swallow the error when creating column families failed

### LINKED ISSUE #437: [BUG]  Failed to open DB with missing CURRENT file
**Describe the bug**

 - OS: MAC OS 10.15.7
 - Version  2.0.5 @6a4ed16a

get following error message when try to start up kvrocks ( first time run )

E1229 15:41:02.217978 276704704 main.cc:314] Failed to open: IO error: No such file or directory: While opening a file for sequentially reading: /tmp/kvrocks/db/CURRENT: No such file or directory

/tmp/kvrocks directory did exist but no sub-directory ./db under it.

git-hulk:
We expected to ignore the column families not opened error
opened error, but we ignored all errors without checking the
error type and message. This may cause other errors being ignored
as well.

This closes: https://github.com/apache/incubator-kvrocks/issues/437

### DIFF
--- a/src/storage.cc
+++ b/src/storage.cc
@@ -20,7 +20,6 @@
 
 #include "storage.h"
 
-#include <strings.h>
 #include <fcntl.h>
 #include <sys/stat.h>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


src/storage.cc
```diff
@@ -209,9 +210,16 @@ Status Storage::CreateColumnFamilies(const rocksdb::Options &options) {
     tmp_db->Close();
     delete tmp_db;
   }
-  // Open db would be failed if the column families have already exists,
```
tisonkun, line 212: Is it possible we check exact that it failed because of "already exists" instead of "NOT column families not opened"? I'm afraid there is still some edge case we miss.

Also, shall we prefer C++ idiom and thus check strings with `<string>`? That means:

* `s.ToString().rfind(notOpenedPrefix, 0) != 0`

over

* `!strncasecmp(s.ToString().c_str(), notOpenedPrefix.c_str(), notOpenedPrefix.size())`

git-hulk: > Is it possible we check exact that it failed because of "already exists" instead of "NOT column families not opened"? I'm afraid there is still some edge case we miss.

Yes, it's a bit tricky and confuse, but we can't change the status message AFAIK since it returned by the RocksDB.

For the second suggestion, We do prefer C++ idiom. my initial intention was to ignore the case when comparing the prefix. But seems this's unnecessary since it can't protect anything if the status message is modified, I will change to `find` to simplify the expression.

tisonkun: Sorry I just (again, haha) revert the boolean logic :P

Ping me when you have a conclusion on whether use rfind or keep strncasecmp.

git-hulk: Thanks @tisonkun, changed to the c++ api and do a bit logic refactor, please take a look again.

tisonkun (approved):
LGTM

ShooterIT (approved):

Issue #572, opened 2022-05-11
Proposal: add multiple platforms and CPUs build workflow in GitHub Actions


git-hulk:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

As far as I know, Kvrocks works well on platforms like CentOS/Ubuntu/Debian and MacOSX, as well as CPU like ARM64 and Apple M1. But we didn't check the build system status now and we can't guarantee it's always fine over time. So I think we should check those statuses periodically.

### Solution

For most PRs won't affect the build system status, but it's a bit time and resource costly. The workaround is to check daily or weekly instead of every PR, and we can fix them in time when anyone was broken. Do you guys have any thoughts? 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

tisonkun:
IIRC [GitHub Actions doesn't provide hosted ARM64 machine](https://github.community/t/ask-about-github-hosted-runners-arm64-platform-support/167483) ([another instance](https://issues.apache.org/jira/browse/CURATOR-634)). We may need a sponsor for self-hosted runners, other CI platform or ask ASF Infra for more information.

git-hulk:
Thanks @tisonkun.

Another solution can run those ARM64 jobs on Travis CI, but I'm wondering whether it makes sense or not.

tisonkun:
Closed as now we're using Travis to achieve multiple platform supports. If GitHub Actions support ARM platform later, we can review this proposal.

PR #626, opened 2022-06-08
CMake: support ninja as a build backend

### LINKED ISSUE #621: Enhancement: support Ninja as a CMake backend
### Search before asking

### Motivation

CMake support Ninja (`-G Ninja`) as a build tools to generate, just like Makefiles, which provides more efficient build processing.
But there are some problem in current CMake configurations in kvrocks so we cannot use it.

I will investigate it and make a fix.

### Solution

_No response_

PragmaTwice:
Close #621.

Benchmark between make and ninja build for kvrocks can be found in #625.
TL;DR: The benchmark shows that ninja is slightly slower than make build in clean build, but has huge advantages in incremental build, which is very friendly to kvrocks developers who need to do frequent incremental builds.

So in this PR, kvrocks still use make by default, but the problem that blocks ninja backend to successfully build in cmake is fixed and now `build.sh` has a `--ninja` option to enable ninja.

### DIFF
--- a/.github/workflows/daily-ci.yaml
+++ b/.github/workflows/daily-ci.yaml
@@ -127,3 +127,23 @@ jobs:
       - name: Redis Tcl Test
         run:
           cd tests/tcl && sh runtest && cd -
+
+  build-on-ubuntu-with-ninja:
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+    
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout Code Base
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 64
+
+      - name: Build
+        run: |
+          ./build.sh build --ninja
+        
+      - name: Unit Test
+        run: |
+          ./build/unittest
--- a/README.md
+++ b/README.md
@@ -76,9 +76,8 @@ $ ./build/kvrocks -c kvrocks.conf
 ### Running test cases
 
 ```shell
-$ # make sure CMake was executed in ./build
+$ ./build.sh build --unittest
 $ cd build
-$ make unittest
 $ ./unittest
 ```
 
--- a/build.sh
+++ b/build.sh
@@ -25,6 +25,7 @@ function usage() {
     echo "-Dvar=value : extra cmake definitions" >&2
     echo "-jN         : execute N build jobs concurrently, default N = 4" >&2
     echo "--unittest  : build unittest target" >&2
+    echo "--ninja     : use ninja to build kvrocks" >&2
     echo "-h, --help  : print this help messages" >&2
     exit 1
 }
@@ -33,6 +34,7 @@ until [ $# -eq 0 ]; do
     case $1 in
         -D*) CMAKE_DEFS="$CMAKE_DEFS $1";;
         --unittest) BUILD_UNITTEST=1;;
+        --ninja) USE_NINJA="-G Ninja";;
         -j*) JOB_CMD=$1;;
         -*) usage;;
         *) BUILD_DIR=$1;;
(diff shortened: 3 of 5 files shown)


git-hulk (dismissed):
LGTM, nice.

tisonkun (commented):
Shall we add a matrix to test build with ninja in CI workflow?

PragmaTwice:
> Shall we add a matrix to test build with ninja in CI workflow?

I think we could add a ninja job to the daily ci now, but since the workflows haven't been refactored, it might not be easy to use an unified matrix, which is described in #572, and I will make a PR to refactor workflows later.

tisonkun (approved):

git-hulk (approved):

tisonkun:
Thanks for your contribution!

ShooterIT:
Hi @PragmaTwice Currently daily CI is failed, please have a look,
one tips for debugging daily CI on github, you can remove the rule of daily https://github.com/apache/incubator-kvrocks/blob/unstable/.github/workflows/daily-ci.yaml#L21 in you branch, you can submit a PR after there is no error report.

tisonkun:
Actually I think whether we should running a daily CI or just run it every PR - with Apache INFRA we may have enough resource to run CI now. At least we may try to trigger daily CI when related files touched (however, "related files" can be any source files and thus it becomes a CI runs for every PR).

ShooterIT:
Hi @tisonkun it will cost much time if we run also daily CI for every PR, i think it is not friendly for contributors.

> At least we may try to trigger daily CI when related files touched (however, "related files" can be any source files and thus it becomes a CI runs for every PR).

This idea sounds good, could we support this feature?

tisonkun:
At least we can add:

```yml
name: Daily CI
on:
  pull_request:
    paths:
      - ".github/workflows/daily-ci.yaml"
  schedule:
    - cron: '0 1 * * *'
  workflow_dispatch:
```

If I find can find some spare time I'll make some investigation. But feel free to do it yourself.

PragmaTwice:
> Actually I think whether we should running a daily CI or just run it every PR - with Apache INFRA we may have enough resource to run CI now. At least we may try to trigger daily CI when related files touched (however, "related files" can be any source files and thus it becomes a CI runs for every PR).

I think we could use `ccache` and `actions/cache` to accelerate the build procedure in CI, so that more jobs could be added to the workflows triggered by PRs.


> Hi @PragmaTwice Currently daily CI is failed, please have a look, one tips for debugging daily CI on github, you can remove the rule of daily https://github.com/apache/incubator-kvrocks/blob/unstable/.github/workflows/daily-ci.yaml#L21 in you branch, you can submit a PR after there is no error report.

Sorry for that, seems there is no pre-installed ninja in the virtual environment, some `apt install` is needed.

ShooterIT:
> ```yaml
> name: Daily CI
> on:
>   pull_request:
>     paths:
>       - ".github/workflows/daily-ci.yaml"
>   schedule:
>     - cron: '0 1 * * *'
>   workflow_dispatch:
> ```

thanks @tisonkun cool, i am not familiar CI, but it seems it is a light and useful way to resolve our problem.

@PragmaTwice please don't say sorry, you already have done much for Kvrocks. Just let's fix it, we can adopt @tisonkun advise as above to check.

PS: In redis repo, maintainers can trigger manually daily CI, maybe it also is a way for us.

PragmaTwice:
> In redis repo, maintainers can trigger manually daily CI, maybe it also is a way for us.

We can also trigger the daily CI manually since `workflow_dispatch` is added to https://github.com/apache/incubator-kvrocks/blob/unstable/.github/workflows/daily-ci.yaml#L24.

Just like this, click the `Run workflow` button from a specific workflow page:
image

ShooterIT:
found, cool, thanks @PragmaTwice

PR #229, opened 2021-04-22
Fix rocksdb can't auto resume after no space error

karelrooted:
Fix can't auto resume after no space error when the no space error is only trigger by db_->Write without any other background action (compact/flush)

PR #628, opened 2022-06-09
Fix RocksDB can't auto resume after disk quota exceeded error

### LINKED ISSUE #627: RocksDB can't auto resume after disk quota exceeded error
### Search before asking

### Version

Kvrocks: 2.0.6

### Minimal reproduce step

1. limit the disk of kvrocks data-dir
2. create some big files when rocksdb compaction, make these files fill up the disk
3. if you keep writing to kvrocks, you'll find that you can't write to it
4. delete the useless files created in step 2 and find that kvrocks still cannot write

When disk space is available, kvrocks can write data normally.

### What did you see instead?

The kvrocks log keeps appearing: 
```
[event_listener.cc:93] [event_listener/background_error] reason: writecallback, status: IO error: While appending to file: /data/db/xxxx.sst: Disk quota exceeded
```
Kvrocks has been unable to write data unless it is restarted. 

### Anything Else?

_No response_

caipengbo:
Fixed: #627

In #229, the issue where RocksDB could not recover from the no Space background error was fixed. This problem RocksDB at https://github.com/facebook/rocksdb/pull/8376 has been repaired, but the issue has not been thoroughly solved, The same problem will still occur when an `EDQUOT Disk Quota Exceeded` error is encountered (see the detailed in https://github.com/facebook/rocksdb/issues/10134).

RocksDB cannot recover from this problem and must be restarted. This problem is more likely to occur when kvrocks is deployed in container.

In order to handle all versions of RocksDB, we manually resume DB when we encounter two retryable io errors: [No space left on device and Disk Quota Exceeded](https://man7.org/linux/man-pages/man3/errno.3.html).

For the Disk Quota Exceeded error, RocksDB did not expose a friendly interface, so I did a string match.

### DIFF
--- a/src/event_listener.cc
+++ b/src/event_listener.cc
@@ -59,6 +59,18 @@ const std::string compressType2String(const rocksdb::CompressionType type) {
   return iter->second;
 }
 
+bool isDiskQuotaExceeded(const rocksdb::Status &bg_error) {
+    // EDQUOT: Disk quota exceeded (POSIX.1-2001)
+    std::string edquot_str = "Disk quota exceeded";
+    std::string err_msg = bg_error.ToString();
+
+    if (err_msg.length() < edquot_str.length()) {
+        return false;
+    }
+
+    return edquot_str == err_msg.substr(err_msg.size()-edquot_str.size());
+}
+
 void EventListener::OnCompactionCompleted(rocksdb::DB *db, const rocksdb::CompactionJobInfo &ci) {
   LOG(INFO) << "[event_listener/compaction_completed] column family: " << ci.cf_name
             << ", compaction reason: " << static_cast<int>(ci.compaction_reason)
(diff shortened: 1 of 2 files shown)


src/event_listener.cc
```diff
@@ -59,6 +59,18 @@ const std::string compressType2String(const rocksdb::CompressionType type) {
   return iter->second;
 }
 
+bool isDiskQuotaExceeded(const rocksdb::Status &bg_error) {
+    // EDQUOT: Disk quota exceeded (POSIX.1-2001)
+    std::string edquot_str = "Disk quota exceeded";
```
git-hulk: edquot_str  => exceeded_quota_str

src/event_listener.cc
```diff
@@ -59,6 +59,18 @@ const std::string compressType2String(const rocksdb::CompressionType type) {
...
+    std::string err_msg = bg_error.ToString();
+
+    if (err_msg.length() < edquot_str.length()) {
+        return false;
+    }
+
+    return edquot_str == err_msg.substr(err_msg.size()-edquot_str.size());
```
git-hulk: can use string find API?

caipengbo: Yes, we can use find. The reason I use tail matching is to know that it can only be at the tail, which is probably a little bit faster than find.
It doesn't seem necessary. I will use find API.

git-hulk: Yes, background error wasn't the hot path, so it will be good to make it more readable.

src/event_listener.cc
```diff
@@ -59,6 +59,15 @@ const std::string compressType2String(const rocksdb::CompressionType type) {
   return iter->second;
 }
 
+bool isDiskQuotaExceeded(const rocksdb::Status &bg_error) {
+    // EDQUOT: Disk quota exceeded (POSIX.1-2001)
+    std::string exceeded_quota_str = "Disk quota exceeded";
+    std::string err_msg = bg_error.ToString();
+
+    std::size_t found = err_msg.find(exceeded_quota_str);
+    return found != std::string::npos;
```
git-hulk:
```suggestion
    return err_msg.find(exceeded_quota_str) != std::string::npos;

```

src/event_listener.cc
```diff
@@ -59,6 +59,15 @@ const std::string compressType2String(const rocksdb::CompressionType type) {
   return iter->second;
 }
 
+bool isDiskQuotaExceeded(const rocksdb::Status &bg_error) {
+    // EDQUOT: Disk quota exceeded (POSIX.1-2001)
+    std::string exceeded_quota_str = "Disk quota exceeded";
+    std::string err_msg = bg_error.ToString();
+
+    return err_msg.find(exceeded_quota_str) != std::string::npos;
+
```
git-hulk:
```suggestion
    return err_msg.find(exceeded_quota_str) != std::string::npos;
```

git-hulk: @caipengbo my bad, didn't remove the last blank line.

ShooterIT:
```suggestion
```

ShooterIT (dismissed):

ShooterIT:
LGTM, it seems, in cloud environments, we may encounter this error.

git-hulk (approved):

PR #634, opened 2022-06-16
typo: fix typo in README

jackwener:


jackwener:
@PragmaTwice PTAL

README.md
```diff
@@ -178,6 +178,6 @@ latency: 99.9% < 10ms
 
 Kvrocks is under the Apache License Version 2.0. See the LICENSE file for details.
 
-## 微信公众号
+## WeChat offical account
 
 <img src="docs/images/wechat_account.jpg" alt="微信公众号" />
```
PragmaTwice:
```suggestion
<img src="docs/images/wechat_account.jpg" alt="WeChat offical account" />
```

PragmaTwice (dismissed):

README.md
```diff
@@ -85,7 +85,7 @@ $ ./unittest
 
 * centos 6/7
 * ubuntu
-* macosx
+* macos
```
tisonkun: I don't think this is a typo. At least you can add reference about the usage clarification.

README.md
```diff
@@ -178,6 +178,6 @@ latency: 99.9% < 10ms
 
 Kvrocks is under the Apache License Version 2.0. See the LICENSE file for details.
 
-## 微信公众号
```
tisonkun, line 181: This information is mainly for Chinese users and thus I don't think it has to be changed.

tisonkun (changes requested):
Changes are not all about "typo". Comments inline.

README.md
PragmaTwice: I think it's OK to be something like `WeChat Official Account (微信公众号)`, since users who do not understand Chinese may be confused by a Chinese title and a qrcode.

And I actually feel that we can add platforms such as zhihu [1] and medium [2] that have kvrocks accounts somewhere, rather than just having a WeChat official account.

[1] https://www.zhihu.com/people/kvrocks
[2] https://kvrocks.medium.com/

jackwener: agree with @PragmaTwice

README.md
jackwener: we can see the history in [wikipedia](https://en.wikipedia.org/wiki/MacOS#History)

tisonkun: Then it should be:

```markdown
* CentOS 6 or 7
* Ubuntu
* macOS
```

README.md
```diff
@@ -55,7 +55,7 @@ sudo yum install -y epel-release && sudo yum install -y git gcc gcc-c++ make cma
 sudo apt update
 sudo apt-get install gcc g++ make cmake autoconf automake libtool
 
-# MACOSX
+# MacOS
```
ShooterIT: Why not `macOS`

jackwener: The standard name is `macOS`😂

git-hulk: It will be good to use `macOS` since apple/wikipedia also using this name and find nowhere using MacOS.

ShooterIT (approved):

tisonkun (approved):

git-hulk (approved):

tisonkun:
@jackwener thanks for your contribution!

PR #638, opened 2022-06-17
bump .asf.yaml

tisonkun:
INFRA has configured `unstable` as the default branch, we need a new commit to trigger reload configuration.

See also https://issues.apache.org/jira/browse/INFRA-23387


PragmaTwice:
I noticed that `Required` in this configuration doesn't seem to work, any ideas?

PragmaTwice (approved):
LGTM

tisonkun:
@PragmaTwice I guess it will take effect after this patch merged and file changed. If it still doesn't work, we should report to INFRA :)

PR #642, opened 2022-06-20
Fix load namespaces from the config file would ignore case-sensitive

### LINKED ISSUE #640: namespace  is case-sensitive but namespace configed in file is not and namespace command does
### Search before asking

### Version

1.3.3

### Minimal reproduce step

add namespace with command 

>> namespace add "Big_case_namespace" "anypass"
ok

then 
restart the instance ,get namespace info
>> namespace get *
1) big_case_namespace
2) anypass

when restart we should get a namespace named "Big_case_namespace" not "big_case_namespace", it makes data could not be access with auth "anypass",it point to another namespace with lowercase

### What did you see instead?

>> namespace get *
1) big_case_namespace
2) anypass

### Anything Else?

config with namespace and command has the same result

fishery86:
[fix]:load namespace from config file with case-sensitive issue:#640

This closes #640

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -523,16 +523,18 @@ Status Config::parseConfigFromString(std::string input, int line_number) {
   if (kv.size() != 2) return Status(Status::NotOK, "wrong number of arguments");
   if (kv[1] == "\"\"") return Status::OK();
 
-  kv[0] = Util::ToLower(kv[0]);
-  auto iter = fields_.find(kv[0]);
+  std::string configKey = Util::ToLower(kv[0]);
+  if (!strncasecmp(kv[0].data(), "namespace.", 10)) {
+      // namespace should keep key case-sensitive
+      configKey = kv[0];
+      tokens[kv[1]] = kv[0].substr(10, kv[0].size() - 10);
+  }
+  auto iter = fields_.find(configKey);
   if (iter != fields_.end()) {
     auto field = iter->second;
     field->line_number = line_number;
     auto s = field->Set(kv[1]);
     if (!s.IsOK()) return s;
-  }
-  if (!strncasecmp(kv[0].data(), "namespace.", 10)) {
-    tokens[kv[1]] = kv[0].substr(10, kv[0].size()-10);
   }
   return Status::OK();
 }
--- a/src/config.cc
+++ b/src/config.cc
@@ -524,10 +524,12 @@ Status Config::parseConfigFromString(std::string input, int line_number) {
   if (kv[1] == "\"\"") return Status::OK();
 
   std::string configKey = Util::ToLower(kv[0]);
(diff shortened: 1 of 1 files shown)


src/config.cc
```diff
@@ -523,17 +523,19 @@ Status Config::parseConfigFromString(std::string input, int line_number) {
...
   if (kv[1] == "\"\"") return Status::OK();
 
-  kv[0] = Util::ToLower(kv[0]);
-  auto iter = fields_.find(kv[0]);
+  std::string configKey = Util::ToLower(kv[0]);
+  if (!strncasecmp(kv[0].data(), "namespace.", 10)) {
+      // namespace should keep key case-sensitive
+      configKey = kv[0];
+      tokens[kv[1]] = kv[0].substr(10, kv[0].size() - 10);
+  }
```
PragmaTwice:
```suggestion
  const char ns_str[] = "namespace.";
  size_t ns_str_size = sizeof(ns_str) - 1;
  if (!strncasecmp(kv[0].data(), ns_str, ns_str_size)) {
      // namespace should keep key case-sensitive
      configKey = kv[0];
      tokens[kv[1]] = kv[0].substr(ns_str_size);
  }
```
I think we should avoid any magic number, e.g. `10` here : )

src/config.cc
```diff
@@ -523,17 +523,21 @@ Status Config::parseConfigFromString(std::string input, int line_number) {
...
-  auto iter = fields_.find(kv[0]);
+  std::string configKey = Util::ToLower(kv[0]);
+  const char ns_str[] = "namespace.";
+  size_t ns_str_size = sizeof(ns_str) - 1;
+  if (!strncasecmp(kv[0].data(), ns_str, ns_str_size)) {
+      // namespace should keep key case-sensitive
+      configKey = kv[0];
+      tokens[kv[1]] = kv[0].substr(ns_str_size);
+  }
+  auto iter = fields_.find(configKey);
```
```suggestion
  std::string field_key = Util::ToLower(kv[0]);
  const char ns_str[] = "namespace.";
  size_t ns_str_size = sizeof(ns_str) - 1;
  if (!strncasecmp(kv[0].data(), ns_str, ns_str_size)) {
      // namespace should keep key case-sensitive
      field_key = kv[0];
      tokens[kv[1]] = kv[0].substr(ns_str_size);
  }
  auto iter = fields_.find(field_key);
```

git-hulk: The variable name should be spliced by `_` instead of CamelCase. Sorry that we didn't claim the code style clearly in our contribution guide.

git-hulk (approved):
LGTM, thanks @fishery86

PragmaTwice (approved):
LGTM

git-hulk:
Thanks for your contribution again @fishery86

PR #643, opened 2022-06-21
Optimize happy path for constructing Status

PragmaTwice:
In kvrocks, `Status` is used to indicate the state of the operation when it is completed.

Obviously, these operation will complete normally (often called a happy path) in most cases (or say, with high probability), and fail in a small number of cases.

Field `msg_` typed `std::string` will be assigned to `ok` when constructing the normal state, which can be more time consuming than the default constructor (although usually `std::string` implementations in the standard library do SSO (small string optimization), thus heap allocation may be avoided).

In this PR, I delayed the `ok` to message dumping to speed up the happy path.

### DIFF
--- a/src/status.h
+++ b/src/status.h
@@ -57,13 +57,19 @@ class Status {
     NetSendErr,
   };
 
-  Status() : Status(cOK, "ok") {}
-  explicit Status(Code code, std::string msg = "") : code_(code), msg_(std::move(msg)) {}
+  Status() : Status(cOK) {}
+  explicit Status(Code code, std::string msg = {}) : code_(code), msg_(std::move(msg)) {}
   bool IsOK() { return code_ == cOK; }
   bool IsNotFound() { return code_ == NotFound; }
   bool IsImorting() { return code_ == SlotImport; }
-  std::string Msg() { return msg_; }
-  static Status OK() { return Status(cOK, "ok"); }
+  std::string Msg() {
+    if(!IsOK())
+      return msg_;
+    else {
+      return "ok";
+    }
+  }
+  static Status OK() { return {}; }
 
  private:
   Code code_;
--- a/src/status.h
+++ b/src/status.h
@@ -63,11 +63,10 @@ class Status {
   bool IsNotFound() { return code_ == NotFound; }
   bool IsImorting() { return code_ == SlotImport; }
   std::string Msg() {
-    if(!IsOK())
-      return msg_;
-    else {
+    if(IsOK()) {
       return "ok";
     }
+    return msg_;
   }
   static Status OK() { return {}; }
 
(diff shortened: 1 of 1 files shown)


git-hulk (approved):
LGTM

src/status.h
```diff
@@ -57,13 +57,18 @@ class Status {
     NetSendErr,
   };
 
-  Status() : Status(cOK, "ok") {}
-  explicit Status(Code code, std::string msg = "") : code_(code), msg_(std::move(msg)) {}
+  Status() : Status(cOK) {}
```
tisonkun, line 60: @ShooterIT @git-hulk may I ask why `cOK` gets its name at the first place instead of `OK`?

src/status.h
```diff
@@ -57,13 +57,18 @@ class Status {
...
+  explicit Status(Code code, std::string msg = {}) : code_(code), msg_(std::move(msg)) {}
   bool IsOK() { return code_ == cOK; }
   bool IsNotFound() { return code_ == NotFound; }
   bool IsImorting() { return code_ == SlotImport; }
```
tisonkun, line 64:
```suggestion
  bool IsImporting() { return code_ == SlotImport; }
```

Typo? May we fix it in another PR though.

src/status.h
PragmaTwice: because we already have a method named `OK` 🤣

src/status.h
```diff
@@ -57,13 +57,18 @@ class Status {
...
-  std::string Msg() { return msg_; }
-  static Status OK() { return Status(cOK, "ok"); }
+  std::string Msg() {
+    if (IsOK()) {
+      return "ok";
+    }
+    return msg_;
```
tisonkun, line 69: Will this change cause branch prediction overhead and affect efficiency which we'd like to improve in this PR? It seems not a net win.

PragmaTwice: The function `Msg` only will be called in a error path (in most cases), not normal path, so the probability of calling is low. 
This fits perfectly with my idea, so I think something like `__builtin_expect` may be unnecessary : )

A classic pattern for `Msg`:
```c++
Status s = ...
if(!s.OK()) {
  dump s.Msg() to somewhere...
}
```

src/status.h
git-hulk: > because we already have a method named `OK` 🤣

Yes, exactly. But `cOK` is NOT a good name since it's inconsistent with other code.

src/status.h
git-hulk: it's typo

tisonkun (approved):
LGTM.

src/status.h
PragmaTwice: Maybe open a help wanted issue to encourage more involvement🤣 
Not sure whether it is a proper way, calling expert @tisonkun

git-hulk: 🤣

tisonkun: There are several typos in our codebase. I don't think they deserve an issue but if you'd like to fix it with another patch, go ahead. Otherwise I may handle it later this week =。=

PragmaTwice: Got it. Thanks.

Issue #581, opened 2022-05-15
Proposal: Just return values instead of passing pointers if possible


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

C++ specifies a number of strategies in the standard to eliminate unnecessary copies, which are called [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision).

One of these optimizations is called *named return value optimization* (NRVO), which is defined in the standard. 
This optimization works on the return value, in short, the returned object is not copied onto the current stack.

GCC and Clang both perform this optimization by default (even in the `-O0` case). 
The following simple code demonstrates it:

```c++
T f() {
    T a;
    ... // operations on a
    return a;
}
 
T a = f(); // T::T(const T&) or T::T(T&&) will not be called
```

You can view the assembly or modify and execute this code online on [godbolt](https://godbolt.org/z/xcYK31nGY).

So passing pointers to optimize efficiency maybe not necessary, and returning values is usually easier to make the operation be expressed in the same expression and more intuitive. Some parts of this project use this approach, like [these functions](https://github.com/apache/incubator-kvrocks/blob/unstable/src/util.h#L79-L81).

So I think it would be better to change them to return the results directly.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

git-hulk:
Cool, thanks @PragmaTwice. We use many C style passing pointers in function parameter,
it's good to me since it won't affect readable.

tisonkun:
Closed as no more insights here. If you find more patterns to adopt, feel free to start a new initiative.

PR #644, opened 2022-06-21
Refactor Trim/Split/Split2KV functions in util.h

PragmaTwice:
- Solve #581
- Refactor Split and Split2KV
- Optimize some signature to const reference
- Add some unit test for string utils

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -550,29 +550,25 @@ std::string StringToHex(const std::string &input) {
   return output;
 }
 
-constexpr std::uint64_t ExpTo1024(size_t n) {
-  return static_cast<std::uint64_t>(1) << (10 * n);
-}
-
 void BytesToHuman(char *buf, size_t size, uint64_t n) {
   double d;
 
-  if (n < ExpTo1024(1)) {
+  if (n < 1024) {
     snprintf(buf, size, "%" PRIu64 "B", n);
-  } else if (n < ExpTo1024(2)) {
-    d = static_cast<double>(n)/ExpTo1024(1);
+  } else if (n < (1024*1024)) {
+    d = static_cast<double>(n)/(1024);
     snprintf(buf, size, "%.2fK", d);
-  } else if (n < ExpTo1024(3)) {
-    d = static_cast<double>(n)/ExpTo1024(2);
+  } else if (n < (1024LL*1024*1024)) {
+    d = static_cast<double>(n)/(1024*1024);
     snprintf(buf, size, "%.2fM", d);
-  } else if (n < ExpTo1024(4)) {
-    d = static_cast<double>(n)/ExpTo1024(3);
+  } else if (n < (1024LL*1024*1024*1024)) {
+    d = static_cast<double>(n)/(1024LL*1024*1024);
     snprintf(buf, size, "%.2fG", d);
-  } else if (n < ExpTo1024(5)) {
-    d = static_cast<double>(n)/ExpTo1024(4);
+  } else if (n < (1024LL*1024*1024*1024*1024)) {
+    d = static_cast<double>(n)/(1024LL*1024*1024*1024);
     snprintf(buf, size, "%.2fT", d);
-  } else if (n < ExpTo1024(6)) {
-    d = static_cast<double>(n)/ExpTo1024(5);
+  } else if (n < (1024LL*1024*1024*1024*1024*1024)) {
+    d = static_cast<double>(n)/(1024LL*1024*1024*1024*1024);
     snprintf(buf, size, "%.2fP", d);
   } else {
     snprintf(buf, size, "%" PRIu64 "B", n);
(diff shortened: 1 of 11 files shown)


tisonkun (commented):
Generally looks good. One comment:

It seems that `Trim` will update the input string while `Split` doesn't. This looks inconsistent.

Also, if `Split` with empty delim, we may return the input string as is.

PragmaTwice:
Thanks @tisonkun !

> `Trim` will update the input string while `Split` doesn't

`Trim` takes a copy of the first argument, not a reference.

> if `Split` with empty delim, we may return the input string as is.

done~

src/util.cc
```diff
@@ -371,7 +371,14 @@ std::string Trim(std::string in, const std::string &chars) {
 std::vector<std::string> Split(const std::string &in, const std::string &delim) {
   std::vector<std::string> out;
 
-  if (in.empty() || delim.empty()) return out;
+  if (in.empty()) {
+    return out;
+  }
+
+  if (delim.empty()) {
```
tisonkun, line 378: Sorry. After take a look at Java's `Spring.split` and C++20's [`split_view`](https://en.cppreference.com/w/cpp/ranges/split_view), its seems empty delim will cause output a string vector contains every character as a string member.

```java
jshell> "123".split("")
$1 ==> String[3] { "1", "2", "3" }

jshell> "1 2 3".split("")
$2 ==> String[5] { "1", " ", "2", " ", "3" }
```

```cpp
#include <iostream>
#include <iomanip>
#include <ranges>
#include <string_view>
 
int main() {
    constexpr std::string_view words{"Hello-_-C++-_-20-_-!"};
    constexpr std::string_view delim{""};
    for (const auto word : std::views::split(words, delim)) {
        std::cout << std::quoted(std::string_view(word.begin(), word.end())) << ' ';
    }
}
// "H" "e" "l" "l" "o" "-" "_" "-" "C" "+" "+" "-" "_" "-" "2" "0" "-" "_" "-" "!" 
// https://wandbox.org/permlink/5BZQHqUYd36t0WbQ
```

PragmaTwice: So thanks, I'll fix it.
BTW, this function has some difference than some well known `split`: it split strings regarding to regex "a|b|c|..." where delim="abc...", so I am considering to add another split that split strings just regarding to the provided delimiter.

tisonkun: Yes. We don't have to follow other system's design, but we should make our own utils semantic clear.

PragmaTwice: done

src/util.cc
```diff
@@ -359,42 +359,54 @@ std::string ToLower(std::string in) {
...
+  size_t begin = 0, end = in.find_first_of(delim);
   do {
-    pos = in.find_first_of(delim);
-    elem = in.substr(0, pos);
-    Trim(elem, delim, &trimed_elem);
-    if (!trimed_elem.empty()) out->push_back(trimed_elem);
-    in = in.substr(pos+1);
-  } while (pos != std::string::npos);
+    std::string elem = in.substr(begin, end - begin);
+    if (!elem.empty()) out.push_back(std::move(elem));
```
tisonkun, line 388: `Split("1  2", " ")` should normally return `{"1", "", "2"}`. But with this condition, it returns `{"1", "2"}`.

I don't know if kvrocks depends on this behavior but point out it's diverged from other libraries.

PragmaTwice: I think it seems more reasonable if `Split("1 2", " ")` returns `{"1", "2"}`, and it is same as previous behavior.
And `Split("1 2","")` returns `{"1", " ", "2"}`.

tisonkun (dismissed):

src/util.cc
git-hulk: Yes, this `split` function has a bit different from most languages
since we want to ignore the empty value.

git-hulk (dismissed):
Cool, this makes a lot sense.

src/util.cc
```diff
@@ -538,25 +550,29 @@ std::string StringToHex(const std::string &input) {
...
 
+constexpr std::uint64_t ExpTo1024(size_t n) {
+  return static_cast<std::uint64_t>(1) << (10 * n);
+}
+
 void BytesToHuman(char *buf, size_t size, uint64_t n) {
   double d;
 
-  if (n < 1024) {
+  if (n < ExpTo1024(1)) {
```
ShooterIT: actually this code is copied from redis.
but for this change, i don't feel it is clear, maybe we can define `KB MB GB TB PB`.

And this change is not related to theme.

PragmaTwice: It just replace `1024*1024*...*1024` to something like `1024^n`, so I don't think it will affect readability. And it is not related to `bytes`.

> And this change is not related to theme

Agree. Maybe move it to a separate PR.

PragmaTwice: These changes are reverted, I'll prepare a seperate PR to illustrate them.

tisonkun (approved):

ShooterIT (approved):

git-hulk (approved):

PragmaTwice:
Thanks everyone! Merging...

PR #646, opened 2022-06-21
Fix unmatched iterator in `Cluster::SetClusterNodes`

PragmaTwice:


git-hulk (approved):
LGTM, good catch!

ShooterIT:
Hi @PragmaTwice is there a bug? what is its danger?

PragmaTwice:
> Hi @PragmaTwice is there a bug? what is its danger?

Yeah, it is an undefined behavior (to compare iterators from different containers), so the compiler and the standard library implementation can literally do anything, e.g. either return true or false, or even spend $1 from your mastercard 🤣  althought maybe libstdc++ and libc++ will not do that.
In some standard library implementation the `end()` iterator of a hashmap is just a `nullptr`, so the compare work well, but I means that I think we should not rely on a specific implementation and an undefined behavior.

git-hulk:
@ShooterIT I think we can disable the test cases which failed frequently before finding the root cause, how do you think?

tisonkun:
@git-hulk if so, we should hold a pinned issue to investigate and move forward fixing them.

git-hulk:
> @git-hulk if so, we should hold a pinned issue to investigate and move forward fixing them.

OK, I'll submit an issue to find out those test cases, then submit another issue to track them.

ShooterIT (approved):
my bad 🤦‍♂️ different variables

tisonkun (approved):

PragmaTwice:
Thanks everyone. Merging...

PR #649, opened 2022-06-22
Set CMAKE_EXPORT_COMPILE_COMMANDS to ON as a default behavior

PragmaTwice:
Both clangd, ccls, clang-tidy and other C++ language tools (like cppcheck) need compile_commands.json (a database for compile commands) to retrieve compile information like include dirs, standard, compile options, and more.
CMake provide an option CMAKE_EXPORT_COMPILE_COMMANDS to generate the file.
So setting it to ON as a default behavior can make these tools more easier to use.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,8 @@ if(ENABLE_ASAN AND ENABLE_TSAN)
     message(FATAL_ERROR "ASan and TSan cannot be used at the same time")
 endif()
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 # GLIBC < 2.17 should explict specify the real time library when use clock_*
 find_library(REALTIME_LIB rt)
 if (REALTIME_LIB)


tisonkun (commented):
Could you describe a concrete use case that could benefits from this change? It seems we don't even configure `CMAKE_EXPORT_COMPILE_COMMANDS` for current cppcheck.

PragmaTwice:
> Could you describe a concrete use case that could benefits from this change? It seems we don't even configure `CMAKE_EXPORT_COMPILE_COMMANDS` for current cppcheck.

Yeah, cppcheck allow user to not provide a `compile_commands.json`, but it will greatly affect the detection accuracy, and user need to provide all include dirs, standard, compile options manually to fix this. 
(of course, cppcheck itself is a tool with the advantage of speed, it does not use some more in-depth static analysis methods, so the detection rate itself is not high)

In static analysis (clang-tidy, cppcheck, and most commercial SAST products) and language server (clangd, ccls) of C++, `compile_commands.json` is just a standard.
Many CMake C++ project like apache/doris, oneflow, clickhouse did something like this PR.

And without this file our c++ language server like clangd or ccls can barely work, e.g. lots of missing header, missing macro definition, mismatch standard, mismatch compile options that can affect semantics... 🤣 (clion maybe generate this file automatically since clion use clangd (and a self-maintained engine) to analyze source code)

git-hulk (approved):
LGTM

tisonkun (approved):

PR #652, opened 2022-06-23
Fix Wrongly parsed the RESP empty/null array

### LINKED ISSUE #633: Wrongly parsed the RESP empty/null array
### Search before asking

### Version

b1eb17f0228b40244694b308639a8876c90444ba

### Minimal reproduce step

According to https://redis.io/docs/reference/protocol-spec/#resp-arrays, 
empty/null array is allowed in RESP but Kvrocks got the protocol error.

```shell
# empty array
$ echo '*0\r\n*2\r\n$3\r\nget\r\n$3\r\nkey\r\n'|nc 127.0.0.1 6666
# null array
$ echo '*-1\r\n*2\r\n$3\r\nget\r\n$3\r\nkey\r\n'|nc 127.0.0.1 6666
```

Kvrocks can parse those protocol correctly and response the right value of the `key`
like Redis behavior:
```shell
❯ echo '*0\r\n*2\r\n$3\r\nget\r\n$3\r\nkey\r\n'|nc 127.0.0.1 6379                                                                                                  
$-1

❯ echo '*-1\r\n*2\r\n$3\r\nget\r\n$3\r\nkey\r\n'|nc 127.0.0.1 6379                                                                                                 
$-1
```

### What did you see instead?

```shell
❯ echo '*0\r\n*2\r\n$3\r\nget\r\n$3\r\nkey\r\n'|nc 127.0.0.1 6666                                                                                                  
-Protocol error: expected '$'

❯ echo '*-1\r\n*2\r\n$3\r\nget\r\n$3\r\nkey\r\n'|nc 127.0.0.1 6666                                                                                                 
-Protocol error: invalid multibulk length
```

### Anything Else?

For the empty array case, Kvrocks should NOT enter the next state when the multi bulk length is 0 at line:[redis_request.cc#L71](https://github.com/apache/incubator-kvrocks/blob/unstable/src/redis_request.cc#L71)

For the null array case, Kvrocks should use `stoll` to parse the number instead of `stoull` since the length maybe negative at line: [redis_request.cc#L62](https://github.com/apache/incubator-kvrocks/blob/unstable/src/redis_request.cc#L62). Also, we need to skip the null array like the empty array case.

willshS:
This closes #633.

**Now** : I only fix like:
`echo -e '*-1\r\n*2\r\n$3\r\nget\r\n$3\r\nkey\r\n' | nc 127.0.0.1 6666` 
`echo -e '*2\r\n$3\r\nget\r\n$3\r\nkey\r\n*0\r\n*2\r\n$3\r\nget\r\n$3\r\nkey\r\n' | nc 127.0.0.1 6666` 

Redis can have **"nil"** value. like:

image

kvrocks also behaves the same way.(I didn't modify anything about "nil" value)


git-hulk:
Cool, can you add some test cases in [protocol.tcl](https://github.com/apache/incubator-kvrocks/blob/unstable/tests/tcl/tests/unit/protocol.tcl) to make sure it has already fixed?

PragmaTwice (commented):
Agree with @git-hulk, the rest looks good. Thanks for contribution.

src/redis_request.cc
```diff
@@ -59,11 +59,15 @@ Status Request::Tokenize(evbuffer *input) {
         svr_->stats_.IncrInbondBytes(len);
         if (line[0] == '*') {
           try {
-            multi_bulk_len_ = std::stoull(std::string(line + 1, len-1));
+            multi_bulk_len_ = std::stoll(std::string(line + 1, len-1));
           } catch (std::exception &e) {
             free(line);
             return Status(Status::NotOK, "Protocol error: invalid multibulk length");
           }
+          if (multi_bulk_len_ <= 0) {
```
PragmaTwice, line 67: Is an assignment `multi_bulk_len_ = 0` needed while the condition `multi_bulk_len_ < 0` holds?

PragmaTwice: There is a condition `if (multi_bulk_len_ == 0)` in `src/redis_request.cc:111`

willshS: `state_ `is still `ArrayLen`, so it will read again and assign to `multi_bulk_len_`.   it's a good habit that to add the assignment `multi_bulk_len_ = 0`. Thank you

willshS:
> Cool, can you add some test cases in [protocol.tcl](https://github.com/apache/incubator-kvrocks/blob/unstable/tests/tcl/tests/unit/protocol.tcl) to make sure it has already fixed?

sure, that's what i should do.

git-hulk:
> > Cool, can you add some test cases in [protocol.tcl](https://github.com/apache/incubator-kvrocks/blob/unstable/tests/tcl/tests/unit/protocol.tcl) to make sure it has already fixed?
> 
> sure, that's what i should do.

Cool, thanks @willshS

src/redis_request.cc
```diff
@@ -35,7 +35,7 @@
 namespace Redis {
 const size_t PROTO_INLINE_MAX_SIZE = 16 * 1024L;
 const size_t PROTO_BULK_MAX_SIZE = 512 * 1024L * 1024L;
-const size_t PROTO_MULTI_MAX_SIZE = 1024 * 1024L;
+const int64_t PROTO_MULTI_MAX_SIZE = 1024 * 1024L;
```
tisonkun: Why do you make this change?

tisonkun:
Shall this patch fix #633?

git-hulk:
> Shall this patch fix #633?

yes

src/redis_request.cc
willshS: compiler warning. `size_t == int64_t`

tisonkun: You may instead change `multi_bulk_len_ > PROTO_MULTI_MAX_SIZE` to `multi_bulk_len_ > (int64_t) PROTO_MULTI_MAX_SIZE`?

PragmaTwice: I also think it is better to keep `PROTO_MULTI_MAX_SIZE` typed `size_t`, since semantically it is a positive constant.

willshS: yes, you are right, it should be positive constant, thank you.

git-hulk (approved):

tisonkun (approved):

PR #660, opened 2022-06-24
fix: replication tests should wait for server restarted

### LINKED ISSUE #656: Executing test client: couldn't open socket: connection refused
Report: https://github.com/apache/incubator-kvrocks/runs/6981007645?check_suite_focus=true#step:10:703

tisonkun:
This fixes #656.

This patch is uncompleted. @git-hulk it seems `$stdout` in tcl files doesn't mean stdout from tty. Do you have any background about this test logic?

The issue we should overcome here is to determinately wait for kvrocks server up, instead of sleep 1 second which is unstable.

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -167,6 +167,8 @@ Status Server::Start() {
     }
   });
 
+  std::cout << "Ready to accept connections\n";
+
   return Status::OK();
 }
 
--- a/src/server.cc
+++ b/src/server.cc
@@ -167,7 +167,7 @@ Status Server::Start() {
     }
   });
 
-  LOG(INFO) << "Ready to accept connections\n";
+  LOG(INFO) << "Ready to accept connections";
 
   return Status::OK();
 }
--- a/src/server.cc
+++ b/src/server.cc
@@ -167,7 +167,7 @@ Status Server::Start() {
     }
   });
 
-  std::cout << "Ready to accept connections" << std::endl;
+  LOG(INFO) << "Ready to accept connections\n";
 
   return Status::OK();
 }
--- a/src/server.cc
+++ b/src/server.cc
@@ -167,7 +167,7 @@ Status Server::Start() {
     }
   });
 
-  std::cout << "Ready to accept connections\n";
+  LOG(INFO) << "Ready to accept connections\n";
 
   return Status::OK();
 }
--- a/src/server.cc
+++ b/src/server.cc
@@ -167,7 +167,7 @@ Status Server::Start() {
     }
   });
 
-  LOG(INFO) << "Ready to accept connections\n";
+  std::cout << "Ready to accept connections" << std::endl;
 
   return Status::OK();
 }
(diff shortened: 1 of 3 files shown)


src/server.cc
```diff
@@ -167,6 +167,8 @@ Status Server::Start() {
     }
   });
 
+  std::cout << "Ready to accept connections\n";
```
tisonkun: Perhaps we should send this line to log and find a way for tests reading logs.

PragmaTwice:
```suggestion
  LOG(INFO) << "Ready to accept connections";
```
Is this what you need?

git-hulk:
@tisonkun You're right, we supported to send logs to stdout recently. Need to set `log-dir` to `stdout` on https://github.com/apache/incubator-kvrocks/blob/unstable/tests/tcl/tests/assets/default.conf

src/server.cc
tisonkun: However TCL tests handle log files hacky and I tend to send this information to stdout now.

tisonkun:
> @tisonkun You're right, we supported to send logs to stdout recently. Need to set `log-dir` to `stdout` on https://github.com/apache/incubator-kvrocks/blob/unstable/tests/tcl/tests/assets/default.conf

Actually if you search `kvrocks.INFO` we hardcode several file path to the default log path..

I'm sending the "Ready to accept connections"  info to stdout for now and see whether we don't suffer from the flaky test. We may later improve stdout and logging logic.

tests/tcl/tests/support/server.tcl
```diff
@@ -403,7 +403,6 @@ proc start_server {options {code undefined}} {
 
     set unixsocket [file normalize [format "%s/%s" [dict get $config "dir"] "socket"]]
     dict set config "unixsocket" $unixsocket
-    dict set config "log-dir" [format "%s/%s" [dict get $config "dir"] "stdout"]
```
tisonkun: We can redirect log to stdout by setting "log-dir" to "stdout", but we should later update all reference hardcode to `$dir/kvrocks.INFO`

PragmaTwice:
A related problem: I think that kvrocks should not output anything on stdout if we set a log file (which is a regular file).

Currently kvrocks still outputs something to stdout when the log file is set, which is confusing (to have multiple output locations), e.g. we could output the version to log file or via `kvrocks -v`.

Of course we can consider this issue later so in this PR stdout is still a possible output location.

@git-hulk @tisonkun

tisonkun:
OK. Let me push a followup commit to switch to `LOG(INFO)` version.

PragmaTwice:
> OK. Let me push a followup commit to switch to `LOG(INFO)` version.

Just address an already-exist problem, so feel free to use stdout. 
I think we can consider this problem later, and it should not block this PR.

tisonkun:
It seems logs are fragmented among files and we should find another way to determinately confirm server has been restarted.

tisonkun:
@git-hulk @PragmaTwice I think this patch is ready to merge. You can give it a review now :)

src/server.cc
```diff
@@ -167,6 +167,8 @@ Status Server::Start() {
     }
   });
 
+  LOG(INFO) << "Ready to accept connections\n";
```
PragmaTwice:
```suggestion
  LOG(INFO) << "Ready to accept connections";
```
newline is automatically applied in glog

tisonkun: Updated.

git-hulk (approved):

PragmaTwice (approved):
LGTM

PR #661, opened 2022-06-24
Fix the role command can't be used when loading data

### LINKED ISSUE #655: Executing test client: LOADING kvrocks is restoring the db from backup
Report: https://github.com/apache/incubator-kvrocks/runs/7026847364?check_suite_focus=true#step:10:713

git-hulk:
This closes #655 

Currently, Kvrocks forbid to use the DB pointer when loading data,
so we add `ok-loading` to exclude those commands that do NOT use the DB pointer.
But we forgot to add `ok-loading` to the role command which is using by the replication test case: [tests/integration/replication.tcl#L254](https://github.com/apache/incubator-kvrocks/blob/unstable/tests/tcl/tests/integration/replication.tcl#L254), and it got error when restoring the backup.


PragmaTwice (approved):
LGTM

tisonkun (approved):

git-hulk:
Thanks all, merging.

PR #668, opened 2022-06-25
Fix get rocksdb ops tcl test case

### LINKED ISSUE #658: get rocksdb ops by COMMAND INFO in tests/unit/command.tcl [expr abs($cmd_qps - $put_qps)] < 10
Report: https://github.com/apache/incubator-kvrocks/runs/6453256361?check_suite_focus=true#step:7:2018

caipengbo:
fix #658

### DIFF
--- a/tests/tcl/tests/unit/command.tcl
+++ b/tests/tcl/tests/unit/command.tcl
@@ -56,13 +56,11 @@ start_server {tags {"command"}} {
         set seek_qps [s seek_per_sec]
         set next_qps [s next_per_sec]
 
-        # Based on the encoding of list, we can calculate the relationship
-        # between put QPS and other QPS.
-        assert {$put_qps != 0}
-        assert {[expr abs($put_qps - $get_qps)] < 10}
-        assert {[expr abs($put_qps/2 - $seek_qps)] < 10}
+        assert {$put_qps > 0}
+        assert {$get_qps > 0}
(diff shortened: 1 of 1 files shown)


git-hulk:
As commented in https://github.com/apache/incubator-kvrocks/issues/658#issuecomment-1166219636, seems our current ops comparison is not so determined since it depends on the server cron to record metrics, also ops will be changed by the time slot. I think we can change the assert condition to guarantee those ops metrics can't be zero. How do you think?

caipengbo:
> I think we can change the assert condition to guarantee those ops metrics can't be zero.

Yes, this operation is the metric from rocksdb, which should return as correct as possible. 

And how about using put ops compare to other ops?

git-hulk:
> > I think we can change the assert condition to guarantee those ops metrics can't be zero.
> 
> Yes, this operation is the metric from rocksdb, which should return as correct as possible.
> 
> And how about using put ops compare to other ops?

For those ops metrics would be affected by other test cases if they have many get/put/seek operations, coz our metrics calculate range is 16s. Can see: https://github.com/apache/incubator-kvrocks/blob/0b2de8ab6f661d90a279a0c11096755d7733b11c/src/stats.cc#L108

I think it's ok to check ops > 0 only for current implementation, we should export those metrics counter if want to check them more precisely.

git-hulk (approved):

Issue #663, opened 2022-06-24
Use `unique_ptr` to eliminate some trivial manual deallocation


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

There are many manual memory (or other system resources like file) allocation and deallocation such as `new`/`delete` in kvrocks.

Memory management is a complex topic (after RC/tracing GC is involved quickly) and there are some smart pointers like `unique_ptr`, `shared_ptr` in C++ to abstract a specific way to manage resources.

For some trivial scenario which do not require to share objects, we can do it via `unique_ptr`, which is a zero-overhead abstraction.

We can switch some manual deallocations to `unique_ptr` to make the code more maintainable and less prone to memory leaks.

NOTE: We try to use `unique_ptr` instead of creating our own RAII class types because `unique_ptr` has good semantics: it prohibits copy construction and only allows move construction, which keeps us from accidentally sharing objects and causing Use After Free or Double Free. (And we should carefully to retrieve raw pointers via `get()` by checking that the callee do not require ownership)

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

jackwener:
I'd like to do it.

PragmaTwice:
> I'd like to do it.

Thanks, feel free to submit a PR.

Since this issue may be divided into many PRs, everyone can directly submit a PR for a specific part (no need to reply this issue), so I think it would be better not to assign this issue to a specific person.

tisonkun:
@PragmaTwice is there any remaining works and do you want to continue driving this effort?

PragmaTwice:
I think most of trivial manual deallocation is removed, but there still remains some raw memory allocation and fd allocation.

tisonkun:
@PragmaTwice is there a way we can find and list them out here? Otherwise, it seems this proposal stalls.

PragmaTwice:
Closed. Will create new issue if more manual deallocation is found.

PR #671, opened 2022-06-25
enhance: use unique_ptr in Config

jackwener:
Part of #663 

In config.cc, use unique_ptr to eliminate some trivial manual deallocation


jackwener:
I have use `emplace_back` place the vector [List-initialization](https://en.cppreference.com/w/cpp/language/list_initialization).

Because it will use `copy constructor`, it will call `FieldWrapper(const FieldWrapper &wrapper)`.

So, I use `emplace_back`

PR #672, opened 2022-06-25
enhance: use unique_ptr in Config

jackwener:
Part of #663 

In config.cc, use unique_ptr to eliminate some trivial manual deallocation

I have remove the vector [List-initialization](https://en.cppreference.com/w/cpp/language/list_initialization).

I use `[]` place it

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -188,7 +188,7 @@ Config::Config() {
   for (auto &wrapper : fields) {
     auto &field = wrapper.field;
     field->readonly = wrapper.readonly;
-    fields_.insert({wrapper.name, std::move(field)});
+    fields_.emplace(std::move(wrapper.name), std::move(field));
   }
   initFieldValidator();
   initFieldCallback();
--- a/src/config.cc
+++ b/src/config.cc
@@ -84,7 +84,7 @@ const char *configEnumGetName(configEnum *ce, int val) {
 
 Config::Config() {
   struct FieldWrapper {
-    const std::string name;
+    std::string name;
     bool readonly;
     std::unique_ptr<ConfigField> field;
 
--- a/src/config.cc
+++ b/src/config.cc
@@ -84,107 +84,240 @@ const char *configEnumGetName(configEnum *ce, int val) {
 
 Config::Config() {
   struct FieldWrapper {
-    const char *name;
+    const std::string name;
     bool readonly;
-    ConfigField *field;
+    std::unique_ptr<ConfigField> field;
+
+    FieldWrapper(std::string name, bool readonly,
+                 std::unique_ptr<ConfigField> field)
+        : name(std::move(name)), readonly(readonly), field(std::move(field)){};
+
+    FieldWrapper(const FieldWrapper &wrapper)
+        : name(std::move(wrapper.name)), readonly(wrapper.readonly) {}
   };
-  std::vector<FieldWrapper> fields = {
-      {"daemonize", true, new YesNoField(&daemonize, false)},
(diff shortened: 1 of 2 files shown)


PragmaTwice:
```c++
template<typename T, typename ...Args> auto UniqueConfig(Args &&... args) {
  return std::unique_ptr<ConfigField>(new T(std::forward<Args>(args)...));
}
```
You can add this utility function to smplify the changes : )

jackwener:
> ```c++
> template<typename T, typename ...Args> auto UniqueConfig(Args &&... args) {
>   return std::unique_ptr<ConfigField>(new T(std::forward<Args>(args)...));
> }
> ```
> 
> You can add this utility function to smplify the changes : )

Looks like changing a constructor is clearest. I have done it.

src/config.cc
```diff
@@ -84,11 +84,15 @@ const char *configEnumGetName(configEnum *ce, int val) {
...
   struct FieldWrapper {
-    const char *name;
+    const std::string name;
     bool readonly;
-    ConfigField *field;
+    std::unique_ptr<ConfigField> field;
+
+    FieldWrapper(std::string name, bool readonly,
+                 ConfigField* field)
+        : name(std::move(name)), readonly(readonly), field(std::move(field)) {}
```
PragmaTwice: Applying `std::move` to a pointer `ConfigField* field` is useless.

src/config.cc
```diff
@@ -484,9 +488,6 @@ void Config::initFieldCallback() {
 }
 
 Config::~Config() {
```
PragmaTwice: Could you delete it and add `~Config() = default;` in `config.h`?
We'd better make sure these trivial function in header to let compilers do more optimization like inlining and call elimination.

jackwener:
Thanks @PragmaTwice ! ❤️👍

src/config.cc
```diff
@@ -181,10 +185,10 @@ Config::Config() {
       {"rocksdb.level_compaction_dynamic_level_bytes",
         false, new YesNoField(&RocksDB.level_compaction_dynamic_level_bytes, false)},
   };
-  for (const auto &wrapper : fields) {
-    auto field = wrapper.field;
+  for (auto &wrapper : fields) {
+    auto &field = wrapper.field;
     field->readonly = wrapper.readonly;
-    fields_.insert({wrapper.name, field});
+    fields_.insert({wrapper.name, std::move(field)});
```
PragmaTwice:
```suggestion
    fields_.emplace(std::move(wrapper.name), std::move(field));
```
We can also move the name string 🤣

PragmaTwice (dismissed):
LGTM

src/config.cc
```diff
@@ -84,11 +84,15 @@ const char *configEnumGetName(configEnum *ce, int val) {
 
 Config::Config() {
   struct FieldWrapper {
-    const char *name;
+    const std::string name;
```
PragmaTwice:
```suggestion
    std::string name;
```

tisonkun: Why do you make this change?

PragmaTwice (approved):
LGTM

git-hulk (approved):

src/config.cc
PragmaTwice: `const std::string name` may prevent `std::move` from FieldWrapper::name to key of `std::map field_` in https://github.com/apache/incubator-kvrocks/blob/8836bc57d18927534c928d0b8b7d19a5d93c9d5f/src/config.cc#L191

tisonkun (approved):

PragmaTwice:
Thanks for you contribution! Merging...

PR #645, opened 2022-06-21
Add some RAII type for evbuffer to avoid manual free

PragmaTwice:
- Remove all manual call to `free` and `evbuffer_free`
- Add RAII type `UniqueEvbufReadln` and `UniqueEvbuf`

PR #673, opened 2022-06-25
use unique_ptr in src/server.h

zz-jason:
Signed-off-by: Jian Zhang <zjsariel@gmail.com>

to #663

### DIFF
--- a/src/server.h
+++ b/src/server.h
@@ -201,7 +201,7 @@ class Server {
   std::unique_ptr<Cluster> cluster_;
   static std::atomic<int> unix_time_;
   std::unique_ptr<class SlotMigrate> slot_migrate_;
-  class SlotImport *slot_import_;
+  class SlotImport *slot_import_ = nullptr;
 
  private:
   void cron();
--- a/src/server.h
+++ b/src/server.h
@@ -198,10 +198,10 @@ class Server {
 
   Stats stats_;
   Engine::Storage *storage_;
-  Cluster *cluster_;
+  std::unique_ptr<Cluster> cluster_;
   static std::atomic<int> unix_time_;
-  class SlotMigrate *slot_migrate_ = nullptr;
-  class SlotImport *slot_import_ = nullptr;
+  std::unique_ptr<class SlotMigrate> slot_migrate_;
+  class SlotImport *slot_import_;
 
  private:
   void cron();
--- a/cppcheck.sh
+++ b/cppcheck.sh
@@ -27,4 +27,4 @@ set -ex
 cppcheck --version
 cppcheck \
     --force --enable=${CHECK_TYPES} -U__GNUC__ -x ${LANG}  src --std=${STANDARD} --error-exitcode=${ERROR_EXITCODE} \
-    --inline-suppr --suppress=noCopyConstructor:src/server.cc --suppress=noOperatorEq:src/server.cc -j$(nproc)
+    --inline-suppr -j$(nproc)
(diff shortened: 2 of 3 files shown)


tisonkun:
This patch can be blocked by #645 as `missingInclude` is tricky to address. See also https://github.com/apache/incubator-kvrocks/pull/645#issuecomment-1162042973.

tisonkun (changes requested):
@zz-jason There're other failure in CI tasks. Please make sure that with you patch you can still compile kvrocks and pass `./cppcheck.sh` at least.

src/server.h
```diff
@@ -198,10 +198,10 @@ class Server {
 
   Stats stats_;
   Engine::Storage *storage_;
-  Cluster *cluster_;
+  std::unique_ptr<Cluster> cluster_;
   static std::atomic<int> unix_time_;
-  class SlotMigrate *slot_migrate_ = nullptr;
-  class SlotImport *slot_import_ = nullptr;
+  std::unique_ptr<class SlotMigrate> slot_migrate_;
+  class SlotImport *slot_import_;
```
PragmaTwice:
```suggestion
  class SlotImport *slot_import_ = nullptr;
```
fix the cppcheck warning in CI

PragmaTwice:
Could you remove `--suppress=noCopyConstructor:src/server.cc --suppress=noOperatorEq:src/server.cc` in `./cppcheck.sh` to make CI pass? Sorry for discommodity.

zz-jason:
I find the help message of `cppcheck` is:

```
--suppress=<spec>    Suppress warnings that match <spec>. The format of
                     <spec> is:
                     [error id]:[filename]:[line]
                     The [filename] and [line] are optional. If [error id]
                     is a wildcard '*', all error ids match.
```

Does `--suppress=noCopyConstructor:src/server.cc` mean to ignore the `noCopyConstructor` warning for `src/server.cc`?

PragmaTwice:
> I find the help message of `cppcheck` is:
> 
> ```
> --suppress=<spec>    Suppress warnings that match <spec>. The format of
>                      <spec> is:
>                      [error id]:[filename]:[line]
>                      The [filename] and [line] are optional. If [error id]
>                      is a wildcard '*', all error ids match.
> ```
> 
> Does `--suppress=noCopyConstructor:src/server.cc` mean to ignore the `noCopyConstructor` warning for `src/server.cc`?

Yeah, my guess is that the `noCopyConstructor` warning no longer appears because you changed the type of some of `Server`'s fields to `unique_ptr` (which has no copy ctor) so that `Server` can no longer copy constructible. 
At this point cppcheck reports `unmatchedSuppression` because we added an useless `--suppress` (after this change).
We will consider using clang tidy instead of cppcheck in the future.
```
src/server.cc:-1:0: information: Unmatched suppression: noCopyConstructor [unmatchedSuppression]
^
src/server.cc:-1:0: information: Unmatched suppression: noOperatorEq [unmatchedSuppression]
^
```
Actually you can just add `--suppress=unmatchedSuppression` instead of remove anything.

PragmaTwice (approved):
LGTM. cc @tisonkun

tisonkun (approved):
LGTM.

@git-hulk @PragmaTwice it seems we don't have `delete cluster_` to eliminate in this patch, did we leak memory previously on these field?

PragmaTwice:
> LGTM.
> 
> @git-hulk @PragmaTwice it seems we don't have `delete cluster_` to eliminate in this patch, did we leak memory previously on these field?

I also wonder that. 🤣

git-hulk:
@tisonkun @PragmaTwice Yes, we do leak this field even it wouldn't cause problem.

git-hulk (approved):

git-hulk:
Thanks for @zz-jason contribution, merging...

PR #681, opened 2022-06-27
enhance: use unique_ptr in SlotMigrate and FeedSlaveThread 

gogim1:
ref #663


caipengbo:
Can we merge this PR and #680 into one? I think they are almost the same.

gogim1:
> Can we merge this PR and #680 into one? I think they are almost the same.

done

PR #682, opened 2022-06-27
 enhance: use unique_ptr in SlotMigrate and FeedSlaveThread

gogim1:
ref #663

### DIFF
--- a/src/slot_migrate.cc
+++ b/src/slot_migrate.cc
@@ -19,10 +19,12 @@
  */
 
 #include "slot_migrate.h"
+
+#include <utility>
+
 #include "batch_extractor.h"
 #include "event_util.h"
 
-#include <utility>
 
 static std::map<RedisType, std::string> type_to_cmd = {
   {kRedisString, "set"},
--- a/src/slot_migrate.cc
+++ b/src/slot_migrate.cc
@@ -104,11 +104,13 @@ Status SlotMigrate::MigrateStart(Server *svr, const std::string &node_id, const 
   dst_node_ = node_id;
 
   // Create migration job
-  slot_job_ = std::unique_ptr<SlotMigrateJob>(new SlotMigrateJob(slot, dst_ip, dst_port,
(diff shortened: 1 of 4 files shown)


src/slot_migrate.cc
```diff
@@ -104,10 +104,9 @@ Status SlotMigrate::MigrateStart(Server *svr, const std::string &node_id, const
   dst_node_ = node_id;
 
   // Create migration job
-  SlotMigrateJob *job = new SlotMigrateJob(slot, dst_ip, dst_port,
-                                           speed, pipeline_size, seq_gap);
+  slot_job_ = std::unique_ptr<SlotMigrateJob>(new SlotMigrateJob(slot, dst_ip, dst_port,
+                                           speed, pipeline_size, seq_gap));
   std::lock_guard<std::mutex> guard(job_mutex_);
```
PragmaTwice: (not related to changes of this PR) It is weird since seems the scope of this `lock_guard` includes a `LOG(INFO)`, which has IO operations.

src/slot_migrate.cc
```diff
@@ -104,10 +104,9 @@ Status SlotMigrate::MigrateStart(Server *svr, const std::string &node_id, const
...
-  slot_job_ = job;
```
PragmaTwice, line 110: Does removing this line make the expected behavior different? This line is in a scope of a `lock_guard`.

gogim1: sorry, my bad. could i just move Line107 to the scope of the 'lock_guard', or maybe i need to alloc memory for a temp unique_ptr and move the temp unique_ptr's value to slot_job_ in the scope of 'lock_guard'

PragmaTwice: I think the latter is better.

git-hulk (approved):

PragmaTwice (approved):

PragmaTwice:

PR #683, opened 2022-06-27
Remove unique_ptr factory

jackwener:
Factory is redundant. We don't need to it make unique_ptr with it.

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -4714,9 +4714,7 @@ class CommandScript : public Commander {
 };
 
 #define ADD_CMD(name, arity, description , first_key, last_key, key_step, fn) \
-{name, arity, description, 0, first_key, last_key, key_step, []() -> std::unique_ptr<Commander> { \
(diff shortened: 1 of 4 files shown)


src/redis_cmd.cc
```diff
@@ -4714,9 +4714,7 @@ class CommandScript : public Commander {
 };
 
 #define ADD_CMD(name, arity, description , first_key, last_key, key_step, fn) \
-{name, arity, description, 0, first_key, last_key, key_step, []() -> std::unique_ptr<Commander> { \
-  return std::unique_ptr<Commander>(new fn()); \
-}}
+{name, arity, description, 0, first_key, last_key, key_step}
```
PragmaTwice, line 4717: Why did you remove `fn`? I think this is a classic factory pattern.

PragmaTwice: The role of the factory is to dispatch from the command name to the corresponding derived class of `Commander`

PR #689, opened 2022-06-29
ci: refactor CI yaml settings

tisonkun:
Sorting out steps and run redis-cli IT cases also on macos.

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -56,7 +56,6 @@ jobs:
           ./cpplint.sh
           ./cppcheck.sh
 
-
   build-and-test:
     name: Build and test
     needs: [license, lint]
@@ -97,7 +96,7 @@ jobs:
       - name: Install redis
         if: steps.cache-redis.outputs.cache-hit != 'true'
         run: |
-          curl https://github.com/redis/redis/archive/6.2.6.tar.gz
+          curl -O https://github.com/redis/redis/archive/6.2.6.tar.gz
           tar -xzvf 6.2.6.tar.gz
           pushd redis-6.2.6 && BUILD_TLS=yes make -j$NPROC redis-cli && sudo mv src/redis-cli /usr/bin/ && popd
 
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -99,8 +99,7 @@ jobs:
           curl -O https://download.redis.io/releases/redis-6.2.6.tar.gz
           tar -xzvf redis-6.2.6.tar.gz
           mkdir -p $HOME/local/bin
-          pushd redis-6.2.6 && BUILD_TLS=yes make -j$NPROC redis-cli && mv src/redis-cli && popd
-          mv redis-6.2.6/src/redis-cli $HOME/local/bin/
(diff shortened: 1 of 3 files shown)


.github/workflows/kvrocks.yaml
```diff
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-name: kvrocks ci actions  # don't edit while the badge was depend on this
```
tisonkun, line 18: The badge now depends on file name, so it's OK to change this field.

tisonkun:
Push 62c2952b8a8792bedfb308ba278303085b81c629 to ensure that cache logic works.

PragmaTwice:
Greate changes.

When I was changing this before, I wondered whether `apt install redis` / `brew install redis` could work well as building from source (which is the current way), but i didn't try. Do you think this is possible?

tests/tcl/tests/test_helper.tcl
```diff
@@ -60,6 +60,7 @@ set ::all_tests {
     unit/scripting
     integration/slotmigrate
     integration/slotimport
+    integration/redis-cli
```
PragmaTwice, line 63: Maybe we need change configuration of daily CI since `integration/redis-cli` is mandatory now? e.g. install redis-cli.

tisonkun: I don't check logics in daily CI. How can it be related?

tisonkun: OK. I can see the related code. Let me push a new commit.

PragmaTwice (approved):
LGTM

tisonkun:
@git-hulk @ShooterIT @PragmaTwice Perhaps we can move build and test tasks from daily CI to per PR CI. They take less than half hour to finish. While building docker image takes more than 2 hours to finish and I'm curious why.

PragmaTwice:
> @git-hulk @ShooterIT @PragmaTwice Perhaps we can move build and test tasks from daily CI to per PR CI. They take less than half hour to finish. While building docker image takes more than 2 hours to finish and I'm curious why.

Maybe it is due to that we use QEMU to simulate an arm64 environment, and I think that is slow.

tisonkun:
@PragmaTwice thanks for your input. Another approach is still run more Build and Test tasks every day but separate docker job into another workflow.

PragmaTwice:
Another thing is that since we build the nightly version of the docker image we can automatically publish it to some registries like the docker hub, e.g. some tags like `nightly`, `nightly-20220629`.

And I also think building docker image in CI triggered by PRs is unnecessary (maybe an arm evironment is useful, but a docker environment is not necessary for build and test).

git-hulk:
> Another thing is that since we build the nightly version of the docker image we can automatically publish it to some registries like the docker hub, e.g. some tags like `nightly`, `nightly-20220629`.
> 
> And I also think building docker image in CI triggered by PRs is unnecessary (maybe an arm evironment is useful, but a docker environment is not necessary for build and test).

Good point that we can push the `nightly` docker image, latest image need to wait for the new release is too slow for most users.

tisonkun:
Let's move the discussion of reviewing build docker image job to #691.

I'll merge this patch now.

PR #692, opened 2022-06-29
Forward C compiler to build deps in CMake

PragmaTwice:
In this PR, we make sure that those C dependencies with makefiles or automake is compiled by the compiler specified by `CMAKE_C_COMPILER`.


tisonkun:
Does this patch fix some issues existing now? BTW as the building logic grow complexity, we may rewrite it with Python or other more structural tools :)

PragmaTwice:
> Does this patch fix some issues existing now? BTW as the building logic grow complexity, we may rewrite it with Python or other more structural tools :)

Not yet, but I think it is a potential risk of not specifying the compiler for building dependencies.

Rewriting `build.sh` in python sounds great.

tisonkun (approved):
No harm to introduce such options as long as we don't promise backward compatibility on build script. `--gcc` and `--clang` looks like a handy option to me, we cannot grow options in this way too much.

git-hulk (approved):

PragmaTwice:
Thanks everyone, merging...

Issue #691, opened 2022-06-29
Review build docker image CI task which takes more than 2 hours to finish


tisonkun:
@git-hulk @ShooterIT @PragmaTwice Perhaps we can move build and test tasks from daily CI to per PR CI. They take less than half hour to finish. While building docker image takes more than 2 hours to finish and I'm curious why.

_Originally posted by @tisonkun in https://github.com/apache/incubator-kvrocks/issues/689#issuecomment-1169606788_

tisonkun:
> @git-hulk @ShooterIT @PragmaTwice Perhaps we can move build and test tasks from daily CI to per PR CI. They take less than half hour to finish. While building docker image takes more than 2 hours to finish and I'm curious why.

Maybe it is due to that we use QEMU to simulate an arm64 environment, and I think that is slow.

_Originally posted by @PragmaTwice in https://github.com/apache/incubator-kvrocks/issues/689#issuecomment-1169609906_

@PragmaTwice thanks for your input. Another approach is still run more Build and Test tasks every day but separate docker job into another workflow.

_Originally posted by @tisonkun in https://github.com/apache/incubator-kvrocks/issues/689#issuecomment-1169611332_

Another thing is that since we build the nightly version of the docker image we can automatically publish it to some registries like the docker hub, e.g. some tags like `nightly`, `nightly-20220629`.

And I also think building docker image in CI triggered by PRs is unnecessary (maybe an arm evironment is useful, but a docker environment is not necessary for build and test).

_Originally posted by @PragmaTwice in https://github.com/apache/incubator-kvrocks/issues/689#issuecomment-1169614450_

> Another thing is that since we build the nightly version of the docker image we can automatically publish it to some registries like the docker hub, e.g. some tags like `nightly`, `nightly-20220629`.
> 
> And I also think building docker image in CI triggered by PRs is unnecessary (maybe an arm evironment is useful, but a docker environment is not necessary for build and test).

Good point that we can push the `nightly` docker image, latest image need to wait for the new release is too slow for most users.

_Originally posted by @git-hulk in https://github.com/apache/incubator-kvrocks/issues/689#issuecomment-1169619824_

git-hulk:
The QEMU was introduce in PR: https://github.com/apache/incubator-kvrocks/pull/550, maybe we can move other jobs to per CI from daily CI except the QEMU docker image.

PragmaTwice:
[...]

PR #694, opened 2022-06-29
ci: add a label to trigger daily CI in PR

### LINKED ISSUE #665: Add a label to trigger daily CI in PR
### Search before asking

### Motivation

We could add a label like `need-daily-CI`, and when the label is applied to a PR, all jobs of daily CI will join in the checking sequence of this PR.

### Solution

_No response_

gogim1:
close #665

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -116,7 +116,7 @@ jobs:
           cd tests/tcl && ./runtest
 
   daily-ci:
-    if: ${{ github.event.label.name == 'need-daily-CI' }}
+    if: github.event.label.name == 'need-daily-CI'
     name: Run Daily CI
     uses: ./.github/workflows/daily-ci.yaml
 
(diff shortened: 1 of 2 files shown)


ShooterIT:
cool, thanks

tisonkun (changes requested):
This change doesn't work because pull_request event doesn't include label changes. See also https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#label.

Also I commented on #665 for an alternative - we may not implement the feature in this way.

.github/workflows/kvrocks.yaml
```diff
@@ -115,6 +115,12 @@ jobs:
           cp $HOME/local/bin/redis-cli tests/tcl/redis-cli
           cd tests/tcl && ./runtest
 
+  daily-ci:
+    if: contains(github.event.pull_request.labels.*.name, 'need-daily-CI')
```
tisonkun, line 119:
```suggestion
    if: contains(github.event.pull_request.labels.*.name, 'needs-daily-ci')
```

tisonkun (commented):
Sorry. I get it wrong that the condition should be event=pull_request + type=labeled.

However, the condition should be handled in `daily-ci.yaml`, where you can change `paths:` to `types: [labeled]`.

tisonkun:
@GoGim1 I'm integrate this patch with a solution to #691.

gogim1:
thanks everyone!

PR #695, opened 2022-06-29
ci: separate daily ci and trigger daily ci only if labeled

### LINKED ISSUE #665: Add a label to trigger daily CI in PR

### LINKED ISSUE #691: Review build docker image CI task which takes more than 2 hours to finish

tisonkun:
This closes #665 and closes #691.

This supersedes and will close #694.

### DIFF
--- a/.github/workflows/daily-ci.yaml
+++ b/.github/workflows/daily-ci.yaml
@@ -27,7 +27,10 @@ on:
 jobs:
   build-docker-image:
     name: Build Docker Image
-    if: (github.event_name != 'pull_request') || (contains(github.event.pull_request.labels.*.name, 'needs-daily-ci'))
+    if: |
+      (github.event_name != 'pull_request') ||
+      (github.event.action == 'labeled' && github.event.label.name == 'needs-daily-ci') ||
+      (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'needs-daily-ci'))
     runs-on: ubuntu-18.04
     steps:
       - name: Checkout Code Base
--- a/.github/workflows/daily-ci.yaml
+++ b/.github/workflows/daily-ci.yaml
@@ -19,22 +19,19 @@ name: Daily CI
 
 on:
   pull_request:
-    paths:
-      - ".github/workflows/daily-ci.yaml"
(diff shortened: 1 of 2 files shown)


PragmaTwice (approved):
LGTM

.github/workflows/daily-ci.yaml
```diff
@@ -19,22 +19,22 @@ name: Daily CI
...
     - cron: '0 1 * * *'
   workflow_dispatch:
 
 jobs:
   build-docker-image:
     name: Build Docker Image
+    if: |
+      (github.event_name != 'pull_request') ||
+      (github.event.action == 'labeled' && github.event.label.name == 'needs-daily-ci') ||
+      (github.event.action != 'labeled' && contains(github.event.pull_request.labels.*.name, 'needs-daily-ci'))
```
PragmaTwice, line 33: I think the `workflow_call` method in #694 maybe better since we can have several jobs in the daily ci workflow (every job need an `if` here).

tisonkun: We can actually run kvrocks.yaml job everyday.

tisonkun: > every job need an `if` here

I don't get the point. It's not a regression.

tisonkun: If we call `daily-ci.yaml` from`kvrocks.yaml`, not only the dependencies logic will mess, but also we should now handle `pull_request` event with `labeled` type in `kvrocks.yaml`, it will cause more complex logic.

In Pulsar and previous SkyWalking CI settings we do repeat some conditions as long as it won't grow too much. We can review this issue if it cause concrete trouble.

PR #697, opened 2022-06-29
✨ feat: luajit replace lua

xiaobiaozhao:
1. Luajit replace lua
2. rm test case "cmsgpack can pack and unpack circular references". Becase luajit's core code diffrent from lua. Also, different compilers produce different code execution effects (GCC&clang on Ubuntu).


PragmaTwice:
Thanks for your hard work.

> rm test case "cmsgpack can pack and unpack circular references". Becase luajit's core code diffrent from lua. Also, different compilers produce different code execution effects (GCC&clang on Ubuntu).

I did not get this point, could you elaborate on it?

I think that the JIT mechanism should not affect the original semantics, and the same interpreter / JIT program should keep same behavior while compiled by different C compilers. So we should consider it as a serious problem if it really has some difference on semantics between the original interpreter and the JIT, or between different C compilers.

And I think the same problem as #609 (#614) for luajit needs to be solved.

tests/tcl/tests/unit/scripting.tcl
```diff
@@ -235,37 +235,37 @@ start_server {tags {"scripting"}} {
...
-                assert(re.x.x.x.x.x.x.x.x.x.x.x.x.x.x.y == re.y)
-                -- now the last x would be b above and has no y
-                assert(re.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x)
-                -- so, the final x.x is at the depth limit and was assigned nil
-                assert(re.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x == nil)
-                return {h, re.x.x.x.x.x.x.x.x.y == re.y, re.y == 5}
-        } 0
-    } {82a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a178c0 1 1}
+    #test {EVAL - cmsgpack can pack and unpack circular references?} {
+    #    r eval {local a = {x=nil,y=5}
```
git-hulk: Why we disable this test case?

xiaobiaozhao: I had fix this test.  The case is from here https://github.com/antirez/lua-cmsgpack/blob/master/test.lua#L393

git-hulk: cool, thanks

tests/tcl/tests/unit/scripting.tcl
```diff
@@ -263,9 +263,10 @@ start_server {tags {"scripting"}} {
                 assert(re.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x)
                 -- so, the final x.x is at the depth limit and was assigned nil
                 assert(re.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x.x == nil)
-                return {h, re.x.x.x.x.x.x.x.x.y == re.y, re.y == 5}
+                assert(h == "82a17881a17882a17881a17882a17881a17882a17881a17882a17881a17882a17881a17882a17881a17882a17881a178c0a17905a17905a17905a17905a17905a17905a17905a17905" or h == "82a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a17882a17905a17881a178c0")
```
PragmaTwice, line 266: Why there are two different values of `h`?

xiaobiaozhao: I had fix this test. The case is from here https://github.com/antirez/lua-cmsgpack/blob/master/test.lua#L393

PragmaTwice: I got it. It is due to the element order of a table is unstable. This reminds me of Map in golang.

CMakeLists.txt
```diff
@@ -59,15 +59,15 @@ include(cmake/snappy.cmake)
 include(cmake/lz4.cmake)
 include(cmake/rocksdb.cmake)
 include(cmake/libevent.cmake)
-include(cmake/lua.cmake)
+include(cmake/luajit.cmake)
```
PragmaTwice, line 62: I think we can provide a build option to switch from luajit and lua for progressive substitution.

PragmaTwice, line 65: I think we can provide a build option to switch between luajit and lua for progressive substitution.

cmake/luajit.cmake
```diff
@@ -0,0 +1,58 @@
...
+# under the License.
+
+include_guard()
+
+include(FetchContent)
+
+FetchContent_Declare(luajit
+  GIT_REPOSITORY https://github.com/xiaobiaozhao/LuaJIT.git
+  GIT_TAG 803487f8b01c672495a2fcd29dcbed09e4fd6319
+)
```
PragmaTwice: Now we have `FetchContent_DeclareGitHubWithMirror`, refer to https://github.com/apache/incubator-kvrocks/blob/unstable/cmake/lua.cmake

cmake/luajit.cmake
```diff
@@ -0,0 +1,58 @@
...
+      WORKING_DIRECTORY ${luajit_SOURCE_DIR}/src
+      BYPRODUCTS ${luajit_SOURCE_DIR}/src/libluajit.a
+    )
+  else()
+    add_custom_target(make_luajit COMMAND make libluajit.a
+      "CFLAGS=${LUA_CFLAGS}"
+      WORKING_DIRECTORY ${luajit_SOURCE_DIR}/src
+      BYPRODUCTS ${luajit_SOURCE_DIR}/src/libluajit.a
+  )
+  endif()
```
PragmaTwice: You can introduce a variable to avoid duplication.

cmake/luajit.cmake
```diff
@@ -0,0 +1,58 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+include_guard()
+
+include(FetchContent)
+
+FetchContent_Declare(luajit
+  GIT_REPOSITORY https://github.com/xiaobiaozhao/LuaJIT.git
```
PragmaTwice: I think we could need to build a new repo under some organization rather than a personal one.

src/scripting.h
```diff
@@ -23,9 +23,7 @@
 #include <string>
 #include <vector>
 
-#include <lua.h>
-#include <lauxlib.h>
-#include <lualib.h>
+#include <lua.hpp>
```
PragmaTwice: I think this header does not solve the issue in #609, and we need some investigation to determine how to avoid memory leaks in c++ in luajit.

PragmaTwice (commented):
As I commented in discussion, Redis did some changes to lua codebases, not only lua extensions. You can check [here](https://github.com/redis/redis/commits/unstable/deps/lua/src
) to see these commits.  I haven't investigated carefully what all these changes affect, and this may need some work. 

Other comments are inline.

tests/tcl/tests/unit/scripting.tcl
xiaobiaozhao: Yes， It behaves differently in different compilers

PragmaTwice: > Yes， It behaves differently in different compilers

I still have deep doubts about this, because I think we should try to avoid semantics behaving differently under different compilers.

cmake/luajit.cmake
git-hulk: Yes, I will checkout this personal repo to KvrocksLabs org

xiaobiaozhao:
BTW, We can do the Luajit extension based on this repository. https://github.com/openresty/luajit2.
The repo add some new function  https://github.com/openresty/luajit2#new-lua-apis

cmake/luajit.cmake
git-hulk: @xiaobiaozhao Can you help to submit your changes to https://github.com/KvrocksLabs/LuaJIT?

git-hulk: I pushed your commits, you can just switch to the new repo

xiaobiaozhao: done https://github.com/apache/incubator-kvrocks/pull/697/commits/c4a66978d9e3c313a186ac577a98ee2a635ca8e5

cmake/luajit.cmake
xiaobiaozhao: done https://github.com/apache/incubator-kvrocks/pull/697/commits/33e874e0e4acc93dac2f7a9b4d5bdf9cf56fe37e

cmake/luajit.cmake
xiaobiaozhao: done https://github.com/apache/incubator-kvrocks/pull/697/commits/33e874e0e4acc93dac2f7a9b4d5bdf9cf56fe37e

src/scripting.h
xiaobiaozhao: http://luajit.org/extensions.html#exceptions

PragmaTwice: Thanks, seems there is a detail comment: https://github.com/luajit/luajit/blob/master/src/lj_err.c#L20-L67

git-hulk: > I think this header does not solve the issue in #609, and we need some investigation to determine how to avoid memory leaks in c++ in luajit.

@xiaobiaozhao you can try to call `DestroyState` in the Server destructor since it leaks by `cjson` open would init the buffer(1024). But I'm wondering why it happens in luajit only.

xiaobiaozhao: The issus still exists. https://github.com/apache/incubator-kvrocks/pull/697/commits/707b177d60daf37dfcfbdc75a9a62ea4238a966d

xiaobiaozhao: I can't reproduce in my ubuntu 20.4

PragmaTwice:
In order to use sanitizers, you not only need to add compile options, but also link options. (which may be why CI is currently failing)
Also, don't forget to remove these sanitizer options when the bug is fixed : )

xiaobiaozhao:
image
This is  mem leak on my ubuntu

src/server.cc
```diff
@@ -174,6 +174,7 @@ Status Server::Start() {
 
 void Server::Stop() {
   stop_ = true;
+  Lua::DestroyState(lua_);
```
git-hulk, line 177: Can move this line to the server destructor and delete the server after the server threads join.

git-hulk:
Many thanks for @xiaobiaozhao hard try

xiaobiaozhao:
> As I commented in discussion, Redis did some changes to lua codebases, not only lua extensions. You can check [here](https://github.com/redis/redis/commits/unstable/deps/lua/src) to see these commits. I haven't investigated carefully what all these changes affect, and this may need some work.
> 
> Other comments are inline.

[here](https://github.com/redis/redis/commits/unstable/deps/lua/src) The part of the lua code is modified for Redis 7.0, mainly the "redis function". I don't think this part needs to be ported to KVROCKS at this time

PragmaTwice:
> [here](https://github.com/redis/redis/commits/unstable/deps/lua/src) The part of the lua code is modified for Redis 7.0, mainly the "redis function". I don't think this part needs to be ported to KVROCKS at this time

Thanks for your investigation. 

How do you think about this idea: https://github.com/apache/incubator-kvrocks/pull/697#discussion_r910672053 ?
I think it is not a hard job to support both since all these APIs are same.

If users encounter incompatibility between luajit and lua that makes the previous code unable to execute properly, or other problems with luajit, they can choose to switch to original lua to avoid it.

git-hulk:
@tisonkun and @ShooterIT can take a look if you have time. For https://github.com/apache/incubator-kvrocks/pull/697#discussion_r910672053, I also prefer providing an option to let user fallback if they have any problem.

General implementation is good to me.

CMakeLists.txt
```diff
@@ -25,8 +25,9 @@ option(ENABLE_ASAN "enable address santinizer" OFF)
 option(ENABLE_TSAN "enable thread santinizer" OFF)
 option(ASAN_WITH_LSAN "enable leak santinizer while address santinizer is enabled" ON)
 option(ENABLE_STATIC_LIBSTDCXX "link kvrocks with static library of libstd++ instead of shared library" ON)
+option(USE_LUAJIT "enable luajit replace lua" ON)
```
```suggestion
option(USE_LUAJIT "use luaJIT instead of lua" ON)
```

src/server.h
```diff
@@ -30,7 +30,13 @@
 #include <memory>
 #include <unordered_map>
 
+#ifdef USE_LUAJIT
+#include <lua.hpp>
+#else
+#include <lauxlib.h>
 #include <lua.h>
+#include <lualib.h>
+#endif
```
PragmaTwice: With https://github.com/KvrocksLabs/lua/pull/4, we do not need `USE_LUAJIT` definition checking. Could you help to review & merge it? @git-hulk

git-hulk: sure, thanks @PragmaTwice 

@xiaobiaozhao you can update the newest commit.

CMakeLists.txt
xiaobiaozhao: done

git-hulk (dismissed):

.github/workflows/kvrocks.yaml
```diff
@@ -101,6 +101,9 @@ jobs:
             os: ubuntu-18.04
             with_ninja: --ninja
             without_jemalloc: -DDISABLE_JEMALLOC=ON
+          - name: Ubuntu GCC without luajit
+            os: ubuntu-18.04
+            without_luajit: -DUSE_LUAJIT=OFF
```
PragmaTwice, line 115: I think you need to add `${{ matrix.without_luajit }}` to line 140, otherwise it is useless

xiaobiaozhao: done https://github.com/apache/incubator-kvrocks/pull/697/commits/d4a6ff9a7a0a38d01f5ea0228714c1cd0d98f644

PragmaTwice:
Maybe you need to add `*.hpp` to line 42 in lua.cmake to pass the "Ubuntu GCC without luajit" job.

xiaobiaozhao:
> Maybe you need to add `*.hpp` to line 42 in lua.cmake to pass the "Ubuntu GCC without luajit" job.

done

git-hulk (dismissed):

tisonkun (dismissed):

git-hulk (dismissed):

PragmaTwice:
There are some memory leaks reported by ASan which are steadily reproduced and need some investigation. 

The rest is good for me. 👍

git-hulk:
> There are some memory leaks reported by ASan which are steadily reproduced and need some investigation.
> 
> The rest is good for me. 👍

Those memory leaks looks not related to this PR.

PragmaTwice:
> Those memory leaks looks not related to this PR.

Seems these reported leaks are steadily (100%) reproduced for more than 10 times, and do not appear in other PRs and the unstable branch, so I think it is hard to say that it is not related to this PR.

I'll take a deeper look when I have time : )

git-hulk:
> > Those memory leaks looks not related to this PR.
> 
> Seems these reported leaks are steadily (100%) reproduced for more than 10 times, and do not appear in other PRs and the unstable branch, so I think it is hard to say that it is not related to this PR.
> 
> I'll take a deeper look when I have time : )

I fixed those issues in #714 and it looks good now, I think this PR would be also became green after #714.

git-hulk (approved):

git-hulk:
@tisonkun @PragmaTwice @ShooterIT I think we can merge this PR if no objections.

tisonkun (approved):
LGTM. Thank you very much for this fruitful contribution @xiaobiaozhao!

@git-hulk one thing to point out is that we may not include this change in 2.1.0 release and at earliest in 2.2.0.

tisonkun (approved):
LGTM. Thank you very much for this fruitful contribution @xiaobiaozhao!

@git-hulk one thing to point out is that we may not include this change in 2.1.0 release and at earliest in 2.2.0.

PragmaTwice (approved):
LGTM, thanks for such a great work.

git-hulk:
@tisonkun Agree, I also think this is a big change and we can release it on the next release.

ShooterIT (approved):

git-hulk:
Thanks all, merging...

git-hulk:
Thanks for @xiaobiaozhao great contribution again.

Issue #677, opened 2022-06-26
Exclude unstable branch while canceling previous workflows


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

![image](https://user-images.githubusercontent.com/20042607/175820242-e318efaa-3b18-4cfa-a235-ec3caba21e10.png)

As shown in the figure above, the retrying workflow was canceled due to a new commit.
We may retry the kvrocks ci action of some commits in the unstable branch. 
At this time, if the unstable branch can only run one workflow at the same time, it will cause some trouble.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

tisonkun:
I don't think this is a big deal. Always we focus on the latest version and it's fine if only the nightly version passes CI.

If you want to debugging an unstable test, you may open a branch on fork or send a draft pull request.

PragmaTwice:
> I don't think this is a big deal. Always we focus on the latest version and it's fine if only the nightly version passes CI.
> 
> If you want to debugging an unstable test, you may open a branch on fork or send a draft pull request.

In fact, there is another situation. 

For example, if two commits are merged into the unstable branch within a short time interval, the workflow of the previous commit will be canceled, and the actions status of the previous commit will display a ❌.

This can lead users to think that our unstable branch will often fail, which may not be what we expected, and this situation may be more serious than the one above.

tisonkun:
@PragmaTwice since we require every PR should be up-to-date before merge, actually we can skip running CI in `unstable` branch.

See also [Apache Pulsar's CI settings](https://github.com/apache/pulsar/blob/6c715e08c67c40221c8c08cbe0b31d3a6aa1bf32/.github/workflows/pulsar-ci.yaml):

```yaml
on:
  pull_request:
    branches:
      - master
  push:
    branches:
      - branch-*
```

PragmaTwice:
> @PragmaTwice since we require every PR should be up-to-date before merge, actually we can skip running CI in `unstable` branch.
> 
> See also [Apache Pulsar's CI settings](https://github.com/apache/pulsar/blob/6c715e08c67c40221c8c08cbe0b31d3a6aa1bf32/.github/workflows/pulsar-ci.yaml):
> 
> ```yaml
> on:
>   pull_request:
>     branches:
>       - master
>   push:
>     branches:
>       - branch-*
> ```

[...]

PR #700, opened 2022-06-30
Disable require branches to be up-to-date

tisonkun:
As development under Kvrocks grows rapidly, it causes unnecessary reruns for patches due to we require branches to be up to date before merging.

We can disable the option and let the committers decide whether we should merge base branch before accepting the patch.

See also this discussion on general@: https://lists.apache.org/thread/w1tvjxzooyrjso80hdd7y04n2q6k7mxl

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -39,7 +39,6 @@ github:
         dismiss_stale_reviews: true
         required_approving_review_count: 1
       required_status_checks:
-        strict: true
         contexts:
           - Required
 


ShooterIT (approved):

git-hulk (approved):

PragmaTwice (approved):
BTW any idea on #677? Seems CI runs on unstable branch become more important after this change.

tisonkun:
@PragmaTwice we still run tests at least in latest commit for unstable branch (and I think we don't have to rerun previous commit but try to fix test to be stable) so I don't think we should over optimize CI workflow. Another perspective is we may run `kvrocks.yaml` with a schedule trigger also.

PR #704, opened 2022-07-02
use const & for avoiding copy

jackwener:
replace `const auto& ` with `const auto &` just to unify format. 

Add `const &` to avoid copy

### DIFF
--- a/tests/cppunit/rwlock_test.cc
+++ b/tests/cppunit/rwlock_test.cc
@@ -27,7 +27,7 @@
 TEST(LockManager, LockKey) {
   LockManager locks(8);
   std::vector<rocksdb::Slice> keys = {"abc", "123", "456", "abc", "123"};
-  for (const auto key : keys) {
+  for (const auto &key : keys) {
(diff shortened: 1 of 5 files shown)


git-hulk (approved):

PragmaTwice (approved):

git-hulk:
Thanks @jackwener, merging...

tests/cppunit/rwlock_test.cc
```diff
@@ -27,7 +27,7 @@
 TEST(LockManager, LockKey) {
   LockManager locks(8);
   std::vector<rocksdb::Slice> keys = {"abc", "123", "456", "abc", "123"};
-  for (const auto key : keys) {
```
mapleFU, line 30: I think copying slice maybe more lightweight then using reference?

PragmaTwice: Yeah, maybe passing by value is more effective in some cases. Don't know why there are many `const Slice&` even in rocksdb itself.

PR #705, opened 2022-07-03
Refactor Task: remove useless struct field

PragmaTwice:
`std::function` is already a type-erased closure container which can be used to carry arguments, so the `void * arg` in `struct Task` seems to be useless.

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -1073,16 +1073,13 @@ Status Server::AsyncCompactDB(const std::string &begin_key, const std::string &e
   }
   db_compacting_ = true;
 
-  Task task;
-  task.arg = this;
-  task.callback = [begin_key, end_key](void *arg) {
-    auto svr = static_cast<Server *>(arg);
+  Task task = [begin_key, end_key, this] {
     Slice *begin = nullptr, *end = nullptr;
     if (!begin_key.empty()) begin = new Slice(begin_key);
     if (!end_key.empty()) end = new Slice(end_key);
-    svr->storage_->Compact(begin, end);
-    std::lock_guard<std::mutex> lg(svr->db_job_mu_);
-    svr->db_compacting_ = false;
+    storage_->Compact(begin, end);
+    std::lock_guard<std::mutex> lg(db_job_mu_);
+    db_compacting_ = false;
     delete begin;
     delete end;
   };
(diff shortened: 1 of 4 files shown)


tisonkun (commented):
@PragmaTwice what's the background of this refactor? For style or efficiency, or sth else?

jackwener:
> @PragmaTwice what's the background of this refactor? For style or efficiency, or sth else?

Looks like is to simplify the code.

Remove the useless code struct field `arg`.

Indeed, the supplementary explanation will make it easier to get to key point.

PragmaTwice:
> @PragmaTwice what's the background of this refactor? For style or efficiency, or sth else?

Added. Thanks for reminding.

tisonkun (commented):
Some subtle difference from the change before and after could be, if you catch the argument in place and later they're changed, result may be different. However, if you confirm that those arguments are constant, this change should be fine.

PragmaTwice:
> Some subtle difference from the change before and after could be, if you catch the argument in place and later they're changed, result may be different. However, if you confirm that those arguments are constant, this change should be fine.

Yeah, but seems `TaskRunner` does not expose any interface to change fields of `Task` elements in it. It just has some interface to submit tasks and run tasks.

tisonkun:
@PragmaTwice I don't think we should support something like "capture argument changes", but if the arguments are captured later, we should pass them instead of capture them at the place.

It's not a blocker for this patch. But I'd like to point out that such "refactor" isn't always trivial and no side effect.

PragmaTwice:
> @PragmaTwice I don't think we should support something like "capture argument changes", but if the arguments are captured later, we should pass them instead of capture them at the place.
> 
> It's not a blocker for this patch. But I'd like to point out that such "refactor" isn't always trivial and no side effect.

Thanks.

In fact, as long as it is not added to `TaskRunner`, we don't need to do unnecessary type erasure, e.g.:
```c++
TaskRunner runner; // a container of `std::function<void()>` elements

auto some_func = [a, b](T v){ do_something(v, a); do_other(v, b); }; // `T v` is delay for capturing later

for(auto i : c) { runner.push([i, some_func]{ some_func(i); }); }
```

tisonkun (approved):
LGTM

PR #716, opened 2022-07-05
kvrocks.conf: deleted block resize command, added perflog ratio details

### LINKED ISSUE #717: Remove deprecated configuration `auto-resize-block-and-sst`
### Search before asking

### Motivation

`auto-resize-block-and-sst` was used to auto-resize block and sst size in flight,
but we deprecated it after using the rocksdb partition index, so we should remove it
from the config file.

### Solution

remove deprecated `auto-resize-block-and-sst` from config file.

AvivPl:
- It is unclear what the range of the perflog ratio is. ([0,1]  ,[0, 100] ,etc.)
- resizing block size is deprecated.

Closes #717

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -301,8 +301,8 @@ supervised no
 # Default: empty
 # profiling-sample-commands ""
 
-# Ratio of the samples would be recorded. We simply use the rand to determine
-# whether to record the sample or not.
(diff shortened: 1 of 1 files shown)


git-hulk:
Thanks @AvivPl

git-hulk (approved):

PR #721, opened 2022-07-08
Update kvrocks.yaml

kezhenxu94:
We have a dedicate GitHub actions for header check so this should be updated, and we don't recommend using `@main` so there is no unexpected behaviors when we have changes in our codebase.

See https://github.com/apache/skywalking-eyes/pull/123

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: apache/skywalking-eyes@main
+      - uses: apache/skywalking-eyes/header@501a28d2fb4a9b962661987e50cf0219631b32ff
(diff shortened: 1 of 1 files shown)


PragmaTwice (approved):
LGTM. BTW is there a suitable tag rather than a commit hash?

kezhenxu94:
> LGTM. BTW is there a suitable tag rather than a commit hash?

We haven't had a release 0.4.0 so for now we have to use a commit hash, we will manage to release 0.4.0 soon so you can change to `@v0.4.0`.

xiaobiaozhao:
LGTM

tisonkun (approved):

PragmaTwice:
> We haven't had a release 0.4.0 so for now we have to use a commit hash, we will manage to release 0.4.0 soon so you can change to `@v0.4.0`.

Got it. Merging, thanks for you contribution!

PR #722, opened 2022-07-09
Return values instead of passing pointers

jackwener:
Utilize NRVO,  return values instead of passing pointers.


PragmaTwice (approved):
LGTM, BTW it is NRVO instead of RVO.

tisonkun (approved):

tisonkun:
Merging...

Thanks for your contribution @jackwener !

PR #725, opened 2022-07-11
replace shell scripts with x.py

tisonkun:
See also https://github.com/apache/incubator-kvrocks/pull/692#issuecomment-1169738759


x.py
```diff
@@ -0,0 +1,146 @@
...
+
+    find_command("autoconf", msg="autoconf is required to build jemalloc")
+    cmake = find_command("cmake", msg="CMake is required")
+
+    output = run([cmake, "-version"], stdout=subprocess.PIPE)
+    output = run(["head", "-n", "1"], stdin=output, stdout=subprocess.PIPE)
+    output = run(["sed", "s/[^0-9.]*//g"], stdin=output, stdout=subprocess.PIPE)
+    cmake_version = output.read().decode().strip()
+    if semver.compare(cmake_version, CMAKE_REQUIRE_VERSION) < 0:
+        raise RuntimeError(f"CMake {CMAKE_REQUIRE_VERSION} or higher is required, got: {cmake_version}")
```
tisonkun: I don't download CMake here if it's missing. Since it causes issue like https://github.com/apache/incubator-kvrocks/discussions/712 and I'd prefer to let the user explicit download such significant and complex configurable dependency.

Anyway, if we don't automatically download autoconf, why CMake?

x.py
```diff
@@ -0,0 +1,146 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+import glob
+import os
+import pathlib
+import click
+import semver
```
PragmaTwice: I think the use of packages in pypi might make it a little more difficult for users to use this script than before?

tisonkun: It's possible to rewrite usage of `click` with `argparse` and `semver` with simple function, while I can see installing pypi libs is a well-known skill. Let me try to rewrite the script but I don't have a strong feeling on this part.

tisonkun:
Look like we run out of OSS only credit on Travis CI and all pending jobs (at least of Kvrocks) are hanging to be queued. I'm asking INFRA on Slack for help.

UPDATE - It seems something wrong with Travis CI. Waiting for verification :(

Dockerfile
```diff
@@ -22,11 +22,11 @@ ENV TZ=Asia/Shanghai
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
 RUN apt update
-RUN apt install -y cmake make git autoconf libtool g++
+RUN apt install -y cmake make git autoconf libtool g++ python3 python3-pip
```
PragmaTwice:
```suggestion
RUN apt install -y cmake make git autoconf libtool g++ python3
```

PragmaTwice (approved):
LGTM

x.py
```diff
@@ -0,0 +1,175 @@
...
+    cmake = find_command("cmake", msg="CMake is required")
+
+    output = run([cmake, "-version"], stdout=subprocess.PIPE)
+    output = run(["head", "-n", "1"], stdin=output, stdout=subprocess.PIPE)
+    output = run(["sed", "s/[^0-9.]*//g"], stdin=output, stdout=subprocess.PIPE)
+    cmake_version = output.read().decode().strip()
+    cmake_require_version = '.'.join(map(str, CMAKE_REQUIRE_VERSION))
+    cmake_semver = SEMVER_REGEX.match(cmake_version)
+    if cmake_semver is None:
+        raise RuntimeError(f"CMake {cmake_require_version} or higher is required, got: {cmake_version}")
```
PragmaTwice, line 85: I think we can provide a URL for users and developers to download suitable cmake, like https://github.com/Kitware/CMake/releases or https://cmake.org/download/

PragmaTwice: Also, I think it would be good to provide a `--cmake-bin` to customize the use of the cmake executable.

These can be done in subsequent PRs.

tisonkun: I think it's too cumbersome to convey possible download links. Still we don't recommend for autoconf and users should be easily to search how to download CMake. For example, as a macOS user I'd prefer to manage CMake via Homebrew instead of download from either links provided here.

It's too much to a downstream project heuristically educating users to install a dependency.

tisonkun: Also we already provide some preferred defaults about getting prerequisites in the README.md file.

tisonkun: > Also, I think it would be good to provide a `--cmake-bin` to customize the use of the cmake executable.

Yes. This sounds reasonable to me.

.github/workflows/kvrocks.yaml
```diff
@@ -42,19 +42,20 @@ jobs:
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.x
+      - name: Setup dependencies
+        run: pip install cpplint==1.5.0
       - name: Lint and check code
         run: |
-          ./cpplint.sh
-          ./cppcheck.sh
+          ./x.py check cpplint
```
git-hulk, line 57: Just curious, why do we call this file `x.py`?

tisonkun: Naming is hard. I stolen the idea from [rust-lang/rust](https://github.com/rust-lang/rust) to call the universal script entrypoint `x.py`.

git-hulk: Got it, thanks @tisonkun

git-hulk (approved):

tisonkun:
@PragmaTwice @git-hulk I'll run `docker build .` locally and merge this patch once it's OK.

Hope Travis CI can fix the queuing issue in time :(

tisonkun:
Merging...

x.py
tisonkun: @PragmaTwice since this patch merged, you may create an issue for `--cmake-bin`.

PR #745, opened 2022-07-21
Implement the data structure and partial commands of the Redis stream 

### LINKED ISSUE #532: [QUESTION] Any plan to support Redis Stream?
Hi, do you have any plans to support Redis Stream? That's good for use KVrocks as replacement for message-oriented system. My company is one of the needs of this.
Detailed: 

- [https://redis.io/commands/?group=stream](https://redis.io/commands/?group=stream) 
- [http://antirez.com/news/114](http://antirez.com/news/114)
- [http://antirez.com/news/128](http://antirez.com/news/128)

torwig:
**Available commands:**
- XADD
- XDEL
- XINFO STREAM
- XLEN
- XRANGE
- XREAD
- XREVRANGE
- XTRIM

**Design notes**
Link to the Redis data type description: https://redis.io/docs/manual/data-types/streams/
A stream is a sequence of entries.
Each entry has a unique ID in the form "1-2" where two 64-bit numbers are divided by a hyphen.
By default, the first number is set to a millisecond timestamp, and the second one is a so-called sequence number - for cases when more than one entry was added at the same millisecond.
The value of an entry is a set of key-value pairs.
In case of the command:
`XADD s1 1-0 key1 val1 key2 val2`
the ID is _1-0_ and the value is _key1 val1 key2 val2_.
In RocksDB entries are represented as key-value pairs where the key is formed as:
`key | version | entry-ID-milliseconds-value | entry-ID-sequence-number-value`
and the value is encoded as:
`key1-length(fixed-32) | key1 | val1-length(fixed-32) | val1 | key2-length(fixed-32) | key2 | val2-length(fixed-32) | val2`.
Thanks to the structure of a key, all entries in a stream are sorted in chronological order.
As for _value_ decoding: this is the first idea that came to my mind and maybe it's not very efficient because has an overhead (4 bytes on every argument).
Why did I introduce such a weird encoding scheme? Because if you are reading entries, Redis responds not with a single string:
```
1) 1) "s1"
   2) 1) 1) "1-0"
         2) 1) "key1"
            2) "val1"
            3) "key2"
            4) "val2"
```
Perhaps, command args can be joined with a ' '(space) into a single string and this string should be saved in RocksDB? After reading, it will be split while constructing the reponse. With this encoding scheme, I was thinking about the possible spaces inside arguments and how to deal with them?

**Differences from Redis**
1. `XTRIM` and `XADD with trim possibility`: nearly exact trimming (via `~`) is not possible due to implementation details (no radix tree here). However, `LIMIT` option is working while in Redis it is allowed only in combination with `~`. `LIMIT` can be disallowed to be consistent with Redis protocol. I didn't do that because I want to hear opinions from kvrocks maintainers.

Replication is not implemented yet. Basically, I didn't test streams in a cluster configuration. Perhaps, the plain `XREAD` on a replica will work, but blocking `XREAD` that unblocks after `XADD` on the master - I'm sure that some code should be written. It would be greatly appreciated if maintainers provide me with some hints about how to implement this.

Consumer groups are not implemented. I'm thinking about the possible implementation.

Right now I'm looking for any maintainers' feedback from the adding-new-data-type perspective (maybe, I didn't add a new column family to some filter/checker/extractor, etc.) and information about proper replicating a stream from master to other nodes.

This closes #532

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -71,6 +71,7 @@ const char *errNoSuchKey = "no such key";
 const char *errUnbalacedStreamList =
     "Unbalanced XREAD list of streams: for each stream key an ID or '$' must be specified.";
 const char *errTimeoutIsNegative = "timeout is negative";
+const char *errLimitOptionNotAllowed = "syntax error, LIMIT cannot be used without the special ~ option";
 
 class CommandAuth : public Commander {
  public:
@@ -4832,18 +4833,7 @@ class CommandXAdd : public Commander {
       }
 
       if (val == "limit" && !entry_id_found) {
-        if (i+1 >= args.size()) {
-          return Status(Status::RedisParseErr, errInvalidSyntax);
-        }
-
-        try {
-          limit_ = std::stoull(args[i+1]);
-          with_limit_ = true;
-        } catch (const std::exception &) {
-          return Status(Status::RedisParseErr, errValueNotInterger);
-        }
-        i += 2;
-        continue;
+        return Status(Status::RedisParseErr, errLimitOptionNotAllowed);
       }
 
       if (val == "*" && !entry_id_found) {
@@ -4889,10 +4879,6 @@ class CommandXAdd : public Commander {
       options.trim_options.strategy = StreamTrimStrategy::MinID;
       options.trim_options.min_id = min_id_;
     }
-    if (with_limit_) {
-      options.trim_options.with_limit = true;
-      options.trim_options.limit = limit_ == 0 ? UINT64_MAX : limit_;
-    }
     if (with_entry_id_) {
       options.with_entry_id = true;
       options.entry_id = entry_id_;
@@ -4920,14 +4906,12 @@ class CommandXAdd : public Commander {
  private:
   std::string stream_name_;
   uint64_t max_len_ = 0;
-  uint64_t limit_;
   Redis::StreamEntryID min_id_;
   Redis::NewStreamEntryID entry_id_;
   std::vector<std::string> name_value_pairs_;
   bool nomkstream_ = false;
   bool with_max_len_ = false;
   bool with_min_id_ = false;
-  bool with_limit_ = false;
   bool with_entry_id_ = false;
 };
 
@@ -5602,25 +5586,19 @@ class CommandXTrim : public Commander {
       return Status(Status::RedisParseErr, errInvalidSyntax);
     }
 
-    std::string limit;
+    bool limit_option_found = false;
     if (eq_sign_found) {
       if (args.size() > 6 && Util::ToLower(args[5]) == "limit") {
-        with_limit_ = true;
-        limit = args[6];
+        limit_option_found = true;
       }
     } else {
       if (args.size() > 5 && Util::ToLower(args[4]) == "limit") {
-        with_limit_ = true;
-        limit = args[5];
+        limit_option_found = true;
       }
     }
 
-    if (with_limit_) {
-      try {
-        limit_ = std::stoull(limit);
-      } catch (const std::exception &) {
-        return Status(Status::RedisParseErr, errValueNotInterger);
-      }
+    if (limit_option_found) {
+      return Status(Status::RedisParseErr, errLimitOptionNotAllowed);
     }
 
     return Status::OK();
@@ -5637,10 +5615,6 @@ class CommandXTrim : public Commander {
     options.strategy = strategy_;
     options.max_len = max_len_;
     options.min_id = min_id_;
-    options.with_limit = with_limit_;
-    if (options.with_limit) {
-      options.limit = limit_ == 0 ? UINT64_MAX : limit_;
-    }
 
     uint64_t removed;
     auto s = stream_db.Trim(args_[1], options, &removed);
@@ -5654,11 +5628,9 @@ class CommandXTrim : public Commander {
   }
 
  private:
-  uint64_t limit_ = 0;
   uint64_t max_len_ = 0;
   StreamEntryID min_id_;
   StreamTrimStrategy strategy_ = StreamTrimStrategy::None;
-  bool with_limit_ = false;
 };
 
 #define ADD_CMD(name, arity, description , first_key, last_key, key_step, fn) \
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -4771,17 +4771,25 @@ class CommandXAdd : public Commander {
       }
 
       if (val == "maxlen" && !entry_id_found) {
-        std::string max_len_str;
+        if (i+1 >= args.size()) {
+          return Status(Status::RedisParseErr, errInvalidSyntax);
+        }
+
+        size_t max_len_idx;
         bool eq_sign_found = false;
         if (args[i+1] == "=") {
-          max_len_str = args[i+2];
+          max_len_idx = i+2;
           eq_sign_found = true;
         } else {
-          max_len_str = args[i+1];
+          max_len_idx = i+1;
+        }
+
+        if (max_len_idx >= args.size()) {
+          return Status(Status::RedisParseErr, errInvalidSyntax);
         }
 
         try {
-          max_len_ = std::stoull(max_len_str);
+          max_len_ = std::stoull(args[max_len_idx]);
           with_max_len_ = true;
         } catch (const std::exception &) {
           return Status(Status::RedisParseErr, errValueNotInterger);
@@ -4792,17 +4800,25 @@ class CommandXAdd : public Commander {
       }
 
       if (val == "minid" && !entry_id_found) {
-        std::string min_id_str;
+        if (i+1 >= args.size()) {
+          return Status(Status::RedisParseErr, errInvalidSyntax);
+        }
+
+        size_t min_id_idx;
         bool eq_sign_found = false;
-        if (args[i + 1] == "=") {
-          min_id_str = args[i+2];
+        if (args[i+1] == "=") {
+          min_id_idx = i+2;
           eq_sign_found = true;
         } else {
-          min_id_str = args[i+1];
+          min_id_idx = i+1;
+        }
+
+        if (min_id_idx >= args.size()) {
+          return Status(Status::RedisParseErr, errInvalidSyntax);
         }
 
         try {
-          auto s = ParseStreamEntryID(min_id_str, &min_id_);
+          auto s = ParseStreamEntryID(args[min_id_idx], &min_id_);
           if (!s.IsOK()) {
             return Status(Status::RedisParseErr, s.Msg());
           }
@@ -4816,6 +4832,10 @@ class CommandXAdd : public Commander {
       }
 
       if (val == "limit" && !entry_id_found) {
+        if (i+1 >= args.size()) {
+          return Status(Status::RedisParseErr, errInvalidSyntax);
+        }
+
         try {
           limit_ = std::stoull(args[i+1]);
           with_limit_ = true;
@@ -5248,6 +5268,10 @@ class CommandXRead : public Commander {
       }
 
       if (arg == "count") {
+        if (i+1 >= args.size()) {
+          return Status(Status::RedisParseErr, errInvalidSyntax);
+        }
+
         try {
           with_count_ = true;
           count_ = static_cast<uint64_t>(std::stoll(args[i+1]));
@@ -5259,6 +5283,10 @@ class CommandXRead : public Commander {
       }
 
       if (arg == "block") {
+        if (i+1 >= args.size()) {
+          return Status(Status::RedisParseErr, errInvalidSyntax);
+        }
+
         block_ = true;
         try {
           auto v = std::stoll(args[i+1]);
@@ -5534,31 +5562,39 @@ class CommandXTrim : public Commander {
     if (trim_strategy == "maxlen") {
       strategy_ = StreamTrimStrategy::MaxLen;
 
-      std::string len;
+      size_t max_len_idx;
       if (args[3] != "=") {
-        len = args[3];
+        max_len_idx = 3;
       } else {
-        len = args[4];
+        max_len_idx = 4;
         eq_sign_found = true;
       }
 
+      if (max_len_idx >= args.size()) {
+        return Status(Status::RedisParseErr, errInvalidSyntax);
+      }
+
       try {
-        max_len_ = std::stoull(len);
+        max_len_ = std::stoull(args[max_len_idx]);
       } catch (const std::exception &) {
         return Status(Status::RedisParseErr, errValueNotInterger);
       }
     } else if (trim_strategy == "minid") {
       strategy_ = StreamTrimStrategy::MinID;
 
-      std::string id;
+      size_t min_id_idx;
       if (args[3] != "=") {
-        id = args[3];
+        min_id_idx = 3;
       } else {
-        id = args[4];
+        min_id_idx = 4;
         eq_sign_found = true;
       }
 
-      auto s = ParseStreamEntryID(id, &min_id_);
+      if (min_id_idx >= args.size()) {
+        return Status(Status::RedisParseErr, errInvalidSyntax);
+      }
+
+      auto s = ParseStreamEntryID(args[min_id_idx], &min_id_);
       if (!s.IsOK()) {
         return s;
       }
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -45,6 +45,7 @@
 #include "redis_pubsub.h"
 #include "redis_sortedint.h"
 #include "redis_slot.h"
+#include "redis_stream.h"
 #include "replication.h"
 #include "util.h"
 #include "storage.h"
@@ -66,6 +67,10 @@ const char *errWrongNumOfArguments = "wrong number of arguments";
 const char *errValueNotInterger = "value is not an integer or out of range";
 const char *errAdministorPermissionRequired = "administor permission required to perform the command";
 const char *errValueMustBePositive = "value is out of range, must be positive";
+const char *errNoSuchKey = "no such key";
+const char *errUnbalacedStreamList =
+    "Unbalanced XREAD list of streams: for each stream key an ID or '$' must be specified.";
+const char *errTimeoutIsNegative = "timeout is negative";
 
 class CommandAuth : public Commander {
  public:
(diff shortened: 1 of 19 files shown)


caipengbo:
Can you add some design documents so that people can understand your design ideas :)

git-hulk:
Thanks to @torwig great contribution.

For the replication part, Kvrocks will replicate the data after writing into db, so we needn't to do anything if it has no special case. For the new column family, I prefer to keeping less column family and feel free to add if we MUST.

torwig:
@caipengbo Yes, sure, I will :)

torwig:
@git-hulk I mean that if I run
`XREAD BLOCK 0 STREAMS s1 $` on a replica (it will block)
and then execute
`XADD s1 * k1 v1` on master
the `XREAD` will be unblocked and return the added entry.
It's very similar to publish-subscribe - there is a fan-out.
Does it require some additional code in processing replicated data?

git-hulk:
> @git-hulk I mean that if I run `XREAD BLOCK 0 STREAMS s1 $` on a replica (it will block) and then execute `XADD s1 * k1 v1` on master the `XREAD` will be unblocked and return the added entry. It's very similar to publish-subscribe - there is a fan-out. Does it require some additional code in processing replicated data?

Yes, you are right. Kvrocks default replication only sends the data to replicas, we need to recognize those special write batch and wake up the waiting clients.

torwig:
@caipengbo Added design notes.

git-hulk:
> @caipengbo Added design notes.

Many thanks to @torwig detail explanation and bring this good feature into Kvrocks community, we will take a look recently.

src/redis_cmd.cc
```diff
@@ -4730,6 +4735,876 @@ class CommandScript : public Commander {
...
+      if (val == "nomkstream" && !entry_id_found) {
+        nomkstream_ = true;
+        ++i;
+        continue;
+      }
+
+      if (val == "maxlen" && !entry_id_found) {
+        std::string max_len_str;
+        bool eq_sign_found = false;
+        if (args[i+1] == "=") {
```
git-hulk, line 4756: Need to check the length of arguments before `i+1`

src/redis_cmd.cc
```diff
@@ -4730,6 +4735,876 @@ class CommandScript : public Commander {
...
+        }
+
+        i += eq_sign_found ? 3 : 2;
+        continue;
+      }
+
+      if (val == "minid" && !entry_id_found) {
+        std::string min_id_str;
+        bool eq_sign_found = false;
+        if (args[i + 1] == "=") {
```
git-hulk, line 4777: this same with above

git-hulk:
I took a look briefly and overall design is good to me. There are two small issues in design internal:

1. I think we should save the entries number as well in the encoded value instead of depending on '\0', we should allow using '\0' in the string.
2. We need to create the column family for stream in [storage::open](https://github.com/apache/incubator-kvrocks/blob/unstable/src/storage.cc#L234) if wants to use the new column family, or it will fallback to the default column family.

torwig:
> I took a look briefly and overall design is good to me. There are two small issues in design internal:
> 
> 1. I think we should save the entries number as well in the encoded value instead of depending on '\0', we should allow using '\0' in the string.
> 2. We need to create the column family for stream in [storage::open](https://github.com/apache/incubator-kvrocks/blob/unstable/src/storage.cc#L234) if wants to use the new column family, or it will fallback to the default column family.

Thank you for your review. I'll fix the issues. 
Just one question for now: what do you mean by depending on '\0'? Currently, the encoded value consists of:
`first-string-len | first-string | second-string-len | second string`
as implemented here: https://github.com/apache/incubator-kvrocks/pull/745/files#diff-eee0baf52c379afb7e25739db14ec62c603d3c8f4c7f46af24e019319b2bb4beR135

git-hulk:
> > I took a look briefly and overall design is good to me. There are two small issues in design internal:
> > 
> > 1. I think we should save the entries number as well in the encoded value instead of depending on '\0', we should allow using '\0' in the string.
> > 2. We need to create the column family for stream in [storage::open](https://github.com/apache/incubator-kvrocks/blob/unstable/src/storage.cc#L234) if wants to use the new column family, or it will fallback to the default column family.
> 
> Thank you for your review. I'll fix the issues. Just one question for now: what do you mean by depending on '\0'? Currently, the encoded value consists of: `first-string-len | first-string | second-string-len | second string` as implemented here: https://github.com/apache/incubator-kvrocks/pull/745/files#diff-eee0baf52c379afb7e25739db14ec62c603d3c8f4c7f46af24e019319b2bb4beR135

Yes, I mean we can also save the number of encoded value. As this example, the encoded value can be `2(4bytes)|first-string-len | first-string | second-string-len | second string` which we can count the number of strings directly instead of depending of end of string. How do you think about this case?

torwig:
@git-hulk Fixed possible crashes via access to command non-existing command arguments by index. Created column family for streams. I will use it also for replication purposes, to unblock blocked readers.

torwig:
> > > I took a look briefly and overall design is good to me. There are two small issues in design internal:
> > > 
> > > 1. I think we should save the entries number as well in the encoded value instead of depending on '\0', we should allow using '\0' in the string.
> > > 2. We need to create the column family for stream in [storage::open](https://github.com/apache/incubator-kvrocks/blob/unstable/src/storage.cc#L234) if wants to use the new column family, or it will fallback to the default column family.
> > 
> > 
> > Thank you for your review. I'll fix the issues. Just one question for now: what do you mean by depending on '\0'? Currently, the encoded value consists of: `first-string-len | first-string | second-string-len | second string` as implemented here: https://github.com/apache/incubator-kvrocks/pull/745/files#diff-eee0baf52c379afb7e25739db14ec62c603d3c8f4c7f46af24e019319b2bb4beR135
> 
> Yes, I mean we can also save the number of encoded value. As this example, the encoded value can be `2(4bytes)|first-string-len | first-string | second-string-len | second string` which we can count the number of strings directly instead of depending of end of string. How do you think about this case?

I've got your idea about writing the number of strings to the encoded value. However, I didn't understand how this information can help us. I mean, if we know that there are 2 strings encoded, how this number can work in the decoding process? Could you please provide me with any example?
Currently, we read length `len1`, the read exactly `len1` bytes as a string, then read `len2`, then read exactly `len2` bytes as a string.

git-hulk:
> > > > I took a look briefly and overall design is good to me. There are two small issues in design internal:
> > > > 
> > > > 1. I think we should save the entries number as well in the encoded value instead of depending on '\0', we should allow using '\0' in the string.
> > > > 2. We need to create the column family for stream in [storage::open](https://github.com/apache/incubator-kvrocks/blob/unstable/src/storage.cc#L234) if wants to use the new column family, or it will fallback to the default column family.
> > > 
> > > 
> > > Thank you for your review. I'll fix the issues. Just one question for now: what do you mean by depending on '\0'? Currently, the encoded value consists of: `first-string-len | first-string | second-string-len | second string` as implemented here: https://github.com/apache/incubator-kvrocks/pull/745/files#diff-eee0baf52c379afb7e25739db14ec62c603d3c8f4c7f46af24e019319b2bb4beR135
> > 
> > 
> > Yes, I mean we can also save the number of encoded value. As this example, the encoded value can be `2(4bytes)|first-string-len | first-string | second-string-len | second string` which we can count the number of strings directly instead of depending of end of string. How do you think about this case?
> 
> I've got your idea about writing the number of strings to the encoded value. However, I didn't understand how this information can help us. I mean, if we know that there are 2 strings encoded, how this number can work in the decoding process? Could you please provide me with any example? Currently, we read length `len1`, the read exactly `len1` bytes as a string, then read `len2`, then read exactly `len2` bytes as a string.

I rethink about this point, current implementation is ok since C++ string didn't depend on '\0' at all, so use the string length to determine where's value end is ok. Sorry that I misunderstand the implementation. Just disregard it.

torwig:
@git-hulk Added batch handling on replicas to unblock clients that were blocked by the XREAD command by the respective XADD.

src/redis_stream_base.h
```diff
@@ -0,0 +1,162 @@
...
+  bool operator<(const StreamEntryID &rhs) const {
+    if (ms < rhs.ms) return true;
+    if (ms == rhs.ms) return seq < rhs.seq;
+    return false;
+  }
+
+  bool operator>=(const StreamEntryID &rhs) const {
+    if (ms > rhs.ms) return true;
+    if (ms == rhs.ms) return seq >= rhs.seq;
+    return false;
```
PragmaTwice:
```suggestion
    return !(*this < rhs);
```

src/redis_stream_base.h
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  bool operator>(const StreamEntryID &rhs) const {
+    if (ms > rhs.ms) return true;
+    if (ms == rhs.ms) return seq > rhs.seq;
+    return false;
+  }
```
PragmaTwice:
```suggestion
  bool operator>(const StreamEntryID &rhs) const {
    return rhs < *this;
  }
  
  bool operator<=(const StreamEntryID &rhs) const {
    return !(rhs < *this);
  }
```

src/redis_stream_base.h
```diff
@@ -0,0 +1,162 @@
...
+};
+
+struct StreamInfo {
+  uint64_t size;
+  uint64_t entries_added;
+  StreamEntryID last_generated_id;
+  StreamEntryID max_deleted_entry_id;
+  StreamEntryID recorded_first_entry_id;
+  StreamEntry *first_entry = nullptr;
+  StreamEntry *last_entry = nullptr;
```
PragmaTwice: Consider to use `std::unique_ptr`?

PragmaTwice (commented):
Such a great work! Not familiar with redis stream, just comment for some little code smell.

src/redis_stream_base.h
torwig: Done

src/redis_stream_base.h
torwig: Done

src/redis_stream_base.h
torwig: Done

git-hulk:
> @git-hulk Added batch handling on replicas to unblock clients that were blocked by the XREAD command by the respective XADD.

Thanks for your great contribution, I will try and review the PR in a few days.

torwig:
@git-hulk I've removed the LIMIT option from trim options to be compliant with Redis protocol.
By default, Redis uses exact trimming (or if `=` is specified). And in this case, the LIMIT option can't be used.
The LIMIT option can be used only with nearly exact trimming (if `~` is specified instead of `=`), but nearly exact trimming is bound to Redis implementation of a stream (there are nodes with elements), so it can't be applied with RocksDB.

Also, I will resolve some conflicts with the `unstable` branch in a few minutes.

git-hulk:
> @git-hulk I've removed the LIMIT option from trim options to be compliant with Redis protocol. By default, Redis uses exact trimming (or if `=` is specified). And in this case, the LIMIT option can't be used. The LIMIT option can be used only with nearly exact trimming (if `~` is specified instead of `=`), but nearly exact trimming is bound to Redis implementation of a stream (there are nodes with elements), so it can't be applied with RocksDB.
> 
> Also, I will resolve some conflicts with the `unstable` branch in a few minutes.

Thanks @torwig, it makes sense to me since the underlay structures are different. I think we can have basic stream functions, then try to resolve the nearly extract trimming issue only if many community users requested it, or just left it behind.

src/redis_cmd.cc
```diff
@@ -4750,6 +4756,883 @@ class CommandScript : public Commander {
...
+
+    if (name_value_pairs_.empty() || name_value_pairs_.size() % 2 != 0) {
+      return Status(Status::RedisParseErr, errWrongNumOfArguments);
+    }
+
+    return Status::OK();
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    if (svr->IsSlave()) {
```
git-hulk: Can remove this check since we have already done it at: https://github.com/apache/incubator-kvrocks/blob/40132067bc7bb696f96675e8a396f0473321a454/src/redis_connection.cc#L407

src/redis_cmd.cc
```diff
@@ -4750,6 +4756,883 @@ class CommandScript : public Commander {
...
+      }
+      ids_.push_back(id);
+    }
+    return Status::OK();
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    if (svr->IsSlave()) {
+      return Status(Status::NotOK, "READONLY You can't write against a read only slave");
+    }
```
git-hulk: can remove as well

git-hulk:
I have reviewed this PR, implementation, test coverage and design are good to me. I will approve this PR after above conversations were resolved.

PTAL @ShooterIT @Alfejik @caipengbo @PragmaTwice, I think we can merge this PR after it got >= 3 approves.

caipengbo:
Sorry, there are some small things recently. I will carry out CR within this week :)

src/redis_cmd.cc
torwig: Done

src/redis_cmd.cc
torwig: Done

git-hulk (dismissed):

PragmaTwice (dismissed):

src/redis_metadata.cc
```diff
@@ -287,3 +289,65 @@ rocksdb::Status ListMetadata::Decode(const std::string &bytes) {
...
+  PutFixed64(dst, last_entry_id.ms);
+  PutFixed64(dst, last_entry_id.seq);
+
+  PutFixed64(dst, entries_added);
+}
+
+rocksdb::Status StreamMetadata::Decode(const std::string &bytes) {
+  // flags(1byte) + expire (4byte)
+  if (bytes.size() < 5) {
+    return rocksdb::Status::InvalidArgument("the metadata is too short");
```
caipengbo: Would it be better if we extract `the metadata is too short` into a const variable? 
I found a lot of the same strings in this file.

src/redis_stream_base.cc
```diff
@@ -0,0 +1,158 @@
...
+  auto pos = input.find("-");
+  if (pos != std::string::npos) {
+    try {
+      auto ms_str = input.substr(0, pos);
+      auto seq_str = input.substr(pos + 1);
+
+      id->ms = std::stoull(ms_str);
+      id->seq = std::stoull(seq_str);
+    } catch (const std::exception &) {
+      return Status(Status::RedisParseErr, "Invalid stream ID specified as stream command argument");
```
caipengbo: make `"Invalid stream ID specified as stream command argument"` a const string? 

It's repeated 7 times.

src/redis_stream_base.cc
```diff
@@ -0,0 +1,158 @@
...
+    }
+  } else {
+    try {
+      id->ms = std::stoull(input);
+      id->seq = 0;
+    } catch (const std::exception &) {
+      return Status(Status::RedisParseErr, "Invalid stream ID specified as stream command argument");
+    }
+  }
+  return Status();
```
caipengbo:
```suggestion
  return Status::OK();
```

src/redis_stream_base.cc
```diff
@@ -0,0 +1,158 @@
...
+    try {
+      auto ms_str = input.substr(0, pos);
+      auto seq_str = input.substr(pos + 1);
+
+      id->ms = std::stoull(ms_str);
+      id->seq = std::stoull(seq_str);
+    } catch (const std::exception &) {
+      return Status(Status::RedisParseErr, "Invalid stream ID specified as stream command argument");
+    }
+  } else {
```
caipengbo: Because we handle exception, I personally prefer to use this format: 
```c++
if () {
    xxx
    return;
}
xxxx;
return;
```
I think this will reduce indentation and make the code cleaner.

src/server.cc
```diff
@@ -491,6 +533,28 @@ Status Server::WakeupBlockingConns(const std::string &key, size_t n_conns) {
...
+Status Server::OnEntryAddedToStream(const std::string &ns, const std::string &key,
+                                    const Redis::StreamEntryID &entry_id) {
+  std::lock_guard<std::mutex> guard(blocking_keys_mu_);
+  auto iter = blocked_stream_consumers_.find(key);
+  if (iter == blocked_stream_consumers_.end() || iter->second.empty()) {
+    return Status(Status::NotOK);
+  }
+
+  for (auto it = iter->second.begin(); it != iter->second.end(); ) {
+    StreamConsumer* consumer = *it;
```
caipengbo: Recommend use smart pointers.

caipengbo:
[redis_cmd.cc](https://github.com/apache/incubator-kvrocks/blob/unstable/src/redis_cmd.cc) file was already long because it contained the command implementation. 

Now we have added a new `Stream` type, which is much longer. Maybe we'll add other new types in the future. They will definitely have a lot of commands too, which will make the file huge.

Should we consider splitting this file, separating the various types of commands into different files? HDYT @git-hulk @PragmaTwice

src/redis_metadata.cc
torwig: Done

src/redis_stream_base.cc
torwig: Done

src/redis_stream_base.cc
torwig: Done

src/redis_stream_base.cc
torwig: I refactored that try-block is now outside the if-block. Please, have a look. Maybe, it's not exactly what you meant.

src/server.cc
torwig: Done

caipengbo:
LGTM

caipengbo (dismissed):

PragmaTwice:
> [redis_cmd.cc](https://github.com/apache/incubator-kvrocks/blob/unstable/src/redis_cmd.cc) file was already long because it contained the command implementation.
> 
> Now we have added a new `Stream` type, which is much longer. Maybe we'll add other new types in the future. They will definitely have a lot of commands too, which will make the file huge.
> 
> Should we consider splitting this file, separating the various types of commands into different files? HDYT @git-hulk @PragmaTwice

It sounds reasonable.

PragmaTwice (dismissed):

git-hulk:
> > [redis_cmd.cc](https://github.com/apache/incubator-kvrocks/blob/unstable/src/redis_cmd.cc) file was already long because it contained the command implementation.
> > Now we have added a new `Stream` type, which is much longer. Maybe we'll add other new types in the future. They will definitely have a lot of commands too, which will make the file huge.
> > Should we consider splitting this file, separating the various types of commands into different files? HDYT @git-hulk @PragmaTwice
> 
> It sounds reasonable.

@caipengbo Yes, agree that we should separate commands into each data structure after this PR. I'll submit a new issue to track this improvement.

git-hulk (dismissed):

git-hulk:
@torwig Can you help to resolve the conflict?

torwig:
@git-hulk Done

git-hulk:
Thanks @torwig

We can merge this PR after one of @Alfejik @ShooterIT approved since the new commit only resolved command number conflict, so I think the previous approves are still valid. 

Of course, guys can approve it again if you're free. cc @caipengbo @PragmaTwice

git-hulk (approved):

PragmaTwice (approved):

tisonkun:
I don't have time to go through this patch. Although, it seems this patch looks good to two reviewers. If we accept such a new feature, remember to track updates on our doc site :)

git-hulk:
> I don't have time to go through this patch. Although, it seems this patch looks good to two reviewers. If we accept such a new feature, remember to track updates on our doc site :)

Thanks to @tison warm remind. Will merge this PR if have no further feedback tomorrow, then update the doc site after merging.

Alfejik (approved):

git-hulk:
Hello, thanks  to @torwig brings this awesome feature for Kvrocks community and everyone who reviewed this PR. I will summary and merge this PR and welcome to create a new thread to further discussion. Cheers!!!

git-hulk:
Thanks @torwig and @aleksraiden again.

ShooterIT:
as we know, prefix bloom filter can improve the performance of range read,  but current encoding make it hard to use since we can't know key length and how to set prefix length. One PR still is in draft https://github.com/apache/incubator-kvrocks/pull/508

> key | version | entry-ID-milliseconds-value | entry-ID-sequence-number-value

for the entry encoding, i think we can put version at the front,  so that prefix bloom filter will be useful.

furthermore, maybe we should redesign the encoding for all complex data structure for better range read performance. Of course, we should keep compatibility, such as, adding an encoding version to distinguish different version for decoding, this way also allow us implement new features easily.

> key1-length(fixed-32) | key1 | val1-length(fixed-32) | val1 | key2-length(fixed-32) | key2 | val2-length(fixed-32) | val2.

i think we use  variable length encoding, such as `EncodeVarint32` in leveldb/rocksdb,  to save disk space.

git-hulk:
> for the entry encoding, i think we can put version at the front, so that prefix bloom filter will be useful.

I think we can change all complex data structures at the next major release to make the prefix bloom filter more effective.

PR #746, opened 2022-07-22
Fix multiple arch nightly images by docker manifest in Travis CI

PragmaTwice:
It seems that only images for a single arch was uploaded to docker hub.

We can use `docker manifest` to compose multiple arch into one tag like other images did.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,31 +17,47 @@
 
 language: cpp
 
-arch:
-  - amd64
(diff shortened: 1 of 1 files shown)


git-hulk (approved):

tisonkun (approved):

PragmaTwice:
Thanks all. Merging...

PR #747, opened 2022-07-22
Fix manifest creation of nightly tag in Travis CI

PragmaTwice:
In previous commit, tag [nightly-20220722-e3eb8f4](https://hub.docker.com/layers/kvrocks/kvrocks/kvrocks/nightly-20220722-e3eb8f4/images/sha256-b85a5c0343ad6dca8a5306c3f7a3c3873b357d6ab420b26137c5bf0471f62f34?context=explore) is successfully pushed to docker hub with both amd64 and arm64 arch, 
but the tag [nightly](https://hub.docker.com/layers/kvrocks/kvrocks/kvrocks/nightly/images/sha256-eaa51003ed6f121d9fe3adcfbeb0c4a7d8c87192ef1f7a70d162b3b411c123e7?context=explore) failed to be updated.

This PR fixed it, which is a stupid mistake. 🤣

PR #748, opened 2022-07-22
Fix updating of nightly images in Travis CI

PragmaTwice:
#747 did not solve the problem of `nightly` tag, although I do not know why `--amend MANIFEST_LIST` did not work.

So I use another way to implement and test in my local device to ensure it works.

Sorry for that.


git-hulk (approved):
LGTM, cool.

tisonkun (approved):
Never mind. We don't verify the case in PR and this is a normal debugging loop.

You should feel good to bring such valuable feature to Kvrocks community :)

PragmaTwice:
Thanks all. Merging...

Issue #639, opened 2022-06-19
BUG: Data race reported by TSan in kvrocks


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Version

unstable, linux

### Minimal reproduce step

```sh
./build.sh build --unittest -DENABLE_TSAN=ON
./build/unittest
```

### What did you expect to see?

No TSan report.

### What did you see instead?

Selected TSan log in unittest:
```
WARNING: ThreadSanitizer: data race (pid=14679)
  Read of size 1 at 0x7b7c0000673b by thread T1:
    #0 memcmp <null> (libtsan.so.0+0x67cb3)
    #1 rocksdb::Slice::compare(rocksdb::Slice const&) const /home/runner/work/incubator-kvrocks/incubator-kvrocks/build/_deps/rocksdb-src/include/rocksdb/slice.h:250 (unittest+0x5d75ef)
    #2 Compare /home/runner/work/incubator-kvrocks/incubator-kvrocks/build/_deps/rocksdb-src/util/comparator.cc:34 (unittest+0x5d75ef)
    #3 Engine::SubKeyFilter::Filter(int, rocksdb::Slice const&, rocksdb::Slice const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >*, bool*) const /home/runner/work/incubator-kvrocks/incubator-kvrocks/src/compact_filter.cc:111 (unittest+0x18605c)
    #4 rocksdb::CompactionFilter::FilterV2(int, rocksdb::Slice const&, rocksdb::CompactionFilter::ValueType, rocksdb::Slice const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >*) const /home/runner/work/incubator-kvrocks/incubator-kvrocks/build/_deps/rocksdb-src/include/rocksdb/compaction_filter.h:171 (unittest+0x67ba24)
    #5 rocksdb::CompactionIterator::InvokeFilterIfNeeded(bool*, rocksdb::Slice*) /home/runner/work/incubator-kvrocks/incubator-kvrocks/build/_deps/rocksdb-src/db/compaction/compaction_iterator.cc:276 (unittest+0x67ba24)
  Previous write of size 1 at 0x7b7c0000673b by main thread (mutexes: write M19766):
[...]

PR #759, opened 2022-07-27
Propagate sanitizer flags to all cmake deps

PragmaTwice:
- solve #639
- apply sanitizer flags to all cmake deps to avoid false positive report
- ban sanitizer build while jemalloc is enabled
- add [tsan-suppressions](https://github.com/apache/incubator-kvrocks/compare/unstable...PragmaTwice:fix-san-build?expand=1#diff-a16454a82fd4f3403a2e89418377f5f34f7f566fba4456bdc9005667c6c4fcc1) to suppress specific TSan reports
- enable TSan builds in the workflow


PragmaTwice:
```
*** [err]: requests on cluster are ok in tests/integration/cluster.tcl
Expected '16383' to be equal to '' (context: type source line 257 file /Users/runner/work/incubator-kvrocks/incubator-kvrocks/tests/tcl/tests/integration/cluster.tcl cmd {assert_equal "16383" [$r3 get $slot_16383_key]} proc ::test)
```
from https://github.com/apache/incubator-kvrocks/runs/7536793002?check_suite_focus=true

.github/workflows/kvrocks.yaml
```diff
@@ -158,10 +156,13 @@ jobs:
...
 
       - name: Run Unit Test
-        run: ./build/unittest
+        run: |
+          export ${{ matrix.runtime_env_vars }}
+          ./build/unittest
 
       - name: Run Redis Tcl Test
         run: |
+          export ${{ matrix.runtime_env_vars }}
```
tisonkun, line 165: You can use `env` field of a step.

See also https://docs.github.com/en/actions/learn-github-actions/environment-variables.

PragmaTwice: I think it is hard to eval something in `env` like `$(realpath ./tests/tsan-suppressions)`, do you have any suggestion here?

PragmaTwice:
Hello everyone, any thought on this PR? ❤️

.github/workflows/kvrocks.yaml
tisonkun: Aha. I get your point. Then let's keep this logic as is.

tisonkun (approved):

git-hulk (approved):
LGTM 👍

PragmaTwice:
Thanks all! Merging...

PR #771, opened 2022-08-05
Add GETDEL command

Ranxy:
Added support for GETDEL. but there is a problem with tcl tests, one of the tests in the test set is not turned on because the `config set` does not currently support `repl-ping-replica-period`

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ $ redis-cli -p 6666
 ### Running test cases
 
 ```shell
-$ ./build.sh build --unittest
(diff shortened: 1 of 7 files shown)


git-hulk:
@Ranxy Thanks for your contribution~

git-hulk (approved):

ShooterIT (approved):

git-hulk:
Thanks @Ranxy, will update `GetDel` command to the supported commands soon. cc @ShooterIT

PR #773, opened 2022-08-06
Remove generic parameter for `Popen` in `x.py`

PragmaTwice:
It seems generic parameter for `Popen` is not supported in python with version less than 3.8 (supported in python >= 3.9).

Hence the docker build process will fail because `ubuntu:focal` has python 3.8 in its apt repo, which makes the nightly build fail to push to the docker hub. :rofl:

### DIFF
--- a/x.py
+++ b/x.py
@@ -51,7 +51,7 @@ SEMVER_REGEX = re.compile(
     re.VERBOSE,
 )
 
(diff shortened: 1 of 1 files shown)


git-hulk (approved):
LGTM

tisonkun (approved):

tisonkun:
Thank you!

PR #781, opened 2022-08-17
replace new instead of std::make_unique or std::make_shared

ghost:
1. replaced new instead of std::make_unique or std::make_shared 
2. replace some NULLs instead of nullptr
3. replace condition == false instead of !condition

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -71,11 +71,11 @@ uint64_t GetTimeStampUS(void);
 // define std::make_unique in c++14
 // refer to https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
 template <typename T, typename... Args>
-inline std::unique_ptr<T> MakeUnique(Args&& ... args) {
+std::unique_ptr<T> MakeUnique(Args&& ... args) {
   return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
 }
 template <typename T, typename... Args>
-inline std::shared_ptr<T> MakeShared(Args&& ... args) {
+std::shared_ptr<T> MakeShared(Args&& ... args) {
   return std::shared_ptr<T>(new T(std::forward<Args>(args)...));
 }
 }  // namespace Util
--- a/src/util.h
+++ b/src/util.h
@@ -77,6 +77,6 @@ std::unique_ptr<T> MakeUnique(Args&& ... args) {
 
 template <typename T, typename... Args>
 std::shared_ptr<T> MakeShared(Args&& ... args) {
-  return std::shared_ptr<T>(new T(std::forward<Args>(args)...));
+  return std::make_shared<T>(std::forward<Args>(args)...);
 }
 }  // namespace Util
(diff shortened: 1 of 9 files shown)


PragmaTwice:
Hi, thanks for your contribution.

Since there is no `std::make_unique` in c++11, you can use `Util::MakeUnique` in `util.h` as an alternative.

git-hulk:
Hello @Furud0Er1ka, I saw many places are only formatted without changing the logic, I think we can rollback those format codes to keep this PR consistent with what's title described, it also helps the code review easier.

src/util.h
```diff
@@ -71,8 +76,14 @@ uint64_t GetTimeStampUS(void);
...
 }
 
-}  // namespace Util
+// define std::make_shared in c++14
+// refer to https://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared
+template <typename T, typename... Args>
+inline std::shared_ptr<T> MakeShared(Args &&...args) {
+  return std::shared_ptr<T>(new T(std::forward<Args>(args)...));
+}
+} // namespace Util
```
PragmaTwice: `std::make_shared` is in c++11, not c++14, so I think we do not need it.

src/util.h
```diff
@@ -71,8 +76,14 @@ uint64_t GetTimeStampUS(void);
 // define std::make_unique in c++14
 // refer to https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique
 template <typename T, typename... Args>
-std::unique_ptr<T> MakeUnique(Args&& ... args) {
+inline std::unique_ptr<T> MakeUnique(Args &&...args) {
```
PragmaTwice: This `inline` is useless. A function template does already imply `inline`.

PragmaTwice:
Hi, as mentioned by @git-hulk, I think you could minimize your PR, and you can check here to see why cpplint failed: https://github.com/apache/incubator-kvrocks/runs/7902217149?check_suite_focus=true

ghost:
> Hi, as mentioned by @git-hulk, I think you could minimize your PR, and you can check here to see why cpplint failed: https://github.com/apache/incubator-kvrocks/runs/7902217149?check_suite_focus=true

thanks for your review! I have removed all unnecessary changes.

src/util.h
```diff
@@ -74,5 +74,8 @@ template <typename T, typename... Args>
 std::unique_ptr<T> MakeUnique(Args&& ... args) {
   return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
 }
-
+template <typename T, typename... Args>
```
git-hulk, line 78:
```suggestion

template <typename T, typename... Args>
```

git-hulk:
cool, thanks

git-hulk (dismissed):

src/util.h
```diff
@@ -75,4 +75,8 @@ std::unique_ptr<T> MakeUnique(Args&& ... args) {
   return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
 }
 
+template <typename T, typename... Args>
+std::shared_ptr<T> MakeShared(Args&& ... args) {
+  return std::shared_ptr<T>(new T(std::forward<Args>(args)...));
```
PragmaTwice:
```suggestion
  return std::make_shared<T>(std::forward<Args>(args)...);
```

PragmaTwice (approved):

git-hulk (approved):

PragmaTwice:
Thanks for your contribution again! Merging...

PR #787, opened 2022-08-22
Add newly designed `ParseInt`

PragmaTwice:
It is the first (small) step to construct the new command parsing module. 

- add `TryParseInt<T>` and `ParseInt<T>` and unit tests
- use union in `StatusOr` since it is more readable and no strict aliasing issue

### DIFF
--- a/src/parse_util.h
+++ b/src/parse_util.h
@@ -71,7 +71,7 @@ struct ParseIntFunc<unsigned long long> { // NOLINT
   constexpr static const auto value = std::strtoull;
 };
 
-} // namespace details
+}  // namespace details
 
 template <typename T>
 using ParseResultAndPos = std::tuple<T, const char *>;
@@ -88,7 +88,7 @@ StatusOr<ParseResultAndPos<T>> TryParseInt(const char *v, int base = 0) {
   }
 
   if (errno) {
-    return {Status::NotOK, std::string{"TryParseInt: "} + std::strerror(errno)};
+    return {Status::NotOK, std::string {"TryParseInt: "} + std::strerror(errno)};
   }
 
   if (!std::is_same<T, decltype(res)>::value &&
--- a/src/parse_util.h
+++ b/src/parse_util.h
@@ -20,9 +20,11 @@
 
 #pragma once
 
-#include <limits>
-#include <status.h>
 #include <cstdlib>
+#include <limits>
+#include <string>
+#include <tuple>
+#include <status.h>
 
 namespace details {
 
@@ -30,7 +32,7 @@ template <typename>
 struct ParseIntFunc;
 
 template <>
-struct ParseIntFunc<short> {
+struct ParseIntFunc<short> { // NOLINT
   constexpr static const auto value = std::strtol;
 };
 
@@ -40,17 +42,17 @@ struct ParseIntFunc<int> {
 };
 
 template <>
-struct ParseIntFunc<long> {
+struct ParseIntFunc<long> { // NOLINT
   constexpr static const auto value = std::strtol;
 };
 
 template <>
-struct ParseIntFunc<long long> {
+struct ParseIntFunc<long long> { // NOLINT
   constexpr static const auto value = std::strtoll;
 };
 
 template <>
-struct ParseIntFunc<unsigned short> {
+struct ParseIntFunc<unsigned short> { // NOLINT
   constexpr static const auto value = std::strtoul;
 };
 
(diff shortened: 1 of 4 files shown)


src/parse_util.h
```diff
@@ -0,0 +1,130 @@
...
+    return {Status::NotOK, "TryParseInt: not started as an integer"};
+  }
+
+  if (errno) {
+    return {Status::NotOK, std::string {"TryParseInt: "} + std::strerror(errno)};
+  }
+
+  if (!std::is_same<T, decltype(res)>::value &&
+    (res < std::numeric_limits<T>::min() || res > std::numeric_limits<T>::max())) {
+    return {Status::NotOK, "TryParseInt: out of range of integer type"};
```
caipengbo: Personally, I don't think we need to prefix the function name in the Status information returned, which is probably a little redundant. 
This is what the caller needs to add, he knows what function he is using.

src/parse_util.h
```diff
@@ -0,0 +1,130 @@
...
+  constexpr static const auto value = std::strtoull;
+};
+
+}  // namespace details
+
+template <typename T>
+using ParseResultAndPos = std::tuple<T, const char *>;
+
+template <typename T = long long> // NOLINT
+StatusOr<ParseResultAndPos<T>> TryParseInt(const char *v, int base = 0) {
```
caipengbo, line 86: Should we put TryParseInt under ParseInt?  Normal first, then special, which I think is logical.

And write a few simple comments on TryParseInt to explain the difference between TryParseInt and ParseInt?
It was in the test cases that I learned the difference.

src/parse_util.h
```diff
@@ -0,0 +1,130 @@
...
+
+  if (v == end) {
+    return {Status::NotOK, "TryParseInt: not started as an integer"};
+  }
+
+  if (errno) {
+    return {Status::NotOK, std::string {"TryParseInt: "} + std::strerror(errno)};
+  }
+
+  if (!std::is_same<T, decltype(res)>::value &&
```
w4096, line 100: This is unnecessary, if the parsed result is out or range, errno will be set as ERANGE.

PragmaTwice: Hi, this is actually necessary. If `T` is `int` or `short`, `strtol` will be used, so `decltype(res)` is `long` (there is no `strtoi` or `strtos`). Actually the implementation of `sto*` in libstdc++ do the totally same thing as here. 

https://code.woboq.org/gcc/libstdc++-v3/include/ext/string_conversions.h.html#73

src/parse_util.h
PragmaTwice: Yeah, I also think it need some comment, I'll add later.

`ParseInt` is implemented by `TryParseInt`, so  `TryParseInt` should be declared above.

src/parse_util.h
fuzhe1989: What if `res` and `T` are signed and unsigned? Guess more careful type check is needed here.

PragmaTwice: Hi @fuzhe1989 , it will not happen, unsigned integers will be parsed by `strtou*`, you could check here:
https://github.com/apache/incubator-kvrocks/pull/787/files#diff-01299a1bd002d7db4c7c4e6c059d1a6a97ac2abea9042acafc5473e474f87d85R34-R72

though some static asserts can be put to ensure it.

fuzhe1989: got it

w4096: @PragmaTwice You are right, I have ignored the return type of ParseIntFunc<T>::value, I miss understand it and think the return type is T.

I think the value of `end` should be checked, if end != v + strlen(v), this means we meet something like "123abc". In this situation should we return an error ?

PragmaTwice: This is the difference between `TryParseInt` and `ParseInt`. `ParseInt` check it and return error while `end != input + size`, but `TryParseInt` just return the `end` pointer, to cover more situation.

src/parse_util.h
PragmaTwice: done

src/parse_util.h
PragmaTwice: done

caipengbo (approved):

tisonkun (commented):
Could you link this PR to the master issue? And is there a description about "the new command parsing module"?

PragmaTwice:
> Could you link this PR to the master issue? And is there a description about "the new command parsing module"?

Hi @tisonkun , there is an old issue #598 to describe it but maybe a little out-dated, and I will create a new issue.

git-hulk (approved):

PragmaTwice:
Thanks all! Merging...

PR #789, opened 2022-08-23
Replace `text` with `universal_newlines` to support python<3.7 in `x.py`

PragmaTwice:
Parameter `text` in `Popen` is introduced in python 3.7, so we can replace it with `universal_newlines` to be compatible with older versions. I encounter this error in a `centos:7` docker container, and I remember some other guys used to report this error in IM.


tisonkun (approved):
Thanks!

git-hulk (approved):

PR #791, opened 2022-08-24
Use PutVarInt instead of PutFixed while encoding stream entry value

torwig:
As @ShooterIT recommended here: https://github.com/apache/incubator-kvrocks/pull/745
I changed `PutFixed32` to `PutVarint32` and `GetFixed32` to `GetVarint32`.
Encoding/decoding functions were 'borrowed' from RocksDB codebase.
Obviously, this change makes it impossible to decode entries if entries were encoded with `PutFixed32` (an error will be returned).
Is it too late to introduce such a change?

### DIFF
--- a/src/encoding.cc
+++ b/src/encoding.cc
@@ -272,38 +272,21 @@ double DecodeDouble(const char *ptr) {
   return value;
 }
 
-char* EncodeVarint32(char *dst, uint32_t v) {
+uint8_t* EncodeVarint32(uint8_t *dst, uint32_t v) {
   // Operate on characters as unsigneds
-  unsigned char* ptr = reinterpret_cast<unsigned char*>(dst);
-  static const int B = 128;
-  if (v < (1 << 7)) {
-    *(ptr++) = v;
-  } else if (v < (1 << 14)) {
-    *(ptr++) = v | B;
-    *(ptr++) = v >> 7;
-  } else if (v < (1 << 21)) {
-    *(ptr++) = v | B;
-    *(ptr++) = (v >> 7) | B;
-    *(ptr++) = v >> 14;
-  } else if (v < (1 << 28)) {
-    *(ptr++) = v | B;
-    *(ptr++) = (v >> 7) | B;
-    *(ptr++) = (v >> 14) | B;
-    *(ptr++) = v >> 21;
-  } else {
-    *(ptr++) = v | B;
-    *(ptr++) = (v >> 7) | B;
-    *(ptr++) = (v >> 14) | B;
-    *(ptr++) = (v >> 21) | B;
-    *(ptr++) = v >> 28;
-  }
-  return reinterpret_cast<char*>(ptr);
+  uint8_t* ptr = dst;
+  do {
+    *ptr = 0x80 | v;
+    v >>= 7, ++ptr;
+  } while (v != 0);
+  *(ptr - 1) &= 0x7F;
+  return ptr;
 }
 
 void PutVarint32(std::string *dst, uint32_t v) {
-  char buf[5];
-  char* ptr = EncodeVarint32(buf, v);
-  dst->append(buf, static_cast<size_t>(ptr - buf));
+  uint8_t buf[5];
+  uint8_t* ptr = EncodeVarint32(buf, v);
+  dst->append(reinterpret_cast<char*>(buf), static_cast<size_t>(ptr - buf));
 }
 
 const char* GetVarint32PtrFallback(const char *p, const char *limit, uint32_t *value) {
@@ -311,9 +294,9 @@ const char* GetVarint32PtrFallback(const char *p, const char *limit, uint32_t *v
   for (uint32_t shift = 0; shift <= 28 && p < limit; shift += 7) {
     uint32_t byte = *(reinterpret_cast<const unsigned char*>(p));
     p++;
-    if (byte & 128) {
+    if (byte & 0x80) {
       // More bytes are present
-      result |= ((byte & 127) << shift);
+      result |= ((byte & 0x7F) << shift);
     } else {
       result |= (byte << shift);
       *value = result;
@@ -326,7 +309,7 @@ const char* GetVarint32PtrFallback(const char *p, const char *limit, uint32_t *v
 const char* GetVarint32Ptr(const char *p, const char *limit, uint32_t *value) {
   if (p < limit) {
     uint32_t result = *(reinterpret_cast<const unsigned char*>(p));
-    if ((result & 128) == 0) {
+    if ((result & 0x80) == 0) {
       *value = result;
       return p + 1;
     }
--- a/src/encoding.cc
+++ b/src/encoding.cc
@@ -271,3 +271,77 @@ double DecodeDouble(const char *ptr) {
   memcpy(&value, &decoded, sizeof(value));
   return value;
 }
+
+char* EncodeVarint32(char *dst, uint32_t v) {
+  // Operate on characters as unsigneds
+  unsigned char* ptr = reinterpret_cast<unsigned char*>(dst);
+  static const int B = 128;
+  if (v < (1 << 7)) {
+    *(ptr++) = v;
+  } else if (v < (1 << 14)) {
+    *(ptr++) = v | B;
+    *(ptr++) = v >> 7;
+  } else if (v < (1 << 21)) {
+    *(ptr++) = v | B;
+    *(ptr++) = (v >> 7) | B;
+    *(ptr++) = v >> 14;
+  } else if (v < (1 << 28)) {
+    *(ptr++) = v | B;
+    *(ptr++) = (v >> 7) | B;
+    *(ptr++) = (v >> 14) | B;
+    *(ptr++) = v >> 21;
+  } else {
+    *(ptr++) = v | B;
+    *(ptr++) = (v >> 7) | B;
+    *(ptr++) = (v >> 14) | B;
+    *(ptr++) = (v >> 21) | B;
+    *(ptr++) = v >> 28;
+  }
+  return reinterpret_cast<char*>(ptr);
+}
+
+void PutVarint32(std::string *dst, uint32_t v) {
+  char buf[5];
+  char* ptr = EncodeVarint32(buf, v);
+  dst->append(buf, static_cast<size_t>(ptr - buf));
+}
+
+const char* GetVarint32PtrFallback(const char *p, const char *limit, uint32_t *value) {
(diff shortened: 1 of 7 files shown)


ShooterIT:
> Is it too late to introduce such a change?

I don't think it is late, since now it is only on `unstable` branch which we may make changes on it.

one problem we need to concern is about open source license about these codes copied from leveldb/rocksdb.
should we put its license header. And is `xxFixed32` copied from leveldb/rocksdb, should we describe this? @git-hulk

git-hulk:
@torwig Thanks for your contribution. As @ShooterIT mentioned that we don't promise stability and compatibility before releasing, so feel free to do that.

@ShooterIT Those Put|GetFixedXXX are NOT copied from rocksdb/leveldb indeed, we rewritten them into BIGENDIAN while the rocksdb is little endian, so I think it's good to keep Apache 2.0 statement only. https://github.com/facebook/rocksdb/blob/0050a73a4fb1ba52d78655b271a0f09ba1fbf7fc/util/coding_lean.h#L25

src/encoding.cc
```diff
@@ -271,3 +271,77 @@ double DecodeDouble(const char *ptr) {
   memcpy(&value, &decoded, sizeof(value));
   return value;
 }
+
+char* EncodeVarint32(char *dst, uint32_t v) {
```
PragmaTwice, line 275: No logical mistake found. I just wonder if this loop unrolling will be more efficient, e.g. protobuf does not seem to unroll it. https://sourcegraph.com/github.com/protocolbuffers/protobuf/-/blob/src/google/protobuf/io/coded_stream.h?L898:42&subtree=true

src/encoding.cc
```diff
@@ -271,3 +271,77 @@ double DecodeDouble(const char *ptr) {
...
+  // Operate on characters as unsigneds
+  unsigned char* ptr = reinterpret_cast<unsigned char*>(dst);
+  static const int B = 128;
```
PragmaTwice: I prefer `0b1000'0000` or `0x80` rather than decimal form here.

PragmaTwice: And same as some other decimal numbers for bit-wise computation below, except these n for x <<(or >>) n.

src/encoding.cc
PragmaTwice: I perform a simple benchmark here: https://quick-bench.com/q/NlktQnulDGxmP1pcIIK6MLvjldk

It seems that this unrolling do not take much performance improvement.

Maybe my benchmark has some flaw, so you guys can review it and correct me : )

cc @ShooterIT @git-hulk

PragmaTwice (commented):
Other code looks good to me. Thanks for your contribution!

And you could add some unit test for `*Varint` functions.

src/encoding.cc
```diff
@@ -271,3 +271,77 @@ double DecodeDouble(const char *ptr) {
...
+    *(ptr++) = (v >> 21) | B;
+    *(ptr++) = v >> 28;
+  }
+  return reinterpret_cast<char*>(ptr);
+}
+
+void PutVarint32(std::string *dst, uint32_t v) {
+  char buf[5];
+  char* ptr = EncodeVarint32(buf, v);
+  dst->append(buf, static_cast<size_t>(ptr - buf));
```
PragmaTwice, line 289: You can cast `buf` here and change these internal function signatures to something like `uint8_t* EncodeVarint32(uint8_t *dst, uint32_t v)`, to avoid many `reinterpret_cast`.

src/encoding.cc
git-hulk: The benchmark looks good, NoUnroll even faster than Unroll in GCC.

src/encoding.cc
git-hulk: Yes, I also prefer to `0x80` over 128. We still need to use `ob1000000` if want to use binary representation since the `'` digital separator was introduced in C++14.

torwig: Refactored this function. Replace it to unrolled version.

src/encoding.cc
torwig: Done

src/encoding.cc
torwig: Done

git-hulk (dismissed):
LGTM, thanks

src/encoding.cc
```diff
@@ -271,3 +271,60 @@ double DecodeDouble(const char *ptr) {
...
+  for (uint32_t shift = 0; shift <= 28 && p < limit; shift += 7) {
+    uint32_t byte = *(reinterpret_cast<const unsigned char*>(p));
+    p++;
+    if (byte & 0x80) {
+      // More bytes are present
+      result |= ((byte & 0x7F) << shift);
+    } else {
+      result |= (byte << shift);
+      *value = result;
+      return reinterpret_cast<const char*>(p);
```
PragmaTwice:
```suggestion
      return p;
```

src/encoding.cc
```diff
@@ -271,3 +271,60 @@ double DecodeDouble(const char *ptr) {
...
+void PutVarint32(std::string *dst, uint32_t v) {
+  uint8_t buf[5];
+  uint8_t* ptr = EncodeVarint32(buf, v);
+  dst->append(reinterpret_cast<char*>(buf), static_cast<size_t>(ptr - buf));
+}
+
+const char* GetVarint32PtrFallback(const char *p, const char *limit, uint32_t *value) {
+  uint32_t result = 0;
+  for (uint32_t shift = 0; shift <= 28 && p < limit; shift += 7) {
+    uint32_t byte = *(reinterpret_cast<const unsigned char*>(p));
```
PragmaTwice:
```suggestion
    uint32_t byte = static_cast<unsigned char>(*p);
```

src/encoding.cc
```diff
@@ -271,3 +271,60 @@ double DecodeDouble(const char *ptr) {
...
+      *value = result;
+      return p;
+    }
+  }
+  return nullptr;
+}
+
+const char* GetVarint32Ptr(const char *p, const char *limit, uint32_t *value) {
+  if (p < limit) {
+    uint32_t result = *(reinterpret_cast<const unsigned char*>(p));
```
PragmaTwice:
```suggestion
    uint32_t result = static_cast<unsigned char>(*p);
```

PragmaTwice (dismissed):
LGTM

src/encoding.cc
ShooterIT: maybe many conditional judgments in unrolling break pipelining processing, this minor computation doesn't influence performance. And now @torwig implement new code, also is free from license problem.

src/encoding.cc
ShooterIT: I think original way is better, `char*` type is more universal, also use `unsigned char` if using `char`.

PragmaTwice: If you mean the difference between `uint8_t` and `unsigned char`, I think they are both OK. I do not prefer anyone, I use `uint8_t` just because it is short. I know `uint8_t` is not guaranteed to be equal to `unsigned char` since 1 byte is not guaranteed to be equal to 8 bits in some platform, but this function obviously treat char as 8 bits, since the function name is `...32`, so just use `unsigned char` will still make this function fail in those platform.

ShooterIT: no, i want to define function as `char* EncodeVarint32(char *dst, uint32_t v)`. If we agree, we use `unsiged char` in `EncodeVarint32`.

PragmaTwice: Fine, I have no opinion, since this is just a small change : )

torwig: @ShooterIT Do you mean code should look like the following:

```
char* EncodeVarint32(char *dst, uint32_t v) {
  // Operate on characters as unsigneds
  unsigned char* ptr = static_cast<unsigned char*>(dst);
  do {
    *ptr = 0x80 | v;
    v >>= 7, ++ptr;
  } while (v != 0);
  *(ptr - 1) &= 0x7F;
  return ptr;
}

void PutVarint32(std::string *dst, uint32_t v) {
  char buf[5];
  char* ptr = EncodeVarint32(buf, v);
  dst->append(buf, static_cast<size_t>(ptr - buf));
}
```

Because I think I didn't get your idea :)

ShooterIT: oh, yes, i want to keep  the interface like `EncodeVarint32` of leveldb/rocksdb

torwig: Done

torwig:
@git-hulk BTW, after reverting snappy version I have Cmake error on Fedora 35

```
CMake Error at /usr/lib64/cmake/GTest/GTestTargets.cmake:37 (message):

  /usr/lib64/cmake/GTest/GTestConfig.cmake:42 (include)
  /usr/share/cmake/Modules/FindGTest.cmake:187 (find_package)
  build/_deps/snappy-src/CMakeLists.txt:40 (find_package)
```
However, I see that all builds you have in your CI are successful.

PragmaTwice (approved):

PragmaTwice:
> @git-hulk BTW, after reverting snappy version I have Cmake error on Fedora 35
> 
> ```
> CMake Error at /usr/lib64/cmake/GTest/GTestTargets.cmake:37 (message):
>   Some (but not all) targets in this export set were already defined.
> 
>   Targets Defined: GTest::gtest;GTest::gtest_main
> 
>   Targets not yet defined: GTest::gmock;GTest::gmock_main
> 
> Call Stack (most recent call first):
>   /usr/lib64/cmake/GTest/GTestConfig.cmake:42 (include)
>   /usr/share/cmake/Modules/FindGTest.cmake:187 (find_package)
>   build/_deps/snappy-src/CMakeLists.txt:40 (find_package)
> ```
> 
> However, I see that all builds you have in your CI are successful.

There are some cmake changes from Snappy 1.1.7 to 1.1.9, `find_package(gtest)` always exists in 1.1.7 regardless of `SNAPPY_BUILD_TESTS` being set to `OFF`, which is fixed in 1.1.9.

@torwig Could you open an issue for this? I will fix it.

PragmaTwice:
And if you want to workaround it now, you can just remove `find_package(GTest QUIET)` in `<build dir>/_deps/snappy-src/CMakeLists.txt`.

torwig:
@PragmaTwice I've created an issue: https://github.com/apache/incubator-kvrocks/issues/792

git-hulk (approved):

git-hulk:
@ShooterIT Can have a look again and merge if have no comment.

ShooterIT (approved):

PR #793, opened 2022-08-26
Optimize get bitmap string and remove redundant num2bit array

ColinChamber:


src/redis_bitmap.cc
```diff
@@ -146,16 +146,27 @@ rocksdb::Status Bitmap::GetString(const Slice &user_key, const uint32_t max_btos
 
     //  If you setbit bit 0 1, the value is stored as 0x01 in Kvrocks but 0x80 in Redis.
     // So we need to swap bits is to keep the same return value as Redis.
+    static const uint8_t swap_table[256] = {
```
PragmaTwice, line 156: Would you like to add some comments to describe how to generate the table? i.e. the correspending mapping definition:
```
// swap_table(x) = (x & 0x80) >> 7 | ...
```

ColinChamber: Thanks for your review. Is there a more elegant way of compile-time calculations to generate this swap_table? @PragmaTwice

PragmaTwice (approved):

src/redis_bitmap.cc
PragmaTwice: Yeah, you can actually use templates. 

For more details, you can refer to [this commit in doris](https://github.com/apache/doris/commit/f49284036e1ed2bb5597cd6472720552a9c19054), which has samiliar ideas.

The commit above is written in c++17, so some modification is needed.

ColinChamber: I tried to use templates before this commit, but it didn't work out due to a lack of mastery. Thanks for your reference. I will learn it.

git-hulk: I think `swap_table` is NOT easy to maintain and previous way looks better than now. You change this for performance concern?

ColinChamber: Yeah, the previous way is more readable. But doing these operations at runtime is unnecessary and will significantly degrade performance.We can easily use redis-benchmark for comparison.
e.g.

1. Write data at random offsets.
```sh
redis-benchmark -r 9000000 -n 200000 -h 127.0.0.1 -p 6666 eval 'local source = ARGV[1];local offset,num = string.gsub(source,"0","");if(num == 12) then offset = "0";end;return redis.call("setbit", "foo", offset, 1)' 0  __rand_int__
```
2. Get and compare QPS before and after optimization.
```sh
redis-benchmark -n 200 -h 127.0.0.1 -p 6666 eval 'return redis.call("get", "foo")' 0
```
In my environment
**previous way : 53.30 requests per second**
**now : 80.39 requests per second**

We cannot ignore this performance gap. So I think the previous implementation seems inappropriate, although no one complains. @git-hulk

git-hulk: @ColinChamber Thanks for your improvement report, it also makes sense for me since it can improve the performance a lot.

git-hulk (approved):

src/redis_bitmap.cc
ShooterIT: only 80.39 requests per second?

ColinChamber: It depends on the running environment and the bitcount of the value. I was running it on a local virtual machine, and the parameters -r -n of the first step are not little. You can reproduce it in your environment. Maybe there are other things to improve.

ColinChamber: By the way,  this command was executed in Lua script.

ShooterIT (approved):

PR #803, opened 2022-09-01
Fix CMake configuration error in snappy

### LINKED ISSUE #792: CMake error building on Fedora 35
### Search before asking

### Version

Unstable branch: fac42b092810f7273385cd57e881c5e4b91e1158
OS: Fedora 35

### Minimal reproduce step

Clone repo.

mkdir build
cd build
cmake ../

No errors

### What did you see instead?

```
CMake Error at /usr/lib64/cmake/GTest/GTestTargets.cmake:37 (message):

  /usr/lib64/cmake/GTest/GTestConfig.cmake:42 (include)
  /usr/share/cmake/Modules/FindGTest.cmake:187 (find_package)
  build/_deps/snappy-src/CMakeLists.txt:40 (find_package)
```

### Anything Else?

_No response_

PragmaTwice:
This closes #792.

### DIFF
--- a/cmake/modules/FindGTest.cmake
+++ b/cmake/modules/FindGTest.cmake
@@ -15,18 +15,6 @@
 # specific language governing permissions and limitations
 # under the License.
 
-include_guard()
+# used for `find_package(GTest)` mechanism in snappy
 
-include(cmake/utils.cmake)
-
-FetchContent_DeclareGitHubWithMirror(snappy
-  google/snappy 1.1.7
-  MD5=40a371a653b7bb5a47df59b937b78557
-)
-
-FetchContent_MakeAvailableWithArgs(snappy
-  CMAKE_MODULE_PATH=${PROJECT_SOURCE_DIR}/cmake/modules
-  SNAPPY_BUILD_TESTS=OFF
-  SNAPPY_BUILD_BENCHMARKS=OFF
-  BUILD_SHARED_LIBS=OFF
-)
+set(GTEST_FOUND FALSE)
--- a/cmake/modules/FindGflags.cmake
+++ b/cmake/modules/FindGflags.cmake
@@ -15,18 +15,6 @@
 # specific language governing permissions and limitations
 # under the License.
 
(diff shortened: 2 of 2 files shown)


PragmaTwice:
@torwig @aleksraiden Hi, could you guys help me to check whether this PR really fixed the CMake error on Fedora or Debian? So thanks!

Here is a simple script for reference to check it:
```sh
git clone -b fix-792 https://github.com/PragmaTwice/incubator-kvrocks
cd incubator-kvrocks && ./x.py build
```

tisonkun (commented):
May you describe a bit how adding these two files helps resolve the issue?

I think it's about CMake resolving order while it looks like we always set `GFLAGS_FOUND` & `GTEST_FOUND` to `false`.

PragmaTwice:
> May you describe a bit how adding these two files helps resolve the issue?
> 
> I think it's about CMake resolving order while it looks like we always set `GFLAGS_FOUND` & `GTEST_FOUND` to `false`.

By default, for `find_package(X)`, CMake will use the builtin `FindX.cmake` file (provided by CMake) to find the X package.
It will try to find the package in system, e.g. packages installed from package manager.
We overwrite this file to make sure CMake does not use some builtin logic to find the package, because we do not need these package in snappy.

It is actually a bug in (old version of) snappy, we defined `SNAPPY_BUILD_TESTS=OFF`, but snappy will always find gtest regardless of our provided option, and we do not need gtest in snappy since we do not build unit tests of snappy.

@tisonkun

tisonkun (approved):
Thanks for your explanation! Sounds good to me.

Although, I don't have a Fedora env to verify it :P

torwig:
@tisonkun I can confirm that the build process goes smoothly on Fedora 35 now.

tisonkun:
@torwig Thanks for your confirmation!

Merging...

aleksraiden:
@tisonkun I can confirm too, that the build process on Debian 11 are OK.

PR #806, opened 2022-09-01
refactor: write integration tests in Golang

tisonkun:
## Motivation

TCL scripts are hard for debugging. It helps to bring Redis tests but we can later refactor them for easy maintaining and debugging.

I ever try to write it in Rust https://github.com/apache/incubator-kvrocks/discussions/693 but finally propose this patch in Golang. As Rust's testing is less than mature and our contributors like @git-hulk & @torwig should be familiar with Golang. Also, go-redis project has [a backlink to Kvrocks](https://github.com/go-redis/redis#ecosystem) and when possibly, I'm glad to use it for Kvrocks' tests.

## Implementation

1. Add go integration cases mod under `tests/gocase`.
2. Implement the start server logic at `util/server.go`
3. Migrate TCL test `util/command` to the new integration framework.
4. Wire logic to the CI workflow.


git-hulk:
Thanks @tisonkun

The tcl is really hard to debug for myself. Go or Rust are good to me but I think Go will be easier for developers to learn and participate. Can see other guys have any feedback on this topic,

.github/workflows/kvrocks.yaml
```diff
@@ -160,6 +160,27 @@ jobs:
...
+        with:
+          go-version: 1.19.x
+
+      - name: Run Go Integration Cases
+        working-directory: tests/gocase
+        run: |
+          export ${{ matrix.runtime_env_vars }}
+          export KVROCKS_BIN_PATH="$GITHUB_WORKSPACE/build/kvrocks"
+          export GO_CASE_WORKSPACE="$GITHUB_WORKSPACE/tests/gocase/workspace"
+          go test -v ./...
```
PragmaTwice, line 174: Could these be added in `./x.py`, like `./x.py test go <build-dir>`?

tisonkun: Yes, it's possible. I'd like to do it as a follow-up, though.

PragmaTwice: got it

tests/tcl/tests/test_helper.tcl
```diff
@@ -56,7 +56,6 @@ set ::all_tests {
     unit/introspection
     unit/limits
     unit/geo
-    unit/command
```
PragmaTwice, line 59: If we delete it in the migration procedure, developers may need a two phase testing in their local:
1. run tcl tests
2. run go tests

Any idea to simplify the test executing?

tisonkun: If we don't delete it, the situation doesn't change. If only the migration can be eventually closed, I don't think it's a big deal.

Since we do have TCL tests and go IT cases, we always run both. There's no way to avoid it. Only you may batch them in one script call but it seems not necessary.

PragmaTwice: Sounds good to me.

tests/gocase/util/server.go
```diff
@@ -0,0 +1,119 @@
...
+	configs["bind"] = addr.IP.String()
+	configs["port"] = fmt.Sprintf("%d", addr.Port)
+
+	dir := os.Getenv("GO_CASE_WORKSPACE")
+	require.NoError(t, err)
+	dir, err = os.MkdirTemp(dir, "Server-*")
+	require.NoError(t, err)
+	configs["dir"] = dir
+
+	f, err := os.CreateTemp(dir, "*.conf")
```
PragmaTwice: I think `kvrocks.conf` is enough rather than a random string? We can locate it more easy if it has a fixed name.

tisonkun: Good point! Will push a commit for it.

tisonkun: Updated.

git-hulk:
I looked through this PR and it’s clear enough for me. To see other guys have any questions? @PragmaTwice @ShooterIT @Alfejik @caipengbo

PragmaTwice (approved):
LGTM

tisonkun:
It happens that I have some time to work on this migration. I'll merge this patch tomorrow if no more objections and continue the tasks.

Feel free to review after merge.

caipengbo (approved):

PR #808, opened 2022-09-03
Fix inline protocol can't parse LF only EOL

### LINKED ISSUE #807: redis-cli --pipe syntax error
### Search before asking

### Version

kvrocks unstable 40989fc26dc4229bd9f797c14eaa4df07574f47e
ubuntu 22.04

### Minimal reproduce step

```
root@localhost:~# cat /tmp/text1.txt
set zzzap zzzzzzzzzzzzzzzzznologin
set vv zzzzzzzzzzzzzzzzpsbinnologin
```

```
root@localhost:~# cat /tmp/text1.txt | redis-cli --pipe --verbose
```

This is with redis or keydb 

```
root@localhost:~# cat /tmp/text1.txt | redis-cli -h xx.79.179.186 --pipe --verbose
All data transferred. Waiting for the last reply...
Last reply received from server.
errors: 0, replies: 2
```

### What did you see instead?

root@localhost:~# cat /tmp/text1.txt | redis-cli --pipe --verbose
All data transferred. Waiting for the last reply...
ERR syntax error
Last reply received from server.
errors: 1, replies: 1

### Anything Else?

> redis-cli --version
redis-cli 6.0.16

git-hulk:
This closes #807

### DIFF
--- a/src/redis_request.cc
+++ b/src/redis_request.cc
@@ -43,7 +43,16 @@ Status Request::Tokenize(evbuffer *input) {
   while (true) {
     switch (state_) {
       case ArrayLen: {
-        UniqueEvbufReadln line(input, EVBUFFER_EOL_CRLF_STRICT);
+        bool isOnlyLF = true;
+        // We don't use the `EVBUFFER_EOL_CRLF_STRICT` here since only LF is allowed in INLINE protocol.
+        // So we need to search LF EOL and figure out current line has CR or not.
+        UniqueEvbufReadln line(input, EVBUFFER_EOL_LF);
+        if (line && line.length > 0 && line[line.length-1] == '\r') {
+          // remove `\r` if exists
+          --line.length;
+          isOnlyLF = false;
+        }
+
         if (!line || line.length <= 0) {
           if (pipeline_size > 128) {
             LOG(INFO) << "Large pipeline detected: " << pipeline_size;
@@ -53,6 +62,7 @@ Status Request::Tokenize(evbuffer *input) {
           }
           return Status::OK();
         }
+
         pipeline_size++;
         svr_->stats_.IncrInbondBytes(line.length);
         if (line[0] == '*') {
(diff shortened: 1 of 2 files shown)


tisonkun (commented):
In the [RESP spec](https://redis.io/docs/reference/protocol-spec/) it seems only CRLF("\r\n") is the allowed delimiter. Does Redis implement a different manner actually?

git-hulk:
> In the [RESP spec](https://redis.io/docs/reference/protocol-spec/) it seems only CRLF("\r\n") is the allowed delimiter. Does Redis implement a different manner actually?

Yes, RESP only allows CRLF but inline mode allows the LF end of line as well: https://github.com/redis/redis/blob/unstable/src/networking.c#L2080

src/redis_request.cc
```diff
@@ -61,13 +71,13 @@ Status Request::Tokenize(evbuffer *input) {
           } catch (std::exception &e) {
             return Status(Status::NotOK, "Protocol error: invalid multibulk length");
           }
+          if (isOnlyLF || multi_bulk_len_ > (int64_t)PROTO_MULTI_MAX_SIZE) {
+            return Status(Status::NotOK, "Protocol error: invalid multibulk length");
+          }
```
tisonkun, line 76: Why do we disallow `\n` as EOL in the multi-bulk scenario?

tisonkun (approved):

tisonkun (commented):
~~I try `echo -e 'PING' | nc localhost 6666` with empty reply while `echo -e 'PING' | nc localhost 6379` with `PONG`. Shall we handle no eol delimiter at the end case?~~

Sorry, I get it wrong. `echo -e` automatically appends a `\n`

src/redis_request.cc
git-hulk: Redis also disallowed \n in multi bulk protocol, so we just keep consistent with Redis here.

PragmaTwice (approved):

src/redis_request.cc
tisonkun: Thanks for your explanation. I get the point now.

PR #809, opened 2022-09-03
If unix socket is specified, don't listen default TCP if addr:port wasn't set

### LINKED ISSUE #805: Can't run only unixsocket, without any TCP.
### Search before asking

### Motivation

In local env, when we configure fro maximal performance, we want use only unixsocket, without any TCP. 
If I comment bind and port options and uncomment unixsocket, after start, kvrocks try to bing localhost:6666 port, ignoring disable option. In other hands, Redis works very well :)

### Solution

Not trying to bind TCP if both options are disabled (bind and port)

torwig:
This allows Kvrocks to listen to only the Unix socket

Close: #805

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -44,6 +44,8 @@ const char *errNotEnableBlobDB = "Must set rocksdb.enable_blob_files to yes firs
 const char *errNotSetLevelCompactionDynamicLevelBytes =
             "Must set rocksdb.level_compaction_dynamic_level_bytes yes first.";
 
+const char *kDefaultBindAddress = "0.0.0.0";
+
 configEnum compression_type_enum[] = {
     {"no", rocksdb::CompressionType::kNoCompression},
     {"snappy", rocksdb::CompressionType::kSnappyCompression},
@@ -94,8 +96,8 @@ Config::Config() {
   };
   FieldWrapper fields[] = {
       {"daemonize", true, new YesNoField(&daemonize, false)},
-      {"bind", true, new StringField(&binds_, "127.0.0.1")},
-      {"port", true, new IntField(&port, 6666, 1, 65535)},
+      {"bind", true, new StringField(&binds_, "")},
+      {"port", true, new IntField(&port, 0, 0, 65535)},
       {"workers", true, new IntField(&workers, 8, 1, 256)},
       {"timeout", false, new IntField(&timeout, 0, 0, INT_MAX)},
       {"tcp-backlog", true, new IntField(&backlog, 511, 0, INT_MAX)},
@@ -538,6 +540,21 @@ Status Config::finish() {
   }
   if ((cluster_enabled) && !tokens.empty()) {
     return Status(Status::NotOK, "enabled cluster mode wasn't allowed while the namespace exists");
+  }
+  if (unixsocket.empty()) {
+    if (binds.size() == 0) {
+      binds.emplace_back(kDefaultBindAddress);
+    }
+    if (port == 0) {
+      port = kDefaultPort;
+    }
+  }
+  if (cluster_enabled && binds.size() == 0) {
+    return Status(Status::NotOK, "node is in cluster mode, but TCP listen address "
+                                 "wasn't specified via configuration file.");
+  }
+  if (master_port != 0 && binds.size() == 0) {
+    return Status(Status::NotOK, "replication doesn't supports unix socket.");
   }
   if (db_dir.empty()) db_dir = dir + "/db";
   if (backup_dir.empty()) backup_dir = dir + "/backup";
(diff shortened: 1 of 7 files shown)


aleksraiden:
Awesome, @torwig, lot of thanks! Please, review @PragmaTwice @ShooterIT @git-hulk

src/config.cc
```diff
@@ -44,6 +44,8 @@ const char *errNotEnableBlobDB = "Must set rocksdb.enable_blob_files to yes firs
 const char *errNotSetLevelCompactionDynamicLevelBytes =
             "Must set rocksdb.level_compaction_dynamic_level_bytes yes first.";
 
+const char *kDefaultBindAddress = "0.0.0.0";
```
ShooterIT: i think default bind address should be "127.0.0.1" which is safer.

ShooterIT (commented):
cool job, thanks

src/config.cc
```diff
@@ -539,6 +538,21 @@ Status Config::finish() {
   if ((cluster_enabled) && !tokens.empty()) {
     return Status(Status::NotOK, "enabled cluster mode wasn't allowed while the namespace exists");
   }
+  if (unixsocket.empty()) {
+    if (binds.size() == 0) {
+      binds.emplace_back(kDefaultBindAddress);
+    }
+    if (port == 0) {
+      port = kDefaultPort;
+    }
```
PragmaTwice: I think if in current config, `port == 0` and `unixsocket` is empty, kvrocks would be better to report an error and then terminate, rather than assigning a default port.

src/config.cc
torwig: @ShooterIT Ok. However, the comment in config file states that: 

`By default kvrocks listens for connections from all the network interfaces available on the server.`

Should I change the related comment?

src/config.cc
torwig: Yes, that was my first intention. But currently, if the config is empty (I mean absolutely empty), kvrocks will start with defaults showing just the warning. If I report an error if port is 0 and unixsocket is empty, the behavior will be changed.

PragmaTwice: Let me see..

How about
```
{"port", true, new IntField(&port, 6666, 0, 65535)},
```
rather than
```
{"port", true, new IntField(&port, 0, 0, 65535)},
```
?

torwig: Yes, it may work. I'll check and add changes if everything is OK.

torwig: Done

src/config.cc
```diff
@@ -88,14 +88,13 @@ Config::Config() {
...
-                 ConfigField* field)
+    FieldWrapper(std::string name, bool readonly, ConfigField *field)
         : name(std::move(name)), readonly(readonly), field(field) {}
   };
   FieldWrapper fields[] = {
       {"daemonize", true, new YesNoField(&daemonize, false)},
-      {"bind", true, new StringField(&binds_, "127.0.0.1")},
-      {"port", true, new IntField(&port, 6666, 1, 65535)},
+      {"bind", true, new StringField(&binds_, "")},
+      {"port", true, new IntField(&port, kDefaultPort, 1, 65535)},
```
PragmaTwice, line 97:
```suggestion
      {"port", true, new IntField(&port, kDefaultPort, 0, 65535)},
```
typo?

torwig: Now `min` parameter should be `1`.

PragmaTwice: Oh I got it.

src/config.cc
torwig: @ShooterIT I changed the default bind address from `0.0.0.0` to `127.0.0.1` and edited the comment in `kvrocks.conf`.

ShooterIT: thanks @torwig i think old comment is wrong since default bind address is 127.0.0.1, also ping @git-hulk

git-hulk: Yes, the default bind address will be better since it can prevent users from starting with external IPs and getting security attacks.

src/config.cc
```diff
@@ -539,6 +538,16 @@ Status Config::finish() {
   if ((cluster_enabled) && !tokens.empty()) {
     return Status(Status::NotOK, "enabled cluster mode wasn't allowed while the namespace exists");
   }
+  if (unixsocket.empty() && binds.size() == 0) {
+    binds.emplace_back(kDefaultBindAddress);
+  }
+  if (cluster_enabled && binds.size() == 0) {
+    return Status(Status::NotOK, "node is in cluster mode, but TCP listen address "
+                                 "wasn't specified via configuration file.");
```
```suggestion
                                 "wasn't specified via configuration file");
```

src/config.cc
```diff
@@ -539,6 +538,16 @@ Status Config::finish() {
...
+  }
+  if (master_port != 0 && binds.size() == 0) {
+    return Status(Status::NotOK, "replication doesn't supports unix socket.");
```
```suggestion
    return Status(Status::NotOK, "replication doesn't supports unix socket");
```

git-hulk (approved):

ShooterIT (approved):

PragmaTwice (approved):

PR #810, opened 2022-09-04
refactor: unit/protocol to gocase

### LINKED ISSUE #812: Move TCL test unit/protocol to Go case

tisonkun:
This closes https://github.com/apache/incubator-kvrocks/issues/812.


tests/gocase/unit/protocol/regression_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	v = rdb.RPush(ctx, "handle", "a")
+	require.EqualValues(t, 1, v.Val())
+
+	// TODO should read the second pushed element
+	//for _, res := range resList {
+	//	r, err := c.ReadLine()
+	//	require.NoError(t, err)
+	//	require.Equal(t, res, r)
+	//}
```
tisonkun, line 63: This is not a regression because upstream tests only the equivalent to:

```go
require.EqualValues(t, 1, v.Val())
```

You can see this comment https://github.com/redis/redis/commit/4bb1ccc6ebaa5944f8283500d69cae505f9d04d8#r83003244.

If I send `BLPOP` commands after `RPUSH`, the reader can read both two responses.

I don't know whether it's a bug/incompatibility in Kvrocks, but it's not a blocker to this patch.

git-hulk (approved):

PR #827, opened 2022-09-06
Fix CMake policy CMP0135

### LINKED ISSUE #826: Warning on latest CMake 3.24.1 - DOWNLOAD_EXTRACT_TIMESTAMP option
### Search before asking

### Motivation

Building a KVRocks with latest CMake produced a lot of warnings like this: 

> 
> CMake Warning (dev) at /usr/local/share/cmake-3.24/Modules/FetchContent.cmake:1264 (message):
>   The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
>   not set.  The policy's OLD behavior will be used.  When using a URL
>   download, the timestamps of extracted files should preferably be that of
>   the time of extraction, otherwise code that depends on the extracted
>   contents might not be rebuilt if the URL changes.  The OLD behavior
>   preserves the timestamps from the archive instead, but this is usually not
>   what you want.  Update your project to the NEW behavior or specify the
>   DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
>   robustness issue.
> Call Stack (most recent call first):
>   cmake/utils.cmake:48 (FetchContent_Declare)
>   cmake/utils.cmake:55 (FetchContent_DeclareWithMirror)
>   cmake/luajit.cmake:22 (FetchContent_DeclareGitHubWithMirror)
>   CMakeLists.txt:88 (include)
> This warning is for project developers.  Use -Wno-dev to suppress it.

Link to this at CMake docs: [https://cmake.org/cmake/help/latest/policy/CMP0135.html](https://cmake.org/cmake/help/latest/policy/CMP0135.html)

### Solution

Add options DOWNLOAD_EXTRACT_TIMESTAMP with value NEW into FetchContent_Declare function (cmake/utils.cmake)

aleksraiden:
Set an DOWNLOAD_EXTRACT_TIMESTAMP options to new behavior (by cmake policy CMP0135). Close #826


aleksraiden:
Not so trivial, need to use macros for define options only for CMake > 3.24.0. I will patch this asap

aleksraiden:
Sorry, I Am newbie in CMake, but tried to make this clear as possible. In dedicated policy.cmake we can add any options in futures.

torwig:
Applying this patch doesn't cause any issues with the build process on Fedora 35 as well.

cmake/policy.cmake
```diff
@@ -0,0 +1,28 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# it must be a macro cause policies have scopes
+# http://www.cmake.org/cmake/help/v3.0/command/cmake_policy.html
+
+macro (kvrocks_policy)
+
+	# Avoid warning about DOWNLOAD_EXTRACT_TIMESTAMP in CMake 3.24:
```
torwig:
```suggestion
  # Avoid warning about DOWNLOAD_EXTRACT_TIMESTAMP in CMake 3.24:
```

cmake/policy.cmake
```diff
@@ -0,0 +1,28 @@
...
+	if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
```
torwig:
```suggestion
  if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
```

cmake/policy.cmake
```diff
@@ -0,0 +1,28 @@
...
+		cmake_policy(SET CMP0135 NEW)
```
torwig:
```suggestion
    cmake_policy(SET CMP0135 NEW)
```

cmake/policy.cmake
```diff
@@ -0,0 +1,28 @@
...
+	endif()
```
torwig:
```suggestion
  endif()
```

cmake/policy.cmake
```diff
@@ -0,0 +1,28 @@
...
+
```
torwig:
```suggestion
```

cmake/policy.cmake
```diff
@@ -0,0 +1,28 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# it must be a macro cause policies have scopes
+# http://www.cmake.org/cmake/help/v3.0/command/cmake_policy.html
+
+macro (kvrocks_policy)
+
```
torwig:
```suggestion
```

torwig (commented):
Small formatting changes

aleksraiden:
Lot of thanks @torwig

tisonkun (changes requested):
I suggest you simply add:

```cmake
if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
    cmake_policy(SET CMP0135 NEW)
endif()
```

in `CMakeLists.txt` under the root path and below `option`s section. It's a global setting and we don't have to "refactor it out". As you mentioned, these lines are scoped. So inlining is native IIUC.

Concretely:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index b6ffa0b..d453beb 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,10 @@ option(ASAN_WITH_LSAN "enable leak santinizer while address santinizer is enable
 option(ENABLE_STATIC_LIBSTDCXX "link kvrocks with static library of libstd++ instead of shared library" ON)
 option(USE_LUAJIT "use luaJIT instead of lua" ON)
 
+if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
+    cmake_policy(SET CMP0135 NEW)
+endif()
+
 set(DEPS_FETCH_PROXY "" CACHE STRING
     "a template URL to proxy the traffic for fetching dependencies, e.g. with DEPS_FETCH_PROXY = https://some-proxy/,
      https://example/some-dep.zip -> https://some-proxy/https://example/some-dep.zip")
```

aleksraiden:
Of course, your version is more clear. Lot of thanks.
But - if we need a other patch like this, this version isn't so good as a proposition to a split code into files, as I think. I tried to watch how  this work into an other top-level projects, like this - [ethereum/solidity](https://github.com/ethereum/solidity/blob/4159d13a996d8b376ae33f2be58b9651269daee4/cmake/EthPolicy.cmake) and they have a lot of other policy. 

But, yes, your patch is awesome simple and clean.

tisonkun:
> if we need a other patch like this

Let's factor policies out if this is true.

[GoogleTest](https://github.com/google/googletest/blob/0e0d9feefab1b51aaab9dfd70132e93c0b6964e5/CMakeLists.txt#L6-L16) follows this simple way, and so do libevent, lz4, rocksdb if you check our dependencies.

We don't have to follow other projects and these references are only for reference. My criteria are keeping simple things simple, and refactoring only if we do meet some problems.

tisonkun (approved):
LGTM. Waiting for CI green...

aleksraiden:
@tisonkun Done, lot of thanks

PragmaTwice (approved):

tisonkun:
@aleksraiden merged. Thanks for your contribution!

Issue #813, opened 2022-09-04
Integrate Go cases into build scripts


tisonkun:
I think basically we can treat Go cases as a standalone module and write a `Makefile` to it for running linters and tests.

Then we wire it into the `x.py` script.

PragmaTwice:
I am thinking whether the makefile is necessary, since it is just a test module and we already have `x.py` which is a general script for developers. 

If there are two scripts that need to be maintained, I think it might cause some confusion. The functionality of the makefile is not orthogonal to our current `x.py`, they have similiar capabilities.

tisonkun:
@PragmaTwice you're right. Previously I'm thinking of adding linters for Go code and then build/test can depend on the lint task. Makefile does a good job for task dependencies while implementing it inside a script is awkward. However, now I believe we can run lint in another `x.py` command and there's no dependency between tasks to handle. So your patch is great.

tisonkun:
Closed by #832.

PR #832, opened 2022-09-06
Add go test to x.py and improve error in gocase

PragmaTwice:


tests/gocase/util/server.go
```diff
@@ -59,6 +59,7 @@ func (s *KvrocksServer) Close() {
 
 func StartServer(t *testing.T, configs map[string]string) *KvrocksServer {
 	b := os.Getenv("KVROCKS_BIN_PATH")
+	require.NotEmpty(t, b, "please set the environment variable `KVROCKS_BIN_PATH`")
```
suica, line 62: can we make some guesses about the path or just provide a predefined value here, and only fail when we really cannot find the binary?

tests/gocase/util/server.go
```diff
@@ -67,7 +68,7 @@ func StartServer(t *testing.T, configs map[string]string) *KvrocksServer {
 	configs["port"] = fmt.Sprintf("%d", addr.Port)
 
 	dir := os.Getenv("GO_CASE_WORKSPACE")
-	require.NoError(t, err)
+	require.NotEmpty(t, dir, "please set the environment variable `GO_CASE_WORKSPACE`")
```
tisonkun, line 71: Good catch!

tisonkun (changes requested):
Changes in `util/server.go` look good to me.

For changes in `x.py`, you may take a look at https://github.com/apache/incubator-kvrocks/issues/813.

I'd prefer to merge changes in `util/server.go` and you can start a new patch about integrating Go cases into build scripts.

tests/gocase/util/server.go
tisonkun: Can you give some guesses here? Actually as suggested in #813 I think basically we can treat Go cases as a standalone module, then we make guesses in `x.py` since there's where we confirm we're in the kvrocks workspace.

x.py
```diff
@@ -306,6 +320,15 @@ def test_tcl(dir: str, rest: List[str]) -> None:
     parser_test_tcl.add_argument('rest', nargs=REMAINDER, help="the rest of arguments to forward to TCL scripts")
     parser_test_tcl.set_defaults(func=test_tcl)
 
+    parser_test_go = parser_test_subparsers.add_parser(
+        'go',
+        description="Test kvrocks via go test cases",
+        help="Test kvrocks via go test cases",
+    )
+    parser_test_go.add_argument('dir', metavar='BUILD_DIR', nargs='?', default='build', help="directory including kvrocks build files")
```
tisonkun, line 328: `require=True`?

tisonkun (commented):
Generally looks good to me now.

x.py
PragmaTwice: We treat `<repo-root>/build` as a default directory for building and testing, e.g.
```
./x.py build [build]
./x.py test tcl [build]
./x.py test go [build]
```
`[build]` can be omitted.

tisonkun: Oops. I omit the `default='build'` part.

tisonkun (dismissed):
LGTM.

PragmaTwice:
@tisonkun Added `-bench=.`

tisonkun (approved):

PragmaTwice:
Thanks all! Merging...

PR #833, opened 2022-09-06
Rewrite test unit/type/list-2 in Go

### LINKED ISSUE #817: Move TCL test unit/type/list-2 to Go case

suica:
fixes #817


tests/gocase/unit/type/list/list-2_test.go
```diff
@@ -0,0 +1,88 @@
...
+			myList = append(myList, largeStr)
+			for i := 0; i < 1000; i++ {
+				s := randInt64Str()
+				rdb.RPush(ctx, key, s)
+				myList = append(myList, s)
+			}
+			for i := 0; i < 1000; i++ {
+				low := int(rand.Float64() * float64(startLen))
+				high := int(float64(low) + rand.Float64()*float64(startLen))
+				myList = myList[low:min(high+1, len(myList))]
```
suica: Here I clamp the end index by `len(myList)`, since original tcl test case uses `lrange` function with an inclusive end index. Otherwise, when `cap(myList)>=high>len(myList)`, result of `myList[low:high+1]` will be not as expected or even panics if `cap(myList)<=high`.

tests/gocase/unit/type/list/list-2_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/apache/incubator-kvrocks/tests/gocase/util"
+	"github.com/stretchr/testify/require"
+)
+
+func randInt64Str() string {
+	return strconv.FormatInt(rand.Int63(), 10)
+}
+
+func min(a, b int) int {
```
suica: Is it ok to place utils here, or such things should be placed elsewhere?

git-hulk: How about moving them into [tests/gocase/util](https://github.com/apache/incubator-kvrocks/tree/unstable/tests/gocase/util)?

PragmaTwice: It will be wonderful if you could implement a generic version and put into tests/gocase/util. 

i.e. `func min[T constraints.Ordered](a, b T) T`

tisonkun:
For the first time I know that Go files can have `-` in name :)

tests/gocase/unit/type/list/list-2_test.go
tisonkun: I'd prefer to use an existing mature lib like https://pkg.go.dev/modernc.org/mathutil#Min.

tests/tcl/tests/unit/type/list-2.tcl
```diff
@@ -1,70 +0,0 @@
...
-# See bundled license file licenses/LICENSE.redis for details.
-
-# This file is copied and modified from the Redis project,
-# which started out as: https://github.com/redis/redis/blob/dbcc0a8/tests/unit/type/list-2.tcl
-
-start_server {
-    tags {"list"}
-    overrides {
-        "list-max-ziplist-size" 4
-    }
```
tisonkun, line 28: Interesting. It seems we don't populate this config in the new test case but it still passes.

tisonkun: To populate config, see:

https://github.com/apache/incubator-kvrocks/blob/3f0fb401030332aff76a4d2638887cbf0de240fc/tests/gocase/unit/auth/auth_test.go#L45-L47

tests/gocase/unit/type/list/list-2_test.go
suica: > I'd prefer to use an existing mature lib like https://pkg.go.dev/modernc.org/mathutil#Min.

done

tests/gocase/util/random.go
```diff
@@ -0,0 +1,29 @@
...
+package util
+
+import (
+	"math/rand"
+	"strconv"
+)
+
+func RandInt64Str() string {
+	return strconv.FormatInt(rand.Int63(), 10)
+}
```
tisonkun: You can just inline this oneliner instead of a handy utility.

tests/tcl/tests/unit/type/list-2.tcl
suica: fixed

tests/gocase/util/random.go
suica: done

suica:
> For the first time I know that Go files can have `-` in name :)

Maybe we need to use `list_common.go` & `list_2_test.go` to comply the Go naming conventions?

tisonkun:
@suica I don't have a strong feeling as long as it can work.

My preference is:

```
list
- common.go
- list_test.go
```

suica:
@tisonkun Oh yes, that naming would be more concise and consistent. 
And since the file is about the cases from `list-2.tcl`, how about keeping the number in its name like this?

```
list
- common.go
- list_2_test.go
```

tisonkun:
@suica I guess those numbers are there for historical issues. We can firstly add them add together, and if there're some issues we should split tests, we write it with a meaningful name instead of counting.

Previously, I split command.tcl to package command and info, and protocol.tcl into two files.

tests/gocase/go.mod
```diff
@@ -7,10 +7,13 @@ require (
 	github.com/stretchr/testify v1.8.0
 )
 
+require github.com/remyoudompheng/bigfft v0.0.0-20200410134404-eec4a21b6bb0 // indirect
```
PragmaTwice: Just curious, why there is a new `require` clause rather than a single `require` with all indirect packages? :rofl:

tests/gocase/go.mod
```diff
@@ -7,10 +7,13 @@ require (
...
+
 require (
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
+	modernc.org/mathutil v1.5.0
```
PragmaTwice: I think you can put it in the first `require` clause, since it is not an indirect package.

tests/gocase/go.mod
suica: Oops, maybe it is because that i misplaced the `modernc.org/mathutil v1.5.0` to the indirect packages area... I'll fix both of them.

tests/gocase/go.mod
suica: also fixed

tisonkun:
@suica I send a patch onto your PR https://github.com/suica/incubator-kvrocks/pull/1.

cc @PragmaTwice I notice that test like "LTRIM stress testing" can adopt a Go benchmark flavor and when you wire it into `x.py`, you may think of add `-bench=.` if possible.

PragmaTwice:
> cc @PragmaTwice I notice that test like "LTRIM stress testing" can adopt a Go benchmark flavor and when you wire it into `x.py`, you may think of add `-bench=.` if possible.

The script will forward all unparsed options, e.g.
```
./x.py test go <build-dir> -bench=...
```
equals to
```
cd ...
KVROCKS_BIN_PATH=... GO_CASE_WORKSPACE=... go test -v ./... -bench=...
```

---

Oh, I think you may mean that we can make `-bench=.` a default option?

tisonkun:
@PragmaTwice yes. We can discuss in detail in the dedicated PR. Just FYI.

tisonkun (approved):

tisonkun:

PR #834, opened 2022-09-07
ci: add golangci_lint 

tisonkun:

PR #835, opened 2022-09-07
Move TCL test unit/type/list-3 to Go case

### LINKED ISSUE #818: Move TCL test unit/type/list-3 to Go case

suica:
closes #818


tisonkun:
You can rebase on latest unstable version to catch up #834 for golangci-lint reports.

tests/gocase/unit/type/list/list_test.go
```diff
@@ -84,3 +84,154 @@ func BenchmarkLTRIM(b *testing.B) {
...
+				} else {
+					where = "after"
+				}
+				require.NoError(t, rdb.LInsert(ctx, key, where, otherEle, ele).Err())
+			}
+		}
+	})
+
+	t.Run("ziplist implementation: value encoding and backlink", func(t *testing.T) {
+		// TODO: check --accurate options for go test
```
suica: TCL testing has `--accurate` flag. When it is enabled, we would run slower test by running more iterations.
Should we port this to Go test cases?

tests/gocase/unit/type/list/list_test.go
```diff
@@ -84,3 +84,154 @@ func BenchmarkLTRIM(b *testing.B) {
...
+func TestZipList(t *testing.T) {
+	srv := util.StartServer(t, map[string]string{
+		"list-max-ziplist-size": "16",
+	})
+	defer srv.Close()
+	ctx := context.Background()
+	rdb := srv.NewClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
+
+	rand.Seed(0)
```
PragmaTwice, line 97: Does there need deterministic of PRNG ?

tests/gocase/util/random.go
```diff
@@ -0,0 +1,103 @@
...
+
+func randomSignedInt(max int) int32 {
+	i := rand.Int31n(int32(max))
+	if rand.Float64() > 0.5 {
+		return -i
+	}
+	return i
+}
+
+func RandString(min, max int, typ string) string {
```
PragmaTwice: It would be better if `typ` is enumeration rather than string.

tests/gocase/util/random.go
```diff
@@ -0,0 +1,103 @@
...
+	switch typ {
+	case "binary":
+		minVal, maxVal = 0, 255
+	case "alpha":
+		minVal, maxVal = 48, 122
+	case "compr":
+		minVal, maxVal = 48, 52
+	}
+	for ; length > 0; length-- {
+		s := fmt.Sprintf("%c", minVal+int(rand.Float64()*float64(maxVal-minVal+1)))
```
PragmaTwice: random generation of a range [min, max) can be defined as min + randInt(max - min), where randInt(n) generates random number from 0 to n-1. You can use `rand.Int*n`.

tests/gocase/util/random.go
```diff
@@ -0,0 +1,103 @@
...
+	index := rand.Int31n(int32(len(funcs)))
+	return funcs[index]()
+}
+
+func RandPathNoResult(funcs ...func()) {
+	index := rand.Int31n(int32(len(funcs)))
+	funcs[index]()
+}
+
+func randomSignedInt(max int) int32 {
```
PragmaTwice: rand.Int63n(max * 2) - max

tests/gocase/util/random.go
suica: fixed

tests/gocase/util/random.go
```diff
@@ -0,0 +1,111 @@
...
+	return i
+}
+
+type RandStringType string
+
+const (
+	Alpha  RandStringType = "alpha"
+	Binary RandStringType = "binary"
+	Compr  RandStringType = "compr"
+)
```
PragmaTwice:
```suggestion
type RandStringType int

const (
	Alpha  RandStringType = iota
	Binary
	Compr
)
```

tests/gocase/util/random.go
suica: The original implementation of `randString` in `tests/tcl/tests/support/util.tcl`, the both ends of minVal-maxVal interval (as well as min-max interval which represents the random length range) should be inclusive. Fixed accordingly.

tests/gocase/util/random.go
suica: Fixed according to the original comment...

tests/gocase/unit/type/list/list_test.go
suica: The main reason to have seed set here is to make the testing failure reproducible (although this is to predetermine the test input). 
Otherwise, we need other machanisms to ensure failed tests can be reproduced, such as dumping all history commands and replay them...

PragmaTwice (dismissed):
LGTM

tests/gocase/unit/type/list/list_test.go
tisonkun: Could you link the result of the total time `list_test` take now? If it takes a few seconds, we can simply run 100 iterations and remove this TODO. Otherwise, we run only 10 iterations and remove this TODO. I don't see in a near future we will take over this Redis TCL tests specific flag.

tisonkun: Could you link the result of the total time `list_test` take now? If it takes a few seconds, we can simply run 100 iterations and remove this TODO. Otherwise, we run only 10 iterations and remove this TODO. I don't see in a near future we will take over this Redis TCL tests specific flag.

tests/gocase/unit/type/list/list_test.go
```diff
@@ -84,3 +84,154 @@ func BenchmarkLTRIM(b *testing.B) {
...
+					}
+				}
+				lis = append(lis, data)
+				require.NoError(t, rdb.RPush(ctx, key, data).Err())
+			}
+			require.Equal(t, int64(len(lis)), rdb.LLen(ctx, key).Val())
+
+			for i := int64(199); i >= 0; i-- {
+				require.Equal(t, lis[i], rdb.LIndex(ctx, key, i).Val())
+			}
```
tisonkun:
```suggestion
			for i := 199; i >= 0; i-- {
				require.Equal(t, lis[i], rdb.LIndex(ctx, key, int64(i)).Val())
			}
```

?

tests/gocase/unit/type/list/list_test.go
```diff
@@ -84,3 +84,154 @@ func BenchmarkLTRIM(b *testing.B) {
...
+		}
+	})
+
+	t.Run("ziplist implementation: encoding stress testing", func(t *testing.T) {
+		// TODO: check --accurate options for go test
```
tisonkun:
```suggestion
```

tests/gocase/unit/type/list/list_test.go
```diff
@@ -84,3 +84,154 @@ func BenchmarkLTRIM(b *testing.B) {
...
+		key := "l"
+		for j := 0; j < 200; j++ {
+			require.NoError(t, rdb.Del(ctx, key).Err())
+			lis := []string{}
+			length := int(rand.Int63n(400))
+			for i := 0; i < int(length); i++ {
```
tisonkun:
```suggestion
			var lis []string{}
			l := int(rand.Int63n(400))
			for i := 0; i < l; i++ {
```

tests/gocase/unit/type/list/list_test.go
```diff
@@ -84,3 +84,154 @@ func BenchmarkLTRIM(b *testing.B) {
...
+						require.NoError(t, rdb.RPush(ctx, key, rv).Err())
+					},
+					func() {
+						lis = append([]string{rv}, lis...)
+						require.NoError(t, rdb.LPush(ctx, key, rv).Err())
+					},
+				)
+			}
+			require.Equal(t, int64(len(lis)), rdb.LLen(ctx, key).Val())
+			for i := 0; i < length; i++ {
```
tisonkun:
```suggestion
			for i := 0; i < l; i++ {
```

tests/gocase/unit/type/list/list_test.go
suica: 100 iters cost 17s to run the whole `TestZipList` test case on my laptop.
And 10 iters cost 10s. 
I think it is a resonable difference.

tisonkun: Then from my perspective, either set the iteration to 10 or 100 is OK - we don't run with `--accurate` now. Please choose one and remove this TODO comment.

tests/gocase/util/random.go
```diff
@@ -37,7 +37,7 @@ func RandPathNoResult(funcs ...func()) {
 
 // Random signed integer between -max and max (both extremes excluded).
```
tisonkun:
```suggestion
// Random signed integer in (-max, max)
```

Math has the sign for exclusive interval.

tisonkun (dismissed):
LGTM. One minor comment inline.

tests/gocase/unit/type/list/list_test.go
suica: 100 iters used.

PragmaTwice (approved):

tisonkun (approved):

PR #839, opened 2022-09-08
Fix successor commands won't be processed before receiving new read event

### LINKED ISSUE #831: [BUG] Should read the multiple blocking pushed element
The following test can pass if set `KVROCKS_BIN_PATH` to `redis-server`.

https://github.com/apache/incubator-kvrocks/blob/dbd968b47a08184b3a0b6492f28f9ba8edf6d11b/tests/gocase/unit/protocol/regression_test.go#L58-L63

We should figure out the cause and to fix it.

git-hulk:
Currently, we will stop processing commands when running into block commands
like BRPOP/BLPOP and there remained commands in the connection, but the connection
continues handling those commands since the read event was triggered.

This closes #831

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -1635,12 +1635,13 @@ class CommandBPop : public Commander {
   static void WriteCB(bufferevent *bev, void *ctx) {
     auto self = reinterpret_cast<CommandBPop *>(ctx);
     auto s = self->TryPopFromList();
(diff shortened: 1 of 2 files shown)


caipengbo (approved):

tisonkun (approved):

git-hulk:
Thanks all, merging...

PR #846, opened 2022-09-10
Move tcl keyspace test cases to Go

### LINKED ISSUE #841: Move TCL test unit/keyspace to Go case

git-hulk:
This closes #841 

I think it's a bit wordy to add the directory for only one file, so I used the `keyspace_test.go` instead of `keyspace/keyspace_test.go`


tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+package unit
+
+import (
+	"context"
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
```
tisonkun: I suggest always using `require` to avoid unnecessary dirty states. We just fail fast.

tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,208 @@
...
+	defer srv.Close()
+
+	ctx := context.Background()
+	rdb := srv.NewClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
+
+	t.Run("DEL against a single item", func(t *testing.T) {
+		key := "x"
+		value := "foo"
+		assert.Nil(t, rdb.Set(ctx, key, value, 0).Err())
```
tisonkun:
```suggestion
		require.NoError(t, rdb.Set(ctx, key, value, 0).Err())
```

tisonkun: ditto others

tisonkun (changes requested):
Two global style comments. Will give a functionality review after they're addressed.

tests/gocase/unit/keyspace_test.go
git-hulk: Fixed, you're right that previous steps failed will affect the successor steps in most test cases, so it's better to use require instead of assert.

tests/gocase/unit/keyspace_test.go
git-hulk: all replace with `NoError`

tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,206 @@
...
+			return keys[i] < keys[j]
+		})
+		require.Equal(t, []string{"foo_a", "foo_b", "foo_c", "key_x", "key_y", "key_z"}, keys)
+	})
+
+	t.Run("DEL all keys", func(t *testing.T) {
+		vals := rdb.Keys(ctx, "*").Val()
+		require.Equal(t, int64(len(vals)), rdb.Del(ctx, vals...).Val())
+		require.NoError(t, rdb.Do(ctx, "dbsize", "scan").Err())
+		time.Sleep(100 * time.Millisecond)
```
tisonkun, line 86: Does `dbsize scan` asynchorously? Why we need to sleep here?

tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,206 @@
...
+	})
+
+	t.Run("KEYS to get all keys", func(t *testing.T) {
+		keys := rdb.Keys(ctx, "*").Val()
+		sort.Slice(keys, func(i, j int) bool {
+			return keys[i] < keys[j]
+		})
+		require.Equal(t, []string{"foo_a", "foo_b", "foo_c", "key_x", "key_y", "key_z"}, keys)
+	})
+
```
tisonkun, line 75: Missing

```tcl
    test {DBSIZE} {
        r dbsize scan
        after 100
        r dbsize
    } {6}
```

tisonkun, line 75: Missing

```tcl
    test {DBSIZE} {
        r dbsize scan
        after 100
        r dbsize
    } {6}
```

tisonkun, line 75: Missing

```tcl
    test {DBSIZE} {
        r dbsize scan
        after 100
        r dbsize
    } {6}
```

tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,206 @@
...
+		time.Sleep(100 * time.Millisecond)
+		require.Equal(t, int64(0), rdb.Do(ctx, "dbsize").Val())
+	})
+
+	t.Run("EXISTS", func(t *testing.T) {
+		newKey := "newkey"
+		require.NoError(t, rdb.Set(ctx, newKey, "test", 0).Err())
+		require.Equal(t, int64(1), rdb.Exists(ctx, newKey).Val())
+		require.Equal(t, int64(1), rdb.Del(ctx, newKey).Val())
+		require.Equal(t, int64(0), rdb.Exists(ctx, newKey).Val())
```
tisonkun: If you use `require.EqualValues`, you don't need to cast between integers. But this is no a requirement. Just for your information and current style is OK for me, too.

tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,206 @@
...
+		r, err = c.ReadLine()
+		require.NoError(t, err)
+		require.Equal(t, "xyzk", r)
+		r, err = c.ReadLine()
+		require.NoError(t, err)
+		require.Equal(t, "+PONG", r)
+	})
+
+	t.Run("Non existing command", func(t *testing.T) {
+		require.Contains(t, rdb.Do(ctx, "foobaredcommand").Err().Error(), "ERR")
```
tisonkun:
```suggestion
		util.ErrorRegexp(t, rdb.Do(ctx, "foobaredcommand").Err(), "ERR.*")
```

tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,206 @@
...
+			randomKey := rdb.RandomKey(ctx).Val()
+			if randomKey == "foo" {
+				fooSeen = 1
+			} else if randomKey == "bar" {
+				barSeen = 1
+			}
+			if fooSeen == 1 || barSeen == 1 {
+				break
+			}
+			time.Sleep(10 * time.Millisecond)
```
tisonkun: Why do we need to sleep here?

tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,206 @@
...
+		}
+		require.Greater(t, fooSeen+barSeen, 0)
```
tisonkun: If you'd like to write as:

```go
			if fooSeen == 1 || barSeen == 1 {
				break
			}
```

I suggest:

```go
		for i := 0; i < 1000; i++ {
			randomKey := rdb.RandomKey(ctx).Val()
			switch randomKey {
			case "foo", "bar":
				return
			}
		}
		require.Fail(t, "RANDOMKEY never hits foo or bar")
```

tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,206 @@
...
+	})
+
+	t.Run("RANDOMKEY regression 1", func(t *testing.T) {
+		rdb.FlushDB(ctx)
+		require.NoError(t, rdb.Set(ctx, "x", 10, 0).Err())
+		require.Equal(t, int64(1), rdb.Del(ctx, "x").Val())
+		require.Equal(t, "", rdb.RandomKey(ctx).Val())
+	})
+
+	t.Run("KEYS * two times with long key, Github issue #1208", func(t *testing.T) {
```
tisonkun: This is for the Redis issue number?
```suggestion
	t.Run("KEYS * two times with long key - Redis Github issue #1208", func(t *testing.T) {
```

tests/gocase/unit/keyspace_test.go
```diff
@@ -0,0 +1,206 @@
...
+		rdb.FlushDB(ctx)
+		require.NoError(t, rdb.Set(ctx, "dlskeriewrioeuwqoirueioqwrueoqwrueqw", "teset", 0).Err())
+		require.Equal(t, []string{"dlskeriewrioeuwqoirueioqwrueoqwrueqw"}, rdb.Keys(ctx, "*").Val())
```
tisonkun, line 159:
```suggestion
		require.Equal(t, []string{"dlskeriewrioeuwqoirueioqwrueoqwrueqw"}, rdb.Keys(ctx, "*").Val())
		require.Equal(t, []string{"dlskeriewrioeuwqoirueioqwrueoqwrueqw"}, rdb.Keys(ctx, "*").Val())
```

I think the things to be tested here is "KEYS * two times"

tisonkun (changes requested):
Thanks for your update. Comments inline.

tests/gocase/unit/keyspace_test.go
git-hulk: Yes, the DB scan is asynchronous, so need to wait a moment. Current behavior is consistent with tcl test cases.

tests/gocase/unit/keyspace_test.go
git-hulk: Good catch

tests/gocase/unit/keyspace_test.go
git-hulk: Yes, I saw this but I think this line should be duplicate.

tisonkun: No. The case is that Redis once encounter an issue that if run `KEY *` twice, it will fail. I tend to keep these two lines or we just rename/remove the test case.

See also https://github.com/redis/redis/issues/1208.

tests/gocase/unit/keyspace_test.go
git-hulk: This's a good suggestion, have done.

tests/gocase/unit/keyspace_test.go
git-hulk: Got it, thanks. Just keep those two lines.

tests/gocase/unit/keyspace_test.go
git-hulk: We random by time seed, so it may always get the same seed if requests fast enough. But the case looks good until now, so let's remove the sleep.

git-hulk: We random by time seed, so it may always get the same seed if requests fast enough. But the case looks good until now, so let's remove the sleep.

tisonkun (approved):
Looks good to me. Thank you!

git-hulk:
Thanks for @tisonkun carefully review, will merge after the CI become green.

PR #850, opened 2022-09-11
kvrocks docker can use environment variables define config

usrtax:
fix https://github.com/apache/incubator-kvrocks/discussions/848


tisonkun (changes requested):
This is an important user-facing API change. I suppose you write a brief design for it.

Basically, `kvrocks` binary supports only `-c` option to specify the location of config file for now. If we provide a different interface for the docker image, it's inconsistent and cause confusions.

Besides, `sed` string substitution is handy. That said, you manually edit options in the default config. It's brittle; e.g., what to add an option if  the default config file doesn't have?

These're reasons we need a design first, instead of making some monkey patches.

PragmaTwice:
I think, if we want to support command line options for configurations, we need to implement it in a more direct way, rather than `sed`, which is more like a workaround.

usrtax:
> I think, if we want to support command line options for configurations, we need to implement it in a more direct way, rather than `sed`, which is more like a workaround.

I agree with this, but I'm not familiar with C code and don't know how to modify it.
Generally speaking, if you not using docker, there is no need to configure in this way.
I created my own version of docker [usertax/kvrocks](https://hub.docker.com/repository/docker/usertax/kvrocks) for easy to use now

I added some text in readme for this pull request

When use docker, some configure, such as : port, masterauth password, log directory, can be set by environment variables ( see [docker/run.sh](./docker/run.sh) for details ) .

For example , in your `docker-compose.yml` , you can write :

```yaml
version: '3'
services:
  kvrocks:
    image: kvrocks/kvrocks
    restart: always
    environment:
      kvrocks_masterauth: ${KVROCKS_PASSWORD}
      kvrocks_port: ${KVROCKS_PORT}
      kvrocks_log_dir: /log
    volumes:
      - ./data/kvrocks:/var/lib/kvrocks
      - /var/log/docker/api.user.tax/kvrocks:/log
    ports:
      - ${KVROCKS_PORT}:${KVROCKS_PORT}
```

with a `.env` file like :

```
KVROCKS_PASSWORD=xxx
KVROCKS_PORT=9998
```

tanruixiang:
Hi. I find Macos does't support realpath. I have submitted a patch to you. Here's the [patch](https://github.com/user-tax-dev/incubator-kvrocks/pull/1). And I feel that two `run`(`run.sh` and `_run.sh`) are a little strange. Can these two `run` be combined into one?@usrtax

tisonkun:
@tanruixiang I suggest we focus on the design decision before jump into implementation details.

@PragmaTwice @git-hulk @ShooterIT @usrtax a quick idea on passing override from environment variable is taking a namespace said `kvrocks.` and when start the server, read from the env vars and overwrite `xxx` with `kvrocks.xxx`.

Anyway, I hope this initiative goes through a design instead of rushing into patches.

tisonkun:
Closed in favor of a discussion on the Discussions forum or dev@kvrocks.apache.org mailing list first. Make it clear that we don't rush into patches.

PR #866, opened 2022-09-12
Move TCL test unit/type/list to Go case

### LINKED ISSUE #816: Move TCL test unit/type/list to Go case

suica:
refer to #816

### DIFF
--- a/tests/gocase/unit/type/list/list_test.go
+++ b/tests/gocase/unit/type/list/list_test.go
@@ -50,7 +50,34 @@ func TestList(t *testing.T) {
 
 	t.Run("LPUSH, RPUSH, LLENGTH, LINDEX, LPOP - ziplist", func(t *testing.T) {
 		myZipList1 := "myziplist1"
+		myZipList2 := "myziplist2"
+
+		// first lpush then rpush
 		require.EqualValues(t, 1, rdb.LPush(ctx, myZipList1, "aa").Val())
+		require.EqualValues(t, 2, rdb.RPush(ctx, myZipList1, "bb").Val())
+		require.EqualValues(t, 3, rdb.RPush(ctx, myZipList1, "cc").Val())
+		require.EqualValues(t, 3, rdb.LLen(ctx, myZipList1).Val())
+		require.EqualValues(t, "aa", rdb.LIndex(ctx, myZipList1, 0).Val())
+		require.EqualValues(t, "bb", rdb.LIndex(ctx, myZipList1, 1).Val())
+		require.EqualValues(t, "cc", rdb.LIndex(ctx, myZipList1, 2).Val())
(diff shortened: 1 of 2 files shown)


tests/gocase/unit/type/list/list_test.go
```diff
@@ -39,6 +39,118 @@ var largeValue = map[string]string{
...
+	}
+
+	for typ, value := range largeValue {
+		t.Run(fmt.Sprintf("BLPOP, BRPOP: single existing list - %s", typ), func(t *testing.T) {
+			c := srv.NewTCPClient()
+			defer func() { require.NoError(t, c.Close()) }()
+			createList("blist", []string{"a", "b", value, "c", "d"})
+			res := rdb.LRange(ctx, "blist", 0, -1).Val()
+			fmt.Println(res)
+			require.NoError(t, c.Write("*3\r\n$5\r\nBLPOP\r\n$5\r\nblist\r\n$1\r\n1\r\n"))
```
suica, line 148: I used raw protocol here.

suica, line 148: @tisonkun Hi, I found out that in `list.tcl`, some cases use `set rd [redis_deferring_client]` as a deferring client.

I wonder how to use a "deferring client" in Go.

---

The study I have done:
- I have tried `NewTCPClient`, and although it has similar APIs (such as `ReadLine`), but it turns out to be that I have to write raw Redis protocol to communicate with the server. Actually it works, but it seems to be a little bit awkward... Any way, I have written a little piece of code at [here](https://github.com/apache/incubator-kvrocks/pull/866/files#diff-6df089ad841b180ccd3a65619ee756b051f26ed20ea3cc984feb985c41c9da02). Please comment on this if I did anything wrong :)
- I also have learned how you dealt with the deferring client in #854 and #810, and it seems to be we still need to write raw protocol in code.

tisonkun:
@ShooterIT @git-hulk I'm unfamiliar how `redis_deferring_client` works. Could you give some background here?

@suica for raw client, in #888 I write a `WriteArgs` method for easily writing normal commands.

git-hulk:
> @ShooterIT @git-hulk I'm unfamiliar how `redis_deferring_client` works. Could you give some background here?
> 
> @suica for raw client, in #888 I write a `WriteArgs` method for easily writing normal commands.

`redis_deferring_client` will NOT read the response after sending commands, so we can send many commands then reading them together. Can see https://github.com/apache/incubator-kvrocks/blob/fca9e6962be5e85ad659545d94b5dafa4aec1325/tests/tcl/tests/support/redis.tcl#L140

tisonkun:
Closed as inactive. Code conflict by the way. @suica you can resubmit the patch when it gets completed.

PR #870, opened 2022-09-13
Replace bench with test

tisonkun:

https://github.com/apache/incubator-kvrocks/pull/858#issuecomment-1244823869

Not sure whether it's a bug cause timeout, it's perhaps benchmark test a higher load. Since we already set iterations, I'd like to run in test now.

### DIFF
--- a/tests/gocase/unit/type/list/list_test.go
+++ b/tests/gocase/unit/type/list/list_test.go
@@ -40,29 +40,29 @@ var largeValue = map[string]string{
 	"linkedList": strings.Repeat("hello", 4),
 }
 
-func BenchmarkLTRIM(b *testing.B) {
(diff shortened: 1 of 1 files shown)


git-hulk (approved):
LGTM, thanks @tisonkun

PragmaTwice (approved):

git-hulk:
https://github.com/apache/incubator-kvrocks/blob/unstable/tests/cppunit/rwlock_test.cc#L170

This unit test sleep time is NOT right, expected 100ms but current is 1ms. I think 1ms is too short to guarantee the else condition thread will be always be scheduled before the if condition. I'll bring this sleep time fix in the next PR.

tisonkun:
@git-hulk thanks! I rerun the CI for now.

Issue #857, opened 2022-09-11
Move TCL test unit/quit to Go case


tisonkun:

IoCing:
I would like to work on it !

tisonkun:
@IoCing assigned. Go ahead!

PR #871, opened 2022-09-13
Move TCL test unit/quit to Go case

IoCing:
I find that go-redis don't support quit command.
so i use tcpclient to send command and read result.


IoCing:
refer to #857

src/lock_manager.h
```diff
@@ -25,6 +25,7 @@
 #include <string>
 #include <functional>
 
+
```
tanruixiang: Hi. Is this an empty line added accidentally? It seems that it has nothing to do with your pr.
```suggestion
```

tanruixiang (changes requested):

src/lock_manager.h
IoCing: sorry，i don't know ..... 
maybe because i run check locally

git-hulk:
Thanks for your contribution!

tests/gocase/unit/quit_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func formatCommand(args []string) string {
+	suffix := "\r\n"
+	str := "*" + fmt.Sprintf("%d", len(args)) + suffix
+	for _, arg := range args {
+		str += "$" + fmt.Sprintf("%d", len(arg)) + suffix
+		str += arg + suffix
+	}
```
git-hulk:
```suggestion
	str := fmt.Sprintf("*%d\r\n", len(args))
	for _, arg := range args {
		str += fmt.Sprintf("$%d\r\n%s\r\n", len(arg), arg)
	}
```

tests/gocase/unit/quit_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func TestQuit(t *testing.T) {
+	srv := util.StartServer(t, map[string]string{})
+	defer srv.Close()
+	rdb := srv.NewTCPClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
+
+	t.Run("QUIT returns OK", func(t *testing.T) {
+		// reconnect
```
PragmaTwice: I think in the original TCL code, `reconnect` is a function, not a comment. Is there any reason for this inconsistency?

tests/gocase/unit/quit_test.go
```diff
@@ -0,0 +1,103 @@
...
+		require.NoError(t, rdb.Close())
+		rdb = srv.NewTCPClient()
+		require.NoError(t, rdb.Write(formatCommand([]string{"quit"})))
+		require.NoError(t, rdb.Write(formatCommand([]string{"ping"})))
```
PragmaTwice: In the original TCL code, I think it is not a raw TCP client here. Is there any reason for this inconsistency?

tests/gocase/unit/quit_test.go
```diff
@@ -0,0 +1,101 @@
...
+		resSet, errSet := rdb.ReadLine()
+		require.Equal(t, "", resSet)
+		require.Error(t, errSet)
+		// reconnect
+		require.NoError(t, rdb.Close())
+		rdb = srv.NewTCPClient()
+		require.NoError(t, rdb.Write(formatCommand([]string{"get", "foo"})))
+		resGet, errGet := rdb.ReadLine()
+		require.Equal(t, "$-1", resGet)
+		require.NoError(t, errGet)
```
PragmaTwice: Same as above.

PragmaTwice (commented):
The rest looks good. Thanks for your contribution!

tests/gocase/unit/quit_test.go
IoCing: In the original TCL code ,  `reconnect` function is close old client and init a new client with old client's host, port and config. I directly implemented this function here .

tests/gocase/unit/quit_test.go
IoCing: because the go-redis do not support quit command in pipelined. I think the same effect can be achieved with tcpclient's `write` and `read`.

tests/gocase/unit/quit_test.go
IoCing: If I don't understand wrong , In the original TCL code , These two test cases are also very similar

tanruixiang:
@PragmaTwice  I guess maybe the go-redis not supports quit ([Redis 3 commands except QUIT, MONITOR, and SYNC.](https://github.com/go-redis/redis)).As @IoCing said on the pr.

IoCing:
@PragmaTwice thanks for your comments

tests/gocase/unit/quit_test.go
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package unit
```
tanruixiang:
```suggestion
package quit
```

tests/gocase/unit/quit_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func TestQuit(t *testing.T) {
+	srv := util.StartServer(t, map[string]string{})
+	defer srv.Close()
+	rdb := srv.NewTCPClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
+
+	t.Run("QUIT returns OK", func(t *testing.T) {
+		// reconnect
+		require.NoError(t, rdb.Close())
```
tanruixiang:
```suggestion
		rdb := srv.NewTCPClient()
		defer func() { require.NoError(t, rdb.Close()) }()
```

tests/gocase/unit/quit_test.go
```diff
@@ -0,0 +1,101 @@
...
+		require.Equal(t, "+OK", resQuit)
+		require.NoError(t, errQuit)
+		resPing, errPing := rdb.ReadLine()
+		require.Equal(t, "", resPing)
+		require.Error(t, errPing)
+	})
+
+	t.Run("Pipelined commands after QUIT must not be executed", func(t *testing.T) {
+		// reconnect
+		require.NoError(t, rdb.Close())
```
tanruixiang:
```suggestion
		rdb := srv.NewTCPClient()
		defer func() { require.NoError(t, rdb.Close()) }()
```

tanruixiang (changes requested):

tests/gocase/unit/quit_test.go
```diff
@@ -0,0 +1,101 @@
...
+		str += fmt.Sprintf("$%d\r\n%s\r\n", len(arg), arg)
+	}
+	return str
+}
+
+func TestQuit(t *testing.T) {
+	srv := util.StartServer(t, map[string]string{})
+	defer srv.Close()
+	rdb := srv.NewTCPClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
```
tanruixiang:
```suggestion
```
I think u can move it into every t.Run.

tanruixiang (changes requested):

tanruixiang:
Hi, the first `rdb := srv.NewTCPClient()` is not used, I think u can move it into every `t.Run` .Thanks for your great contribution.

tests/gocase/unit/quit_test.go
PragmaTwice: Oh I got it. Thanks.

PragmaTwice:
I am wondering whether we could use the pipeline support in go-redis instead of current way. Any ideas?

Refer to https://github.com/go-redis/redis/blob/master/pipeline.go

Just an idea, I am ok with current way : )

tests/gocase/unit/quit_test.go
PragmaTwice: I think, if it is possible, we would be better when we use the wrapped client instead of raw TCP client : )

tests/gocase/unit/quit_test.go
PragmaTwice: Got it. Thanks.

tanruixiang:
> I am wondering whether we could use the pipeline support in go-redis instead of current way. Any ideas?
> 
> Refer to https://github.com/go-redis/redis/blob/master/pipeline.go
> 
> Just an idea, I am ok with current way : )


@PragmaTwice  I'm try this way ,but get `unit/pool.go:348: Conn has unread data`,I guess the pipeline still not supports quit? 
```
		rdb := srv.NewClient().Pipeline()
		result := rdb.Do(ctx, formatCommand([]string{"quit"}))
		rdb.Exec(ctx)
		t.Log(result.Val())
```

PragmaTwice:
> > I am wondering whether we could use the pipeline support in go-redis instead of current way. Any ideas?
> > Refer to https://github.com/go-redis/redis/blob/master/pipeline.go
> > Just an idea, I am ok with current way : )
> 
> @PragmaTwice I'm try this way ,but get `unit/pool.go:348: Conn has unread data`,I guess the pipeline still not supports quit?
> 
> ```
> 		rdb := srv.NewClient().Pipeline()
> 		result := rdb.Do(ctx, formatCommand([]string{"quit"}))
> 		rdb.Exec(ctx)
> 		t.Log(result.Val())
> ```

I think you can just write `rdb.Do(ctx, "quit")` instead of these `formatCommand` stuff.

tanruixiang:
> > I am wondering whether we could use the pipeline support in go-redis instead of current way. Any ideas?
> > Refer to https://github.com/go-redis/redis/blob/master/pipeline.go
> > Just an idea, I am ok with current way : )
> 
> @PragmaTwice I'm try this way ,but get `unit/pool.go:348: Conn has unread data`,I guess the pipeline still not supports quit?
> 
> ```
> 		rdb := srv.NewClient().Pipeline()
> 		result := rdb.Do(ctx, formatCommand([]string{"quit"}))
> 		rdb.Exec(ctx)
> 		t.Log(result.Val())
> ```



> > I am wondering whether we could use the pipeline support in go-redis instead of current way. Any ideas?
> > Refer to https://github.com/go-redis/redis/blob/master/pipeline.go
> > Just an idea, I am ok with current way : )
> 
> @PragmaTwice I'm try this way ,but get `unit/pool.go:348: Conn has unread data`,I guess the pipeline still not supports quit?
> 
> ```
> 		rdb := srv.NewClient().Pipeline()
> 		result := rdb.Do(ctx, formatCommand([]string{"quit"}))
> 		rdb.Exec(ctx)
> 		t.Log(result.Val())
> ```



> > I am wondering whether we could use the pipeline support in go-redis instead of current way. Any ideas?
> > Refer to https://github.com/go-redis/redis/blob/master/pipeline.go
> > Just an idea, I am ok with current way : )
> 
> @PragmaTwice I'm try this way ,but get `unit/pool.go:348: Conn has unread data`,I guess the pipeline still not supports quit?
> 
> ```
> 		rdb := srv.NewClient().Pipeline()
> 		result := rdb.Do(ctx, formatCommand([]string{"quit"}))
> 		rdb.Exec(ctx)
> 		t.Log(result.Val())
> ```



> > > I am wondering whether we could use the pipeline support in go-redis instead of current way. Any ideas?
> > > Refer to https://github.com/go-redis/redis/blob/master/pipeline.go
> > > Just an idea, I am ok with current way : )
> > 
> > 
> > @PragmaTwice I'm try this way ,but get `unit/pool.go:348: Conn has unread data`,I guess the pipeline still not supports quit?
> > ```
> > 		rdb := srv.NewClient().Pipeline()
> > 		result := rdb.Do(ctx, formatCommand([]string{"quit"}))
> > 		rdb.Exec(ctx)
> > 		t.Log(result.Val())
> > ```
> 
> I think you can just write `rdb.Do(ctx, "quit")` instead of these `formatCommand` stuff.



> > > I am wondering whether we could use the pipeline support in go-redis instead of current way. Any ideas?
> > > Refer to https://github.com/go-redis/redis/blob/master/pipeline.go
> > > Just an idea, I am ok with current way : )
> > 
> > 
> > @PragmaTwice I'm try this way ,but get `unit/pool.go:348: Conn has unread data`,I guess the pipeline still not supports quit?
> > ```
> > 		rdb := srv.NewClient().Pipeline()
> > 		result := rdb.Do(ctx, formatCommand([]string{"quit"}))
> > 		rdb.Exec(ctx)
> > 		t.Log(result.Val())
> > ```
> 
> I think you can just write `rdb.Do(ctx, "quit")` instead of these `formatCommand` stuff.

Oh,sorry ,I did a stupid thing by copy-pasting, I'll try it again.

tanruixiang:
@PragmaTwice It will panic because of EOF. We do get results for this case if we ignore panic.
```
pipe := srv.NewClient().Pipeline()
tmp1 := pipe.Do(ctx, "quit")
tmp2 := pipe.Do(ctx, "ping")
_, err := pipe.Exec(ctx)
if err != nil { // eof error
	panic(err)
}
t.Log(tmp1.Val()) //  OK
t.Log(tmp2.Val()) //  nil
```

PragmaTwice:
> @PragmaTwice It will panic because of EOF. We do get results for this case if we ignore panic.
> 
> ```
> pipe := srv.NewClient().Pipeline()
> tmp1 := pipe.Do(ctx, "quit")
> tmp2 := pipe.Do(ctx, "ping")
> _, err := pipe.Exec(ctx)
> if err != nil { // eof error
> 	panic(err)
> }
> t.Log(tmp1.Val()) //  OK
> t.Log(tmp2.Val()) //  nil
> ```

A little confused. I think the `panic` is added by yourself rather than in `pipe.Exec`?

tanruixiang:
> > @PragmaTwice It will panic because of EOF. We do get results for this case if we ignore panic.
> > ```
> > pipe := srv.NewClient().Pipeline()
> > tmp1 := pipe.Do(ctx, "quit")
> > tmp2 := pipe.Do(ctx, "ping")
> > _, err := pipe.Exec(ctx)
> > if err != nil { // eof error
> > 	panic(err)
> > }
> > t.Log(tmp1.Val()) //  OK
> > t.Log(tmp2.Val()) //  nil
> > ```
> 
> A little confused. I think the `panic` is added by yourself rather than in `pipe.Exec`?

[Example code](https://redis.uptrace.dev/guide/go-redis-pipelines.html#pipelines)  shows we should check this err. I do not know if we sholud ignore it.

tests/gocase/unit/quit_test.go
IoCing: > Hi, the first `rdb := srv.NewTCPClient()` is not used, I think u can move it into every `t.Run` .Thanks for your great contribution.

I think that the reconnect is not reflected after these changes.

IoCing:
> > > @PragmaTwice It will panic because of EOF. We do get results for this case if we ignore panic.
> > > ```
> > > pipe := srv.NewClient().Pipeline()
> > > tmp1 := pipe.Do(ctx, "quit")
> > > tmp2 := pipe.Do(ctx, "ping")
> > > _, err := pipe.Exec(ctx)
> > > if err != nil { // eof error
> > > 	panic(err)
> > > }
> > > t.Log(tmp1.Val()) //  OK
> > > t.Log(tmp2.Val()) //  nil
> > > ```
> > 
> > 
> > A little confused. I think the `panic` is added by yourself rather than in `pipe.Exec`?
> 
> [Example code](https://redis.uptrace.dev/guide/go-redis-pipelines.html#pipelines) shows we should check this err. I do not know if we sholud ignore it.

@PragmaTwice @tanruixiang Should I now use `pipeline.do` instead of `tcpclient` reimplementation ?

PragmaTwice:
> @PragmaTwice @tanruixiang Should I now use `pipeline.do` instead of `tcpclient` reimplementation ?

I prefer `pipeline.do`, although I think both of them are OK to me.

tanruixiang:
@IoCing Hi,I just try `QUIT returns OK`, If we igrone the err, It can run as expected. You can try others.

IoCing:
> > @PragmaTwice @tanruixiang Should I now use `pipeline.do` instead of `tcpclient` reimplementation ?
> 
> I prefer `pipeline.do`, although I think both of them are OK to me.

ok  ,  i will try to use pipeline.Do to implement

IoCing:
> @IoCing Hi,I just try `QUIT returns OK`, If we igrone the err, It can run as expected. You can try others.

ok,thanks for your help

git-hulk (dismissed):

PragmaTwice (approved):

PragmaTwice (changes requested):
So sorry, but I need to withdraw my approve. :cry:

tests/gocase/unit/quit_test.go
```diff
@@ -0,0 +1,87 @@
...
+	t.Run("QUIT returns OK", func(t *testing.T) {
+		require.NoError(t, reconnect(srv, &rdb))
+		pipe := rdb.Pipeline()
+		cmd1 := pipe.Do(ctx, "quit")
+		cmd2 := pipe.Do(ctx, "ping")
+		_, err := pipe.Exec(ctx)
+		require.Equal(t, "OK", cmd1.Val())
+		require.Equal(t, nil, cmd2.Val())
+		require.Error(t, err)
+	})
```
PragmaTwice: I think there is no pipelined command in this test?

Here is the corresponding TCL code:
```
    test "QUIT returns OK" {
        reconnect
        assert_equal OK [r quit]
        assert_error * {r ping}
    }
```

IoCing:
```go
cmd1 := rdb.Do(ctx, "quit")
cmd2 := rdb.Ping(ctx)
require.Equal(t, "OK", cmd1.Val())
require.Equal(t, nil, cmd2.Val())
```

because go-redis don't support `quit`.And i find use `rdb.Do` to excute `quit` at here is also not work. 

like above,`ping` should return `nil`, but returns `pong` when actually running.

git-hulk: @IoCing Can see this example: https://github.com/apache/incubator-kvrocks/blob/unstable/tests/gocase/unit/keyspace/keyspace_test.go#L106

PragmaTwice: If you really want to use pipeline here, I think the identical code will be:
```
pipe := rdb.Pipeline()
cmd1 := pipe.Do(ctx, "quit")
_, err := pipe.Exec(ctx)
// process cmd1
pipe = rdb.Pipeline()
cmd2 := pipe.Do(ctx, "ping")
_, err = pipe.Exec(ctx)
```

But in your code:
```
pipe := rdb.Pipeline()
cmd1 := pipe.Do(ctx, "quit")
cmd2 := pipe.Do(ctx, "ping")
_, err := pipe.Exec(ctx)
```

these two command is packed into one pipeline, which is not identical to the TCL code.

I think in the original code, the behavior here is request->reply->request->reply, not (packed request)->reply.

IoCing: ok，i revise it now

IoCing:
```go
require.NoError(t, reconnect(srv, &rdb))
pipe := rdb.Pipeline()
cmd1 := pipe.Do(ctx, "quit")
_, err := pipe.Exec(ctx)
require.Equal(t, "OK", cmd1.Val())
require.NoError(t, err)
// process cmd1
pipe = rdb.Pipeline()
cmd2 := pipe.Do(ctx, "ping")
_, err = pipe.Exec(ctx)
require.Equal(t, nil, cmd2.Val())
require.Error(t, err)
```

This method is not work.

image-20220917223238253

 i think i can use tcpclient like what I first implemented.

```go
trdb := srv.NewTCPClient()
require.NoError(t, trdb.Write("*1\r\n$4\r\nquit\r\n"))
resQuit, errQuit := trdb.ReadLine()
require.Equal(t, "+OK", resQuit)
require.NoError(t, errQuit)
require.NoError(t, trdb.Write("*1\r\n$4\r\nping\r\n"))
resPing, errPing := trdb.ReadLine()
require.Equal(t, "", resPing)
require.Error(t, errPing)
```

PragmaTwice (approved):
LGTM

git-hulk (approved):

tisonkun (approved):

PR #882, opened 2022-09-17
Support DISK Command

### LINKED ISSUE #874: The requirement of two commands for Kvrocks
### Search before asking

### Motivation

 I have a requirement which I hope you dear admins can fix.  When we use Redis, we can easily know the memory usage of the specific key by using command "memory usage key " as well as the keyspace hits percentage of the specific key by using command "info".  **But for Kvrocks, we can't easily get the information by using commands.**  In conclusion, i'll appreciate it if you can solve this problem.  Looking forward to your reply, thank you!

### Solution

_No response_

tanruixiang:
It closes #874


src/redis_disk.cc
```diff
@@ -0,0 +1,166 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
```
PragmaTwice, line 20: Hi @tanruixiang , thanks for you contribution!

I noticed that there are many duplicate code snippets in this file, I think you can think about how to reduce/eliminate them.

tanruixiang: Ok. Thank you for your review.

src/redis_disk.h
```diff
@@ -0,0 +1,50 @@
...
+                                      Slice subkeyright = Slice());
+  rocksdb::Status GetStringSize(const Slice &user_key, uint64_t *key_size);
+  rocksdb::Status GetHashSize(const Slice &user_key, uint64_t *key_size);
+  rocksdb::Status GetSetSize(const Slice &user_key, uint64_t *key_size);
+  rocksdb::Status GetListSize(const Slice &user_key, uint64_t *key_size);
+  rocksdb::Status GetZsetSize(const Slice &user_key, uint64_t *key_size);
+  rocksdb::Status GetBitmapSize(const Slice &user_key, uint64_t *key_size);
+  rocksdb::Status GetSortedintSize(const Slice &user_key, uint64_t *key_size);
+ private:
+    rocksdb::SizeApproximationOptions option;
```
torwig: Across Kvrocks codebase private members of the class are named with the `_` suffix. So you can rename `option` to `option` and omit using `this->option` everywhere.

src/redis_disk.cc
```diff
@@ -0,0 +1,146 @@
...
+rocksdb::Status Disk::GetSortedintSize(const Slice &user_key, uint64_t *key_size) {
+    *key_size = 0;
+    std::string ns_key;
+    AppendNamespacePrefix(user_key, &ns_key);
+    SortedintMetadata metadata(false);
+    rocksdb::Status s = Database::GetMetadata(kRedisSortedint, ns_key, &metadata);
+    if (!s.ok()) return s.IsNotFound() ? rocksdb::Status::OK() : s;
+    std::string prefix_key, next_version_prefix_key, start_buf;
+    PutFixed64(&start_buf, 0);
+    return this->GetApproximateSizes(metadata, ns_key,
```
torwig: You can omit `this` when calling class functions here and in other functions as well.

src/redis_disk.cc
```diff
@@ -0,0 +1,146 @@
...
+#include "rocksdb/status.h"
+#include "redis_zset.h"
+#include "redis_bitmap.h"
+#include "redis_disk.h"
+#include "status.h"
+
+namespace Redis {
+rocksdb::Status Disk::GetApproximateSizes(const Metadata &metadata, const Slice &ns_key,
+                                          rocksdb::ColumnFamilyHandle *column_family,
+                                          uint64_t *key_size, Slice subkeyleft ,
```
torwig: I'd recommend making the `key_size` argument the last one because it is the output one and it is passed by a pointer.

src/redis_disk.cc
```diff
@@ -0,0 +1,146 @@
...
+    PutFixed64(&buf, metadata.head);
+    return this->GetApproximateSizes(metadata, ns_key,
+                                     storage_->GetCFHandle(Engine::kSubkeyColumnFamilyName),
+                                     key_size, buf);
+}
+
+rocksdb::Status Disk::GetZsetSize(const Slice &user_key, uint64_t *key_size) {
+    *key_size = 0;
+    std::string ns_key;
+    AppendNamespacePrefix(user_key, &ns_key);
```
torwig: To be consistent with the rest of Kvrock's code, please use the indentation of 2 spaces instead of 4.

src/redis_disk.h
tanruixiang: > Across Kvrocks codebase private members of the class are named with the `_` suffix. So you can rename `option` to `option` and omit using `this->option` everywhere.

thank you. It has been modified.

src/redis_disk.cc
tanruixiang: > You can omit `this` when calling class functions here and in other functions as well.

thank you. It has been modified.

src/redis_disk.cc
tanruixiang: > 



> I'd recommend making the `key_size` argument the last one because it is the output one and it is passed by a pointer.

Hi. The last two are parameters with default values. In fact, `key_size` are the last parameters without default values.

torwig: Aha. I see that you can't just move that parameter to the end because there are two args with default values in the end.

src/redis_disk.cc
tanruixiang: > To be consistent with the rest of Kvrock's code, please use the indentation of 2 spaces instead of 4.

Thank you. It has been modified.

src/redis_disk.cc
torwig: My bad, my apologies.

tanruixiang: > My bad, my apologies.

Don't apologize. Thank you very much for your help. 😁

src/redis_disk.cc
```diff
@@ -0,0 +1,146 @@
...
+  return db_->GetApproximateSizes(option_, metadata_cf_handle_, &key_range, 1, key_size);
+}
+
+rocksdb::Status Disk::GetHashSize(const Slice &user_key, uint64_t *key_size) {
+  *key_size = 0;
+  std::string ns_key;
+  AppendNamespacePrefix(user_key, &ns_key);
+  HashMetadata metadata(false);
+  rocksdb::Status s = Database::GetMetadata(kRedisHash, ns_key, &metadata);
+  if (!s.ok()) return s.IsNotFound() ? rocksdb::Status::OK() : s;
```
PragmaTwice: These code appears in almost every method in this file, which is like a copy-and-paste programming pattern, so I think you can just define something like
```
rocksdb::Status GetHashSize(auto ns_key, auto metadata, auto key_size) // fill these types
```
and then
```
RedisType type = ...;
std::string ns_key;
AppendNamespacePrefix(user_key, &ns_key);
Metadata metadata(type, false);
rocksdb::Status s = Database::GetMetadata(kRedisHash, ns_key, &metadata);
if (!s.ok()) return s.IsNotFound() ? rocksdb::Status::OK() : s;

switch (redisType) {
  case kRedisHash: GetHashSize(...); break;
  case kRedisSomething: GetSomethingSize(...) 
  ...
}
```

PragmaTwice: Or you can even do not define one method per redis type, I think it is ok to just implement it in the switch clause.

tanruixiang: > Or you can even do not define one method per redis type, I think it is ok to just implement it in the switch clause.

Thank you. I'm doing it now.

tests/cppunit/disk_test.cc
```diff
@@ -0,0 +1,202 @@
...
+    EXPECT_TRUE(string->Set(key_, std::string(p, 'a')).ok());
+    std::string got_value;
+    EXPECT_TRUE(string->Get(key_,  &got_value).ok());
+    EXPECT_EQ(got_value, std::string(p, 'a'));
+    std::string value;
+    string->Get(key_, &value);
+    uint64_t result = 0;
+    EXPECT_TRUE(disk->GetKeySize(key_, kRedisString, &result).ok());
+    EXPECT_GE(result, p);
+    break;
```
torwig: I'm wondering why this operator (`break`) is here. With it, the loop will run only one time. But if you want to do something just one time, there is no reason to introduce the loop.

tests/cppunit/disk_test.cc
```diff
@@ -0,0 +1,202 @@
...
+  }
+  bitmap->Del(key_);
+}
+
+TEST_F(RedisDiskTest, SortedintDisk) {
+  std::unique_ptr<Redis::Sortedint> sortedint = Util::MakeUnique<Redis::Sortedint>(storage_, "disk_ns_sortedint");
+  std::unique_ptr<Redis::Disk> disk = Util::MakeUnique<Redis::Disk>(storage_, "disk_ns_sortedint");
+  key_ = "sortedintdisk_key";
+  int ret;
+  for(int i=0;i<10;i++){
```
torwig: Just formatting (add spaces like on line 174).

tests/cppunit/disk_test.cc
```diff
@@ -0,0 +1,202 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#include <chrono>
+#include <gtest/gtest.h>
```
torwig: It's not a big deal, but I'd recommend grouping headers of the standard library together (`chrono` after third-party `gtest` header)

tests/cppunit/disk_test.cc
tanruixiang: I noticed that the running results of ci are somewhat different from those of my native machine. At the same time, because this is a draft pr, I am using ci for testing.I found that a byte of data is often not detected due to the problem with the settings of Writebatch(sync is default false).

tests/cppunit/disk_test.cc
tanruixiang: > It's not a big deal, but I'd recommend grouping headers of the standard library together (`chrono` after third-party `gtest` header)

Ok. Very good advice.

src/redis_disk.cc
```diff
@@ -0,0 +1,163 @@
...
+  return GetApproximateSizes(metadata, ns_key,
+                             storage_->GetCFHandle(Engine::kSubkeyColumnFamilyName),
+                             key_size, std::to_string(0), std::to_string(0));
+}
+
+rocksdb::Status Disk::GetSortedintSize(const Slice &ns_key, uint64_t *key_size) {
+  SortedintMetadata metadata(false);
+  rocksdb::Status s = Database::GetMetadata(kRedisSortedint, ns_key, &metadata);
+  if (!s.ok()) return s.IsNotFound() ? rocksdb::Status::OK() : s;
+  std::string prefix_key, next_version_prefix_key, start_buf;
```
torwig: `prefix_key` and `next_version_prefix_key` variables are declared but not used.

src/redis_disk.cc
```diff
@@ -0,0 +1,163 @@
...
+      case RedisType::kRedisSortedint:
+        s = GetSortedintSize(ns_key, key_size);
+        break;
+      case RedisType::kRedisZSet:
+        s = GetZsetSize(ns_key, key_size);
+        break;
+      case RedisType::kRedisNone:
+        s = GetHashSize(ns_key, key_size);
+        return rocksdb::Status::NotFound("Not found ", user_key);
+        break;
```
torwig: I think you don't need `break` here because of `return`.

tanruixiang: thank you. It has been fixed.

tests/cppunit/disk_test.cc
tanruixiang: > I'm wondering why this operator (`break`) is here. With it, the loop will run only one time. But if you want to do something just one time, there is no reason to introduce the loop.

I noticed that the running results of ci are somewhat different from those of my native machine. At the same time, because this is a draft pr, I am using ci for testing.I found that a byte of data is often not detected due to the problem with the settings of Writebatch(sync is default false). So don't worry. I'll fix it after the test. Thanks again for your serious review

tanruixiang:
@PragmaTwice Hi. Do you have any more suggestions on reducing redundant code?

src/redis_disk.cc
```diff
@@ -0,0 +1,160 @@
...
+  if (!s.ok()) return s;
+  return GetApproximateSizes(metadata, ns_key,
+                             storage_->GetCFHandle(Engine::kSubkeyColumnFamilyName),
+                             key_size);
+}
+
+rocksdb::Status Disk::GetBitmapSize(const Slice &ns_key, uint64_t *key_size) {
+  BitmapMetadata metadata(false);
+  rocksdb::Status s = Database::GetMetadata(kRedisBitmap, ns_key, &metadata);
+  if (!s.ok()) return s.IsNotFound() ? rocksdb::Status::OK() : s;
```
PragmaTwice, line 134: As in https://github.com/apache/incubator-kvrocks/pull/882#discussion_r973585761,
you can put them out of the `switch`, then it can be eliminated from every method in this file.

tanruixiang:
@PragmaTwice @git-hulk @torwig I noticed a problem. In the following tests. When we write a value with only one bit(like 'a'). Occasionally, the size of the key we get is 0. It occurs frequently when the value is very small. I guess the reason for this problem is that sync in writeoption defaults to false. Or because the `GetApproximateSizes` can only yield approximate values, which may be ignored for small values.
```
  std::vector<int> value_size{1, 1024};
  for(auto &p : value_size){
    EXPECT_TRUE(string->Set(key_, std::string(p, 'a')).ok());
    std::string got_value;
    EXPECT_TRUE(string->Get(key_,  &got_value).ok());
    EXPECT_EQ(got_value, std::string(p, 'a'));
    std::string value;
    string->Get(key_, &value);
    uint64_t result = 0;
    EXPECT_TRUE(disk->GetKeySize(key_, kRedisString, &result).ok());
    EXPECT_GE(result, p);
  }
```

src/redis_disk.cc
tanruixiang: We need to set different types of metadata variables. The code seems to be more unreadable if the Use inheritance mechanism is used?

PragmaTwice: Oh you are right, a factory function is needed for `Metadata`. So weird since the ctor of `Metadata` is public, I think protected is enough. It is a mess while `Metadata` itself is contructed by users.

src/redis_disk.cc
```diff
@@ -0,0 +1,160 @@
...
+  return rocksdb::Status::OK();
+}
+rocksdb::Status Disk::GetKeySize(const Slice &user_key, RedisType type, uint64_t *key_size) {
+  *key_size = 0;
+  std::string ns_key;
+  rocksdb::Status s;
+  AppendNamespacePrefix(user_key, &ns_key);
+    switch (type) {
+      case RedisType::kRedisString:
+        s = GetStringSize(ns_key, key_size);
```
PragmaTwice: You can return the status directly. BTW the switch block is not aligned well.

tanruixiang: Thank you, it has been fixed.

tanruixiang:
> @PragmaTwice @git-hulk @torwig I noticed a problem. In the following tests. When we write a value with only one bit(like 'a'). Occasionally, the size of the key we get is 0. It occurs frequently when the value is very small. I guess the reason for this problem is that sync in writeoption defaults to false. Or because the `GetApproximateSizes` can only yield approximate values, which may be ignored for small values.
> 
> ```
>   std::vector<int> value_size{1, 1024};
>   for(auto &p : value_size){
>     EXPECT_TRUE(string->Set(key_, std::string(p, 'a')).ok());
>     std::string got_value;
>     EXPECT_TRUE(string->Get(key_,  &got_value).ok());
>     EXPECT_EQ(got_value, std::string(p, 'a'));
>     std::string value;
>     string->Get(key_, &value);
>     uint64_t result = 0;
>     EXPECT_TRUE(disk->GetKeySize(key_, kRedisString, &result).ok());
>     EXPECT_GE(result, p);
>   }
> ```

@git-hulk  Hi. Do you have time to look at this question? Because the approximate usage space is obtained, it may encounter a situation of 0 during testing. Increasing the sleep time also doesn't completely solve the problem. It seems that the sync method is not supported either.

git-hulk:
> > @PragmaTwice @git-hulk @torwig I noticed a problem. In the following tests. When we write a value with only one bit(like 'a'). Occasionally, the size of the key we get is 0. It occurs frequently when the value is very small. I guess the reason for this problem is that sync in writeoption defaults to false. Or because the `GetApproximateSizes` can only yield approximate values, which may be ignored for small values.
> > ```
> >   std::vector<int> value_size{1, 1024};
> >   for(auto &p : value_size){
> >     EXPECT_TRUE(string->Set(key_, std::string(p, 'a')).ok());
> >     std::string got_value;
> >     EXPECT_TRUE(string->Get(key_,  &got_value).ok());
> >     EXPECT_EQ(got_value, std::string(p, 'a'));
> >     std::string value;
> >     string->Get(key_, &value);
> >     uint64_t result = 0;
> >     EXPECT_TRUE(disk->GetKeySize(key_, kRedisString, &result).ok());
> >     EXPECT_GE(result, p);
> >   }
> > ```
> 
> @git-hulk Hi. Do you have time to look at this question? Because the approximate usage space is obtained, it may encounter a situation of 0 during testing. Increasing the sleep time also doesn't completely solve the problem. It seems that the sync method is not supported either.

Did you enable the `include_memtabtles` in SizeApproximationOptions?

tanruixiang:
> > > @PragmaTwice @git-hulk @torwig I noticed a problem. In the following tests. When we write a value with only one bit(like 'a'). Occasionally, the size of the key we get is 0. It occurs frequently when the value is very small. I guess the reason for this problem is that sync in writeoption defaults to false. Or because the `GetApproximateSizes` can only yield approximate values, which may be ignored for small values.
> > > ```
> > >   std::vector<int> value_size{1, 1024};
> > >   for(auto &p : value_size){
> > >     EXPECT_TRUE(string->Set(key_, std::string(p, 'a')).ok());
> > >     std::string got_value;
> > >     EXPECT_TRUE(string->Get(key_,  &got_value).ok());
> > >     EXPECT_EQ(got_value, std::string(p, 'a'));
> > >     std::string value;
> > >     string->Get(key_, &value);
> > >     uint64_t result = 0;
> > >     EXPECT_TRUE(disk->GetKeySize(key_, kRedisString, &result).ok());
> > >     EXPECT_GE(result, p);
> > >   }
> > > ```
> > 
> > 
> > @git-hulk Hi. Do you have time to look at this question? Because the approximate usage space is obtained, it may encounter a situation of 0 during testing. Increasing the sleep time also doesn't completely solve the problem. It seems that the sync method is not supported either.
> 
> Did you enable the `include_memtabtles` in SizeApproximationOptions?

Of course yes.
```
    this->option_.include_memtabtles = true;
    this->option_.include_files = true;
```

tanruixiang:
Is it possible that it is a bug of rocksdb? I can get the value of the corresponding key, but the size occupied by the key is 0. And this problem occurs occasionally.

git-hulk:
> > > > @PragmaTwice @git-hulk @torwig I noticed a problem. In the following tests. When we write a value with only one bit(like 'a'). Occasionally, the size of the key we get is 0. It occurs frequently when the value is very small. I guess the reason for this problem is that sync in writeoption defaults to false. Or because the `GetApproximateSizes` can only yield approximate values, which may be ignored for small values.
> > > > ```
> > > >   std::vector<int> value_size{1, 1024};
> > > >   for(auto &p : value_size){
> > > >     EXPECT_TRUE(string->Set(key_, std::string(p, 'a')).ok());
> > > >     std::string got_value;
> > > >     EXPECT_TRUE(string->Get(key_,  &got_value).ok());
> > > >     EXPECT_EQ(got_value, std::string(p, 'a'));
> > > >     std::string value;
> > > >     string->Get(key_, &value);
> > > >     uint64_t result = 0;
> > > >     EXPECT_TRUE(disk->GetKeySize(key_, kRedisString, &result).ok());
> > > >     EXPECT_GE(result, p);
> > > >   }
> > > > ```
> > > 
> > > 
> > > @git-hulk Hi. Do you have time to look at this question? Because the approximate usage space is obtained, it may encounter a situation of 0 during testing. Increasing the sleep time also doesn't completely solve the problem. It seems that the sync method is not supported either.
> > 
> > 
> > Did you enable the `include_memtabtles` in SizeApproximationOptions?
> 
> Of course yes.
> 
> ```
>     this->option_.include_memtabtles = true;
>     this->option_.include_files = true;
> ```

OK, I need to take some time to inspect the reason. But I think it's ok to skip this corner case first since we should NOT care if the usage is small enough.

tanruixiang:
> > > > > @PragmaTwice @git-hulk @torwig I noticed a problem. In the following tests. When we write a value with only one bit(like 'a'). Occasionally, the size of the key we get is 0. It occurs frequently when the value is very small. I guess the reason for this problem is that sync in writeoption defaults to false. Or because the `GetApproximateSizes` can only yield approximate values, which may be ignored for small values.
> > > > > ```
> > > > >   std::vector<int> value_size{1, 1024};
> > > > >   for(auto &p : value_size){
> > > > >     EXPECT_TRUE(string->Set(key_, std::string(p, 'a')).ok());
> > > > >     std::string got_value;
> > > > >     EXPECT_TRUE(string->Get(key_,  &got_value).ok());
> > > > >     EXPECT_EQ(got_value, std::string(p, 'a'));
> > > > >     std::string value;
> > > > >     string->Get(key_, &value);
> > > > >     uint64_t result = 0;
> > > > >     EXPECT_TRUE(disk->GetKeySize(key_, kRedisString, &result).ok());
> > > > >     EXPECT_GE(result, p);
> > > > >   }
> > > > > ```
> > > > 
> > > > 
> > > > @git-hulk Hi. Do you have time to look at this question? Because the approximate usage space is obtained, it may encounter a situation of 0 during testing. Increasing the sleep time also doesn't completely solve the problem. It seems that the sync method is not supported either.
> > > 
> > > 
> > > Did you enable the `include_memtabtles` in SizeApproximationOptions?
> > 
> > 
> > Of course yes.
> > ```
> >     this->option_.include_memtabtles = true;
> >     this->option_.include_files = true;
> > ```
> 
> OK, I need to take some time to inspect the reason. But I think it's ok to skip this corner case first since we should NOT care if the usage is small enough.

It's not only small cases where problems can arise. Sometimes a larger value also has a situation where the result is 0. For example, the failure of this CI has a relatively large value but get key size is zero. In this case ,we Lpush 1000 number value but get 0 size.[big value but get size 0 case](https://github.com/apache/incubator-kvrocks/actions/runs/3077380954/jobs/4972217783)

git-hulk:
> It's not only small cases where problems can arise. Sometimes a larger value also has a situation where the result is 0. For example, the failure of this CI has a relatively large value but get key size is zero. In this case ,we Lpush 1000 number value but get 0 size.[big value but get size 0 case](https://github.com/apache/incubator-kvrocks/actions/runs/3077380954/jobs/4972217783)

OK, got it. I need to read this PR first and try to reproduce it on my side.

src/redis_disk.h
```diff
@@ -0,0 +1,51 @@
...
+ public:
+  explicit Disk(Engine::Storage *storage, const std::string &ns): Database(storage, ns) {
+    this->option_.include_memtabtles = true;
+    this->option_.include_files = true;
+  }
+  rocksdb::Status GetApproximateSizes(const Metadata &metadata, const Slice &ns_key,
+                                      rocksdb::ColumnFamilyHandle *column_family,
+                                      uint64_t *key_size, Slice subkeyleft = Slice(),
+                                      Slice subkeyright = Slice());
+  rocksdb::Status GetStringSize(const Slice &user_key, uint64_t *key_size);
```
caipengbo: Should we put these functions in each type of files (Like `redis_string.h` and `redis_string.cpp`)?

torwig: @caipengbo Do you mean that each type should implement its own function (e.g. `GetApproximateKeySize`) ? Otherwise, I think, it's not the best idea to spread functions of `Disk` class among different files (one function per file).

caipengbo: I think `GetApproximateKeySize` should be a virtual function and every data type should rewrite it

tanruixiang: > I think `GetApproximateKeySize` should be a virtual function and every data type should rewrite it

Using virtual functions is also a good approach. May I ask  using virtual functions still require switch to create different objects? If so,that looks like it's about the same as my current implementation. Maybe my c++ is not proficient enough, do you have a better solution? Thank you for your review.

caipengbo: Indeed, we need to create different objects. We added an additional command to calculate the size of different types, and your implementation is also a good method.

src/redis_disk.cc
```diff
@@ -0,0 +1,151 @@
...
+    case RedisType::kRedisSet:
+      return GetSetSize(ns_key, key_size);
+    case RedisType::kRedisSortedint:
+      return GetSortedintSize(ns_key, key_size);
+    case RedisType::kRedisZSet:
+      return GetZsetSize(ns_key, key_size);
+    case RedisType::kRedisNone:
+      return rocksdb::Status::NotFound("Not found ", user_key);
+    case RedisType::kRedisStream:
+      return rocksdb::Status::NotSupported("Not support stream");
```
torwig: Why Stream type is not supported? I think getting its size is very similar to `set`, except Stream has its own column family. By the way, you can add a new line before and after `GetKeySize` and delete one line between `GetHashSize` and `GetSetSize`.

tanruixiang: @torwig  I'm sorry for missing the `stream`. I added the corresponding type based on [design document](https://kvrocks.apache.org/docs/Design/design-structure-on-rocksdb). There is no documentation for `stream` here. I noticed that `stream` are implemented by you, do you have time to add `stream`  into [this website](https://kvrocks.apache.org/docs/Design/design-structure-on-rocksdb)?

torwig: Yes, I can add the design of the Stream data type to that document. I just thought that that document can be updated with the changes that are released. And right now the Stream type isn't released. @git-hulk @PragmaTwice Am I right?

git-hulk: Sure, I had updated the supported command in the Kvrocks website: https://github.com/apache/incubator-kvrocks-website/blob/main/docs/01-supported-commands.md

torwig: I see, then I will update the design docs.

git-hulk: Thanks @torwig

tanruixiang:
> OK, got it. I need to read this PR first and try to reproduce it on my side.

@git-hulk  I used the parameters configured for `GetApproximateSizes` testing in `rocksdb`. However, there is still a probability of zero results. Do you have any suggestions?

src/redis_disk.cc
torwig: @tanruixiang I've added Stream design here https://github.com/apache/incubator-kvrocks-website/pull/17 .

tanruixiang: > @tanruixiang I've added Stream design here [apache/incubator-kvrocks-website#17](https://github.com/apache/incubator-kvrocks-website/pull/17) .

Thank you very much! Thanks for your contribution!

tanruixiang: I'm still looking for the reason why sometimes the value obtained is 0. If found, I will add  `GetStreamSize`.

git-hulk:
> > OK, got it. I need to read this PR first and try to reproduce it on my side.
> 
> @git-hulk I used the parameters configured for `GetApproximateSizes` testing in `rocksdb`. However, there is still a probability of zero results. Do you have any suggestions?

Sorry for missing this comment, I have no enough time to inspect this case yet.

tanruixiang:
> Sorry for missing this comment, I have no enough time to inspect this case yet.

@git-hulk  Do not worry. I've been looking at this question. I found a strange phenomenon, if you get the corresponding key directly, you can find the value, but calling GetApproximateSizes will get 0. Maybe I'm almost sure what the reason is, please give me a little more time.(In general it can indeed be a problem with `rocksdb`.)

tanruixiang:
@git-hulk  I think I already know why. According to [this](https://github.com/facebook/rocksdb/issues/10733), to prevent the situation of 0 size during the test, it is necessary to save the data in different data blocks of sst as much as possible. This requires us to enter a lot of data when testing.

git-hulk:
> @git-hulk I think I already know why. According to [this](https://github.com/facebook/rocksdb/issues/10733), to prevent the situation of 0 size during the test, it is necessary to save the data in different data blocks of sst as much as possible. This requires us to enter a lot of data when testing.

Nice!

src/redis_disk.h
```diff
@@ -0,0 +1,52 @@
...
+  rocksdb::Status GetHashSize(const Slice &ns_key, uint64_t *key_size);
+  rocksdb::Status GetSetSize(const Slice &ns_key, uint64_t *key_size);
+  rocksdb::Status GetListSize(const Slice &ns_key, uint64_t *key_size);
+  rocksdb::Status GetZsetSize(const Slice &ns_key, uint64_t *key_size);
+  rocksdb::Status GetBitmapSize(const Slice &ns_key, uint64_t *key_size);
+  rocksdb::Status GetSortedintSize(const Slice &ns_key, uint64_t *key_size);
+  rocksdb::Status GetStreamSize(const Slice &ns_key, uint64_t *key_size);
+  rocksdb::Status GetKeySize(const Slice &user_key, RedisType type, uint64_t *key_size);
+ private:
+    rocksdb::SizeApproximationOptions option_;
```
torwig: @tanruixiang Hi, I'll try to review it later today. 
After a brisk view:
- fix indentation here
- keep one blank line between method definitions in the .cc file
- using `this` on line 32, 33 is redundant

tests/cppunit/disk_test.cc
```diff
@@ -0,0 +1,202 @@
...
+#include "redis_list.h"
+#include "redis_zset.h"
+#include "redis_bitmap.h"
+#include "redis_sortedint.h"
+#include "redis_stream.h"
+
+class RedisDiskTest : public TestBase {
+protected:
+  explicit RedisDiskTest() : TestBase() {
+    delete storage_;
```
torwig, line 40: Have you tried to use this code without deleting and creating new storage? `config_` is passed through the pointer and it is used mainly within the `Open()` call.

src/redis_disk.h
tanruixiang: Thanks a lot, it has been fixed.

tests/cppunit/disk_test.cc
tanruixiang: > Have you tried to use this code without deleting and creating new storage? `config_` is passed through the pointer and it is used mainly within the `Open()` call.

It seems that there is no interface to directly modify the config. Will it cause a memory leak if I don't delete it?

torwig: I was thinking about removing the line `storage_ = new Engine::Storage(config_);` Does it work?

torwig: And of course without deleting existing config instance.

torwig: `config_` is a protected member so you can modify its members from the derived class (if I'm not mistaken)

tanruixiang: Let me try it.

tanruixiang: Is this your idea? I didn't delve into it, but it crashes.
```
  explicit RedisDiskTest() : TestBase() {
    // delete storage_;
    storage_->debugconfig();
    config_->RocksDB.compression = rocksdb::CompressionType::kNoCompression;
    config_->RocksDB.write_buffer_size = 1;
    // storage_ = new Engine::Storage(config_);
    Status s = storage_->Open();
    if (!s.IsOK()) {
      std::cout << "Failed to open the storage, encounter error: " << s.Msg() << std::endl;
      assert(s.IsOK());
    }
    storage_->debugconfig();
  }
```

torwig: Yes, you are right, that was my idea. If it crashes, discard my idea :)

torwig: Now I see. The `TestBase` constructor calls `storage_->Open()`

tanruixiang: > 

Yes. My guess is that there are some resources that must be destructed to be freed, so it can't just be reopened.

tanruixiang: > Now I see. The TestBase constructor calls storage_->Open()

Yes. My guess is that there are some resources that must be destructed to be freed, so it can't just be reopened.

torwig:
@tanruixiang Hi, I've done some tests locally. For example, if I set `x` to a string of length 50 and then use `disk` command - I get something about 80 as a result. Then I restart `kvrocks` and repeat `disk` command. And this time the result is something about 1500. Is it some kind of approximation by the `rocksdb`?

tests/gocase/unit/disk/disk_test.go
```diff
@@ -0,0 +1,111 @@
...
+	defer srv.Close()
+	ctx := context.Background()
+	rdb := srv.NewClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
+	estimationFactor := 0.5
+	t.Run("Disk usage String", func(t *testing.T) {
+		require.NoError(t, rdb.Set(ctx, "stringkey", "aaaaaaaaaaaaaaaaaaaaaaaa", 0).Err())
+		val, err := rdb.Do(ctx, "Disk", "usage", "stringkey").Int()
+		require.NoError(t, err)
+		require.GreaterOrEqual(t, val, 1)
```
PragmaTwice: IMHO, the expected value is weird. And I wonder if it can be a range rather than only a lower bound.

tanruixiang:
> @tanruixiang Hi, I've done some tests locally. For example, if I set `x` to a string of length 50 and then use `disk` command - I get something about 80 as a result. Then I restart `kvrocks` and repeat `disk` command. And this time the result is something about 1500. Is it some kind of approximation by the `rocksdb`?

Hi. I think it's as expected. It should be the reason for the statistical granularity of rocksdb. You can see in [there](https://github.com/facebook/rocksdb/issues/10733).

tests/gocase/unit/disk/disk_test.go
tanruixiang: > IMHO, the expected value is weird. And I wonder if it can be a range rather than only a lower bound.

sounds good. Let me try.

tanruixiang: @PragmaTwice Now my judgment is that the result will not exceed the upper bound or the lower bound within an order of magnitude

tests/gocase/unit/disk/disk_test.go
```diff
@@ -0,0 +1,145 @@
...
+		approximateSize := 0
+		for i := 0; i < 10000; i++ {
+			require.NoError(t, rdb.HSet(ctx, "hashkey", "key"+strconv.Itoa(i), "value"+strconv.Itoa(i)).Err())
+			approximateSize += len("hashkey") + len("key"+strconv.Itoa(i)) + len("value"+strconv.Itoa(i)) + 8
+		}
+		require.NoError(t, rdb.Set(ctx, "hashkey", strings.Repeat("A", 1024), 0).Err())
+		val, err := rdb.Do(ctx, "Disk", "usage", "hashkey").Int()
+		require.NoError(t, err)
+		require.GreaterOrEqual(t, val, int(float64(approximateSize)*estimationFactor))
+		require.LessOrEqual(t, val, int(float64(approximateSize)/estimationFactor))
```
PragmaTwice, line 62: I am wondering if this could be narrowed down a bit more. This range currently has a hundred times the gap between the upper and lower bound, which I think is still relatively large. But if we can confirm that the estimation accuracy of rocksdb can only be this at most, it is fine.

tanruixiang: I think it is theoretically possible, but we need to accurately calculate the key size constructed by kvrocks as the expected size. Also consider rocksdb's space compression for keys (such as prefix compression). In fact, the [test](https://github.com/facebook/rocksdb/blob/main/db/db_test.cc#L1465) of rocksdb is also a relatively large range. I personally don't think it makes much sense to do these jobs, because as long as there is no problem with the rocksdb we depend on, then there will be no problem with this function, If there is a change in the kvrocks level, the scope now is enough to test the problem. Of course, I can give it a try if you feel the need to.

tanruixiang:
@git-hulk @torwig @caipengbo Hi. Do you have time to review?

tests/gocase/unit/disk/disk_test.go
PragmaTwice: I think it is fine if rocksdb also has a large range.

git-hulk:
> @git-hulk @torwig @caipengbo Hi. Do you have time to review?

Sure, I have read through the implementation, and overall is good to me. I'm still wondering if we have a more general way to reduce duplicate codes, I will approve this PR soon if can't figure out a better way.

src/redis_cmd.cc
```diff
@@ -3476,6 +3477,32 @@ class CommandInfo : public Commander {
...
 };
 
+class CommandDisk : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    std::string opname = Util::ToLower(args[1]);
+    if (opname != "usage")
+      return Status(Status::RedisInvalidCmd, "Unknown operation");
+    if (args.size() != 3)
+      return Status(Status::RedisInvalidCmd, "Incorrect number of parameters");
```
git-hulk: Can remove this since it's done before executing the command
```suggestion
```

tanruixiang:
> Sure, I have read through the implementation, and overall is good to me. I'm still wondering if we have a more general way to reduce duplicate codes, I will approve this PR soon if can't figure out a better way.

Do you have any ideas? It seems like it should be impossible to reduce each type of custom operation.

git-hulk:
> > Sure, I have read through the implementation, and overall is good to me. I'm still wondering if we have a more general way to reduce duplicate codes, I will approve this PR soon if can't figure out a better way.
> 
> Do you have any ideas? It seems like it should be impossible to reduce each type of custom operation.

The main issue is `GetMetadata` requires passing the type, so another way is to allow skipping type check in `GetMetadata`, but the current implementation is also fine to me. To see others have thoughts on this cc @ShooterIT @PragmaTwice @caipengbo @torwig

git-hulk (approved):
LGTM

git-hulk:
@PragmaTwice @caipengbo @torwig Do you have any suggestion on this PR? Or I will merge it if we have no further comment.

caipengbo:
LGTM

caipengbo (approved):

git-hulk:
Thanks for @tanruixiang contribution and patient, also thanks to everyone who reviewed this PR. Merging...

caipengbo:
We also need to update our website :) @git-hulk

git-hulk:
Yes, @caipengbo you can help to update as well when you're free. https://github.com/apache/incubator-kvrocks-website/blob/main/docs/01-supported-commands.md

caipengbo:
> Yes, @caipengbo you can help to update as well when you're free. https://github.com/apache/incubator-kvrocks-website/blob/main/docs/01-supported-commands.md

If I update the repository, will the site automatically update?

tanruixiang:
> We also need to update our website :) @git-hulk

Do you want to do it? If you don't have time, I can update the website later today.

git-hulk:
Yes, GitHub actions will deploy the update automatically.

git-hulk:
> > We also need to update our website :) @git-hulk
> 
> Do you want to do it? If you don't have time, I can update the website later today.

Cool, thanks. @tanruixiang

caipengbo:
Yes, you can do it :) @tanruixiang

Issue #878, opened 2022-09-15
disableWAL option


tufitko:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

We are using kvrocks with client sharding, also we can lose data. So, WAL is useless for us and it just makes additional write pressure. 

### Solution

I made some changes to ensure about this: https://github.com/apache/incubator-kvrocks/commit/bb1ba2eb0e2897b17e73ec3af8320d82d4872a07

I can replace `rocksdb::WriteOptions()` to global WriteOptions with optional disabledWAL, what do you think about it?

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

git-hulk:
Thanks for @tufitko feedback. So you don't need the replication, right?

I thinks we can add a new config option to let users to disable the WAL, and also need to forbid the `slaveof` command when the WAL is disabled since the replication can't work without it. How do you think?

tufitko:
@git-hulk yeah, we don't need replicas. About `slaveof` sounds good, so I'm gonna make PR at the weekends

git-hulk:
> @git-hulk yeah, we don't need replicas. About `slaveof` sounds good, so I'm gonna make PR at the weekends

Cool, thanks. Can just reply error in the slaveof command if the WAL is disabled.

aleksraiden:
What is the benefits from disabling WAL? This mechanism, as I remember, used for parallels writers, thats can work without blocking too. And, of course, data safety are poor. This only reduce a disk I/O, but in sequential mode (like WAL) I think, its no a lot of problem. maybe you want to reduce storage - you can disable wal history (in archive directory inside). Maybe this can help you quickly?

tufitko:
@aleksraiden It's exactly for reducing a disk I/O. We have a lot of data to write, so any reduction of IO helps us.

aleksraiden:
@tufitko, ok, interested to me too. I have a medium-load system (streams, ~20K rps, 100M+ msg in history, but 7.2.2 rocksdb) based on kvrocks, I will interest to test this patch too.

PR #885, opened 2022-09-18
feat: add rocksdb write_options to configuration

tufitko:
#878
also, I've updated Dockerfile with Prerequisite command, because it didn't build
C++ isn't my common lang, so any comments are welcome

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -203,6 +203,14 @@ Config::Config() {
         false, new IntField(&RocksDB.max_bytes_for_level_multiplier, 10, 1, 100)},
       {"rocksdb.level_compaction_dynamic_level_bytes",
         false, new YesNoField(&RocksDB.level_compaction_dynamic_level_bytes, false)},
+
+      /* rocksdb write options */
+      {"rocksdb.write_options.sync", true, new YesNoField(&RocksDB.write_options.sync, false)},
+      {"rocksdb.write_options.disable_wal", true, new YesNoField(&RocksDB.write_options.disable_WAL, false)},
+      {"rocksdb.write_options.no_slowdown", true, new YesNoField(&RocksDB.write_options.no_slowdown, false)},
+      {"rocksdb.write_options.low_pri", true, new YesNoField(&RocksDB.write_options.low_pri, false)},
+      {"rocksdb.write_options.memtable_insert_hint_per_batch",
+        true, new YesNoField(&RocksDB.write_options.memtable_insert_hint_per_batch, false)},
   };
   for (auto &wrapper : fields) {
     auto &field = wrapper.field;
--- a/src/storage.cc
+++ b/src/storage.cc
@@ -73,6 +73,7 @@ Storage::Storage(Config *config)
   SetCheckpointCreateTime(0);
   SetCheckpointAccessTime(0);
   backup_creating_time_ = std::time(nullptr);
+  SetWriteOptions(config->RocksDB.write_options);
 }
 
 Storage::~Storage() {
@@ -92,6 +93,14 @@ void Storage::CloseDB() {
   for (auto handle : cf_handles_) db_->DestroyColumnFamilyHandle(handle);
   delete db_;
   db_ = nullptr;
+}
+
+void Storage::SetWriteOptions(const Config::RocksDB::WriteOptions& config) {
+  write_opts_.sync = config.sync;
+  write_opts_.disableWAL = config.disable_WAL;
+  write_opts_.no_slowdown = config.no_slowdown;
+  write_opts_.low_pri = config.low_pri;
+  write_opts_.memtable_insert_hint_per_batch = config.memtable_insert_hint_per_batch;
 }
 
 rocksdb::BlockBasedTableOptions Storage::InitTableOptions() {
@@ -533,7 +542,7 @@ rocksdb::Status Storage::DeleteRange(const std::string &first_key, const std::st
   if (!s.ok()) {
     return s;
   }
-  return Write(rocksdb::WriteOptions(), &batch);
+  return Write(write_opts_, &batch);
 }
 
 rocksdb::Status Storage::FlushScripts(const rocksdb::WriteOptions &options, rocksdb::ColumnFamilyHandle *cf_handle) {
@@ -547,7 +556,7 @@ rocksdb::Status Storage::FlushScripts(const rocksdb::WriteOptions &options, rock
   if (!s.ok()) {
     return s;
   }
-  return Write(rocksdb::WriteOptions(), &batch);
+  return Write(options, &batch);
 }
 
 Status Storage::ReplicaApplyWriteBatch(std::string &&raw_batch) {
@@ -555,7 +564,7 @@ Status Storage::ReplicaApplyWriteBatch(std::string &&raw_batch) {
     return Status(Status::NotOK, "reach space limit");
   }
   auto bat = rocksdb::WriteBatch(std::move(raw_batch));
-  auto s = db_->Write(rocksdb::WriteOptions(), &bat);
+  auto s = db_->Write(write_opts_, &bat);
   if (!s.ok()) {
     return Status(Status::NotOK, s.ToString());
   }
@@ -647,7 +656,7 @@ Status Storage::WriteToPropagateCF(const std::string &key, const std::string &va
 
   auto cf = GetCFHandle(kPropagateColumnFamilyName);
   batch.Put(cf, key, value);
-  auto s = Write(rocksdb::WriteOptions(), &batch);
+  auto s = Write(write_opts_, &batch);
   if (!s.ok()) {
     return Status(Status::NotOK, s.ToString());
   }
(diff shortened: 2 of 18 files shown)


tanruixiang:
May I ask why change the storage interface. I think it is necessary to reserve this `WriteOptions` parameter. Not all types share the same storage's writeoption(Even though all are now using the same defaults, I think it's still necessary to keep).

tufitko:
@tanruixiang I can keep old Write interface, I've removed it because I didn't see non-default write options. Can you give an example of when this might be needed?

tanruixiang:
> @tanruixiang I can keep old Write interface, I've removed it because I didn't see non-default write options. Can you give an example of when this might be needed?

There are many options in `WriteOptions`, not just `disableWAL`. If someone needs to change other options like you, and use different options in different places, they will need this `WriteOptions` parameter. In particular, the `rocksdb::WriteOptions write_opts_ ` you add are private, which means we can't make changes in places where options might need to be changed.

tufitko:
🤔 @tanruixiang  I think I need to return the old Write interface and add a method to return default write options from storage. Is it ok?

mapleFU:
Hi, I think we can provide a `defaultRocksDBOption` or `writeWithDefault` in `storage`, I think just create default `WriteOptions` everywhere is not a good idea, and only those who know what rocksdb's options means can define them themselves

tanruixiang:
> 🤔 @tanruixiang I think I need to return the old Write interface and add a method to return default write options from storage. Is it ok?

We can discuss it. @git-hulk @PragmaTwice What form do you think is the best?

Dockerfile
```diff
@@ -15,20 +15,20 @@
...
-FROM ubuntu:focal as build
+FROM ubuntu:jammy as build
 
 # workaround tzdata install hanging
 ENV TZ=Asia/Shanghai
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
 RUN apt update
-RUN apt install -y cmake make git autoconf libtool g++ python3
+RUN apt install -y git gcc g++ make cmake autoconf automake libtool python3 libssl-dev
```
PragmaTwice: I pick this line to #887, since this PR may not be merged quickly. Thanks for your contribution!

git-hulk:
I also prefer to add the `WriteWithDefault` to avoid creating the option everywhere, then we can keep the `Write` method to be consistent with RocksDB.

Dockerfile
tufitko: ok! Then I remove it from my PR

tufitko:
So, I've reworked PR with `WriteWithDefault`. What about `Delete`? There was write_options and I removed it.

PragmaTwice:
> So, I've reworked PR with `WriteWithDefault`. What about `Delete`? There was write_options and I removed it.

I think you can recover the WriteOptions parameter, and just expose a method `WriteOptions Storage::DefaultWriteOptions()`.

And you can also add something like `DeleteWithDefault(...) { Delete(DefaultWriteOptions(), ...) }` if you like.

tufitko:
> > So, I've reworked PR with `WriteWithDefault`. What about `Delete`? There was write_options and I removed it.
> 
> I think you can recover the WriteOptions parameter, and just expose a method `WriteOptions Storage::DefaultWriteOptions()`.
> 
> And you can also add something like `DeleteWithDefault(...) { Delete(DefaultWriteOptions(), ...) }` if you like.

reworked again 🙈

src/storage.cc
```diff
@@ -511,7 +512,7 @@ rocksdb::Status Storage::Write(const rocksdb::WriteOptions &options, rocksdb::Wr
     updates->PutLogData(ServerLogData(kReplIdLog, replid_).Encode());
   }
 
-  return db_->Write(options, updates);
+  return db_->Write(write_opts_, updates);
```
PragmaTwice:
```suggestion
  return db_->Write(options, updates);
```

tufitko: ooops

tufitko: fixed

src/config.cc
```diff
@@ -174,6 +174,7 @@ Config::Config() {
       {"rocksdb.delayed_write_rate", false, new Int64Field(&RocksDB.delayed_write_rate, 0, 0, INT64_MAX)},
       {"rocksdb.wal_ttl_seconds", true, new IntField(&RocksDB.WAL_ttl_seconds, 3*3600, 0, INT_MAX)},
       {"rocksdb.wal_size_limit_mb", true, new IntField(&RocksDB.WAL_size_limit_MB, 16384, 0, INT_MAX)},
+      {"rocksdb.disableWAL", true, new YesNoField(&RocksDB.disableWAL, false)},
```
PragmaTwice: I think `disable_wal` is more suitable in configuration file? since other options are all underscore-case and case-insensitive.

tanruixiang:
As @PragmaTwice  said. I think it is possible to increase the configuration of default writeoption in the configuration. Doing this eliminates the need to recompile the code every time the writeoption is modified.

src/config.cc
tufitko: `disable_wal` looks better, but in RocksDb it calls `disableWAL`. shouldn't we save the original name?

tufitko:
> As @PragmaTwice said. I think it is possible to increase the configuration of default writeoption in the configuration. Doing this eliminates the need to recompile the code every time the writeoption is modified.

hmmm, I can add `sync`, `no_slowdown`, `low_pri`, `memtable_insert_hint_per_batch`. But `ignore_missing_column_families` and `timestamp` looks useless to setting through the config. do you agree with this?

tanruixiang:
> > As @PragmaTwice said. I think it is possible to increase the configuration of default writeoption in the configuration. Doing this eliminates the need to recompile the code every time the writeoption is modified.
> 
> hmmm, I can add `sync`, `no_slowdown`, `low_pri`, `memtable_insert_hint_per_batch`. But `ignore_missing_column_families` and `timestamp` looks useless to setting through the config. do you agree with this?

I think these parameters can be reserved, and synchronization in the config file(For example, adding an interface to change the `default writeoption`. If the configuration file is configured with `default writeoption`, then make corresponding modifications.). Of course, if it is not in the configuration file, the default value is used.

tanruixiang:
> I think these parameters can be reserved, and synchronization in the config file(For example, adding an interface to change the `default writeoption`. If the configuration file is configured with `default writeoption`, then make corresponding modifications.). Of course, if it is not in the configuration file, the default value is used.

After this modification, if you want to `disableWAL`, you only need to modify the configuration file.

src/config.cc
git-hulk: I also found this point, I think it should be inconsistent in RocksDB NOT the intent. `disable_wal` even `disable_WAL` is better than `disableWAL`, what do you think?

tufitko:
@tanruixiang updated, I'm not sure I understood correctly. 
I didn't add `timestamp` because I still think it's useless. also, there isn't it in newer versions of rocksdb.

PragmaTwice:
I prefer to expose only the options that may be used in practice. If we increase the number of options exposed, then we need to consider how kvrocks will behave if these options are used, and whether all aspects of the functionality in kvrocks will still work. 

If we can't be sure of that, then just exposing `disbale_wal` is fine. If you expose other options, you may need to do some additional explaining: such as whether you tested how kvrocks works when those options are used, and under what circumstances they will be used.

BTW, do not forget to update `kvrocks.conf` : )

tufitko:
I need a decision about exposing write options. 
I think `ignore_missing_column_families` and `timestamp` are unneeded for configuring. Others can be useful

tanruixiang:
There are some parameter settings in writeoption that should be safe, such as `sync`. I think it's okay to keep these parameters, and if the user uses a parameter in config file that we're not sure about, then we give a warning (I think for users unfamiliar with writeoption, they'll just use the default values, and users who modify those values are sure Is to have a certain understanding of rocksdb).

PragmaTwice (dismissed):
LGTM

src/config.cc
```diff
@@ -603,6 +611,9 @@ Status Config::finish() {
     auto s = rocksdb::Env::Default()->CreateDirIfMissing(name);
     if (!s.ok()) return Status(Status::NotOK, s.ToString());
   }
+  if (!slaveof_.empty() && RocksDB.write_options.disable_WAL) {
+    return Status(Status::NotOK, "cannot disable the WAL while using Master-Slave replication");
+  }
```
git-hulk: I think we should check this in the `slaveof` command.

git-hulk (commented):
Others are good to me except one comment.

src/config.cc
tufitko: ok, moved to `Execution` of slaveof

git-hulk (approved):

src/config.cc
git-hulk: Thank you!

PragmaTwice (approved):

PR #895, opened 2022-09-19
Support hrange command

### LINKED ISSUE #894: Support Hrange command
### Search before asking

### Motivation

According to [kvrocks 3.0 roadmap](https://github.com/apache/incubator-kvrocks/projects/2#card-81014194), add hrange command.

### Solution

Since kvrocks uses rocksdb as the storage engine, keys are ordered in sstable. So we can easily provide a range query command for hash.
We give the following command:
hrange hashkey $start $end [LIMIT $number]
For the specified hashkey, We will get the kv-pair with the kv-pair's key located in [start, stop) interval in the hash. If the limit is set, at most limit kv pairs will be returned.

tanruixiang:
It closes #894

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -1462,14 +1462,10 @@ class CommandHRange : public Commander {
     if (args.size() == 6 && Util::ToLower(args[4]) != "limit") {
       return Status(Status::RedisInvalidCmd, errInvalidSyntax);
     }
-    try {
-      start_ = std::stol(args[2]);
-      stop_ = std::stol(args[3]);
-      if (args.size() == 6) {
-        limit_ = std::stol(args[5]);
-      }
-    } catch (const std::exception& ) {
-      return Status(Status::RedisParseErr, errValueNotInterger);
+    if (args.size() == 6) {
+      auto parseResult = ParseInt<int>(args_[5], /* base= */ 10);
+      if (!parseResult.IsOK())return Status(Status::RedisParseErr, errValueNotInterger);
+      limit_ = parseResult.GetValue();
     }
     return Commander::Parse(args);
   }
@@ -1480,7 +1476,7 @@ class CommandHRange : public Commander {
     }
     Redis::Hash hash_db(svr->storage_, conn->GetNamespace());
     std::vector<FieldValue> field_values;
-    rocksdb::Status s = hash_db.Range(args_[1], start_, stop_, limit_, &field_values);
+    rocksdb::Status s = hash_db.Range(args_[1], args_[2], args_[3], limit_, &field_values);
     if (!s.ok()) {
       return Status(Status::RedisExecErr, s.ToString());
     }
@@ -1493,7 +1489,7 @@ class CommandHRange : public Commander {
   }
 
  private:
-  int start_ = 0, stop_ = 0, limit_ = -1;
+  int limit_ = -1;
 };
 
 class CommandPush : public Commander {
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -23,12 +23,14 @@
 #include <sys/socket.h>
 #include <algorithm>
 #include <cctype>
+#include <climits>
 #include <cmath>
 #include <chrono>
 #include <thread>
 #include <utility>
 #include <memory>
 #include <glog/logging.h>
+#include <vector>
 
 #include "redis_db.h"
 #include "redis_cmd.h"
@@ -1427,10 +1429,11 @@ class CommandHVals : public Commander {
     if (!s.ok()) {
       return Status(Status::RedisExecErr, s.ToString());
     }
-    *output = "*" + std::to_string(field_values.size()) + CRLF;
-    for (const auto &fv : field_values) {
-      *output += Redis::BulkString(fv.value);
+    std::vector<std::string> values;
+    for(const auto &p : field_values) {
+      values.emplace_back(p.value);
     }
+    *output = MultiBulkString(values);
     return Status::OK();
   }
 };
@@ -1444,11 +1447,12 @@ class CommandHGetAll : public Commander {
     if (!s.ok()) {
       return Status(Status::RedisExecErr, s.ToString());
     }
-    *output = "*" + std::to_string(field_values.size() * 2) + CRLF;
-    for (const auto &fv : field_values) {
-      *output += Redis::BulkString(fv.field);
-      *output += Redis::BulkString(fv.value);
+    std::vector<std::string> kv_pairs;
+    for(const auto &p : field_values) {
+      kv_pairs.emplace_back(p.field);
+      kv_pairs.emplace_back(p.value);
     }
+    *output = MultiBulkString(kv_pairs);
     return Status::OK();
   }
 };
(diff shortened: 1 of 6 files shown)


src/redis_cmd.cc
```diff
@@ -1453,6 +1453,48 @@ class CommandHGetAll : public Commander {
...
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() != 6 && args.size() != 4) {
+      return Status(Status::RedisParseErr, errWrongNumOfArguments);
+    }
+    if (args.size() == 6 && Util::ToLower(args[4]) != "limit") {
+      return Status(Status::RedisInvalidCmd, errInvalidSyntax);
+    }
+    try {
+      start_ = std::stol(args[2]);
```
git-hulk: @tanruixiang The start/stop should be the string range instead of the index.

src/redis_cmd.cc
```diff
@@ -1453,6 +1453,48 @@ class CommandHGetAll : public Commander {
...
+      stop_ = std::stol(args[3]);
+      if (args.size() == 6) {
+        limit_ = std::stol(args[5]);
```
git-hulk: Can use `ParseInt` here

src/redis_cmd.cc
tanruixiang: > @tanruixiang The start/stop should be the string range instead of the index.

Ok, I misunderstood the meaning of start and stop. I'll revise it right away.

src/redis_cmd.cc
tanruixiang: > Can use `ParseInt` here

I noticed that std::stol is used in quite a few places in the kvrocks. Should I open another issue to replace these std::stol?

git-hulk: Yes, I think we can do it in another PR to make the context clear.

src/redis_cmd.cc
git-hulk: 👍 The issue also didn't clarify those parameters clearly.

src/redis_cmd.cc
```diff
@@ -1453,6 +1453,45 @@ class CommandHGetAll : public Commander {
...
+    rocksdb::Status s = hash_db.Range(args_[1], args_[2], args_[3], limit_, &field_values);
+    if (!s.ok()) {
+      return Status(Status::RedisExecErr, s.ToString());
+    }
+    *output = "*" + std::to_string(field_values.size() * 2) + CRLF;
+    for (const auto &fv : field_values) {
+      *output += Redis::BulkString(fv.field);
+      *output += Redis::BulkString(fv.value);
+    }
+    return Status::OK();
```
caipengbo: Can we use `Redis::MultiBulkString`?

src/redis_cmd.cc
```diff
@@ -1453,6 +1453,45 @@ class CommandHGetAll : public Commander {
...
+      if (!parseResult.IsOK())return Status(Status::RedisParseErr, errValueNotInterger);
+      limit_ = parseResult.GetValue();
+    }
+    return Commander::Parse(args);
+  }
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    if (args_.size() == 6 && limit_ < 0) {
+      *output = "*" + std::to_string(0) + CRLF;
+      return Status::OK();
+    }
```
caipengbo: This part of the logic, I think we should put it into the `Hash::Range`

tanruixiang: > This part of the logic, I think we should put it into the `Hash::Range`

Thank you for your review. Cannot get `args_` in `Hash::Range`. It seems unreasonable to pass it as a parameter.

src/redis_cmd.cc
tanruixiang: > Can we use `Redis::MultiBulkString`?

Yes, But it need convert  `std::vector<FieldValue> field_values` to `std::vector<string> field_values`. Is there an elegant iterator conversion method in c++?

src/redis_cmd.cc
caipengbo: I mean use `limit_`  instead of `args_`.
If you don't want that, maybe it's better to return `Redis::NilString()` here.

src/redis_cmd.cc
caipengbo: I personally don't think it's very elegant to assemble RESP here. Maybe we could wrap it as a function that can passes `FieldValue`, or we could assemble it as `vector<string>` and pass it to our existing `Redis::MultiBulkString`.

src/redis_cmd.cc
tanruixiang: > I mean use `limit_` instead of `args_`. If you don't want that, maybe it's better to return `Redis::NilString()` here.

OK. I need to modify the current logic, after the modification, it should be able to meet the requirements you said.

src/redis_cmd.cc
tanruixiang: > I personally don't think it's very elegant to assemble RESP here. Maybe we could wrap it as a function that can passes `FieldValue`, or we could assemble it as `vector<string>` and pass it to our existing `Redis::MultiBulkString`

Yes. There are several places in the code where this direct assembly method is used. I can assemble them into a vector<string> and send it to MultiBulkString to solve it.

git-hulk:
Thanks for @tanruixiang contribution, this PR looks good to me. Do you mind adding the `gocase` for the hrange command as well first, then we can migrate the other test cases later.

tanruixiang:
> Thanks for @tanruixiang contribution, this PR looks good to me. Do you mind adding the `gocase` for the hrange command as well first, then we can migrate the other test cases later.

Yes. Originally, I wanted to complete the migration of `hash.tcl` by the way. But I see that issue seems to be assigned. ~Can I add to gocase after he is done migrating?~ I think I can write gocase for hrange first.

git-hulk:
> > Thanks for @tanruixiang contribution, this PR looks good to me. Do you mind adding the `gocase` for the hrange command as well first, then we can migrate the other test cases later.
> 
> Yes. Originally, I wanted to complete the migration of `hash.tcl` by the way. But I see that issue seems to be assigned. ~Can I add to gocase after he is done migrating?~ I think I can write gocase for hrange first.

Sorry that I missed this comment, it's ok to me.

git-hulk (dismissed):

src/redis_cmd.cc
```diff
@@ -1444,13 +1447,50 @@ class CommandHGetAll : public Commander {
...
+    for (const auto &p : field_values) {
+      kv_pairs.emplace_back(p.field);
+      kv_pairs.emplace_back(p.value);
+    }
+    *output = MultiBulkString(kv_pairs);
     return Status::OK();
   }
+
+ private:
+  int64_t limit_ = LONG_MAX;
```
PragmaTwice, line 1493: I think maybe `limit_` should be an unsigned integer?

You can use `ParseInt<unsigned ...>` which will return error when parsing a negative number.

caipengbo: Maybe the syntax is such that the user can enter a negative number, but it will return an empty array.

tanruixiang: > 

hash can store at most uint32_t entry, So it is ok.

caipengbo (approved):

src/redis_cmd.cc
tanruixiang: > Maybe the syntax is such that the user can enter a negative number, but it will return an empty array.

Yes, When user input negative number, It will return empty array.

tanruixiang: > > Maybe the syntax is such that the user can enter a negative number, but it will return an empty array.
> 
> Yes, When user input negative number, It will return empty array.

You can check this case in hash_test.go.

PragmaTwice: got it. although I prefer more constrained and strongly typed behaviors.

PragmaTwice (approved):

git-hulk (approved):

git-hulk:
Thanks all, merging...

PR #897, opened 2022-09-19
Replace std::stol with ParseInt

### LINKED ISSUE #896: Replace std::stol with ParseInt.
### Search before asking

### Motivation

There are many places in the code that use std::stol. Replace std::stol with ParseInt. Use ParseInt instead of std::stol for safer parsing.

### Solution

_No response_

tanruixiang:
It closes #896


src/redis_cmd.cc
```diff
@@ -488,16 +488,12 @@ class CommandSet : public Commander {
...
-          std::string s = args_[++i];
-          std::string::size_type sz;
-          ttl_ms = std::stol(s, &sz);
-          if (sz != s.size()) {
-            return Status(Status::RedisParseErr, errValueNotInterger);
-          }
-        } catch (std::exception &e) {
+        std::string s = args_[++i];
+        auto parseResult = ParseInt<int64_t>(s, /* base= */ 10);
+        if (!parseResult.IsOK()) {
```
```suggestion
        if (!parseResult) {
```

src/redis_cmd.cc
```diff
@@ -820,12 +816,11 @@ class CommandDel : public Commander {
...
 Status getBitOffsetFromArgument(std::string arg, uint32_t *offset) {
-  int64_t offset_arg = 0;
-  try {
-    offset_arg = std::stoll(arg);
-  } catch (std::exception &e) {
+  auto parseResult = ParseInt<int64_t>(arg, /* base= */ 10);
+  if (!parseResult.IsOK()) {
     return Status(Status::RedisParseErr, errValueNotInterger);
   }
+  int64_t offset_arg = offset_arg = parseResult.GetValue();
```
```suggestion
  int64_t offset_arg = *parseResult;
```

src/util.cc
```diff
@@ -341,26 +342,26 @@ int GetLocalPort(int fd) {
...
-    }
-  } catch (std::exception &e) {
+  auto parseResult = ParseInt<int64_t>(str, /* base= */ 10);
+  if (!parseResult.IsOK()) {
     return Status(Status::NotOK, "value is not an integer or out of range");
   }
+  *n = parseResult.GetValue();
+  if (max > min && (*n < min || *n > max)) {
+    return Status(Status::NotOK, "value should between "+std::to_string(min)+" and "+std::to_string(max));
+  }
```
PragmaTwice: you can use another overload of ParseInt which can accept a range.

src/redis_cmd.cc
```diff
@@ -488,16 +488,12 @@ class CommandSet : public Commander {
...
-        try {
-          std::string s = args_[++i];
-          std::string::size_type sz;
-          ttl_ms = std::stol(s, &sz);
-          if (sz != s.size()) {
-            return Status(Status::RedisParseErr, errValueNotInterger);
-          }
-        } catch (std::exception &e) {
+        std::string s = args_[++i];
+        auto parseResult = ParseInt<int64_t>(s, /* base= */ 10);
```
PragmaTwice, line 492: Actually the `<int64_t>` can be omitted.

PragmaTwice: not `int64_t`, be careful with integer types.

src/redis_cmd.cc
```diff
@@ -564,17 +560,17 @@ class CommandSetEX : public Commander {
...
-    try {
-      auto ttl_ms = std::stol(args[2]);
-      if (ttl_ms <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
-      if (ttl_ms > 0 && ttl_ms < 1000) {
-        ttl_ = 1;
-      } else {
-        ttl_ = ttl_ms / 1000;
-      }
-    } catch (std::exception &e) {
+    auto parseResult = ParseInt<int64_t>(args[2], /* base= */ 10);
```
PragmaTwice: not `int64_t`

src/redis_cmd.cc
```diff
@@ -889,12 +884,16 @@ class CommandBitCount : public Commander {
       return Status(Status::RedisParseErr, errInvalidSyntax);
     }
     if (args.size() == 4) {
-      try {
-        start_ = std::stol(args[2]);
-        stop_ = std::stol(args[3]);
-      } catch (std::exception &e) {
+      auto parseArgs2Result = ParseInt<int64_t>(args[2], /* base= */ 10);
```
PragmaTwice: same as above

src/redis_cmd.cc
```diff
@@ -1104,20 +1109,20 @@ class CommandPExpire : public Commander {
...
-        seconds_ = 1;
-      } else {
-        seconds_ = ttl_ms / 1000;
-        if (seconds_ >= INT32_MAX - now) {
-          return Status(Status::RedisParseErr, "the expire time was overflow");
-        }
-      }
-      seconds_ += now;
-    } catch (std::exception &e) {
+    auto parseResult = ParseInt<int64_t>(args[2], /* base= */ 10);
```
PragmaTwice: you can just remove `/* base= */`

tests/cppunit/t_hash_test.cc
```diff
@@ -114,7 +115,11 @@ TEST_F(RedisHashTest, HIncr) {
   }
   std::string bytes;
   hash->Get(key_, field, &bytes);
-  value = std::stoll(bytes);
+  auto parseResult = ParseInt<int64_t>(bytes, /* base= */ 10);
+  if (!parseResult.IsOK()) {
+    EXPECT_TRUE(false);
```
PragmaTwice:
```suggestion
    FAIL();
```

tests/cppunit/t_hash_test.cc
```diff
@@ -114,7 +115,11 @@ TEST_F(RedisHashTest, HIncr) {
...
+  }
+  value = parseResult.GetValue();
   EXPECT_EQ(32, value);
```
PragmaTwice:
```suggestion
  EXPECT_EQ(32, *parseResult);
```

tanruixiang:
@PragmaTwice  Thanks for your warm help. I'm just going to upload a draft to save my progress and keep doing it tomorrow. I didn't even check it myself. You have helped me. Thank you very much😆

src/redis_cmd.cc
```diff
@@ -5363,13 +5368,12 @@ class CommandXRead : public Commander {
...
-        try {
-          with_count_ = true;
-          count_ = static_cast<uint64_t>(std::stoll(args[i+1]));
-        } catch (const std::exception &) {
+        with_count_ = true;
+        auto parseResult = ParseInt<int64_t>(args[i+1], /* base= */ 10);
+        if (!parseResult.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
+        count_ = static_cast<uint64_t>(parseResult.GetValue());
```
PragmaTwice: you can use `ParseInt<unsigned something>`

src/redis_cmd.cc
tanruixiang: > 

Hi. This type is int64_t.

src/redis_cmd.cc
tanruixiang: Hi. This type is int64_t.

tanruixiang:
@PragmaTwice  Hi. The original code uses `std:: stol` , but using `ParseInt<long int> `instead the lint will report an error. Do you have any better suggestions?

src/redis_cmd.cc
```diff
@@ -564,17 +560,17 @@ class CommandSetEX : public Commander {
...
-    try {
-      auto ttl_ms = std::stol(args[2]);
-      if (ttl_ms <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
-      if (ttl_ms > 0 && ttl_ms < 1000) {
-        ttl_ = 1;
-      } else {
-        ttl_ = ttl_ms / 1000;
-      }
-    } catch (std::exception &e) {
+    auto parseResult = ParseInt<long int>(args[2], 10);
```
```suggestion
    auto parseResult = ParseInt<int>(args[2], 10);
```
`ttl_` is typed `int`

src/redis_cmd.cc
```diff
@@ -1104,20 +1109,20 @@ class CommandPExpire : public Commander {
...
-        seconds_ = 1;
-      } else {
-        seconds_ = ttl_ms / 1000;
-        if (seconds_ >= INT32_MAX - now) {
-          return Status(Status::RedisParseErr, "the expire time was overflow");
-        }
-      }
-      seconds_ += now;
-    } catch (std::exception &e) {
+    auto parseResult = ParseInt<long int>(args[2], 10);
```
PragmaTwice: maybe uint64_t is ok here.

src/redis_cmd.cc
```diff
@@ -1517,16 +1522,16 @@ class CommandPop : public Commander {
...
-    } catch (const std::exception& ) {
+    auto parseResult = ParseInt<int32_t>(args[2], 10);
+    if (!parseResult) {
       return Status(Status::RedisParseErr, errValueNotInterger);
     }
+    int32_t v = *parseResult;
+    if (v < 0) {
+      return Status(Status::RedisParseErr, errValueMustBePositive);
+    }
+    count_ = v;
```
PragmaTwice, line 1533: `v` is unnecessary

src/redis_cmd.cc
```diff
@@ -1517,16 +1522,16 @@ class CommandPop : public Commander {
...
-        return Status(Status::RedisParseErr, errValueMustBePositive);
-      }
-      count_ = v;
-      with_count_ = true;
-    } catch (const std::exception& ) {
+    auto parseResult = ParseInt<int32_t>(args[2], 10);
+    if (!parseResult) {
       return Status(Status::RedisParseErr, errValueNotInterger);
     }
+    int32_t v = *parseResult;
```
PragmaTwice: actually we prefer underscore-case for local variable, and for here, you can use `v` instead of `parseResult`

src/redis_cmd.cc
```suggestion
    auto v = ParseInt<int32_t>(args[2], 10);
    if (!v) {
      return Status(Status::RedisParseErr, errValueNotInterger);
    }
    if (*v < 0) {
      return Status(Status::RedisParseErr, errValueMustBePositive);
    }
    count_ = *v;
```

src/redis_cmd.cc
tanruixiang: `Long int` and `int` are not equal on some machines. Do you have any good suggestions about this?

tanruixiang: Maybe we can just use int64_t, because the next code will check the range.
`
if (ttl_ms > 0 && ttl_ms < 1000) {
      ttl_ = 1;
    } else {
      ttl_ = ttl_ms / 1000;
    }
`

src/redis_cmd.cc
tanruixiang: Yes. I think it would be more simpler to replace parseResult with v.

src/redis_cmd.cc
tanruixiang: This is same above.

src/redis_cmd.cc
```diff
@@ -820,12 +815,11 @@ class CommandDel : public Commander {
 };
 
 Status getBitOffsetFromArgument(std::string arg, uint32_t *offset) {
-  int64_t offset_arg = 0;
-  try {
-    offset_arg = std::stoll(arg);
-  } catch (std::exception &e) {
+  auto parseResult = ParseInt<int64_t>(arg, 10);
```
PragmaTwice: uint32_t

tanruixiang: @PragmaTwice  Hi. I think that in this simple pr, we do not need to modify the data type according to logic, and it is important to restore the original data type as much as possible. I think we can reopen a PR to make these logical adjustments. What do you think.

PragmaTwice: If you parse a int64_t, and then check whether its range is exactly in (0, UINT32_MAX), you can just parse a uint32_t instead.

PragmaTwice: BTW, you can just return the original status.
```
return res.ToStatus();
```

src/redis_cmd.cc
```diff
@@ -889,12 +883,16 @@ class CommandBitCount : public Commander {
...
-      try {
-        start_ = std::stol(args[2]);
-        stop_ = std::stol(args[3]);
-      } catch (std::exception &e) {
+      auto parseArgs2Result = ParseInt<int64_t>(args[2], 10);
+      if (!parseArgs2Result) {
+        return Status(Status::RedisParseErr, errValueNotInterger);
+      }
+      start_ = *parseArgs2Result;
+      auto parseArgs3Result = ParseInt<int64_t>(args[3], 10);
```
PragmaTwice: just `arg_stop`, `parse_stop`, `res` or something like that is ok

src/redis_cmd.cc
tanruixiang: > If you parse a int64_t, and then check whether its range is exactly in (0, UINT32_MAX), you can just parse a uint32_t instead.

yes I know what you mean. I mean, should this pr be defined as trying to restore the original code.

tanruixiang: That is to say, the type used in the original code shoule be the same here. In this pr, regardless of whether the type can be adjusted logically. If the type can be modified, we can do it in the next Pr?

src/util.cc
```diff
@@ -341,26 +342,20 @@ int GetLocalPort(int fd) {
...
-  try {
-    *n = static_cast<int64_t>(std::stoll(str, nullptr, 8));
-    if (max > min && (*n < min || *n > max)) {
-      return Status(Status::NotOK, "value should between "+std::to_string(min)+" and "+std::to_string(max));
-    }
-  } catch (std::exception &e) {
-    return Status(Status::NotOK, "value is not an integer or out of range");
+  auto parseResult = ParseInt<int64_t>(str, NumericRange<int64_t>{min, max}, 8);
+  if (!parseResult) {
+    return Status(Status::NotOK, parseResult.Msg());
```
```suggestion
    return parseResult.ToStatus();
```

tanruixiang:
> That is to say, the type used in the original code shoule be the same here. In this pr, regardless of whether the type can be adjusted logically. If the type can be modified, we can do it in the next Pr?

@PragmaTwice  HI.  What do you think about this. If you want to modify it in this pr, I will modify it immediately, otherwise I will modify it to another pr.

PragmaTwice:
> > That is to say, the type used in the original code shoule be the same here. In this pr, regardless of whether the type can be adjusted logically. If the type can be modified, we can do it in the next Pr?
> 
> @PragmaTwice HI. What do you think about this. If you want to modify it in this pr, I will modify it immediately, otherwise I will modify it to another pr.

I think it is ok to change it, since it is trivial.

tanruixiang:
> > > That is to say, the type used in the original code shoule be the same here. In this pr, regardless of whether the type can be adjusted logically. If the type can be modified, we can do it in the next Pr?
> > 
> > 
> > @PragmaTwice HI. What do you think about this. If you want to modify it in this pr, I will modify it immediately, otherwise I will modify it to another pr.
> 
> I think it is ok to change it, since it is trivial.

Ok, I will change it .

PragmaTwice (approved):
LGTM, thanks for your contribution!

tests/tcl/tests/unit/type/hash.tcl
```diff
@@ -335,8 +335,8 @@ start_server {tags {"hash"}} {
         catch {r hincrby smallhash str 1} smallerr
         catch {r hincrby smallhash str 1} bigerr
         set rv {}
-        lappend rv [string match "ERR*not an integer*" $smallerr]
-        lappend rv [string match "ERR*not an integer*" $bigerr]
+        lappend rv [string match "ERR*non-integer*" $smallerr]
+        lappend rv [string match "ERR*non-integer*" $bigerr]
```
tisonkun, line 339: why do you need these changes?

tanruixiang: > why do you need these changes?

Use the `ToStatus()` method of class `StatusOr` to get error information. It is expressed slightly differently than it was hardcoded before.

PragmaTwice: I think these changes are acceptable. These messages come from `ParseInt`, and are forwarded to these callers.

tisonkun (approved):

PragmaTwice:
Hi @tanruixiang, I notice there is also some code using `std::stoul` or `std::stoull` in kvrocks, do you want to also replace them in this PR? Or it is OK to merge it and replace them in another PR.

tanruixiang:
> Hi @tanruixiang, I notice there is also some code using `std::stoul` or `std::stoull` in kvrocks, do you want to also replace them in this PR? Or it is OK to merge it and replace them in another PR.

OK I am willing to do this task. I think this PR information is a bit too much, we can continue to do this in the next PR.

PragmaTwice:
> > Hi @tanruixiang, I notice there is also some code using `std::stoul` or `std::stoull` in kvrocks, do you want to also replace them in this PR? Or it is OK to merge it and replace them in another PR.
> 
> OK I am willing to do this task. I think this PR information is a bit too much, we can continue to do this in the next PR.

Thanks. Maybe there are also some `atoi` and `std::stoi`, you can replace them too in another PR.

PragmaTwice:
Thanks. I will merge it soon.

Issue #893, opened 2022-09-19
After the single test, sending SIGTERM or SIGINT


ColinChamber:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

After the single test, sending SIGTERM or SIGINT that kvrocks can handle may be better than sending it directly to [KILL](https://github.com/apache/incubator-kvrocks/blob/unstable/tests/gocase/util/server.go#L62). We can also check whether kvrocks can exit gracefully.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #898, opened 2022-09-20
Send SIGTERM after finishing the unit test

ColinChamber:
#893 solved


tests/gocase/util/server.go
```diff
@@ -59,8 +60,13 @@ func (s *KvrocksServer) NewTCPClient() *tcpClient {
 }
 
 func (s *KvrocksServer) Close() {
-	require.NoError(s.t, s.cmd.Process.Kill())
-	require.EqualError(s.t, s.cmd.Wait(), "signal: killed")
+	require.NoError(s.t, s.cmd.Process.Signal(syscall.SIGTERM))
```
PragmaTwice, line 71: I am wondering, if kvrocks crashed, whether it will report an error? cc @tisonkun

tests/gocase/util/server.go
```diff
@@ -59,8 +60,13 @@ func (s *KvrocksServer) NewTCPClient() *tcpClient {
...
+	// A Pod is granted a term to terminate gracefully, which defaults to 30 seconds.
```
tisonkun: What does `Pod` stand for here? It seems a comment copied from somewhere that doesn't fit this situation.

tests/gocase/util/server.go
tisonkun: I see the `s.cmd.Wait` below so it should be fine.

tests/gocase/util/server.go
ColinChamber: Kubernetes will send a SIGTERM signal to the containers in the pod after deleting the pod. It waits for a specified time, called the termination grace period. By default, this is 30 seconds. If the containers are still running after the grace period, they are sent the SIGKILL signal and forcibly removed.

tisonkun: So here is a local integration test environment. For nothing, we should talk about Kubernetes things?

ColinChamber: Yeah, it is two separate things. I just want the wait time not to be a magic number.

ColinChamber: Users may want to exit gracefully in 30 seconds, so I change it like this.

tests/gocase/util/server.go
```diff
@@ -26,13 +26,16 @@ import (
...
 	"path/filepath"
+	"syscall"
 	"testing"
 	"time"
 
 	"github.com/go-redis/redis/v9"
 	"github.com/stretchr/testify/require"
 )
 
+const k8sDefaultGracePeriod = 30
```
tisonkun: Let's move this const to `util/constats.go` and bring the comment to this constant instead of inline.

tisonkun (commented):
Rest LGTM

tests/gocase/util/server.go
ColinChamber: OK. Thanks for your help!

tisonkun (approved):

tisonkun:
Thank you!

PR #899, opened 2022-09-20
test: harden quit_test and list_test

tisonkun:


tests/gocase/unit/quit/quit_test.go
```diff
@@ -45,7 +45,7 @@ func TestPipeQuit(t *testing.T) {
 
 		require.NoError(t, c.WriteArgs("PING"))
 		r, err = c.ReadLine()
-		require.EqualError(t, err, "EOF")
+		require.Error(t, err)
```
tisonkun, line 48: After `QUIT`, the connection can be disposed. So we can possibly receive "read tcp 127.0.0.1:53224->127.0.0.1:36579: read: connection reset by peer" instead of "EOF".

tests/gocase/unit/type/list/list_test.go
```diff
@@ -93,7 +94,7 @@ func TestZipList(t *testing.T) {
 	defer srv.Close()
 	ctx := context.Background()
 	rdb := srv.NewClientWithOption(&redis.Options{
-		MaxRetries: -1,
+		ReadTimeout: 10 * time.Second,
```
tisonkun, line 97: @vmihailenco could you help with reviewing whether this option is a good choice?

For issues described at https://github.com/apache/incubator-kvrocks/issues/872#issuecomment-1248825118.

git-hulk (approved):

tests/gocase/unit/type/list/list_test.go
vmihailenco: Yes, this seems worth a try. The default is 3 seconds.

In go-redis tests we also have a busy loop to wait until Redis Server is fully bootstrapped, but I am not sure whether it is needed for kvrocks:

```go
for i := 0; i < 100; i++ {
  if err := rdb.Ping().Err(); err != nil {
    time.Sleep(100*time.Millisecond)
    continue
  }
  break
}
```

Issue #891, opened 2022-09-19
Kvrocks sends negative ttl while migrating slot


ellutionist:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Version

OS:
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.1 LTS
Release:        22.04

Kvrocks:
v2.1.0

### Minimal reproduce step

1. Create a kvrocks cluster of 2 nodes:
```
127.0.0.1:6666> cluster nodes
KEp2lfd524nItcLDNm=nkrTgxxRa2OnyfCorGqiw 127.0.0.1:6666@16666 myself,master - 1663582260999 1663582261000 1 connected 0-8191
iv3zxA39N1kXg23BmGAchLynaNCG2K284QvpgklT 127.0.0.1:6667@16667 master - 1663582260999 1663582261000 1 connected 8192-16383
```

2. Set a key into one node, which expres in 5 seconds:
```
127.0.0.1:6666> set key00000000000000007100 key00000000000000007100 ex 5
OK
```

3. Check the slot:
```
127.0.0.1:6666> cluster keyslot key00000000000000007100
(integer) 35
```

4. After 5 seconds, migrate the slot to the other node, the migration will fail:
```
127.0.0.1:6666> clusterx migrate 35 iv3zxA39N1kXg23BmGAchLynaNCG2K284QvpgklT
OK
```

```
127.0.0.1:6666> cluster info
cluster_state:ok
cluster_slots_assigned:16384
cluster_slots_ok:16384
cluster_slots_pfail:0
cluster_slots_fail:0
cluster_known_nodes:2
cluster_size:2
cluster_current_epoch:1
cluster_my_epoch:1
migrating_slot: 35
destination_node: iv3zxA39N1kXg23BmGAchLynaNCG2K284QvpgklT
migrating_state: fail
```

5. Error log:
```
E20220919 18:14:55.837342 639656 slot_migrate.cc:495] [migrate] Got invalid response: -ERR invalid expire time, line length: 24
E20220919 18:14:55.837354 639656 slot_migrate.cc:793] [migrate] Wrong response
E20220919 18:14:55.837363 639656 slot_migrate.cc:320] [migrate] Failed to send left data in pipeline
E20220919 18:14:55.837378 639656 slot_migrate.cc:182] [migrate] Failed to send snapshot
[...]

PR #901, opened 2022-09-20
Support EXAT/PEXAT option in the set command

ellutionist:
PR to solve issue #891

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -477,28 +477,25 @@ class CommandSet : public Commander {
       } else if (opt == "xx" && !nx_) {
         xx_ = true;
       } else if (opt == "ex" && !ttl_ && !last_arg) {
-        std::string s = args_[++i];
-        auto parse_status = ParseInt<int>(s);
-        if (!parse_status.IsOK()) {
+        auto parse_result = ParseInt<int>(args_[++i], 10);
+        if (!parse_result) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
-        ttl_ = parse_status.GetValue();
+        ttl_ = *parse_result;
         if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
       } else if (opt == "exat" && !ttl_ && !expire_ && !last_arg) {
-        std::string s = args_[++i];
-        auto parse_status = ParseInt<int64_t>(s);
-        if (!parse_status.IsOK()) {
+        auto parse_result = ParseInt<int64_t>(args_[++i], 10);
+        if (!parse_result) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
-        expire_ = parse_status.GetValue();
+        expire_ = *parse_result;
         if (expire_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
       } else if (opt == "pxat" && !ttl_ && !expire_ && !last_arg) {
-        std::string s = args_[++i];
-        auto parse_status = ParseInt<uint64_t>(s);
-        if (!parse_status.IsOK()) {
+        auto parse_result = ParseInt<uint64_t>(args[++i], 10);
+        if (!parse_result) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
-        uint64_t expire_ms = parse_status.GetValue();
+        uint64_t expire_ms = *parse_result;
         if (expire_ms <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
         if (expire_ms < 1000) {
           expire_ = 1;
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -478,7 +478,7 @@ class CommandSet : public Commander {
         xx_ = true;
       } else if (opt == "ex" && !ttl_ && !last_arg) {
         std::string s = args_[++i];
-        StatusOr<int> parse_status = ParseInt(s);
+        auto parse_status = ParseInt<int>(s);
         if (!parse_status.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
@@ -486,7 +486,7 @@ class CommandSet : public Commander {
         if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
       } else if (opt == "exat" && !ttl_ && !expire_ && !last_arg) {
         std::string s = args_[++i];
-        StatusOr<int64_t> parse_status = ParseInt(s);
+        auto parse_status = ParseInt<int64_t>(s);
         if (!parse_status.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
@@ -494,7 +494,7 @@ class CommandSet : public Commander {
         if (expire_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
       } else if (opt == "pxat" && !ttl_ && !expire_ && !last_arg) {
         std::string s = args_[++i];
-        StatusOr<uint64_t> parse_status = ParseInt(s);
+        auto parse_status = ParseInt<uint64_t>(s);
         if (!parse_status.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -477,41 +477,28 @@ class CommandSet : public Commander {
       } else if (opt == "xx" && !nx_) {
         xx_ = true;
       } else if (opt == "ex" && !ttl_ && !last_arg) {
-        try {
-          std::string s = args_[++i];
-          std::string::size_type sz;
-          ttl_ = std::stoi(s, &sz);
-          if (sz != s.size()) {
-            return Status(Status::RedisParseErr, errValueNotInterger);
-          }
-        } catch (std::exception &e) {
+        std::string s = args_[++i];
+        StatusOr<int> parse_status = ParseInt(s);
+        if (!parse_status.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
+        ttl_ = parse_status.GetValue();
         if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
       } else if (opt == "exat" && !ttl_ && !expire_ && !last_arg) {
-        try {
-          std::string s = args_[++i];
-          std::string::size_type sz;
-          expire_ = std::stol(s, &sz);
-          if (sz != s.size()) {
-            return Status(Status::RedisParseErr, errValueNotInterger);
-          }
-        } catch (std::exception &e) {
+        std::string s = args_[++i];
+        StatusOr<int64_t> parse_status = ParseInt(s);
+        if (!parse_status.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
+        expire_ = parse_status.GetValue();
         if (expire_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
       } else if (opt == "pxat" && !ttl_ && !expire_ && !last_arg) {
-        uint64_t expire_ms = 0;
-        try {
-          std::string s = args_[++i];
-          std::string::size_type sz;
-          expire_ms = std::stoul(s, &sz);
-          if (sz != s.size()) {
-            return Status(Status::RedisParseErr, errValueNotInterger);
-          }
-        } catch (std::exception &e) {
+        std::string s = args_[++i];
+        StatusOr<uint64_t> parse_status = ParseInt(s);
+        if (!parse_status.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
+        uint64_t expire_ms = parse_status.GetValue();
         if (expire_ms <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
         if (expire_ms < 1000) {
           expire_ = 1;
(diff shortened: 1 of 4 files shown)


src/slot_migrate.cc
```diff
@@ -568,6 +568,12 @@ Status SlotMigrate::MigrateOneKey(rocksdb::Slice key, std::string *restore_cmds)
     return Status(Status::cOK, "empty");;
   }
 
+  int64_t now;
+  rocksdb::Env::Default()->GetCurrentTime(&now);
+  if (metadata.expire <= now) {
```
git-hulk: expire = 0 means never be expired, can use `metadata.Expired()`.

datavisorchengpeng: Thank you for reminding me to check that `metadata.expire` must be greater than 0.
The reason why I did not use `metadata.Expired()` is that this method returns `false` when `metadata.expire == now`, which will make the source send a ttl of 0 to the destination, thus ruin the whole slot migration. I happened to meet such case before.

ellutionist: Thank you for reminding me to check that `metadata.expire` must be greater than 0.
The reason why I did not use `metadata.Expired()` is that this method returns `false` when `metadata.expire == now`, which will make the source send a ttl of 0 to the destination, thus ruin the whole slot migration. I happened to meet such case before.

git-hulk: I think there are two issues in this case:

1. Source node should never send known expired keys and we need to check whether is expired or not.
2. Target node should also skip expired keys instead of stoping the migration flow, since we can't guarantee the clock between source and target nodes are always the same and the network delay also may cause TTL = 0 problem.

ellutionist: What do you think of:
1.  Source node still check whether the key has expired and skip it if it has. 
2. Source node employs `SET key value EXAT unix-time-seconds` (not supported yet as far as I know, but not hard to implement) instead of `SET key value EX ttl` to send the exact expire ttl during the migration, in order to address the clock syncing or network latency issue.
3. Target node checks the expire unix time when receiving `EXAT` and skip the key when it is behind the current system clock (Does the response "OK" looks reasonable?). The target node cannot respond an error since the source node will abort the migration upon receiving any error reply.

git-hulk: Yes, exactly. I think it's ok to response ok when it's skipped.

xiaobiaozhao: LGTM

ellutionist: Just implemented "EXAT|PXAT" in the [new commit](https://github.com/apache/incubator-kvrocks/pull/901/commits/3a541e57ddbb8389095a6a691e1a234957bfe655). The implementation is trying to keep the change to the minimal and keep the bahavior consistent with the original redis (deleting the key if it should be expired).

git-hulk: cool, thanks.

src/redis_cmd.cc
```diff
@@ -486,6 +486,36 @@ class CommandSet : public Commander {
...
+        try {
+          std::string s = args_[++i];
+          std::string::size_type sz;
+          expire_ = std::stol(s, &sz);
+          if (sz != s.size()) {
+            return Status(Status::RedisParseErr, errValueNotInterger);
+          }
+        } catch (std::exception &e) {
+          return Status(Status::RedisParseErr, errValueNotInterger);
+        }
```
xiaobiaozhao:
```suggestion
      } else if (opt == "exat" && !ttl_ && !expire_ && !last_arg) {
          std::string s = args_[++i];
          std::string::size_type sz;
          
          if (sz != s.size()) {
            return Status(Status::RedisParseErr, errValueNotInterger);
          }
          try {
            expire_ = std::stol(s, &sz);
          } catch (std::exception &e) {
          return Status(Status::RedisParseErr, errValueNotInterger);
          }
```

src/redis_cmd.cc
```diff
@@ -486,6 +486,36 @@ class CommandSet : public Commander {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
         if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "exat" && !ttl_ && !expire_ && !last_arg) {
+        try {
+          std::string s = args_[++i];
+          std::string::size_type sz;
+          expire_ = std::stol(s, &sz);
```
tanruixiang: Better to use parseInt<>

src/redis_cmd.cc
```diff
@@ -486,6 +486,36 @@ class CommandSet : public Commander {
...
+        } catch (std::exception &e) {
+          return Status(Status::RedisParseErr, errValueNotInterger);
+        }
+        if (expire_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "pxat" && !ttl_ && !expire_ && !last_arg) {
+        uint64_t expire_ms = 0;
+        try {
+          std::string s = args_[++i];
+          std::string::size_type sz;
+          expire_ms = std::stoul(s, &sz);
```
tanruixiang: Better to use parseInt<>

src/redis_cmd.cc
```diff
@@ -475,17 +475,34 @@ class CommandSet : public Commander {
...
-        try {
-          std::string s = args_[++i];
-          std::string::size_type sz;
-          ttl_ = std::stoi(s, &sz);
-          if (sz != s.size()) {
-            return Status(Status::RedisParseErr, errValueNotInterger);
-          }
-        } catch (std::exception &e) {
+        std::string s = args_[++i];
+        StatusOr<int> parse_status = ParseInt(s);
```
```suggestion
        auto parse_status = ParseInt<int>(s);
```

src/redis_cmd.cc
```diff
@@ -475,17 +475,34 @@ class CommandSet : public Commander {
...
+        if (!parse_status.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
+        ttl_ = parse_status.GetValue();
         if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "exat" && !ttl_ && !expire_ && !last_arg) {
+        std::string s = args_[++i];
+        StatusOr<int64_t> parse_status = ParseInt(s);
```
```suggestion
        auto parse_status = ParseInt<int64_t>(s);
```

src/redis_cmd.cc
```diff
@@ -475,17 +475,34 @@ class CommandSet : public Commander {
...
+        if (!parse_status.IsOK()) {
+          return Status(Status::RedisParseErr, errValueNotInterger);
+        }
+        expire_ = parse_status.GetValue();
+        if (expire_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "pxat" && !ttl_ && !expire_ && !last_arg) {
+        std::string s = args_[++i];
+        StatusOr<uint64_t> parse_status = ParseInt(s);
```
```suggestion
        auto parse_status = ParseInt<uint64_t>(s);
```

ellutionist: Thanks for correction!

git-hulk:
@ellutionist Can you help to resolve the conflict in redis_cmd.cc

ellutionist:
> @ellutionist Can you help to resolve the conflict in redis_cmd.cc

@git-hulk Conflict resolved.

git-hulk (dismissed):

src/redis_cmd.cc
```diff
@@ -477,17 +477,34 @@ class CommandSet : public Commander {
...
-          std::string s = args_[++i];
-          std::string::size_type sz;
-          ttl_ = std::stoi(s, &sz);
-          if (sz != s.size()) {
-            return Status(Status::RedisParseErr, errValueNotInterger);
-          }
-        } catch (std::exception &e) {
+        std::string s = args_[++i];
+        auto parse_status = ParseInt<int>(s);
+        if (!parse_status.IsOK()) {
```
tanruixiang:
```suggestion
        if (!parse_status) {
```

src/redis_cmd.cc
```diff
@@ -477,17 +477,34 @@ class CommandSet : public Commander {
...
         }
+        ttl_ = parse_status.GetValue();
         if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "exat" && !ttl_ && !expire_ && !last_arg) {
+        std::string s = args_[++i];
+        auto parse_status = ParseInt<int64_t>(s);
+        if (!parse_status.IsOK()) {
+          return Status(Status::RedisParseErr, errValueNotInterger);
+        }
+        expire_ = parse_status.GetValue();
```
tanruixiang:
```suggestion
        expire_ = *parse_status;
```

src/redis_cmd.cc
```diff
@@ -477,17 +477,34 @@ class CommandSet : public Commander {
...
-          if (sz != s.size()) {
-            return Status(Status::RedisParseErr, errValueNotInterger);
-          }
-        } catch (std::exception &e) {
+        std::string s = args_[++i];
+        auto parse_status = ParseInt<int>(s);
+        if (!parse_status.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
+        ttl_ = parse_status.GetValue();
```
tanruixiang:
```suggestion
        ttl_ = *parse_status;
```

src/redis_cmd.cc
```diff
@@ -477,17 +477,34 @@ class CommandSet : public Commander {
...
         if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "exat" && !ttl_ && !expire_ && !last_arg) {
+        std::string s = args_[++i];
+        auto parse_status = ParseInt<int64_t>(s);
+        if (!parse_status.IsOK()) {
```
tanruixiang:
```suggestion
        if (!parse_status) {
```

src/redis_cmd.cc
```diff
@@ -477,17 +477,34 @@ class CommandSet : public Commander {
...
+          return Status(Status::RedisParseErr, errValueNotInterger);
+        }
+        expire_ = parse_status.GetValue();
+        if (expire_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "pxat" && !ttl_ && !expire_ && !last_arg) {
+        std::string s = args_[++i];
+        auto parse_status = ParseInt<uint64_t>(s);
+        if (!parse_status.IsOK()) {
```
tanruixiang:
```suggestion
        if (!parse_status) {
```

src/redis_cmd.cc
```diff
@@ -477,17 +477,34 @@ class CommandSet : public Commander {
...
+          return Status(Status::RedisParseErr, errValueNotInterger);
+        }
+        uint64_t expire_ms = parse_status.GetValue();
```
tanruixiang:
```suggestion
        uint64_t expire_ms = *parse_status;
```

src/redis_cmd.cc
```diff
@@ -477,17 +477,34 @@ class CommandSet : public Commander {
...
+        std::string s = args_[++i];
+        auto parse_status = ParseInt<int>(s);
+        if (!parse_status.IsOK()) {
           return Status(Status::RedisParseErr, errValueNotInterger);
         }
+        ttl_ = parse_status.GetValue();
         if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "exat" && !ttl_ && !expire_ && !last_arg) {
+        std::string s = args_[++i];
+        auto parse_status = ParseInt<int64_t>(s);
```
tanruixiang:
```suggestion
        auto parse_status = ParseInt<int64_t>(args_[++i]);
```

src/redis_cmd.cc
```diff
@@ -477,17 +477,34 @@ class CommandSet : public Commander {
...
+        if (!parse_status.IsOK()) {
+          return Status(Status::RedisParseErr, errValueNotInterger);
+        }
+        expire_ = parse_status.GetValue();
+        if (expire_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+      } else if (opt == "pxat" && !ttl_ && !expire_ && !last_arg) {
+        std::string s = args_[++i];
+        auto parse_status = ParseInt<uint64_t>(s);
```
tanruixiang:
```suggestion
        auto parse_status = ParseInt<int64_t>(args_[++i]);
```

tanruixiang (changes requested):
HI. Thank you very much for your contribution. The code could be a little more concise.

xiaobiaozhao (approved):

tanruixiang:
@ellutionist Thank you very much for your contribution. LGTM. Do you have time to add a unit test for cpp? It should be in `t_string_test.cc`.

ellutionist:
> @ellutionist Thank you very much for your contribution. LGTM. Do you have time to add a unit test for cpp? It should be in `t_string_test.cc`.

@tanruixiang Since this PR does not bring any new interface like `SetEX` or `MSet` to the class `Redis::String`, I cannot see any proper entry point for cpp unit test. Any good idea?

tanruixiang:
> @tanruixiang Since this PR does not bring any new interface like `SetEX` or `MSet` to the class `Redis::String`, I cannot see any proper entry point for cpp unit test. Any good idea?

Sorry, I revisited the code. You're right. I guess tcl's tests are enough. Thank you very much for your contribution.

tanruixiang (approved):

git-hulk (approved):

git-hulk:
Many thanks for your contribution, will merge if there are no more comments by tomorrow. cc @PragmaTwice @caipengbo

Issue #886, opened 2022-09-18
rework compaction filter when used blobs


tufitko:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

We have quite big messages (~100Mib), so we decided to use BlobDB to decrease pressure on disk. But this action decreases write throughput and increase read iops because RocksDB reads data from blobfiles.
Current SubKeyFilter gets values of keys only to check bitmap emptiness. I think it's too wasteful. Every compaction is painful. 
It's possible to implement `FilterBlobByKey`?  

### Solution

I've made implementation of `FilterBlobByKey`. We don't use bitmaps, so it works for us. This almost completely removes the reads of blobs
What do you think about it?
https://github.com/apache/incubator-kvrocks/commit/299f45a680970c84ac75759848b6fb1b9f4dd6fa

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

git-hulk:
Cool, @ShooterIT @caipengbo @shangxiaoxiong Can help to have a look if you're free?

caipengbo:
Yeah, I will CR as soon as possible

caipengbo:
@tufitko This is a good optimization point, in most cases we can make a decision about a key-value solely based on the key in compaction filter. 

Just as you said we can introduce new `FilterBlobByKey` interface to optimize it:
- For the bitmap type, we can return `kUndetermined` in `FilterBlobByKey` to read the value further. 
- For the other types, we just need to read the key through `FilterBlobByKey`.

tufitko:
@caipengbo so, I can create PR based on my commit? Should it be optional or nothing to worry about bitmap?

caipengbo:
@tufitko Good job! I don't think there's anything to worry about with bitmap, just do it!

PR #902, opened 2022-09-20
feat: impl FilterBlobByKey for SubKeyFilter

tufitko:
#886 
implement FilterBlobByKey to avoid blobs reading in compactions when using BlobDB

### DIFF
--- a/src/compact_filter.cc
+++ b/src/compact_filter.cc
@@ -140,9 +140,8 @@ bool SubKeyFilter::Filter(int level,
     return false;
   }
 
-  bool result = IsMetadataExpired(ikey, metadata) ||
+  return IsMetadataExpired(ikey, metadata) ||
     (metadata.Type() == kRedisBitmap && Redis::Bitmap::IsEmptySegment(value));
-  return result;
 }
 
 }  // namespace Engine
--- a/src/compact_filter.cc
+++ b/src/compact_filter.cc
@@ -50,14 +50,13 @@ bool MetadataFilter::Filter(int level,
   return metadata.Expired();
 }
 
-bool SubKeyFilter::IsKeyExpired(const InternalKey &ikey, const Slice &value) const {
+Status SubKeyFilter::GetMetadata(const InternalKey &ikey, Metadata* metadata) const {
   std::string metadata_key;
 
   auto db = stor_->GetDB();
   const auto cf_handles = stor_->GetCFHandles();
   // storage close the would delete the column family handler and DB
-  if (!db || cf_handles->size() < 2)  return false;
-
+  if (!db || cf_handles->size() < 2)  return Status(Status::NotOK, "storage is closed");
   ComposeNamespaceKey(ikey.GetNamespace(), ikey.GetKey(), &metadata_key, stor_->IsSlotIdEncoded());
 
   if (cached_key_.empty() || metadata_key != cached_key_) {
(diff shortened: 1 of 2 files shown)


src/compact_filter.cc
```diff
@@ -108,7 +110,12 @@ bool SubKeyFilter::Filter(int level,
                                   std::string *new_value,
                                   bool *modified) const {
   InternalKey ikey(key, stor_->IsSlotIdEncoded());
-  bool result = IsKeyExpired(ikey, value);
+  Metadata metadata(kRedisNone, false);
+  if (GetMetadata(ikey, &metadata)) {
+    return true;
+  }
```
caipengbo: I don't think the name `GetMetadata` is clear, so I'm confused here.

src/compact_filter.cc
```diff
@@ -117,4 +124,27 @@ bool SubKeyFilter::Filter(int level,
...
+    return rocksdb::CompactionFilter::Decision::kUndetermined;
+  }
+
+  bool result = IsMetadataExpired(ikey, metadata);
+  DLOG(INFO) << "[compact_filter/subkey] "
+            << "namespace: " << ikey.GetNamespace().ToString()
+            << ", metadata key: " << ikey.GetKey().ToString()
+            << ", subkey: " << ikey.GetSubKey().ToString()
+            << ", verison: " << ikey.GetVersion()
+            << ", result: " << (result ? "deleted" : "reserved");
```
caipengbo: I think this log content should be distinguished with the log in Filter (line 119), although the location of the log can be determined by the line of the file and even though it is a Debug log.

src/compact_filter.cc
tufitko: I didn't invent a better name. `FindMetadata`, `FillMetadata`?

src/compact_filter.cc
tufitko: `[compact_filter/subkey_blob]` is it ok?

git-hulk: We can remove all debug logs in compaction filter since it will affect the compaction performance even it's disabled in production.

caipengbo: Agree with it.

src/compact_filter.cc
caipengbo: Let me think about it. I still don't think the logic here is clear enough.

tufitko: this function is looking for metadata and fills it, also returned true if kv can be deleted.

src/compact_filter.cc
tufitko: okey

src/compact_filter.cc
caipengbo: So it does more than one thing, and we should separate those things.

tufitko: Hm, I don't know how I can split it. 
there are 4 actions:
- get_db
- build_metadata_key
- find_metdata
- decode_metadata

 And all of the actions need to get metadata..

src/compact_filter.h
```diff
@@ -64,9 +64,13 @@ class SubKeyFilter : public rocksdb::CompactionFilter {
...
 
   const char *Name() const override { return "SubkeyFilter"; }
-  bool IsKeyExpired(const InternalKey &ikey, const Slice &value) const;
+  bool GetMetadata(const InternalKey &ikey, Metadata* metadata) const;
+  bool IsMetadataExpired(const InternalKey &ikey, const Metadata& metadata) const;
   bool Filter(int level, const Slice &key, const Slice &value,
               std::string *new_value, bool *modified) const override;
+  rocksdb::CompactionFilter::Decision FilterBlobByKey(int /*level*/, const Slice& /*key*/,
+                                   std::string* /*new_value*/,
+                                   std::string* /*skip_until*/) const override;
```
tanruixiang:
```suggestion
  rocksdb::CompactionFilter::Decision FilterBlobByKey(int level, const Slice &key,
                                   std::string *new_value,
                                   std::string *skip_until) const override;
```

I think we should unify the style here.

tanruixiang (changes requested):

src/compact_filter.cc
caipengbo: I thought about it a little more, `GetMetadata` is OK.

git-hulk: @tufitko @caipengbo I think another idea is `GetMetadata` returns Status instead of bool, so that caller determine whether to keep the subkey/blob or not. Not sure if I clarify clearly enough.

git-hulk: The pseudo-code should be like below:

```
Filter() {

  Status s = GetMetadata(ikey, &metadata);
  if (s.IsNotFound()) {
      return true;
  } 
  if (!s.IsOK()) {
     // log error
     return false;
  } 
  ... check is expired or other logic here
}
```

src/compact_filter.h
caipengbo: `FilterBlobByKey()` will be executed before `Filter()`, so I think `FilterBlobByKey()` should be declared and defined before `Filter()`.

src/compact_filter.cc
caipengbo: > @tufitko @caipengbo I think another idea is GetMetadata returns Status instead of bool, so that caller determine whether to keep the subkey/blob or not. Not sure if I clarify clearly enough.

Yeah, that makes it clearer!

tufitko: Oh, okey, it’s good, I will be commited later

src/compact_filter.h
tufitko: Ok, it’s no problem

tufitko: Good, thanks

caipengbo: Yeah, it's not a big deal, but it's more logical :)

src/compact_filter.cc
git-hulk: ❤️ Thanks

src/compact_filter.cc
```diff
@@ -108,13 +127,22 @@ bool SubKeyFilter::Filter(int level,
...
+    LOG(ERROR) << "[compact_filter/subkey] Failed to get metadata"
+            << ", namespace: " << ikey.GetNamespace().ToString()
+            << ", key: " << ikey.GetKey().ToString()
+            << ", err: " << s.Msg();
+    return false;
+  }
+
+  bool result = IsMetadataExpired(ikey, metadata) ||
+    (metadata.Type() == kRedisBitmap && Redis::Bitmap::IsEmptySegment(value));
   return result;
```
git-hulk:
```suggestion
  return IsMetadataExpired(ikey, metadata) ||
    (metadata.Type() == kRedisBitmap && Redis::Bitmap::IsEmptySegment(value));
```

git-hulk (approved):

tanruixiang (approved):
LGTM. thanks for your contribution

caipengbo (approved):

PR #906, opened 2022-09-22
Avoid using Get interface to obtain metadata when iterating data during migration

### LINKED ISSUE #904: Avoid using Get interface to obtain Metadata when iterating data during migration
### Motivation

When migrating data, I found that when migrating the same number of keys (about 10W), the migration time was very variable, sometimes very short, sometimes very long. When the migration took a long time, I found that the CPU was full. 

Looking at the flame diagram below:

image

we can see that there is a **deep call stack** on the left. 

Zoomed in it:

image

When we parsed and sent the full data (`SlotMigrate::SendSnapshot()`), we used the `rocksdb::DB::Get()` interface to get the metadata, which was CPU consuming.

### How much does this affect migration speed?

1. Because we set the iterator to not fill the cache, when the amount of data is larger than cache, it is very likely that we will call `rocksdb::DB::Get()` to read disk, which greatly affects the performance!
2. If the retrieval in memtable is bad( it depends on workload), then it will consume more CPU, which will bring more serious impact.

### Solution

In fact, we do not need to get the value through the `rocksdb::DB::Get()` interface. When we iterate data, value and key are together, so we can directly get the value through the iterator and then encode into metadata.

caipengbo:
Fixed #904

### DIFF
--- a/src/slot_migrate.cc
+++ b/src/slot_migrate.cc
@@ -546,7 +546,7 @@ bool SlotMigrate::CheckResponseWithCounts(int sock_fd, int total) {
 Status SlotMigrate::MigrateOneKey(const rocksdb::Slice &key, const rocksdb::Slice &value, std::string *restore_cmds) {
   std::string prefix_key;
   AppendNamespacePrefix(key, &prefix_key);
-  std::string bytes = value.ToString();
+  std::string bytes = std::move(value.ToString());
   Metadata metadata(kRedisNone, false);
   metadata.Decode(bytes);
   if (metadata.Type() != kRedisString && metadata.size == 0) {
--- a/src/slot_migrate.cc
+++ b/src/slot_migrate.cc
@@ -302,7 +302,7 @@ Status SlotMigrate::SendSnapshot(void) {
 
     // Add key's constructed cmd to restore_cmds, send pipeline
     // or not according to current_pipeline_size_
-    auto stat = MigrateOneKey(rocksdb::Slice(user_key), iter->value(), &restore_cmds);
+    auto stat = MigrateOneKey(user_key, iter->value(), &restore_cmds);
     if (stat.IsOK()) {
       if (stat.Msg() == "ok") migratedkey_cnt++;
       if (stat.Msg() == "expired") expiredkey_cnt++;
--- a/src/slot_migrate.cc
+++ b/src/slot_migrate.cc
@@ -546,7 +546,7 @@ bool SlotMigrate::CheckResponseWithCounts(int sock_fd, int total) {
 Status SlotMigrate::MigrateOneKey(const rocksdb::Slice &key, const rocksdb::Slice &value, std::string *restore_cmds) {
   std::string prefix_key;
   AppendNamespacePrefix(key, &prefix_key);
-  std::string bytes = std::move(value.ToString());
+  std::string bytes = value.ToString();
   Metadata metadata(kRedisNone, false);
   metadata.Decode(bytes);
   if (metadata.Type() != kRedisString && metadata.size == 0) {
--- a/src/slot_migrate.cc
+++ b/src/slot_migrate.cc
@@ -302,7 +302,7 @@ Status SlotMigrate::SendSnapshot(void) {
 
     // Add key's constructed cmd to restore_cmds, send pipeline
     // or not according to current_pipeline_size_
-    auto stat = MigrateOneKey(rocksdb::Slice(user_key), &restore_cmds);
+    auto stat = MigrateOneKey(rocksdb::Slice(user_key), iter->value(), &restore_cmds);
     if (stat.IsOK()) {
       if (stat.Msg() == "ok") migratedkey_cnt++;
       if (stat.Msg() == "expired") expiredkey_cnt++;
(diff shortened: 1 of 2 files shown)


src/slot_migrate.cc
```diff
@@ -302,7 +302,7 @@ Status SlotMigrate::SendSnapshot(void) {
 
     // Add key's constructed cmd to restore_cmds, send pipeline
     // or not according to current_pipeline_size_
-    auto stat = MigrateOneKey(rocksdb::Slice(user_key), &restore_cmds);
+    auto stat = MigrateOneKey(rocksdb::Slice(user_key), iter->value(), &restore_cmds);
```
git-hulk:
```suggestion
    auto stat = MigrateOneKey(user_key, iter->value(), &restore_cmds);
```

git-hulk (dismissed):

datavisorxiaobiaozhao:
The problem we had was that when we migrated, we had a lot of disk reads, but with this optimized version, these abnormal disk reads disappeared, so I think GetMetaData calls the Get interface

caipengbo:
> The problem we had was that when we migrated, we had a lot of disk reads, but with this optimized version, these abnormal disk reads disappeared, so I think GetMetaData calls the Get interface.

@datavisorxiaobiaozhao Yes. During the iteration, we set the cache not to be filled:

https://github.com/apache/incubator-kvrocks/blob/e8721a270d23345f385af6db0617c3549d6e6812/src/slot_migrate.cc#L273-L277

so that when we call `DB::Get()`, we are likely to read the disk.

src/slot_migrate.cc
```diff
@@ -543,29 +543,15 @@ bool SlotMigrate::CheckResponseWithCounts(int sock_fd, int total) {
...
-}
-
-Status SlotMigrate::MigrateOneKey(rocksdb::Slice key, std::string *restore_cmds) {
+Status SlotMigrate::MigrateOneKey(const rocksdb::Slice &key, const rocksdb::Slice &value, std::string *restore_cmds) {
   std::string prefix_key;
   AppendNamespacePrefix(key, &prefix_key);
-  std::string bytes;
-  bool st = GetSlotKeyMetadata(prefix_key, &bytes);
-  if (!st) return Status(Status::NotOK, "[migrate] Failed to get key's metadata");
+  std::string bytes = value.ToString();
```
xiaobiaozhao, line 549: Optimization big value
```suggestion
  std::string bytes = std::move(value.ToString());
```

PragmaTwice: This `std::move` is useless, and it will block the (N)RVO ([return value optimization](https://en.cppreference.com/w/cpp/language/copy_elision)).

explainations:

###  1. this `std::move` is useless
Signature of `ToString`:
```
std::string Slice::ToString(...);
```
So the return value `x.ToString()` is already an rvalue (or concretely, a [prvalue](https://en.cppreference.com/w/cpp/language/value_category#prvalue)).
Move an rvalue is useless, since it will essentially dispatch to the rvalue reference overload.

### 2. it will block the (N)RVO
RVO and NRVO are important optimization approaches in C++, and are guaranteed in new standards.

So, in `std::string x = v.ToString()`, the local result variable of `ToString` will be eliminated and it just construct one `std::string x`, no copy contruction, and even no move construction.

If you move it, then RVO will just not work here.

So, in conclusion, this change will only downgrade the performance.

caipengbo: That's great. That's something I don't know about. I'll dig into it. thank you @PragmaTwice

git-hulk (approved):

src/slot_migrate.cc
xiaobiaozhao: That's great!!!

shangxiaoxiong:
There may be some problem. The Get interface may aim to get the latest expire time of the key.

caipengbo:
> There may be some problem. The Get interface may aim to get the latest expire time of the key.

@shangxiaoxiong Oh Yes, I missed that. Then we only need GetMetaData for those with expiration time. Or we can record the time snapshot was created and subtract the expiration time to determine whether it is expired!

shangxiaoxiong:
During sending snapshot, exipre time of some key may be changed to larger by clients. GetMetaData is needed to deal with this case.

shangxiaoxiong:
l think GetMetaData is needed for complex key with expire time to get the latest expire time.

caipengbo:
> During sending snapshot, exipre time of some key may be changed to larger by clients. GetMetaData is needed to deal with this case.

During sending snapshot:
- for simple type, user changes to expire are incremental data and will be synchronized through WAL.
- for complex types, we really should get the latest metadata.

shangxiaoxiong:
> > During sending snapshot, exipre time of some key may be changed to larger by clients. GetMetaData is needed to deal with this case.
> 
> During sending snapshot, user changes to expire are incremental data and will be synchronized through WAL.

Simple key is ok but complex key is not.

git-hulk:
> > > During sending snapshot, exipre time of some key may be changed to larger by clients. GetMetaData is needed to deal with this case.
> > 
> > 
> > During sending snapshot, user changes to expire are incremental data and will be synchronized through WAL.
> 
> Simple key is ok but complex key is not.

It can't resolve this issue even we use the latest snapshot since the change may come after.

shangxiaoxiong:
Yeah. Using GetMetaData with latest snapshot can't solve the problem completely. Just a little improvement.

caipengbo:
Indeed, the whole process is not atomic.

git-hulk:
Yes, the root cause is the snapshot can be different between writers and background threads. I prefer to use current implementation, then we can try to seek another way to resolve this issue completely. How do you think? @caipengbo @shangxiaoxiong

shangxiaoxiong:
Current implementation makes slotmigration faster.  Faster slotmigration is, smaller risk of losing key is.  On the whole, current implementation may be good.

caipengbo:
I think OK, I will continue to think about how to solve this kind of problem.

xiaobiaozhao:
I think this is the best solution for now, merge this pr first, and then contionue to look for new solutions.

xiaobiaozhao (approved):

shangxiaoxiong (approved):

git-hulk (approved):

caipengbo:
Thanks all for the great discussion, I'll explain the problems with this implementation in #904

PragmaTwice (approved):

PR #911, opened 2022-09-22
Disable the TLS by default

### LINKED ISSUE #907: Disable the TLS by default when building
### Search before asking

### Motivation

For most users should NOT need TLS version,  so I think we can disable it by default like `Redis` and guide users to enable it if they need this feature.

### Solution

Change `ENABLE_OPENSSL` from `ON` to `OFF`

git-hulk:
This closes #907

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -72,24 +72,29 @@ jobs:
           - name: Darwin Clang
             os: macos-11
             compiler: auto
+            with_openssl: -DENABLE_OPENSSL=ON
           - name: Darwin Clang without Jemalloc
             os: macos-11
             compiler: auto
             without_jemalloc: -DDISABLE_JEMALLOC=ON
+            with_openssl: -DENABLE_OPENSSL=ON
           - name: Ubuntu GCC
             os: ubuntu-18.04
             compiler: gcc
           - name: Ubuntu Clang
             os: ubuntu-18.04
             compiler: clang
+            with_openssl: -DENABLE_OPENSSL=ON
           - name: Ubuntu GCC ASan
             os: ubuntu-18.04
             without_jemalloc: -DDISABLE_JEMALLOC=ON
             with_sanitizer: -DENABLE_ASAN=ON
+            with_openssl: -DENABLE_OPENSSL=ON
             compiler: gcc
           - name: Ubuntu GCC TSan
             os: ubuntu-18.04
             without_jemalloc: -DDISABLE_JEMALLOC=ON
+            with_openssl: -DENABLE_OPENSSL=ON
             with_sanitizer: -DENABLE_TSAN=ON
             compiler: gcc
             runtime_env_vars: TSAN_OPTIONS="suppressions=$(realpath ./tests/tsan-suppressions)"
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -93,6 +93,13 @@ $ ./x.py build # `./x.py build -h` to check more options;
                # especially, `./x.py build --ghproxy` will fetch dependencies via ghproxy.com.
 ```
 
+To build with TLS support, you'll need OpenSSL development libraries (e.g. libssl-dev on Debian/Ubuntu) and run:
```
PragmaTwice, line 96: Should we remove openssl packages like `libssl-dev` in Prerequisite above?

git-hulk: Does it sound good to preserve the OpenSSL in prerequisite? I think it's ok since it's a common library and harms nothing.

PragmaTwice:
Do we need to enable TLS in docker images? Currently TLS is enabled in nightly dockerhub images.

git-hulk:
> Do we need to enable TLS in docker images? Currently TLS is enabled in nightly dockerhub images.

Yes, we need to enable TLS in docker image, let me fix it later.

.github/workflows/kvrocks.yaml
```diff
@@ -206,7 +206,7 @@ jobs:
         run: |
           export ${{ matrix.runtime_env_vars }}
           ./x.py test tcl build --cli-path $HOME/local/bin/redis-cli --dont-clean
-          if [[ ! -z "${{ matrix.with_openssl }}" ]] && [[ "${{ matrix.os }}" == ubuntu* ]]; then
+          if [[ ! -z ${{ matrix.with_openssl }} ]] && [[ "${{ matrix.os }}" == ubuntu* ]]; then
```
PragmaTwice: I think if `matrix.with_openssl` does not exist, it will be replaced to `[[ ! -z ]]` rather than `[[ ! -z "" ]]`, hence maybe it will fail due to syntax error.

PragmaTwice: BTW  `[[ -n "something" ]]` is another approach to check if it is not empty : )

git-hulk: Thanks @PragmaTwice

PragmaTwice (approved):

.github/workflows/kvrocks.yaml
```diff
@@ -164,10 +174,8 @@ jobs:
 
       - name: Build Kvrocks
         run: |
-          ./x.py build -j$NPROC --unittest --compiler \
-            ${{ matrix.compiler }} ${{ matrix.with_ninja }} ${{ matrix.with_sanitizer }} \
-            ${{ matrix.without_jemalloc }} ${{ matrix.without_openssl }} ${{ matrix.without_luajit }} \
-            ${{ env.CMAKE_EXTRA_DEFS }}
+          ./x.py build -j$NPROC --unittest --compiler ${{ matrix.compiler }} ${{ matrix.without_jemalloc }} ${{ matrix.without_luajit }} \
```
git-hulk, line 177: it seems that some parameters will be lost if there has an extra slash. For example, if `${{ env.CMAKE_EXTRA_DEFS }}` was empty and will become:
```shell
./x.py build -j8 --unittest --compiler \
  gcc --ninja \
  -DENABLE_OPENSSL=ON \
```
then it causes the TLS server to start failed. cc @PragmaTwice

PragmaTwice: weird :rofl:

git-hulk: 😆 I tried to reorder parameters to work around, but didn't get what's the root cause. Really hope my conclusion is wrong.

PragmaTwice: If it still not work well, I think we can write it in one line, no slash.

git-hulk: ok~

PR #919, opened 2022-09-25
Fix many typos in cluster.cc

### LINKED ISSUE #873: Unit test has wrong sleep time
https://github.com/apache/incubator-kvrocks/blob/unstable/tests/cppunit/rwlock_test.cc#L170

This unit test sleep time is NOT right, expected 100ms but current is 1ms. I think 1ms is too short to guarantee the else condition thread will be always be scheduled before the if condition. I'll bring this sleep time fix in the next PR.

_Originally posted by @git-hulk in https://github.com/apache/incubator-kvrocks/issues/870#issuecomment-1244844809_

git-hulk:
1. Fix typos in the cluster.cc
2. Use const string when appearing many times
3. Fix incorrect sleep time in rwlock test case

Close #873

### DIFF
--- a/tests/cppunit/rwlock_test.cc
+++ b/tests/cppunit/rwlock_test.cc
@@ -167,7 +167,7 @@ TEST(ReadWriteLock, WriteLockGurad_First) {
       });
     } else {
       ths[i] = std::thread([&rwlock, &val]() {
-        usleep(1000);  // To avoid it is the first to run, just sleep 100ms
+        usleep(100000);  // To avoid it is the first to run, just sleep 100ms
         auto ptr = std::unique_ptr<RWLock::ReadLock>(new RWLock::ReadLock(rwlock));
         ASSERT_EQ(val, 3);
       });
--- a/src/cluster.cc
+++ b/src/cluster.cc
@@ -29,7 +29,7 @@
 #include "redis_cmd.h"
 #include "replication.h"
 
-const char *errInvalidNodeID = "Invalid node id";
+const char *errInvalidNodeID = "Invalid cluster node id";
 const char *errInvalidSlotID = "Invalid slot id";
 const char *errSlotOutOfRange = "Slot is out of range";
 const char *errInvalidClusterVersion = "Invalid cluster version";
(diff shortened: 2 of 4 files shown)


tests/cppunit/rwlock_test.cc
```diff
@@ -167,7 +167,7 @@ TEST(ReadWriteLock, WriteLockGurad_First) {
       });
     } else {
       ths[i] = std::thread([&rwlock, &val]() {
-        usleep(1000);  // To avoid it is the first to run, just sleep 100ms
+        usleep(100000);  // To avoid it is the first to run, just sleep 100ms
```
PragmaTwice, line 170:
```suggestion
        using namespace std::chrono_literals;
        std::this_thread::sleep_for(100ms);  // To avoid it is the first to run, just sleep 100ms
```

We can use the time literal `100ms` (defined in chrono part of std library) directly, haha : )

caipengbo (dismissed):

caipengbo (dismissed):

PragmaTwice:
I notice that CI is down in darwin clang. Maybe `#include <chrono>` is needed to use namespace `std::chrono_literals`.

git-hulk:
> I notice that CI is down in darwin clang. Maybe `#include <chrono>` is needed to use namespace `std::chrono_literals`.

Yes, should include `<chrono>`, fixed.

git-hulk:
hi @PragmaTwice, seems the namespace of `chrono_literals` is a bit different between compilers, do you suggest rollbacking to use the previous way?

PragmaTwice:
> hi @PragmaTwice, seems the namespace of `chrono_literals` is a bit different between compilers, do you suggest rollbacking to use the previous way?

Sure. Maybe it need c++14 :rofl:

git-hulk:
> > hi @PragmaTwice, seems the namespace of `chrono_literals` is a bit different between compilers, do you suggest rollbacking to use the previous way?
> 
> Sure. Maybe it need c++14 🤣

I was a bit confuse when looking its docs, it says `std::literals::chrono_literals` required at least C++14, but std::chrono_literals can be found in C++11's chrono. Let's use the legacy way first.

caipengbo (approved):

PragmaTwice (approved):

git-hulk:
Thanks all, merging...

PR #923, opened 2022-09-26
Return from HRANGE if key wasn't found

torwig:
Small fix to the implementation of `HRANGE` command for the case if a key wasn't found.


git-hulk (approved):
LGTM, thanks @torwig

PR #924, opened 2022-09-27
Use ParseInt to replace sto*, ato*

tanruixiang:
Use ParseInt to replace sto*, ato*.
The previous [PR](https://github.com/apache/incubator-kvrocks/pull/921) was closed due to misoperation, I am very sorry.


src/batch_extractor.cc
```diff
@@ -66,7 +67,8 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
     } else if (metadata.expire > 0) {
       auto args = log_data_.GetArguments();
       if (args->size() > 0) {
-        RedisCommand cmd = static_cast<RedisCommand >(std::stoi((*args)[0]));
+        auto parse_result = ParseInt<int>((*args)[0], 10);
+        RedisCommand cmd = static_cast<RedisCommand >(*parse_result);
```
PragmaTwice: If you do not check whether the `parse_result` is ok before getting its value, the program will abort when facing an error status in `ParseInt`.

tanruixiang: > If you do not check whether the `parse_result` is ok before getting its value, it will abort the program when facing an error status in `ParseInt`.

Hi. The previous code also didn't check the correctness of parsing. My understanding is that this is a certain successful parsing. Do you think we need to do extra checks?

tanruixiang:
@caipengbo @shangxiaoxiong  I'm very sorry that the last pr was closed because of my misoperation. I hope we can continue to review in this pr.

src/batch_extractor.cc
PragmaTwice: The behavior is not quite same as previous. While using  `ato*`, the program does not abort if facing some error input and do not check result. For `sto*`, it will just throw the exception to the upper layer.

tanruixiang: > The behavior is not quite same as previous. While using `ato*`, the program does not abort if facing some error input and do not check result. For `sto*`, it will just throw the exception to the upper layer.

Yes. But if the parsing is expected must to be successful, core dump may be a good choice. Adding defensive check is also a good solution. I'll add the check later.

PragmaTwice: > > The behavior is not quite same as previous. While using `ato*`, the program does not abort if facing some error input and do not check result. For `sto*`, it will just throw the exception to the upper layer.
> 
> Yes. But if the parsing is expected must to be successful, core dump may be a good choice. Adding defensive check is also a good solution. I'll add the check later.

If the parsing is really expected to be successful, you can add some comment to simply prove it (i.e. the condition "result is not ok" is always unreachable whenever the execution path varies). Then you can omit the status checking.

src/redis_cmd.cc
```diff
@@ -4495,15 +4497,11 @@ class CommandReplConf : public Commander {
...
-          throw std::overflow_error("listening-port out of range");
-        }
-        port_ = static_cast<uint32_t>(p);
-      } catch (const std::exception &e) {
-        return Status(Status::RedisParseErr, "listening-port should be number");
+      auto parse_result = ParseInt<uint32_t>(value, 10);
+      if (!parse_result) {
+          return Status(Status::RedisParseErr, "listening-port should be number");
       }
+      port_ = *parse_result;
```
git-hulk, line 4544: Need to check the port range here

src/cluster.cc
```diff
@@ -583,9 +587,14 @@ Status Cluster::ParseClusterNodes(const std::string &nodes_str, ClusterNodes *no
           }
         }
       } else if (ranges.size() == 2) {
-        start = std::atoi(ranges[0].c_str());
-        stop = std::atoi(ranges[1].c_str());
-        if (start >= stop || start < 0 || stop >= kClusterSlots) {
+        auto parse_start = ParseInt<int>(ranges[0].c_str(), valid_range, 10);
+        auto parse_stop = ParseInt<int>(ranges[1].c_str(), valid_range, 10);
```
PragmaTwice:
```suggestion
        auto parse_start = ParseInt<int>(ranges[0], valid_range, 10);
        auto parse_stop = ParseInt<int>(ranges[1], valid_range, 10);
```

src/cluster.cc
```diff
@@ -583,9 +587,14 @@ Status Cluster::ParseClusterNodes(const std::string &nodes_str, ClusterNodes *no
...
+        if (!parse_start || !parse_stop || *parse_start >= *parse_stop) {
+          return Status(Status::ClusterInvalidInfo, errSlotOutOfRange);
+        }
+        start = *parse_start;
+        stop = *parse_stop;
+        if (start >= stop) {
           return Status(Status::ClusterInvalidInfo, errSlotOutOfRange);
         }
```
PragmaTwice:
```suggestion
```

src/cluster.cc
```diff
@@ -530,10 +531,11 @@ Status Cluster::ParseClusterNodes(const std::string &nodes_str, ClusterNodes *no
     std::string host = fields[1];
 
     // 3) port
-    int port = std::atoi(fields[2].c_str());
-    if (port <= 0 || port >= (65535-kClusterPortIncr)) {
+    auto parse_result = ParseInt<int>(fields[2].c_str(), NumericRange<int64_t>{1, 65535 - kClusterPortIncr - 1}, 10);
```
PragmaTwice:
```suggestion
    auto parse_result = ParseInt<int>(fields[2], NumericRange<int64_t>{1, 65535 - kClusterPortIncr - 1}, 10);
```

src/cluster.cc
```diff
@@ -566,14 +568,16 @@ Status Cluster::ParseClusterNodes(const std::string &nodes_str, ClusterNodes *no
...
 
     // 6) slot info
+    auto valid_range = NumericRange<int>{0, kClusterSlots - 1};
     for (unsigned i = 5; i < fields.size(); i++) {
       int start, stop;
       std::vector<std::string> ranges = Util::Split(fields[i], "-");
       if (ranges.size() == 1) {
-        start = std::atoi(ranges[0].c_str());
-        if (IsValidSlot(start) == false) {
+        auto parse_result = ParseInt<int>(ranges[0].c_str(), valid_range, 10);
```
PragmaTwice:
```suggestion
        auto parse_result = ParseInt<int>(ranges[0], valid_range, 10);
```

src/compaction_checker.cc
```diff
@@ -76,10 +77,22 @@ void CompactionChecker::PickCompactionFiles(const std::string &cf_name) {
     if (file_creation_time > static_cast<uint64_t>(now-3600)) continue;
     for (const auto &property_iter : iter.second->user_collected_properties) {
       if (property_iter.first == "total_keys") {
-        total_keys = std::atoi(property_iter.second.data());
+        auto parse_result = ParseInt<int>(property_iter.second.data(), 10);
```
PragmaTwice:
```suggestion
        auto parse_result = ParseInt<int>(property_iter.second, 10);
```

src/compaction_checker.cc
```diff
@@ -76,10 +77,22 @@ void CompactionChecker::PickCompactionFiles(const std::string &cf_name) {
...
+        if (!parse_result) {
+          LOG(ERROR) << "[compaction checker] Parse total_keys error: "
+                    << parse_result.Msg();
+          continue;
+        }
+        total_keys = *parse_result;
       }
       if (property_iter.first == "deleted_keys") {
-        deleted_keys = std::atoi(property_iter.second.data());
+        auto parse_result = ParseInt<int>(property_iter.second.data(), 10);
```
PragmaTwice:
```suggestion
        auto parse_result = ParseInt<int>(property_iter.second, 10);
```

tanruixiang:
@git-hulk @PragmaTwice Thank you so much. They have been fixed.

git-hulk (dismissed):

git-hulk:
Thanks for @tanruixiang contribution. A little tip, We can click the `resolve conservation` if it's resolved, then reviewers can know there're resolved.

caipengbo (dismissed):

src/util.cc
```diff
@@ -603,12 +603,14 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value) {
           tokens.clear();
           return tokens;
         }
-        array_len = std::stoull(std::string(start+1, p));
+        // parse_result expects to must be parsed successfully here
+        auto parse_result = ParseInt<uint64_t>(std::string(start + 1, p), 10);
+        array_len = *parse_result;
```
```suggestion
        // parse_result expects to must be parsed successfully here
        array_len = *ParseInt<uint64_t>(std::string(start + 1, p), 10);
```

src/util.cc
```diff
@@ -617,20 +619,23 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value) {
           tokens.clear();
           return tokens;
         }
-        bulk_len = std::stoull(std::string(start+1, p));
+        // parse_result expects to must be parsed successfully here
+        auto parse_result = ParseInt<uint64_t>(std::string(start + 1, p), 10);
+        bulk_len = *parse_result;
```
```suggestion
        // parse_result expects to must be parsed successfully here
        bulk_len = *ParseInt<uint64_t>(std::string(start + 1, p), 10);
```

src/cluster.cc
```diff
@@ -530,10 +531,11 @@ Status Cluster::ParseClusterNodes(const std::string &nodes_str, ClusterNodes *no
     std::string host = fields[1];
 
     // 3) port
-    int port = std::atoi(fields[2].c_str());
-    if (port <= 0 || port >= (65535-kClusterPortIncr)) {
+    auto parse_result = ParseInt<int>(fields[2], NumericRange<int64_t>{1, 65535 - kClusterPortIncr - 1}, 10);
```
PragmaTwice:
```suggestion
    auto parse_result = ParseInt<int>(fields[2], NumericRange<int>{1, 65535 - kClusterPortIncr - 1}, 10);
```

src/redis_cmd.cc
```diff
@@ -779,11 +779,13 @@ class CommandCAS : public Commander {
       std::string opt = Util::ToLower(args[i]);
       if (opt == "ex") {
         if (last_arg) return Status(Status::NotOK, errWrongNumOfArguments);
-        ttl_ = atoi(args_[++i].c_str());
+        auto parse_result = ParseInt<int>(args_[++i].c_str(), 10);
```
PragmaTwice, line 784: It is expected to check the status, otherwise wrong user input will just terminate the kvrocks instance, which will cause serious security problems.

src/redis_cmd.cc
```diff
@@ -779,11 +779,13 @@ class CommandCAS : public Commander {
...
+        ttl_ = *parse_result;
         if (ttl_ <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
       } else if (opt == "px") {
         if (last_arg) return Status(Status::NotOK, errWrongNumOfArguments);
-        auto ttl_ms = atol(args[++i].c_str());
+        auto parse_result = ParseInt<int>(args[++i].c_str(), 10);
```
PragmaTwice, line 792: Same as above.

src/redis_cmd.cc
```diff
@@ -4800,7 +4806,8 @@ class CommandClusterX : public Commander {
 
     // CLUSTERX SETSLOT $SLOT_ID NODE $NODE_ID $VERSION
     if (subcommand_ == "setslot" && args_.size() == 6) {
-      slot_id_ = atoi(args_[2].c_str());
+      auto parse_id = ParseInt<int>(args[2].c_str(), 10);
```
PragmaTwice, line 4817: Same as above.

src/redis_cmd.cc
```diff
@@ -4810,7 +4817,8 @@ class CommandClusterX : public Commander {
       if (args_[4].size() != kClusterNodeIdLen) {
         return Status(Status::RedisParseErr, "Invalid node id");
       }
-      set_version_ = atoll(args_[5].c_str());
+      auto parse_version = ParseInt<uint64_t>(args[5].c_str(), 10);
```
PragmaTwice, line 4831: Same as above.

PragmaTwice (commented):
The rest looks good to me!

src/redis_cmd.cc
tanruixiang: I'm sorry to have missed it. It seems that the previous code did not crash because of the use of `ato*`.

PragmaTwice (dismissed):

caipengbo:
LGTM

tisonkun:
@PragmaTwice @tanruixiang may I ask why `ParseInt` is better than `stoi` and `atoi`? I can see we write more boilerplate code.

tanruixiang:
> 



> @PragmaTwice @tanruixiang may I ask why `ParseInt` is better than `stoi` and `atoi`? I can see we write more boilerplate code.

My opinion is that the string to number process can be managed uniformly. What the `ParseInt` does is essentially the Wrapper of `std::strt*`. In fact, if you use `std::strt*` directly, you also need boilerplate code (at least try catch), but during the replacement process, we found that try catch was not done in some places, so it was supplemented.

PragmaTwice:
> @PragmaTwice @tanruixiang may I ask why `ParseInt` is better than `stoi` and `atoi`? I can see we write more boilerplate code.

We currently prefer to use `Status`/`StatusOr` for error handling for several reasons:
- There are some misuses of exceptions in the current code: for example, a lot of `try...catch` statements catch `std::exception` directly, but apparently this catches additional exceptions, such as `std::bad_alloc`. Also, some uses of functions such as `atoi`/`stoi` do not take into account parsing failures at all, which have been fixed in this PR, so the amount of code has increased.
- `StatusOr` is able to work better with `Status` to forward error messages from lower layers, and has better composability, which is not very obvious at the moment, but will become more obvious as `StatusOr` is used more and more.

tisonkun (dismissed):
@tanruixiang sorry that we merge #943 first so there's some conflict. Please resolve the conflict and I think this patch is ready to merge.

tanruixiang:
> @tanruixiang sorry that we merge #943 first so there's some conflict. Please resolve the conflict and I think this patch is ready to merge.

Solved, but CI doesn't seem to start?

tisonkun (approved):

tisonkun:
Thank you!

PR #933, opened 2022-09-29
fix: ping return bulk string

### LINKED ISSUE #926: PING returns PONG if no argument is provided, otherwise return a copy of the argument as a bulk.
### Search before asking

### Version

unstable

### Minimal reproduce step

ping message

message

### What did you see instead?

wrong number of arguments

### Anything Else?

[redis ping](https://redis.io/commands/ping/)

ColinChamber:
This closes #926.


tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -0,0 +1,42 @@
+/*
```
PragmaTwice, line 1: May I ask the meaning of the filename? I think something like `ping_test.go` will be more simple.

ColinChamber: I did this to imitate [redis](https://github.com/redis/redis/blob/unstable/tests/unit/introspection.tcl#L3-L5). Kvrocks also has introspection.tcl

src/redis_cmd.cc
```diff
@@ -242,7 +242,13 @@ class CommandFlushAll : public Commander {
 class CommandPing : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    *output = Redis::SimpleString("PONG");
+    if (args_.size() == 1) {
+      *output = Redis::SimpleString("PONG");
+    } else if (args_.size() == 2) {
+      *output = Redis::SimpleString(args_[1]);
```
tisonkun: IIRC Redis will return bulk string here. That is,

```bash
$ echo -e '*1\r\n$4\r\nPING\r\n' | nc localhost 6379
+PONG

$ echo -e '*2\r\n$4\r\nPING\r\n$4\r\nPONG\r\n' | nc localhost 6379
$4
PONG
```

tests/gocase/unit/introspection/introspection_test.go
PragmaTwice: Oh, got it. Thanks.

src/redis_cmd.cc
caipengbo: How do we determine simple string and bulk string in the go test case? @tisonkun

tisonkun: Using `server.NewTCPClient` to obtain a low-level client to verify the raw response.

caipengbo: Ok, thank you. That's exactly what I need :)

ColinChamber: Thank you. This is my mistake. My branch name is ping-return-bulk-string, but I still misuse the simple string. 😓

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -0,0 +1,53 @@
...
+	rdb := srv.NewClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
+
+	t.Run("PING", func(t *testing.T) {
+		c := srv.NewTCPClient()
+		defer func() { require.NoError(t, c.Close()) }()
+		require.NoError(t, c.Write("*1\r\n$4\r\nPING\r\n"))
+		r, err := c.ReadLine()
+		require.NoError(t, err)
+		require.Contains(t, r, "+PONG")
```
caipengbo: Should we write a new test case for the redis protocol validation about PING?

It seems that semantic verification and protocol verification have been put into one test case, which I think should be split into two cases.

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -0,0 +1,53 @@
...
+	srv := util.StartServer(t, map[string]string{})
+	defer srv.Close()
+	ctx := context.Background()
+	rdb := srv.NewClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
+
+	t.Run("PING", func(t *testing.T) {
+		c := srv.NewTCPClient()
+		defer func() { require.NoError(t, c.Close()) }()
+		require.NoError(t, c.Write("*1\r\n$4\r\nPING\r\n"))
```
tisonkun: use `WriteArgs`

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -0,0 +1,53 @@
...
+		r, err := c.ReadLine()
+		require.NoError(t, err)
+		require.Contains(t, r, "+PONG")
+		require.NoError(t, c.Write("*2\r\n$4\r\nPING\r\n$4\r\nPONG\r\n"))
```
tisonkun: use `WriteArgs`

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -0,0 +1,53 @@
...
+		r, err = c.ReadLine()
+		require.NoError(t, err)
+		require.Contains(t, r, "$4")
+		r, err = c.ReadLine()
+		require.NoError(t, err)
+		require.Contains(t, r, "PONG")
+		require.EqualError(t, rdb.Do(ctx, "ping", "hello", "redis").Err(), "ERR wrong number of arguments")
```
tisonkun: consistently use `TCPClient` and just remove `rdb` outside.

tisonkun (commented):

tests/gocase/unit/introspection/introspection_test.go
ColinChamber: Yeah. Maybe it is not necessary to verify the protocol about PING. Some things are apparent. It makes the test too cumbersome.

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -0,0 +1,52 @@
...
+	srv := util.StartServer(t, map[string]string{})
+	defer srv.Close()
+
+	t.Run("PING", func(t *testing.T) {
+		c := srv.NewTCPClient()
+		defer func() { require.NoError(t, c.Close()) }()
+		require.NoError(t, c.WriteArgs("PING"))
+		r, err := c.ReadLine()
+		require.NoError(t, err)
+		require.Contains(t, r, "+PONG")
```
tisonkun: I think you should use `require.Equal` here? Also below assertions seem looser than it should be.

ColinChamber: Thank you. Updated.

tisonkun (approved):

src/redis_cmd.cc
```diff
@@ -5833,7 +5839,7 @@ class CommandXTrim : public Commander {
 
 CommandAttributes redisCommandTable[] = {
     ADD_CMD("auth", 2, "read-only ok-loading", 0, 0, 0, CommandAuth),
-    ADD_CMD("ping", 1, "read-only", 0, 0, 0, CommandPing),
+    ADD_CMD("ping", -1, "read-only", 0, 0, 0, CommandPing),
```
tisonkun, line 5842: May I ask the reason to make this change? Is there some reference from the Redis document?

The overall change looks good to me. Merging...

ColinChamber: This digit indicates the number of parameters allowed(including command name), "1 " means there is precisely one, and "-1 "means at least one.  Do you want this [source code](https://github.com/redis/redis/blob/6.2/src/server.c#L715) and the [comment](https://github.com/redis/redis/blob/6.2/src/server.c#L89)?

tisonkun: Thank you!

PR #944, opened 2022-10-04
Move TCL test unit/type/string to Go case 

### LINKED ISSUE #814: Move TCL test unit/type/string to Go case

gogim1:
closes #814


tests/gocase/unit/type/str/str_test.go
```diff
@@ -0,0 +1,649 @@
...
+*
+* Unless required by applicable law or agreed to in writing,
+* software distributed under the License is distributed on an
+* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+* KIND, either express or implied.  See the License for the
+* specific language governing permissions and limitations
+* under the License.
+ */
+
+package str
```
tisonkun: I suggest you keep the name `string` instead of `str`.

tisonkun:
You possibly need to remove the entry of string tests for TCL test helper. Otherwise, the TCL suite can fail.

See https://github.com/apache/incubator-kvrocks/commit/43cca6a1ec50bfc850b8d9d08c552a61de8c0034#diff-44b16c150e71ef9382d728c10c9098bf483e57731d98f84f94a394e81c9172c3 as an example.

tests/gocase/unit/type/str/str_test.go
gogim1: it would report an [error](https://github.com/apache/incubator-kvrocks/actions/runs/3178996337/jobs/5181038227#step:7:270) if we do that since string is a predeclared identifier

tisonkun: You can use `string_test` or `strings` :)

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -0,0 +1,649 @@
...
+		require.NoError(t, err)
+		require.Equal(t, "$-1", r)
+
+		require.Equal(t, "1", rdb.Get(ctx, "foo").Val())
+	})
+
+	t.Run("Extended SET XX option", func(t *testing.T) {
+		require.NoError(t, rdb.Del(ctx, "foo").Err())
+
+		require.NoError(t, c.WriteArgs("SET", "foo", "1", "xx"))
```
tisonkun: Please try to use `redis.Client`'s method instead of sending commands with raw client.

ditto all others

tisonkun (changes requested):

tests/gocase/unit/type/strings/strings_test.go
gogim1: `(redis.Client).Set` does not support `Extended SET XX option`, it only take `context`, `key`, `value` and `expiration` as arguements, so i use raw client to work around.

gogim1: and `(redis.Client).Set` does not support `Cas` and `Cad` methods.

gogim1: i get your point now, maybe i should use `(redis.Client).Do` for instead.

tisonkun: > Extended SET XX option

There is a `SetXX` method.

> Cas and Cad

Yes. For Kvrocks specific command, use `Do` method. See also `sint` tests.

gogim1:
https://github.com/apache/incubator-kvrocks/pull/944#discussion_r986722195

i think use `Do("SET", "foo", "1", "xx")` method is better than `SetXX`  method in `Extended SET XX option` unit test, since we are testing the `set` command with option instead of `setxx` command, and the two commands don't seem to have the same implementation:
```
root@DESKTOP-RHSJVVQ:~/incubator-kvrocks# !11
redis-cli -p 6666
127.0.0.1:6666> set foo 1 nx
OK
127.0.0.1:6666> get foo
"1"
127.0.0.1:6666> setnx foo 2
(integer) 0
127.0.0.1:6666> set foo 2 nx
(nil)
127.0.0.1:6666> 
```
what's more, in some cases, such as `Extended SET PXAT option`, you even can't find the correspond `SetPXAT` method. So i prefer to use `Do` for consistency. Any ideas? @tisonkun

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -0,0 +1,513 @@
...
+		require.Equal(t, "10", rdb.Get(ctx, "x").Val())
+	})
+
+	t.Run("SETNX against expired volatile key", func(t *testing.T) {
+		require.NoError(t, rdb.Set(ctx, "x", "10", 1*time.Second).Err())
+		require.Eventually(t, func() bool {
+			require.NoError(t, rdb.SetNX(ctx, "x", "20", 0).Err())
+			return rdb.Get(ctx, "x").Val() == "20"
+		}, 5*time.Second, 100*time.Millisecond)
+	})
```
tisonkun: This case seems different from the original case. Can you explain a bit?

tests/tcl/tests/unit/type/string.tcl
```diff
@@ -1,742 +0,0 @@
...
-        r expire x 1
-
-        # Wait for the key to expire
-        after 2000
-
-        assert_equal 1 [r setnx x 20]
-        assert_equal 20 [r get x]
-    }
-
-    # test "GETEX EX option" {
```
tisonkun, line 128: @git-hulk @ShooterIT @PragmaTwice do we simply not implement `GETEX`? Perhaps it worth an issue.

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+	t.Run("MGET command", func(t *testing.T) {
+		require.NoError(t, rdb.Set(ctx, "foo", "FOO", 0).Err())
+		require.NoError(t, rdb.Set(ctx, "bar", "BAR", 0).Err())
+		require.Equal(t, []interface{}{"FOO", "BAR"}, rdb.MGet(ctx, "foo", "bar").Val())
+	})
+
+	t.Run("MGET against non existing key", func(t *testing.T) {
+		require.NoError(t, rdb.Set(ctx, "foo", "FOO", 0).Err())
+		require.NoError(t, rdb.Set(ctx, "bar", "BAR", 0).Err())
```
tisonkun: I think the original case is stateful so that you don't re-set these keys. Also, you may keep all `r flushdb` with `require.NoError(t, rdb.FlushDB(ctx).Err())`.

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -0,0 +1,513 @@
...
+		require.Equal(t, []interface{}{"FOO", nil, "BAR"}, rdb.MGet(ctx, "foo", "baazz", "bar").Val())
+	})
+
+	t.Run("MGET against non-string key", func(t *testing.T) {
+		require.NoError(t, rdb.Set(ctx, "foo", "FOO", 0).Err())
+		require.NoError(t, rdb.Set(ctx, "bar", "BAR", 0).Err())
```
tisonkun: ditto

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -0,0 +1,513 @@
...
+		require.NoError(t, rdb.Set(ctx, "x", "10", 0).Err())
+		require.EqualValues(t, 0, rdb.Exists(ctx, "x1").Val())
+		require.EqualValues(t, 0, rdb.Exists(ctx, "y2").Val())
+		require.EqualValues(t, 1, rdb.Exists(ctx, "x").Val())
+		r := rdb.MSetNX(ctx, map[string]interface{}{
+			"x1": "xxx",
+			"y2": "yyy",
+			"x":  "20",
+		})
+		require.Equal(t, false, r.Val())
```
tisonkun:
```suggestion
		require.False(t, r.Val())
```

tisonkun: This test is in the wrong order from the original one.

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -0,0 +1,513 @@
...
+
+	t.Run("MSETNX with not existing keys", func(t *testing.T) {
+		require.EqualValues(t, 0, rdb.Exists(ctx, "x1").Val())
+		require.EqualValues(t, 0, rdb.Exists(ctx, "y2").Val())
+		r := rdb.MSetNX(ctx, map[string]interface{}{
+			"x1": "xxx",
+			"y2": "yyy",
+		})
+		require.Equal(t, true, r.Val())
+	})
```
tisonkun: ditto

tisonkun (changes requested):
Review part of this patch. Comments inline.

Please try to keep the logic identical - that is, no eager refactor. Otherwise, it's hard to review.

As a note, I reviewed until "STRLEN against plain string"

tests/tcl/tests/unit/type/string.tcl
git-hulk: Yes, we can implement `GetEX` command, will create an issue to track.

gogim1:
> Review part of this patch. Comments inline.
> 
> Please try to keep the logic identical - that is, no eager refactor. Otherwise, it's hard to review.
> 
> As a note, I reviewed until "STRLEN against plain string"

I've kept the logic identical. Sorry to bring so much trouble. 😂

GoGim1:
[error message](https://github.com/apache/incubator-kvrocks/actions/runs/3210776460/jobs/5251928178#step:11:987)

gogim1:
1. It seems something wrong with [TestExpire](https://github.com/apache/incubator-kvrocks/actions/runs/3210776460/jobs/5251928178#step:11:202)?
2. No ideas why [this](https://github.com/apache/incubator-kvrocks/actions/runs/3210776460/jobs/5251928178#step:11:987) happened. Maybe due to the unstable test?

tanruixiang:
> 1. It seems something wrong with [TestExpire](https://github.com/apache/incubator-kvrocks/actions/runs/3210776460/jobs/5251928178#step:11:202)?
> 2. No ideas why [this](https://github.com/apache/incubator-kvrocks/actions/runs/3210776460/jobs/5251928178#step:11:987) happened. Maybe due to the unstable test?

I remember having similar unstable results before, long time no see after tison fix.  Do we need to use the same settings as List here? https://github.com/apache/incubator-kvrocks/pull/869 @tisonkun

tisonkun (approved):
LGTM with minor changes.

git-hulk (approved):

PragmaTwice (approved):

tisonkun:
Thank you @GoGim1!

PR #953, opened 2022-10-05
feat: add mutil_path to conf

### LINKED ISSUE #918: support Multi-disk(Multi-path)
### Search before asking

### Motivation

When the host has multiple disks, multiple disks can be used for data storage to increase the performance of KVROCKS.
Hot data can be stored on local SSDS and cold data can be stored on cloud disks

### Solution

```
option.db_paths = {
                     {"/disk1", 1000 * 1000 * 1000},
                     {"/disk2", 1000 * 1000 * 1000},
                     {"/disk3", 1000 * 1000 * 1000},
                     {"/disk4", 1000 * 1000 * 1000}};
```

https://github.com/facebook/rocksdb/blob/main/include/rocksdb/options.h#L672

xiaobiaozhao:
It close https://github.com/apache/incubator-kvrocks/issues/918

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -1,3 +1,4 @@
+rocksdb.db_paths "/mnt/ssd a10; /mnt/hdd 10"
 ################################ GENERAL #####################################
 
 # By default kvrocks listens for connections from localhost interface.
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -760,5 +760,56 @@ rocksdb.write_options.low_pri no
 # Default: no
 rocksdb.write_options.memtable_insert_hint_per_batch no
 
+# A list of paths where SST files can be put into, with its target size.
+# Newer data is placed into paths specified earlier in the vector while
+# older data gradually moves to paths specified later in the vector.
+#
+# For example, you have a flash device with 10GB allocated for the DB,
+# as well as a hard drive of 2TB, you should config it to be:
+#   [{"/flash_path", 10GB}, {"/hard_drive", 2TB}]
+#
+# The system will try to guarantee data under each path is close to but
+# not larger than the target size. But current and future file sizes used
+# by determining where to place a file are based on best-effort estimation,
+# which means there is a chance that the actual size under the directory
+# is slightly more than target size under some workloads. User should give
+# some buffer room for those cases.
+#
+# If none of the paths has sufficient room to place a file, the file will
+# be placed to the last path anyway, despite to the target size.
+#
+# Placing newer data to earlier paths is also best-efforts. User should
+# expect user files to be placed in higher levels in some extreme cases.
+#
+# If left empty, only one path will be used, which is db_name passed when
+# opening the DB.
+# Default: empty
+#
+# If KVRocks stores a lot of cold data, you can use this parameter to
+# reduce the amount of high performance disk usage.
+#
+# Note that when enable_db_paths is set to yes, kvrocks db date will save to
+# rocksdb.db_paths0' path & rocksdb.db_paths1's path
+#
+# Default: no
+# rocksdb.enable_db_paths yes
+#
+# rocksdb.db_paths0's value will be set into rocksdb.db_paths[0]'s path
+#
+# rocksdb.db_paths0 /mnt/ssd
+#
+# rocksdb.db_paths0_size_gb's value will be set into rocksdb.db_paths[0]'s size(GB)
+#
+# rocksdb.db_paths0_size_gb 10
+#
+# rocksdb.db_paths1's value will be set into rocksdb.db_paths[1]'s path
+#
+# rocksdb.db_paths1 /mnt/hdd
+#
+# rocksdb.db_paths1_size_gb's value will be set into rocksdb.db_paths[1]'s size(GB)
+# If value is 0, means that rocksdb.db_paths off
+#
+# rocksdb.db_paths1_size_gb 10
+
 ################################ NAMESPACE #####################################
 # namespace.test change.me
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -1,4 +1,3 @@
-rocksdb.db_paths "/mnt/ssd a10; /mnt/hdd 10"
 ################################ GENERAL #####################################
 
 # By default kvrocks listens for connections from localhost interface.
@@ -789,28 +788,8 @@ rocksdb.write_options.memtable_insert_hint_per_batch no
 # If KVRocks stores a lot of cold data, you can use this parameter to
 # reduce the amount of high performance disk usage.
 #
-# Note that when enable_db_paths is set to yes, kvrocks db date will save to
-# rocksdb.db_paths0' path & rocksdb.db_paths1's path
-#
-# Default: no
-# rocksdb.enable_db_paths yes
-#
-# rocksdb.db_paths0's value will be set into rocksdb.db_paths[0]'s path
-#
-# rocksdb.db_paths0 /mnt/ssd
-#
-# rocksdb.db_paths0_size_gb's value will be set into rocksdb.db_paths[0]'s size(GB)
-#
-# rocksdb.db_paths0_size_gb 10
-#
-# rocksdb.db_paths1's value will be set into rocksdb.db_paths[1]'s path
-#
-# rocksdb.db_paths1 /mnt/hdd
-#
-# rocksdb.db_paths1_size_gb's value will be set into rocksdb.db_paths[1]'s size(GB)
-# If value is 0, means that rocksdb.db_paths off
-#
-# rocksdb.db_paths1_size_gb 10
+# Default: empty
+# rocksdb.db_paths "/mnt/ssd 10G; /mnt/hdd 1T;"
 
 ################################ NAMESPACE #####################################
 # namespace.test change.me
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -187,8 +187,19 @@ rocksdb::Options Storage::InitOptions() {
     std::vector<std::string> paths = Util::Split(config_->RocksDB.db_paths, ";");
     for (auto & path_size : paths) {
       std::vector<std::string> ps = Util::Split(path_size, " \t");
-      // auto parse_result = ParseInt<uint32_t>(ps[1]);
-      // options.db_paths.emplace_back(rocksdb::DbPath(ps[0], *parse_result * GiB));
+      auto parse_result = TryParseInt<uint32_t>(ps[1].c_str());
+      std::string unit = Util::ToLower(std::get<1>(*parse_result));
+      u_int64_t size = std::get<0>(*parse_result);
+      if (unit == "k") {
+        size *= KiB;
+      } else if (unit == "m") {
+        size *= MiB;
+      } else if (unit == "g") {
+        size *= GiB;
+      } else if (unit == "t") {
+        size *= TiB;
+      }
+      options.db_paths.emplace_back(rocksdb::DbPath(ps[0], size));
     }
   }
 
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -183,13 +183,13 @@ rocksdb::Options Storage::InitOptions() {
   options.max_bytes_for_level_multiplier = config_->RocksDB.max_bytes_for_level_multiplier;
   options.level_compaction_dynamic_level_bytes = config_->RocksDB.level_compaction_dynamic_level_bytes;
 
-  if (config_->RocksDB.enable_db_paths &&
-      !config_->RocksDB.db_paths0.empty() && !config_->RocksDB.db_paths1.empty()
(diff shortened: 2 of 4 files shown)


kvrocks.conf
```diff
@@ -745,10 +745,61 @@ rocksdb.write_options.low_pri no
...
+# is slightly more than target size under some workloads. User should give
+# some buffer room for those cases.
+#
+# If none of the paths has sufficient room to place a file, the file will
+# be placed to the last path anyway, despite to the target size.
+#
+# Placing newer data to earlier paths is also best-efforts. User should
+# expect user files to be placed in higher levels in some extreme cases.
+#
+# If left empty, only one path will be used, which is db_name passed when
```
torwig, line 784: What option do these lines describe (with Default: empty)?

kvrocks.conf
```diff
@@ -745,10 +745,61 @@ rocksdb.write_options.low_pri no
...
+# rocksdb.db_paths0_size_gb 10
+#
+# rocksdb.db_paths1's value will be set into rocksdb.db_paths[1]'s path
+#
+# rocksdb.db_paths1 /mnt/hdd
+#
+# rocksdb.db_paths1_size_gb's value will be set into rocksdb.db_paths[1]'s size(GB)
+# If value is 0, means that rocksdb.db_paths off
+#
+# rocksdb.db_paths1_size_gb 10
```
PragmaTwice: Why is there only two hard-indexed option for path? I think this design can hardly be called elegant.

src/config.cc
```diff
@@ -206,6 +206,11 @@ Config::Config() {
         false, new IntField(&RocksDB.max_bytes_for_level_multiplier, 10, 1, 100)},
       {"rocksdb.level_compaction_dynamic_level_bytes",
         false, new YesNoField(&RocksDB.level_compaction_dynamic_level_bytes, false)},
+      {"rocksdb.enable_db_paths", true, new YesNoField(&RocksDB.enable_db_paths, false)},
+      {"rocksdb.db_paths0", true, new StringField(&RocksDB.db_paths0, "")},
+      {"rocksdb.db_paths0_size_gb", true, new IntField(&RocksDB.db_paths0_size_gb, 0, 0, INT_MAX)},
```
PragmaTwice, line 211: I prefer MiB, and you can just remove the `_gb` suffix, and describe it in the configuration file to users.

PragmaTwice:
I do not quite love the current design for configuration, I think a better way would be something like:

```
rocksdb.db_paths /mnt/ssd /mnt/hdd
rocksdb.db_paths_size 10 10
```

Or simply:
```
rocksdb.db_paths /mnt/ssd 10 /mnt/hdd 10
```

Since we have no quoted multiple field configuration now, I think we can just forbid pathes that include spaces.

kvrocks.conf
xiaobiaozhao: I've tested that db_paths supports up to four paths, but only the first and last items will write data

PragmaTwice: db_paths in rocksdb is typed vector, so we should just forward an array to it regardless of its behavior now. If its behavior changes in future versions, we can hardly change our configuration options since we need to keep compatibility.

xiaobiaozhao: Good idea. I can redesign the plan.

xiaobiaozhao:
> I do not quite love the current design for configuration, I think a better way would be something like:
> 
> ```
> rocksdb.db_paths /mnt/ssd /mnt/hdd
> rocksdb.db_paths_size 10 10
> ```
> 
> Or simply:
> 
> ```
> rocksdb.db_paths /mnt/ssd 10 /mnt/hdd 10
> ```
> 
> Since we have no quoted multiple field configuration now, I think we can just forbid pathes that include spaces.

I think this is better
rocksdb.db_paths "/mnt/ssd 10; /mnt/hdd 10"

aleksraiden:
maybe 10Gb / 100Mb - with size dimensions?

git-hulk:
> maybe 10Gb / 100Mb - with size dimensions?

Yes, I think it will be better if we can set the unit explicitly. 

`rocksdb.db_paths /mnt/ssd 10Gb /mnt/hdd 10Mb` or `/mnt/ssd 10Gb; /mnt/hdd 10Mb` is good.

PragmaTwice:
Good idea. I think just something like `1` `1k` `1M` `1G` `1T` is fine, without the `B`.

git-hulk:
> Good idea. I think just something like `1` `1k` `1M` `1G` `1T` is fine, without the `B`.

Agree with this, looks more intuitive.

tanruixiang:
> I've tested that db_paths supports up to four paths, but only the first and last items will write data

@xiaobiaozhao I noticed that the four paths of rocksdb are all writable. Is it possible that there is a problem with the configuration? Try the same as this [rocksdb‘s testcase](https://github.com/facebook/rocksdb/blob/main/db/db_universal_compaction_test.cc#L1228) configuration, will there be any changes?

xiaobiaozhao:
> > I've tested that db_paths supports up to four paths, but only the first and last items will write data
> 
> @xiaobiaozhao I noticed that the four paths of rocksdb are all writable. Is it possible that there is a problem with the configuration? Try the same as this [rocksdb‘s testcase](https://github.com/facebook/rocksdb/blob/main/db/db_universal_compaction_test.cc#L1228) configuration, will there be any changes?

I look at this demo, and when I try it out, I find that I only write 0,3 paths. There may be other configurations that I need to keep looking into.

kvrocks.conf
caipengbo: > I've tested that db_paths supports up to four paths, but only the first and last items will write data

I think we need to fully understand rocksdb's behavior first, so that we can use this feature in kvrocks. We won't be able to use this feature until we know if it will behave as we expect.

caipengbo:
> I've tested that db_paths supports up to four paths, but only the first and last items will write data

I think we need to fully understand rocksdb's behavior first, so that we can use this feature in kvrocks. We won't be able to use this feature until we know if it will behave as we expect.

tisonkun:
We may not simply forward options to what RocksDB supports but clarify what it benefits and how it does it.

I don't think simply adding dbpaths options helps for separating cold and hot data, it's that you extend the space to write data. If there's no real-world use case that benefits from binding to multiple paths (multiple disks?), I tend not to forward the option.

tanruixiang:
> We may not simply forward options to what RocksDB supports but clarify what it benefits and how it does it.
> 
> I don't think simply adding dbpaths options helps for separating cold and hot data, it's that you extend the space to write data. If there's no real-world use case that benefits from binding to multiple paths (multiple disks?), I tend not to forward the option.

I think this option is useful if multiple storage media (Such as SSD and HDD) of different speeds are used.

tisonkun:
> I think this option is useful if multiple storage media (Such as SSD and HDD) of different speeds are used.

@tanruixiang can we control which data are written to a specific path? It seems not and then when I config multiple paths, the only purpose to me is that the data volume is large while one disk cannot serve it and I need another ones.

tanruixiang:
> > I think this option is useful if multiple storage media (Such as SSD and HDD) of different speeds are used.
> 
> @tanruixiang can we control which data are written to a specific path? It seems not and then when I config multiple paths, the only purpose to me is that the data volume is large while one disk cannot serve it and I need another ones.

`Rocksdb` will be stored in the order of `db_path`, and the low-level sstable will be placed in the earlier path. For example, if we have an SSD and an HDD, we always want the low-level sstable to be in the SSD.

xiaobiaozhao:
There's a rocksdb issue about db_paths . https://github.com/facebook/rocksdb/issues/10841

tisonkun:
I think we may give this patch a try.

Before asking @xiaobiaozhao to resolve the conflict, I'm calling for a consensus here:

cc @PragmaTwice @git-hulk @ShooterIT @tanruixiang @torwig @caipengbo do you see any blockers to resolve or information needed to provide more here to move this patch forward?

git-hulk:
@xiaobiaozhao The only issue is only parts of DB paths are used which are mentioned in https://github.com/apache/incubator-kvrocks/pull/953#issuecomment-1286391126, it looks still no follow-up from the RocksDB community.

xiaobiaozhao:
> @xiaobiaozhao The only issue is only parts of DB paths are used which are mentioned in [#953 (comment)](https://github.com/apache/incubator-kvrocks/pull/953#issuecomment-1286391126), it looks still no follow-up from the RocksDB community.

I've tested a lot of case that I can only write two paths

kvrocks.conf
```diff
@@ -760,5 +760,36 @@ rocksdb.write_options.low_pri no
...
+# expect user files to be placed in higher levels in some extreme cases.
+#
+# If left empty, only one path will be used, which is db_name passed when
+# opening the DB.
+# Default: empty
+#
+# If KVRocks stores a lot of cold data, you can use this parameter to
+# reduce the amount of high performance disk usage.
+#
+# Default: empty
```
PragmaTwice, line 791: duplicated "default"

src/storage/storage.cc
```diff
@@ -183,6 +183,26 @@ rocksdb::Options Storage::InitOptions() {
...
+      if (unit == "k") {
+        size *= KiB;
+      } else if (unit == "m") {
+        size *= MiB;
+      } else if (unit == "g") {
+        size *= GiB;
+      } else if (unit == "t") {
+        size *= TiB;
+      }
+      options.db_paths.emplace_back(rocksdb::DbPath(ps[0], size));
```
PragmaTwice, line 202: Hi, now you can use `ParseSizeAndUnit` to achieve this function rather than manually parsing.

kvrocks.conf
PasunuriSrinidhi: These lines describe the behavior of the "datadir" option in a database configuration file.

By default, if "datadir" is left empty, only one path will be used for storing the database files. However, if multiple paths are specified, the database will attempt to place newer data in the earlier paths.

The lines also state that this behavior is "best-efforts," meaning that the database will try its best to follow this rule but may not always be able to do so. In some extreme cases, user files may be placed in higher levels, which is something the user should be aware of.

Overall, these lines describe how the "datadir" option works and what behavior the user can expect if multiple paths are specified.

aleksraiden:
Hi all, is there any updates? This feature is very interesting for us. Could we (with @torwig) helping for this PR?

xiaobiaozhao:
> Hi all, is there any updates? This feature is very interesting for us. Could we (with @torwig) helping for this PR?
The feature compilation is OK, but I didn't go result as expected, so we want to fix that first. Check it out if you're interested.
https://github.com/facebook/rocksdb/issues/10841

aleksraiden:
> > Hi all, is there any updates? This feature is very interesting for us. Could we (with @torwig) helping for this PR?
> > The feature compilation is OK, but I didn't go result as expected, so we want to fix that first. Check it out if you're interested.
> > [facebook/rocksdb#10841](https://github.com/facebook/rocksdb/issues/10841)

Could you check your test db_path example for comparability with latest 8.3.2 version of rocksdb? If I try to build, I have segfault.  maybe this feature can be fixed in latest release, lets check. Lot of thanks, @xiaobiaozhao

xiaobiaozhao:
> > > Hi all, is there any updates? This feature is very interesting for us. Could we (with @torwig) helping for this PR?
> > > The feature compilation is OK, but I didn't go result as expected, so we want to fix that first. Check it out if you're interested.
> > > [facebook/rocksdb#10841](https://github.com/facebook/rocksdb/issues/10841)
> 
> Could you check your test db_path example for comparability with latest 8.3.2 version of rocksdb? If I try to build, I have segfault. maybe this feature can be fixed in latest release, lets check. Lot of thanks, @xiaobiaozhao

I'll try as soon as possible.

xiaobiaozhao:
> > > > Hi all, is there any updates? This feature is very interesting for us. Could we (with @torwig) helping for this PR?
> > > > The feature compilation is OK, but I didn't go result as expected, so we want to fix that first. Check it out if you're interested.
> > > > [facebook/rocksdb#10841](https://github.com/facebook/rocksdb/issues/10841)
> > 
> > 
> > Could you check your test db_path example for comparability with latest 8.3.2 version of rocksdb? If I try to build, I have segfault. maybe this feature can be fixed in latest release, lets check. Lot of thanks, @xiaobiaozhao
> 
> I'll try as soon as possible.

I just tested it with 8.3.2 version, rocksdb still using only 0/3 paths.
```
du -h /mnt/rocksdb/
471M    /mnt/rocksdb/a
4.0K    /mnt/rocksdb/c
1.5G    /mnt/rocksdb/d
4.0K    /mnt/rocksdb/b
2.0G    /mnt/rocksdb/
```
I think we should try to study this feature

https://rocksdb.org/blog/2022/11/09/time-aware-tiered-storage.html

git-hulk:
Close this PR due to lack of active, feel free to reopen it if want to continue working on this.

PR #956, opened 2022-10-07
Move TCL test unit/introspection to Go case

### LINKED ISSUE #890: Move TCL test unit/introspection to Go case

tisonkun:
This closes #890.


tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -20,33 +20,230 @@
...
+		// normal clients
+		c2 := srv.NewClient()
+		require.NoError(t, c2.Do(ctx, "CLIENT", "SETNAME", "normal").Err())
+		defer func() { require.NoError(t, c2.Close()) }()
+
+		require.EqualValues(t, 2, rdb.ClientKillByFilter(ctx, "type", "pubsub").Val())
+
+		// now the pubsub client should no longer be listed
+		// but normal client should not be dropped
+		t.Skip("cannot run the following assertions")
```
tisonkun: The following assertions failed. Still debugging. Others work well for me.

tisonkun: Fixed at https://github.com/apache/incubator-kvrocks/pull/956/commits/ba0cb9f8d6e3ae9f58ae57552c448e00af542d17.

cc @@vmihailenco you may be interested in this case:

When I'm migrating Redis's TCL tests to Go cases using go-redis, I notice that if I use go-redis's `(P)Subscribe` interface, it will hold some extra resources so that the connections underlayer cannot be released. It causes this case failed: even with `CLIENT KILL TYPE PUBSUB`, the connections remain.

git-hulk:
@tisonkun  The test `DEBUG will freeze server` also failed casually in Go, because it may use a different connection between `Debug Sleep 2.2` and `Set a b` command and `Set a b` won't wait for the debug command when it's served by the different worker.

Below is TCP packets, the debug sleep command is sent by connection(port 49225) and the set command is in connection(port 49221):
```
11:24:50.845605 IP localhost.49225 > localhost.16379: Flags [P.], seq 1:36, ack 1, win 6379, options [nop,nop,TS val 1201924138 ecr 643167600], length 35
....E..W..@.@............I?./..C.L.......K.....
G..*&U.p*3
$5
DEBUG
$5
sleep
$3
2.2

11:24:50.845681 IP localhost.49221 > localhost.16379: Flags [P.], seq 97:124, ack 301, win 6375, options [nop,nop,TS val 3933309289 ecr 2680271241], length 27
....E..O..@.@............E?. .F..........C.....
.q.i....*3
$3
set
$1
a
$1
b
```

May be we can limit the pool size to 1 to workaround this issue.

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -20,33 +20,223 @@
...
+			fmt.Printf("CLIENT LIST: %v\n", r)
+			return strings.Count(r, "pubsub") == 0 && strings.Count(r, "normal") == 1
+		}, 5*time.Second, 100*time.Millisecond)
+	})
+
+	t.Run("DEBUG will freeze server", func(t *testing.T) {
+		// use TCPClient to avoid waiting for reply
+		c := srv.NewTCPClient()
+		defer func() { require.NoError(t, c.Close()) }()
+		require.NoError(t, c.WriteArgs("DEBUG", "sleep", "2.2"))
```
git-hulk, line 175:
```suggestion
		require.NoError(t, c.WriteArgs("DEBUG", "sleep", "2.2"))
		
	    // Sleep 100ms to prevent the successive set command to be executed
		// before the debug command since there are in the different connection.
		time.Sleep(100 * time.Millisecond)
```

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -20,33 +20,227 @@
...
+		
+	    // Sleep 100ms to prevent the successive set command to be executed
+		// before the debug command since there are in the different connection.
+		time.Sleep(100 * time.Millisecond)
```
git-hulk:
```suggestion
        // sleep 100ms to prevent the successive set command to be executed
        // before the debug command since there are in the different connection.
		time.Sleep(100 * time.Millisecond)
```

git-hulk:
> @tisonkun The test `DEBUG will freeze server` also failed casually in Go, because it may use a different connection between `Debug Sleep 2.2` and `Set a b` command and `Set a b` won't wait for the debug command when it's served by the different worker.
> 
> Below is TCP packets, the debug sleep command is sent by connection(port 49225) and the set command is in connection(port 49221):
> 
> ```
> 11:24:50.845605 IP localhost.49225 > localhost.16379: Flags [P.], seq 1:36, ack 1, win 6379, options [nop,nop,TS val 1201924138 ecr 643167600], length 35
> ....E..W..@.@............I?./..C.L.......K.....
> G..*&U.p*3
> $5
> DEBUG
> $5
> sleep
> $3
> 2.2
> 
> 11:24:50.845681 IP localhost.49221 > localhost.16379: Flags [P.], seq 97:124, ack 301, win 6375, options [nop,nop,TS val 3933309289 ecr 2680271241], length 27
> ....E..O..@.@............E?. .F..........C.....
> .q.i....*3
> $3
> set
> $1
> a
> $1
> b
> ```
> 
> Maybe we can limit the pool size to 1 to work around this issue.

@tisonkun this conclusion is incorrect. The debug command is an exclusive command, so it will behave like the Redis(single thread on the server side). The root cause is the `set` command can be executed before the `debug` command since they are in a different connection.

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -20,33 +20,227 @@
...
+		require.NoError(t, c2.Do(ctx, "CLIENT", "SETNAME", "normal").Err())
+		defer func() { require.NoError(t, c2.Close()) }()
+
+		require.EqualValues(t, 2, rdb.ClientKillByFilter(ctx, "type", "pubsub").Val())
+
+		// now the pubsub client should no longer be listed
+		// but normal client should not be dropped
+		require.Eventually(t, func() bool {
+			r := rdb.ClientList(ctx).Val()
+			fmt.Printf("CLIENT LIST: %v\n", r)
```
git-hulk:
```suggestion
```

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -20,33 +20,227 @@
...
+		defer func() { require.NoError(t, c0.Close()) }()
+		require.NoError(t, c0.Do(ctx, "CLIENT", "SETNAME", "pubsub").Err())
+		r := c0.Do(ctx, "SUBSCRIBE", "foo")
+		require.NoError(t, r.Err())
+		require.Equal(t, "[subscribe foo 1]", fmt.Sprintf("%v", r.Val()))
+
+		// psubscribe clients
+		c1 := srv.NewClient()
+		defer func() { require.NoError(t, c1.Close()) }()
+		require.NoError(t, c1.Do(ctx, "CLIENT", "SETNAME", "pubsub").Err())
```
git-hulk:
```suggestion
		require.NoError(t, c1.Do(ctx, "CLIENT", "SETNAME", "pubsub_patterns").Err())
```

tisonkun:
@git-hulk Thank you! Updated.

git-hulk (approved):

tests/gocase/unit/introspection/introspection_test.go
vmihailenco: @tisonkun PubSub will try to reconnect when connection is lost. Perhaps that is what you are seeing?

tisonkun: @vmihailenco Thank you! It explains the case.

tisonkun: Is there a way to suppress this manner?

PR #958, opened 2022-10-07
Support save rdb

ColinChamber:
This refers to #1001.

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -55,6 +55,7 @@
 #include "slot_migrate.h"
 #include "storage.h"
 #include "util.h"
+#include "redis_rdb.h"
 
 namespace Redis {
 
@@ -3578,6 +3579,18 @@ class CommandBGSave: public Commander {
   }
 };
 
+class CommandSave: public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    Redis::RedisDatabase RDB(svr->storage_, conn->GetNamespace());
+    Config *config = svr->GetConfig();
+    Status s = RDB.Dump(config->dir + "/" + config->rdb_name);
+    if (!s.IsOK()) return s;
+    *output = Redis::SimpleString("OK");
+    return Status::OK();
+  }
+};
+
 class CommandFlushBackup : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
@@ -6033,6 +6046,7 @@ CommandAttributes redisCommandTable[] = {
     ADD_CMD("script", -2, "exclusive no-script", 0, 0, 0, CommandScript),
 
     ADD_CMD("compact", 1, "read-only no-script", 0, 0, 0, CommandCompact),
+    ADD_CMD("save", 1, "read-only no-script exclusive", 0, 0, 0, CommandSave),
     ADD_CMD("bgsave", 1, "read-only no-script", 0, 0, 0, CommandBGSave),
     ADD_CMD("flushbackup", 1, "read-only no-script", 0, 0, 0, CommandFlushBackup),
     ADD_CMD("slaveof", 3, "read-only exclusive no-script", 0, 0, 0, CommandSlaveOf),
--- a/src/redis_rdb.cc
+++ b/src/redis_rdb.cc
@@ -33,6 +33,33 @@
 
 
 namespace Redis {
+
+/* 
+*  Making Kvrocks write RDB is more straightforward than making it parse RDB.
+*  Because Redis is backward compatible, which can load the original RDB file. 
+*  We can easily utilize the primitive RDB object types, which follow the simple 
+*  encoding pattern:  <type> <len> <key> <len> <value> 
+*  or <type> <len> <key> <size> <data> (<data> consists of <len><key(value)>)
+*  RDB_TYPE_STRING, RDB_TYPE_LIST, RDB_TYPE_SET, RDB_TYPE_ZSET, RDB_TYPE_HASH,
+*  these five types follow this pattern. We ignore the RDB_TYPE_HASH_ZIPMAP, 
+*  RDB_TYPE_LIST_ZIPLIST, RDB_TYPE_SET_INTSET, and other complicated types 
+*  only implement the five simple type.
+* 
+*  There is a mapping relationship between Redis object type and RDB object type.
+*  But you don't have to worry about Redis using ineffective object types after loading RDB. 
+*  Redis will convert it to an efficient type automatically.
+*  E.g
+*  	kvrocks> hmset foo foo1 bar1 foo2 bar2
+*	kvrocks> save (use RDB_TYPE_HASH rather than RDB_TYPE_HASH_ZIPLIST)
+*   --------  Redis loads the RDB that kvrocks dumped  --------
+*   redis> OBJECT ENCODING foo
+*   "ziplist"
+*   
+*  There are more detailed references to RDB.
+*   https://github.com/sripathikrishnan/redis-rdb-tools/wiki/Redis-RDB-Dump-File-Format
+*   https://github.com/sripathikrishnan/redis-rdb-tools/blob/master/docs/RDB_Version_History.textile
+*   https://rdb.fnordig.de/file_format.html
+*/
 Status RedisDatabase::Dump(const std::string &file_name) {
   Redis::String string_db(storage_, namespace_);
   Redis::Hash hash_db(storage_, namespace_);
@@ -88,7 +115,7 @@ Status RedisDatabase::Dump(const std::string &file_name) {
             if (rs.ok()) {
                 s = util.SaveLen(field_values.size());
                 if (!s.IsOK()) return s;
-                for (const auto fv : field_values) {
+                for (const auto &fv : field_values) {
                     s = util.SaveStringObject(fv.field);
                     if (!s.IsOK()) return s;
                     s = util.SaveStringObject(fv.value);
@@ -101,7 +128,7 @@ Status RedisDatabase::Dump(const std::string &file_name) {
             if (rs.ok()) {
                 s = util.SaveLen(elems.size());
                 if (!s.IsOK()) return s;
-                for (const auto item : elems) {
+                for (const auto &item : elems) {
                     s = util.SaveStringObject(item);
                     if (!s.IsOK()) return s;
                 }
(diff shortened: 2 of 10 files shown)


src/redis_rdb.h
```diff
@@ -0,0 +1,82 @@
...
+#include "redis_db.h"
+
+namespace Redis {
+class RedisDatabase : public Database {
+ public:
+  RedisDatabase(Engine::Storage *storage, const std::string &ns) : Database(storage, ns) {}
+  Status Dump(const std::string &file_name);
+
+ private:
+  class Util {
```
torwig: I think that `Util` is not a very suitable name for a class. Could you please choose something more descriptive?

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+
+#include "redis_string.h"
+#include "redis_list.h"
+#include "redis_set.h"
+#include "redis_hash.h"
+#include "redis_zset.h"
+#include "redis_bitmap.h"
+#include "db_util.h"
+
+#include "redis_rdb.h"
```
torwig: Perhaps `cpplint` will be arguing about the header file sequence. `#include "redis_rdb.h"` shoudl be the first in `.cc` file.

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+    return saveRaw(&expiretime, 4);
+}
+
+Status RedisDatabase::Util::saveType(unsigned char type) {
+    return saveRaw(&type, 1);
+}
+
+Status RedisDatabase::Util::saveRaw(const void *p, const int &len) {
+    file_.write((const char *)p, len);
+    if (file_.bad())
```
torwig: I think that if you prefer without curly brackets, then the `if` clause and expression should be in one line. Otherwise, it's better to put curly brackets, not relying on the indentation.

src/redis_rdb.h
```diff
@@ -0,0 +1,82 @@
...
+#define RDB_TYPE_STRING 0
+#define RDB_TYPE_LIST   1
+#define RDB_TYPE_SET    2
+#define RDB_TYPE_ZSET   3
+#define RDB_TYPE_HASH   4
+
+#define RDB_ENCVAL 3
+#define RDB_ENC_INT8  0 /* 8 bit signed integer */
+#define RDB_ENC_INT16 1 /* 16 bit signed integer */
+#define RDB_ENC_INT32 2 /* 32 bit signed integer */
```
PragmaTwice: You can use const or constexpr variables rather than macros.

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+            break;
+    }
+    if (!rs.ok()) return Status(Status::NotOK, rs.ToString());
+  }
+  s = util.SaveTail();
+  return s;
+}
+
+RedisDatabase::Util::Util(const std::string &file_name) {
+    file_.open(file_name, std::ofstream::binary);
```
PragmaTwice, line 204: What happens if a file fails to open? Will the user be informed?

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+        case kRedisZSet:
+            return saveType(RDB_TYPE_ZSET);
+        case kRedisBitmap:
+            return saveType(RDB_TYPE_STRING);
+        default:
+            return Status(Status::NotOK, "illegal type");
+    }
+}
+
+Status RedisDatabase::Util::saveSecondTime(const int32_t &remain) {
```
PragmaTwice:
```suggestion
Status RedisDatabase::Util::saveSecondTime(int32_t remain) {
```

I think this const ref is useless.

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+    int64_t now;
+    int32_t expiretime;
+    rocksdb::Env::Default()->GetCurrentTime(&now);
+    expiretime = static_cast<int32_t>(now) + remain;
```
PragmaTwice:
```suggestion
    rocksdb::Env::Default()->GetCurrentTime(&now);
    int32_t expiretime = static_cast<int32_t>(now) + remain;
```
declare variables at where you really use it

PragmaTwice: same as all other places

src/redis_rdb.h
```diff
@@ -0,0 +1,82 @@
...
+                            const int &remain,
+                            const RedisType &type);
+    Status SaveStringObject(const std::string &str);
+    Status SaveLen(uint64_t len);
+    Status SaveMeta();
+    Status SaveTail();
+
+   private:
+    std::ofstream file_;
+    Status saveType(unsigned char type);
```
PragmaTwice, line 46: you can define an enum class for your `type` here

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+    return saveRaw(&expiretime, 4);
+}
+
+Status RedisDatabase::Util::saveType(unsigned char type) {
+    return saveRaw(&type, 1);
+}
+
+Status RedisDatabase::Util::saveRaw(const void *p, const int &len) {
```
PragmaTwice:
```suggestion
Status RedisDatabase::Util::saveRaw(const char *p, int len) {
```

And, why is it `int` rather than unsigned integer here?

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+    file_.open(file_name, std::ofstream::binary);
+}
+
+RedisDatabase::Util::~Util() {
+    file_.close();
+}
+
+Status RedisDatabase::Util::SaveMetaKeyPair(const std::string &key,
+                                            const int32_t &remain,
+                                            const RedisType &type) {
```
PragmaTwice:
```suggestion
Status RedisDatabase::Util::SaveMetaKeyPair(const std::string &key,
                                            int32_t remain,
                                            const RedisType &type) {
```

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+        if (!s.IsOK()) return s;
+    } else if (len <= UINT32_MAX) {
+        /* Save a 32 bit len */
+        buf[0] = RDB_32BITLEN;
+         s = saveRaw(buf, 1);
+        if (!s.IsOK()) return s;
+        uint32_t len32 = htonl(len);
+        s = saveRaw(&len32, 4);
+    }
+    return s;
```
PragmaTwice, line 271:
```suggestion
    if (len < (1 << 6)) {
        /* Save a 6 bit len */
        buf[0] = (len & 0xFF) | (RDB_6BITLEN << 6);
        s = saveRaw(buf, 1);
    } else if (len < (1 << 14)) {
        /* Save a 14 bit len */
        buf[0] = ((len >> 8) & 0xFF) | (RDB_14BITLEN << 6);
        buf[1] = len & 0xFF;
        s = saveRaw(buf, 1);
    } else if (len <= UINT32_MAX) {
        /* Save a 32 bit len */
        buf[0] = RDB_32BITLEN;
         s = saveRaw(buf, 1);
        if (!s.IsOK()) return s;
        uint32_t len32 = htonl(len);
        s = saveRaw(&len32, 4);
    }
    return s;
```

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+// Length encoding is a variable byte encoding designed to use as few bytes as possible,
+// which use redis logic directly
+Status RedisDatabase::Util::SaveLen(uint64_t len) {
+    unsigned char buf[2];
+    Status s;
+    if (len < (1 << 6)) {
+        /* Save a 6 bit len */
+        buf[0] = (len & 0xFF) | (RDB_6BITLEN << 6);
+        s = saveRaw(buf, 1);
+        if (!s.IsOK()) return s;
```
PragmaTwice:
```suggestion
```

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+    } else if (len < (1 << 14)) {
+        /* Save a 14 bit len */
+        buf[0] = ((len >> 8) & 0xFF) | (RDB_14BITLEN << 6);
+        buf[1] = len & 0xFF;
+        s = saveRaw(buf, 1);
+        if (!s.IsOK()) return s;
```
PragmaTwice:
```suggestion
```

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+    } else if (len <= UINT32_MAX) {
```
PragmaTwice, line 263:
```suggestion
    } else {
```

src/redis_rdb.cc
PragmaTwice, line 292: What happens if the length exceeds uint32?

src/redis_rdb.cc
```diff
@@ -0,0 +1,429 @@
...
+  return s;
+}
+
+RedisDatabase::Util::Util(const std::string &file_name) {
+    file_.open(file_name, std::ofstream::binary);
+}
+
+RedisDatabase::Util::~Util() {
+    file_.close();
+}
```
PragmaTwice:
```suggestion
```

it is useless

src/redis_rdb.h
```diff
@@ -0,0 +1,82 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#define RDB_VERSION 6
+#define SELECT_DB 0
```
PragmaTwice: I prefer a more concrete name here.

git-hulk:
@ColinChamber Can you help to add more background information?

src/redis_rdb.h
ColinChamber: Thank you. I use `Composer` to replace `Util`. Do you think that's ok?

src/redis_rdb.cc
ColinChamber: Thank you.Do you think it's OK to throw an exception?

src/redis_rdb.cc
ColinChamber: According to the kvrocks metadata size(4byte)  design, this situation does not exist.

src/redis_cmd.cc
```diff
@@ -3578,6 +3579,18 @@ class CommandBGSave: public Commander {
   }
 };
 
+class CommandSave: public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    Redis::RedisDatabase RDB(svr->storage_, conn->GetNamespace());
```
torwig, line 3585: I'd prefer `rdb` instead of `RDB` variable name

src/redis_rdb.h
ColinChamber: May I ask for a suggested name?

src/redis_rdb.h
```diff
@@ -0,0 +1,57 @@
...
+
+#pragma once
+
+#include <fstream>
+#include <string>
+
+#include "redis_db.h"
+
+namespace Redis {
+class RedisDatabase : public Database {
```
torwig, line 29: @ColinChamber I think you can rename `RedisDatabase` to something like `DumpCreator` and move all the functionality from `Composer` to `DumpCreator`. Thus, you won't have `RedisDatabase` with just 1 method and inside it one private class and such long things like `RedisDatabase::Composer::MethodName()`. What do you think?

src/redis_rdb.cc
```diff
@@ -0,0 +1,450 @@
...
+
+const int RDB_VERSION = 6;
+// always use the redis logical database 0
+const int SELECT_DB = 0;
+const int RDB_6BITLEN = 0;
+const int RDB_14BITLEN = 1;
+const int RDB_32BITLEN = 0x80;
+
+enum RDBOpCode: unsigned char {
+    RDB_OPCODE_EXPIRETIME = 253,  /* Expire time in seconds. */
```
torwig, line 44: @ColinChamber Additionally you can have a look at this commit (https://github.com/apache/incubator-kvrocks/commit/1dcf225acad001175df0ea142b5e498f29d1b18b) and pay attention to `const` and `enum` naming to make relevant changes in your PR.

src/redis_rdb.h
ColinChamber: Is the class name `DumpCreator` ok if the `parser` is added in the future?

torwig: Do you mean the parser for importing the dump?

ColinChamber: Parse the Redis dump file. I know few people need it, but if someone wants to achieve it.

torwig: So you think about the `RedisDatabase` class and its inner classes, for example, `DumpCreator` and `DumpLoader`/`DumpImporter`/etc. ?

src/redis_rdb.cc
PragmaTwice: I think you can use `uint32_t` if it will not happen.

tests/gocase/integration/save_test.go
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package integration
```
tanruixiang, line 20:
```suggestion
package save
```

tanruixiang (changes requested):
Hi. I found a simple package name typo.

tisonkun (changes requested):
I'd prefer you create an issue or at least update the PR description about why you made this change. Otherwise, we review the implementation details but lose the aim of doing all this stuff.

PragmaTwice:
> I'd prefer you create an issue or at least update the PR description about why you made this change. Otherwise, we review the implementation details but lose the aim of doing all this stuff.

There is an issue #1001, although it has not been linked to this PR.

tisonkun:
@PragmaTwice Thank you! I'll check it.

ColinChamber:
There's a little tricky to dump RDB in Kvrocks. My current test for this PR is far from enough. I'm sorry that I should close it first. Maybe I will resubmit after I understand every detail. Thanks to every reviewer.

PR #961, opened 2022-10-08
Implement the GETEX command

### LINKED ISSUE #960: Support `GetEX` command
### Search before asking

### Motivation

Redis supports the `GetEX` command since 6.2.0. We can follow-up as well since this command can benefit users if they want to extend or reduce the expire time when touching.

### Solution

Implement the command described in: https://redis.io/commands/getex/

maochongxin:
This closes #960 

Has passed the [string.tcl](https://github.com/redis/redis/blob/unstable/tests/unit/type/string.tcl#L105) test 

```
➜  redis git:(6.2) ✗ tclsh tests/test_helper.tcl --host 0.0.0.0 --port 6666
Cleanup: may take some time... OK
Starting test server at port 21079
[ready]: 37258
Testing unit/type/string
[ok]: SET and GET an item
[ok]: SET and GET an empty item
[ok]: Very big payload in GET/SET
[ok]: Very big payload random access
[ok]: SET 10000 numeric keys and access all them in reverse order
[ok]: SETNX target key missing
[ok]: SETNX target key exists
[ok]: SETNX against not-expired volatile key
[ok]: SETNX against expired volatile key
[ok]: GETEX EX option
[ok]: GETEX PX option
[ok]: GETEX EXAT option
[ok]: GETEX PXAT option
[ok]: GETEX PERSIST option
[ok]: GETEX no option
[ok]: GETEX syntax errors
[ok]: GETEX no arguments
[ok]: GETDEL command
```


git-hulk:
cool, thanks for your contribution.

@maochongxin Can we put the ttl argument parse into a single function? so that we can avoid duplicate codes, the expire argument can be translated into ttl as well.

maochongxin:
> cool, thanks for your contribution.
> 
> @maochongxin Can we put the ttl argument parse into a single function? so that we can avoid duplicate codes, the expire argument can be translated into ttl as well.

Of course, I'm going to optimize it

maochongxin:
> cool, thanks for your contribution.
> 
> @maochongxin Can we put the ttl argument parse into a single function? so that we can avoid duplicate codes, the expire argument can be translated into ttl as well.

@git-hulk Complete the refactoring, but Is it waiting for #963  conclusion

PragmaTwice:
```
SET key value [NX | XX] [GET] [EX seconds | PX milliseconds |
  EXAT unix-time-seconds | PXAT unix-time-milliseconds | KEEPTTL]

GETEX key [EX seconds | PX milliseconds | EXAT unix-time-seconds |
  PXAT unix-time-milliseconds | PERSIST]
```

I think we should notice their difference: there is no `PERSIST` in `SET`, and no `NX`, `XX` in `GETEX`, hence we should give a syntax error to user e.g. while encountering `NX` in `GETEX`.

So I do not think the whole parsing procedure for `SET` and `GETEX` can be put in one function (or you can try it, but I think it might make the logic more complicated), you could just unify the EX/PX/EXAT/PXAT part.

src/redis_cmd.cc
```diff
@@ -318,6 +380,48 @@ class CommandGet : public Commander {
   }
 };
 
+class CommandGetEx : public Commander {
+public:
+    Status Parse(const std::vector<std::string> &args) override {
+        if (std::find_if(args.begin(), args.end(), [](const std::string& str) -> bool {
```
PragmaTwice: I think it is more like a workaround rather than a good solution, as you first check whether there are non-exist options and then use  ParseTtlHelper which is designed for SET.

src/redis_cmd.cc
```diff
@@ -318,6 +380,48 @@ class CommandGet : public Commander {
...
+            return Util::ToLower(str) == "persist";
```
PragmaTwice: `ToLower` will produce a new string copy, which is actually unnecessary for case-insensitive string comparison.

PragmaTwice: You can define a new util function for case-insensitive string comparison.

src/redis_cmd.cc
```diff
@@ -97,6 +97,68 @@ AuthResult AuthenticateUser(Connection *conn, Config* config, const std::string&
   return AuthResult::OK;
 }
 
+Status ParseTtlHelper(const std::vector<std::string> &args, int *result) {
```
PragmaTwice: I think something like `ParseTTL` is fine for the function name.

src/redis_cmd.cc
PragmaTwice: And I think this function is hard to maintain. You can imagine that we need to add a new parameter for SET, then we need to also modify GetEX, which is weird.

maochongxin: I think `PERSIST` is a special case, because it is mutually exclusive with the other ttl arguments, in addition, other cases are the intersection of `Set` and `Getex` commands, the parsing of parameters such as ex, px, etc.

> I think it is more like a workaround rather than a good solution, as you first check whether there are non-exist options and then use ParseTtlHelper which is designed for SET.

maochongxin: > And I think this function is hard to maintain. You can imagine that we need to add a new parameter for SET, then we need to also modify GetEX, which is weird.

Yeah, need to think about how to make it more versatile

maochongxin: @PragmaTwice  Hi, I thought of adding a parameter to decouple the redis command and the parse function, I think it is more maintainable than before, do you have any ideas about that

src/util.cc
```diff
@@ -368,6 +368,15 @@ std::string ToLower(std::string in) {
   return in;
 }
 
+bool CompareString(const std::string& lhs, const std::string& rhs, bool case_insensitive) {
```
PragmaTwice: I think we just care about case-insensitive situation is fine, we do not need the function while case-sensitive, and it has a worse performance since it has a runtime `case_insensitive` checking.

PragmaTwice: I think we just care about case-insensitive situation is fine, we do not need the function while case-sensitive, and it has a worse performance since it has a runtime `case_insensitive` checking.

src/redis_cmd.cc
```diff
@@ -97,6 +97,68 @@ AuthResult AuthenticateUser(Connection *conn, Config* config, const std::string&
...
+            }
+            ttl_ms = *parse_result;
+            if (ttl_ms <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
+            if (ttl_ms > 0 && ttl_ms < 1000) {
+                ttl = 1;  // round up the pttl to second
+            } else {
+                ttl = static_cast<int>(ttl_ms / 1000);
+            }
+        } else if (white_list.find(opt) != white_list.end()) {
+            // pass
```
PragmaTwice: If you want to use unordered_map, you can set `iter->second` to `true` here, and remove those `for_eash` in `CommandGetEX::Parse` and `CommandSet::Parse`.

maochongxin: Good idea, thanks

src/util.cc
```diff
@@ -368,9 +368,9 @@ std::string ToLower(std::string in) {
...
 
-bool CompareString(const std::string& lhs, const std::string& rhs, bool case_insensitive) {
-    return lhs.size() == rhs.size() && std::equal(lhs.begin(), lhs.end(), rhs.begin(),[case_insensitive](const char& l, const char& r) {
-        if (l == r || case_insensitive && std::tolower(l) == std::tolower(r)) {
+bool CaseInsensitiveCompare(const std::string& lhs, const std::string& rhs) {
+    return lhs.size() == rhs.size() && std::equal(lhs.begin(), lhs.end(), rhs.begin(),[](const char& l, const char& r) {
+        if (l == r || std::tolower(l) == std::tolower(r)) {
             return true;
         }
         return false;
```
```suggestion
        return std::tolower(l) == std::tolower(r);
```

src/redis_cmd.cc
```diff
@@ -143,8 +143,8 @@ Status ParseTTL(const std::vector<std::string> &args, const std::unordered_map<s
             } else {
                 ttl = static_cast<int>(ttl_ms / 1000);
             }
-        } else if (white_list.find(opt) != white_list.end()) {
-            // pass
+        } else if (white_list->find(opt) != white_list->end()) {
+            (*white_list)[opt] = true;
```
PragmaTwice: You can reuse the iterator rather than find the element again.

src/util.cc
maochongxin: but is it possible to reduce function calls here?

PragmaTwice: I think the probability of `l != r` is higher than `l == r` for normal string, so the current code may walk more branches, which is bad for pipelining and branch prediction. And, unless an obvious benchmark can be provided, I prefer simpler code.

src/redis_cmd.cc
```diff
@@ -97,6 +97,68 @@ AuthResult AuthenticateUser(Connection *conn, Config* config, const std::string&
...
+            if (ttl_ms > 0 && ttl_ms < 1000) {
+                ttl = 1;  // round up the pttl to second
+            } else {
+                ttl = static_cast<int>(ttl_ms / 1000);
+            }
+        } else if (white_list->find(opt) != white_list->end()) {
+            (*white_list)[opt] = true;
+        } else {
+            return Status(Status::NotOK, errInvalidSyntax);
+        }
```
```suggestion
        } else {
          auto iter = white_list->find(opt);
          if (iter != white_list->end()) {
              iter->second = true;
          } else {
              return Status(Status::NotOK, errInvalidSyntax);
          }
        }
```

src/util.cc
```diff
@@ -368,6 +368,12 @@ std::string ToLower(std::string in) {
   return in;
 }
 
+bool CaseInsensitiveCompare(const std::string& lhs, const std::string& rhs) {
+  return lhs.size() == rhs.size() && std::equal(lhs.begin(), lhs.end(), rhs.begin(), [](const char& l, const char& r) {
```
```suggestion
  return lhs.size() == rhs.size() && std::equal(lhs.begin(), lhs.end(), rhs.begin(), [](char l, char r) {
```

src/redis_cmd.cc
```diff
@@ -318,6 +386,50 @@ class CommandGet : public Commander {
   }
 };
 
+class CommandGetEx : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    white_list_register();
```
```suggestion
    white_list_ = {{"persist", false}};
```

src/redis_cmd.cc
```diff
@@ -318,6 +386,50 @@ class CommandGet : public Commander {
...
+    }
+    *output = s.IsNotFound() ? Redis::NilString() : Redis::BulkString(value);
+    return Status::OK();
+  }
+
+ private:
+  void white_list_register() {
+    white_list_["persist"] = false;
+  }
+
```
```suggestion
```

src/redis_cmd.cc
```diff
@@ -471,86 +583,37 @@ class CommandAppend: public Commander {
...
-        if (ttl_ms <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
-        if (ttl_ms > 0 && ttl_ms < 1000) {
-          ttl_ = 1;  // round up the pttl to second
-        } else {
-          ttl_ = static_cast<int>(ttl_ms/1000);
-        }
-      } else {
-        return Status(Status::NotOK, errInvalidSyntax);
-      }
+    white_list_register();
```
```suggestion
    white_list_ = {{"nx", false}, {"xx", false}};
```

src/redis_cmd.cc
```diff
@@ -559,10 +622,14 @@ class CommandSet : public Commander {
...
 
  private:
-  bool xx_ = false;
-  bool nx_ = false;
+  void white_list_register() {
+    white_list_["nx"] = false;
+    white_list_["xx"] = false;
+  }
+
+ private:
```
```suggestion
```

git-hulk (dismissed):

PragmaTwice (commented):
It seems these TCL tests are migrated to golang, so maybe we should add these GetEX tests in go test cases again. cc @maochongxin

maochongxin:
> It seems these TCL tests are migrated to golang, so maybe we should add these GetEX tests in go test cases again. cc @maochongxin

Ok, let me add it

maochongxin:
> > It seems these TCL tests are migrated to golang, so maybe we should add these GetEX tests in go test cases again. cc @maochongxin
> 
> Ok, let me add it

done

PragmaTwice (approved):
LGTM, thanks for your contriubution!

PragmaTwice:
Just for record: https://github.com/apache/incubator-kvrocks/actions/runs/3231715201/jobs/5291557196
```
FAIL: TestString/Extended_SET_EXAT_option
```

git-hulk (approved):

git-hulk:
Thanks all, merging...

git-hulk:
submitted the doc updated PR: https://github.com/apache/incubator-kvrocks-website/pull/19

PR #963, opened 2022-10-08
(WIP) Add parse util function(Waiting for parsing framework design)

### LINKED ISSUE #962: Align the expire option behavior of redis
### Search before asking

### Version

unstable branch.

### Minimal reproduce step

```
127.0.0.1:6666> cas key old new ex 1 ex 10
(integer) 0
```

(error) ERR syntax error

### What did you see instead?

Executed successfully.

### Anything Else?

Maybe other commands have this problem too. I'll fix them after double-checking. Even if the command was coded to prevent this issue (such as CommandSet), the code was written with no generality. I will provide a more general parsing function in this pr.

tanruixiang:
For commands with many variable parameters, the current parsing method is still relatively rigid.  In order to uniformly parse this multi-optional parameter type command, a util function is added.
```
Status ParseCommandSyntax(const std::vector<std::string> &args,
                            const std::unordered_map<std::string, int> &key_words,
                            std::unordered_map<std::string, std::vector<std::string>> *result)
```
`args` stores the command that needs to be parsed. 

`key_words` indicates the keyword need to parse the command and the number of parameters required. For example, the number `ex` needs to parse is 2: the ex and the `ex's value`.

`result`  stores the value parsed by the `key_words`. For example, what `ex` parses is an array with a length of 1 and the array's only value is `ex's value`.

it close #962

### DIFF
--- a/src/redis_cmd.cc
+++ b/src/redis_cmd.cc
@@ -31,6 +31,8 @@
 #include <vector>
 #include <thread>
 #include <utility>
+#include <string>
+#include <unordered_map>
 
 #include "cluster.h"
 #include "log_collector.h"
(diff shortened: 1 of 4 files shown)


tanruixiang:
@git-hulk @PragmaTwice @torwig @caipengbo Do you have any better ideas? After discussing a suitable method, I will follow up the replacement work.

git-hulk:
hmmm... It looks NOT generalized enough, I think the main is that has no good way to map the receiver name and type, so we can't know how to parse the field value.

tanruixiang:
> hmmm... It looks NOT generalized enough, I think the main is that has no good way to map the receiver name and type, so we can't know how to parse the field value.

The main reason is that different command keywords and analysis methods are also different, so it is difficult to unify. Is there any more elegant way to do it?

PragmaTwice:
I think the function generally looks fine, but it seems we now need a well-designed parsing framework rather than some non-generic util functions now.

I am making some attempts at this, although I may not be able to devote much time to it these days.

tanruixiang:
> I think the function generally looks fine, but it seems we now need a well-designed parsing framework rather than some non-generic util functions now.
> 
> I am making some attempts at this, although I may not be able to devote particularly much time to it these days.

I agree. If we need to implement a very generalized function, a single util function is not enough. 

Maybe you can split your work into some issues for everyone to implement together?

PragmaTwice:
> > I think the function generally looks fine, but it seems we now need a well-designed parsing framework rather than some non-generic util functions now.
> > I am making some attempts at this, although I may not be able to devote particularly much time to it these days.
> 
> I agree. If we need to implement a very generalized function, a single util function is not enough.
> 
> Maybe you can split your work into some issues for everyone to implement together?

Of cause, if everything goes well I will come up with some design ideas and plan out some implementation tasks in the next week or so.

maochongxin:
I think it can be roughly classified into several types of parameters. Can it be classified by some features such as SFINAE?

PragmaTwice:
Another concern is that, there are so many unnecessary string copy in both current code and this PR. Maybe we need to find a way to reduce them, and take advantages of move semantics of `std::string` and usage of `std::string_view`.

caipengbo:
One idea I have is that we implement syntactic state machine for each command, and then use a utility function to parse the state machine.

tisonkun:
Since the code base changes a lot, I suggest you reimplement this patch if you're still interested on it.

Closing...

Issue #884, opened 2022-09-18
Add redis-benchmark to CI


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

We can design a phase in CI (triggerred by PR or daily) to automatically run redis-benchmark to test the stability and performance of kvrocks under a certain commit, and report it while an evident performance downgrade or crash occurs in the benchmark phase.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

aleksraiden:
I thinked about this feature. Very cool, we need this tests.

IMHO, we can add to x.py file something like this:

`./x.py benchmark --bench-path=` 

with default value of client/request etc. 

In CI script, we change step "Install redis" and copy redis-benchmark into  $HOME/local/bin/.

Then we use -q options and run benchmark in quite format, just output req/sec. 

In a first stage this numbers only for view, but maybe we can store a some historical tests results into file and after bench are running, check deviations. 

So, I'am not a excellent C/Python/CI dev, but I can try to realize this step-by-step

PragmaTwice:
Thanks for your interest, please go ahead!

aleksraiden:
Thanks, gets to work

thomas-0816:
I used "apk add redis" within the standard-container (apache/kvrocks:2.8.0) and Ryzen7 PRO 7840U.

benchmark command:
- redis-benchmark -r 10000 -c 64 --threads 8 -p 6379 -t set -n 10000000 -d 64

Redis 7.0.15 (default config):
- "io-threads 1": rps=159115.5
- "io-threads 2": rps=210508.0
- "io-threads 4": rps=249928.0
- "io-threads 8": rps=248720.0
- "io-threads 12": rps=166864.0
- "io-threads 16": rps=81694.4

kvrocks v2.8
- "worker 1": rps=80468.0
- "worker 2": rps=151312.0
- "worker 4": rps=223028.0
- "worker 8": rps=265768.9
- "worker 12": rps=206741.0
- "worker 16": rps=188637.5

Quite impressive!

PR #967, opened 2022-10-09
Add redis-benchmark to CI

aleksraiden:
Add redis-benchmark and check it after all test passed. Draft #884

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -154,6 +154,7 @@ jobs:
         with:
           path: |
             ~/local/bin/redis-cli
+            ~/local/bin/redis-benchmark
           key: ${{ runner.os }}-redis-cli
 
       - name: Setup macOS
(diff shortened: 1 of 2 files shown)


aleksraiden:
Something wrong in CI-mode, when I try to start default kvrocks server  from x.py

aleksraiden:
@torwig Could you please check my dirty Python and clean-up in? As I think we maybe can add all options from original redis-benchmark into options params to bench stage?

tisonkun:
@aleksraiden I suggest you take a look at how `./x.py test go` implement. From my perspective, these two commands can be very similar.

aleksraiden:
@tisonkun thanks, correct, previously I see how to add new integration test with benchmark, but I'am not so perfect Go-dev )

If a path with bench-integration test in Go is more clear, I think, @torwig can help me in this PR

tisonkun:
@aleksraiden I don't mean you should use Golang here. I mean:

```bash
./x.py bench [BUILD]             -p 6666 -q
#            ^ optional binpath  ^ rest parameters
```

tisonkun:
Closed since I don't think this functionality should go into CI tasks per PR commit.

You may think of making a perf page like https://perf.rust-lang.org/ whose source is open. Or at least running a daily job while trying to post the result instead of bury them into CI output internally.

PR #973, opened 2022-10-11
Add `UniqueFD` and `ScopeExit`

PragmaTwice:
`UniqueFD` is a wrapped file descriptor with owership management, i.e. a `unique_ptr`-version fd.

`ScopeExit` is a scope guard like `std::experimental::scope_exit` in the library fundamentals v3 from WG21 or `boost::scope_guard` in [Boost](https://www.boost.org/doc/libs/1_80_0/libs/scope_exit/doc/html/index.html).


src/util.cc
```diff
@@ -162,52 +162,59 @@ Status SockSetTcpKeepalive(int fd, int interval) {
...
     tv.tv_sec = timeout / 1000;
     tv.tv_usec = (timeout % 1000) * 1000;
     if (setsockopt(*fd, SOL_SOCKET, SO_RCVTIMEO, reinterpret_cast<char *>(&tv), sizeof(tv)) < 0) {
-      close(*fd);
-      *fd = -1;
       return Status(Status::NotOK, std::string("setsockopt failed: ") + strerror(errno));
     }
   }
+
+  exit.Disable();
```
git-hulk, line 218: Why do we need to disable the exit here? the `exit` should be released after leaving the function.

PragmaTwice: It is a successful `return` clause here, so we do not need to close the fd.

git-hulk: oh, I see.

src/redis_cmd.cc
```diff
@@ -4659,22 +4661,21 @@ class CommandFetchFile : public Commander {
...
           LOG(INFO) << "[replication] Succeed sending file " << file << " to "
                     << ip;
         } else {
           LOG(WARNING) << "[replication] Fail to send file " << file << " to "
                        << ip << ", error: " << strerror(errno);
-          close(fd);
           break;
         }
-        close(fd);
+        fd.Reset();
```
git-hulk: How about using `fd.Close()` here, it looks more intuitive than fd.Reset()

src/stats.cc
```diff
@@ -53,19 +54,20 @@ int64_t Stats::GetMemoryRSS() {
...
-  if ((fd = open(filename, O_RDONLY)) == -1) return 0;
-  if (read(fd, buf, sizeof(buf)) <= 0) {
-    close(fd);
+  auto fd = UniqueFD(open(filename, O_RDONLY));
+  if (!fd) return 0;
+  if (read(*fd, buf, sizeof(buf)) <= 0) {
     return 0;
   }
-  close(fd);
+  fd.Reset();
```
git-hulk: Same with the above

git-hulk (commented):
Overall looks good to me

src/redis_cmd.cc
```suggestion
        fd.Close();
```

src/stats.cc
PragmaTwice:
```suggestion
  fd.Close();
```

PragmaTwice: done

src/redis_cmd.cc
PragmaTwice: done

src/replication.cc
```diff
@@ -748,9 +749,9 @@ Status ReplicationThread::parallelFetchFile(const std::string &dir,
           if (!s.IsOK()) {
             return Status(Status::NotOK, "connect the server err: " + s.Msg());
           }
+          UniqueFD unique_fd{sock_fd};
           s = this->sendAuth(sock_fd);
```
torwig, line 753: I think the `this` keyword can be omitted.

torwig: My bad, I'm sorry.

tisonkun (dismissed):
LGTM. Although, as we write more and more code, it can be a good time to try modularizing source files into folders.

src/fd_util.h
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+ private:
+  int fd_;
+};
+
+
+// ref to https://en.cppreference.com/w/cpp/experimental/scope_exit
+template <typename F>
+struct ScopeExit {
```
caipengbo: Is `ScopeExit` an `fd` related utility? It's put in `fd_util.h`

PragmaTwice: Thanks. I will create a new file for it.

caipengbo: How about putting it in `util.h`?

PragmaTwice:
> LGTM. Although, as we write more and more code, it can be a good time to try modularizing source files into folders.

Good idea, I will try to do this.

src/fd_util.h
PragmaTwice: I think it is better to put in a new file. I think the current `util.h` is too clutter, and I actually have a plan to separate the functions in `util.h` into different files

PragmaTwice: A header file named `util.h` can lead to a disaster: all functions that we do not know where to put will be placed in this file, making this file cluttered and bloated. We need more specific file names.

caipengbo: That's true, `util.h` is messy. Maybe we can break it down into categories later.

caipengbo (approved):

tisonkun (approved):

git-hulk (approved):

Issue #800, opened 2022-08-30
Use clang-format and clang-tidy instead of cpplint and cppcheck


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

clang-tidy is described in #602.

clang-format is currently the de facto standard of C++ code formatting.

It has lots of benefits, like:
- currently, cpplint reports some format error, but it can hardly automatically fix the code. So we need check errors in CI, and manually fix them one by one. But in clang-format, we can format code to a consistent style with just one command. And we can even add the auto-fix phase to CI (lots of projects do so). That is a huge convenience for developers. 
- we cannot control the version of cpplint well: it has different report in different version for same code. So different report in local and CI, it is unacceptable.
- cpplint cannot handle some detail of C++ language well. clang-format uses Clang AST and every powerful compiler technology in Clang. C++ is a super complicated language, no one can handle the syntax and semantics in different C++ standard well, except compilers like GCC or Clang. It is the core reason why cpplint and cppcheck failed to become great tools (recently cppcheck supports to use libClang instead of its own toy parser and type/semantics checker).

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

tanruixiang:
By the way, I get the same result with clang14 which is the default version of macos. The latest clang is 15. If you can be sure that the result of 15 is the same, maybe there is no need to make judgments here. I think at least we can give a warning here and give the recommended clang version. Direct RuntimeError seems a bit inappropriate.@PragmaTwice 
https://github.com/apache/incubator-kvrocks/blob/unstable/x.py#L140

PragmaTwice:
> By the way, I get the same result with clang14 which is the default version of macos. The latest clang is 15. If you can be sure that the result of 15 is the same, maybe there is no need to make judgments here. I think at least we can give a warning here and give the recommended clang version. Direct RuntimeError seems a bit inappropriate.@PragmaTwice https://github.com/apache/incubator-kvrocks/blob/unstable/x.py#L140

There are some difference (may not be encountered often) in format result between different version of clang-format. So I recommand developers use the same version of clang-format as in github actions.

You can install a clang-format 12 or download from the URL provided in the message.

I think it can be a warning, but developers may become angry when they format the code but still fail in the CI :rofl:

tanruixiang:
[...]

PR #979, opened 2022-10-12
Use clang-format instead of cpplint

PragmaTwice:
It closes the clang-format part of #800.

Note: Although it is not recommanded, but developers can disable clang-format for specific code snippets via:
```
// clang-format off

code that do not want to be formatted

// clang-format on
```

------

To format code:
```
./x.py format
```
To check whether code is formatted:
```
./x.py check format
```

To use a custom clang-format:
```
./x.py format --clang-format-path <path>
```


x.py
```diff
@@ -74,7 +74,7 @@ def run_pipe(*args: str, msg: Optional[str]=None, verbose: bool=False, **kwargs:
 def find_command(command: str, msg: Optional[str]=None) -> str:
     return run_pipe("which", command, msg=msg).read().strip()
 
-def check_version(current: str, required: Tuple[int, int, int], prog_name: Optional[str] = None) -> None:
+def check_version(current: str, required: Tuple[int, int, int], prog_name: Optional[str] = None, exact: bool = False) -> None:
```
tisonkun: This parameter is not used?

x.py
```diff
@@ -262,11 +285,13 @@ def test_go(dir: str, rest: List[str]) -> None:
...
-        description="Lint code with cpplint (https://github.com/cpplint/cpplint)",
-        help="Lint code with cpplint (https://github.com/cpplint/cpplint)")
-    parser_check_cpplint.set_defaults(func=cpplint)
+    parser_check_format = parser_check_subparsers.add_parser(
+        'format',
+        description="Format source code",
+        help="Format source code")
+    parser_check_format.set_defaults(func=check_format)
+    parser_check_format.add_argument('--clang-format-path', default='clang-format', help="path of clang-format used to check source")
+    parser_check_format.add_argument('-i', default=False, action='store_true', help="format source files")
```
tisonkun: A check should not change files. Perhaps you add both `./x.py check clang-format` and `./x.py fix clang-format`.

x.py
```diff
@@ -116,14 +116,37 @@ def build(dir: str, jobs: int, ghproxy: bool, ninja: bool, unittest: bool, compi
...
+        *glob("tests/cppunit/**/*.cc", recursive=True),
+        *glob("tools/kvrocks2redis/**/*.h", recursive=True),
+        *glob("tools/kvrocks2redis/**/*.cc", recursive=True),
+    ]
+
+def check_format(clang_format_path: str, i: bool) -> None:
+    print("WARNING: We use clang-format 12 in CI,\n"
+        "  so we recommend that you also use this version locally to avoid inconsistencies.\n"
+        "  You can install it from your package manager (usually in clang-12 package)\n"
+        "  or download it from https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.1")
```
tisonkun: Shall we always show warning?

.github/workflows/kvrocks.yaml
```diff
@@ -38,27 +38,35 @@ jobs:
...
+        if: always() && steps.check-format.outcome != 'success'
+        run: |
+          ./x.py check format -i --clang-format-path clang-format-12
+          git diff -p > clang-format.patch
+          cat clang-format.patch
+      - name: Upload format patch
+        uses: actions/upload-artifact@v3
+        if: always() && steps.check-format.outcome != 'success'
+        with:
+          path: clang-format.patch
```
tisonkun: No need. I'd prefer add a CI summary or annotation to ask the user perform the format step.

* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary
* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-creating-an-annotation-for-an-error

tisonkun (changes requested):
This changes seems in a good direction. Comments inline.

x.py
PragmaTwice: oh, thanks, I will delete it.

PragmaTwice: done

x.py
PragmaTwice: done, it is `./x.py format` now.

x.py
PragmaTwice: now it will only appear when clang-format in not versioned 12

torwig:
@PragmaTwice Good job!

.github/workflows/kvrocks.yaml
PragmaTwice: I will add a summary, but I think it is useful for some circumstances. Developers do not need a clang-format 12 (code formatted by other version of clang-format may not pass the CI) since we will generate patches, which is friendly for first-time contributors.

git-hulk:
> @PragmaTwice Good job!

I also like this change, cpplint can't auto fix is too frustrate.

git-hulk:
@PragmaTwice Can you help to rebase those commits, I don't know which commit should I start looking.

.github/workflows/kvrocks.yaml
tisonkun: Yep. I noticed that you upload a patch file instead of pushing a commit. Then it's great to contain!

.clang-format
```diff
@@ -0,0 +1,200 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
```
tisonkun, line 3: Generated? Could you share how you gain this config file?

PragmaTwice:
> @PragmaTwice Can you help to rebase those commits, I don't know which commit should I start looking.

You can just review three file: `x.py`, `.github/workflows/kvrocks.yml` and `.clang-format`, all of other file changes are made by clang-format.

.clang-format
PragmaTwice: 1. `cd` to a non-project directory
2. `clang-format-12 -style=google -dump-config > .clang-format`
3. change `ColumnLimit` to 120

torwig:
@PragmaTwice You can squash a bunch of `fix` commits so they won't be merged in the `unstable` branch in the current representation :)

x.py
```diff
@@ -116,14 +119,43 @@ def build(dir: str, jobs: int, ghproxy: bool, ninja: bool, unittest: bool, compi
...
+    version_res = run_pipe(command, '--version').read().strip()
+    version_str = re.search(r'version\s+((?:\w|\.)+)', version_res).group(1)
+
+    version = check_version(version_str, CLANG_FORMAT_REQUIRED_VERSION, "clang-format")
+
+    if version[0] > 12:
+        raise RuntimeError("We use clang-format 12 in CI,\n"
+            "  so we recommend that you also use this version locally to avoid inconsistencies.\n"
+            "  You can install it from your package manager (usually in clang-12 package)\n"
+            "  or download it from https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.1")
```
tisonkun:
```suggestion
        raise RuntimeError("We use clang-format 12 in CI,"
            "  so we recommend that you also use this version locally to avoid inconsistencies."
            "  You can install it from your package manager (usually in clang-12 package)"
            "  or download it from https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.1")
```

I suggest not to manually wrap lines.

x.py
```diff
@@ -116,14 +119,43 @@ def build(dir: str, jobs: int, ghproxy: bool, ninja: bool, unittest: bool, compi
...
+    return [
+        *glob("src/**/*.h", recursive=True),
+        *glob("src/**/*.cc", recursive=True),
+        *glob("tests/cppunit/**/*.h", recursive=True),
+        *glob("tests/cppunit/**/*.cc", recursive=True),
+        *glob("tools/kvrocks2redis/**/*.h", recursive=True),
+        *glob("tools/kvrocks2redis/**/*.cc", recursive=True),
+    ]
+
+def check_format(clang_format_path: str, fix: bool = False) -> None:
```
tisonkun:
```suggestion
def clang_format(clang_format_path: str, fix: bool = False) -> None:
```

... also its references.

tisonkun (commented):
Thanks for your updates! Two comments inline.

.clang-format
tisonkun: I'd prefer `ColumnLimit` to 200...Nowadays display device is far wider than 80 or 120 already, and wrap function calls or string literals are really clumsy.

tisonkun:
> @PragmaTwice You can squash a bunch of fix commits so they won't be merged in the unstable branch in the current representation :)

@torwig I think we're using squash and merge here :)

x.py
```diff
@@ -256,17 +288,25 @@ def test_go(dir: str, rest: List[str]) -> None:
...
     parser_check_subparsers = parser_check.add_subparsers()
-    parser_check_cpplint = parser_check_subparsers.add_parser(
-        'cpplint',
-        description="Lint code with cpplint (https://github.com/cpplint/cpplint)",
-        help="Lint code with cpplint (https://github.com/cpplint/cpplint)")
-    parser_check_cpplint.set_defaults(func=cpplint)
+    parser_check_format = parser_check_subparsers.add_parser(
+        'format',
+        description="Check source format by clang-tidy",
+        help="Check source format by clang-tidy")
```
tisonkun: I think `clang-tidy` here should be `clang-format`

torwig:
@tisonkun Perhaps I'm too old, but I do "squashing" by myself :)

x.py
PragmaTwice: done

x.py
PragmaTwice: done

x.py
PragmaTwice: done

.clang-format
PragmaTwice: The old limit is 120, so I just inherit it. I think maybe 200 is not friendly to my small 1080p laptop 🤣 , but I am open for it. What do other guys think about it? cc @git-hulk @torwig @caipengbo @ShooterIT

caipengbo: I also think 120 is enough, we should rarely have too long lines?

torwig: I think 120 is a good decision. Of course, we can use even 200, but 120 makes developers think if the length of the line is more than 120 characters.

tanruixiang: I also think 120 is enough.

tisonkun (dismissed):

tisonkun (approved):

git-hulk (approved):

tanruixiang (approved):

caipengbo (approved):

PR #981, opened 2022-10-13
Replace constant macros to constexpr variables and enums in headers

PragmaTwice:
Since macros have some defect because only the text replacement is performed, I think we can encourage developers to use more constexpr variables, function templates and enums while it is possible.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -41,7 +41,7 @@ class Storage;
 
 constexpr const uint16_t PORT_LIMIT = 65535;
 
-enum SupervisedMode { SUPERVISED_NONE = 0, SUPERVISED_AUTODETECT, SUPERVISED_SYSTEMD, SUPERVISED_UPSTART };
+enum SupervisedMode { kSupervisedNone = 0, kSupervisedAutoDetect, kSupervisedSystemd, kSupervisedUpStart };
 
 constexpr const char *TLS_AUTH_CLIENTS_NO = "no";
 constexpr const char *TLS_AUTH_CLIENTS_OPTIONAL = "optional";
@@ -90,7 +90,7 @@ struct Config {
   int slowlog_log_slower_than = 100000;
   int slowlog_max_len = 128;
   bool daemonize = false;
-  int supervised_mode = SUPERVISED_NONE;
+  int supervised_mode = kSupervisedNone;
   bool slave_readonly = true;
   bool slave_serve_stale_data = true;
   bool slave_empty_db_before_fullsync = false;
--- a/src/config.h
+++ b/src/config.h
@@ -41,7 +41,7 @@ class Storage;
 
 constexpr const uint16_t PORT_LIMIT = 65535;
 
-enum { SUPERVISED_NONE = 0, SUPERVISED_AUTODETECT, SUPERVISED_SYSTEMD, SUPERVISED_UPSTART };
+enum SupervisedMode { SUPERVISED_NONE = 0, SUPERVISED_AUTODETECT, SUPERVISED_SYSTEMD, SUPERVISED_UPSTART };
 
 constexpr const char *TLS_AUTH_CLIENTS_NO = "no";
 constexpr const char *TLS_AUTH_CLIENTS_OPTIONAL = "optional";
--- a/src/config.h
+++ b/src/config.h
@@ -38,20 +38,18 @@ class Server;
 namespace Engine {
 class Storage;
 }
-#define PORT_LIMIT 65535
 
-#define SUPERVISED_NONE 0
-#define SUPERVISED_AUTODETECT 1
-#define SUPERVISED_SYSTEMD 2
-#define SUPERVISED_UPSTART 3
+constexpr const uint16_t PORT_LIMIT = 65535;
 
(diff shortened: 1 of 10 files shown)


tanruixiang:
Hi. Is it better to use `constexpr` here too?
https://github.com/apache/incubator-kvrocks/blob/unstable/src/redis_cmd.cc#L63

src/config.h
```diff
@@ -38,20 +38,18 @@ class Server;
...
 
-#define SUPERVISED_NONE 0
-#define SUPERVISED_AUTODETECT 1
-#define SUPERVISED_SYSTEMD 2
-#define SUPERVISED_UPSTART 3
+constexpr const uint16_t PORT_LIMIT = 65535;
 
-#define TLS_AUTH_CLIENTS_NO "no"
-#define TLS_AUTH_CLIENTS_OPTIONAL "optional"
+enum { SUPERVISED_NONE = 0, SUPERVISED_AUTODETECT, SUPERVISED_SYSTEMD, SUPERVISED_UPSTART };
```
torwig: Would you like to give a name to this enum?

src/redis_zset.h
```diff
@@ -76,11 +76,13 @@ typedef struct {
   double score;
 } MemberScore;
 
-#define ZSET_INCR 1
-#define ZSET_NX (1 << 1)
-#define ZSET_XX (1 << 2)
-#define ZSET_REVERSED (1 << 3)
-#define ZSET_REMOVED 1 << 4
+enum {
```
torwig: And to give a name for `enum` here to add some extra meaning to these values?

src/config.h
PragmaTwice: done

src/redis_zset.h
PragmaTwice: done

PragmaTwice:
> Hi. Is it better to use `constexpr` here too? https://github.com/apache/incubator-kvrocks/blob/unstable/src/redis_cmd.cc#L63

Sure. I think at least it should be typed `const char * const` (if no `constexpr`), since the top-level `const` qualifier matters for constants (in linkage and storage durations).

src/config.h
```diff
@@ -38,20 +38,18 @@ class Server;
...
 
-#define SUPERVISED_NONE 0
-#define SUPERVISED_AUTODETECT 1
-#define SUPERVISED_SYSTEMD 2
-#define SUPERVISED_UPSTART 3
+constexpr const uint16_t PORT_LIMIT = 65535;
 
-#define TLS_AUTH_CLIENTS_NO "no"
-#define TLS_AUTH_CLIENTS_OPTIONAL "optional"
+enum SupervisedMode { SUPERVISED_NONE = 0, SUPERVISED_AUTODETECT, SUPERVISED_SYSTEMD, SUPERVISED_UPSTART };
```
caipengbo: Now that enum has a name, can the prefix of enum values be removed?

src/redis_zset.h
```diff
@@ -76,11 +76,13 @@ typedef struct {
...
-#define ZSET_XX (1 << 2)
-#define ZSET_REVERSED (1 << 3)
-#define ZSET_REMOVED 1 << 4
+enum ZSetFlags {
+  ZSET_INCR = 1,
+  ZSET_NX = 1 << 1,
+  ZSET_XX = 1 << 2,
+  ZSET_REVERSED = 1 << 3,
+  ZSET_REMOVED = 1 << 4,
+};
```
caipengbo, line 86: Can `ZSET_` be removed?

src/config.h
PragmaTwice: I think we cannot remove it. Since it is not an `enum class`, the names in the `enum` will be directly exported to the outer namespace, so if we remove the prefix, we will get some name like `NONE` in the global namespace, which will be a mess.

PragmaTwice: And the reason I do not use `enum class` here is that `enum class` cannot be implicitly casted to integer, so it will involve more code changes.

caipengbo: Ok, got it!

tisonkun (dismissed):

src/redis_zset.h
```diff
@@ -76,11 +76,13 @@ typedef struct {
   double score;
 } MemberScore;
 
-#define ZSET_INCR 1
-#define ZSET_NX (1 << 1)
-#define ZSET_XX (1 << 2)
-#define ZSET_REVERSED (1 << 3)
-#define ZSET_REMOVED 1 << 4
+enum ZSetFlags {
```
torwig, line 79: What do you think about the naming of enumerating options? Because they aren't macro definitions now, maybe they could have naming like `MigrateTaskState` (camel case with the `k` prefix as described here https://google.github.io/styleguide/cppguide.html#Enumerator_Names)?

PragmaTwice: done

torwig (approved):
LGTM

caipengbo (approved):

tisonkun (approved):

PragmaTwice:
Thanks all. Merging...

PR #991, opened 2022-10-14
Change server redis version string

juancresc:
Hello! we are trying to use the Redis spark connector and found this possible issue:
Here, the Redis version is reported as version:
https://github.com/apache/incubator-kvrocks/blob/81d37927dfb2798138c8569af7c6a44dd376c3f4/src/server.cc#L783
But the Redis client expects a redis_version instead
https://github.com/RedisLabs/spark-redis/blob/master/src/main/scala/com/redislabs/provider/redis/RedisConfig.scala#L229
Apparently, for that reason, it is not able to connect.

### DIFF
--- a/src/server.cc
+++ b/src/server.cc
@@ -780,7 +780,7 @@ void Server::GetServerInfo(std::string *info) {
   }
   time(&now);
   string_stream << "# Server\r\n";
-  string_stream << "version:" << VERSION << "\r\n";
+  string_stream << "redis_version:" << VERSION << "\r\n";
   string_stream << "git_sha1:" << GIT_COMMIT << "\r\n";
   string_stream << "os:" << name.sysname << " " << name.release << " " << name.machine << "\r\n";
 #ifdef __GNUC__


src/server.cc
```diff
@@ -780,7 +780,7 @@ void Server::GetServerInfo(std::string *info) {
   }
   time(&now);
   string_stream << "# Server\r\n";
-  string_stream << "version:" << VERSION << "\r\n";
+  string_stream << "redis_version:" << VERSION << "\r\n";
```
torwig, line 783: I think it's **not** the version of `Redis`, it's the version of `kvrocks`.

git-hulk: Yes, Kvrocks is Redis protocol compatible but should NOT change its version to Redis directly. 

But we can discuss how to work around this scenario. From my proposition, I think you can fork and modify your repository and then have a test. If all go smoothly, then we can see if possible to compatible this on `spark-redis`. What do you think? @juancrescente

git-hulk: @tisonkun @PragmaTwice @ShooterIT Do you have comments on this case?

PragmaTwice: I think if we need a `redis_version` in server info, maybe we need to select a real redis version rather than kvrocks version. e.g.
```
redis_version: <selected compatible redis version>
kvrocks_version: 2.x.x
```

git-hulk: Good advice, we are aligning the Redis 4.0 in Redis protocol but many commands in the newer version are also supported.

caipengbo: @juancrescente  I also think you should simply modify `spark-redis`, after all, it is only an external tool.

juancresc: @caipengbo Makes sense, although I’d love this tool to be compatible for others to use with spark.  
I wonder what is the use of this version string and if the function could just return both, version (kvrocks version) and redis_version (protocol compatibility).

juancresc: @git-hulk ill give it a try to your suggestions and get back here

git-hulk: okay, thanks. @caipengbo I think it makes sense to let users know the Redis version which is compatible with Kvrocks.

tisonkun: > know the Redis version which is compatible with Kvrocks.

Yes. From my experience in TiDB, the `version()` function returns a string compatible with MySQL while `tidb_version()` returns a TiDB dedicated version string. We may print both info entry for a Redis compatible version string and kvrocks specific version string.

Reference:

```
mysql> SELECT VERSION();
+--------------------+
| VERSION()          |
+--------------------+
| 5.7.25-TiDB-v6.3.0 |
+--------------------+
1 row in set (0.00 sec)

mysql> SELECT TIDB_VERSION();
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| TIDB_VERSION()                                                                                                                                                                                                                                                                                   |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Release Version: v6.3.0
Edition: Community
Git Commit Hash: ecd67531f1721d3e49eb15a202ac7c0ae02291ec
Git Branch: heads/refs/tags/v6.3.0
UTC Build Time: 2022-09-23 14:21:08
GoVersion: go1.19.1
Race Enabled: false
TiKV Min Version: 6.2.0-alpha
Check Table Before Drop: false
Store: unistore |
+--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0.01 sec)
```

tisonkun:
Closed as invalid.

Feel free to open an issue if there's still issue stopping you from using Redis Spark Connector.

xiaobiaozhao:
please assign to me

tisonkun:
@xiaobiaozhao I don't think we have something to do here. You can create a new issue to description the motivation for modifying the version string and the design. We don't do thing randomly but with a real-world requirement.

juancresc:
> @xiaobiaozhao I don't think we have something to do here. You can create a new issue to description the motivation for modifying the version string and the design. We don't do thing randomly but with a real-world requirement.

I was supposed to test this with spark-Redis and comment here the outcome. Haven’t had time yet but will report the result here soon

juancresc:
@xiaobiaozhao and after discussing, the solution was not to modify the version string itself but add one more with the redis_version pointing to the supported protocol which is 4.0

xiaobiaozhao:
> and after discussing, the solution was not to modify the version string itself but add one more with the redis_version pointing to the supported protocol which is 4.0

OK，Please continue to deal with this PR

PR #1012, opened 2022-10-19
Move TCL test unit/tls to Go case

### LINKED ISSUE #995: Move TCL test unit/tls to Go case

PragmaTwice:
It closes #995.


tests/gocase/util/server.go
```diff
@@ -181,11 +238,19 @@ func StartServer(t testing.TB, configs map[string]string) *KvrocksServer {
...
+
+	proc, err := process.NewProcess(int32(cmd.Process.Pid))
+	require.NoError(t, err)
+
+	procStatus := []string{}
 	require.Eventually(t, func() bool {
 		err := c.Ping(context.Background()).Err()
-		return err == nil || err.Error() == "NOAUTH Authentication required."
+		procStatus, _ = proc.Status()
+		return err == nil || err.Error() == "NOAUTH Authentication required." || reflect.DeepEqual(procStatus, []string{process.Zombie})
```
PragmaTwice: If kvrocks aborted due to an config error or whatever (it will enter zombie state in *nix since `wait` has not been called), we can exit the ping loop quickly rather than waiting 1 minute.

tests/gocase/util/server.go
```diff
@@ -41,9 +46,11 @@ var workspace = flag.String("workspace", "", "directory of cases workspace")
...
 
 type KvrocksServer struct {
-	t    testing.TB
-	cmd  *exec.Cmd
-	addr *net.TCPAddr
+	t   testing.TB
+	cmd *exec.Cmd
+
+	addr    *net.TCPAddr
+	tlsAddr *net.TCPAddr
```
PragmaTwice, line 46: A TLS-enabled server has both a non-TLS port and a TLS port which can work at the same time, so I think we need two fields here.

tests/gocase/util/server.go
```diff
@@ -21,29 +21,39 @@ package util
...
 	"github.com/stretchr/testify/require"
 )
 
 var binPath = flag.String("binPath", "", "directory including kvrocks build files")
 var workspace = flag.String("workspace", "", "directory of cases workspace")
 var deleteOnExit = flag.Bool("deleteOnExit", false, "whether to delete workspace on exit")
+var TLSServerName = flag.String("TLSServerName", "localhost", "server name for TLS connection")
+
+var EnableTLSTests = flag.Bool("enableTLSTests", false, "enable TLS-related test cases")
+var CliPath = flag.String("cliPath", "redis-cli", "path to redis-cli")
```
PragmaTwice: If these options are in other places like `tls_test.go`, then `go test -enableTLSTests ./...` will fail for all tests except `tls_test.go`, so we should always put options to util rather than a single test file.

tisonkun (commented):
I push a few commits and notice that TLS tests may work on darwin now (work for me locally, I suspect it's about TLS server name). Turn on to verify.

tisonkun:
Emmm...Unfortunately it doesn't. Reverting...

tests/gocase/integration/cli/cli_test.go
```diff
@@ -38,14 +39,16 @@ import (
 
 const defaultByteBufLen = math.MaxUint16
 
+var cliPath = flag.String("cliPath", "redis-cli", "path to redis-cli")
```
PragmaTwice: As mentioned here https://github.com/apache/incubator-kvrocks/pull/1012#discussion_r999529162, I think `go test` will fail if `cliPath` is passed explicitly, e.g.
```
./x.py test go build -cliPath=something
```

tisonkun: Oops, I misunderstand the rule here. Reverting...

tisonkun:
```
--- FAIL: TestBitmap (403.30s)
    --- FAIL: TestBitmap/SETBIT/GETBIT/BITCOUNT/BITPOS_boundary_check_(type_bitmap) (15.33s)
        bitmap_test.go:146: 
            	Error Trace:	/Users/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/unit/type/bitmap/bitmap_test.go:146
            	Error:      	Not equal: 
            	            	expected: int(1)
            	            	actual  : int64(0)
            	Test:       	TestBitmap/SETBIT/GETBIT/BITCOUNT/BITPOS_boundary_check_(type_bitmap)
```

@PragmaTwice @git-hulk I don't know why it keeps failing...Cannot reproduce locally and should not be related to this patch (although I try to "fix" it, it fails continuously).

git-hulk:
> ow why it keeps failing...Cannot reproduce locally and should not be related to this patch (although I try to "fix" it, it fail

Yes, not related to current PR, I'm wondering if it's caused by Go parallel run tests.

tisonkun:
@git-hulk No. Golang only runs tests in parallel if you write `t.Parallel()`.

tisonkun:
> ```
> --- FAIL: TestBitmap (403.30s)
>     --- FAIL: TestBitmap/SETBIT/GETBIT/BITCOUNT/BITPOS_boundary_check_(type_bitmap) (15.33s)
>         bitmap_test.go:146: 
>             	Error Trace:	/Users/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/unit/type/bitmap/bitmap_test.go:146
>             	Error:      	Not equal: 
>             	            	expected: int(1)
>             	            	actual  : int64(0)
>             	Test:       	TestBitmap/SETBIT/GETBIT/BITCOUNT/BITPOS_boundary_check_(type_bitmap)
> ```
> 
> @PragmaTwice @git-hulk I don't know why it keeps failing...Cannot reproduce locally and should not be related to this patch (although I try to "fix" it, it fails continuously).

@vmihailenco do you have any ideas here? This test is our last test to be migrated with go-redis XD

tisonkun:
It seems that the bitmap failure does related to this patch. I'm digging into the test scripts.

See identical CI runs: https://github.com/apache/incubator-kvrocks/actions/runs/3287762862/jobs/5417324374

vmihailenco:
>@vmihailenco do you have any ideas here? 

Perhaps try `maxOffset/8 + 1`?

tisonkun:
@vmihailenco Let me try it if this run still fails. It's strange since it happens only for darwin platform.

tests/gocase/tls/tls_test.go
```diff
@@ -0,0 +1,138 @@
...
+	defer func() { require.NoError(t, rdb.Close()) }()
+
+	doWithTLSClient := func(tlsConfig *tls.Config, f func(c *redis.Client)) {
+		c := srv.NewClientWithOption(&redis.Options{TLSConfig: tlsConfig, Addr: srv.TLSAddr()})
+		defer func() { require.NoError(t, c.Close()) }()
+		f(c)
+	}
+
+	t.Run("TLS: Not accepting non-TLS connections on a TLS port", func(t *testing.T) {
+		c := srv.NewClientWithOption(&redis.Options{TLSConfig: defaultTLSConfig})
```
PragmaTwice:
```suggestion
		c := srv.NewClientWithOption(&redis.Options{Addr: srv.TLSAddr()})
```
I think the default port is a non-TLS port, and it requires a non-TLS connection so we do not need TLSConfig here.

tisonkun: Good catch.

tisonkun (approved):
Pending for merging...

tisonkun:
After I add back `-bench=.` in `x.py`, tests passed. Really weird >_<

PR #1017, opened 2022-10-20
Align the expire option behavior of redis

tanruixiang:
This refers to #962


tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -141,33 +141,107 @@ func TestString(t *testing.T) {
...
 		require.EqualValues(t, -1, rdb.TTL(ctx, "foo").Val())
+		require.ErrorContains(t, rdb.Do(ctx, "getex", "foo", "ex", 10, "persist").Err(), "syntax err")
+		require.ErrorContains(t, rdb.Do(ctx, "getex", "foo", "px", 10000, "persist").Err(), "syntax err")
+		require.ErrorContains(t, rdb.Do(ctx, "getex", "foo", "pxat", time.Now().Add(10000*time.Millisecond).UnixMilli(), "persist").Err(), "syntax err")
+		require.ErrorContains(t, rdb.Do(ctx, "getex", "foo", "exat", time.Now().Add(100*time.Second).Unix(), "persist").Err(), "syntax err")
+	})
+
+	t.Run("GETEX with incorrect use of multi options should result in syntax err", func(t *testing.T) {
+		options := []string{"px", "ex", "pxat", "exat"}
+		for i := 0; i < (1 << 4); i++ {
```
torwig: I think since it's not part of a hot path somewhere in the production code, it would be more beneficial and clear to use a more explicit integer value instead of a shift operation.
The same could be applied to the statement `i&(1<<j)` in the condition clause a few lines below and inside another loop.

tanruixiang: Hi. This is subset enumeration in binary. If you use 16 directly, it will be more difficult to understand. If this confuses people, I'll change to a way that doesn't use binary enums later.

tanruixiang: It has been solved.

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -141,33 +141,75 @@ func TestString(t *testing.T) {
...
+
 	t.Run("GETEX PERSIST option", func(t *testing.T) {
 		require.NoError(t, rdb.Del(ctx, "foo").Err())
 		require.NoError(t, rdb.Set(ctx, "foo", "bar", 10*time.Second).Err())
 		util.BetweenValues(t, rdb.TTL(ctx, "foo").Val(), 5*time.Second, 10*time.Second)
 		require.NoError(t, rdb.Do(ctx, "getex", "foo", "persist").Err())
 		require.EqualValues(t, -1, rdb.TTL(ctx, "foo").Val())
+		require.ErrorContains(t, rdb.Do(ctx, "getex", "foo", "ex", 10, "persist").Err(), "syntax err")
+		require.ErrorContains(t, rdb.Do(ctx, "getex", "foo", "px", 10000, "persist").Err(), "syntax err")
+		require.ErrorContains(t, rdb.Do(ctx, "getex", "foo", "pxat", time.Now().Add(10000*time.Millisecond).UnixMilli(), "persist").Err(), "syntax err")
```
git-hulk: I see some places are using a second unit but some are milliseconds, any difference between those cases? if no, I think we can keep the second unit which is easier for reading.

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -497,7 +573,12 @@ func TestString(t *testing.T) {
 	})
 
 	t.Run("Extended SET with incorrect use of multi options should result in syntax err", func(t *testing.T) {
-		require.ErrorContains(t, rdb.Do(ctx, "SET", "foo", "bar", "ex", "10", "px", "10000").Err(), "syntax err")
```
tisonkun, line 500: You can still retain this case? Although it's semantically equal to `px 100 ex 10`, we're testing the parsing logic here so `ex 10 px 10000` can be a different case.

src/commands/redis_cmd.cc
```diff
@@ -103,24 +104,28 @@ Status ParseTTL(const std::vector<std::string> &args, std::unordered_map<std::st
   int ttl = 0;
   int64_t expire = 0;
   bool last_arg = false;
+  int ex_exist = 0, exat_exist = 0, pxat_exist = 0, px_exist = 0;
```
tisonkun: I'd prefer `has_xxx`, but it's on personal bias.

tisonkun:
```suggestion
  bool has_ex = false, has_exat = false, has_pxat = false, has_px = false;
```

You can still use `has_ex + ... + > 1` below. I think it's an idiom in C++. cc @PragmaTwice

src/commands/redis_cmd.cc
```diff
@@ -154,6 +160,9 @@ Status ParseTTL(const std::vector<std::string> &args, std::unordered_map<std::st
       }
     }
   }
+  if (px_exist + ex_exist + exat_exist + pxat_exist >= 2) {
```
tisonkun:
```suggestion
  if (px_exist + ex_exist + exat_exist + pxat_exist > 1) {
```

nit: More than 1 means has duplicates. Also, it's my personal bias. Not a blocker from merging.

src/commands/redis_cmd.cc
```diff
@@ -858,6 +870,9 @@ class CommandCAS : public Commander {
         return Status(Status::NotOK, errInvalidSyntax);
       }
     }
+    if (ex_exist && px_exist) {
```
tisonkun:
```suggestion
    if (ex_exist + px_exist > 1) {
```

Keep patten consistent.

tisonkun (commented):
Generally looks good. Comments inline. The style comments are not required - let's find the common style preferences in the community.

tests/gocase/unit/type/strings/strings_test.go
tanruixiang: I used the same form as the previous test, the purpose of this should be to use milliseconds when using px and seconds when using ex

git-hulk: It's not different for `time.Now.Add()`, so I think it'd be better to keep the second unit.

tanruixiang: Our current tests are multiples of thousands of milliseconds, and indeed all can be replaced by seconds. It has now been replaced.

tests/gocase/unit/type/strings/strings_test.go
tanruixiang: OK. Tests for different parameter orders have been added.

tisonkun (approved):

git-hulk (approved):

PR #1020, opened 2022-10-20
Fix don't duplicate killed clients which have already flagged kCloseAfterReply

### LINKED ISSUE #1014: Flaky test "kill normal client"
### Search before asking

### Motivation

```
--- FAIL: TestIntrospection (3.80s)
    --- FAIL: TestIntrospection/Kill_normal_client (0.00s)
        introspection_test.go:120: 
            	Error Trace:	/home/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/unit/introspection/introspection_test.go:120
            	Error:      	Not equal: 
            	            	expected: int(1)
            	            	actual  : int64(2)
            	Test:       	TestIntrospection/Kill_normal_client
```

https://github.com/apache/incubator-kvrocks/actions/runs/3286622831/jobs/5414949088

### Solution

Perhaps we should tolerate 1 or 2.

@vmihailenco is this behavior related to go-redis auto reconnecting?

cc @git-hulk 

git-hulk:
This may close #1014

After looking through the test case "Kill normal client" many times, I can't find any clues why it fails randomly. But found it may duplicate killing and count the client if it has already flagged as `CloseAfterReply`, so I guess maybe it's one of the reasons and fix it first.

### DIFF
--- a/src/server/worker.cc
+++ b/src/server/worker.cc
@@ -430,7 +430,7 @@ void Worker::KillClient(Redis::Connection *self, uint64_t id, std::string addr, 
   for (const auto &iter : conns_) {
     Redis::Connection *conn = iter.second;
     if (skipme && self == conn) continue;
-    // no need to kill again the client if flags as kCloseAfterReply
+    // no need to kill the client again if flags as kCloseAfterReply
     if (conn->IsFlagEnabled(Redis::Connection::kCloseAfterReply)) {
       continue;
     }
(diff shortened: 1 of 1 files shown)


src/server/worker.cc
```diff
@@ -430,6 +430,10 @@ void Worker::KillClient(Redis::Connection *self, uint64_t id, std::string addr,
   for (const auto &iter : conns_) {
     Redis::Connection *conn = iter.second;
     if (skipme && self == conn) continue;
+    // no need to kill again the client if flags as kCloseAfterReply
```
tisonkun: So we kill clients asynchronously?

git-hulk: Yes, it will close the client after the socket buffer were sent out.

git-hulk:
```suggestion
    // no need to kill the client again if flags as kCloseAfterReply
```

tisonkun (approved):

PR #1022, opened 2022-10-21
Implement the ZADD options

### LINKED ISSUE #1018: Support ZADD options
### Search before asking

### Motivation

ZADD supports a list of options.
Starting with Redis version 3.0.2: Added the XX, NX, CH and INCR options.
Starting with Redis version 6.2.0: Added the GT and LT options.

I need these in my business scenario.

### Solution

https://redis.io/commands/zadd/

manchurio:
This closes #1018 

Has passed the unit test with go.


PragmaTwice:
You can just run `./x.py format` to format the source code rather than manual formatting.

PragmaTwice:
Oh, I notice the cppcheck false alarm. I think you can change `--std=c++11` to `17` here: https://github.com/apache/incubator-kvrocks/blob/unstable/x.py#L162

@manchurio

manchurio:
> You can just run `./x.py format` to format the source code rather than manual formatting.

got it,thanks.

manchurio:
> Oh, I notice the cppcheck false alarm. I think you can change `--std=c++11` to `17` here: https://github.com/apache/incubator-kvrocks/blob/unstable/x.py#L162
> 
> @manchurio
```
src/commands/redis_cmd.cc:2432:19: warning: Suspicious condition. The result of find() is an iterator, but it is not properly checked. [stlIfFind]
    if (auto it = options.find(option); it != options.end()) {
                  ^
```
It seems cppcheck  wrongly reports @PragmaTwice

PragmaTwice:
It is sad that cppcheck still fail and report a FALSE ALARM.

We will replace cppcheck by clang-tidy later, and I think you can now workaround it by `auto iter = ...; if(iter != end()) ...`

cc @git-hulk

git-hulk:
@manchurio Need to use gofmt to format the test case

src/commands/redis_cmd.cc
```diff
@@ -2385,18 +2393,74 @@ class CommandZAdd : public Commander {
...
+  constexpr unsigned max_index = 6;
+  for (unsigned i = 2; i < max_index; i++) {
+    auto option = Util::ToLower(args[i]);
+    auto it = options.find(option);
+    if (it != options.end()) {
+      flags_ |= it->second;
+      index++;
+    } else {
+      break;
+    }
```
tanruixiang:
```suggestion
    if (options.count(option)) {
      flags_ |= options[option];
      index++;
    } else {
      break;
    }
```

tanruixiang (changes requested):
If you only need to determine whether the key exists, you can use count directly, the iterator is redundant. Also doing this will not trigger an error from cppcheck. (By the way, C++20 has [contains](https://en.cppreference.com/w/cpp/container/unordered_map/contains) to do this thing. )

src/commands/redis_cmd.cc
manchurio: Not only determine whether the key exists, but also get the associated value. And in your code, it finds value twice, bad for performance.

tanruixiang: Yes. But after all, the average time complexity is O(1), and the `options` are very small, so there is basically no gap in performance. Readability is probably more important here. Of course, your implementation is fine too.

PragmaTwice:
> If you only need to determine whether the key exists, you can use count directly, the iterator is redundant. Also doing this will not trigger an error from cppcheck. (By the way, C++20 has [contains](https://en.cppreference.com/w/cpp/container/unordered_map/contains) to do this thing. )

Hi @tanruixiang, 

Actually I like the original code more, since the iterator is used in the true-branch block to avoid a double search (the upper bound is actually not O(1) due to collision, and even O(1) DOES NOT imples no cost). 

And I think this change does not improve readability, as the use of iterators is natural and easy to understand in c++. The false positives of cppcheck are entirely due to his own problems, and we should not modify the code logic just for that.

And in my opinion, I prefer just a comment rather than "request for changes" for just a trivial and purely code-smell review message, in order to be nice to new contributors.

tanruixiang:
> And in my opinion, I prefer just a comment rather than "request for changes" for just a trivial and purely code-smell review message, in order to be nice to new contributors.

Got it, thanks a lot for your suggestion. It should be that I misunderstood this "request for changes". I originally thought that only use "request for changes"  can be clicked on github to modify. (Because if someone else suggests a little modification, I prefer to click on github to complete the modification🤣)

tanruixiang:
> Actually I like the original code more, since the iterator is used in the true-branch block to avoid a double search (the upper bound is actually not O(1) due to collision, and even O(1) DOES NOT imples no cost).
> 
> And I think this change does not improve readability, as the use of iterators is natural and easy to understand in c++. The false positives of cppcheck are entirely due to his own problems, and we should not modify the code logic just for that.

Maybe it's my bad habit, I always try not to use iterators in code snippets that are similar to algorithm problems. I will try to use iterators more in the future.

PragmaTwice:
> > And in my opinion, I prefer just a comment rather than "request for changes" for just a trivial and purely code-smell review message, in order to be nice to new contributors.
> 
> Got it, thanks a lot for your suggestion. It should be that I misunderstood this "request for changes". I originally thought that only use "request for changes" can be clicked on github to modify. (Because if someone else suggests a little modification, I prefer to click on github to complete the modification🤣)

In the github review page, you can select "comment", "request for changes" or "approve". I think the first option is relatively "milder".

src/commands/redis_cmd.cc
```diff
@@ -2385,18 +2393,74 @@ class CommandZAdd : public Commander {
...
+
+  void parseOptions(const std::vector<std::string> &args, unsigned &index);
+  Status validateFlags();
+};
+
+void CommandZAdd::parseOptions(const std::vector<std::string> &args, unsigned &index) {
+  std::unordered_map<std::string, ZSetFlags> options = {{"xx", kZSetXX}, {"nx", kZSetNX}, {"ch", kZSetCH},
+                                                        {"lt", kZSetLT}, {"gt", kZSetGT}, {"incr", kZSetIncr}};
+  constexpr unsigned max_index = 6;
+  for (unsigned i = 2; i < max_index; i++) {
```
tanruixiang: Thank you very much for your contribution. There may be duplicate flags here, so the upper limit is not necessarily 6. For example, the following is the running result of redis.
```
127.0.0.1:6379> ZADD myzset 3 "test"
(integer) 1
127.0.0.1:6379>  ZADD myzset  XX XX XX XX XX XX XX INCR 3 "test"
"6"
```

manchurio: Thanks for your review.I will fix it later.

src/commands/redis_cmd.cc
```diff
@@ -2365,12 +2366,19 @@ class CommandSInterStore : public Commander {
...
-      return Status(Status::RedisParseErr, errInvalidSyntax);
+    unsigned index = 2;
+    parseOptions(args, index);
+    if (auto s = validateFlags(); !s.IsOK()) {
+      return s;
+    }
+    if (auto left = (args.size() - index); left > 0) {
+      if ((flags_ & kZSetIncr) && left != 2) {
+        return Status(Status::RedisParseErr, "INCR option supports a single increment-element pair");
+      } else if (left % 2 != 0)
```
torwig: To be consistent you should add curly braces to the statement inside `else if` clause.

src/commands/redis_cmd.cc
```diff
@@ -2385,19 +2393,74 @@ class CommandZAdd : public Commander {
...
+    if (it != options.end()) {
+      flags_ |= it->second;
+      index++;
+    } else {
+      break;
+    }
+  }
+}
+
+Status CommandZAdd::validateFlags() {
```
torwig: This function can be marked as `const`.

src/commands/redis_cmd.cc
```diff
@@ -2385,19 +2393,74 @@ class CommandZAdd : public Commander {
...
+  if (!flags_) {
+    return {};
```
torwig: I'd prefer to write `Status::OK()` explicitly. At least it's used over the codebase (consistency).
The same is at the end of this function.

src/commands/redis_cmd.cc
```diff
@@ -2385,19 +2393,74 @@ class CommandZAdd : public Commander {
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     int ret;
+    double old_score = member_scores_[0].score;
+    bool incr = (flags_ & kZSetIncr) != 0;
```
torwig: This line can be safely moved before the actual use of `incr` - before `if`-statement.

src/types/redis_zset.cc
```diff
@@ -71,19 +72,35 @@ rocksdb::Status ZSet::Add(const Slice &user_key, uint8_t flags, std::vector<Memb
     }
     added_member_keys.insert(member_key);
 
+    bool lt = (flags & kZSetLT) != 0;
+    bool gt = (flags & kZSetGT) != 0;
+
     if (metadata.size > 0) {
+      if (flags & kZSetNX) {
```
torwig: I think it's not a correct condition. Because if there is a non-empty sorted set and you try to add a new member WITH the option `NX`, `kvrocks` doesn't add this element but `Redis` does. The root cause is that if the flag `NX` is set => then for loop does nothing because every time `continue` is executed (of course, if there is at least one element in the sorted set).

For example:

`Kvrocks`
```
[yaroslav@localhost redis]$ ./src/redis-cli -p 6666
127.0.0.1:6666> zadd ss 1 a 2 b 3 c
(integer) 3
127.0.0.1:6666> zrange ss 0 100
1) "a"
2) "b"
3) "c"
127.0.0.1:6666> zadd ss NX 4 d
(integer) 0     (<-- NOT added but this element is NEW)
127.0.0.1:6666> zrange ss 0 100
1) "a"
2) "b"
3) "c"
```
Compared to `Redis`:

```
[yaroslav@localhost redis]$ ./src/redis-cli
127.0.0.1:6379> zadd ss 1 a 2 b 3 c
(integer) 3
127.0.0.1:6379> zrange ss 0 100
1) "a"
2) "b"
3) "c"
127.0.0.1:6379> zadd ss NX 4 d
(integer) 1 (<-- ADDED)
127.0.0.1:6379> zrange ss 0 100
1) "a"
2) "b"
3) "c"
4) "d"
```

torwig:
@manchurio By the way, where are the unit tests? :)

tisonkun (changes requested):
Please revert changes on `x.py`.

PragmaTwice:
The file permission of `x.py` has changed from 755 to 644, which is weird. Could you revert it?

manchurio:
> @manchurio By the way, where are the unit tests? :)

The unit tests of zadd options are  in zset_test.go

manchurio:
> Please revert changes on `x.py`.
> The file permission of `x.py` has changed from 755 to 644, which is weird. Could you revert it?

OK. I'll figure out what happened, then fix it later.

tanruixiang:
> > @manchurio By the way, where are the unit tests? :)
> 
> The unit tests of zadd options are in zset_test.go

Hi. What he meant maybe the `tests/cppunit/t_zset_test.cc`.

torwig:
@manchurio @tanruixiang  Yes, I meant the unit tests that are written in C++. Related to the sorted set unit tests are located in `tests/cppunit/t_zset_test.cc` and test only one component - `class ZSet`.
On the other hand, tests in Go/TCL test the entire `kvrocks` (storage + protocol + network).

tanruixiang:
> @manchurio @tanruixiang Yes, I meant the unit tests that are written in C++. Related to the sorted set unit tests are located in `tests/cppunit/t_zset_test.cc` and test only one component - `class ZSet`. On the other hand, tests in Go/TCL test the entire `kvrocks` (storage + protocol + network).

By the way, after everyone's efforts, there are only Go tests now.😄

manchurio:
> The file permission of `x.py` has changed from 755 to 644, which is weird. Could you revert it?

The file permission of `x.py` has changed from 644 to 755 now. @PragmaTwice  @tisonkun

src/types/redis_zset.cc
```diff
@@ -71,19 +72,35 @@ rocksdb::Status ZSet::Add(const Slice &user_key, uint8_t flags, std::vector<Memb
...
+    bool lt = (flags & kZSetLT) != 0;
+    bool gt = (flags & kZSetGT) != 0;
+
     if (metadata.size > 0) {
       std::string old_score_bytes;
       s = db_->Get(rocksdb::ReadOptions(), member_key, &old_score_bytes);
       if (!s.ok() && !s.IsNotFound()) return s;
       if (s.ok()) {
+        if (!s.IsNotFound() && (flags & kZSetNX)) {
+          continue;
```
git-hulk, line 81: Can just return instead of continuing here?

src/types/redis_zset.cc
```diff
@@ -71,19 +72,35 @@ rocksdb::Status ZSet::Add(const Slice &user_key, uint8_t flags, std::vector<Memb
...
+        }
         double old_score = DecodeDouble(old_score_bytes.data());
-        if (flags == kZSetIncr) {
+        if (flags & kZSetIncr) {
+          if (lt && (*mscores)[i].score >= 0) {
+            continue;
+          } else if (gt && (*mscores)[i].score <= 0) {
+            continue;
+          }
```
git-hulk:
```suggestion
          if (lt && (*mscores)[i].score >= 0) {
            continue;
          } else if (gt && (*mscores)[i].score <= 0) {
            continue;
          }
```
Can merge into one line with `||`

src/types/redis_zset.cc
```diff
@@ -71,19 +72,35 @@ rocksdb::Status ZSet::Add(const Slice &user_key, uint8_t flags, std::vector<Memb
...
           (*mscores)[i].score += old_score;
           if (std::isnan((*mscores)[i].score)) {
             return rocksdb::Status::InvalidArgument("resulting score is not a number (NaN)");
           }
         }
         if ((*mscores)[i].score != old_score) {
+          if (lt && (*mscores)[i].score >= old_score) {
```
git-hulk: Same with above, can merge those two conditions

src/commands/redis_cmd.cc
```diff
@@ -2380,12 +2381,20 @@ class CommandSInterStore : public Commander {
 class CommandZAdd : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    if (args.size() % 2 != 0) {
-      return Status(Status::RedisParseErr, errInvalidSyntax);
+    unsigned index = 2;
+    parseOptions(args, index);
```
git-hulk: Can use consistent name like `parseFlags`

src/commands/redis_cmd.cc
```diff
@@ -2400,19 +2409,74 @@ class CommandZAdd : public Commander {
...
+  }
+  if (lt && gt) {
+    return Status(Status::RedisParseErr, errZSetLTGTNX);
+  }
+  if (lt && nx) {
+    return Status(Status::RedisParseErr, errZSetLTGTNX);
+  }
+  if (gt && nx) {
+    return Status(Status::RedisParseErr, errZSetLTGTNX);
+  }
```
git-hulk:
```suggestion
  if ((lt && gt) || (lt && nx) || (gt && nx)) {
    return Status(Status::RedisParseErr, errZSetLTGTNX);
  }
```

torwig:
@manchurio You can also consider using something like the following to encapsulate the logic with options and eliminate code duplication:

```
struct ZAddOptions {
  ZAddOptions() {}
  explicit ZAddOptions(uint8_t flags) : flags(flags) {}

  bool IsXX() const { return (flags & kZSetNX) != 0; }
  // other methods to ask if the specific flag was set

  uint8_t flags = 0;
};
```

Of course, you can add methods like `SetXX` to encapsulate set operations as well (to not spread bit operations over the codebase).

src/commands/redis_cmd.cc
torwig: @git-hulk I think that `options` is OK since the `Redis` documentation names them as `ZADD options`. Perhaps the function name should be more specific like `parseAddOptions`. However, this method belongs to the class `CommandZAdd` so it may be clear that they are ZAdd-options. What do you think?

git-hulk: Yes, flags or options are OK. I saw that it calls 'options' when parsing but used `validateFlags` when validating, so I think it will be better to keep consistent if possible.

torwig: @git-hulk Now I see. Yes, you are right here. There should be one name everywhere for them (flags or options).

git-hulk: 👍 Yes, I didn't claim the context clearly in the first comment.

tests/gocase/unit/type/zset/zset_test.go
```diff
@@ -93,6 +93,82 @@ func basicTests(t *testing.T, rdb *redis.Client, ctx context.Context, encoding s
...
+		rdb.Del(ctx, "ztmp")
+		require.Equal(t, 1.5, rdb.ZAddArgsIncr(ctx, "ztmp", redis.ZAddArgs{NX: true, Members: []redis.Z{{Member: "abc", Score: 1.5}}}).Val())
+		require.Equal(t, 0.0, rdb.ZAddArgsIncr(ctx, "ztmp", redis.ZAddArgs{LT: true, Members: []redis.Z{{Member: "abc", Score: -1.5}}}).Val())
+		require.Equal(t, 0.0, rdb.ZAddArgsIncr(ctx, "ztmp", redis.ZAddArgs{LT: true, Members: []redis.Z{{Member: "abc", Score: 1.5}}}).Val())
+		require.Equal(t, redis.Nil, rdb.ZAddArgsIncr(ctx, "ztmp", redis.ZAddArgs{LT: true, Members: []redis.Z{{Member: "abc", Score: 1.5}}}).Err())
+	})
+
+	t.Run(fmt.Sprintf("ZSET ZADD LT/GT with other options - %s", encoding), func(t *testing.T) {
+		rdb.Del(ctx, "ztmp")
+		require.Equal(t, int64(1), rdb.ZAddArgs(ctx, "ztmp", redis.ZAddArgs{Members: []redis.Z{{Member: "abc", Score: 1.5}}}).Val())
```
tisonkun:
```suggestion
		require.EqualValues(t, 1, rdb.ZAddArgs(ctx, "ztmp", redis.ZAddArgs{Members: []redis.Z{{Member: "abc", Score: 1.5}}}).Val())
```

nit: better assertions.

ditto others.

manchurio: Great, it's more convenient.

src/types/redis_zset.cc
manchurio: It should not returns here.
eg. 
del z1
zadd z1 1 a
zadd z1 nx 2 b 2 a  //  b should be added since it is new.

If it returns here, b has no chance to be added.

git-hulk: Yes, you're right, my bad for misunderstanding the break logic.

src/types/redis_zset.cc
manchurio: merged

src/commands/redis_cmd.cc
manchurio: Ok, changed to parseFlags now.

src/commands/redis_cmd.cc
manchurio: merged.

src/types/redis_zset.cc
manchurio: merged

manchurio:
> @manchurio You can also consider using something like the following to encapsulate the logic with options and eliminate code duplication:
> 
> ```
> struct ZAddOptions {
>   ZAddOptions() {}
>   explicit ZAddOptions(uint8_t flags) : flags(flags) {}
> 
>   bool IsXX() const { return (flags & kZSetNX) != 0; }
>   // other methods to ask if the specific flag was set
> 
>   uint8_t flags = 0;
> };
> ```
> 
> Of course, you can add methods like `SetXX` to encapsulate set operations as well (to not spread bit operations over the codebase).


This refactoring has been completed

torwig:
@manchurio Thank you, I'll have a look at it as soon as possible.

src/commands/redis_cmd.cc
```diff
@@ -2400,19 +2409,65 @@ class CommandZAdd : public Commander {
...
+    rocksdb::Status s = zset_db.Add(args_[1], flags_, &member_scores_, &ret);
     if (!s.ok()) {
       return Status(Status::RedisExecErr, s.ToString());
     }
-    *output = Redis::Integer(ret);
+    bool incr = flags_.HasIncr();
+    if (incr) {
+      auto new_score = member_scores_[0].score;
+      bool nx = flags_.HasNX(), xx = flags_.HasXX(), lt = flags_.HasLT(), gt = flags_.HasGT();
+      if ((nx || xx || lt || gt) && old_score == new_score &&
```
tanruixiang: `incr` ,`nx`, `xx`, `lt` and `gt` are rebundant, It can be obtained directly using `Has*`.

src/commands/redis_cmd.cc
```diff
@@ -2400,19 +2409,65 @@ class CommandZAdd : public Commander {
...
+  if (!flags_.HasFlag()) {
+    return Status::OK();
+  }
+  bool nx = flags_.HasNX(), xx = flags_.HasXX(), lt = flags_.HasLT(), gt = flags_.HasGT();
+  if (nx && xx) {
+    return Status(Status::RedisParseErr, "XX and NX options at the same time are not compatible");
+  }
+  if ((lt && gt) || (lt && nx) || (gt && nx)) {
+    return Status(Status::RedisParseErr, errZSetLTGTNX);
+  }
```
tanruixiang: ditto.

src/types/redis_zset.cc
```diff
@@ -71,19 +72,31 @@ rocksdb::Status ZSet::Add(const Slice &user_key, uint8_t flags, std::vector<Memb
     }
     added_member_keys.insert(member_key);
 
+    bool lt = flags.HasLT();
+    bool gt = flags.HasGT();
```
tanruixiang: ditto.

tanruixiang (commented):
Hi. Since there is a unified interface to get the flag's status, then we don't need additional variables to store. Even if there is reuse, there is no difference in speed.

src/types/redis_zset.h
```diff
@@ -82,15 +82,39 @@ enum ZSetFlags {
   kZSetXX = 1 << 2,
   kZSetReversed = 1 << 3,
   kZSetRemoved = 1 << 4,
+  kZSetGT = 1 << 5,
+  kZSetLT = 1 << 6,
+  kZSetCH = 1 << 7,
 };
 
+typedef struct ZAddFlags {
```
torwig: @manchurio You can omit `typedef` here.

src/types/redis_zset.h
```diff
@@ -82,15 +82,39 @@ enum ZSetFlags {
...
+
+  void SetFlag(ZSetFlags setFlags) { flags |= setFlags; }
+
+  static const ZAddFlags Incr() { return ZAddFlags{kZSetIncr}; }
+
+  static const ZAddFlags Default() { return ZAddFlags{0}; }
+
+ private:
+  uint8_t flags = 0;
+} ZAddFlags;
```
torwig: And leave only `};` here, removing `ZAddFlags`.

src/types/redis_zset.h
```diff
@@ -82,15 +82,39 @@ enum ZSetFlags {
...
+typedef struct ZAddFlags {
+  explicit ZAddFlags(uint8_t flags = 0) : flags(flags) {}
+
+  bool HasNX() const { return (flags & kZSetNX) != 0; }
+  bool HasXX() const { return (flags & kZSetXX) != 0; }
+  bool HasLT() const { return (flags & kZSetLT) != 0; }
+  bool HasGT() const { return (flags & kZSetGT) != 0; }
+  bool HasCH() const { return (flags & kZSetCH) != 0; }
+  bool HasIncr() const { return (flags & kZSetIncr) != 0; }
+  bool HasFlag() const { return flags != 0; }
```
torwig: I'd rename this, for example, to `HasAnyFlags`.

src/types/redis_zset.h
manchurio: @torwig Could you explain why?
In my opinion, I prefer to  use "using"  instead of 'typedef', but for consistency with other code. I choose this way.

src/types/redis_zset.h
manchurio: Yes, you are right, I'd solved it later.

src/types/redis_zset.h
torwig: @manchurio Because it's just a C++ type definition:

```
struct Name {
// members
// methods
};

class OtherName {
// members
// methods
};
```

PragmaTwice: > @torwig Could you explain why? In my opinion, I prefer to use "using" instead of 'typedef', but for consistency with other code. I choose this way.

Just use `class A{};` instead of `typedef class A{} A;`. And you do not need to care consistency on `typedef`, we will replace them by `using`.

manchurio: OK, I also prefer class.

src/types/redis_zset.h
manchurio: Use class instead of typedef struct

src/commands/redis_cmd.cc
```diff
@@ -2377,12 +2378,20 @@ class CommandSInterStore : public Commander {
...
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    if (args.size() % 2 != 0) {
-      return Status(Status::RedisParseErr, errInvalidSyntax);
+    unsigned index = 2;
+    parseFlags(args, index);
+    if (auto s = validateFlags(); !s.IsOK()) {
+      return s;
+    }
+    if (auto left = (args.size() - index); left > 0) {
```
git-hulk: How about left = 0? if requests have only flags but no member and score?

git-hulk (commented):
Overall is good for me, only one comment.

src/commands/redis_cmd.cc
manchurio: It's a bug. It's fixed now.

torwig (approved):
LGTM

git-hulk (approved):

tanruixiang (approved):

PragmaTwice (approved):

tisonkun (approved):

PragmaTwice:
Sorry, touched wrong button. :rofl:
The new branch generated by the revert button has been deleted.

PR #1023, opened 2022-10-21
Add mailing list to README

PragmaTwice:
Since `dev@` is already overwhelmed with discussion messages, and I always get duplicated discussion notification mails from both github and `dev@`), I thought I could move the discussion notifications to `issues@` as well.

I also removed the google group link because it has not been used for a long time and many developers canot get notifications. Is this appropriate? @git-hulk

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@
 
 ---
 
-* [Google Group](https://groups.google.com/g/kvrocks)
 * [Slack Channel](https://join.slack.com/t/kvrockscommunity/shared_invite/zt-p5928e3r-OUAK8SUgC8GOceGM6dAz6w)
+* [Mailing List](https://lists.apache.org/list.html?dev@kvrocks.apache.org) ([Tutorial](https://www.apache.org/foundation/mailinglists.html))
 
 **Apache Kvrocks(Incubating)** is a distributed key value NoSQL database that uses RocksDB as storage engine and is compatible with Redis protocol. Kvrocks intends to decrease the cost of memory and increase the capacity while compared to Redis. The design of replication and storage was inspired by `rocksplicator` and `blackwidow`.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -25,8 +25,8 @@
 
 ---
 
-* [Google Group](https://groups.google.com/g/kvrocks)
 * [Slack Channel](https://join.slack.com/t/kvrockscommunity/shared_invite/zt-p5928e3r-OUAK8SUgC8GOceGM6dAz6w)
+* [Mailing List](https://lists.apache.org/list.html?dev@kvrocks.apache.org) ([Tutorial](https://www.apache.org/foundation/mailinglists.html))
```
tisonkun:
```suggestion
* [Mailing List](https://lists.apache.org/list.html?dev@kvrocks.apache.org) ([how to subscribe](https://www.apache.org/foundation/mailinglists.html#subscribing))
```

tisonkun (commented):
LGTM. One comment inline.

README.md
PragmaTwice: Thanks!

git-hulk (approved):
LGTM

tisonkun:
@PragmaTwice we can have a path filter for skipping CI on docs change.

But since we required the status, it can be a bit tricky and we should make use of https://github.com/dorny/paths-filter to skip steps.

It's not a requirement of this patch, of course.

tisonkun (approved):

PragmaTwice:
> @PragmaTwice we can have a path filter for skipping CI on docs change.
> 
> But since we required the status, it can be a bit tricky and we should make use of https://github.com/dorny/paths-filter to skip steps.
> 
> It's not a requirement of this patch, of course.

Sure, we need to skip CI if no actual code changes.

tisonkun:
Let me create an issue for it.

PR #1037, opened 2022-10-24
Support CLI options for kvrocks server

### LINKED ISSUE #1034: Add support for CLI arguments/environment variables to overwrite conf file values
### Search before asking

### Motivation

It will be good idea to support some configuration overwrites via CLI arguments or Environment variables. Specially it will be so much useful in containerized environment. We can pack the default conf file in the docker image and overwrite respective values by setting environment variables or CLI args.

### Solution

_No response_

PragmaTwice:
It closes #1034

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -601,6 +601,25 @@ void Config::ClearMaster() {
   }
 }
 
+Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &input, int line_number) {
+  std::string field_key = Util::ToLower(input.first);
+  const char ns_str[] = "namespace.";
+  size_t ns_str_size = sizeof(ns_str) - 1;
+  if (!strncasecmp(input.first.data(), ns_str, ns_str_size)) {
+    // namespace should keep key case-sensitive
+    field_key = input.first;
+    tokens[input.second] = input.first.substr(ns_str_size);
+  }
+  auto iter = fields_.find(field_key);
+  if (iter != fields_.end()) {
+    auto &field = iter->second;
+    field->line_number = line_number;
+    auto s = field->Set(input.second);
+    if (!s.IsOK()) return s;
+  }
+  return Status::OK();
+}
+
 Status Config::parseConfigFromString(const std::string &input, int line_number) {
   auto parsed = ParseConfigLine(input);
   if (!parsed) return parsed.ToStatus();
@@ -609,22 +628,7 @@ Status Config::parseConfigFromString(const std::string &input, int line_number) 
 
   if (kv.first.empty() || kv.second.empty()) return Status::OK();
 
-  std::string field_key = Util::ToLower(kv.first);
-  const char ns_str[] = "namespace.";
-  size_t ns_str_size = sizeof(ns_str) - 1;
-  if (!strncasecmp(kv.first.data(), ns_str, ns_str_size)) {
-    // namespace should keep key case-sensitive
-    field_key = kv.first;
-    tokens[kv.second] = kv.first.substr(ns_str_size);
-  }
-  auto iter = fields_.find(field_key);
-  if (iter != fields_.end()) {
-    auto &field = iter->second;
-    field->line_number = line_number;
-    auto s = field->Set(kv.second);
-    if (!s.IsOK()) return s;
-  }
-  return Status::OK();
+  return parseConfigFromPair(kv, line_number);
 }
 
 Status Config::finish() {
@@ -657,9 +661,9 @@ Status Config::finish() {
   return Status::OK();
 }
 
-Status Config::Load(const std::string &path) {
-  if (!path.empty()) {
-    path_ = path;
+Status Config::Load(const CLIOptions &opts) {
+  if (!opts.conf_file.empty()) {
+    path_ = opts.conf_file;
     std::ifstream file(path_);
     if (!file.is_open()) return Status(Status::NotOK, strerror(errno));
 
@@ -669,7 +673,6 @@ Status Config::Load(const std::string &path) {
       std::getline(file, line);
       Status s = parseConfigFromString(line, line_num);
       if (!s.IsOK()) {
-        file.close();
         return Status(Status::NotOK, "at line: #L" + std::to_string(line_num) + ", err: " + s.Msg());
       }
       line_num++;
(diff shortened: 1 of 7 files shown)


git-hulk:
wow, so speedy.

tests/gocase/integration/cli_options/cli_options_test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/apache/incubator-kvrocks/tests/gocase/util"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCLIOptions(t *testing.T) {
+	srv := util.StartServerWithCLIOptions(t, map[string]string{}, []string{"--maxclients", "23333"})
```
tisonkun, line 31: I suggest that we have a prefix to overwrite config values, like `-Dmaxclients` or use another metavar to avoid conflict with common prefixes.

tisonkun (changes requested):
We'd better guard config keys with a namespace.

tests/gocase/integration/cli_options/cli_options_test.go
tisonkun: Emmm...But Redis just use this syntax. Let me think of it.

tisonkun: What's your opinion on this topic? cc @git-hulk @caipengbo @PragmaTwice @torwig @tanruixiang

torwig: @tisonkun In my opinion, it's OK to use the _same_ option name without any special prefix because one value _overwrites_ another value of the _same_ option.

PragmaTwice: I prefer no prefix. There are only 3 other non-config options (help, version, config), and it will NOT increase.

git-hulk: From my side, it can make Redis users happy if we obey the Redis way, so the current implementation is ok to me.

tanruixiang: I think if we do encounter conflicting scenarios, we can again consider making some changes.

src/main.cc
```diff
@@ -112,30 +109,47 @@ void setupSigSegvAction() {
...
+  std::cout << program << description << std::endl
+            << std::left << std::setw(width) << "-c, --config <filename>"
+            << "set config file to <filename>, or `-' for stdin" << std::endl
+            << std::setw(width) << "-v, --version"
+            << "print version information" << std::endl
+            << std::setw(width) << "-h, --help"
+            << "print this help message" << std::endl
+            << std::setw(width) << "--<config-key> <config-value>"
+            << "overwrite specific config option <config-key> to <config-value>" << std::endl;
 }
```
tisonkun: You may try to improve this usage section. Currently, it shows:

```
➜  kvrocks git:(cli-options) ./cmake-build-debug/kvrocks -h
./cmake-build-debug/kvrocksa distributed key value NoSQL database that uses RocksDB as storage engine and is compatible with Redis protocol
-c, --config <filename>         set config file to <filename>, or `-' for stdin
-v, --version                   print version information
-h, --help                      print this help message
--<config-key> <config-value>   overwrite specific config option <config-key> to <config-value>
➜  kvrocks git:(cli-options) redis-server -h
Usage: ./redis-server [/path/to/redis.conf] [options] [-]
       ./redis-server - (read config from stdin)
       ./redis-server -v or --version
       ./redis-server -h or --help
       ./redis-server --test-memory <megabytes>
       ./redis-server --check-system

Examples:
       ./redis-server (run the server with default conf)
       echo 'maxmemory 128mb' | ./redis-server -
       ./redis-server /etc/redis/6379.conf
       ./redis-server --port 7777
       ./redis-server --port 7777 --replicaof 127.0.0.1 8888
       ./redis-server /etc/myredis.conf --loglevel verbose -
       ./redis-server /etc/myredis.conf --loglevel verbose

Sentinel mode:
       ./redis-server /etc/sentinel.conf --sentinel
```

tisonkun (changes requested):
OK. Then I have no objection on the option format. More comments inline.

src/main.cc
PragmaTwice: I think the current option description is pretty clear, and we do not need the same usage information as redis. If you have a good way to describe it, you can submit other PRs later.

tisonkun: You may notice the format is malformed and whitespace isn't pretty inserted.

Also, I don't think that we should print a long description like "a distributed key value NoSQL database that uses RocksDB as storage engine and is compatible with Redis protocol".

PragmaTwice: Done. Can check it again : )

src/main.cc
```diff
@@ -24,6 +24,9 @@
 #include <stdlib.h>
```
tisonkun, line 23: Perhaps remove `<getopt.h>` now.

src/config/config.h
```diff
@@ -61,6 +61,14 @@ struct CompactionCheckerRange {
   bool Enabled() { return Start != -1 || Stop != -1; }
 };
 
+struct CLIOptions {
```
tisonkun, line 64: nit: I'd prefer as simple as `Options`.

tisonkun (changes requested):
I test with `./build/kvrocks --unknown k` and it starts well. I'm thinking of whether we should fail if the config key is unknown.

PragmaTwice:
> I test with `./build/kvrocks --unknown k` and it starts well. I'm thinking of whether we should fail if the config key is unknown.

If you put an `unkown k` in the config file, it will also start well. It is not a problem of this PR.

src/config/config.h
PragmaTwice: I think `Options` is not specific enough.

src/main.cc
PragmaTwice: done

tisonkun (approved):

src/config/config.cc
```diff
@@ -657,30 +661,40 @@ Status Config::finish() {
...
-    if (!file.is_open()) return Status(Status::NotOK, strerror(errno));
+Status Config::Load(const CLIOptions &opts) {
+  if (!opts.conf_file.empty()) {
+    std::ifstream file;
+    std::istream *in = nullptr;
+    if (opts.conf_file == "-") {
+      in = &std::cin;
+    } else {
+      path_ = opts.conf_file;
+      file.open(path_);
```
git-hulk, line 672: iostream will close the file after exiting the scope, right?

PragmaTwice: Yeah

git-hulk: Thanks~

git-hulk (approved):

PragmaTwice:
Thanks all. Merging...

PR #1038, opened 2022-10-24
Use a consistent way to get the current time

### LINKED ISSUE #983: Flaky test on expire precision
### Search before asking

### Version

https://github.com/apache/incubator-kvrocks/pull/981/commits/461d0e8cc95cf959df50db2d75fecffbfd919875

### Minimal reproduce step

Flaky test on CI env.

Tests passed.

### What did you see instead?

```
--- FAIL: TestString (12.24s)
    --- FAIL: TestString/GETEX_EXAT_option (0.00s)
        assertions.go:36: 
            	Error Trace:	/home/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/unit/type/strings/assertions.go:36
            	            				/home/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/unit/type/strings/strings_test.go:155
            	Error:      	"11s" is not less than or equal to "10s"
            	Test:       	TestString/GETEX_EXAT_option
    --- FAIL: TestString/GETEX_PXAT_option (0.00s)
        assertions.go:36: 
            	Error Trace:	/home/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/unit/type/strings/assertions.go:36
            	            				/home/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/unit/type/strings/strings_test.go:162
            	Error:      	"11s" is not less than or equal to "10s"
            	Test:       	TestString/GETEX_PXAT_option
```

### Anything Else?

cc @git-hulk please take a look.

git-hulk:
As mentioned in https://github.com/apache/incubator-kvrocks/issues/983#issuecomment-1288931255, @mapleFU also figured out that Kvrocks use three ways to get the current time and confused developers, so we fixed it here by using a consistent way. 

This also may close #983, since I suspect the root cause is the different behavior of getting current time, so using the new way `clock_gettime` to keep consistent with Go will never be got any random failure in the expiry test cases.

### DIFF
--- a/src/common/util.cc
+++ b/src/common/util.cc
@@ -673,6 +673,8 @@ int aeWait(int fd, int mask, uint64_t timeout) {
   }
 }
 
+int64_t GetTimeStamp() { return static_cast<int64_t>(GetTimeStampMS() / 1000); }
+
 uint64_t GetTimeStampMS() {
   auto tp = std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::system_clock::now());
   auto ts = std::chrono::duration_cast<std::chrono::milliseconds>(tp.time_since_epoch());
--- a/src/common/util.cc
+++ b/src/common/util.cc
@@ -672,19 +672,4 @@ int aeWait(int fd, int mask, uint64_t timeout) {
     return retval;
   }
 }
-
-int64_t GetTimeStamp() { return static_cast<int64_t>(GetTimeStampMS() / 1000); }
-
-uint64_t GetTimeStampMS() {
-  auto tp = std::chrono::time_point_cast<std::chrono::milliseconds>(std::chrono::system_clock::now());
-  auto ts = std::chrono::duration_cast<std::chrono::milliseconds>(tp.time_since_epoch());
-  return ts.count();
-}
-
-uint64_t GetTimeStampUS() {
-  auto tp = std::chrono::time_point_cast<std::chrono::microseconds>(std::chrono::system_clock::now());
-  auto ts = std::chrono::duration_cast<std::chrono::microseconds>(tp.time_since_epoch());
-  return ts.count();
-}
-
 }  // namespace Util
--- a/src/common/util.h
+++ b/src/common/util.h
@@ -68,6 +68,7 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value);
 
 void ThreadSetName(const char *name);
 int aeWait(int fd, int mask, uint64_t milliseconds);
+
 template <typename Duration = std::chrono::seconds>
 auto GetTimeStamp() {
   return std::chrono::duration_cast<Duration>(std::chrono::system_clock::now().time_since_epoch()).count();
--- a/src/common/util.h
+++ b/src/common/util.h
@@ -72,7 +72,7 @@ template <typename Duration = std::chrono::seconds>
 auto GetTimeStamp() {
   return std::chrono::duration_cast<Duration>(std::chrono::system_clock::now().time_since_epoch()).count();
 }
-uint64_t GetTimeStampMS() { return GetTimeStamp<std::chrono::milliseconds>(); }
-uint64_t GetTimeStampUS() { return GetTimeStamp<std::chrono::microseconds>(); }
+inline uint64_t GetTimeStampMS() { return GetTimeStamp<std::chrono::milliseconds>(); }
+inline uint64_t GetTimeStampUS() { return GetTimeStamp<std::chrono::microseconds>(); }
 
 }  // namespace Util
--- a/src/common/util.h
+++ b/src/common/util.h
@@ -69,7 +69,7 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value);
 void ThreadSetName(const char *name);
 int aeWait(int fd, int mask, uint64_t milliseconds);
 template <typename Duration = std::chrono::seconds>
-auto GetTimeStamp() -> typename Duration::rep {
+auto GetTimeStamp() -> Duration::rep {
   return std::chrono::duration_cast<Duration>(std::chrono::system_clock::now().time_since_epoch()).count();
 }
 uint64_t GetTimeStampMS() { return GetTimeStamp<std::chrono::milliseconds>(); }
--- a/src/common/util.h
+++ b/src/common/util.h
@@ -69,7 +69,7 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value);
 void ThreadSetName(const char *name);
 int aeWait(int fd, int mask, uint64_t milliseconds);
 template <typename Duration = std::chrono::seconds>
-auto GetTimeStamp() -> Duration::rep {
+auto GetTimeStamp() {
   return std::chrono::duration_cast<Duration>(std::chrono::system_clock::now().time_since_epoch()).count();
 }
 uint64_t GetTimeStampMS() { return GetTimeStamp<std::chrono::milliseconds>(); }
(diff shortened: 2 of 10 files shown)


mapleFU:
Hi, `clock_gettime` is a posix api. maybe `std::steady_clock` is better here.

src/common/util.h
```diff
@@ -68,6 +68,7 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value);
 
 void ThreadSetName(const char *name);
 int aeWait(int fd, int mask, uint64_t milliseconds);
+int64_t GetTimeStamp();
 uint64_t GetTimeStampMS();
 uint64_t GetTimeStampUS();
```
PragmaTwice:
```suggestion
template <typename Duration = std::chrono::seconds>
auto GetTimeStamp() -> typename Duration::rep {
  return std::chrono::duration_cast<Duration>(std::chrono::system_clock::now().time_since_epoch()).count();
}
uint64_t GetTimeStampMS() { return GetTimeStamp<std::chrono::milliseconds>(); }
uint64_t GetTimeStampUS() { return GetTimeStamp<std::chrono::microseconds>(); }
```

PragmaTwice: We can get second/millisecond/microsecond directly without manual division.

src/common/util.h
```diff
@@ -68,7 +68,11 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value);
 
 void ThreadSetName(const char *name);
 int aeWait(int fd, int mask, uint64_t milliseconds);
-uint64_t GetTimeStampMS();
-uint64_t GetTimeStampUS();
+template <typename Duration = std::chrono::seconds>
+auto GetTimeStamp() -> typename Duration::rep {
```
PragmaTwice:
```suggestion
auto GetTimeStamp() -> Duration::rep {
```

PragmaTwice: Sorry, a typo :rofl:

git-hulk:
> Hi, `clock_gettime` is a posix api. maybe `std::steady_clock` is better here.

@mapleFU Yes, we are using C++ `std::chrono::system_clock` API instead of `clock_gettime` directly, can see `Util::GetTimeStamp*`.

src/common/util.h
git-hulk: aha, never mind, just append a new commit.

PragmaTwice:
> > Hi, `clock_gettime` is a posix api. maybe `std::steady_clock` is better here.
> 
> @mapleFU Yes, we are using C++ `std::chrono::system_clock` API instead of `clock_gettime` directly, can see `Util::GetTimeStamp*`.

Maybe it is [steady_clock](https://en.cppreference.com/w/cpp/chrono/steady_clock), which is a monotonic clock rather than [system_clock](https://en.cppreference.com/w/cpp/chrono/system_clock) (a system-wide real time wall clock) in c++11 and later.

src/common/util.h
PragmaTwice: Maybe we need also remove these old definitions in `util.cc`. @git-hulk

PragmaTwice:
@mapleFU But I have a question. I notice that unlike `system_clock`, standard of `steady_clock` does not specify that it measures Unix Time.

src/common/util.h
```diff
@@ -68,7 +68,11 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value);
 
 void ThreadSetName(const char *name);
 int aeWait(int fd, int mask, uint64_t milliseconds);
-uint64_t GetTimeStampMS();
-uint64_t GetTimeStampUS();
+template <typename Duration = std::chrono::seconds>
+auto GetTimeStamp() -> Duration::rep {
```
PragmaTwice:
```suggestion
auto GetTimeStamp() {
```
So sorry, the `typename` is needed. But I now just want to remove the type.

src/common/util.cc
```diff
@@ -673,6 +673,8 @@ int aeWait(int fd, int mask, uint64_t timeout) {
   }
 }
 
+int64_t GetTimeStamp() { return static_cast<int64_t>(GetTimeStampMS() / 1000); }
```
tisonkun: Add a comment that the return value is in seconds? Also, why cannot we ass this method as `GetTimeStampSeconds` in the `.h` file? And thus `GetTimeStampMillis`/`GetTimeStampMicros`. At least I guess the common default of timestamp is in milliseconds.

PragmaTwice: We need to remove this function definition. It is already defined in `.h`.

src/common/util.h
```diff
@@ -68,7 +68,11 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value);
...
 void ThreadSetName(const char *name);
 int aeWait(int fd, int mask, uint64_t milliseconds);
-uint64_t GetTimeStampMS();
-uint64_t GetTimeStampUS();
+template <typename Duration = std::chrono::seconds>
+auto GetTimeStamp() {
+  return std::chrono::duration_cast<Duration>(std::chrono::system_clock::now().time_since_epoch()).count();
+}
+uint64_t GetTimeStampMS() { return GetTimeStamp<std::chrono::milliseconds>(); }
+uint64_t GetTimeStampUS() { return GetTimeStamp<std::chrono::microseconds>(); }
```
PragmaTwice:
```suggestion
inline uint64_t GetTimeStampMS() { return GetTimeStamp<std::chrono::milliseconds>(); }
inline uint64_t GetTimeStampUS() { return GetTimeStamp<std::chrono::microseconds>(); }
```

mapleFU:
> But I have a question. I notice that unlike `system_clock`, standard of `steady_clock` does not specify that it measures Unix Time.

`time_point` can call `time_since_epoch`. Like the case here: https://github.com/facebook/rocksdb/blob/main/env/env_posix.cc#L142-L162

src/common/util.h
git-hulk: Looks we can't put the function body in header file, it will cause duplicate symbol conflicts.

PragmaTwice:
> > But I have a question. I notice that unlike `system_clock`, standard of `steady_clock` does not specify that it measures Unix Time.
> 
> `time_point` can call `time_since_epoch`. Like the case here: https://github.com/facebook/rocksdb/blob/main/env/env_posix.cc#L142-L162

`time_since_epoch` DOES NOT imply the Unix epoch.

src/common/util.cc
git-hulk: re @tisonkun, Yes, I think the timestamp is usually referred as the timestamp second.

mapleFU:
> `time_since_epoch` DOES NOT imply the Unix epoch.

You're right. C++20 says that `time_since_epoch` should follow the system epoch, and C++17's `time_since_epoch` is de facto unix epoch. But standard doesn't force it. I think just follow rocksdb's code is ok. Do you have any ideas? @PragmaTwice

PragmaTwice:
> > `time_since_epoch` DOES NOT imply the Unix epoch.
> 
> You're right. C++20 says that `time_since_epoch` should follow the system epoch, and C++17's `time_since_epoch` is de facto unix epoch. But standard doesn't force it. I think just follow rocksdb's code is ok. Do you have any ideas? @PragmaTwice

I notice that although no guarantee, most of implementations do use Unix time, so I think it is fine.

tisonkun (approved):

PragmaTwice:
But another problem is that, even in c++20, the `system_clock` is guaranteed to use Unix Time, but it seems `steady_clock` does not have this guarantee.

PragmaTwice (approved):

PragmaTwice:
Thanks all. Merging...

PR #1041, opened 2022-10-25
Fix missing dependency and add license files in Dockerfile

PragmaTwice:


tisonkun (approved):

git-hulk (dismissed):

Dockerfile
```diff
@@ -38,5 +40,8 @@ COPY ./kvrocks.conf  ./conf/
 RUN sed -i -e 's%dir /tmp/kvrocks%dir /var/lib/kvrocks%g' ./conf/kvrocks.conf
 VOLUME /var/lib/kvrocks
 
+COPY ./LICENSE ./NOTICE ./DISCLAIMER .
+COPY ./licenses ./licenses
```
tisonkun: Step 16/19 : COPY ./LICENSE ./NOTICE ./DISCLAIMER .
When using COPY with more than one source file, the destination must be a directory and end with a /

tisonkun (changes requested):

Dockerfile
tisonkun: Please verify locally first :)

PragmaTwice: It is weird. My local docker does not have this error : (

Dockerfile
```diff
@@ -38,5 +40,8 @@ COPY ./kvrocks.conf  ./conf/
 RUN sed -i -e 's%dir /tmp/kvrocks%dir /var/lib/kvrocks%g' ./conf/kvrocks.conf
 VOLUME /var/lib/kvrocks
 
+COPY ./LICENSE ./NOTICE ./DISCLAIMER .
```
PragmaTwice:
```suggestion
COPY ./LICENSE ./NOTICE ./DISCLAIMER ./
```

Dockerfile
PragmaTwice: done

tisonkun (approved):

PR #1045, opened 2022-10-25
rewrite *write function

xiaobiaozhao:
1. Resolve compile-time warning
2. Improves code robustness

### DIFF
--- a/src/common/util.cc
+++ b/src/common/util.cc
@@ -672,4 +672,28 @@ int aeWait(int fd, int mask, uint64_t timeout) {
     return retval;
   }
 }
+
+Status Write(int fd, const std::string &data) {
+  ssize_t n = 0;
+  while (n < static_cast<ssize_t>(data.size())) {
+    ssize_t nwritten = write(fd, data.c_str() + n, data.size() - n);
+    if (nwritten == -1) {
+      return Status(Status::NotOK, strerror(errno));
+    }
+    n += nwritten;
+  }
+  return Status::OK();
+}
+
+Status Pwrite(int fd, const std::string &data, std::istream::off_type offset) {
+  ssize_t n = 0;
+  while (n < static_cast<ssize_t>(data.size())) {
+    ssize_t nwritten = pwrite(fd, data.c_str() + n, data.size() - n, offset);
+    if (nwritten == -1) {
+      return Status(Status::NotOK, strerror(errno));
+    }
+    n += nwritten;
+  }
+  return Status::OK();
+}
 }  // namespace Util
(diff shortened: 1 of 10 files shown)


git-hulk (dismissed):

src/common/util.cc
```diff
@@ -672,4 +672,28 @@ int aeWait(int fd, int mask, uint64_t timeout) {
     return retval;
   }
 }
+
+Status Write(int fd, const std::string &data) {
```
PragmaTwice, line 676: I wonder if there is any difference between `SockSend` and this new function.

xiaobiaozhao: I'm going to create a pr that fixes the socksend function and replaces write with send. Becase send has more flags than write.

src/common/util.cc
```diff
@@ -672,4 +672,28 @@ int aeWait(int fd, int mask, uint64_t timeout) {
...
+    ssize_t nwritten = write(fd, data.c_str() + n, data.size() - n);
+    if (nwritten == -1) {
+      return Status(Status::NotOK, strerror(errno));
+    }
+    n += nwritten;
+  }
+  return Status::OK();
+}
+
+Status Pwrite(int fd, const std::string &data, std::istream::off_type offset) {
```
PragmaTwice:
```suggestion
Status Pwrite(int fd, const std::string &data, off_t offset) {
```
This function is not related to `std::istream`

src/common/util.cc
PragmaTwice: Do we really use the flag provided by `send`?

Currently `SockSend` has a totally same function body as this function. It is weird.

src/common/util.h
```diff
@@ -75,5 +75,7 @@ auto GetTimeStamp() {
 }
 inline uint64_t GetTimeStampMS() { return GetTimeStamp<std::chrono::milliseconds>(); }
 inline uint64_t GetTimeStampUS() { return GetTimeStamp<std::chrono::microseconds>(); }
-
+// file util
+Status Write(int fd, const std::string &data);
+Status Pwrite(int fd, const std::string &data, std::istream::off_type offset);
```
PragmaTwice: Same as above.

caipengbo (dismissed):

src/common/util.cc
xiaobiaozhao: done

git-hulk (approved):

caipengbo (approved):

PragmaTwice (approved):
Thanks. Generally looks good to me.
But as I mentioned above, currently `SockSend` has a totally same function body as the newly added function `Write`. So I will do some code clean if no further PR is proposed.

git-hulk:
Thanks all, merging..

PR #1047, opened 2022-10-27
Fix rename-command didn't work when using it multiple times

### LINKED ISSUE #1040: rename commands don't work
### Search before asking

### Version

2.1.0

### Minimal reproduce step

1、add rename-command in kvrocks.conf
rename-command flushall                             dont_run_flushall_rudely
rename-command flushdb                             dont_run_flushdb_rudely
rename-command bgsave                              dont_run_bgsave_rudely
rename-command config                               dont_run_config_rudely
2、start kvrocks

3、exec dont_run_bgsave_rudely
127.0.0.1:6379> dont_run_bgsave_rudely
(error) ERR unknown command dont_run_bgsave_rudely

exec dont_run_config_rudely   can success

127.0.0.1:6379> config get rename-command
1) "rename-command"
2) "config                              dont_run_config_rudely"

as it shows in config info，only dont_run_config_rudely effects。

4、exec config rewrite
127.0.0.1:6379> dont_run_config_rudely rewrite
OK
127.0.0.1:6379> dont_run_config_rudely get rename-command
1) "rename-command"
2) "config                              dont_run_config_rudely"

but rename-commands in kvrocks.conf still are old contents.
root@redis-kvsenior01-290192:/home/redis/etc# grep rename /home/redis/data/conf/kvrocks.conf
# rename
# rename
rename-command flushall                             dont_run_flushall_rudely
rename-command flushdb                             dont_run_flushdb_rudely
rename-command bgsave                              dont_run_bgsave_rudely
...
rename-command config                              dont_run_config_rudely

all rename commands can work

### What did you see instead?

no

### Anything Else?

_No response_

tanruixiang:
This close #1040

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -765,7 +765,7 @@ Status Config::Rewrite() {
   std::map<std::string, std::string> new_config;
   for (const auto &iter : fields_) {
     if (iter.second->IsMultiConfig()) {
-      // We should NOT overwrite the rename command since it cannot be rewritten in-flight,
+      // We should NOT overwrite the commands which are MultiConfig since it cannot be rewritten in-flight,
       // so skip it here to avoid rewriting it as new item.
       continue;
     }
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -725,12 +725,13 @@ void Config::Get(std::string key, std::vector<std::string> *values) {
   values->clear();
   for (const auto &iter : fields_) {
     if (key == "*" || Util::ToLower(key) == iter.first) {
-      values->emplace_back(iter.first);
       if (iter.second->GetConfigType() == configType::MultiConfig) {
         for (const auto &p : Util::Split(iter.second->ToString(), "\n")) {
-          values->push_back(p);
+          values->emplace_back(iter.first);
+          values->emplace_back(p);
         }
       } else {
+        values->emplace_back(iter.first);
         values->emplace_back(iter.second->ToString());
       }
     }
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -149,7 +149,7 @@ Config::Config() {
       {"profiling-sample-commands", false, new StringField(&profiling_sample_commands_, "")},
       {"slowlog-max-len", false, new IntField(&slowlog_max_len, 128, 0, INT_MAX)},
       {"purge-backup-on-fullsync", false, new YesNoField(&purge_backup_on_fullsync, false)},
-      {"rename-command", true, new StringField(&rename_command_, "")},
+      {"rename-command", true, new MultiStringField(&rename_command_, std::vector<std::string>{})},
       {"auto-resize-block-and-sst", false, new YesNoField(&auto_resize_block_and_sst, true)},
       {"fullsync-recv-file-delay", false, new IntField(&fullsync_recv_file_delay, 0, 0, INT_MAX)},
       {"cluster-enabled", true, new YesNoField(&cluster_enabled, false)},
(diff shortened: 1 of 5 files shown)


src/config/config.cc
```diff
@@ -723,7 +726,13 @@ void Config::Get(std::string key, std::vector<std::string> *values) {
...
     if (key == "*" || Util::ToLower(key) == iter.first) {
       values->emplace_back(iter.first);
-      values->emplace_back(iter.second->ToString());
+      if (iter.second->GetConfigType() == configType::MultiConfig) {
+        for (const auto &p : Util::Split(iter.second->ToString(), "\n")) {
+          values->push_back(p);
+        }
+      } else {
+        values->emplace_back(iter.second->ToString());
+      }
```
```suggestion
      if (iter.second->GetConfigType() == configType::MultiConfig) {
        for (const auto &p : Util::Split(iter.second->ToString(), "\n")) {
          values->emplace_back(iter.first);
          values->push_back(p);
        }
      } else {
        values->emplace_back(iter.first);
        values->emplace_back(iter.second->ToString());
      }
```
I think we need to keep one-key-one-value pairs, in order to make user parse the result of `config get *` easily.

src/config/config.cc
```diff
@@ -723,7 +726,13 @@ void Config::Get(std::string key, std::vector<std::string> *values) {
   for (const auto &iter : fields_) {
     if (key == "*" || Util::ToLower(key) == iter.first) {
       values->emplace_back(iter.first);
```
```suggestion
```

src/config/config.cc
```diff
@@ -273,23 +273,26 @@ void Config::initFieldValidator() {
...
-         auto commands = Redis::GetCommands();
-         auto cmd_iter = commands->find(Util::ToLower(args[0]));
-         if (cmd_iter == commands->end()) {
-           return Status(Status::NotOK, "No such command in rename-command");
-         }
-         if (args[1] != "\"\"") {
-           auto new_command_name = Util::ToLower(args[1]);
-           if (commands->find(new_command_name) != commands->end()) {
-             return Status(Status::NotOK, "Target command name already exists");
+         std::vector<std::string> all_args = Util::Split(v, "\n");
```
git-hulk, line 276: The rename command should not support:
```
rename-command a b
c d
e f 
``` 
so it makes no sense to split with newline here?

src/config/config.cc
```diff
@@ -722,8 +725,15 @@ void Config::Get(std::string key, std::vector<std::string> *values) {
   values->clear();
   for (const auto &iter : fields_) {
     if (key == "*" || Util::ToLower(key) == iter.first) {
-      values->emplace_back(iter.first);
-      values->emplace_back(iter.second->ToString());
+      if (iter.second->GetConfigType() == configType::MultiConfig) {
```
git-hulk: Can you add a test case for this line, it looks not correct for the rename-command now.

PragmaTwice:
https://github.com/apache/incubator-kvrocks/blob/debd8988d6f6978593ad35b7e06d0c3e773ee825/src/config/config.cc#L767-L771

Since **ALL `MultiStringField` cannot be rewritten**, so I think you need to change it to something like
```
if (iter->second.IsMulti())
```
in case of someone use `MultiStringField` but forget to include the config key in this check.

git-hulk:
> https://github.com/apache/incubator-kvrocks/blob/debd8988d6f6978593ad35b7e06d0c3e773ee825/src/config/config.cc#L767-L771
> 
> Since **ALL `MultiStringField` cannot be rewritten**, so I think you need to change it to something like
> 
> ```
> if (iter->second.IsMulti())
> ```
> 
> in case of someone use `MultiStringField` but forget to include the config key in this check.

Yes, that's right.

src/config/config.cc
tanruixiang: This is because we store the rename-command in the vector, and the code logic is obtained with Tostring. So we used `\n` as delimiter to concat this vector. For example: `KEYS KEYSNEW \n GET  GETNEW \n`

src/config/config.cc
tanruixiang: Test cases have been added.

tests/cppunit/config_test.cc
```diff
@@ -133,6 +133,33 @@ TEST(Config, GetAndSet) {
...
+
+  std::ostringstream string_stream;
+  string_stream << "rename-command KEYS KEYS_NEW"
+                << "\n";
+  string_stream << "rename-command GET GET_NEW"
+                << "\n";
+  string_stream << "rename-command SET SET_NEW"
+                << "\n";
+  std::ofstream output_file(path, std::ios::out);
+  output_file.write(string_stream.str().c_str(), string_stream.str().size());
```
PragmaTwice: You can write something like `output_file << "rename-command KEYS KEYS_NEW\n" << ...` directly.

PragmaTwice (approved):

src/config/config.cc
git-hulk: Okay, very sad that I made the field value as a string.

git-hulk (approved):

git-hulk:
Thanks @tanruixiang @PragmaTwice, merging...

PR #1055, opened 2022-10-30
Fix CAS and CAD return statuses

torwig:


git-hulk (approved):
LGTM, thanks @torwig

tisonkun (approved):
LGTM. Although, it seems that the method comment is incorrect also. Would you go through this method?

```cpp
//  1 if the operation is successful
//  -1 if the user_key does not exist
//  0 if the operation fails
```

PragmaTwice (approved):

caipengbo (approved):

tisonkun:
Merging...

Further improvement can be done later. They're not blockers.

PR #1060, opened 2022-11-01
Fix log-dir and backup-dir appear at the same time.

### LINKED ISSUE #1059: When log-dir and backup-dir appear at the same time, log-dir does not take effect
### Search before asking

### Version

unstable branch.
This bug was introduced on October 24.
https://github.com/apache/incubator-kvrocks/blob/684bcc6ce7060ab507627b8299b8326e9ffbf133/src/config/config.cc#L357-L359

### Minimal reproduce step

There are two ways to reproduce

1. Using the command flags
./kvrocks --log-dir /data/log --backup-dir /data/backup

2. Use configuration files(kvrocks.conf)
backup-dir /data/kvrocks/backup
log-dir /data/log

Log appears in /data/log

### What did you see instead?

Log appears in /tmp

### Anything Else?

If backup-dir appears, It will trigger the execution of this line of code.
https://github.com/apache/incubator-kvrocks/blob/684bcc6ce7060ab507627b8299b8326e9ffbf133/src/config/config.cc#L357-L359

Just remove this line and the bug will be fixed.
Or we reinitialize glog,because macro LOG will  initialize a singleton object before setting the log location.

https://github.com/google/glog/blob/05fbc65278db1aa545ca5cb743c31bc717a48d0f/src/logging.cc#L930-L940

manchurio:
This closes #1059

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -303,6 +303,8 @@ int main(int argc, char *argv[]) {
     return 1;
   }
   initGoogleLog(&config);
+  google::ShutdownGoogleLogging();
+  google::InitGoogleLogging("kvrocks");
   printVersion(LOG(INFO));
   // Tricky: We don't expect that different instances running on the same port,
   // but the server use REUSE_PORT to support the multi listeners. So we connect


git-hulk:
@manchurio Thanks for your report and PR, I think the root cause is using before the glog dir was specified in config::Load, so it will send logs to the default dir(/tmp), so this fix is not exactly correct.

manchurio:
> @manchurio Thanks for your report and PR, I think the root cause is using before the glog dir was specified in config::Load, so it will send logs to the default dir(/tmp), so this fix is not exactly correct.

@git-hulk Thanks for the explanation, do you have any suggestion?

git-hulk:
@manchurio I think we can simply remove the logging when the previous backup dir was changed. So that we won't use glog before `initGoogleLog` was called. cc @mapleFU

manchurio:
But that doesn't stop others from doing it again.

git-hulk:
> backup dir was changed. So that we won't use glog before `initGoogleLog` was called. cc @mapleFU

> But that doesn't stop others from doing it again.

We avoid this by moving `google::InitGoogleLogging` into initGoogleLog(), so that all logs will be sent to stderr.

mapleFU:
> I think we can simply remove the logging when the previous backup dir was changed. So that we won't use glog before `initGoogleLog` was called.

@git-hulk It's ok, but changing the backup dir may cause some problems, maybe we can logging in `set` command?

PragmaTwice:
We do not need to log the first assignment of backup dir:
```diff
--- if (!previous_backup.empty()) { 
+++ if (srv && !previous_backup.empty()) { 
```

git-hulk:
@manchurio #1061 has fixed this bug and will close this PR, feel free to reopen it if need.

Issue #1052, opened 2022-10-29
Upload go case log if integration tests failed


tisonkun:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

Otherwise, it's hard to debug and get the root cause of failure in CI environment.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

tisonkun:
The go case log locates at `$WORKSPACE/tests/gocase/workspace`.

tisonkun:
See also https://github.com/apache/incubator-kvrocks/pull/1065/files#r1012663263

PR #1065, opened 2022-11-02
Dump stacktrace for crashed instance started by go test in CI

PragmaTwice:
It closes #1052.

PR #1067, opened 2022-11-03
Fix dangling reference in ReplicationThread::ParseWriteBatch

PragmaTwice:
~~This may fixes the frequent crash problem in CI.~~

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -219,8 +219,18 @@ jobs:
           ./x.py test go build $GOCASE_RUN_ARGS
           SANITIZER_OUTPUT=$(grep "Sanitizer:" tests/gocase/workspace -r || true)
           if [[ $SANITIZER_OUTPUT ]]; then
+            echo "found sanitizer reports:"
             echo "$SANITIZER_OUTPUT"
-            echo "\ndetail reports:\n"
+            echo "detail log:"
+            cat $(find tests/gocase/workspace -iname stderr)
+            echo "sanitizer error was reported, exiting..."
+            exit 1
+          fi
+          CRASHES=$(grep "Ooops!" tests/gocase/workspace -r || true)
+          if [[ $CRASHES ]]; then
+            echo "found crashes:"
+            echo "$CRASHES"
+            echo "detail log:"
             cat $(find tests/gocase/workspace -iname stderr)
             echo "sanitizer error was reported, exiting..."
             exit 1
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -199,7 +199,9 @@ jobs:
       
       - name: Setup Coredump
         if: ${{ startsWith(matrix.os, 'ubuntu') }}
-        run: echo "$(pwd)/coredumps/corefile-%e-%p-%t" | sudo tee /proc/sys/kernel/core_pattern
+        run: |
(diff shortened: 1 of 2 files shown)


torwig (approved):
LGTM

torwig:
@PragmaTwice Good catch! I needed a few minutes to see the dangling reference in the code.

.github/workflows/kvrocks.yaml
```diff
@@ -217,15 +225,35 @@ jobs:
...
             exit 1
           fi
 
+      - uses: actions/upload-artifact@v2
+        if: ${{ failure() && startsWith(matrix.os, 'ubuntu') }}
+        with:
+          name: kvrocks-coredumps-${{ matrix.name }}
+          path: |
+            ./build/kvrocks
+            ./coredumps/*
```
tisonkun: Perhaps we also upload gocase workspace logs so that close https://github.com/apache/incubator-kvrocks/issues/1052?

.github/workflows/kvrocks.yaml
```diff
@@ -217,15 +225,35 @@ jobs:
...
+            cat $(echo "$CRASHES" | awk -F ':' '{print $1}')
             exit 1
           fi
 
+      - uses: actions/upload-artifact@v2
+        if: ${{ failure() && startsWith(matrix.os, 'ubuntu') }}
+        with:
+          name: kvrocks-coredumps-${{ matrix.name }}
+          path: |
+            ./build/kvrocks
```
tisonkun: Why do we upload the binary?

tisonkun (commented):

.github/workflows/kvrocks.yaml
PragmaTwice: You can review these changes in #1065.

.github/workflows/kvrocks.yaml
PragmaTwice: these coredump are useless without the binary. We cannot get the totally same binary in local due to compiler/libstdcxx/glibc version.

.github/workflows/kvrocks.yaml
tisonkun: So this patch is based on #1065 and #1065 closes #1052?

If so, I suggest you update the PR description (or at least the next time:))

.github/workflows/kvrocks.yaml
tisonkun: Get it.

tisonkun (approved):

.github/workflows/kvrocks.yaml
PragmaTwice: Yeah, it is based on #1065.

PragmaTwice:
It seems it does not solve the frequent crashing in CI. Merging...

torwig:
@PragmaTwice Crashes started about a few days ago (if I'm right). Could they be linked to a specific commit or related to the beginning of a new month (of course, it's impossible that day/month/year/moon can cause the program to crash)?

PragmaTwice:
> @PragmaTwice Crashes started about a few days ago (if I'm right). Could they be linked to a specific commit or related to the beginning of a new month (of course, it's impossible that day/month/year/moon can cause the program to crash)?

Sure. But since I cannot reproduce the crash locally (run go tests many times), it is hard for me to quickly determine which commit is at fault.

mapleFU:
How about print or tracing `bulk_string`'s WriteBatch? Maybe it can produce some important infos

Or we can bisect the results?

PR #1071, opened 2022-11-03
Export more rocksdb statistics to the info command

### LINKED ISSUE #1049: Export the block cache hit/miss statistics
### Search before asking

### Motivation

Block cache hit/miss is the key metric for the read performance, so it can help a lot when profiling if we can export the block cache hit/miss statistics to users.

### Solution

Read the block cache statistics from RocksDB, then export through the info command.

xiaobiaozhao:
close https://github.com/apache/incubator-kvrocks/issues/1049


src/stats/stats.h
```diff
@@ -37,6 +37,51 @@ enum StatsMetricFlags {
...
+  STATS_METRIC_BLOOM_FILTER_PREFIX_USEFUL,
+
+  STATS_METRIC_COMPACT_READ_BYTES,   // Bytes read during compaction
+  STATS_METRIC_COMPACT_WRITE_BYTES,  // Bytes written during compaction
+  STATS_METRIC_FLUSH_WRITE_BYTES,    // Bytes written during flush
+
+  // Row cache.
+  STATS_METRIC_ROW_CACHE_HIT,
+  STATS_METRIC_ROW_CACHE_MISS,
+
```
git-hulk: Remove them, we didn't use row cache row.

src/stats/stats.h
```diff
@@ -37,6 +37,51 @@ enum StatsMetricFlags {
...
+  STATS_METRIC_GET_HIT_L2_AND_UP,
+  STATS_METRIC_COMPACTION_KEY_DROP_NEWER_ENTRY,         // key was written with a newer value.
+                                                        // Also includes keys dropped for range del.
+  STATS_METRIC_COMPACTION_KEY_DROP_OBSOLETE,            // The key is obsolete.
+  STATS_METRIC_COMPACTION_KEY_DROP_RANGE_DEL,           // key was covered by a range tombstone.
+  STATS_METRIC_COMPACTION_KEY_DROP_USER,                // user compaction function has dropped the key.
+  STATS_METRIC_COMPACTION_RANGE_DEL_DROP_OBSOLETE,      // all keys in range were deleted.
+  STATS_METRIC_COMPACTION_OPTIMIZED_DEL_DROP_OBSOLETE,  // Deletions obsoleted before bottom level due to file gap
+                                                        // optimization.
+  STATS_METRIC_COMPACTION_CANCELLED,                    // If a compaction was canceled in sfm to prevent ENOSPC
```
git-hulk, line 61: Why do we need to care about this stat?

src/stats/stats.h
```diff
@@ -37,6 +37,51 @@ enum StatsMetricFlags {
...
+  STATS_METRIC_BYTES_WRITTEN,         // The number of uncompressed bytes issued by DB::Put(), DB::Delete(),
+                                      // DB::Merge(), and DB::Write().
+  STATS_METRIC_BYTES_READ,            // The number of uncompressed bytes read from DB::Get().  It could be
+                                      // either from memtables, cache, or table files.
+                                      // For the number of logical bytes read from DB::MultiGet(),
+                                      // please use NUMBER_MULTIGET_BYTES_READ.
+  STATS_METRIC_NUMBER_DB_SEEK_FOUND,  // The number of calls to seek/next/prev that returned data
+  STATS_METRIC_NUMBER_DB_NEXT_FOUND,
+  STATS_METRIC_NUMBER_DB_PREV_FOUND,
```
git-hulk, line 70: duplicated, please take a look at the old metrics we have.

src/stats/stats.h
```diff
@@ -37,6 +37,51 @@ enum StatsMetricFlags {
...
+  STATS_METRIC_BLOCK_CACHE_MISS,
+  STATS_METRIC_BLOCK_CACHE_HIT,
+  STATS_METRIC_BLOCK_CACHE_ADD,
+  STATS_METRIC_BLOCK_CACHE_BYTES_READ,
+  STATS_METRIC_BLOCK_CACHE_BYTES_WRITE,
+  STATS_METRIC_BLOOM_FILTER_USEFUL,
+  STATS_METRIC_BLOOM_FILTER_FULL_POSITIVE,
+  STATS_METRIC_BLOOM_FILTER_FULL_TRUE_POSITIVE,
+  STATS_METRIC_PERSISTENT_CACHE_HIT,
+  STATS_METRIC_PERSISTENT_CACHE_MISS,
```
git-hulk: We have no persistent cache now, so it makes no sense to add those two.

src/stats/stats.h
xiaobiaozhao: Do you mean NUMBER_DB_SEEK?

src/stats/stats.h
xiaobiaozhao: It  planned for TS

src/stats/stats.h
xiaobiaozhao: done

src/stats/stats.h
xiaobiaozhao: done

git-hulk:
@xiaobiaozhao Looks like you wrongly merged the previous commits and they appeared as new code. Please fix them first.

tisonkun (commented):
You should merge or rebase onto unstable instead of pick unstable commit thus they're different commit object to show in this patch.

src/stats/stats.h
git-hulk: Yes, I think those three metrics should no mean a lot for us.

xiaobiaozhao:
> You should merge or rebase onto unstable instead of pick unstable commit thus they're different commit object to show in this patch.

done

xiaobiaozhao:
> @xiaobiaozhao Looks like you wrongly merged the previous commits and they appeared as new code. Please fix them first.

done

xiaobiaozhao:
@git-hulk @tisonkun Merge?

ShooterIT:
does this change damage performance?

maybe we can add a new section which include these statistics, only show these when calling `info all` or `info $new_section`

git-hulk:
> does this change damage performance?
> 
> maybe we can add a new section which include these statistics, only show these when calling `info all` or `info $new_section`

Those metrics should NOT affect the performance because there're from memory statistics and are already in the RocksDB section.

xiaobiaozhao:
> > does this change damage performance?
> > maybe we can add a new section which include these statistics, only show these when calling `info all` or `info $new_section`
> 
> Those metrics should NOT affect the performance because there're from memory statistics and are already in the RocksDB section.

I think it's time to merge  @git-hulk @ShooterIT

ShooterIT:
I don't think current change is OK, the output of `info rocksdb` is tool long, it is not easy to read,
i prefer to show these info when calling `info all`, or add a new section to include them

git-hulk:
> I don't think current change is OK, the output of `info rocksdb` is tool long, it is not easy to read, i prefer to show these info when calling `info all`, or add a new section to include them

Another choice maybe to add another command to export those statistics, dose it sound good?

PragmaTwice (commented):
I think maybe we need a better way to oganize code about metric track & output, otherwise many repeated code will be added in `server.cc` to do this task for new metrics 🤣 , OR we can move them to `stats.cc`.

xiaobiaozhao:
> I don't think current change is OK, the output of `info rocksdb` is tool long, it is not easy to read, i prefer to show these info when calling `info all`, or add a new section to include them

Maybe the new section could be called "rocksdb-detail"

ShooterIT:
i found we already have "stats" command, it can report lots of rocksdb info.

xiaobiaozhao:
> i found we already have "stats" command, it can report lots of rocksdb info.

@git-hulk What do you think?

git-hulk:
Yes, can move them into the stat command

ShooterIT:
Hi @xiaobiaozhao did you see `stats` command output? it reports lots of info, is it enough?

xiaobiaozhao:
> Hi @xiaobiaozhao did you see `stats` command output? it reports lots of info, is it enough?

I'll take a look at that this weekend

xiaobiaozhao:
I think `stats` is enough

Issue #1029, opened 2022-10-23
Tracking issue: Fix clang-tidy reports in kvrocks


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

```
    clang-analyzer-core.CallAndMessage
    clang-analyzer-core.CallAndMessageModeling
    clang-analyzer-core.DivideZero
    clang-analyzer-core.DynamicTypePropagation
    clang-analyzer-core.NonNullParamChecker
    clang-analyzer-core.NonnilStringConstants
    clang-analyzer-core.NullDereference
    clang-analyzer-core.StackAddrEscapeBase
    clang-analyzer-core.StackAddressEscape
    clang-analyzer-core.UndefinedBinaryOperatorResult
    clang-analyzer-core.VLASize
    clang-analyzer-core.builtin.BuiltinFunctions
    clang-analyzer-core.builtin.NoReturnFunctions
    clang-analyzer-core.uninitialized.ArraySubscript
    clang-analyzer-core.uninitialized.Assign
    clang-analyzer-core.uninitialized.Branch
    clang-analyzer-core.uninitialized.CapturedBlockVariable
    clang-analyzer-core.uninitialized.UndefReturn
    clang-analyzer-cplusplus.InnerPointer
    clang-analyzer-cplusplus.Move
    clang-analyzer-cplusplus.NewDelete
    clang-analyzer-cplusplus.NewDeleteLeaks
    clang-analyzer-cplusplus.PlacementNew
    clang-analyzer-cplusplus.PureVirtualCall
    clang-analyzer-cplusplus.SelfAssignment
    clang-analyzer-cplusplus.SmartPtrModeling
    clang-analyzer-cplusplus.VirtualCallModeling
    clang-analyzer-deadcode.DeadStores
    clang-analyzer-nullability.NullPassedToNonnull
    clang-analyzer-nullability.NullReturnedFromNonnull
    clang-analyzer-nullability.NullabilityBase
    clang-analyzer-nullability.NullableDereferenced
    clang-analyzer-nullability.NullablePassedToNonnull
    clang-analyzer-nullability.NullableReturnedFromNonnull
    clang-analyzer-security.FloatLoopCounter
    clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling
    clang-analyzer-security.insecureAPI.SecuritySyntaxChecker
[...]

PR #1075, opened 2022-11-04
Fix part of the clang-tidy warning

tanruixiang:
Related to #1029 .

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -43,12 +43,17 @@ const char *errInvalidImportState = "Invalid import state";
 
 ClusterNode::ClusterNode(std::string id, std::string host, int port, int role, std::string master_id,
                          std::bitset<kClusterSlots> slots)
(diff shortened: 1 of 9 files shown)


PragmaTwice (approved):
Looks good to me.

I just wonder if there is any check whose reports in kvrocks are all fixed. And then we can put them in `.clang-tidy`, according to #1029.

git-hulk (approved):

tanruixiang:
> Looks good to me.
> 
> I just wonder if there is any check whose reports in kvrocks are all fixed. And then we can put them in `.clang-tidy`, according to #1029.

I am fixing them in file order. I will try to see which checks have been fully fixed when I have time.

PragmaTwice:
> > Looks good to me.
> > I just wonder if there is any check whose reports in kvrocks are all fixed. And then we can put them in `.clang-tidy`, according to #1029.
> 
> I am fixing them in file order. I will try to see which checks have been fully fixed when I have time.

Got it. Thanks.

PR #1076, opened 2022-11-05
Fix compiler warning in replication

PragmaTwice:
```
/home/twice/incubator-kvrocks/src/cluster/replication.cc: In ‘rocksdb::Status ReplicationThread::ParseWriteBatch(const std::string&)’:
/home/twice/incubator-kvrocks/src/cluster/replication.cc:903:10: Warning: switch does not handle ‘kBatchTypeNone’ [-Wswitch]
  903 |   switch (write_batch_handler.Type()) {
      |          ^
[ 92%] Built target kvrocks_objs
```

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -912,7 +912,7 @@ rocksdb::Status ReplicationThread::ParseWriteBatch(const std::string &batch_stri
         }
       }
       break;
-    case kBatchTypeStream:
+    case kBatchTypeStream: {
       auto key = write_batch_handler.Key();
       InternalKey ikey(key, storage_->IsSlotIdEncoded());
       Slice entry_id = ikey.GetSubKey();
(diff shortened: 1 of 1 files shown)


tisonkun (dismissed):

src/cluster/replication.cc
```diff
@@ -921,6 +921,9 @@ rocksdb::Status ReplicationThread::ParseWriteBatch(const std::string &batch_stri
       GetFixed64(&entry_id, &id.seq);
       srv_->OnEntryAddedToStream(ikey.GetNamespace().ToString(), ikey.GetKey().ToString(), id);
       break;
+    }
+    default:
```
tisonkun: I don't know whether C++ can resolve exhaustive `case` and we just write `case kBatchTypeNone:` here.

PragmaTwice: I think both of them is OK here.

tisonkun: Yep. I'd prefer explicit than implicit but it's not a blocker from merging :)

```suggestion
    case kBatchTypeNone:
```

PragmaTwice: done

git-hulk (approved):

PR #1080, opened 2022-11-08
New status methods

torwig:
The goal was to eliminate the direct usage of enum values of the `Status` class (`kNotOK`, `kNotFound`, etc.)
Some enum values were deleted: `kUnknownCommand` (replaced with `kInvalidCommand`), `KDBGetWALErr` (replaced with `kNotOK`).

### DIFF
--- a/tests/cppunit/status_test.cc
+++ b/tests/cppunit/status_test.cc
@@ -26,7 +26,7 @@
 TEST(StatusOr, Scalar) {
   auto f = [](int x) -> StatusOr<int> {
     if (x > 10) {
-      return {Status::NotOK, "x large than 10"};
+      return {Status::kNotOK, "x large than 10"};
     }
 
     return 2 * x + 5;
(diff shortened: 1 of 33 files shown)


PragmaTwice (commented):
I think I am not a fan of these changes, since after it we need to add two extra methods (e.g. `bool IsSomeCode() const { ... }` and `static Status SomeCode(std::string msg = {}) { ... }`) if we want to add a new status code `SomeCode`.

I think maybe using X macros would make maintenance easier (no need to manually add/remove extra methods), but I am not sure other developers would accept using X macros.

tests/cppunit/status_test.cc
```diff
@@ -26,7 +26,7 @@
 TEST(StatusOr, Scalar) {
   auto f = [](int x) -> StatusOr<int> {
     if (x > 10) {
-      return {Status::NotOK, "x large than 10"};
+      return Status::NotOK("x large than 10");
```
PragmaTwice, line 29: Actually the original code `{Status::NotOK, "x large than 10"}` here is typed `StatusOr<int>`, i.e. it equals `StatusOr<int>{Status::NotOK, "x large than 10"}` (call the ctor `StatusOr<T>(Code, std::string)`), and the new code `Status::NotOK("x large than 10")` is typed `Status`. Although since we have a constructor `StatusOr<T>(Status)` so it will go well.

torwig:
@PragmaTwice Yes, you are right about two methods. Perhaps we can leave it as it is for now and don't make such refactoring if it doesn't make programming easier.

PR #1112, opened 2022-11-13
Fix modernize-loop-convert warning reported by clang-tidy

### LINKED ISSUE #1095: Fix `modernize-loop-convert` warning reported by clang-tidy
### Search before asking

### Motivation

*This issue is tracked by #1029.*

***We welcome new contributors to take these issues as a beginning of a deep dive to kvrocks***

Currently we have enabled lots of clang-tidy checks, but there are so many reports already exist in the kvrocks code, so we cannot treat these report as errors to block future PR with some clang-tidy reported warnings in CI.

Hence the goal of this issue is to solve all `modernize-loop-convert` tagged clang-tidy reports, and then enable `warnings-as-errors` for this specific check in `.clang-tidy`.

To get clang-tidy reports for latest kvrocks code, there are several ways: you can run `./x.py check tidy` locally, or check the log of the latest run of GitHub Actions on the unstable branch (e.g. https://github.com/apache/incubator-kvrocks/actions/runs/3449328741/jobs/5757148924#step:8:916). To be friendly for new contributors, we list all `modernize-loop-convert` tagged reports below, so in normal cases you can just follow the log below and fix them one by one.

```log
/home/runner/work/incubator-kvrocks/incubator-kvrocks/src/common/task_runner.cc:57:3: warning: use range-based for loop instead [modernize-loop-convert]
  for (size_t i = 0; i < threads_.size(); i++) {
  ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      (auto & thread : threads_)
/home/runner/work/incubator-kvrocks/incubator-kvrocks/src/stats/stats.cc:33:5: warning: use range-based for loop instead [modernize-loop-convert]
    for (int j = 0; j < STATS_METRIC_SAMPLES; j++) {
    ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        (unsigned long & sample : im.samples)
/home/runner/work/incubator-kvrocks/incubator-kvrocks/src/stats/stats.cc:108:3: warning: use range-based for loop instead [modernize-loop-convert]
  for (int j = 0; j < STATS_METRIC_SAMPLES; j++) sum += inst_metrics[metric].samples[j];
  ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      (unsigned long sample : inst_metrics[metric].samples) sample
/home/runner/work/incubator-kvrocks/incubator-kvrocks/src/storage/redis_db.cc:614:3: warning: use range-based for loop instead [modernize-loop-convert]
  for (size_t i = 0; i < args_.size(); i++) {
  ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      (auto & arg : args_)
/home/runner/work/incubator-kvrocks/incubator-kvrocks/src/types/redis_set.cc:311:3: warning: use range-based for loop instead [modernize-loop-convert]
  for (size_t i = 0; i < keys.size(); i++) {
  ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      (auto key : keys)
/home/runner/work/incubator-kvrocks/incubator-kvrocks/src/types/redis_string.cc:140:3: warning: use range-based for loop instead [modernize-loop-convert]
  for (size_t i = 0; i < ns_keys.size(); i++) {
  ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      (auto & ns_key : ns_keys)
```

Some of these reports will give modification suggestions in the log, you can refer to them directly; otherwise, you need to understand the content of the warning and combine your C++ knowledge to fix it. (Fortunately, most reports (except those by clang static analyzer) are intuitive and easy to fix.)

When you fixed these all reports listed above, you can now add the check `modernize-loop-convert` to `warnings-as-errors` in `.clang-tidy` (in the root dir of the repo): We show a sample diff below to illustrate how to modify `.clang-tidy`:

```diff
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,7 +1,7 @@
 # refer to https://clang.llvm.org/extra/clang-tidy/checks/list.html
 Checks: ...

-WarningsAsErrors: ...
+WarningsAsErrors: ..., modernize-loop-convert
```

Then, if all of the above are completed, congratulations, you can submit a PR for your changes now!

### Solution

_No response_

MaximSmolskiy:
Fix #1095


src/types/redis_string.cc
```diff
@@ -137,8 +137,8 @@ std::vector<rocksdb::Status> String::MGet(const std::vector<Slice> &keys, std::v
   // don't use range-based for loop here, coz the slice member
   // would refer the address instead of copy the value, and use
   // range-based for loop may cause all members refer to the same addr
-  for (size_t i = 0; i < ns_keys.size(); i++) {
-    slice_keys.emplace_back(ns_keys[i]);
+  for (const auto &ns_key : ns_keys) {
```
torwig, line 137: @MaximSmolskiy This place is really tricky. The comment above it explains why we should NOT use a range-based loop here.

MaximSmolskiy: @torwig Can you please explain this comment in more details?
`ns_keys` is `std::vector<std::string>`. `std::vector` `operator[](size_type pos)` always returns reference (as said [here](https://en.cppreference.com/w/cpp/container/vector/operator_at)), so `ns_keys[i]` was reference to `std::string`. And I don't understand how `range-based for loop` can cause problems here, because `ns_key` now is reference to `std::string` too (as it was).

PragmaTwice: `emplace_back` will invalidate all references and iterators (if capacity changes), so it is not allowed to use an iterator or reference based loop (as well as for-range) here. Of cource, index is always safe.

MaximSmolskiy: @PragmaTwice I'm sorry, but I still don't understand what the problem is. Here we iterate through vector `ns_keys` and put its values to another vector `slice_keys`. So, `ns_keys` shouldn't be updated during this loop. So, range-based for loop should be equivalent to index based loop here.

PragmaTwice: Oh sorry. I did not notice that there are two different vectors. Some time to wait me to understand these code again :rofl:

PragmaTwice: Actually I cannot understand the comment. `Slice` will never *copy* a string, it will only *take reference to* a string. cc @git-hulk @ShooterIT

torwig: I've just tested the following code:

```
std::vector<std::string> ns_keys = {"one", "two", "three", "four"};
std::vector<rocksdb::Slice> slice_keys;
for (size_t i = 0; i < ns_keys.size(); i++) {
  slice_keys.emplace_back(ns_keys[i]);
}

for (auto key : slice_keys) {
  std::cout << key.ToString() << std::endl;
}
```

And the output is:

```
one
two
three
four
```

Then I changed it to (as in the `clang` recommendation):

```
std::vector<std::string> ns_keys = {"one", "two", "three", "four"};
std::vector<rocksdb::Slice> slice_keys;
for (auto &ns_key : ns_keys) {
  slice_keys.emplace_back(ns_key);
}

for (auto key : slice_keys) {
  std::cout << key.ToString() << std::endl;
}
```

And the output is still the same.
Perhaps at some point in time, the code in `kvrocks` was changed but the comment is still here (an doutdated). Or there is something else.

PragmaTwice: I think it can be something like:
```
std::vector<rocksdb::Slice> slice_keys(ns_keys.size());
std::copy(ns_keys.begin(), ns_keys.end(), slice_keys.begin());
```

PragmaTwice:
I think we can still use range-based for loop and remove these comment to continue with this PR if it is confirmed that these comments are out-of-date and not correct now. 
cc @git-hulk

src/types/redis_string.cc
git-hulk: @PragmaTwice @torwig @MaximSmolskiy If my memory serves, the root cause is the for loop would reuse the same variable `ns_key` address if we use `for (auto ns_key : ns_keys)` instead of `for (auto &ns_key : ns_keys)`. 

So we can remove this comment if `for (auto &ns_key : ns_keys)` works well.

PragmaTwice (approved):

PragmaTwice:
Thanks for your contribution!

PR #1116, opened 2022-11-13
feat: add support for Redis sintercard command

### LINKED ISSUE #1083: [NEW] Add `SINTERCARD` command aligned Redis
### Search before asking

### Motivation

add support for [SINTERCARD](https://redis.io/commands/sintercard/) command in Redis, which is available since Redis 7.0.0

### Solution

Since `SINTER` and `SCARD` were implemented already, I think it's quite straightforward to combine those two commands' logic.

boatrainlsz:
Add support for Redis sintercard command, this close #1083


tanruixiang:
Thank you very much for your contribution. Maybe you also need to add the go test, you can refer to `incubator-kvrocks/tests/gocase/unit/type/set/set_test.go`

src/commands/redis_cmd.cc
```diff
@@ -2270,6 +2277,54 @@ class CommandSInter : public Commander {
...
+    Redis::Set set_db(svr->storage_, conn->GetNamespace());
+    int64_t ret = 0;
+    auto s = set_db.InterCard(keys, limit, &ret);
+    if (!s.ok()) {
+      return Status(Status::RedisExecErr, s.ToString());
+    }
+    *output = Redis::Integer(ret);
+    return Status::OK();
+  }
+};
```
tanruixiang, line 2492: You can refer to #1032  to use a more concise parsing approach.

boatrainlsz: @tanruixiang Thanks for advice.

src/commands/redis_cmd.cc
```diff
@@ -2270,6 +2277,54 @@ class CommandSInter : public Commander {
   }
 };
 
+class CommandSInterCard : public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
```
PragmaTwice, line 2448: I think you can try to split this method to two methods: `Parse` and `Execute`.

tests/cppunit/t_set_test.cc
```diff
@@ -188,6 +188,27 @@ TEST_F(RedisSetTest, Inter) {
   set->Del(k3);
 }
 
+TEST_F(RedisSetTest, InterCard) {
+  int ret;
```
PragmaTwice, line 192: Would you like to also write some unit tests in golang? e.g. `tests/gocase/unit/type/...`

src/commands/redis_cmd.cc
PragmaTwice: It is optional to use the command parser, but I think at least we need to extract the parsing part in this method `Execute` to `Parse`.

src/types/redis_set.cc
```diff
@@ -133,6 +133,17 @@ rocksdb::Status Set::Card(const Slice &user_key, int *ret) {
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status Set::InterCard(const std::vector<Slice> &keys, int64_t limit, int64_t *ret) {
+  std::vector<std::string> members;
+  rocksdb::Status s = Inter(keys, &members);
+  if (!s.ok()) return s;
+  *ret = members.size();
+  if (limit > 0 && *ret > limit) {
```
torwig, line 141: The Redis docs states:

`By default, the command calculates the cardinality of the intersection of all given sets. When provided with the optional LIMIT argument (which defaults to 0 and means unlimited), if the intersection cardinality reaches limit partway through the computation, the algorithm will exit and yield limit as the cardinality. Such implementation ensures a significant speedup for queries where the limit is lower than the actual intersection cardinality.`

I think, your implementation doesn't speed up the process of intersection, it performs an intersection as usual (full) and adjusts the result by comparing it with the `limit` value. In order to truly use the `limit` value, you should use it somewhere inside `Inter()`, passing it as a parameter or introducing a completely separate method.

boatrainlsz:
Everybody, Thank you for all your suggestions, I will recheck related codes.

tisonkun:

torwig:
@tisonkun Resolved. However, this feature is not ready to be merged.

PragmaTwice:
Hi @boatrainlsz , do you still have time to continue this PR?

infdahai:
@PragmaTwice Can I fix this pr because of long-term loss of contact? I don't know how to collaborate with this.

PragmaTwice:
> @PragmaTwice Can I fix this pr because of long-term loss of contact? I don't know how to collaborate with this.

Sure. I think you can pick up this PR.

git-hulk:
Close this PR due to out-of-date and no response.

PR #1119, opened 2022-11-13
Support more `ZRANGE` options

### LINKED ISSUE #1117: Support more `ZRANGE` options
### Search before asking

### Motivation

Redis no longer supports `ZRANGEBYLEX` and `ZRANGEBYSCORE` since `6.2.0`, merging this feature into options in `ZRANGE`. Currently our `ZRANGE` has almost no options and needs to be aligned with redis.
https://redis.io/commands/zrange/
https://redis.io/commands/zrangebylex/
https://redis.io/commands/zrangebyscore/

Related to #1082 's discussion.

### Solution

_No response_

tanruixiang:
This close #1117 .


src/commands/redis_cmd.cc
```diff
@@ -2554,44 +2555,119 @@ class CommandZRange : public Commander {
...
-    auto parse_start = ParseInt<int>(args[2], 10);
-    auto parse_stop = ParseInt<int>(args[3], 10);
-    if (!parse_start || !parse_stop) {
-      return Status(Status::RedisParseErr, errValueNotInteger);
+    CommandParser parser(args, 4);
+    while (parser.Good()) {
+      if (parser.EatEqICaseFlag("BYSCORE", by_flag_)) {
+        by_flag_ = "BYSCORE";
+      } else if (parser.EatEqICase("BYLEX")) {
+        by_flag_ = "BYLEX";
```
```suggestion
      if (parser.EatEqICaseFlag("BYSCORE", by_flag_)) {
      } else if (parser.EatEqICaseFlag("BYLEX", by_flag_)) {
```

src/commands/redis_cmd.cc
```diff
@@ -2554,44 +2555,117 @@ class CommandZRange : public Commander {
...
  private:
+  std::string_view by_flag_ = "";
+  int offset_ = 0;
+  int count_ = -1;
   int start_ = 0;
   int stop_ = 0;
-  bool reversed_;
+  bool reversed_ = false;
   bool with_scores_ = false;
+  ZRangeLexSpec specLex_;
```
torwig: Change camel case: `specLex_` => `spec_lex_`

src/commands/redis_cmd.cc
```diff
@@ -2554,44 +2555,117 @@ class CommandZRange : public Commander {
...
+      }
+    } else {
+      by_flag_ = "BYINDEX";
+      auto parse_start = ParseInt<int>(args[2], 10);
+      auto parse_stop = ParseInt<int>(args[3], 10);
+      if (!parse_start || !parse_stop) {
+        return Status(Status::RedisParseErr, errValueNotInteger);
+      }
+      start_ = *parse_start;
+      stop_ = *parse_stop;
```
PragmaTwice: Forget to handle `REV` here.

tanruixiang: Here it is straightforward to use `reversed_`. It is handled in the following `execute`.
```
uint8_t flags = !reversed_ ? 0 : kZSetReversed;
```

src/commands/redis_cmd.cc
```diff
@@ -2554,44 +2555,117 @@ class CommandZRange : public Commander {
...
+      }
+      if (!with_scores_) {
+        output->append(Redis::MultiLen(static_cast<int64_t>(member_scores.size())));
+      } else {
+        output->append(Redis::MultiLen(static_cast<int64_t>(member_scores.size() * 2)));
+      }
+      for (const auto &ms : member_scores) {
+        output->append(Redis::BulkString(ms.member));
+        if (with_scores_) output->append(Redis::BulkString(Util::Float2String(ms.score)));
+      }
```
PragmaTwice: When writing code, it is important to prevent duplicate code fragments. 

Instead of mindlessly coding, I think it is better to think more about composition and abstraction.

src/commands/redis_cmd.cc
PragmaTwice: Oh thanks. Got it.
But I think processing same logic using different methods is an easy way to produce duplicated code.

tanruixiang: Here it is just to use the previous interface. I can use another wrapper interface to solve this problem.

src/commands/redis_cmd.cc
```diff
@@ -2554,152 +2556,135 @@ class CommandZPopMax : public CommandZPop {
...
-      if (!parse_offset || !parse_count) {
-        return Status(Status::RedisParseErr, errValueNotInteger);
-      }
-      spec_.offset = *parse_offset;
-      spec_.count = *parse_count;
-    }
-    return Commander::Parse(args);
+    const_cast<std::vector<std::string> &>(args).emplace_back("BYLEX");
+    if (reversed_) const_cast<std::vector<std::string> &>(args).emplace_back("REV");
+    return CommandZRange::Parse(args);
```
PragmaTwice, line 2667: I think you can find a better way to organize these code here.

To be honest, it is ugly to use `const_cast` (to break the original semantics) and takes more runtime overhead (`zrangebylex` can be more efficient to parse than `zrange`). 

And more important thing is, it is not the correct way to parse `zrangebylex`. For example, it will allow some commands like `zrangebylex key x y BYLEX`.

src/commands/redis_cmd.cc
```diff
@@ -2554,152 +2556,135 @@ class CommandZPopMax : public CommandZPop {
...
-    auto parse_start = ParseInt<int>(args[2], 10);
-    auto parse_stop = ParseInt<int>(args[3], 10);
-    if (!parse_start || !parse_stop) {
-      return Status(Status::RedisParseErr, errValueNotInteger);
+    int offset = 0;
+    int count = -1;
+    CommandParser parser(args, 4);
+    while (parser.Good()) {
+      if (parser.EatEqICaseFlag("BYSCORE", by_flag_)) {
+        spec_ = std::make_shared<ZRangeSpec>();
```
PragmaTwice: Could you explain a bit about the use of `shared_ptr` here?

tanruixiang: Concurrent instructions result in destructions on line `auto s = svr_->LookupAndCreateCommand(cmd_tokens.front(), &current_cmd_);`.
```
void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
  Config *config = svr_->GetConfig();
  std::string reply, password = config->requirepass;

  while (!to_process_cmds->empty()) {
    auto cmd_tokens = to_process_cmds->front();
    to_process_cmds->pop_front();

    if (IsFlagEnabled(Redis::Connection::kCloseAfterReply) && !IsFlagEnabled(Connection::kMultiExec)) break;

    auto s = svr_->LookupAndCreateCommand(cmd_tokens.front(), &current_cmd_);
```

src/commands/redis_cmd.cc
tanruixiang: It looks like I need to parse it all over again for these instructions. But I think using `const_cast` essentially splits the `zrangebylex` command into `zrange byindex`, which should make sense.
 I wanted to make these commands use the same code as zrange's option as much as possible, not only to facilitate debugging, but also to reflect the fact that the commands have been deprecated.

src/types/redis_zset.cc
```diff
@@ -224,7 +224,10 @@ rocksdb::Status ZSet::Pop(const Slice &user_key, int count, bool min, std::vecto
 }
 
 rocksdb::Status ZSet::Range(const Slice &user_key, int start, int stop, uint8_t flags,
-                            std::vector<MemberScore> *mscores) {
+                            std::vector<MemberScore> *mscores, int limit_offset, int limit_count) {
```
git-hulk, line 227: Can simplify the naming: `limit_offset` => `offset`, same as the `limit_count`

src/commands/redis_cmd.cc
git-hulk: I still can't understand this point, the lookup command is unique ptr and it should not have the race between workers.

tanruixiang: ~~Two tasks `t1` ,` t2` can be executed concurrently, and `t2` starts allocating resources while `t1` is executing. `t2` uses the same `current_cmd_` to get allocation space while t1 is still executing, which causes t1 to use the resources held by `unique_ptr` to be freed.~~
Strange phenomenon.
Here are the failure stacks.
https://github.com/apache/incubator-kvrocks/actions/runs/3458835932/jobs/5773641529

PragmaTwice: It is not correct. If `CommandZrange` has this problem, then other commands will also have this problem.

PragmaTwice:
I want to explain why kvrocks crashes when @tanruixiang is using `unique_ptr`.

Firstly, he creates `ZrangeCommon` and some derived structs of it, like this:
```
struct ZrangeCommon { ... };
struct ZrangeLex : ZrangeCommon { std::string x };
struct ZrangeIndex : ZrangeCommon { ... };
```

Then, he use `unique_ptr` to store `ZrangeCommon`:
```
unique_ptr<ZrangeCommon> spec_;

...

if (...) {
  spec_ = make_unique<ZrangeLex>(...);
} else if (...) {
  spec_ = make_unique<ZrangeIndex>(...);
}
```

Note: Unlike `shared_ptr`, `unique_ptr` does not do type erasure, which means that the default deleter in `unique_ptr` is determined at compile time for the destructor. In this case, destructor of `ZrangeCommon` will be called when `spec_` is destructed. Note that the actual object type at this point may be `ZRangeLex`, and calling the wrong destructor may not only lead to resource leaks, but also to other catastrophic consequences due to UB.

The design of `shared_ptr` is different in that `shared_ptr` does type erasure and its deleter is stored in runtime, which makes it store the exact function pointer of the destructor at construction (which, of course, adds runtime overhead), so there is no such problem.

So a better solution is to use a virtual destructor: `shared_ptr` is not needed here, because `shared_ptr` has extra overhead and other usefulness:

```
struct ZrangeCommon {
  ...
  virtual ~ZrangeCommon() {}
};
```

tanruixiang:
@PragmaTwice Thanks for the analysis. I have another question, why clang can help us to solve this problem.

PragmaTwice:
> @PragmaTwice Thanks for the analysis. I have another question, why clang can help us to solve this problem.

Calling wrong dtor is an UB, and UB can cause anything.

May crash, may not. Also may spent $1 from your digital wallet : )

src/types/redis_zset.h
```diff
@@ -35,46 +36,34 @@ const double kMinScore = (std::numeric_limits<float>::is_iec559 ? -std::numeric_
...
-    max = kMaxScore;
-    minex = maxex = false;
-    offset = -1;
-    count = -1;
-    removed = reversed = false;
-  }
-} ZRangeSpec;
-
-typedef struct ZRangeLexSpec {
+struct ZrangeCommon {
```
git-hulk: @tanruixiang I prefer the old way to separate `ZRangeSpec` and `ZRangeLexSpec`, it's more clear when parsing even with a bit of duplicate code and we also don't need to use the ptr in the Parse function.

src/commands/redis_cmd.cc
git-hulk: I think it's not necessary to generalize the `sepc_` if it makes the code not easy to read. We even can have the union of byLexSpec and byIndexSpec, which is easier to understand than using share_ptr here.

PragmaTwice: I think union may have similar problem: we need to determine the correct dtor since there are some non-POD in spec_ (i.e. std::string in LexSpec), maybe we can just use multiple values or std::variant here.

src/types/redis_zset.h
tanruixiang: We have three modes, and the possibility of increasing the options in future. If we don't use inheritance, the code is simply copied and pasted in three copies. Are there any other pitfalls in using pointers? @PragmaTwice

tisonkun:

tanruixiang:
I will try to finish this pr as soon as possible before the big refactoring.

tisonkun:
Closing as conflicts...

Any progress requires rework now.

PR #1120, opened 2022-11-13
`HrangebyLex` supports specify intervals

### LINKED ISSUE #1082: Doubt about the HRANGE command
### Search before asking

### Motivation

Currently, we implement HRANGE command, but i think the name is not suitable.

For zset, we have `ZRANGEBYLEX` which could get members with lexicographical ordering,
of course, `ZRANGE` has `BYLEX` option.
HRANGE also get fields with lexicographical ordering, so i think its name should be
`HRANGEBYLEX` aligned with `ZRANGEBYLEX`. If we use `HRANGE`, users may think it is
similar with `ZRANGE` which `start` and `stop` are the position in it.

What's more, i think the `min` and `max` arguments of `HRANGE` should start with
( or [, in order to specify if the range item is respectively exclusive or inclusive, also align
with `ZRANGEBYLEX`

### Solution

I suggest to change its name to `HRANGEBYLEX` and the `min` and `max` arguments of it
should start with `(` or `[` 

of course, we can implement `HRANGE` command with different options, but now, i think it
is not urgent since there is no demand of HRANGE(the same semantic of ZRANGE).

### LINKED ISSUE #1118: Support more `HRANGE` options
### Search before asking

### Motivation

Based on the discussion in #1082 , we will align the `HRANGE` instruction with the `ZRANGE`.
The format is as follows:
```
HRANGE key start stop [BYLEX] [REV] [LIMIT offset count]
```
Similar to `zrange`, if we use `bylex`, then `start` and `stop` are the indexes. If we use `bylex`, then `start` and `stop` are strings. `Limit` can specify `offset` and `count` regardless of whether `bylex` is used.
### Solution

_No response_

tanruixiang:
This close #1118 and #1082 .


src/commands/redis_cmd.cc
```diff
@@ -1508,25 +1508,60 @@ class CommandHGetAll : public Commander {
...
+      if (spec_.reversed) {
+        s = Redis::Hash::ParseRangeLexSpec(args[3], args[2], &spec_);
+      } else {
+        s = Redis::Hash::ParseRangeLexSpec(args[2], args[3], &spec_);
+      }
+      if (!s.IsOK()) {
+        return Status(Status::RedisParseErr, s.Msg());
+      }
+    } else {
+      by_flag_ = "BYINDEX";
```
git-hulk: BYINDEX should make no sense for the hash?

tanruixiang: Maybe there is a need for a minimum of k values or something similar?

tanruixiang: For example, to take the smallest element, you can use `hrange key 0 0 ` or `hrange key - + limit 0 1`. You do not need to specify a limit when using index, just the desired range.

torwig: @tanruixiang What do you mean by "the smallest element in a hash"?

tanruixiang: > @tanruixiang What do you mean by "the smallest element in a hash"?

The dictionary order in hash is minimum.

src/types/redis_hash.h
```diff
@@ -29,13 +29,23 @@
...
-typedef struct FieldValue {
+struct FieldValue {
   std::string field;
   std::string value;
-} FieldValue;
+};
 
 enum class HashFetchType { kAll = 0, kOnlyKey = 1, kOnlyValue = 2 };
 
+struct HashSpec {
```
torwig: I think that it should be `HashRangeSpec` or something like that to describe that this is a spec for a range command in a hash.

src/commands/redis_cmd.cc
git-hulk: I can get your point, but "the smallest element in a hash" is implicitly in LSM storage and may be confusing. IMHO, supporting the BYLEX should be enough for the hash range.

tanruixiang: Yes, they can do the same thing. Do you think it is better if we change the command to `hrange` or `hrangebylex`.Because `zrangebylex` is no longer supported. I'm not sure if we need to align our commands to `zrangebylex`

git-hulk: Yes, I think it's good to keep the hrange command with BYLEX option only.

tisonkun:

tanruixiang:
I will try to finish this pr as soon as possible before the big refactoring.

git-hulk (dismissed):
Cool, LGTM

src/commands/redis_cmd.cc
```diff
@@ -1652,7 +1660,10 @@ class CommandHRange : public Commander {
   }
 
  private:
-  int64_t limit_ = LONG_MAX;
+  bool reversed_ = false;
+  int64_t offset_ = 0;
+  int64_t count_ = -1;
+  HashRangeSpec spec_;
```
torwig: You can correct me if I'm wrong.
Suggestion: you can remove `offset` and `count` since you have them as members of `HashRangeSpec`.
Additionally, to avoid casting, perhaps you can change the type of `count` from `int64_t` to `uint64_t` (of course if there is no special behavior when `count == -1` means something because the initial value of it is -1).

tanruixiang: Yes, we need negative numbers. There are some redundant variables that can be removed.

src/commands/redis_cmd.cc
```diff
@@ -1616,31 +1616,36 @@ class CommandHGetAll : public Commander {
 class CommandHRange : public Commander {
```
```suggestion
class CommandHRangeByLex : public Commander {
```

src/commands/redis_cmd.cc
```diff
@@ -1614,34 +1614,39 @@ class CommandHGetAll : public Commander {
...
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     Redis::Hash hash_db(svr->storage_, conn->GetNamespace());
     std::vector<FieldValue> field_values;
-    auto s = hash_db.Range(args_[1], args_[2], args_[3], limit_, &field_values);
+    rocksdb::Status s = hash_db.RangeByLex(args_[1], spec_, &field_values);
     if (!s.ok()) {
-      return {Status::RedisExecErr, s.ToString()};
+      return Status(Status::RedisExecErr, s.ToString());
```
```suggestion
      return {Status::RedisExecErr, s.ToString()};
```
I prefer to stay at the original style.

src/types/redis_hash.h
```diff
@@ -36,6 +36,16 @@ struct FieldValue {
 
 enum class HashFetchType { kAll = 0, kOnlyKey = 1, kOnlyValue = 2 };
 
+struct HashRangeSpec {
```
PragmaTwice:
```suggestion
struct HashRangeLexSpec {
```

src/types/redis_hash.cc
```diff
@@ -355,4 +383,36 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
   return SubKeyScanner::Scan(kRedisHash, user_key, cursor, limit, field_prefix, fields, values);
 }
 
+Status Hash::ParseRangeLexSpec(const std::string &min, const std::string &max, HashRangeSpec *spec) {
```
PragmaTwice: Is it totally same as ZSet::ParseRangeLexSpec?

src/types/redis_hash.h
PragmaTwice: Is it totally same as ZRangeLexSpec?

tanruixiang: Yes, do we need to merge these two into one?

PragmaTwice: I think we can do it in the next PR.

src/types/redis_hash.cc
PragmaTwice: Same as above.

src/types/redis_hash.h
tanruixiang: Ok, I will unify it in the next pr.

src/commands/redis_cmd.cc
```diff
@@ -1614,34 +1614,39 @@ class CommandHGetAll : public Commander {
...
 
-class CommandHRange : public Commander {
+class CommandHRangeByLex : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    if (args.size() != 6 && args.size() != 4) {
-      return {Status::RedisParseErr, errWrongNumOfArguments};
+    CommandParser parser(args, 4);
+    while (parser.Good()) {
+      if (parser.EatEqICase("REV")) {
```
ShooterIT, line 1622: why support `REV` option?

ShooterIT (commented):
`zrangebylex` also has `lex range` parser, we should unite the two into one.

src/commands/redis_cmd.cc
tanruixiang: This can be aligned with zrange and makes sense, for example if you need to take the largest data.

tanruixiang: About aligning zrange I will deal with it in the new pr.

PragmaTwice (dismissed):

src/types/redis_hash.cc
PragmaTwice: We can merge these same structure and methods of HashRangeSpec and ZRangeSpec in next PR.

ShooterIT: why not this pr, i don't think we should import different function to deal with similar things

src/commands/redis_cmd.cc
ShooterIT: i think `hrangebylex` is enough currently since i heard there is such a demand. Developers may not be used to that `hash` support range operations, we can implement when someone wants it, WDYT?

src/types/redis_hash.cc
PragmaTwice: I have no strong opinion for this. Could you merge these structures and functions in this PR? @tanruixiang

tanruixiang: > I have no strong opinion for this. Could you merge these structures and functions in this PR? @tanruixiang

Ok. Sorry to have missed this comment.

src/types/redis_hash.cc
```diff
@@ -355,4 +383,36 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
   return SubKeyScanner::Scan(kRedisHash, user_key, cursor, limit, field_prefix, fields, values);
 }
 
+Status Hash::ParseRangeLexSpec(const std::string &min, const std::string &max, CommonRangeLexSpec *spec) {
```
PragmaTwice: Could you move the function to range_spec.cc, so it can be reused in zset and hash?

tanruixiang: OK, I am very sorry for missing this function.

src/common/range_spce.cc
```diff
@@ -0,0 +1,53 @@
+/*
```
PragmaTwice: there is a typo in the filename :)

src/common/range_spce.cc
```diff
@@ -0,0 +1,53 @@
...
+      spec->maxex = true;
+    } else if (max[0] == '[') {
+      spec->maxex = false;
+    } else {
+      return Status(Status::NotOK, "the max is illegal");
+    }
+    spec->max = max.substr(1);
+  }
+  return Status::OK();
+}
```
PragmaTwice: we prefer a newline in the end of the file

src/common/range_spec.h
```diff
@@ -0,0 +1,36 @@
...
+  std::string min, max;
+  bool minex, maxex; /* are min or max exclusive */
+  bool max_infinite; /* are max infinite */
+  int64_t offset, count;
+  bool removed, reversed;
+  CommonRangeLexSpec()
+      : minex(false), maxex(false), max_infinite(false), offset(-1), count(-1), removed(false), reversed(false) {}
+};
+
+Status ParseRangeLexSpec(const std::string &min, const std::string &max, CommonRangeLexSpec *spec);
```
PragmaTwice: Same as above, we prefer a newline in the end of the file

PragmaTwice (approved):
LGTM. cc @git-hulk  @ShooterIT

git-hulk (approved):

ShooterIT (approved):
Thanks for your impressive work @tanruixiang

PragmaTwice:
Thanks all. Merging...

PR #1125, opened 2022-11-15
Use core number as the default concurrent compile task number in x.py

PragmaTwice:
`4` is not a good value for default concurrent compile task number. 

In devices which core number is less than 4 (like vm of github actions), it can make the compile time even longer than `-j $(nproc)` (since there will be more CPU context switching), and if core number is more than 4, it will also take longer than `-j $(nproc)` in normal situation.

### DIFF
--- a/x.py
+++ b/x.py
@@ -27,6 +27,7 @@ import sys
 from typing import List, Any, Optional, TextIO, Tuple
 from shutil import copyfile
 from warnings import warn
+from multiprocessing import cpu_count
 
 CMAKE_REQUIRE_VERSION = (3, 16, 0)
 CLANG_FORMAT_REQUIRED_VERSION = (12, 0, 0)
@@ -337,7 +338,7 @@ if __name__ == '__main__':
     parser_check_tidy.set_defaults(func=clang_tidy)
     parser_check_tidy.add_argument('dir', metavar='BUILD_DIR', nargs='?', default='build',
                                    help="directory to store cmake-generated and build files")
-    parser_check_tidy.add_argument('-j', '--jobs', default=4, metavar='N', help='execute N build jobs concurrently')
+    parser_check_tidy.add_argument('-j', '--jobs', default=cpu_count(), metavar='N', help='execute N build jobs concurrently')
     parser_check_tidy.add_argument('--clang-tidy-path', default='clang-tidy',
                                    help="path of clang-tidy used to check source")
     parser_check_tidy.add_argument('--run-clang-tidy-path', default='run-clang-tidy',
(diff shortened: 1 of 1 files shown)


git-hulk:
I like this change

git-hulk (approved):

tisonkun (approved):
Thanks for your contribution!

tisonkun:
https://app.travis-ci.com/github/apache/incubator-kvrocks/jobs/588512318 Compilation failed. Please take a look @PragmaTwice .

PragmaTwice:
> https://app.travis-ci.com/github/apache/incubator-kvrocks/jobs/588512318 Compilation failed. Please take a look @PragmaTwice .

I notice that `cpu_count` returns `80` in travis ci, and then `make -j80` is run, causing OOM. It is weird.

git-hulk:
> > https://app.travis-ci.com/github/apache/incubator-kvrocks/jobs/588512318 Compilation failed. Please take a look @PragmaTwice .
> 
> I notice that `cpu_count` returns `80` in travis ci, and then `make -j80` is run, causing OOM. It is weird.

Maybe the `getcpu()` retrieves the number of physical host cores, but the building process is running in the docker environment with limited resources.

PragmaTwice:
> > > https://app.travis-ci.com/github/apache/incubator-kvrocks/jobs/588512318 Compilation failed. Please take a look @PragmaTwice .
> > 
> > 
> > I notice that `cpu_count` returns `80` in travis ci, and then `make -j80` is run, causing OOM. It is weird.
> 
> Maybe the `getcpu()` retrieves the number of physical host cores, but the building process is running in the docker environment with limited resources.

I also suspect so. But I do not know how to retrieve the correct core number. :rofl:

tanruixiang:
> I also suspect so. But I do not know how to retrieve the correct core number. 🤣

Try this? @PragmaTwice 
https://donghao.org/2022/01/20/how-to-get-the-number-of-cpu-cores-inside-a-container/

PragmaTwice:
> > I also suspect so. But I do not know how to retrieve the correct core number. 🤣
> 
> Try this? @PragmaTwice https://donghao.org/2022/01/20/how-to-get-the-number-of-cpu-cores-inside-a-container/

Thanks! I will try it. It would be nice if it can be cross-platform (e.g. linux/mac).

PragmaTwice:
```
$ cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us
cat: /sys/fs/cgroup/cpu/cpu.cfs_quota_us: No such file or directory
```

It is weird. It always reports "no such file" in my local host machine and docker container.

PragmaTwice:
I think it is hard to find a sound & robust way to retrieve current core number with respect to cpuset. 

`psutil` did this, but its implementation is complex.

git-hulk:
> I think it is hard to find a sound & robust way to retrieve current core number with respect to cpuset.
> 
> `psutil` did this, but its implementation is complex.

It should be not worth to do that if we need to import a new module or add too much complexity.

tisonkun (changes requested):
What if we simply remove this argument from required?

In the CI workflow, we can calculate CPU with shell command, and users can leverage they own utils to calculate it. If parallelism is not important for the use case, simple omit the config.

tanruixiang:
> ```
> $ cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us
> cat: /sys/fs/cgroup/cpu/cpu.cfs_quota_us: No such file or directory
> ```
> 
> It is weird. It always reports "no such file" in my local host machine and docker container.

In cgroup2, it seems to have been migrated to `cpu.max`.

https://stackoverflow.com/questions/65551215/get-docker-cpu-memory-limit-inside-container/65554131#65554131

PR #1128, opened 2022-11-19
Increase blocking timeout to harden the blocking pop test case

### LINKED ISSUE #1126: Unstable Go case list_test
### Search before asking

### Motivation

https://github.com/apache/incubator-kvrocks/actions/runs/3476132400/jobs/5811155755

```
--- FAIL: TestList (6.74s)
    --- FAIL: TestList/brpop:_arguments_are_empty (1.00s)
        tcp_client.go:79: 
            	Error Trace:	/home/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/unit/type/list/tcp_client.go:79
            	            				/home/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/unit/type/list/list_test.go:435
            	Error:      	Not equal: 
            	            	expected: int32(42)
            	            	actual  : uint8(0x24)
            	Test:       	TestList/brpop:_arguments_are_empty
```

### Solution

_No response_

git-hulk:
close #1126 

Currently, the blocking pop with the empty argument may fail if took a long time(more than 1s) to execute the RPUSH command. The leading cause is the previous BRPOP command would get the empty bulk string(`$-1\r\n`) instead of the element array when it's timed out.

### DIFF
--- a/tests/gocase/unit/type/list/list_test.go
+++ b/tests/gocase/unit/type/list/list_test.go
@@ -430,7 +430,7 @@ func TestList(t *testing.T) {
 
 		t.Run(fmt.Sprintf("%s: arguments are empty", popType), func(t *testing.T) {
 			require.NoError(t, rdb.Del(ctx, "blist1", "blist2").Err())
-			require.NoError(t, rd.WriteArgs(popType, "blist1", "blist2", "1"))
+			require.NoError(t, rd.WriteArgs(popType, "blist1", "blist2", "4"))
 			require.NoError(t, rdb.RPush(ctx, "blist1", "foo").Err())
 			rd.MustReadStrings(t, []string{"blist1", "foo"})
 			require.EqualValues(t, 0, rdb.Exists(ctx, "blist1").Val())


git-hulk:
I think 4 seconds is long enough since the default read/write timeout of the Redis client is 3 seconds.

tisonkun (approved):

git-hulk:
https://github.com/apache/incubator-kvrocks/actions/runs/3502202012/jobs/5866498105#step:12:209

Looks like the bug was caused by the blocking client wasn't wake up after pushing the new element and then response nil bulk string after the timeout was reached.

Update: I found the root cause should be a corner case that the blocking pop may be blocking until timeout even that has elements in the list. 

The condition like below:

```
Time       Thread A                       Thread B

T0:     received BRPOP                  received RPUSH

T1:     try to pop but got empty 

T2:                                     push the element

T3:                                     try to notify blocking clients but no one is blocking

T4:     add to blocking clients 
```

I may some time to think about how to fix this issue. To see folks have any comments or thoughts on this? @tisonkun @ShooterIT @torwig @torwig

git-hulk:
> https://github.com/apache/incubator-kvrocks/actions/runs/3502202012/jobs/5866498105#step:12:209
> 
> Looks like the bug was caused by the blocking client wasn't wake up after pushing the new element and then response nil bulk string after timeout reached.

PragmaTwice:
How about introduce some mutexes? like `map(key -> mutex)`. (After a cursory glance)

tisonkun:
Merging as a net win...

PR #1130, opened 2022-11-21
Remove default value of parallelism in x.py

tisonkun:
This closes #1125.


PragmaTwice (approved):
I think we need expose a parameter in dockerfile to specify the concurrent job level. Maybe in another PR.

tisonkun:
@PragmaTwice make sense. Perhaps `ARG BUILD_PARALLELISM`. You can read https://docs.docker.com/engine/reference/builder/#arg.

PR #1137, opened 2022-11-24
Tidy files inside src/commands

torwig:
Edit files according to clang-tidy tips.
Found a few places with missed `return` keywords.
Fixed some typos. 
Inserted blank lines to break the code into logical blocks.
Use `size_t` in `for` loops.
Uninitialized variables were initialized.
Condition statements `if (a == nullptr)` and `if (a != nullptr)` were replaced with `if (!a)` and `if (a)` respectively. 
The occurrences of `rocksdb::Status` were replaced with `auto`.

Narrowing conversions were left untouched for now.

P.S. Diff is really big because the file `src/commands/redis_cmd.cc` is big by itself.

### DIFF
--- a/src/commands/redis_cmd.cc
+++ b/src/commands/redis_cmd.cc
@@ -22,7 +22,6 @@
 
 #include <fcntl.h>
 #include <glog/logging.h>
-#include <sys/socket.h>
 
 #include <chrono>
 #include <climits>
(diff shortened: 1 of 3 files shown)


src/commands/redis_cmd.cc
```diff
@@ -632,20 +659,23 @@ class CommandPSetEX : public Commander {
...
     if (!ttl_ms) {
-      return Status(Status::RedisParseErr, errValueNotInteger);
+      return {Status::RedisParseErr, errValueNotInteger};
     }
-    if (*ttl_ms <= 0) return Status(Status::RedisParseErr, errInvalidExpireTime);
-    if (*ttl_ms > 0 && *ttl_ms < 1000) {
+
+    if (*ttl_ms <= 0) return {Status::RedisParseErr, errInvalidExpireTime};
+
+    if (*ttl_ms < 1000) {
```
torwig, line 667: Because the condition `*ttl_ms <= 0` was checked a few lines before, here we can omit the `*ttl_ms > 0` check.

git-hulk (approved):

PragmaTwice (approved):
LGTM. So many lines to review :rofl:

git-hulk:
Thanks all, merging...

PR #1139, opened 2022-11-25
Introduce fmt dependency to simplify string formatting

PragmaTwice:
In this PR, we introduce [**fmt**](https://github.com/fmtlib/fmt), the de facto standard of string formatting in C++ and the origin of `std::format` (c++20) & `std::print` (c++23), which is also the most popular formatting library in C++, used in softwares like 
ceph, mariadb, scylladb and mongodb.

And we can replace the old formatting methods (this PR replaced part of them, far from all):
- `snprintf` or similiar C functions: need to specify types like `d`, `f`, `s`, `g`, and hard to format basic types like `uint64_t` (need some ugly macros). cannot format custom types. need to create a fixed-length buffer by our own.
- `std::stringstream` or similiar stream-like C++ types: very slow, and verbose to write.
- `std::to_string` and string concat (like `"..." + to_string(num1) + "..." + to_string(num2)`): inefficient, and hard to write & read.

fmt is using the MIT license with an exemption clause: https://github.com/fmtlib/fmt/blob/master/LICENSE.rst


torwig (dismissed):
@PragmaTwice Great Job! It's a really handy and compact way to format strings.
 
LGTM.

git-hulk:
@PragmaTwice Changes are good to me, but we need to update the NOTICE as well.

PragmaTwice:
> @PragmaTwice Changes are good to me, but we need to update the NOTICE as well.

Done

git-hulk (approved):

tisonkun (approved):

tisonkun:
Merging...

BTW, NOTICE file can be concise. Depencies' licenses and notices can be simply copied under the `licenses` folder. Then we don't maintain a big text file (manually) :)

PR #1143, opened 2022-11-26
Replace c-style typedef with direct definition or using

PragmaTwice:


torwig (approved):

tisonkun (approved):
Can we add some linter/checker rules to prevent regressions?

PragmaTwice:
> Can we add some linter/checker rules to prevent regressions?

Headers are actually hard to precisely check by clang-tidy due to some reason, maybe I will dive into them later.

PR #1144, opened 2022-11-26
Split `util.h` to more concrete util headers

PragmaTwice:
As discussed in https://github.com/apache/incubator-kvrocks/pull/973#discussion_r992920261, it seems a header file named `util.h` can lead to a disaster: all functions that we do not know where to put will be placed in this file, making this file cluttered and bloated. 

We need more specific file names rather than `util.h`, so in this PR, we separate functions in `util.h` into different files: `string_util.h`, `io_util.h`, `time_util.h` and `thread_util.h`.


caipengbo (approved):

git-hulk (approved):

PragmaTwice:
Thanks all. Merging...

PR #1149, opened 2022-12-01
refactor: move VERSION to src/folder #1148

### LINKED ISSUE #1148: Move VERSION file to src folder
### Search before asking

### Motivation

Move `VERSION` file to `src` folder, so that we have a tidy root layout.

### Solution

_No response_

CNLHC:
This fixes #1148.


CNLHC:
`x.py` is only invoked from the project root so using relative dir in `x.py` seems to be ok.

tisonkun:
I'd prefer to write `VERSION.txt` instead of `.VERSION`. There's no reason for the VERSION file to be hidden.

@PragmaTwice for https://github.com/apache/incubator-kvrocks/issues/1148#issuecomment-1333696239, I'd say that `version.h.in` is not a typical source file also, but we host all source files and meta source files (that generates source files under `src`). You can think of proto definitions.

CNLHC:
> I'd prefer to write `VERSION.txt` instead of `.VERSION`. There's no reason for the VERSION file to be hidden.
> 
> @PragmaTwice for [#1148 (comment)](https://github.com/apache/incubator-kvrocks/issues/1148#issuecomment-1333696239), I'd say that `version.h.in` is not a typical source file also, but we host all source files and meta source files (that generates source files under `src`). You can think of proto definitions.

I agree with @tisonkun that the `VERSION` should be placed in the `src` because this file does influence the output binary and can be regarded as some kind of source file.

tisonkun (commented):
@PragmaTwice Another approach is that we thoroughly remove the `VERSION` file but pass the version by CMake option and default to `999.999.999` :)

PragmaTwice:
> @PragmaTwice Another approach is that we thoroughly remove the `VERSION` file but pass the version by CMake option and default to `999.999.999` :)

It seems hard to do so in a source distribution, while users are required to specify a version number to build.

tisonkun:
> It seems hard to do so in a source distribution, while users are required to specify a version number to build.

Yep..

tisonkun (approved):

tisonkun:

PR #1171, opened 2022-12-08
feat: add config for log cleaner

### LINKED ISSUE #1168: Support deleting old logs by day
### Search before asking

### Motivation

Currently, Kvrocks didn't have a way to auto-delete old logs which may occupy the disk capacity, so it'd be better to offer a way to keep recent logs only. 

related discussion: https://github.com/apache/incubator-kvrocks/discussions/1167

### Solution

We can do it by simply adding a configuration to enable or disable it since the glog has already supported removing old logs.

https://github.com/google/glog/blob/c18db3903b9f0abd80ae740fde94f7e32aa40b92/README.rst#automatically-remove-old-logs

IoCing:
This closes #1168.


kvrocks.conf
```diff
@@ -96,6 +96,15 @@ dir /tmp/kvrocks
...
 log-dir stdout
+# log-dir log
+
+# To enable the log cleaner. if set yes , glog will check if there are
+# overdue logs whenever a flush is performed
+enablelogcleaner yes
+
+# INFO level log file whose last modified time is greater than logcleanerday
+# will be unlinked
+logcleanerday 0
```
torwig: Could you please also mention in the comment what is the default value for `logcleanerday` and set it instead of 0? By the way, what will happen if the value of this setting is `0`? If 0 means "don't clean logs", you can mention it as well.

src/config/config.cc
```diff
@@ -159,6 +159,8 @@ Config::Config() {
       {"migrate-sequence-gap", false, new IntField(&sequence_gap, 10000, 1, INT_MAX)},
       {"unixsocket", true, new StringField(&unixsocket, "")},
       {"unixsocketperm", true, new OctalField(&unixsocketperm, 0777, 1, INT_MAX)},
+      {"enablelogcleaner", false, new YesNoField(&enablelogcleaner, false)},
+      {"logcleanerday", false, new IntField(&logcleanerday, 365, 0, INT_MAX)},
```
PragmaTwice: Why the default value here is 365?

src/main.cc
```diff
@@ -178,6 +178,15 @@ static void initGoogleLog(const Config *config) {
...
   } else {
     FLAGS_log_dir = config->log_dir;
+    if (config->enablelogcleaner) {
+      google::EnableLogCleaner(config->logcleanerday);
+      google::SetLogFilenameExtension(".log");
+      for (int level = google::INFO; level <= google::FATAL; level++) {
+        std::string des = google::LogSeverityNames[level];
+        des = config->log_dir + "/kvrocks_" + des + "_";
+        google::SetLogDestination(level, des.c_str());
+      }
```
PragmaTwice: I think we cannot have different log filenames just by enabling or disabling log cleaner.

kvrocks.conf
```diff
@@ -96,6 +96,15 @@ dir /tmp/kvrocks
 # We also can send logs to stdout/stderr is as simple as:
 #
 log-dir stdout
+# log-dir log
+
+# To enable the log cleaner. if set yes , glog will check if there are
+# overdue logs whenever a flush is performed
+enablelogcleaner yes
```
caipengbo: I think a single `logcleanerday` item is sufficient, with a value of 0(or -1) to turn off logcleaner.

kvrocks.conf
IoCing: thanks , i will add

src/main.cc
IoCing: OK ， i will try to resolve it。

kvrocks.conf
IoCing: thanks, i get it

kvrocks.conf
```diff
@@ -96,6 +96,15 @@ dir /tmp/kvrocks
...
 log-dir stdout
+# log-dir log
+
+# You can configure logcleanerday to control whether to enable logcleaner
+# and the maximum number of days that the INFO level logs will be kept.
+# if set logcleanerday to -1 , this mean disable logcleaner.
+# if set logcleanerday between 0 to INT_MAX , this mean enable logcleaner ,
+# and INFO level log file whose last modified time is greater than logcleanerday will be unlinked.
+# By default the logcleanerday is -1.
+logcleanerday -1
```
caipengbo: What does 0 mean？

git-hulk: how about renaming it to `log-retention-days`

IoCing: if set to 0 , during the flush, the previous INFO level logs will be immediately unlinked

src/config/config.cc
```diff
@@ -159,6 +159,7 @@ Config::Config() {
       {"migrate-sequence-gap", false, new IntField(&sequence_gap, 10000, 1, INT_MAX)},
       {"unixsocket", true, new StringField(&unixsocket, "")},
       {"unixsocketperm", true, new OctalField(&unixsocketperm, 0777, 1, INT_MAX)},
+      {"logcleanerday", false, new IntField(&logcleanerday, -1, -1, INT_MAX)},
```
git-hulk: We need to implement the callback function for this configuration if we expected to call to action when the value was changed. https://github.com/apache/incubator-kvrocks/blob/2ddb300b316486af20cb202c00d9c6e8fed27408/src/config/config.cc#L312

kvrocks.conf
git-hulk: So it means won't retent any logs except the current one？

IoCing: yes ，but only INFO level will be unlinked。WARNING，ERROR，FATAL level log will be kept

src/config/config.cc
IoCing: ok ， i will do it

kvrocks.conf
git-hulk: Got it, thanks for your explanation.

src/config/config.cc
IoCing: how about renaming it to log-overdue-days ，becaue i find related var in the glog is named overdue_days_

git-hulk: I prefer to use `retention` instead of overdue since it looks more clear than overdue. To see if other folks have thoughts on this. @PragmaTwice @torwig @caipengbo

IoCing: thanks ，i get it

caipengbo: > I prefer to use retention instead of overdue since it looks more clear than overdue. To see if other folks have thoughts on this. 

I also think `retention` is better, which is a more general word.

kvrocks.conf
```diff
@@ -96,6 +96,16 @@ dir /tmp/kvrocks
...
+# log-dir log
+
+# You can configure log-retention-days to control whether to enable logcleaner
+# and the maximum number of days that the INFO level logs will be kept.
+# if set to -1 , this mean disable logcleaner.
+# if set to between 0 to INT_MAX , this mean enable logcleaner ,
+# and INFO level log file whose last modified time is greater than log-retention-days will be unlinked.
+# if set to 0 , during the flush, the previous INFO level logs will be immediately unlinked.
+# By default the log-retention-days is -1.
+log-retention-days -1
```
```suggestion

# You can configure log-retention-days to control whether to enable the log cleaner
# and the maximum retention days that the INFO level logs will be kept.
#
# if set to -1, that means to disable the log cleaner.
# if set to 0, all previous INFO level logs will be immediately removed.
# if set to between 0 to INT_MAX, that means it will retent latest N(log-retention-days) day logs. 

# By default the log-retention-days is -1.
log-retention-days -1
```

src/config/config.cc
```diff
@@ -457,6 +458,15 @@ void Config::initFieldCallback() {
          if (cluster_enabled) srv->slot_migrate_->SetSequenceGapSize(sequence_gap);
          return Status::OK();
        }},
+      {"log-retention-days",
+       [this](Server *srv, const std::string &k, const std::string &v) -> Status {
+         if (log_retention_days != -1) {
```
```suggestion
       [this](Server *srv, const std::string &k, const std::string &v) -> Status {
         if (!srv) return Status::OK();
         if (Util::ToLower(log_dir) == "stdout") {
           return {Status::NotOK, "can't set the 'log-retention-days' when the log dir is stdout"};
         }

         if (log_retention_days != -1) {
```

git-hulk:
General looks good to me, except for two suggestions.

IoCing:
> General looks good to me, except for two suggestions.

thanks

git-hulk:
@IoCing Would you mind adding Go test cases for this configuration, to make sure it can be changed in-flight?

kvrocks.conf
```diff
@@ -97,6 +97,16 @@ dir /tmp/kvrocks
 #
 log-dir stdout
 
+# You can configure log-retention-days to control whether to enable the log cleaner
+# and the maximum retention days that the INFO level logs will be kept.
+#
+# if set to -1, that means to disable the log cleaner.
+# if set to 0, all previous INFO level logs will be immediately removed.
```
tisonkun, line 104: I read we use the same config for all logs. How can logs in other levels have different manners?

IoCing: > @IoCing Would you mind adding Go test cases for this configuration, to make sure it can be changed in-flight?

ok , i will try

kvrocks.conf
IoCing: > yes ，but only INFO level will be unlinked。WARNING，ERROR，FATAL level log will be kept

Sorry ,  this describe is wrong.

I went to see the code of blog . I find that it will unlink all level log.

In the glog , every time a log is written, it will be judged whether to do logclean.

glog set a variable FLAGS_logcleansecs to control the time interval of logclean operation (Default is 5 minutes)。

Each level has a logfileobject , but only one logclean object. So if INFO level logfileobject

do logclean at time 0 , other level logfileobject can only logclean after five minutes.

the code of glog
https://github.com/google/glog/blob/master/src/logging.cc#:~:text=void%20LogCleaner%3A%3AUpdateCleanUpTime,UpdateCleanUpTime()%3B

kvrocks.conf
IoCing: > I read we use the same config for all logs. How can logs in other levels have different manners?

At present, glog does not provide an interface to configure the expiration time of different levels
I think it is possible to submit a pr to glog, and then we can set up accordingly。

kvrocks.conf
git-hulk: Thanks for your re-claimed, it'd be better if it could remove all level logs instead of only the INFO level.

kvrocks.conf
git-hulk: @IoCing I think we can move on with all logs with the same retention behavior first, then optimize it if necessary.

tisonkun (approved):

tisonkun:
It seems Travis CI takes a longer time to finish. Perhaps we should restore the GitHub Actions + QEMU solution XD.

Merging...

tisonkun:
@IoCing Thank you!

PR #1172, opened 2022-12-09
Add timeout when replica connect master

### LINKED ISSUE #1170: SLAVEOF command may block a long time
### Search before asking

### Version

2.2.0

### Minimal reproduce step

1. `echo 6 > /proc/sys/net/ipv4/tcp_syn_retries`
2. `SLAVEOF` a non-routable IP address(such as `10.255.255.1`)
3. `SLAVEOF new_master_ip port`

Step 3 should not block for a long time because we are switching to the new master, which should be a quick operation.

### What did you see instead?

Step 3 will block a long time(**> 2min** based on above reproduced scenario)

### Anything Else?

This is because we use blocking IO when use socket [connect(2)](https://man7.org/linux/man-pages/man2/connect.2.html). In some cases, `connect(2)` will be blocked for a long time.

caipengbo:
Fixed: #1170


src/cluster/replication.cc
```diff
@@ -236,18 +237,21 @@ void ReplicationThread::CallbacksStateMachine::Start() {
     handlers_.emplace_front(CallbacksStateMachine::WRITE, "auth write", authWriteCB);
   }
 
+  uint64_t connect_timestamp = 0, connect_timeout_ms = 3100;
+
   while (!repl_->stop_flag_ && bev == nullptr) {
-    Status s = Util::SockConnect(repl_->host_, repl_->port_, &cfd);
+    if (Util::GetTimeStampMS() - connect_timestamp < 1000) {
```
git-hulk: this condition should always be false? coz the connect_timestamp isn't updated at all.

caipengbo: I'm terribly sorry, I forgot one line :(

src/cluster/replication.cc
```diff
@@ -236,18 +237,22 @@ void ReplicationThread::CallbacksStateMachine::Start() {
...
```
git-hulk: @caipengbo How about renaming connect_timestamp to last_connect_timestamp, and also adding a comment line to explain why we need this logic. 

// Sleep here to prevent frequent re-connect when the master is down with the connection refused error

caipengbo: Done

src/cluster/replication.cc
```diff
@@ -236,18 +237,23 @@ void ReplicationThread::CallbacksStateMachine::Start() {
     handlers_.emplace_front(CallbacksStateMachine::WRITE, "auth write", authWriteCB);
   }
 
+  uint64_t last_connect_timestamp = 0, connect_timeout_ms = 3100;
```
tanruixiang: If `connect_timeout_ms` is not dynamically updated, it would be better to use constants?

caipengbo: I had thought about using a constant before, but found that it wasn't used elsewhere, so I wrote it as a local variable to make it more intuitive.

caipengbo:
Does CI fail because timeout is too short? @git-hulk

git-hulk:
> Does CI fail because timeout is too short? @git-hulk

Not sure, maybe connecting with timeout didn't work at all, you can have a test on your side.

caipengbo:
I found some problems with `ScopeExit` and `SockConnect(timeout)` @git-hulk @PragmaTwice

PragmaTwice (approved):
LGTM. Sorry for these mistakes.

src/common/io_util.cc
```diff
@@ -151,9 +151,7 @@ Status SockConnect(const std::string &host, uint32_t port, int *fd, int conn_tim
     sin.sin_port = htons(port);
 
     fcntl(*fd, F_SETFL, O_NONBLOCK);
-    if (connect(*fd, reinterpret_cast<sockaddr *>(&sin), sizeof(sin))) {
-      return Status::FromErrno();
-    }
+    connect(*fd, reinterpret_cast<sockaddr *>(&sin), sizeof(sin));
```
PragmaTwice, line 154: I think maybe we can filter some specific error codes, e.g.
```cpp
if (int err = connect(...); err && errno != EINPROGRESS)
```

caipengbo: Is it enough for `Util::aeWait()` on the next line to handle the error code?

PragmaTwice: Looks good to me.

git-hulk (approved):

tanruixiang (approved):
LGTM.

git-hulk:
Thanks all, merging...

PR #1173, opened 2022-12-09
Fix StatusOr copy assignment

SGZW:


PragmaTwice (approved):
Great catch! So thanks, that is an embarrassing typo.

PragmaTwice:
Thanks for your contribution. Merging...

PR #1183, opened 2022-12-16
Fix SockConect doesn't resolve domain before connecting

### LINKED ISSUE #1182: SockConnect with timeout doesn't support resolve domain
### Search before asking

### Version

unstable

### Minimal reproduce step

1. Start Kvrocks serverA and serverB on the same machine
2. Server A becomes the slave of serverB with localhost domain

It's ok to use the domain(local) to connect the master

### What did you see instead?

Cannot connect to the master, related issue also can see： https://github.com/apache/incubator-kvrocks/discussions/1178

### Anything Else?

_No response_

git-hulk:
This closes #1182 

Currently, Kvrocks supports connecting the server with timeout or not, but the behavior is a bit different between them. The timeout version doesn't support resolving the domain before connecting to the server, so it'll be broken if use it to connect the domain.

SockConnect with timeout is the first time to use in #1172, so this bug won't affect the previous releases.

### DIFF
--- a/src/common/io_util.cc
+++ b/src/common/io_util.cc
@@ -106,10 +106,8 @@ Status SockConnect(const std::string &host, uint32_t port, int *fd, int conn_tim
   if (int rv = getaddrinfo(host.c_str(), std::to_string(port).c_str(), &hints, &servinfo); rv != 0) {
     return {Status::NotOK, gai_strerror(rv)};
   }
-  if (int rv = getaddrinfo(host.c_str(), std::to_string(port).c_str(), &hints, &servinfo); rv != 0) {
-    return {Status::NotOK, gai_strerror(rv)};
-  }
   auto exit = MakeScopeExit([servinfo] { freeaddrinfo(servinfo); });
+
   for (p = servinfo; p != nullptr; p = p->ai_next) {
     auto cfd = UniqueFD(socket(p->ai_family, p->ai_socktype, p->ai_protocol));
     if (!cfd) continue;
@@ -128,8 +126,9 @@ Status SockConnect(const std::string &host, uint32_t port, int *fd, int conn_tim
       if ((retmask & AE_WRITABLE) == 0 || (retmask & AE_ERROR) != 0 || (retmask & AE_HUP) != 0) {
         return Status::FromErrno();
       }
+
+      // restore to the block mode
       int socket_arg = 0;
-      // Set to blocking mode again...
       if ((socket_arg = fcntl(*cfd, F_GETFL, NULL)) < 0) {
         return Status::FromErrno();
       }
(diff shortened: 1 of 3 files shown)


juancresc (dismissed):

src/common/io_util.cc
```diff
@@ -131,57 +97,67 @@ Status SockSetTcpKeepalive(int fd, int interval) {
...
+  hints.ai_socktype = SOCK_STREAM;
 
-    fcntl(*fd, F_SETFL, O_NONBLOCK);
-    connect(*fd, reinterpret_cast<sockaddr *>(&sin), sizeof(sin));
+  if (int rv = getaddrinfo(host.c_str(), std::to_string(port).c_str(), &hints, &servinfo); rv != 0) {
+    return {Status::NotOK, gai_strerror(rv)};
+  }
+  if (int rv = getaddrinfo(host.c_str(), std::to_string(port).c_str(), &hints, &servinfo); rv != 0) {
+    return {Status::NotOK, gai_strerror(rv)};
+  }
```
PragmaTwice: Duplicated?

git-hulk: Yes, it's duplicated code. I will revise this PR again and add a test case for this issue.

tisonkun (approved):

PragmaTwice (approved):

src/common/io_util.cc
```diff
@@ -131,57 +97,66 @@ Status SockSetTcpKeepalive(int fd, int interval) {
...
-    if ((socket_arg = fcntl(*fd, F_GETFL, NULL)) < 0) {
-      return Status::FromErrno();
+    Status s = SockSetTcpKeepalive(*cfd, 120);
+    if (s.IsOK()) {
+      s = SockSetTcpNoDelay(*cfd, 1);
     }
-    socket_arg &= (~O_NONBLOCK);
-    if (fcntl(*fd, F_SETFL, socket_arg) < 0) {
-      return Status::FromErrno();
+    if (!s.IsOK()) {
```
PragmaTwice, line 145:
```suggestion
    else {
```

src/common/io_util.cc
```diff
@@ -131,57 +97,66 @@ Status SockSetTcpKeepalive(int fd, int interval) {
...
+      continue;
```
PragmaTwice, line 146: continue or return error directly, which is better here?

caipengbo (approved):

src/common/io_util.cc
git-hulk: Looks better to continue if it has other addresses to have a try.

git-hulk:
Thanks all, merging...

PR #1191, opened 2022-12-17
Remove old string parsing util and replace them by ParseInt

PragmaTwice:
We remove `DecimalStringToNum` and `OctalStringToNum` in this PR, since they are hard to use and too less general than `ParseInt` (e.g. the parsed integer type cannot be changed), and they are just wrappers of `ParseInt` after we introduce `ParseInt`.


torwig (approved):
LGTM

git-hulk (approved):

PragmaTwice:
Thanks all. Merging...

PR #1195, opened 2022-12-18
Check if the thread is joinable before joining in replication thread

### LINKED ISSUE #1194: The replication thread core dump when stopping the thread
### Search before asking

### Version

unstable

### Minimal reproduce step

https://github.com/apache/incubator-kvrocks/actions/runs/3719371631/jobs/6308230988#step:12:138

CI runs with green status

### What did you see instead?

Core dump when stopping the server

### Anything Else?

_No response_

git-hulk:
This may close #1194 which is a flaky test CI and it may coredump when joining the replication thread in server::Stop. So let's check if the thread is joinable and still throws the exception if it's NOT caused by the thread join state.

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -333,7 +333,7 @@ void ReplicationThread::Stop() {
 
   stop_flag_ = true;  // Stopping procedure is asynchronous,
                       // handled by timer
-  t_.join();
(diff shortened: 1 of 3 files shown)


PragmaTwice (approved):

torwig (approved):

git-hulk:
Thanks all, merging...

PR #1197, opened 2022-12-19
Add support of stream migration during slot migration process

### LINKED ISSUE #1087: Add support for migrating STREAM in cluster mode
### Search before asking

### Motivation

The STREAM type(partial commands) is supported in #745, but didn't add support for migrating in the cluster mode.

For more detail can see: https://github.com/apache/incubator-kvrocks/blob/unstable/src/cluster/slot_migrate.cc#L617

### Solution

_No response_

torwig:
The `XSETID` command was implemented. This command is used only during stream key migration.
It differs from the same `Redis` command just in one case: in `Redis`, you **can't** apply it to a non-existing stream; in `Kvrocks` you **can** because streams are allowed to be empty and there should be a way to migrate an empty stream (with no prior `XADD` command which can create a stream if it doesn't exist).
Related C++ unit tests were added.

Stream key migration was implemented. The key of the  `stream`  type is migrated via a separate function.
Related Go tests were added.

Refactor and tidy cluster-related code (`src/cluster` folder).

Closes #1087

### DIFF
--- a/src/cluster/slot_migrate.cc
+++ b/src/cluster/slot_migrate.cc
@@ -23,28 +23,23 @@
 #include <memory>
 #include <utility>
 
+#include "db_util.h"
 #include "event_util.h"
 #include "fmt/format.h"
 #include "io_util.h"
 #include "storage/batch_extractor.h"
 #include "thread_util.h"
 #include "time_util.h"
+#include "types/redis_stream_base.h"
+#include "types/redis_string.h"
 
 static std::map<RedisType, std::string> type_to_cmd = {
     {kRedisString, "set"}, {kRedisList, "rpush"},    {kRedisHash, "hmset"},      {kRedisSet, "sadd"},
-    {kRedisZSet, "zadd"},  {kRedisBitmap, "setbit"}, {kRedisSortedint, "siadd"},
+    {kRedisZSet, "zadd"},  {kRedisBitmap, "setbit"}, {kRedisSortedint, "siadd"}, {kRedisStream, "xadd"},
 };
 
-SlotMigrate::SlotMigrate(Server *svr, int speed, int pipeline_size, int seq_gap)
-    : Database(svr->storage_, kDefaultNamespace),
-      svr_(svr),
-      state_machine_(kSlotMigrateNone),
-      current_pipeline_size_(0),
-      last_send_time_(0),
-      slot_job_(nullptr),
-      slot_snapshot_time_(0),
-      wal_begin_seq_(0),
-      wal_increment_seq_(0) {
+SlotMigrate::SlotMigrate(Server *svr, int migration_speed, int pipeline_size_limit, int seq_gap)
+    : Database(svr->storage_, kDefaultNamespace), svr_(svr) {
   // Let db_ and metadata_cf_handle_ be nullptr, and get them in real time to avoid accessing invalid pointer,
   // because metadata_cf_handle_ and db_ will be destroyed if DB is reopened.
   // [Situation]:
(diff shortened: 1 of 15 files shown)


src/cluster/slot_migrate.cc
```diff
@@ -554,28 +564,29 @@ bool SlotMigrate::CheckResponseWithCounts(int sock_fd, int total) {
...
   std::string prefix_key;
   AppendNamespacePrefix(key, &prefix_key);
-  std::string bytes = value.ToString();
+  std::string bytes = encoded_metadata.ToString();
   Metadata metadata(kRedisNone, false);
   metadata.Decode(bytes);
-  if (metadata.Type() != kRedisString && metadata.size == 0) {
+  if (metadata.Type() != kRedisString && metadata.Type() != kRedisStream && metadata.size == 0) {
     LOG(INFO) << "[migrate] No elements of key: " << prefix_key;
-    return Status(Status::cOK, "empty");
```
PragmaTwice, line 565: The original code here is weird. We do not allow an OK status has a message, which will break the semantics of Status.

src/commands/redis_cmd.cc
```diff
@@ -6254,6 +6254,66 @@ class CommandXTrim : public Commander {
...
+    if (!s.IsOK()) {
+      return {Status::RedisParseErr, s.Msg()};
+    }
+
+    if (args.size() == 3) {
+      return Status::OK();
+    }
+
+    for (size_t i = 3; i < args.size(); /* manual increment */) {
+      if (Util::ToLower(args[i]) == "entriesadded" && i + 1 < args.size()) {
```
```suggestion
      if (Util::EqualICase(args[i], "entriesadded") && i + 1 < args.size()) {
```

src/cluster/slot_migrate.cc
torwig: Yes, I introduced a dedicated enum type for that but forgot to use this very line. I'll fix it soon.

torwig: Wow, I didn't forget :)
I used `KeyMigrationResult` to not rely on error messages because it can be error-prone.

src/commands/redis_cmd.cc
torwig: Done.

PragmaTwice (approved):

git-hulk (approved):

src/cluster/slot_migrate.h
```diff
@@ -77,89 +83,92 @@ struct SlotMigrateJob {
...
   bool CheckResponseOnce(int sock_fd);
   bool CheckResponseWithCounts(int sock_fd, int total);
 
-  Status MigrateOneKey(const rocksdb::Slice &key, const rocksdb::Slice &value, std::string *restore_cmds);
+  KeyMigrationResult MigrateOneKey(const rocksdb::Slice &key, const rocksdb::Slice &encoded_metadata,
+                                   std::string *restore_cmds);
   bool MigrateSimpleKey(const rocksdb::Slice &key, const Metadata &metadata, const std::string &bytes,
                         std::string *restore_cmds);
   bool MigrateComplexKey(const rocksdb::Slice &key, const Metadata &metadata, std::string *restore_cmds);
+  bool MigrateStream(const rocksdb::Slice &key, const StreamMetadata &metadata, std::string *restore_cmds);
```
caipengbo, line 133: By the way, this type of functions uses `bool` as its return value. I think it's a more elegant way to return `Status`, and move the LOG information from these internal functions to `Status` Msg, and return it to the upper level.

caipengbo (approved):

src/cluster/slot_migrate.h
PragmaTwice: We can refactor them in further PRs : )

PragmaTwice:
Thanks all. Merging...

PR #1198, opened 2022-12-20
Inplace construct error message in StatusOr while data type is large enough

PragmaTwice:
We currently use `std::unique_ptr<std::string>` in `StatusOr<T>` to represent the error message (in the union type: `T` or error message), in order to make the error message as small as possible. Hence the size of `StatusOr<int>` will be equal to `2 * sizeof(void*)` rather than `sizeof(void*) + sizeof(std::string)`, where `sizeof(std::string)` is `32` in libstdc++ (so basically `16` vs `40`).
But if `sizeof(T)` is large than `sizeof(std::string)` by itself, it is unnecessary to construct the error message on heap for a more compact stack size (since it will never be more compact). Thus we can construct the error message in StatusOr in-place while data type is large enough (i.e. `sizeof(T)` >= `sizeof(std::string)`).

### DIFF
--- a/src/common/status.h
+++ b/src/common/status.h
@@ -105,11 +105,61 @@ class [[nodiscard]] Status {
   friend struct StatusOr;
 };
 
+namespace type_details {
 template <typename... Ts>
-using first_element = typename std::tuple_element<0, std::tuple<Ts...>>::type;
+using first_element_t = typename std::tuple_element_t<0, std::tuple<Ts...>>;
 
 template <typename T>
-using remove_cvref_t = typename std::remove_cv<typename std::remove_reference<T>::type>::type;
+using remove_cvref_t = typename std::remove_cv_t<typename std::remove_reference_t<T>>;
+};  // namespace type_details
+
+template <typename, typename = void>
+struct StringInStatusOr : private std::string {
+  using base_type = std::string;
+  static constexpr bool inplace = true;
+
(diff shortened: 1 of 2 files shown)


src/common/status.h
```diff
@@ -105,11 +105,61 @@ class [[nodiscard]] Status {
...
-using remove_cvref_t = typename std::remove_cv<typename std::remove_reference<T>::type>::type;
+using remove_cvref_t = typename std::remove_cv_t<typename std::remove_reference_t<T>>;
+};  // namespace type_details
+
+template <typename, typename = void>
+struct StringInStatusOr : private std::string {
+  using base_type = std::string;
+  static constexpr bool inplace = true;
+
+  explicit StringInStatusOr(std::string&& v) : base_type(std::move(v)) {}
```
torwig, line 121: Across the `Kvrocks` codebase, we place `&` and `*` near the value instead of near the type. So in order to keep the same code style, you can place them close to variables or the `operator` keyword. However, if there is an exception for templates or clang-format tool forces such syntax, it's OK.

PragmaTwice: I tried to change it to `std::string &&v`, but clang-format changed it back.

torwig (approved):
LGTM

PragmaTwice:
Thanks all. Merging...

PR #1199, opened 2022-12-20
Add new lines after every section in Server::GetInfo

kukey:
after every section add new line


torwig (approved):
LGTM.
Personally, I prefer `if (section_cnt++ > 0)` instead of `if (section_cnt++)` but perhaps in this particular case it's not a big deal and even slightly reduces a line length.

PragmaTwice (approved):

git-hulk (approved):
LGTM, good catch!

PR #1201, opened 2022-12-21
Fetch dependencies from local archives in x.py

PragmaTwice:


tisonkun (approved):
LGTM. Two comments here:

1. We can use [context manager](https://realpython.com/python-with-statement/) for `LocalArchiveServer` to write RAII alike code in `with` block.
2. Do you have a real-world use case that requires this feature? We should somehow reduce the complexity unless it's necessary.

PragmaTwice:
> Do you have a real-world use case that requires this feature? We should somehow reduce the complexity unless it's necessary.

Some users have complained that they now have difficulty compiling kvrocks without internet connection, although I do not see this as a big problem.

tisonkun:
@PragmaTwice I suggest you involve those users directly here to test the implementation, since such local archive server is a customized syntax to use.

tisonkun:
Also, we have already ghproxy to use. Perhaps it's for company internal repositories?

PragmaTwice:
> @PragmaTwice I suggest you involve those users directly here to test the implementation, since such local archive server is a customized syntax to use.

Good advice, I will follow it.

tisonkun:
Closing for now. I tend to keep the logic simple unless there are real requirements and use cases.

From a technical perspective, we _can_ support many things, but from an engineering perspective, it's always important to reduce complexity (entropy).

PR #1206, opened 2022-12-26
Adapt kvrocks with ToplingDB

rockeet:
[ToplingDB](https://github.com/topling/toplingdb) is compatible with RocksDB and provide higher performance, [WebView](https://github.com/topling/rockside/wiki/WebView), native support to prometheus, secondary instance on shared storage ...

ToplingDB distributed compaction is only supported for `metadata`, can not be supported for other data because CompactionFilter need to access DB and calling `DB::Get`.

## 1. Build with ToplingDB
By default ToplingDB is not enabled, to compile kvrocks with ToplingDB, following instructions below:

## 1.1 Build ToplingDB
```bash
# ubuntu
sudo apt-get install -y git libjemalloc-dev libaio-dev libgflags-dev zlib1g-dev libbz2-dev libcurl4-gnutls-dev liburing-dev
# centos
sudo yum -y install git libaio-devel gcc-c++ gflags-devel zlib-devel bzip2-devel libcurl-devel liburing-devel

git clone https://github.com/topling/toplingdb
cd toplingdb
make -j`nproc` DEBUG_LEVEL=0 PREFIX=/path/to/toplingdb/install/dir install
```

## 1.2 Build kvrocks with ToplingDB
```bash
# build with ToplingDB, you should build and install ToplingDB first
export LD_LIBRARY_PATH=/path/to/toplingdb/install/dir/lib:${LD_LIBRARY_PATH}
env TOPLINGDB_HOME=/path/to/toplingdb-repo TOPLINGDB_LIBDIR=/path/to/toplingdb/install/dir/lib \
    ./x.py build -j`nproc` -DCMAKE_BUILD_TYPE=Release

# copy html files to kvrocks's ToplingDB document root, this makes webview pretty
cp /path/to/toplingdb-repo/sideplugin/rockside/src/topling/web/{index.html,style.css} \
   /path/to/kvrocks/toplingdb/document-root
```

## 2. Run with ToplingDB
env var `TOPLING_CONF` specifies ToplingDB json/yaml config which instructs kvrocks to run with ToplingDB.

If env var `TOPLING_CONF` is not specified, kvrocks will run without ToplingDB sideplugin and WebView...

User should change datadir in json config.

```bash
$ export LD_LIBRARY_PATH=/path/to/toplingdb/install/dir/lib:${LD_LIBRARY_PATH}
$ env TOPLING_CONF=kvtopling-community.json ./build/kvrocks -c kvrocks.conf
```

## 3. Run with ToplingDB shared storage secondary instance
Secondary instance on shared storage sharing same files of primary instance and keep catch up with primary.

```bash
$ env TOPLING_CONF=kvtopling-community-2nd.json ./build/kvrocks -c kvrocks.conf
```


tisonkun (changes requested):
Topling's [LICENSE](https://github.com/topling/toplingdb/blob/a04132fd83ef08932e5279f24e46ef0be79a9031/LICENSE.Apache) is in [Category X](https://www.apache.org/legal/resolved.html#category-x) from the ASF License Policy. So we cannot integrate it within the Apache release.

That said, you can still provide the integration within topling's org and make necessary interface refactoring in the upstream (support pluggable storage backend, for example), and even send a blog to show the integration is supported by third-party at https://kvrocks.apache.org/blog.

tisonkun:
Closing...

See https://github.com/apache/incubator-kvrocks/pull/1206#pullrequestreview-1229819085.

rockeet:
Benchmark with wikipedia data: **109GB**, **38.5 M** rows, average **2.8 KB** per row.

Hardware: aliyun 2c16g, in random read, we also tested with 8c16g, disk is high performance ESSD(iops up to 60k).

RAM is 16G to make it can not fit all (compressed) data.

**community** conf [kvtopling-community.json](https://github.com/rockeet/incubator-kvrocks/blob/adapt-toplingdb/kvtopling-community.json) using BlockBasedTable(block size is 8K).
**enterprise** conf [kvtopling-enterprise.json](https://github.com/rockeet/incubator-kvrocks/blob/adapt-toplingdb/kvtopling-enterprise.json) using ToplingZipTable.

image

image

Random read after full compaction, in random read, it is CPU bounded, rand 8c means 8c16g.

block cache is 1G for BlockBasedTable, we have tried block cache 4G & 8G, but performance getting worse.
image

rockeet:
> Topling's [LICENSE](https://github.com/topling/toplingdb/blob/a04132fd83ef08932e5279f24e46ef0be79a9031/LICENSE.Apache) is in [Category X](https://www.apache.org/legal/resolved.html#category-x) from the ASF License Policy. So we cannot integrate it within the Apache release.
> 
> That said, you can still provide the integration within topling's org and make necessary interface refactoring in the upstream (support pluggable storage backend, for example), and even send a blog to show the integration is supported by third-party at https://kvrocks.apache.org/blog.

Now ToplingDB license has updated to standard Apache 2.0: [license](https://github.com/topling/toplingdb#license)

git-hulk:
Hi @rockeet 

Great thanks for your benchmark data and impressive work on the toplingdb. I have some questions about the benchmark data:

For chart 1, does the build speed mean the rocksdb SST building key-value speed?

For chart 3, how can we explain the rand key per second? and how it's related to the QPS.

rockeet:
> Hi @rockeet
> 
> Great thanks for your benchmark data and impressive work on the toplingdb. I have some questions about the benchmark data:
> 
> For chart 1, does the build speed mean the rocksdb SST building key-value speed?
> 
> For chart 2, how can we explain the rand key per second? and how it's related to the QPS.

My fault, I should have write the details:

For chart 1, it is KV pair per second from redis client.

For chart 3, it is 16 threads(16 client connections) in which each thread use `mget` with 64 keys per request.

git-hulk:
Thanks for your clarification.

wacdev:
https://github.com/wacfork/incubator-kvrocks/tree/unstable/docker-compose#readme
kvrocks + rocksdb / topling 社区版 性能测试对比

dongdongwcpp:
> For chart 3, it is 16 threads(16 client connections) in which each thread use `mget` with 64 keys per request.

hi , for chart 3,any idea why when blockcache was 4G & 8G, but performance will get worse.

rockeet:
> > For chart 3, it is 16 threads(16 client connections) in which each thread use `mget` with 64 keys per request.
> 
> hi , for chart 3,any idea why when blockcache was 4G & 8G, but performance will get worse.

I think it should because larger block cache evict OS page cache memory, thus page cache hit rate becomming lower, and OS page cache data is compressed, which can hold more data in same memory.

rockeet:
Now distributed compaction is supportted for `metadata`, which support `strings` data type and `main key` for other data types.

distributed compaction for subkey is not supportted, which needs `DB::Get`, thus can not be efficiently supportted.

PR #1215, opened 2023-01-02
Allow to enable the async_io option to improve the performance

xiaobiaozhao:
Here are some comparisons between `async_io` and `normal`

## big value 500B

image

## small value 50B

image

## monitor diff
The green line is kvrocks compiled by async io，this pr.
`ts*` command using seek function to scan data from db.
image

## summarize

|        | cpu  | mem  | latency  | qps  |     |
| ------ | ---- | ---- | -------- | ---- | --- |
| async  | 110% | 100% | 100%     | 120% |     |
| normal | 100% | 100% | 130-400% | 100% |     |

`async_io` is better in seek read, higher qps, lower latency, and slightly higher cpu usage. Perfect for introducing projects

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -107,7 +107,7 @@ log-level info
 #
 # if set to -1, that means to disable the log cleaner.
 # if set to 0, all previous INFO level logs will be immediately removed.
-# if set to between 0 to INT_MAX, that means it will retent latest N(log-retention-days) day logs. 
+# if set to between 0 to INT_MAX, that means it will retent latest N(log-retention-days) day logs.
 
 # By default the log-retention-days is -1.
 log-retention-days -1
@@ -725,6 +725,13 @@ rocksdb.max_bytes_for_level_base 268435456
 #
 # Default: 10
 rocksdb.max_bytes_for_level_multiplier 10
+
+# If yes, RocksDB will prefetch some of data asynchronously.
+# RocksDB apply it if reads are sequential and its internal automatic
+# prefetching.
+
+# Deafult yes
+rocksdb.read_options.async_io yes
 
 # If yes, the write will be flushed from the operating system
 # buffer cache before the write is considered complete.
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -730,8 +730,8 @@ rocksdb.max_bytes_for_level_multiplier 10
 # RocksDB apply it if reads are sequential and its internal automatic
 # prefetching.
 
-# Default yes
-rocksdb.read_options.async_io yes
+# Default no
+rocksdb.read_options.async_io no
 
 # If yes, the write will be flushed from the operating system
 # buffer cache before the write is considered complete.
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -730,7 +730,7 @@ rocksdb.max_bytes_for_level_multiplier 10
 # RocksDB apply it if reads are sequential and its internal automatic
 # prefetching.
 
-# Deafult yes
+# Default yes
 rocksdb.read_options.async_io yes
 
 # If yes, the write will be flushed from the operating system
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -726,9 +726,10 @@ rocksdb.max_bytes_for_level_base 268435456
 # Default: 10
 rocksdb.max_bytes_for_level_multiplier 10
 
-# If yes, RocksDB will prefetch some of data asynchronously.
-# RocksDB apply it if reads are sequential and its internal automatic
-# prefetching.
+# This feature only takes effect in Iterators and MultiGet.
+# If yes, RocksDB will try to read asynchronously and in parallel as much as possible to hide IO latency.
+# In iterators, it will prefetch data asynchronously in the background for each file being iterated on. 
+# In MultiGet, it will read the necessary data blocks from those files in parallel as much as possible.
 
 # Default no
 rocksdb.read_options.async_io no
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -212,6 +212,8 @@ Config::Config() {
       {"rocksdb.write_options.low_pri", true, new YesNoField(&RocksDB.write_options.low_pri, false)},
       {"rocksdb.write_options.memtable_insert_hint_per_batch", true,
        new YesNoField(&RocksDB.write_options.memtable_insert_hint_per_batch, false)},
+      /* rocksdb read options */
+      {"rocksdb.read_options.async_io", false, new YesNoField(&RocksDB.read_options.async_io, true)},
   };
   for (auto &wrapper : fields) {
     auto &field = wrapper.field;
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -214,7 +214,7 @@ Config::Config() {
        new YesNoField(&RocksDB.write_options.memtable_insert_hint_per_batch, false)},
 
       /* rocksdb read options */
-      {"rocksdb.read_options.async_io", false, new YesNoField(&RocksDB.read_options.async_io, true)},
+      {"rocksdb.read_options.async_io", false, new YesNoField(&RocksDB.read_options.async_io, false)},
   };
   for (auto &wrapper : fields) {
     auto &field = wrapper.field;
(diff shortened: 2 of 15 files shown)


git-hulk:
Cool, I think it'd be better if we have the option to let users determine whether to enable it or not.

torwig:
Since this option is marked as `Experimental` in the `rocksdb` codebase, I agree to make this option configurable.
@xiaobiaozhao Thank you for the benchmarks. Nice job!

kvrocks.conf
```diff
@@ -721,6 +721,13 @@ rocksdb.max_bytes_for_level_base 268435456
 # Default: 10
 rocksdb.max_bytes_for_level_multiplier 10
 
+# If yes, RocksDB will prefetch some of data asynchronously.
+# RocksDB apply it if reads are sequential and its internal automatic
+# prefetching.
+
+# Deafult yes
+rocksdb.read_options.async_io yes
```
torwig: Perhaps, it would be better to set the default value to `false` because this feature is experimental in RocksDB.

git-hulk:
@xiaobiaozhao It's not right to involve the server in the storage layer. You should add the async_io flag in storage or implement like the write option https://github.com/apache/incubator-kvrocks/blob/unstable/src/storage/storage.cc#L76

kvrocks.conf
xiaobiaozhao: @git-hulk

kvrocks.conf
```diff
@@ -721,6 +721,13 @@ rocksdb.max_bytes_for_level_base 268435456
 # Default: 10
 rocksdb.max_bytes_for_level_multiplier 10
 
+# If yes, RocksDB will prefetch some of data asynchronously.
+# RocksDB apply it if reads are sequential and its internal automatic
+# prefetching.
+
+# Deafult yes
```
PragmaTwice: typo

xiaobiaozhao: done

kvrocks.conf
git-hulk: Agree with @torwig, can turn it to false by default and wait for this feature production ready. But we can encourage users to have a try.

src/config/config.cc
```diff
@@ -204,6 +204,8 @@ Config::Config() {
       {"rocksdb.write_options.low_pri", true, new YesNoField(&RocksDB.write_options.low_pri, false)},
       {"rocksdb.write_options.memtable_insert_hint_per_batch", true,
        new YesNoField(&RocksDB.write_options.memtable_insert_hint_per_batch, false)},
+      /* rocksdb read options */
+      {"rocksdb.read_options.async_io", false, new YesNoField(&RocksDB.read_options.async_io, true)},
```
```suggestion
      {"rocksdb.read_options.async_io", false, new YesNoField(&RocksDB.read_options.async_io, true)},
```

src/storage/storage.h
```diff
@@ -68,6 +68,7 @@ class Storage {
   ~Storage();
 
   void SetWriteOptions(const Config::RocksDB::WriteOptions &config);
+  void FillSeekReadOptions(rocksdb::ReadOptions &read_options);
```
git-hulk: I think we can rename `FillSeekReadOptions` => `SetReadOptions ` to keep consistent with `SetWriteOptions`.

src/config/config.cc
```diff
@@ -204,6 +204,9 @@ Config::Config() {
       {"rocksdb.write_options.low_pri", true, new YesNoField(&RocksDB.write_options.low_pri, false)},
       {"rocksdb.write_options.memtable_insert_hint_per_batch", true,
        new YesNoField(&RocksDB.write_options.memtable_insert_hint_per_batch, false)},
+
+      /* rocksdb read options */
+      {"rocksdb.read_options.async_io", false, new YesNoField(&RocksDB.read_options.async_io, true)},
```
torwig: Should it be YesNoField(&RocksDB.read_options.async_io, **false**) here?

src/storage/storage.cc
```diff
@@ -103,6 +103,11 @@ void Storage::SetWriteOptions(const Config::RocksDB::WriteOptions &config) {
   write_opts_.memtable_insert_hint_per_batch = config.memtable_insert_hint_per_batch;
 }
 
+void Storage::SetReadOptions(rocksdb::ReadOptions &read_options) {
```
torwig, line 106: It sounds like the `SetReadOptions` function sets the specified `read_options` to the whole `Storage` but the function doesn't do that. For me, the prior name of the function wasn't ideal but definitely more clear :)
The code you are trying to eliminate is repetitive but it's tricky to do that.
I recommend passing `read_options` by a pointer to stress that this parameter will be changed and can be considered as output.

src/config/config.cc
xiaobiaozhao: yes, I missed this

src/storage/storage.cc
xiaobiaozhao: According to the official document, mutil_get will also improve the performance. I think we can add this flag to all readoptions

torwig: I meant that the `Storage` class stores the `write_options` variable as its member, but the `ReadOptions` is passed to the iterator's constructor.

aleksraiden:
Good feature, any updates? In latest rocksdb update fixed some bugs in async_io support, so this PR is very good point to improve our performance.

src/storage/storage.cc
git-hulk: Yes, maybe `SetSeekOptions` is more precise.

git-hulk:
> Good feature, any updates? In latest rocksdb update fixed some bugs in async_io support, so this PR is very good point to improve our performance.

@xiaobiaozhao has updated the PR, to see if @torwig and @PragmaTwice have comments.

xiaobiaozhao:
I've tested mget and hgetall and async_io obviously gives hgetall a big boost.
If rocksdb has better updates in the future, we can continue to follow up.


image

image

torwig:
@xiaobiaozhao Good research.
If you update your branch, and make a final change as mentioned @git-hulk, I think we can merge this PR.

torwig (dismissed):
LGTM

kvrocks.conf
```diff
@@ -726,6 +726,13 @@ rocksdb.max_bytes_for_level_base 268435456
 # Default: 10
 rocksdb.max_bytes_for_level_multiplier 10
 
+# If yes, RocksDB will prefetch some of data asynchronously.
+# RocksDB apply it if reads are sequential and its internal automatic
+# prefetching.
```
PragmaTwice: > RocksDB apply it if reads are sequential and its internal automatic prefetching.

I cannot exactly understand the meaning of this comment, could you improve it?

```suggestion
# This feature only takes effect in Iterators and MultiGet.
# If yes, RocksDB will try to read asynchronously and in parallel as much as possible to hide IO latency.
# In iterators, it will prefetch data asynchronously in the background for each file being iterated on. 
# In MultiGet, it will read the necessary data blocks from those files in parallel as much as possible.
```

git-hulk: @PragmaTwice Does this comment look better?

PragmaTwice: Sure, it looks more clear to me.

git-hulk (approved):

PragmaTwice (approved):

git-hulk:
Thanks all, merging...

jishengming1:
Hi @xiaobiaozhao , Excuse me again, one more question.
Is the response and request order consistent at the same link? For example, send request 1, request 2 and expect to receive response 1, response 2 .

git-hulk:
@jishengming1 Yes, exactly. The order of the response is always exactly the same as the request.

jishengming1:
> Yes, exactly. The order of the response is always exactly the same as the request.

Thank you very much.

jishengming1:
> 
Excuse me again, one more question.
In the same link, if request 1 blocks, will all subsequent requests block? Is request2 affected by request1?

git-hulk:
Hi @jishengming1 I think we can move to the discussion to make the PR context clear.

jishengming1:
Okay, I'll go to the discussion forum to discuss this issue.

PR #1217, opened 2023-01-05
Make config load errors more friendly

### LINKED ISSUE #637: Enhancement: More user-friendly error message
### Search before asking

### Motivation

I get two kinds of error messages when I exec kvrock2redis like this:
* if the `data-dir` in config file is an unexist path,  it shows `Failed to load config, err: NotFound: `
* if the config file is not found,  it shows `Failed to load config, err: No such file or directory`

I think it need a more accurate error message to distinguish this two cases. 

### Solution

Maybe the simpliest way to solve this problem is to append `+ data_dir` after `s.ToString()`:
https://github.com/apache/incubator-kvrocks/blob/c391953f92fc03b4d22e100a42024a898f3fae47/tools/kvrocks2redis/config.cc#L143

torwig:
Enrich error messages that can occur during the loading of the configuration for both `Kvrocks` and `Kvrocks2Redis`.
Minor refactoring of edited source files.
Add a description of default values to `Kvrocks2Redis` configuration file (`cluster-enable` changed to `no` because in the source code, the default value is `false`).

Examples of new error messages
1. Kvrocks
```
Failed to load config. Error: failed to open file '/home/yaroslav/tests/kvrocks/kvrocks3.conf': No such file or directory
Failed to load config. Error: at line #L52: failed to set value of field 'cluster-enabled': argument must be 'yes' or 'no'
Failed to load config. Error: at line #L332: compaction-checker-range is invalid: invalid range format, the range should be between 0 and 24
Failed to load config. Error: at line #L24: malformed config line: config line ends unexpectedly in quoted string
```

2. Kvrocks2Redis
```
Failed to load config. Error: the specified Kvrocks working directory './kvrocks-data/' doesn't exist
Failed to load config. Error: at line #L22: Kvrocks port value should be between 0 and 65535
Failed to load config. Error: failed to open file '/home/yaroslav/tests/kvrocks/kvrocks2redis2.conf': No such file or directory
```

Closes: #637

### DIFF
--- a/utils/kvrocks2redis/kvrocks2redis.conf
+++ b/utils/kvrocks2redis/kvrocks2redis.conf
@@ -30,7 +30,7 @@ kvrocks 127.0.0.1 6666
 # Enable cluster mode.
 #
 # Default: no
-cluster-enable yes
+cluster-enable no
 
 ################################ NAMESPACE AND Sync Target Redis #####################################
 # Synchronize the specified namespace data to the specified Redis DB.
--- a/utils/kvrocks2redis/kvrocks2redis.conf
+++ b/utils/kvrocks2redis/kvrocks2redis.conf
@@ -2,18 +2,24 @@
 
 # The value should be INFO, WARNING, ERROR, FATAL.
 #
-# Default is INFO
+# Default: INFO
 loglevel INFO
 
 # By default kvrocks2redis does not run as a daemon. Use 'yes' if you need it.
 # Note that kvrocks2redis will write a pid file in /var/run/kvrocks2redis.pid when daemonized.
+#
+# Default: no
 daemonize no
 
 # The kvrocks working directory.
 # Note that you must specify a directory here, not a file name.
+#
+# Default: ./data
 data-dir ./data
 
 # Intermediate files are output to this directory when the kvrocks2redis program runs.
+#
+# Default: ./
 output-dir ./
 
 # Sync kvrocks node. Use the node's Psync command to get the newest wal raw write_batch.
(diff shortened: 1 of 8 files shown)


git-hulk (dismissed):

utils/kvrocks2redis/kvrocks2redis.conf
```diff
@@ -23,7 +29,7 @@ kvrocks 127.0.0.1 6666
 
 # Enable cluster mode.
 #
-# Default: yes
+# Default: no
 cluster-enable yes
```
PragmaTwice: Is there a typo?

torwig: @PragmaTwice You mean the default is `no` but the actual value in the config file is `yes`? I was thinking about the change. Should I change it to `no`?

PragmaTwice: Yeah but I am not familiar with these components. cc @git-hulk

git-hulk: This configuration should be meanless to kvrocks2redis

git-hulk: Yes, the default value should be `no`

git-hulk (approved):

utils/kvrocks2redis/kvrocks2redis.conf
torwig: Changed.

torwig:
It seems like the `cluster-enable` config option isn't used across the `kvrocks2redis` codebase. But I'll double-check it later and do the refactoring if it's true (in another PR).

PragmaTwice (approved):

git-hulk:
> It seems like the `cluster-enable` config option isn't used across the `kvrocks2redis` codebase. But I'll double-check it later and do the refactoring if it's true (in another PR).

It's used when decoding the key, coz the key should have the slot prefix if the cluster mode was enabled.

torwig:
Thanks all. Merging...

PR #1219, opened 2023-01-07
Persist the cluster nodes info after applying the cluster topology

### LINKED ISSUE #1021: Persist the cluster topology after applying
### Search before asking

### Motivation

Currently, Kvrocks didn't persist the cluster nodes topology and it needs to resync every time after restarting which it's not very convenient. I think we can store and reload it in the local configuration.

### Solution

_No response_

git-hulk:
This closes #1021 

Currently, the cluster nodes' info is only stored in memory and we need to re-sync the cluster topology after restarting. It's very inconvenient and confusing for most users.

## Solutoin

* Persist the cluster nodes' info in the local disk if the topology was changed. The file location is `{config->dir}/nodes.conf` and the format is below:
```
version 2
id 07c37dfeb235213a872192d90877d0cd55635b92
node 07c37dfeb235213a872192d90877d0cd55635b91 127.0.0.1 59129 master -  1-2 4-8193 10000 10002-11002 16381-16383
node 07c37dfeb235213a872192d90877d0cd55635b92 127.0.0.1 59137 master -  0
```
* Load and parse the `nodes.conf` if the cluster mode is enabled and the nodes file is exists

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -22,6 +22,7 @@
 
 #include <algorithm>
 #include <cstring>
+#include <fstream>
 #include <memory>
 
 #include "commands/redis_cmd.h"
@@ -41,6 +42,8 @@ const char *errNoMasterNode = "The node isn't a master";
 const char *errClusterNoInitialized = "CLUSTERDOWN The cluster is not initialized";
 const char *errInvalidClusterNodeInfo = "Invalid cluster nodes info";
 const char *errInvalidImportState = "Invalid import state";
+
+enum class LoadState { Unknown, Version, ID, NodesInfo };
 
 ClusterNode::ClusterNode(std::string id, std::string host, int port, int role, std::string master_id,
                          std::bitset<kClusterSlots> slots)
@@ -487,29 +490,7 @@ Status Cluster::GetClusterNodes(std::string *nodes_str) {
 }
 
 std::string Cluster::GenNodesDescription() {
-  // Generate slots info firstly
-  int start = -1;
-  std::shared_ptr<ClusterNode> n = nullptr;
-  for (int i = 0; i <= kClusterSlots; i++) {
-    // Find start node and slot id
-    if (n == nullptr) {
-      if (i == kClusterSlots) break;
-      n = slots_nodes_[i];
-      start = i;
-      continue;
-    }
-    // Generate slots info when occur different node with start or end of slot
-    if (i == kClusterSlots || n != slots_nodes_[i]) {
-      if (start == i - 1) {
-        n->slots_info_ += fmt::format("{} ", start);
-      } else {
-        n->slots_info_ += fmt::format("{}-{} ", start, i - 1);
-      }
-      if (i == kClusterSlots) break;
-      n = slots_nodes_[i];
-      start = i;
-    }
-  }
+  UpdateSlotsInfo();
 
   std::string nodes_desc;
   for (const auto &item : nodes_) {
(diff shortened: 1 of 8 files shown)


src/config/config.h
```diff
@@ -194,6 +194,7 @@ struct Config {
   mutable std::mutex backup_mu_;
 
  public:
+  const std::string NodesFilePath();
```
torwig: Did you mean `std::string NodesFilePath() const;`? Because, in my opinion, it makes no sense to return std::string by value and apply const-modifier to that value.

src/cluster/cluster.cc
```diff
@@ -511,37 +552,118 @@ std::string Cluster::GenNodesDescription() {
...
+  } else if (in == "nodes") {
+    return LoadState::NodesInfo;
+  } else {
+    return LoadState::Unknown;
+  }
+}
+
+Status Cluster::LoadClusterNodes(const std::string &file_path) {
+  if (rocksdb::Env::Default()->FileExists(file_path).IsNotFound()) {
+    return Status::OK();
```
torwig, line 609: Probably we can add the following log message before returning in order to report what is going on:

`LOG(INFO) << fmt::format("File '{}' not found. The cluster topology is not defined. Use CLUSTERX subcommands to specify it.");`

src/cluster/cluster.cc
```diff
@@ -511,37 +552,118 @@ std::string Cluster::GenNodesDescription() {
...
+  }
+
+  std::ifstream file;
+  file.open(file_path);
+  if (!file.is_open()) return Status::FromErrno();
```
torwig: Here we can return a more verbose error message:

`return {Status::NotOK, fmt::format("error opening the file '{}': {}", file_path, strerror(errno))};`

src/config/config.h
git-hulk: Yes, you're right, let's remove the const modifier.

src/cluster/cluster.cc
git-hulk: 👍 This message is better.

src/commands/redis_cmd.cc
```diff
@@ -5304,6 +5308,9 @@ class CommandClusterX : public Commander {
     } else {
       *output = Redis::Error("Invalid cluster command options");
     }
+    if (needPersistNodesInfo) {
```
torwig: @git-hulk What do you think about hiding this "knowledge" inside `Cluster`? I mean that the `Cluster` class calls the `DumpClusterNodes` function inside ` SetNodeId`/`SetSlot`/`SetClusterNodes` and commands know nothing about `NodesFilePath` and all the internals about storing the cluster topology in a file, etc.

PragmaTwice:
I have some comments on the file format. 

I think it is a new file format that relies on what is in the comments (starting with `#`) as part of the parsing, which is puzzling. I think we can try not to create a new file format as well as a new parsing logic. If this file is named `*.conf` (it is weird to have two different file format in one program that both have suffix `.conf`), I think we could better refer to the previous kvrocks conf file format, i.e.

```
# a comment that does not affect parsing...
version 1
id 0123456789012345678901234567890123456789
node 07c37dfeb235213a872192d90877d0cd55635b91 127.0.0.1 63262 master -  0-2 4-8193 10000 10002-11002 16381-16383
node ...
```

And there are some util functions in `config/config_util.h` that can help parsing text in such file format.

src/commands/redis_cmd.cc
git-hulk: Yes, it makes sense. Will resolve it later.

git-hulk:
> I have some comments on the file format.
> 
> I think it is a new file format that relies on what is in the comments (starting with `#`) as part of the parsing, which is puzzling. I think we can try not to create a new file format as well as a new parsing logic. If this file is named `*.conf` (it is weird to have two different file format in one program that both have suffix `.conf`), I think we could better refer to the previous kvrocks conf file format, i.e.
> 
> ```
> # a comment that does not affect parsing...
> version 1
> id 0123456789012345678901234567890123456789
> node 07c37dfeb235213a872192d90877d0cd55635b91 127.0.0.1 63262 master -  0-2 4-8193 10000 10002-11002 16381-16383
> node ...
> ```
> 
> And there are some util functions in `config/config_util.h` that can help parsing text in such file format.

Yes, it'd be better to keep the same format, I will reconsider if it's other issues.

src/cluster/cluster.cc
mapleFU: Hi, when first start the cluster, we would meet that message, so it may confuse the user?

src/server/server.cc
```diff
@@ -137,12 +137,17 @@ Status Server::Start() {
   }
 
   if (config_->cluster_enabled) {
+    auto s = cluster_->LoadClusterNodes(config_->NodesFilePath());
```
mapleFU, line 140: Hi hulk, the reload make kvrocks a bit more trickey, if:
1. a server(A) mount on a disk crashed
2. user add some machines, and update the cluster version
3. A restarted, load the file, but has lower version. Previously, it can recover and load from other machines, but now, user need drop all it's data.

How can we handle this problem?

mapleFU (commented):
And still I have two questions:
1. do we have spec on the conf? @PragmaTwice 's advice is ok, and why don't we using something like JSON? It's our standard compatible with other format?
2. Should we ignore some "stale" config? What should we do if a server is crashed?

src/server/server.cc
git-hulk: Hi @mapleFU, thanks for your warm review. 

Currently, the cluster topology is expected to maintain out of the Kvrocks node(Like kvrocks-controller) and the external component will re-sync the cluster topology if it's changed. 

So for question 1/2, it's not a problem since the external component will re-sync if it's changed or lost.

For question 3, we can regard this topology as a local cache, the external controller component will re-sync the cluster topology and won't depend on this persistent topology.

Briefly, this PR is only to cache the cluster topo in the local disk, so it will bring the side effect is the user may get stale data if the client cached the topology and the controller failed to update the topology in time. But I think it's better than before.

mapleFU: @git-hulk Let's dive into the code:
1. Assume we write a lower version(version = 1) on Server A
2. Server write and crash, topology changed, version = 2.
3. Restart, and load with version = 1. The file is ok, then we call `SetClusterNodes` with `force = false`. So, return status not ok. So `Start` will failed

mapleFU: @git-hulk Let's dive into the code:
1. Assume we write a lower version(version = 1) on Server A
2. Server write and crash, topology changed, version = 2.
3. Restart, and load with version = 1. The file is ok, then we call `SetClusterNodes` with `force = false`. So, return status not ok. So `Start` will failed

mapleFU: So, maybe we should ignore error if `Load` is failed?

git-hulk: The load action will run before the server start, so it won't have any cluster version from the time point. The external controller only can re-sync the new topology after the server was started.

mapleFU: Okay, thanks for your patient explaination. But I still think we should ignore it's error:

```
      case LoadState::NodesInfo:
        // std::getline swallowed the newline in the node info, so need to add it back here
        nodesInfo.append(line + "\n");
        break;
```

I think write to os is not atomic, so, we may got a "teared" file if we don't have any checksum. And, `Load` will failed if parsing failed.

mapleFU: Okay, thanks for your patient explaination. But I still think we should ignore it's error:

```
      case LoadState::NodesInfo:
        // std::getline swallowed the newline in the node info, so need to add it back here
        nodesInfo.append(line + "\n");
        break;
```

I think write to os is not atomic, so, we may got a "teared" file if we don't have any checksum. And, `Load` will failed if parsing failed.

git-hulk: Yes, that's a good point. I think we can simply write in memory and flush them at once.

mapleFU: @git-hulk unless you are writing whole block to the device, you cannot make sure that multiple blocks are atomic sync to OS. Think about the cases below:

1. Write to OS: BLK1 ... BLKn
2. BLK1 - BLKn-1 is sync to disk. But BLKn not
3. System crashed

So, the most convenient way is using RocksDB to store it, or write a checksum.

mapleFU: Using OS `rename` API is also possible here. Maybe we can write a atomic-write using `rename`

git-hulk: @mapleFU oh, I forgot that it's written to the temp file and renamed after the finish. So it has not this issue.

src/cluster/cluster.cc
```diff
@@ -487,8 +490,46 @@ Status Cluster::GetClusterNodes(std::string *nodes_str) {
...
+
+    // Flags
+    if (n->id_ == myid_) node_str.append("myself,");
+    if (n->role_ == kClusterMaster) {
+      node_str.append("master - ");
+    } else {
+      node_str.append("slave " + n->master_id_ + " ");
+    }
+
+    // Ping sent, pong received, config epoch, link status
```
mapleFU, line 513: Why do we need to `GetTimeStampMS` every time? Is there some heavy api spending a lot of time?

src/cluster/cluster.cc
```diff
@@ -487,8 +490,46 @@ Status Cluster::GetClusterNodes(std::string *nodes_str) {
...
+    nodes_desc.append(node_str + "\n");
+  }
+  return nodes_desc;
+}
+
+void Cluster::UpdateSlotsInfo() {
   int start = -1;
+  // reset the previous slots info
+  for (const auto &item : nodes_) {
+    const std::shared_ptr<ClusterNode> n = item.second;
```
mapleFU: why not using `const std::shared_ptr<ClusterNode>& n = item.second;`

src/cluster/cluster.cc
```diff
@@ -487,8 +490,46 @@ Status Cluster::GetClusterNodes(std::string *nodes_str) {
 }
 
 std::string Cluster::GenNodesDescription() {
-  // Generate slots info firstly
+  UpdateSlotsInfo();
+
+  std::string nodes_desc;
+  for (const auto &item : nodes_) {
+    const std::shared_ptr<ClusterNode> n = item.second;
```
mapleFU, line 498: why not using `const std::shared_ptr<ClusterNode>& n = item.second;`

src/server/server.cc
mapleFU: Seems ok. It was my fault, `ofstream::close` will flush the buffer and close the fd. So it's safe to rename it.

What will happen if a crashed server thinks itself is a master node? Will it harm the cluster?

src/commands/redis_cmd.cc
mapleFU: Seems that command will failed if `DumpClusterNodes` failed. However, the cluster info is already updated.

mapleFU (commented):
If any bad record or new format is supported, the code running on old server would reject the format, and failed to start.

So, why do we failed when `LoadClusterNodes` failed?

src/cluster/cluster.cc
```diff
@@ -511,37 +552,122 @@ std::string Cluster::GenNodesDescription() {
...
+        if (id.length() != kClusterNodeIdLen) {
+          return {Status::NotOK, errInvalidNodeID};
+        }
+        break;
+      case LoadState::NodesInfo:
+        // std::getline swallowed the newline in the node info, so need to add it back here
+        nodesInfo.append(line + "\n");
+        break;
+      default:
+        return {Status::NotOK, "got unknown parse state"};
```
mapleFU: In the future, if we support other states, seems load would fail?

src/cluster/cluster.cc
git-hulk: Good catch. This code is extracted from the previous implementation, it should not get time every time.

src/cluster/cluster.cc
git-hulk: Yes, can do it.

src/cluster/cluster.cc
git-hulk: It should make sense that the older version can't parse new state.

src/cluster/cluster.cc
git-hulk: ditto

src/cluster/cluster.cc
mapleFU: > It should make sense that the older version can't parse new state.

It's ok but, if a user want to rollback(maybe because some other reason), and the protocol is updated. He will failed to start, unless he delete the file.

src/commands/redis_cmd.cc
git-hulk: re @mapleFU

Yes, cannot promise the atomic now.

git-hulk: re @torwig, I recalled why to specify the path here. It will convenient for adding test cases without depending on server->config, as well as avoid modifying a lot of previous test cases.

src/cluster/cluster.cc
git-hulk: Yes, I have no strong point in this scenario. My initial intention is to prevent unexpected configurations.

git-hulk: To see other folks have any advice? @PragmaTwice @ShooterIT @torwig

torwig (approved):

torwig:
I have a suspicion that all the members of the `Cluster` instance are not thread-safe, e.g. simultaneous calls of `SetClusterNodes` and `GetClusterNodes` can lead to a data race. Am I right?

git-hulk:
> I have a suspicion that all the members of the `Cluster` instance are not thread-safe, e.g. simultaneous calls of `SetClusterNodes` and `GetClusterNodes` can lead to a data race. Am I right?

Yes, we should reconsider making those commands exclusive.

ShooterIT:
all cluster  writing commands have `exclusive`, so they are safe now.

ShooterIT:
is there a way which allow we persist the cluster topology into some rocksdb CF?

ShooterIT:
another issue, since after rebooting, replica would load topology and try to sync with master, so when master receive the sync request from slave, master should check if the replica is in its current cluster topology, if yes, master allow, if no, it should reject.

git-hulk:
> another issue, since after rebooting, replica would load topology and try to sync with master, so when master receive the sync request from slave, master should check if the replica is in its current cluster topology, if yes, master allow, if no, it should reject.

We have no node id or port in the replication process, so it cannot identify the replica and reject it on the master side. But we can compare the cluster version before connecting on the replica side.

git-hulk:
> all cluster writing commands have exclusive, so they are safe now.

My bad, I forget the `cluster` and `clusterx` will be in exclusive mode even if it has no exclusive flag.

git-hulk:
> is there a way which allow we persist the cluster topology into some rocksdb CF?

I think it's unnecessary, the local file will make it easier to modify or drop it manually.

PragmaTwice:
It seems CI was not triggered.  We can make an empty commit to retry.

git-hulk:
> It seems CI was not triggered. We can make an empty commit to retry.

ok

PragmaTwice (approved):

git-hulk:
@ShooterIT To make the context clear, I'll merge this PR first, then file another PR to enhance the replication.

git-hulk:
Thanks all, merging...

PR #1222, opened 2023-01-10
Kvrocks2Redis: decode slot ID in cluster mode

torwig:
Currently, `Kvrocks2Redis` doesn't decode the slot ID if a cluster mode is enabled.
For example, if we write to `Kvrocks` in a cluster mode the following value:
`set book1 "Anything here"`
And run `Kvrocks2Redis` with `cluster-enable` set to `yes`, we will get in `Redis` the following key:
`\x16\x9bbook1`
where `169b` is a hex representation of slot 5787.

The reason: the `Kvrocks` config key `slot_id_encoded` is not set. Setting only the `cluster_enable` config value is not enough for `Kvrocks2Redis`. In `Kvrocks` we have [this](https://github.com/apache/incubator-kvrocks/blob/unstable/src/config/config.cc#L357).

### DIFF
--- a/utils/kvrocks2redis/sync.cc
+++ b/utils/kvrocks2redis/sync.cc
@@ -155,7 +155,7 @@ Status Sync::incrementBatchLoop() {
   while (!IsStopped()) {
     if (evbuffer_read(evbuf, sock_fd_, -1) <= 0) {
       evbuffer_free(evbuf);
-      return {Status::NotOK, std::string("[kvrocks2redis] read increament batch err: ") + strerror(errno)};
+      return {Status::NotOK, std::string("[kvrocks2redis] read increment batch err: ") + strerror(errno)};
     }
     if (incr_state_ == IncrementBatchLoopState::Incr_batch_size) {
       // Read bulk length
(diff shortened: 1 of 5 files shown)


git-hulk (approved):

utils/kvrocks2redis/sync.cc
```diff
@@ -180,8 +180,12 @@ Status Sync::incrementBatchLoop() {
         if (bulk_data_str != "ping") {
           auto bat = rocksdb::WriteBatch(bulk_data_str);
           int count = static_cast<int>(bat.Count());
-          parser_->ParseWriteBatch(bulk_data_str);
-          auto s = updateNextSeq(next_seq_ + count);
+          auto s = parser_->ParseWriteBatch(bulk_data_str);
+          if (!s.IsOK()) {
+            return s.Prefixed(fmt::format("failed to parse write batch '{}'", Util::StringToHex(bulk_data_str)));
```
caipengbo, line 185: Does it help to print out the write batch? This is a long string, and I personally don't think it's very helpful for troubleshooting.

torwig: @caipengbo I've just recalled that we use the same approach in replication [(here)](https://github.com/apache/incubator-kvrocks/blob/unstable/src/cluster/replication.cc#L552) and thought that perhaps it was intended to help with debugging.

caipengbo: Get it, I did encounter this error in the replication, because there is a background error, replica can not be written in. I found that the string is very long, and it is not useful for troubleshoting problem, maybe you said, is used for debugging.

caipengbo (approved):

utils/kvrocks2redis/sync.cc
git-hulk: Yes, it can help us inspect the batch if it goes wrong, so it's fine.

torwig:
Merging...

PR #1225, opened 2023-01-15
Fix error message passing in Server::Start

PragmaTwice:


git-hulk (approved):

tisonkun (commented):
May you add a description on what is fixed and the different before and after?

PragmaTwice:
> May you add a description on what is fixed and the different before and after?


https://github.com/apache/incubator-kvrocks/blob/f1f7c10a30dc3387f6ffd77fe3c6efe2f393ed38/src/server/server.cc#L135

For example, before the PR, the error message in this line can never be dumped into log. So users can hardly know why the server fail to start.

tisonkun (approved):

tisonkun:
Merging...

src/main.cc
```diff
@@ -319,8 +319,7 @@ int main(int argc, char *argv[]) {
     uint32_t ports[] = {config.port, config.tls_port, 0};
     for (uint32_t *port = ports; *port; ++port) {
       if (Util::IsPortInUse(*port)) {
-        LOG(ERROR) << "Could not create server TCP since the specified port[" << *port << "] is already in use"
-                   << std::endl;
+        LOG(ERROR) << "Could not create server TCP since the specified port[" << *port << "] is already in use";
```
tisonkun, line 322: Does LOG already append endl?

PragmaTwice: Sure.

PR #1227, opened 2023-01-16
Add ParseFloat and replace stod with it

PragmaTwice:
We add `ParseFloat` to `parse_util.h` and replace most of `std::stod` with this newly added function in this PR.

### DIFF
--- a/src/commands/redis_cmd.cc
+++ b/src/commands/redis_cmd.cc
@@ -64,23 +64,23 @@
 
 namespace Redis {
 
-const char *kCursorPrefix = "_";
+constexpr const char *kCursorPrefix = "_";
 
-const char *errInvalidSyntax = "syntax error";
-const char *errInvalidExpireTime = "invalid expire time";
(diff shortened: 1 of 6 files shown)


torwig (dismissed):
LGTM

PragmaTwice:
I notice that gcc implemented floating-point version `std::from_chars` since 11.1, which is a sad story. It means we cannot use `std::from_chars` because it comes too late and the gcc version is too new for some users. So I will implement it using `strto*` in c std library instead.

PragmaTwice:
> I notice that gcc implemented floating-point version `std::from_chars` since 11.1, which is a sad story. It means we cannot use `std::from_chars` because it comes too late and the gcc version is too new for some users. So I will implement it using `strto*` in c std library instead.

done

git-hulk (approved):
LGTM

PragmaTwice:
Thanks all. Merging...

PR #1228, opened 2023-01-16
Fix deadlock when unblocking connection on stream

torwig:
I forgot to advance the iterator.
Which leads to an infinite loop. That eventually stops the execution of all writable commands.

Thanks to @aleksraiden for catching (and experiencing) this bug.


aleksraiden:
Thanks, @torwig 

Now all works fine!

PragmaTwice (approved):

git-hulk (approved):

PragmaTwice:
Thanks all. Merging...

PR #1232, opened 2023-01-24
XREAD returns dedicated status if it blocks

torwig:
In order to be consistent with such blocking commands like BLPOP and BRPOP.

### DIFF
--- a/src/commands/redis_cmd.cc
+++ b/src/commands/redis_cmd.cc
@@ -6079,7 +6079,7 @@ class CommandXRead : public Commander {
       evtimer_add(timer_, &tm);
     }
 
-    return Status::OK();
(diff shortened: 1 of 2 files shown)


PragmaTwice (commented):
LGTM, however, I wonder if there could be some test cases to identify this change (if it is not hard to add).

torwig:
@PragmaTwice No because it worked **correctly** even without this change :) 
In this case, the loop wasn't stopped on `s.Is<Status::BlockingCmd>()` clause. But, according to the code, the actual reply wasn't sent because the `reply` variable remained empty.

PragmaTwice (approved):
got it.

torwig:
Merging...

Issue #1146, opened 2022-11-28
Docker container exits with code 132 and no output


SamSpiri:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Version

Ubuntu 20.04.1 LTS \n \l
Linux home 5.4.0-122-generic #138-Ubuntu SMP Wed Jun 22 15:00:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
Docker version 19.03.5, build 633a0ea838

docker image apache/kvrocks:2.2.0

### Minimal reproduce step

$ docker run -i --rm apache/kvrocks:2.2.0
$ echo $?
132

### What did you expect to see?

1. Output logs with the reason why kvrocks exited
2. Kvrocks don't exit but works

### What did you see instead?

no logs, no output, just exit 132

### Anything Else?

tried tags nightly, 2.1.0 with same bad result.
but kvrocks/kvrocks:latest image did output version number and worked until i triggered CTRL-C:
```
$ docker run -i --rm kvrocks/kvrocks
Version: 2.0.6 @4e5883e9
^C
```

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PragmaTwice:
Hi, I am afraid kvrocks aborted due to the `Illegal Instruction` (128 + 4) signal, that is, the kvrocks executable binary in the docker image uses some instruction sets which is not supported by your cpu (in x64, it may be avx512 or something, you can use gdb to determine the concrete instruction which leads to termination).

I will investigate this problem, and you can build kvrocks on your local device to avoid this problem. (or build the docker image on your local.)

SamSpiri:
> not supported by your cpu 

I tried on intel and it also worked until i triggered CTRL-C,
but it didn't produce any output while running and then before exiting it prints all the logs.
That's my main concern, because i have no logs of kvrocks in my kubernetes.
```
$ docker run -i --rm apache/kvrocks:2.2.0
^CI20221128 17:10:47.559381     1 main.cc:306] kvrocks version 2.2.0 @17b31e7
I20221128 17:10:47.606953     1 storage.cc:326] [storage] Success to load the data from disk: 21 ms
I20221128 17:10:47.608379     1 worker.cc:67] [worker] Listening on: 0.0.0.0:6666
I20221128 17:10:47.608695     1 worker.cc:67] [worker] Listening on: 0.0.0.0:6666
[...]

PR #1233, opened 2023-01-27
Make docker image more portable

PragmaTwice:
It is a known issue that some users encountered "illegal instruction" crash while using our docker images, see #1146.

After investigation, RocksDB use `march=native` while the `PORTABLE` option is set to OFF.

We keep the default behavior for best performance, but use `PORTABLE=ON` in docker images to prevent illegal instructions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,16 +17,17 @@
 
 FROM ubuntu:focal as build
 
+ARG MORE_BUILD_ARGS
+
 # workaround tzdata install hanging
 ENV TZ=Asia/Shanghai
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
-RUN apt update
-RUN apt install -y git gcc g++ make cmake autoconf automake libtool python3 libssl-dev
+RUN apt update && apt install -y git gcc g++ make cmake autoconf automake libtool python3 libssl-dev
 WORKDIR /kvrocks
 
 COPY . .
-RUN ./x.py build -DENABLE_OPENSSL=ON -DPORTABLE=ON
+RUN ./x.py build -DENABLE_OPENSSL=ON -DPORTABLE=ON $MORE_BUILD_ARGS
 
 FROM ubuntu:focal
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN apt install -y git gcc g++ make cmake autoconf automake libtool python3 libs
 WORKDIR /kvrocks
 
 COPY . .
-RUN ./x.py build -DENABLE_OPENSSL=ON
+RUN ./x.py build -DENABLE_OPENSSL=ON -DPORTABLE=ON
```
torwig: Why portable is ON if the binary is built inside Ubuntu and is launching inside Ubuntu?
I guess arch-specific optimizations could be applied in this case because we are fully aware of the target platform (Ubuntu). Correct me if I'm wrong.

PragmaTwice: For example, if the build machine (Travis CI here) has avx512 instructions, the built binary will use these instructions to perform vectorize optimization if `-march=native` is ON. But maybe user's device does not support avx512.

tisonkun: I'd suggest adding an `ARG` and defaulting to `ON`.

Our image can always use the default but if users are certain for the platform they can build their own with `PORTABLE=OFF`.

tisonkun (commented):
Comment above.

Dockerfile
PragmaTwice: done. since other build arguments can also have this problem like ENABLE_SSL, DISABLE_JEMALLOC, so I add $MORE_BUILD_ARGS instead of $PORTABLE for more general purpose.

tisonkun (approved):
For confirmation, will `MORE_BUILD_ARGS` override `ENABLE_OPENSSL` and `PORTABLE` if specified?

git-hulk (approved):

PragmaTwice:
> For confirmation, will MORE_BUILD_ARGS override ENABLE_OPENSSL and PORTABLE if specified?

Yeah.

PragmaTwice:
Thanks all. Merging...

PR #1245, opened 2023-02-05
Remove useless buf and prealloc in InternalKey

PragmaTwice:
We remove `buf_` and `prealloc_` field in InternalKey since they are useless (we can write to the result string directly, these allocation and stack space are unnecessary).


torwig (approved):
LGTM

git-hulk (approved):

PragmaTwice:
Thanks all. Merging...

Issue #776, opened 2022-08-12
Report a failure fast if `CMAKE_OSX_DEPLOYMENT_TARGET` unset on MacOS


tisonkun:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

See also https://github.com/apache/incubator-kvrocks/commit/341a989d75783d8dad76bbd5a4473dd345472337#r80998035.

cc @PragmaTwice @xiaobiaozhao 

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

tanruixiang:
Hi. I am using MacBook M1. According to https://github.com/apache/incubator-kvrocks/commit/341a989d75783d8dad76bbd5a4473dd345472337#r80998035, If I replace `${CMAKE_OSX_DEPLOYMENT_TARGET}` with `11.0`, The compilation can be successful. I am not very familiar with cmake. Do you have time to repair it? I noticed that you don't have an apple device to test. I can compile and test for you. @PragmaTwice

PragmaTwice:
> Hi. I am using MacBook M1. According to [341a989#r80998035](https://github.com/apache/incubator-kvrocks/commit/341a989d75783d8dad76bbd5a4473dd345472337#r80998035), If I replace `${CMAKE_OSX_DEPLOYMENT_TARGET}` with `11.0`, The compilation can be successful. I am not very familiar with cmake. Do you have time to repair it? I noticed that you don't have an apple device to test. I can compile and test for you. @PragmaTwice

Hi @tanruixiang, actually you can directly pass the option to cmake, e.g. `cmake -DCMAKE_OSX_DEPLOYMENT_TARGET=10.3 ..`, `MACOSX_DEPLOYMENT_TARGET=10.3 cmake ..` or `./x.py build  -DCMAKE_OSX_DEPLOYMENT_TARGET=10.3` rather than modifying cmake files.

If you want to try to fix this problem, you can add an `if ... else ...` to CMakeLists, i.e. when it is currently in macos and `CMAKE_OSX_DEPLOYMENT_TARGET` is not configured, an error should be reported and the user will be reminded to configure it.

tanruixiang:
[...]

PR #1250, opened 2023-02-07
Automatically set the env MACOSX_DEPLOYMENT_TARGET to the current macOS version unless the env is already set

git-hulk:
Related issue: #776 

We now can automatically set the env MACOSX_DEPLOYMENT_TARGET to the current macOS version like below:

```
-- CMAKE_OSX_DEPLOYMENT_TARGET was set to: 12.5
```

This change can make developers who are using macOS happy.

### DIFF
--- a/cmake/luajit.cmake
+++ b/cmake/luajit.cmake
@@ -19,27 +19,42 @@ include_guard()
 
 include(cmake/utils.cmake)
 
-if((${CMAKE_SYSTEM_NAME} MATCHES "Darwin") AND (NOT CMAKE_OSX_DEPLOYMENT_TARGET))
+if ((${CMAKE_SYSTEM_NAME} MATCHES "Darwin") AND (NOT CMAKE_OSX_DEPLOYMENT_TARGET))
+  set(sw_cmd "sw_vers")
+  set(sw_arg "-productVersion")
+  execute_process(COMMAND ${sw_cmd} ${sw_arg}
+          WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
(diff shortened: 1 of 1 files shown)


cmake/luajit.cmake
```diff
@@ -19,27 +19,42 @@ include_guard()
...
+          WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+          OUTPUT_VARIABLE OUTPUT
+          RESULT_VARIABLE STATUS
+          OUTPUT_STRIP_TRAILING_WHITESPACE)
+  if (STATUS EQUAL 0)
+    string(REGEX MATCH "[0-9]+.[0-9]+"
+            DEPLOYMENT_TARGET ${OUTPUT})
+    set(CMAKE_OSX_DEPLOYMENT_TARGET "${DEPLOYMENT_TARGET}")
+    message(STATUS "CMAKE_OSX_DEPLOYMENT_TARGET was set to: ${DEPLOYMENT_TARGET}")
+  endif ()
```
PragmaTwice: CMake variables has no regular scope (in `if` or `include`), so I prefer more concrete names, e.g. `SW_CMD_OUTPUT` rather than `OUTPUT`, `SW_CMD_STATUS` rather than `STATUS`.

Otherwise we can access this `OUTPUT` variable anywhere after the code above is executed.

git-hulk: Good advice, will update

PragmaTwice (approved):
LGTM

tisonkun (approved):
Verified

git-hulk:
Thanks all, merging...

PR #1251, opened 2023-02-08
Add optional args to XLEN to specify start ID and direction

torwig:
Syntax: `XLEN stream_name [entry_id] [direction]`. 
Where direction can be '-' or '+' (special IDs mean respectively the first and the last elements in the stream). Default value: '+'. 

If an optional entry ID is specified, the XLEN command returns the number of entries between that ID and the first/last element of the stream. 

The element with the specified ID is not considered - it serves as an exclusive boundary. If the stream doesn't contain an entry with the specified ID, the command starts counting from the closest entry to the first/last element in the stream. 

If the specified ID is equal to or greater than the last entry ID and: 
a) direction is '+' - the command returns 0
b) direction is '-' - the command returns the stream's size 

If the specified ID is equal to or less than the first entry ID and: 
a) direction is '-' the command returns 0
b) direction is '+' the command returns stream's size 

This feature is helpful in finding:
a) how many entries in the stream were processed so far 
b) how many unprocessed entries  (ahead of the current one) are in the stream

Complexity: O(n)

### DIFF
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -213,8 +213,6 @@ class CommandXDel : public Commander {
 class CommandXLen : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    stream_name_ = args[1];
-
     if (args.size() > 2) {
       Redis::StreamEntryID id;
       auto s = Redis::ParseStreamEntryID(args[2], &id);
@@ -249,7 +247,6 @@ class CommandXLen : public Commander {
   }
 
  private:
-  std::string stream_name_;
   Redis::StreamLenOptions options_;
 };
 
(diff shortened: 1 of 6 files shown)


src/commands/cmd_stream.cc
```diff
@@ -212,10 +212,33 @@ class CommandXDel : public Commander {
 
 class CommandXLen : public Commander {
  public:
+  Status Parse(const std::vector<std::string> &args) override {
+    stream_name_ = args[1];
```
git-hulk: `stream_name_` looks unused.

git-hulk (commented):
LGTM, has only one comment. @torwig's test cases in PR are always impressive.

src/commands/cmd_stream.cc
torwig: @git-hulk Yes, you're right. I'll fix it right now.

git-hulk (approved):

PragmaTwice (approved):

PragmaTwice (approved):

torwig:
Thanks all. Merging...

Issue #1254, opened 2023-02-10
ASan/TSan does not work well with JeMalloc


xiaobiaozhao:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Version

Mac  & Ubuntu all versions

### Minimal reproduce step

`./x.py build --unittest --compiler gcc -DDISABLE_JEMALLOC=ON -DENABLE_TSAN=ON`
`./x.py build --unittest --compiler gcc -DDISABLE_JEMALLOC=ON -DENABLE_ASAN=ON`

### What did you expect to see?

build succeed

### What did you see instead?

./x.py build --unittest --compiler gcc -DDISABLE_JEMALLOC=ON -DENABLE_TSAN=ON
$ /usr/bin/cmake /root/work/incubator-kvrocks -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DENABLE_TSAN=ON
-- The CXX compiler identification is GNU 11.3.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/g++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:67 (message):
  **### ASan/TSan does not work well with JeMalloc**

-- Configuring incomplete, errors occurred!
See also "/root/work/incubator-kvrocks/build/CMakeFiles/CMakeOutput.log".
Traceback (most recent call last):
  File "/root/work/incubator-kvrocks/./x.py", line 397, in <module>
    args.func(**arg_dict)
  File "/root/work/incubator-kvrocks/./x.py", line 122, in build
    run(cmake, str(basedir), *cmake_options, verbose=True, cwd=dir)
  File "/root/work/incubator-kvrocks/./x.py", line 68, in run
    raise RuntimeError(err)
RuntimeError: 
failed to run: ('/usr/bin/cmake', '/root/work/incubator-kvrocks', '-DCMAKE_BUILD_TYPE=RelWithDebInfo', '-DCMAKE_C_COMPILER=gcc', '-DCMAKE_CXX_COMPILER=g++', '-DENABLE_TSAN=ON')
exit with code: 1

### Anything Else?

[...]

PR #1256, opened 2023-02-10
Use append action instead of nargs for cmake defs in x.py

PragmaTwice:
Currently, we use `nargs=*` for parsing `-D` arguments, e.g. `-D X=1 Y=2` -> `["X=1", "Y=2"]`, `-D X=1 -D Y=2` -> `["Y=2"]`.

We need to change it to `action=append` to align the behavior with CMake, e.g. `-D X=1 -D Y=2` -> `["X=1", "Y=2"]`.

It solved #1254 and fix all wrong argument parsing in GitHub Actions.


git-hulk (dismissed):

tisonkun (dismissed):
LGTM. May I ask the `disk_stats` related changes? We'd better avoid mixing changes, or at least describe what it is.

PragmaTwice:
> LGTM. May I ask the `disk_stats` related changes? We'd better avoid mixing changes, or at least describe what it is.

Previously, all ASan and TSan checks in CI have never been applied (only the last `-D` was kept, all other cmake defs were ignored), so I try to solve all ASan and TSan reports to make the CI pass.

tisonkun:
Thanks for your explanation! That's reasonable.

git-hulk (approved):

PragmaTwice:
Thanks all. Merging...

PR #1259, opened 2023-02-14
fix: license headers in third-party code

tisonkun:
Avoid adding ASF license header for files copied. And use public domain sha1 source, remove misleading "from valrgind" description.

### DIFF
--- a/.github/config/licenserc.yml
+++ b/.github/config/licenserc.yml
@@ -31,3 +31,8 @@ header:
     - '**/*.tcl'
     - '**/*.yml'
     - '**/*.yaml'
+  paths-ignore:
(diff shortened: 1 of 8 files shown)


git-hulk (dismissed):

kezhenxu94 (dismissed):
Thanks for fixing this 🙇 +1

tisonkun:
Push new commits to handle also rand and scripting. Scripting, from my point of view, is a total rewrite and we remain the credit that it's inspired by Redis' implementation.

PragmaTwice (approved):

tisonkun:
Merging...

PR #1264, opened 2023-02-14
Move the kvrocks.conf into volume in Dockerfile

### LINKED ISSUE #1263: Namespace is not persisted after pod restarted.
### Discussed in https://github.com/apache/incubator-kvrocks/discussions/1262

<sup>Originally posted by **Goutami-Gowda** February 14, 2023</sup>
hello, i created kvrocks statefulset with /var/lib/kvrocks volume. but after pod restart , all the namespace that i have added or gone. how can i persist this?</div>

git-hulk:
This closes #1263 

We now store namespaces in kvrocks.conf, but it's NOT in the volume path in Docker and it will be lost after restarting.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,6 @@ RUN apt update && apt install -y libssl-dev
 
 WORKDIR /kvrocks
 
-
 COPY --from=build /kvrocks/build/kvrocks ./bin/
 
 VOLUME /var/lib/kvrocks
--- a/Dockerfile
+++ b/Dockerfile
@@ -42,7 +42,6 @@ VOLUME /var/lib/kvrocks
 COPY ./LICENSE ./NOTICE ./DISCLAIMER ./
 COPY ./licenses ./licenses
 COPY ./kvrocks.conf  /var/lib/kvrocks/conf/
-RUN sed -i -e 's%dir /tmp/kvrocks%dir /var/lib/kvrocks%g' /var/lib/kvrocks/conf/kvrocks.conf
 
 EXPOSE 6666:6666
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -35,14 +35,15 @@ RUN apt update && apt install -y libssl-dev
 
 WORKDIR /kvrocks
 
+
```
git-hulk:
```suggestion
```

tisonkun (commented):
Generally looks good. One comment here:

As we have https://github.com/apache/incubator-kvrocks/pull/1037 now I'm thinking of whether we can replace "RUN  sed ..." part with "ENTRYPOINT" "--dir ...".

git-hulk:
> Generally looks good. One comment here:
> 
> As we have #1037 now I'm thinking of whether we can replace "RUN sed ..." part with "ENTRYPOINT" "--dir ...".

Good advice, will update.

PragmaTwice:
Hi @git-hulk, I remember that, while I run some docker images like caddy, there are several volumes and even a single file volume for configuration, e.g.
```
docker run -p 80:80 -p 443:443 \
    -v somepath/Caddyfile:/etc/caddy/Caddyfile \
    -v somepath:/data -v somepath:/config caddy/caddy
```

So I think maybe the current file structure is enough?

BTW, I think maybe `/kvrocks/data` is more simple that `/var/lib/kvrocks` for data volume.

git-hulk:
> BTW, I think maybe /kvrocks/data is more simple that /var/lib/kvrocks for data volume.

Yes, `/kvrocks/data` is a better path name, but we cannot change the path now since it will affect old users.

PragmaTwice (approved):

PragmaTwice:
Thanks all. Merging...

PR #1271, opened 2023-02-21
Force disable the IPO option in macOS

git-hulk:
Enabling the IPO option will make the debug mode deactivated. The object
the file will become LLVM bitcode and cannot be recognized by the
lldb/gdb.

```
❯ file commands/cmd_string.cc.o
commands/cmd_string.cc.o: LLVM bitcode, wrapper x86_64
```


PragmaTwice (approved):

git-hulk:
Thanks all, merging...

PR #1272, opened 2023-02-21
Use nproc instead of default for job numbers in docker build to avoid timeout

PragmaTwice:
We use `$(nproc)` instead of default value for job numbers in docker build to avoid timeout.

**Reduce docker build time in kvrocks CI from 39min to 20min.**

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -249,11 +249,13 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
+      - name: Get core numbers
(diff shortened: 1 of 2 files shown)


torwig (approved):
Nice improvement!

tisonkun (approved):
LGTM

... while we already unset the `-j` arg by default so it's not 4 :)

https://github.com/apache/incubator-kvrocks/pull/1130

PragmaTwice:
> LGTM
> 
> ... while we already unset the `-j` arg by default so it's not 4 :)
> 
> #1130

Oh got it, fixed.

git-hulk (approved):

PragmaTwice:
Thanks all. Merging...

PR #1279, opened 2023-02-24
Support watch/unwatch commands

### LINKED ISSUE #315: Support Watch/Unwatch
I would like to use Watch and Unwatch for full transaction support. 

yinqiwen/ardb has a working watch/unwatch impl in case that is useful as a reference.

https://github.com/yinqiwen/ardb

PragmaTwice:
It closes #315.

As described in https://github.com/apache/incubator-kvrocks/issues/315#issuecomment-1442063019, we follow the "by command & key tracing" method instead of "by snapshots".

NOTE: key expire / rocksdb storage change monitoring has not been implemented

### DIFF
--- a/src/commands/cmd_txn.cc
+++ b/src/commands/cmd_txn.cc
@@ -45,6 +45,8 @@ class CommandMulti : public Commander {
 class CommandDiscard : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    auto reset_watch = MakeScopeExit([svr, conn] { svr->ResetWatchedKeys(conn); });
+
     if (!conn->IsFlagEnabled(Connection::kMultiExec)) {
       *output = Redis::Error("ERR DISCARD without MULTI");
       return Status::OK();
@@ -96,6 +98,11 @@ class CommandExec : public Commander {
 class CommandWatch : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    if (conn->IsFlagEnabled(Connection::kMultiExec)) {
+      *output = Redis::Error("ERR WATCH inside MULTI is not allowed");
+      return Status::OK();
+    }
+
     svr->WatchKey(conn, std::vector<std::string>(args_.begin() + 1, args_.end()));
     *output = Redis::SimpleString("OK");
     return Status::OK();
--- a/src/commands/cmd_txn.cc
+++ b/src/commands/cmd_txn.cc
@@ -91,7 +91,7 @@ class CommandExec : public Commander {
 class CommandWatch : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    svr->WatchKey(conn, std::vector(args_.begin() + 1, args_.end()));
+    svr->WatchKey(conn, std::vector<std::string>(args_.begin() + 1, args_.end()));
     *output = Redis::SimpleString("OK");
     return Status::OK();
   }
@@ -100,7 +100,7 @@ class CommandWatch : public Commander {
 class CommandUnwatch : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    svr->UnwatchKey(conn, std::vector(args_.begin() + 1, args_.end()));
+    svr->ResetWatchedKeys(conn);
     *output = Redis::SimpleString("OK");
     return Status::OK();
   }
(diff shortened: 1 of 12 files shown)


git-hulk:
This PR is generally good for me, to see if other folks have comments. @torwig @ShooterIT @caipengbo

git-hulk:
There's one corner case that may also modify the key as well:

- Blocking pop from the list: https://github.com/apache/incubator-kvrocks/blob/unstable/src/commands/cmd_list.cc#L228

so we should also update them if the key is modified, others are good to me.

PragmaTwice:
> There's one corner case that may also modify the key as well:
> 
> * Blocking pop from the list: https://github.com/apache/incubator-kvrocks/blob/unstable/src/commands/cmd_list.cc#L228
> 
> so we should also update them if the key is modified, others are good to me.

Done.

git-hulk (approved):

PragmaTwice:
Thanks all. Merging...

Feel free to open an issue if there is anything you are concerning about this PR later.

src/server/server.cc
```diff
@@ -1637,6 +1639,95 @@ Status ServerLogData::Decode(const rocksdb::Slice &blob) {
...
+
+  watched_key_size_ = watched_key_map_.size();
+}
+
+bool Server::IsWatchedKeysModified(Redis::Connection *conn) { return conn->watched_keys_modified_; }
+
+void Server::ResetWatchedKeys(Redis::Connection *conn) {
+  if (watched_key_size_ != 0) {
+    std::unique_lock lock(watched_key_mutex_);
+
```
mapleFU, line 1718: could `watched_key_size_ == 0` here? Because `watched_key_size_` is not checked within lock

mapleFU (commented):
So, here it has:

```
Server:
watched shared lock
map: <key, connections>
watched_key_size_

Connection:
1. watch changed
2. watched key sets
```

All watched data is controlled by shared lock in Server, and only the function modify the Server maping would grab the write-lock. Am I right?

src/server/server.cc
PragmaTwice: `watched_key_size_` is an atomic variable.

src/commands/cmd_txn.cc
```diff
@@ -79,13 +91,37 @@ class CommandExec : public Commander {
...
+
+    svr->WatchKey(conn, std::vector<std::string>(args_.begin() + 1, args_.end()));
+    *output = Redis::SimpleString("OK");
+    return Status::OK();
+  }
+};
+
+class CommandUnwatch : public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
```
mapleFU, line 114: Should unwatch need `kMultiExec` support?

src/server/server.cc
PragmaTwice: It is ok that `watched_key_size_ == 0` here. We check the  `watched_key_size_` just for better performance (to skip the lock when possible).

mapleFU (commented):
Rest LGTM

src/commands/cmd_txn.cc
PragmaTwice: It seems weird to me that `unwatch` actually does not (instantly) fail in multi in Redis, e.g. in redis-cli
```
MULTI
WATCH  ... # error
...
```
```
MULTI
UNWATCH # ok
...
```

src/server/server.cc
```diff
@@ -1637,6 +1639,95 @@ Status ServerLogData::Decode(const rocksdb::Slice &blob) {
...
+void Server::UpdateWatchedKeysFromArgs(const std::vector<std::string> &args, const Redis::CommandAttributes &attr) {
+  if (attr.is_write() && watched_key_size_ > 0) {
+    if (attr.key_range.first_key > 0) {
+      updateWatchedKeysFromRange(args, attr.key_range);
+    } else if (attr.key_range.first_key < 0) {
+      Redis::CommandKeyRange range = attr.key_range_gen(args);
+
+      if (range.first_key > 0) {
+        updateWatchedKeysFromRange(args, range);
+      }
```
mapleFU, line 1677: after `attr.key_range_gen`, if `range.first_key == 0`, would it be a command like flushdb?

src/commands/cmd_txn.cc
mapleFU: It's funny. Seems ok

src/server/server.cc
PragmaTwice: That is a good question. 

I have some concern here: if a command has the write attribute and a static all-zero range, we give the upper-bound approximation as the changed key set. But if a writable command has a dynamic range, I think we cannot make such 
strict assumption, i.e. it can be that, no key is changed. Hence now I do not update all keys in such case.

mapleFU:
Great work!

PR #1285, opened 2023-02-28
Fix type command bug

### LINKED ISSUE #1278: For the expired key, the TYPE command returns dirty data
### Search before asking

### Version

kvrocks v2.3.0

### Minimal reproduce step

```
127.0.0.1:31001> set k v ex 10
OK
127.0.0.1:31001> type k
"string"
127.0.0.1:31001> ttl k
(integer) 3
127.0.0.1:31001> ttl k
(integer) -2
127.0.0.1:31001> type k
"string"
127.0.0.1:31001> get k
(nil)
127.0.0.1:31001> type k
"string"
```

**None** should been returned for the expired key

### What did you see instead?

For the expired key, the TYPE command returns dirty data

### Anything Else?

_No response_

jjz921024:
1. fix #1278 
2. For compatibility with the redis command format, the type command should return simpleString instead of bulkString


PragmaTwice (changes requested):
Hi, I notice that the source branch of your PR is based on tag `2.3.0`, which is wrong.

You need to rebase it into `unstable`, to avoid unnecessary conflicts and changes.

Refer to https://git-scm.com/docs/git-rebase if you are not familiar with Git.

PR #1287, opened 2023-02-28
Use the RocksDB WriteBatchWithIndex to implement the read-your-own-writes in transaction

### LINKED ISSUE #487: [BUG] Transaction can't guarantee atomicity
**Describe the bug**
The exclusive guard lock ensures you run the transaction in isolation, but it cannot guarantee the atomicity of the EXEC stage.
You would like to either have all commands applied to the db or none of them applied.
However, if some commands are executed and persisted to rocksdb and the process fails before the tx completes the db sees an intermediate state which is not desired.

**Expected behavior**
Guarantee the atomicity and consistency.

### LINKED ISSUE #1281: Atomic commit write batches in the command EXEC
### Search before asking

### Motivation

As mentioned in #487, the current transaction is NOT an atomic operation, some operations may be committed to DB but some are not if they crashed in the middle way of executing the command EXEC.

To fix this issue, we need to gather all write operations into one RocksDB's WriteBatch, then commit it at once.

### Solution

We can add a new function to create a shared WriteBatch for the Multi-Exec command like the below:

```C++
Status Storage::Begin() {
    is_txn_mode = true;
    txn_write_batch = make_shared<*rocksdb::WriteBatch>();
}

rocksdb::WriteBatch *Storage::GetWriteBatch() {
   if (is_txn_mode) {
      return txn_write_batch.get();
   }
   return make_shared<*rocksdb::WriteBatch>();
}

Status Storage::Commit() {
    is_txn_mode = false;
    // write txn WriteBatch to RocksDB
    txn_write_batch.reset();
}
```
For the Command Exec, we need to explicitly call the Begin() and Commit() to enter and leave the transaction mode. So that Kvrocks will create a new WriteBatch for each writes operation if it's NOT in transaction mode, and use the shared WriteBatch to collect all write operations if it's in the transaction mode.

git-hulk:
* This closes #1281
* This closes #487 

By default, the writes cannot be seen in transactions if the WriteBatch
wasn't committed. To mitigate this issue, RocksDB offers the WriteBatchWithIndex
to make the uncommitted writes visible if needed.

For more information can see: https://rocksdb.org/blog/2015/02/27/write-batch-with-index.html

This PR looks not easy to review at first glance, but it will be easier if review commits one by one:

- The first commit adds the function `GetWriteBatch` in storage to wrap the WriteBatch for the write operations, making it possible to group and write at once in the EXEC command
- The second commit supports using WriteBatchIndex in transaction mode so that those uncommitted writes can be seen in the same transaction
- The third commit replaces the DB's Get and NewIterator API with WriteBatchIndex to allow visit its own writes in transaction mode.

### DIFF
--- a/src/cluster/slot_import.cc
+++ b/src/cluster/slot_import.cc
@@ -27,9 +27,8 @@ SlotImport::SlotImport(Server *svr)
       import_status_(kImportNone),
       import_fd_(-1) {
   std::lock_guard<std::mutex> guard(mutex_);
-  // Let db_ and metadata_cf_handle_ be nullptr, then get them in real time while use them.
+  // Let metadata_cf_handle_ be nullptr, then get them in real time while use them.
   // See comments in SlotMigrate::SlotMigrate for detailed reason.
-  db_ = nullptr;
   metadata_cf_handle_ = nullptr;
 }
 
--- a/src/commands/cmd_txn.cc
+++ b/src/commands/cmd_txn.cc
@@ -21,6 +21,7 @@
 #include "commander.h"
 #include "error_constants.h"
 #include "server/redis_connection.h"
+#include "server/server.h"
 
 namespace Redis {
 
@@ -68,13 +69,16 @@ class CommandExec : public Commander {
       return Status::OK();
     }
 
+    auto storage = svr->storage_;
     // Reply multi length first
     conn->Reply(Redis::MultiLen(conn->GetMultiExecCommands()->size()));
     // Execute multi-exec commands
     conn->SetInExec();
+    storage->BeginTxn();
     conn->ExecuteCommands(conn->GetMultiExecCommands());
+    auto s = storage->CommitTxn();
     conn->ResetMultiExec();
-    return Status::OK();
+    return s;
   }
 };
 
--- a/src/commands/cmd_txn.cc
+++ b/src/commands/cmd_txn.cc
@@ -74,9 +74,11 @@ class CommandExec : public Commander {
     conn->Reply(Redis::MultiLen(conn->GetMultiExecCommands()->size()));
     // Execute multi-exec commands
     conn->SetInExec();
-    storage->BeginTxn();
-    conn->ExecuteCommands(conn->GetMultiExecCommands());
-    auto s = storage->CommitTxn();
+    auto s = storage->BeginTxn();
+    if (s.IsOK()) {
+      conn->ExecuteCommands(conn->GetMultiExecCommands());
+      s = storage->CommitTxn();
+    }
     conn->ResetMultiExec();
     return s;
   }
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -553,13 +553,13 @@ rocksdb::Status Storage::Write(const rocksdb::WriteOptions &options, rocksdb::Wr
 
 rocksdb::Status Storage::Delete(const rocksdb::WriteOptions &options, rocksdb::ColumnFamilyHandle *cf_handle,
                                 const rocksdb::Slice &key) {
-  auto batch = GetWriteBatch();
+  auto batch = GetWriteBatchBase();
   batch->Delete(cf_handle, key);
   return Write(options, batch->GetWriteBatch());
 }
 
 rocksdb::Status Storage::DeleteRange(const std::string &first_key, const std::string &last_key) {
-  auto batch = GetWriteBatch();
+  auto batch = GetWriteBatchBase();
   rocksdb::ColumnFamilyHandle *cf_handle = GetCFHandle("metadata");
   auto s = batch->DeleteRange(cf_handle, first_key, last_key);
   if (!s.ok()) {
@@ -578,7 +578,7 @@ rocksdb::Status Storage::FlushScripts(const rocksdb::WriteOptions &options, rock
   // didn't contain the end key.
   end_key[end_key.size() - 1] += 1;
 
-  auto batch = GetWriteBatch();
+  auto batch = GetWriteBatchBase();
   auto s = batch->DeleteRange(cf_handle, begin_key, end_key);
   if (!s.ok()) {
     return s;
@@ -685,22 +685,21 @@ void Storage::BeginTxn() {
 Status Storage::CommitTxn() {
   is_txn_mode_ = false;
   auto s = Write(write_opts_, txn_write_batch_->GetWriteBatch());
-  txn_write_batch_.reset();
   if (s.ok()) {
     return {Status::cOK};
   }
   return {Status::NotOK, s.ToString()};
 }
 
-std::shared_ptr<rocksdb::WriteBatchBase> Storage::GetWriteBatch() {
+ObserverOrUniquePtr<rocksdb::WriteBatchBase> Storage::GetWriteBatchBase() {
   if (is_txn_mode_) {
-    return txn_write_batch_;
+    return ObserverOrUniquePtr<rocksdb::WriteBatchBase>(ktxn_write_batch_.get(), ObserverOrUnique::Observer);
   }
-  return std::make_unique<rocksdb::WriteBatch>();
+  return ObserverOrUniquePtr<rocksdb::WriteBatchBase>(new rocksdb::WriteBatch(), ObserverOrUnique::Unique);
 }
 
 Status Storage::WriteToPropagateCF(const std::string &key, const std::string &value) {
-  auto batch = GetWriteBatch();
+  auto batch = GetWriteBatchBase();
   auto cf = GetCFHandle(kPropagateColumnFamilyName);
   batch->Put(cf, key, value);
   auto s = Write(write_opts_, batch->GetWriteBatch());
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -504,6 +504,10 @@ rocksdb::Status Storage::Write(const rocksdb::WriteOptions &options, rocksdb::Wr
   if (reach_db_size_limit_) {
     return rocksdb::Status::SpaceLimit();
   }
+  if (is_txn_mode_) {
+    // The batch won't be flushed until the transaction was committed or rollback
+    return rocksdb::Status::OK();
+  }
 
   // Put replication id logdata at the end of write batch
   if (replid_.length() == kReplIdLength) {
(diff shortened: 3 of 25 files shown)


src/storage/storage.cc
```diff
@@ -639,12 +677,33 @@ void Storage::SetIORateLimit(int64_t max_io_mb) {
...
+  is_txn_mode_ = false;
+  auto s = Write(write_opts_, txn_write_batch_->GetWriteBatch());
+  txn_write_batch_.reset();
+  if (s.ok()) {
+    return {Status::cOK};
+  }
+  return {Status::NotOK, s.ToString()};
+}
+
+std::shared_ptr<rocksdb::WriteBatchBase> Storage::GetWriteBatch() {
```
mapleFU: Should we name it `GetWriteBatchBase` here? And why we use shared_ptr rather than raw pointer here?

src/storage/storage.cc
```diff
@@ -639,12 +677,33 @@ void Storage::SetIORateLimit(int64_t max_io_mb) {
 
 rocksdb::DB *Storage::GetDB() { return db_; }
 
-Status Storage::WriteToPropagateCF(const std::string &key, const std::string &value) {
-  rocksdb::WriteBatch batch;
+void Storage::BeginTxn() {
+  is_txn_mode_ = true;
+  txn_write_batch_ = std::make_unique<rocksdb::WriteBatchWithIndex>();
```
mapleFU, line 692: Does this txn has consistent read snapshot?

src/commands/cmd_txn.cc
```diff
@@ -68,13 +69,16 @@ class CommandExec : public Commander {
...
     }
 
+    auto storage = svr->storage_;
     // Reply multi length first
     conn->Reply(Redis::MultiLen(conn->GetMultiExecCommands()->size()));
     // Execute multi-exec commands
     conn->SetInExec();
+    storage->BeginTxn();
     conn->ExecuteCommands(conn->GetMultiExecCommands());
+    auto s = storage->CommitTxn();
```
mapleFU: Just a naive question. If a exception is thrown in `conn->ExecuteCommands` (I didn't review carefully, so didn't know would it throw). The transaction will never recover until another txn begin and do the thing

mapleFU (commented):
I didn't review it carefully, so maybe I made a mistake, but my point is that this changes too much.

Assume that a transaction is execution, <del>it doesn't block all execution threads, so, the things would be trickey:</del>

1. <del>Storage (which is shared among system) would be mark as txn. My point is that if multiple thread call exec, which would win?</del>
2. <del>When an non-txn get executing when txn is executing, what would it happens? </del>

I suggest that rename it to ctx, and rather than save txn flag to storage, encapsulate it into a per-request context.  <del>And rather than transaction, this maybe better called a batch, because maybe no conflict detection here.</del>

Updated:

Seems that `exclusive` flag would make exec holds and write lock. But still argue that would a exception cause the txn abort.

src/storage/storage.cc
mapleFU: oh it has snapshot, ignore this

src/commands/cmd_txn.cc
git-hulk: Kvrocks won't throw any exception itself, but if the underlying components do, then Kvrocks should exit without committing if we didn't catch it. Thats to say, for the current implementation, all write operations are grouped into one WriteBatch, so it will be committed all or none.

src/storage/storage.cc
git-hulk: Yes, WriteBatchBase is a better name.

>  And why we use shared_ptr rather than raw pointer here

It's ok to use the raw pointer for the global transaction WriteBatch, but we should use the unique_ptr if it's not in transaction mode. So it needs to use shared_ptr if we want to put them together. BTW @PragmaTwice has offered a way to work around this, will have a try this week.

src/commands/cmd_txn.cc
mapleFU: By the way, do we have a flag to check that `ExecuteCommands` triggers error. If it has 5 commands, and the 4th command error. Here it will still try to commit. Am I right?

git-hulk: Yes, that's right. Redis also don't check the execute command status, for example:
```
❯ redis-cli
127.0.0.1:6379> MULTI
OK
127.0.0.1:6379(TX)> SET k1 v1
QUEUED
127.0.0.1:6379(TX)> hset k1 field value
QUEUED
127.0.0.1:6379(TX)> EXEC
1) OK
2) (error) WRONGTYPE Operation against a key holding the wrong kind of value
```

src/storage/storage.cc
```diff
@@ -639,12 +677,33 @@ void Storage::SetIORateLimit(int64_t max_io_mb) {
 
 rocksdb::DB *Storage::GetDB() { return db_; }
 
-Status Storage::WriteToPropagateCF(const std::string &key, const std::string &value) {
-  rocksdb::WriteBatch batch;
+void Storage::BeginTxn() {
+  is_txn_mode_ = true;
```
mapleFU, line 691: Here can we check `is_txn_mode_` is false? If not false, at lease we should add some warn or error log

src/storage/storage.cc
```diff
@@ -639,12 +677,33 @@ void Storage::SetIORateLimit(int64_t max_io_mb) {
...
+  txn_write_batch_ = std::make_unique<rocksdb::WriteBatchWithIndex>();
+}
+
+Status Storage::CommitTxn() {
+  is_txn_mode_ = false;
```
mapleFU, line 702: can we check `is_txn_mode_` is true here? Seems currently we cannot has multiple CommitTxn. At lease we should add some warn or error log

mapleFU (commented):
Other LGTM. Though I still think put `txn_write_batch_` in storage in a bit trickey, can we add some description here, for it can only used by `exclusive` command?

git-hulk:
> Other LGTM. Though I still think put txn_write_batch_ in storage in a bit trickey, can we add some description here, for it can only used by exclusive command?

Thanks for your kind review. Agreed to add more comments to explain why the current implementation is right. I'm also thinking about how to allow multi transactions to improve performance because it's unnecessary to be exclusive command if we can promise the commit all or none. But can retrospect this after this PR and the watch/unwatch feature are ready.

mapleFU:
Just parallel exec can be put into a "Session.Context" rather than `Storage`. A exec can write to it own session, and read first it own session.

Though it's ok, but it needs detail considering. For know, I think txn related is all under exclusive mode, because they only used by "Exec", which is exclusive.

git-hulk:
> Just parallel exec can be put into a "Session.Context" rather than Storage. A exec can write to it own session, and read first it own session.

Yes, the transaction state indeed is a connection state, but it needs some refactoring work to support binding transaction conetext in underlying storage type implementation. We won't do that until this PR and watch/unwatch are supported.

src/storage/storage.cc
PragmaTwice: I think the `Base` suffix seems to be redundant.

> offered a way to work around this

@mwish You can access the prototype via:
https://godbolt.org/z/T1xzfvWMa

torwig:
In general, LGTM.

src/storage/storage.cc
mapleFU: Sounds ok. I just want to distinct `rocksdb::WriteBatchBase` and the real WriteBatch. But current solution is ok.

mapleFU (dismissed):
Other LGTM here, but I still think we need more log and comments.

src/storage/storage.cc
```diff
@@ -500,10 +500,48 @@ Status Storage::GetWALIter(rocksdb::SequenceNumber seq, std::unique_ptr<rocksdb:
...
+                             const rocksdb::Slice &key, std::string *value) {
+  if (is_txn_mode_ && txn_write_batch_->GetWriteBatch()->Count() > 0) {
+    return txn_write_batch_->GetFromBatchAndDB(db_, options, column_family, key, value);
+  }
+  return db_->Get(options, column_family, key, value);
+}
+
+rocksdb::Iterator *Storage::NewIterator(const rocksdb::ReadOptions &options) {
+  return NewIterator(options, db_->DefaultColumnFamily());
+}
```
torwig, line 517: A little formatting here: add a new line after `}`.

src/storage/storage.cc
```diff
@@ -639,12 +677,41 @@ void Storage::SetIORateLimit(int64_t max_io_mb) {
...
+  txn_write_batch_ = std::make_unique<rocksdb::WriteBatchWithIndex>();
+  return Status::OK();
+}
 
+Status Storage::CommitTxn() {
+  if (!is_txn_mode_) {
+    return Status{Status::NotOK, "cannot commit while not in transaction mode"};
+  }
+  is_txn_mode_ = false;
+  auto s = Write(write_opts_, txn_write_batch_->GetWriteBatch());
```
torwig: Could we reset `txn_write_batch` after the transaction commit? Otherwise, it will be reset only on the next call to `BeginTxn`.

src/storage/storage.cc
```diff
@@ -639,12 +677,41 @@ void Storage::SetIORateLimit(int64_t max_io_mb) {
...
+  if (s.ok()) {
+    return {Status::cOK};
```
torwig, line 705: You can just return `Status::OK();` here instead of `{Status::cOK}`.

src/storage/storage.cc
```diff
@@ -639,12 +677,41 @@ void Storage::SetIORateLimit(int64_t max_io_mb) {
...
+  is_txn_mode_ = true;
+  txn_write_batch_ = std::make_unique<rocksdb::WriteBatchWithIndex>();
+  return Status::OK();
+}
 
+Status Storage::CommitTxn() {
+  if (!is_txn_mode_) {
+    return Status{Status::NotOK, "cannot commit while not in transaction mode"};
+  }
+  is_txn_mode_ = false;
```
torwig, line 702: I think `is_txn_mode_` should be set to `false` only **after** the transaction was committed (**after** the call to `Write`). Otherwise `Write` could be called concurrently. Am I right?

git-hulk: The origin Write function will check is_txn_mode status. If true, it won't write the batch to db, so the transaction needs to be reset before committing. But it won't have the data race since the exec command is exclusive.

There's a bit tricky in this implementation, I add a new function `writeToDB` which won't check the transaction status and it should be more clear. @torwig You can take a look at the commit: https://github.com/apache/incubator-kvrocks/pull/1287/commits/b8b5a6bb4332fedd622f8f8f4248e925d122c84b

src/cluster/slot_import.cc
```diff
@@ -27,9 +27,8 @@ SlotImport::SlotImport(Server *svr)
       import_status_(kImportNone),
       import_fd_(-1) {
   std::lock_guard<std::mutex> guard(mutex_);
-  // Let db_ and metadata_cf_handle_ be nullptr, then get them in real time while use them.
+  // Let metadata_cf_handle_ be nullptr, then get them in real time while use them.
   // See comments in SlotMigrate::SlotMigrate for detailed reason.
-  db_ = nullptr;
```
mapleFU, line 32: db_ always ref system db here? Would this cause bug in transaction?

mapleFU (approved):
Seems that incoming changes is that `write` with `txn_batch_mode_` changed to `writeToDb`, and don't rely on `txn_batch_mode_`.

Currently I've no problem, but I hope we can moke error response to test this in the future. Maybe you can use sync_point in rocksdb ( I've port one of it: https://github.com/mapleFU/sync_point , but you can use it in rocksdb) to mock partial exec error in testing.

src/cluster/slot_import.cc
git-hulk: No, it won't. It's right for the migration should always use the raw DB instead of the wrapped one. Because we don't expect to read uncommitted writes when migrating.

torwig (approved):
LGTM

src/storage/storage.h
```diff
@@ -183,7 +198,18 @@ class Storage {
 
   std::atomic<bool> db_in_retryable_io_error_{false};
 
+  std::atomic<bool> is_txn_mode_ = false;
+  // txn_write_batch_ is used as the global write batch for the transaction mode,
+  // all writes will be grouped in this write batch when entering the transaction mode,
+  // then write it at once when committing.
+  //
+  // Notice: the reason why we can use the global transaction? because the EXEC is an exclusive
```
caipengbo, line 206:
```suggestion
  // Notice: the reason why we can use the global transaction is that the EXEC is an exclusive
```

caipengbo (approved):
LGTM, a clear implementation.

git-hulk:
Thanks all, merging…

PR #1296, opened 2023-03-02
Bump rocksdb to 7.10.2

aleksraiden:
RocksDB updated to to 7.10.2. A lot of bugs are fixed, a little bit improved CPU efficiency of DB reads from block cache. 

All test passed OK.

**Bug fixed**:
- Fixed a bug in DB open/recovery from a compressed WAL that was caused due to incorrect handling of certain record fragments with the same offset within a WAL block.
- Fixed a data race on ColumnFamilyData::flush_reason caused by concurrent flushes.
- Fixed DisableManualCompaction() and CompactRangeOptions::canceled to cancel compactions even when they are waiting on conflicting compactions to finish
- Fixed a bug in which a successful GetMergeOperands() could transiently return Status::MergeInProgress()
- Return the correct error (Status::NotSupported()) to MultiGet caller when ReadOptions::async_io flag is true and IO uring is not enabled
- A lot of memory leak are fixed
- Fixes heap usage bugs

**Behavior changes**:
- Make best-efforts recovery verify SST unique ID before Version construction (https://github.com/facebook/rocksdb/pull/10962)
- Introduce epoch_number and sort L0 files by epoch_number instead of largest_seqno. epoch_number represents the order of a file being flushed or ingested/imported. Compaction output file will be assigned with the minimum epoch_number among input files'. For L0, larger epoch_number indicates newer L0 file.

**New Features**:
- When an SstPartitionerFactory is configured, CompactRange() now automatically selects for compaction any files overlapping a partition boundary that is in the compaction range, even if no actual entries are in the requested compaction range. With this feature, manual compaction can be used to (re-)establish SST partition points when SstPartitioner changes, without a full compaction.
- Add BackupEngine feature to exclude files from backup that are known to be backed up elsewhere, using CreateBackupOptions::exclude_files_callback. To restore the DB, the excluded files must be provided in alternative backup directories using RestoreOptions::alternate_dirs.

**Public API Changes**:
- Substantial changes have been made to the Cache class to support internal development goals. Direct use of Cache class members is discouraged and further breaking modifications are expected in the future. SecondaryCache has some related changes and implementations will need to be updated.
- Add MergeOperationOutput::op_failure_scope for merge operator users to control the blast radius of merge operator failures. Existing merge operator users do not need to make any change to preserve the old behavior

### DIFF
--- a/cmake/rocksdb.cmake
+++ b/cmake/rocksdb.cmake
@@ -26,8 +26,8 @@ endif()
 include(cmake/utils.cmake)
 
 FetchContent_DeclareGitHubWithMirror(rocksdb
-  facebook/rocksdb v7.9.2
-  MD5=f9e250f3afdc82288925778cf5513acd
+  facebook/rocksdb v7.10.2
+  MD5=660cddf6ed0857e80860cd86abafc2c3
 )
 
 FetchContent_GetProperties(jemalloc)


torwig (approved):

git-hulk (approved):

xiaobiaozhao (approved):

git-hulk:
@aleksraiden Thanks for your contribution.

PR #1297, opened 2023-03-06
Fix for execution timing of commands from lua

DenizPiri:
Before this, the timing of the commands executed by lua scripts wasn't accounted for correctly. This should resolve that issue.


tisonkun (approved):
Literally looks correct. +1 for merging.

It would be better if you can add a test (e.g. a new go case).

git-hulk (approved):
LGTM

git-hulk:
Thanks all, merging...

git-hulk:
@DenizPiri Thanks for your good catch and contribution again.

PR #1302, opened 2023-03-08
Add CreateThread to avoid manual try-catch

PragmaTwice:
Add `CreateThread`, `ThreadJoin` and `ThreadDetach` to wrap `std::thread` operations, to avoid manual try-catch.

### DIFF
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -243,7 +243,10 @@ class CommandFetchMeta : public Commander {
       auto now = static_cast<time_t>(Util::GetTimeStamp());
       svr->storage_->SetCheckpointAccessTime(now);
     }));
-    t.detach();
+
+    if (auto s = Util::ThreadDetach(t); !s) {
+      return s;
+    }
 
     return Status::OK();
   }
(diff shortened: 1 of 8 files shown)


src/commands/cmd_replication.cc
```diff
@@ -271,48 +273,52 @@ class CommandFetchFile : public Commander {
...
-                    << " ms since of sending files too quickly";
-          usleep(shortest - duration);
-        }
-      }
-      auto now = static_cast<time_t>(Util::GetTimeStamp());
-      svr->storage_->SetCheckpointAccessTime(now);
-      svr->DecrFetchFileThread();
-    });
-    t.detach();
+    std::thread t =
```
torwig: I think you can use `auto` instead of `std::thread` here too.

torwig (dismissed):
LGTM

git-hulk (dismissed):

src/commands/cmd_replication.cc
PragmaTwice: done

torwig (approved):

caipengbo (approved):

PragmaTwice:
Thanks all. Merging...

PR #1303, opened 2023-03-09
Support replica-announce-ip and replica-announce-port

paragor:
Add support
- replica-announce-ip
- replica-announce-port

Generally, it is useful for NAT. 

And for me - it is useful for deploy with redis sentinel in k8s environment (when kvrocks is statefulset), because in rollout restart pod's ip changes. And replica-announce-ip can expose dns name that will be the same between restarts.

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -446,7 +446,7 @@ ReplicationThread::CBState ReplicationThread::replConfReadCB(bufferevent *bev, v
   if (!line) return CBState::AGAIN;
 
   // on unknown option: first try without announce ip, if it fails again - do nothing (to prevent infinite loop)
-  if (line[0] == '-' && isUnknownOption(line.get()) && !self->next_try_without_announce_ip_address_) {
+  if (isUnknownOption(line.get()) && !self->next_try_without_announce_ip_address_) {
     self->next_try_without_announce_ip_address_ = true;
     LOG(WARNING) << "The old version master, can't handle ip-address, "
                  << "try without it again";
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -425,18 +425,34 @@ ReplicationThread::CBState ReplicationThread::checkDBNameReadCB(bufferevent *bev
 
 ReplicationThread::CBState ReplicationThread::replConfWriteCB(bufferevent *bev, void *ctx) {
   auto self = static_cast<ReplicationThread *>(ctx);
-  send_string(bev,
-              Redis::MultiBulkString({"replconf", "listening-port", std::to_string(self->srv_->GetConfig()->port)}));
+  auto config = self->srv_->GetConfig();
+
+  auto port = config->replica_announce_port > 0 ? config->replica_announce_port : config->port;
+  std::vector<std::string> data_to_send{"replconf", "listening-port", std::to_string(port)};
+  if (!self->next_try_without_announce_ip_address_ && !config->replica_announce_ip.empty()) {
+    data_to_send.emplace_back("ip-address");
+    data_to_send.emplace_back(config->replica_announce_ip);
+  }
+  send_string(bev, Redis::MultiBulkString(data_to_send));
   self->repl_state_.store(kReplReplConf, std::memory_order_relaxed);
   LOG(INFO) << "[replication] replconf request was sent, waiting for response";
   return CBState::NEXT;
 }
 
 ReplicationThread::CBState ReplicationThread::replConfReadCB(bufferevent *bev, void *ctx) {
+  auto self = static_cast<ReplicationThread *>(ctx);
   auto input = bufferevent_get_input(bev);
   UniqueEvbufReadln line(input, EVBUFFER_EOL_CRLF_STRICT);
   if (!line) return CBState::AGAIN;
 
+  // on unknown option: first try without announce ip, if it fails again - do nothing (to prevent infinite loop)
+  if (line[0] == '-' && isUnknownOption(line.get()) && !self->next_try_without_announce_ip_address_) {
+    self->next_try_without_announce_ip_address_ = true;
+    LOG(WARNING) << "The old version master, can't handle ip-address, "
+                 << "try without it again";
+    // Retry previous state, i.e. send replconf again
+    return CBState::PREV;
+  }
   if (line[0] == '-' && isRestoringError(line.get())) {
     LOG(WARNING) << "The master was restoring the db, retry later";
     return CBState::RESTART;
@@ -960,6 +976,8 @@ bool ReplicationThread::isRestoringError(const char *err) {
 bool ReplicationThread::isWrongPsyncNum(const char *err) {
   return std::string(err) == "-ERR wrong number of arguments";
 }
+
+bool ReplicationThread::isUnknownOption(const char *err) { return std::string(err) == "-ERR unknown option"; }
 
 rocksdb::Status WriteBatchHandler::PutCF(uint32_t column_family_id, const rocksdb::Slice &key,
                                          const rocksdb::Slice &value) {
--- a/src/server/redis_connection.h
+++ b/src/server/redis_connection.h
@@ -91,7 +91,8 @@ class Connection {
   void SetListeningPort(int port) { listening_port_ = port; }
   int GetListeningPort() { return listening_port_; }
   void SetAnnounceIP(std::string ip) { announce_ip_ = std::move(ip); }
-  std::string GetAnnounceIPOrIP() { return !announce_ip_.empty() ? announce_ip_ : ip_; }
+  std::string GetAnnounceIP() { return !announce_ip_.empty() ? announce_ip_ : ip_; }
+  std::string GetAnnounceAddr() { return GetAnnounceIP() + ":" + std::to_string(listening_port_); }
   uint64_t GetClientType();
   Server *GetServer() { return svr_; }
 
--- a/src/server/redis_connection.h
+++ b/src/server/redis_connection.h
@@ -90,6 +90,8 @@ class Connection {
   uint32_t GetPort() { return port_; }
   void SetListeningPort(int port) { listening_port_ = port; }
   int GetListeningPort() { return listening_port_; }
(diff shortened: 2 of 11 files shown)


git-hulk:
Thanks for your contribution, this's a very useful PR.

@paragor For the CI check error can use `./x.py format` to auto fix the format

src/server/redis_connection.h
```diff
@@ -90,6 +90,9 @@ class Connection {
   uint32_t GetPort() { return port_; }
   void SetListeningPort(int port) { listening_port_ = port; }
   int GetListeningPort() { return listening_port_; }
+  void SetAnnounceIP(std::string ip) { announce_ip_ = std::move(ip); }
+  std::string GetAnnounceIP() { return announce_ip_; }
+  std::string GetAnnounceIPOrIP() { return !announce_ip_.empty() ? announce_ip_ : ip_; }
```
caipengbo: We can do this directly in the `GetIP()` function. If you want a different name, delete `GetIP()` which not used anywhere else.

src/server/redis_connection.h
```diff
@@ -90,6 +90,9 @@ class Connection {
   uint32_t GetPort() { return port_; }
   void SetListeningPort(int port) { listening_port_ = port; }
   int GetListeningPort() { return listening_port_; }
+  void SetAnnounceIP(std::string ip) { announce_ip_ = std::move(ip); }
+  std::string GetAnnounceIP() { return announce_ip_; }
```
caipengbo: There seems to be no place to use `GetAnnounceIP()`, and I personally don't feel the need to add this function

git-hulk (dismissed):
LGTM

src/cluster/replication.cc
```diff
@@ -425,18 +425,35 @@ ReplicationThread::CBState ReplicationThread::checkDBNameReadCB(bufferevent *bev
...
-  send_string(bev,
-              Redis::MultiBulkString({"replconf", "listening-port", std::to_string(self->srv_->GetConfig()->port)}));
+  auto config = self->srv_->GetConfig();
+
+  auto port = config->replica_announce_port > 0 ? config->replica_announce_port : config->port;
+  if (!self->next_try_without_announce_ip_address_ && !config->replica_announce_ip.empty()) {
+    send_string(bev, Redis::MultiBulkString({"replconf", "listening-port", std::to_string(port), "ip-address",
+                                             config->replica_announce_ip}));
+  } else {
+    send_string(bev, Redis::MultiBulkString({"replconf", "listening-port", std::to_string(port)}));
```
caipengbo: Personally, I think use `vector` more concise.

caipengbo (dismissed):
LGTM, except for a few minor comments

src/server/redis_connection.h
```diff
@@ -90,6 +90,9 @@ class Connection {
   uint32_t GetPort() { return port_; }
   void SetListeningPort(int port) { listening_port_ = port; }
   int GetListeningPort() { return listening_port_; }
+  void SetAnnounceIP(std::string ip) { announce_ip_ = std::move(ip); }
```
paragor, line 93: i think i forgot to add opportunity to kill client via  announce addr

src/cluster/replication.cc
git-hulk: I think this place is fine since we are using this initialized vector style in many other places.

src/server/redis_connection.h
git-hulk: GetAnnounceIP is only meaning to slave connections and others should also use GetIP, so we need to keep the GetIP function.

src/cluster/replication.cc
caipengbo: We repeat `'replconf", "listening-port", std::to_string(port)}`, `send_string()`, and we have other options in the future that we can append more concisely.

src/server/redis_connection.h
caipengbo: There is no problem keeping this, individuals prefer when to use and when to add.

git-hulk: Yes, it's using indeed now.

src/server/redis_connection.h
git-hulk: You can help to add this logic if you don't mind.

paragor: yeah, i add it, wait local tests pass and then will push

paragor: done

src/server/redis_connection.h
paragor: done

src/cluster/replication.cc
```diff
@@ -425,18 +425,35 @@ ReplicationThread::CBState ReplicationThread::checkDBNameReadCB(bufferevent *bev
...
 }
 
 ReplicationThread::CBState ReplicationThread::replConfReadCB(bufferevent *bev, void *ctx) {
+  auto self = static_cast<ReplicationThread *>(ctx);
   auto input = bufferevent_get_input(bev);
   UniqueEvbufReadln line(input, EVBUFFER_EOL_CRLF_STRICT);
   if (!line) return CBState::AGAIN;
 
+  // on unknown option: first try without announce ip, if it fails again - do nothing (to prevent infinite loop)
+  if (line[0] == '-' && isUnknownOption(line.get()) && !self->next_try_without_announce_ip_address_) {
```
PragmaTwice: `line[0] == '-'` seems useless here. (`operator==` in string does the same thing)

caipengbo (dismissed):

src/server/redis_connection.h
```diff
@@ -90,6 +90,11 @@ class Connection {
...
+  std::string GetAnnounceIPOrIP() { return !announce_ip_.empty() ? announce_ip_ : ip_; }
```
PragmaTwice: I prefer just `GetAnnounceIP` here.

git-hulk (dismissed):

src/server/redis_connection.h
```diff
@@ -90,6 +90,11 @@ class Connection {
...
+  std::string GetAnnounceAddr() {
+    return GetAnnounceIPOrIP() + ":" + std::to_string(listening_port_ > 0 ? listening_port_ : port_);
```
PragmaTwice: AFAIK `port_` is the peer of `listening_port_`, so I can hardly find the meaning of `listening_port_ > 0 ? listening_port_ : port_`. Could you make some hints to me?

src/server/server.cc
```diff
@@ -910,7 +910,7 @@ void Server::GetReplicationInfo(std::string *info) {
     if (slave->IsStopped()) continue;
 
     string_stream << "slave" << std::to_string(idx) << ":";
-    string_stream << "ip=" << slave->GetConn()->GetIP() << ",port=" << slave->GetConn()->GetListeningPort()
+    string_stream << "ip=" << slave->GetConn()->GetAnnounceIPOrIP() << ",port=" << slave->GetConn()->GetListeningPort()
```
PragmaTwice: Here seems we use `slave->GetConn()->GetListeningPort()` instead of the announce-port, is it right?

PragmaTwice (commented):
Thanks for your contribution! My comments are inline.

BTW, I prefer `announced-ip/port` rather than `announce-ip/port` here.

src/server/server.cc
```diff
@@ -947,7 +947,7 @@ void Server::GetRoleInfo(std::string *info) {
       if (slave->IsStopped()) continue;
 
       list.emplace_back(Redis::MultiBulkString({
-          slave->GetConn()->GetIP(),
+          slave->GetConn()->GetAnnounceIPOrIP(),
           std::to_string(slave->GetConn()->GetListeningPort()),
```
PragmaTwice, line 951: ditto

src/cluster/replication.cc
```diff
@@ -425,18 +425,35 @@ ReplicationThread::CBState ReplicationThread::checkDBNameReadCB(bufferevent *bev
...
-              Redis::MultiBulkString({"replconf", "listening-port", std::to_string(self->srv_->GetConfig()->port)}));
+  auto config = self->srv_->GetConfig();
+
+  auto port = config->replica_announce_port > 0 ? config->replica_announce_port : config->port;
+  if (!self->next_try_without_announce_ip_address_ && !config->replica_announce_ip.empty()) {
+    send_string(bev, Redis::MultiBulkString({"replconf", "listening-port", std::to_string(port), "ip-address",
+                                             config->replica_announce_ip}));
+  } else {
+    send_string(bev, Redis::MultiBulkString({"replconf", "listening-port", std::to_string(port)}));
+  }
```
```suggestion
  std::vector<std::string> data_to_send{"replconf", "listening-port", std::to_string(port)};
  if (!self->next_try_without_announce_ip_address_ && !config->replica_announce_ip.empty()) {
    data_to_send.push_back("ip-address");
    data_to_send.push_back(config->replica_announce_ip);
  }
  send_string(bev, Redis::MultiBulkString());
```

src/server/redis_connection.h
paragor: i was thinking about old versions of kvrocks that can't send listening_port

but now i see git blame and it seems listening_port_ will always > 0. So i just will use to_string(listening_port_)

src/server/server.cc
paragor: GetListeningPort contains announce-port

see function `ReplicationThread::CBState ReplicationThread::replConfWriteCB`

src/cluster/replication.cc
paragor: done

src/server/redis_connection.h
paragor: done

src/server/redis_connection.h
paragor: done

src/cluster/replication.cc
paragor: done

PragmaTwice (approved):
LGTM

git-hulk (approved):

git-hulk:
Thanks all, merging...

git-hulk:
Thanks for your contribution and patient again. @paragor

PR #1310, opened 2023-03-09
Allow HSETNX to receive multiple field-value pairs like HSET

### LINKED ISSUE #1294: Perform HSETNX with multiple values, just like HSET
### Search before asking

### Motivation

With HSET we can add a multiple key/field at once. But HSETNX, that's insert value only if they does not yet exist, use only one key/field pair at once. 

Could we changed HSETNX for usage with multiple values? 

### Solution

_No response_

torwig:
Close #1294

### DIFF
--- a/src/commands/cmd_hash.cc
+++ b/src/commands/cmd_hash.cc
@@ -44,10 +44,22 @@ class CommandHGet : public Commander {
 
 class CommandHSetNX : public Commander {
  public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() % 2 != 0) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+    return Commander::Parse(args);
+  }
+
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    std::vector<FieldValue> field_values;
+    for (size_t i = 2; i < args_.size(); i += 2) {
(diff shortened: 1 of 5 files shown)


src/commands/cmd_hash.cc
```diff
@@ -44,10 +44,22 @@ class CommandHGet : public Commander {
...
+    if (args.size() % 2 != 0) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+    return Commander::Parse(args);
+  }
+
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    std::vector<FieldValue> field_values;
+    for (size_t i = 2; i < args_.size(); i += 2) {
+      field_values.emplace_back(FieldValue{args_[i], args_[i + 1]});
```
PragmaTwice:
```suggestion
      field_values.emplace_back(args_[i], args_[i + 1]);
```

PragmaTwice: BTW I think it is more proper to put the for loop into `Parse`.

torwig: @PragmaTwice Thank you for your reasonable suggestions.

torwig: Done

PragmaTwice (approved):

git-hulk (approved):

PragmaTwice:
Thanks @torwig, merging...

PR #1312, opened 2023-03-10
Improve code style and do some minor fix in scripting

PragmaTwice:
Related to almost all functions in scripting.

### DIFF
--- a/src/storage/scripting.cc
+++ b/src/storage/scripting.cc
@@ -370,7 +370,7 @@ int redisGenericCommand(lua_State *lua, int raise_error) {
     pushError(lua, "READONLY You can't write against a read only slave.");
     return raise_error ? raiseError(lua) : 1;
   }
-  
+
   if (!config->slave_serve_stale_data && srv->IsSlave() && cmd_name != "info" && cmd_name != "slaveof" &&
       srv->GetReplicationState() != kReplConnected) {
     pushError(lua,
@@ -400,7 +400,7 @@ int redisGenericCommand(lua_State *lua, int raise_error) {
     pushError(lua, s.Msg().data());
     return raise_error ? raiseError(lua) : 1;
   }
-  
+
   redisProtocolToLuaType(lua, output.data());
   return 1;
 }
--- a/src/storage/scripting.cc
+++ b/src/storage/scripting.cc
@@ -249,7 +249,7 @@ Status evalGenericCommand(Redis::Connection *conn, const std::vector<std::string
     }
 
     std::string sha;
-    auto s = createFunction(srv, body, &sha, lua);
+    auto s = createFunction(srv, body, &sha, lua, false);
     if (!s.IsOK()) {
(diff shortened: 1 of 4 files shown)


torwig (dismissed):

src/storage/scripting.cc
```diff
@@ -109,12 +110,10 @@ void loadFuncs(lua_State *lua, bool read_only) {
...
-  if (read_only) {
-    /* redis.read_only */
-    lua_pushstring(lua, "read_only");
-    lua_pushboolean(lua, 1);
-    lua_settable(lua, -3);
-  }
+  /* redis.read_only */
+  lua_pushstring(lua, "read_only");
+  lua_pushboolean(lua, read_only);
+  lua_settable(lua, -3);
```
PragmaTwice, line 116: It is not intuitionistic if the boolean value only exists when read_only is true

src/storage/scripting.cc
```diff
@@ -175,26 +174,20 @@ int redisLogCommand(lua_State *lua) {
...
   }
-  level = static_cast<int>(lua_tonumber(lua, -argc));
+  int level = static_cast<int>(lua_tonumber(lua, -argc));
   if (level < LL_DEBUG || level > LL_WARNING) {
     lua_pushstring(lua, "Invalid debug level.");
     return lua_error(lua);
   }
-  if (level < GetServer()->GetConfig()->log_level) {
-    return 0;
-  }
```
PragmaTwice, line 185: This level cannot be compared with loglevel in config.

src/storage/scripting.cc
```diff
@@ -306,48 +294,54 @@ Status evalGenericCommand(Redis::Connection *conn, const std::vector<std::string
...
   auto commands = Redis::GetCommands();
   auto cmd_iter = commands->find(Util::ToLower(args[0]));
   if (cmd_iter == commands->end()) {
     pushError(lua, "Unknown Redis command called from Lua script");
     return raise_error ? raiseError(lua) : 1;
   }
+
   auto redisCmd = cmd_iter->second;
-  if (read_only && redisCmd->is_write()) {
+  if (read_only && ~(redisCmd->flags & Redis::kCmdReadOnly)) {
```
PragmaTwice: There is no assumption that a command is read-only if it has no write flag

src/storage/scripting.cc
```diff
@@ -260,8 +247,9 @@ Status evalGenericCommand(Redis::Connection *conn, const std::vector<std::string
     } else {
       body = args[1];
     }
+
     std::string sha;
-    auto s = createFunction(srv, body, &sha, lua);
+    auto s = createFunction(srv, body, &sha, lua, false);
```
PragmaTwice, line 252: We do not need to store the script body here.

git-hulk (dismissed):

torwig (dismissed):

git-hulk (approved):

PR #1322, opened 2023-03-16
Replace list with other cache-friendly containers

PragmaTwice:


src/server/worker.cc
```diff
@@ -75,7 +75,9 @@ Worker::Worker(Server *svr, Config *config) : svr_(svr), base_(event_base_new())
 }
 
 Worker::~Worker() {
-  std::list<Redis::Connection *> conns;
```
git-hulk, line 78: It will be better to use the list if we need to remove the connection with the pointer?

PragmaTwice: This is a temporary local list, and we only use `emplace_back`. I think there is no reason we remove something from it.

PragmaTwice: And, even if we will do some `remove` from it (although I think it is not possible since we just copy some pointer from other to construct it, and later it is destroyed), we can change its type later.

git-hulk (approved):

torwig (approved):
LGTM

PR #1323, opened 2023-03-16
Introduce TBB to refactor task runners

PragmaTwice:
We refactor the TaskRunner class using blocking queues (tbb::concurrent_bounded_queue) in oneTBB (https://github.com/oneapi-src/oneTBB).

oneTBB provides more utilities about threads which can be used in kvrocks, like spin locks, rw mutex (more efficient), concurrent map, set, vector, and queue, concurrent algorithms, task arena, task scheduler (which has some difference than task runners), and task graphs.


.github/workflows/kvrocks.yaml
```diff
@@ -35,7 +35,7 @@ concurrency:
 jobs:
   check-and-lint:
     name: Lint and check code
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
PragmaTwice, line 38: We need newer clang-tidy (version 12 -> 14) so that a bug (false report) in the old version (clang-tidy 12) will be fixed, and does not fail the CI.

git-hulk (approved):
LGTM

caipengbo:
Do we plan to optimize more places with TBB in the future?

PragmaTwice:
> Do we plan to optimize more places with TBB in the future?

Sure.

PragmaTwice:
Merging...

PR #1325, opened 2023-03-17
Allow to configure the compaction checker policy

### LINKED ISSUE #1291: Enable quick compaction-checker
### Discussed in https://github.com/apache/incubator-kvrocks/discussions/1283

<sup>Originally posted by **xiaobiaozhao** February 27, 2023</sup>
In some case，we need save a lot of data with short ttl, eg ttl = 3600s.
In current kvrocks, it takes at least two days for compation to take place, which can cause disk space not to be released quickly. I think kvrocks needs to come up with a configuration to figure out how long compation will take.

New a conf item `compaction-checker-force-compact-seconds` https://github.com/apache/incubator-kvrocks/blob/c439f3af9d5c7ccb97ff4c7bdab088287acd73cc/src/storage/compaction_checker.cc#L58

And new a conf item `compaction-checker-interval-seconds`
https://github.com/apache/incubator-kvrocks/blob/c439f3af9d5c7ccb97ff4c7bdab088287acd73cc/src/server/server.cc#L184
</div>

ellutionist:
This PR close #1291 .


kvrocks.conf
```diff
@@ -462,6 +462,16 @@ profiling-sample-record-threshold-ms 100
...
 
+# When the compaction checker is triggered, the db will periodically pick the SST file
+# with the highest "deleted percentage" (i.e. the percentage of deleted keys in the SST 
+# file) to compact, in order to free disk space. 
+# However, if a specific SST file was created before "force-compact-file-age" seconds ago, 
+# and its percentage of deleted keys is higher than 
+# "force-compact-file-min-deleted-percentage", it will be forcely compacted as well.
+
+# force-compact-file-age 172800
+# force-compact-file-min-deleted-percentage 10
```
xiaobiaozhao:
```suggestion
force-compact-file-age 172800
force-compact-file-min-deleted-percentage 10
```

Add defalut value description，min & max value.

src/config/config.h
```diff
@@ -131,6 +131,8 @@ struct Config {
   Cron compact_cron;
   Cron bgsave_cron;
   CompactionCheckerRange compaction_checker_range{-1, -1};
+  int64_t force_compact_file_age;
+  int64_t force_compact_file_min_deleted_percentage;
```
xiaobiaozhao:
```suggestion
  uint64_t force_compact_file_age;
  uint64_t force_compact_file_min_deleted_percentage;
```

src/server/server.cc
```diff
@@ -174,7 +174,7 @@ Status Server::Start() {
   compaction_checker_thread_ = GET_OR_RET(Util::CreateThread("compact-check", [this] {
     uint64_t counter = 0;
     time_t last_compact_date = 0;
-    CompactionChecker compaction_checker(this->storage_);
+    CompactionChecker compaction_checker{this->storage_, this->config_};
```
xiaobiaozhao: storage_ already had config‘ pinter

[main.cc](https://github.com/apache/incubator-kvrocks/blob/unstable/src/main.cc#L360)

src/config/config.h
ellutionist: A signed integer type can help avoid negative config value.

src/server/server.cc
ellutionist: But the `config_` is a private field of `Storage` thus cannot be accessed by the `CompactionChecker`.

kvrocks.conf
ellutionist: Done.

src/server/server.cc
git-hulk: Can add a method `GetConfig` for storage. Others are general good to me.

ellutionist: Thanks, `GetConfig` added.

src/config/config.h
PragmaTwice: > A signed integer type can help avoid negative config value.

Sorry but I cannot understand your reply. Could you explain it a bit more?

ellutionist: > > A signed integer type can help avoid negative config value.
> 
> Sorry but I cannot understand your reply. Could you explain it a bit more?

If the config item is defined as a unsigned type (say `uint64_t`), to parse it, eventually `std::strtoul` will be called.

In this situation, if a user set the config value (in the file) to a negative value such as "-1", then the value will be parsed as a ridiculously large number (18446744073709551615).

git-hulk:
@ellutionist can you help to resolve conflicts?

src/config/config.h
PragmaTwice: Thanks, got it. A better way is to make some changes to `parse_util`, but I think the current way can be a temporary workaround.

git-hulk (dismissed):

src/config/config.cc
```diff
@@ -127,6 +127,9 @@ Config::Config() {
       {"replica-announce-ip", false, new StringField(&replica_announce_ip, "")},
       {"replica-announce-port", false, new UInt32Field(&replica_announce_port, 0, 0, PORT_LIMIT)},
       {"compaction-checker-range", false, new StringField(&compaction_checker_range_, "")},
+      {"force-compact-file-age", false, new Int64Field(&force_compact_file_age, 2 * 24 * 3600, 60, INT64_MAX)},
+      {"force-compact-file-min-deleted-percentage", false,
+       new Int64Field(&force_compact_file_min_deleted_percentage, 10, 1, 100)},
```
torwig: Why don't you use `IntField` like the one for `profiling-sample-ratio`? The range of values is just from 1 to 100.

kvrocks.conf
```diff
@@ -462,6 +462,18 @@ profiling-sample-record-threshold-ms 100
 # 0-7am every day.
 compaction-checker-range 0-7
 
+# When the compaction checker is triggered, the db will periodically pick the SST file
+# with the highest "deleted percentage" (i.e. the percentage of deleted keys in the SST 
+# file) to compact, in order to free disk space. 
+# However, if a specific SST file was created before "force-compact-file-age" seconds ago,
```
torwig: I would replace `was created before "force-compact-file-age" seconds ago` with `was created more than "force-compact-file-age" seconds ago`.

src/config/config.cc
ellutionist: Yes, `int` type should suffice.

kvrocks.conf
ellutionist: Thanks, made the change.

src/storage/compaction_checker.cc
```diff
@@ -104,16 +106,23 @@ void CompactionChecker::PickCompactionFiles(const std::string &cf_name) {
...
+
+    // pick the file according to force compact policy
+    if (file_creation_time < static_cast<uint64_t>(now - force_compact_file_age) &&
+        delete_ratio >= force_compact_min_ratio) {
+      LOG(INFO) << "[compaction checker] Going to compact the key in file (force compact policy): " << iter.first;
+      auto s = storage_->Compact(&start_key, &stop_key);
+      LOG(INFO) << "[compaction checker] Compact the key in file (force compact policy): " << iter.first
                 << " finished, result: " << s.ToString();
       maxFilesToCompact--;
+      continue;
```
ellutionist, line 119: @git-hulk @PragmaTwice @xiaobiaozhao I kindly ask you to pay attention to this new `continue` sentence because this will actually change how the `compaction_checker` works.

I add this because if a SST file has been compacted due to the "force compact policy", it should not be considered as a possible candidate with the `best_delete_ratio`. If it is not so skipped, the file may be compacted twice in a single call of `PickCompactionFiles`, from my point of view.

If any of you think this change is not proper please let me know.

git-hulk: @ellutionist Thanks for your point out. You're right that is unnecessary to check again.

git-hulk (approved):

torwig (approved):
LGTM

git-hulk:
Thanks all, merging...

git-hulk:
Thanks for @ellutionist contribution again.

PR #1335, opened 2023-03-20
Relax race condition between subkey compaction thread

### LINKED ISSUE #1298: race condition between subkey compaction thread and Expire command
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Version

all version had the same problem.

### step 1:slowdown  compacton thread@ src/storage/compact_filter.cc
```c++
bool SubKeyFilter::IsMetadataExpired(const InternalKey &ikey, const Metadata &metadata) const {
  if (metadata.Type() == kRedisString  // metadata key was overwrite by set command
      || metadata.Expired() || ikey.GetVersion() != metadata.version) {
      std::cout <<"exp:"<< metadata.expire << ",mk:"<< ikey.GetKey().ToString() << "=>" << ikey.GetSubKey().ToString() <<std::endl;
      cached_key_.clear(); // meta TTL will be update in another thread,so don't cache it
      usleep(100000); // slow down
    return true;
  }
  return false;
}
```

### step 2: slow down for update TTL after check expired@ src/storage/redis_db.cc
```c++
rocksdb::Status Database::Expire(const Slice &user_key, int timestamp) {
....
  if (metadata.Expired()) {
    return rocksdb::Status::NotFound(kErrMsgKeyExpired);
  }
...
sleep(2);
....
}
```

### step 3: a unit test @ tests/cppunit/t_metadata_test.cc
```c++
#include <random>
#include <thread>
using namespace std;
std::string GenRandomStr(size_t len){
  static constexpr const char* DEFAULT_CHARSET = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";

  int64_t now = rocksdb::Env::Default()->NowNanos();
  static std::default_random_engine random{static_cast<unsigned>(now)};
  static std::mt19937 random_generator(random());
  std::string rand_str_out  = std::string(DEFAULT_CHARSET);
  while(len > rand_str_out.length()) rand_str_out +=rand_str_out ;
  std::shuffle(rand_str_out.begin(), rand_str_out.end(), random_generator);
  return rand_str_out.substr(0,len);
}
TEST_F(RedisTypeTest, ExpireBug) {


  bool start_compact = false;
  bool compact_started = false;
  bool stop_running = false;

  std::thread compact_thread([&](){
    while(!stop_running){
      if(start_compact){
        rocksdb::CompactRangeOptions compact_opts;
        compact_opts.change_level = true;
        compact_started = true;
        auto s = storage_->GetDB()->CompactRange(compact_opts,nullptr,nullptr);// compaction subkey,default cf only
        ASSERT_TRUE(s.ok()) << s.ToString();
      }else{
        // waiting to start compact
        usleep(10000);
      }
    }
  });


  int uk_count = 50;  

  auto cfs = storage_->GetCFHandles();
  std::vector<FieldValue> mk_field_values;    
  std::string mk = GenRandomStr(10);
  set<string> field_no_dup ;

  for(int j=0;j<uk_count;j++){
    string field = GenRandomStr(200);
    field_no_dup.insert(field);
    mk_field_values.push_back(FieldValue{field,"0000"});
  }
  EXPECT_EQ(field_no_dup.size(),uk_count);

  int ret = 0;
  rocksdb::Status s = hash->MSet(mk, mk_field_values, false, &ret);
  EXPECT_TRUE(s.ok() && uk_count == ret);
  int64_t now;
  rocksdb::Env::Default()->GetCurrentTime(&now);
  int ttl_exp = now + 5;
  hash->Expire(mk,ttl_exp);      
  

  for(auto&item:*cfs){
    rocksdb::FlushOptions fs;
    auto s = storage_->GetDB()->Flush(fs,item);
    cout << "flush done for:"<< item->GetName() <<",res:" << s.ToString()<<endl;
  }
  sleep(2);
  start_compact = true;    
  cout << "start compact " <<endl;

  while(!compact_started ){
    usleep(10);
  }
  
  //part field will be expired..
  bool lease_TTL_successed = false;
  rocksdb::Env::Default()->GetCurrentTime(&now);
  s = hash->Expire(mk,(int)(now+100));
  if(s.ok()){
    lease_TTL_successed = true;        
  }
    
  if(!lease_TTL_successed){
    start_compact = false;
    cout << "try another round"<<endl;      
    return;      
  }    
  sleep(2);
  vector<FieldValue> vret;
  s = hash->GetAll(mk,&vret);
  if(s.ok()){
    EXPECT_EQ(vret.size(),uk_count);
  }else{
    cout << "mk:"<< mk << " is gone"<<endl;
  }
  stop_running = true;
  compact_thread.join();
}
```
### step 4 compile and run:
`make unittest && rm -rf testsdb/&& ./unittest  --gtest_filter=RedisTypeTest.ExpireBug`

no race condition, the key either expired or kept

### What did you see instead?

a race condition between the Expire command and the compaction thread leads to the deletion of certain fields/members from a hash/zset/set/list。
image

Specifically,this issue occurs when a key's TTL  was expiring while the Expire command's Exprire checking passed and the same time,the `SubKeyFilter(SubKeyFilter::IsMetadataExpired)` removes some subkey.
image

now the key's TTL is updated success but some field/membert were disappeared. 

### Anything Else?

a lazy delete of the subkey may by plus 1 min when a compaction thread checking meta's Expred for subkey may help.

dongdongwcpp:
lazy delete to best effort relax race condition between command `Expire` and subkey Compaction.

It closes #1298.


tisonkun (changes requested):
Could you elaborate a bit how you fix this race condition logically? Generally, "wait for a while" is not logically correct but best effort.

src/storage/compact_filter.cc
```diff
@@ -86,8 +87,11 @@ Status SubKeyFilter::GetMetadata(const InternalKey &ikey, Metadata *metadata) co
 }
 
 bool SubKeyFilter::IsMetadataExpired(const InternalKey &ikey, const Metadata &metadata) const {
+  // lazy delete to avoid race condition between command Expire and subkey Compaction
+  // Related issue:https://github.com/apache/incubator-kvrocks/issues/1298
+  int64_t lazy_expired_ts = Util::GetTimeStamp() + 1000;
```
git-hulk:
```suggestion
  //  `Util::GetTimeStamp() - 1000` means extending 1000s for expired items,
  // or they will be recycled once they reach the expire time.
  int64_t lazy_expired_ts = Util::GetTimeStamp() - 1000;
```

src/storage/compact_filter.cc
```diff
@@ -86,8 +87,13 @@ Status SubKeyFilter::GetMetadata(const InternalKey &ikey, Metadata *metadata) co
 }
 
 bool SubKeyFilter::IsMetadataExpired(const InternalKey &ikey, const Metadata &metadata) const {
+  // lazy delete to avoid race condition between command Expire and subkey Compaction
+  // Related issue:https://github.com/apache/incubator-kvrocks/issues/1298
+  //  `Util::GetTimeStamp() - 1000` means extending 1000s for expired items,
+  // or they will be recycled once they reach the expire time.
+  int64_t lazy_expired_ts = Util::GetTimeStamp() - 1000;
```
git-hulk:
```suggestion
  // lazy delete to avoid race condition between command Expire and subkey Compaction
  // Related issue:https://github.com/apache/incubator-kvrocks/issues/1298
  //
  // `Util::GetTimeStamp() - 300` means extending 5 minutes for expired items,
  // to prevent them from being recycled once they reach the expiration time.
  int64_t lazy_expired_ts = Util::GetTimeStamp() - 300;
```

PragmaTwice:
> Could you elaborate a bit how you fix this race condition logically? Generally, "wait for a while" is not logically correct but best effort.

Hi @tisonkun , this race is triggered by modifying the expire time of a key while compacting. Rare in real-world but indeed a race. 

In my view, since it is hard and expensive to add locking to both compacting and command executing, waiting is the currently best way to avoid the race.

tisonkun:
> > Could you elaborate a bit how you fix this race condition logically? Generally, "wait for a while" is not logically correct but best effort.
> 
> Hi @tisonkun , this race is triggered by modifying the expire time of a key while compacting. Rare in real-world but indeed a race.
> 
> In my view, since it is hard and expensive to add locking to both compacting and command executing, waiting is the currently best way to avoid the race.

Thank you. Then let's update the PR title as "Relax ..." instead of "Fix ...".

tisonkun (approved):

PragmaTwice (approved):

git-hulk (approved):

git-hulk:
Thanks all, merging...

git-hulk:
Thanks for your contribution again @dongdongwcpp

PR #1339, opened 2023-03-20
Use waitForImportState instead of the sleep in Go test case

git-hulk:
Fix flaky test case failure: https://github.com/apache/incubator-kvrocks/actions/runs/4467548694/jobs/7847311171?pr=1328

I can't reproduce this issue, but it may be caused by the migrate may not starting before checking the status. So it will be still in success status even though the cluster nodes were changed.


PragmaTwice (approved):

torwig (approved):

git-hulk:
Thanks, merging...

PR #1341, opened 2023-03-21
Make Status::Code::cOK private to avoid direct access

PragmaTwice:


tisonkun (approved):

git-hulk (approved):

PragmaTwice:
Thanks all. Merging...

ShooterIT:
a taste problem, why use `cOK` instead of `kOK`, for const variables, we use `k` prefix but this one 🙃

PragmaTwice:
> a taste problem, why use `cOK` instead of `kOK`, for const variables, we use `k` prefix but this one 🙃

Maybe you need to ask the author of Status. I just use the original name.

git-hulk:
😭 This should be a typo from me, we copied the status from rocksdb and forgot to remove the `c` prefix.

PR #1342, opened 2023-03-21
Implement the new encoding with 64bit size and expire time in milliseconds

### LINKED ISSUE #1033: Proposal: New encoding with compatibility to old data
### Search before asking

### Motivation

[The current encoding](https://kvrocks.apache.org/docs/Design/design-structure-on-rocksdb) of kvrocks has several notable problems:
1. The length of `size` field is 32 bits, which makes it impossible to store large collection exceeding `4Gi` (items);
2. The length of `expire` field is 32 bits, which means that kvrocks has the Y2038 problem, i.e. kvrocks will lose availability in 2038.

### Solution

Since currently the higher 4 bits of `flags` field are still unused, we will use one of them to distinguish between the old encoding and the new one, i.e. 
```
 0 0 0 0 <4-bit redis-type> -> old encoding
 1 0 0 0 <4-bit redis-type> -> new encoding
```

The new protocol uses fixed 64bit and varint encoding to represent `expire` and `size` field, and all other fields remain the same as before. In this way, the new protocol can be extended to 64bit (or even higher) to solve the above two problems without significantly increasing the space ~~(actually shrinking, because for small number, 4 bytes are not needed to storage).~~

In fact, I discussed this solution with @torwig in slack before, but only considered the `size` field and used fixed-length 64bit encoding. But now I think varint encoding might be a better fit. ~I do not know if the `version` field is suitable for varint encoding, so some advice might be needed.~

For example, the new hash metadata:
```
        +------------------+------------+-----------+-----------+
key =>  |  flags           |  expire    |  version  |  size     |
        |(1000 + 4bit type)| (64bit ms) |  (64bit)  | (64bit)   |
        +------------------+------------+-----------+-----------+
```

The old hash metadata just for compare:
```
        +------------------+------------+-----------+-----------+
key =>  |  flags           |  expire    |  version  |  size     |
        |(0000 + 4bit type)| (32bit sec)|  (64bit)  | (32bit)   |
        +------------------+------------+-----------+-----------+
```

Note that we store millisecond timestamp in the new `expire` field.

In implementation, we can read data with both old and new encoding, and **always write data by the new encoding**.

cc @git-hulk @torwig @ShooterIT @caipengbo 

PragmaTwice:
It closes #1033.

See proposal in #1033, after this change, these old data can still be readable/writable, but all new data will be written via the new encoding.

NOTE: although the encoding now support 64bit size, maybe some place in code still use `int32_t`/`int` so it cannot work well in large number of items more than 32bit. We can fix them after this PR.

After some discussion, we create a new build option ENABLE_NEW_ENCODING (currently default OFF), and users need to turn this option on to use this feature.

### DIFF
--- a/src/storage/compact_filter.cc
+++ b/src/storage/compact_filter.cc
@@ -92,7 +92,7 @@ bool SubKeyFilter::IsMetadataExpired(const InternalKey &ikey, const Metadata &me
   //
   // `Util::GetTimeStamp() - 300` means extending 5 minutes for expired items,
   // to prevent them from being recycled once they reach the expiration time.
-  int64_t lazy_expired_ts = Util::GetTimeStamp() - 300;
+  uint64_t lazy_expired_ts = Util::GetTimeStampMS() - 300000;
   if (metadata.Type() == kRedisString  // metadata key was overwrite by set command
       || metadata.ExpireAt(lazy_expired_ts) || ikey.GetVersion() != metadata.version) {
     return true;
--- a/src/storage/compact_filter.cc
+++ b/src/storage/compact_filter.cc
@@ -90,7 +90,7 @@ bool SubKeyFilter::IsMetadataExpired(const InternalKey &ikey, const Metadata &me
   // lazy delete to avoid race condition between command Expire and subkey Compaction
   // Related issue:https://github.com/apache/incubator-kvrocks/issues/1298
   //
-  // `Util::GetTimeStamp() - 300` means extending 5 minutes for expired items,
+  // `Util::GetTimeStampMS() - 300000` means extending 5 minutes for expired items,
   // to prevent them from being recycled once they reach the expiration time.
   uint64_t lazy_expired_ts = Util::GetTimeStampMS() - 300000;
   if (metadata.Type() == kRedisString  // metadata key was overwrite by set command
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -205,7 +205,7 @@ void Database::Keys(const std::string &prefix, std::vector<std::string> *keys, K
         if (ttl != -1) {
           stats->n_key++;
           stats->n_expires++;
-          if (ttl > 0) ttl_sum += ttl / 1000;
+          if (ttl > 0) ttl_sum += ttl;
         }
       }
       if (keys) {
@@ -224,7 +224,7 @@ void Database::Keys(const std::string &prefix, std::vector<std::string> *keys, K
   }
 
   if (stats && stats->n_expires > 0) {
-    stats->avg_ttl = ttl_sum / stats->n_expires;
+    stats->avg_ttl = ttl_sum / stats->n_expires / 1000;
   }
 }
 
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -90,16 +90,18 @@ rocksdb::Status Database::Expire(const Slice &user_key, uint64_t timestamp) {
   if (metadata.Type() != kRedisString && metadata.size == 0) {
     return rocksdb::Status::NotFound("no elements");
   }
-  if (metadata.expire / 1000 == timestamp) return rocksdb::Status::OK();
+  if (metadata.expire == timestamp) return rocksdb::Status::OK();
 
-  auto buf = std::make_unique<char[]>(value.size());
-  memcpy(buf.get(), value.data(), value.size());
   // +1 to skip the flags
-  EncodeFixed32(buf.get() + 1, (uint32_t)timestamp);
+  if (metadata.Is64BitEncoded()) {
+    EncodeFixed64(value.data() + 1, timestamp);
+  } else {
+    EncodeFixed32(value.data() + 1, timestamp / 1000);
+  }
   auto batch = storage_->GetWriteBatchBase();
   WriteBatchLogData log_data(kRedisNone, {std::to_string(kRedisCmdExpire)});
   batch->PutLogData(log_data.Encode());
-  batch->Put(metadata_cf_handle_, ns_key, Slice(buf.get(), value.size()));
+  batch->Put(metadata_cf_handle_, ns_key, value);
   s = storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
   return s;
 }
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -157,10 +157,8 @@ rocksdb::Status Database::TTL(const Slice &user_key, int64_t *ttl) {
 
   Metadata metadata(kRedisNone, false);
   metadata.Decode(value);
-  *ttl = metadata.TTL();
-  if (*ttl > 0) {
-    *ttl /= 1000;
-  }
+  auto res = metadata.TTL();
+  *ttl = res > 0 ? res / 1000 : res;
 
   return rocksdb::Status::OK();
 }
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -30,6 +30,7 @@
 #include "rocksdb/iterator.h"
 #include "server/server.h"
 #include "storage/redis_metadata.h"
+#include "time_util.h"
 
 namespace Redis {
 
@@ -156,9 +157,9 @@ rocksdb::Status Database::TTL(const Slice &user_key, int64_t *ttl) {
 
   Metadata metadata(kRedisNone, false);
   metadata.Decode(value);
-  *ttl = metadata.TTL() / 1000;
-  if (*ttl < 0) {
-    *ttl = -2;
+  *ttl = metadata.TTL();
+  if (*ttl > 0) {
+    *ttl /= 1000;
   }
 
   return rocksdb::Status::OK();
@@ -203,9 +204,11 @@ void Database::Keys(const std::string &prefix, std::vector<std::string> *keys, K
       }
       if (stats) {
         int64_t ttl = metadata.TTL();
-        stats->n_key++;
-        stats->n_expires++;
-        if (ttl > 0) ttl_sum += ttl / 1000;
+        if (ttl != -1) {
+          stats->n_key++;
+          stats->n_expires++;
+          if (ttl > 0) ttl_sum += ttl / 1000;
+        }
       }
       if (keys) {
         ExtractNamespaceKey(iter->key(), &ns, &user_key, storage_->IsSlotIdEncoded());
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -404,8 +404,12 @@ rocksdb::Status Database::Dump(const Slice &user_key, std::vector<std::string> *
   infos->emplace_back(std::to_string(metadata.version));
   infos->emplace_back("expire");
   infos->emplace_back(std::to_string(Metadata::ExpireMsToS(metadata.expire)));
+  infos->emplace_back("pexpire");
+  infos->emplace_back(std::to_string(metadata.expire));
   infos->emplace_back("size");
   infos->emplace_back(std::to_string(metadata.size));
+  infos->emplace_back("is_64bit_common_field");
+  infos->emplace_back(std::to_string(metadata.Is64BitEncoded()));
 
   infos->emplace_back("created_at");
   struct timeval created_at = metadata.Time();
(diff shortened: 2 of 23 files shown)


torwig:
@PragmaTwice Wow! Awesome feature, especially 64-bit size.

aleksraiden:
Awesome, I very need this feature!!

src/storage/compact_filter.cc
```diff
@@ -92,7 +92,7 @@ bool SubKeyFilter::IsMetadataExpired(const InternalKey &ikey, const Metadata &me
   //
   // `Util::GetTimeStamp() - 300` means extending 5 minutes for expired items,
```
git-hulk:
```suggestion
  // `Util::GetTimeStampMS() - 300000` means extending 5 minutes for expired items,
```

PragmaTwice:
It is ready for review now.

src/storage/redis_metadata.h
```diff
@@ -97,22 +97,45 @@ class InternalKey {
...
+  // the current version: 53bit timestamp + 11bit counter
   uint64_t version;
-  uint32_t size;
 
- public:
-  explicit Metadata(RedisType type, bool generate_version = true);
+  // element size of the key-value
+  uint64_t size;
+
+  explicit Metadata(RedisType type, bool generate_version = true, bool use_64bit_common_field = true);
```
torwig: Is there a case when we need to pass explicitly the `use_64bit_common_field` parameter?

PragmaTwice: No, we do not need to change this parameter. The old data will be decoded normally, and the new data will be encoded use the new encoding (64bit).

PragmaTwice: But I am thinking if we set the parameter to `false` to allow users rollback to previous kvrocks versions (i.e. users cannot rollback (e.g. to 2.3.0) when they use the new encoding to write some new data to the database, although these old data will be treat as well). Do you have some idea?

torwig: Yes, I also had the same idea. 
So, maybe we can set 64-bit encoding in the constructor by default, and it will be overwritten by the Decode method in case of decoding from the 'old' format?

PragmaTwice: > maybe we can set 64-bit encoding in the constructor by default, and it will be overwritten by the Decode method in case of decoding from the 'old' format?

Yes, you are right.

torwig: I just thought that the `use_64bit_common_field` can be useful in unit tests to test backward compatibility, for example.

PragmaTwice: > I just thought that the use_64bit_common_field can be useful in unit tests to test backward compatibility, for example.

Good idea.

xiaobiaozhao:
Wow! Awesome feature

git-hulk (dismissed):

git-hulk (dismissed):

torwig:
@PragmaTwice Hi, I'm trying to add a few unit tests, and the following one is failing:

```
TEST(Metadata, MetadataDecodingBackwardCompatibleSimpleKey) {
  auto expire_at = Util::GetTimeStamp() + 10;
  Metadata md_old(kRedisString, true, false);
  EXPECT_FALSE(md_old.Is64BitEncoded());
  md_old.expire = expire_at;
  std::string encoded_bytes;
  md_old.Encode(&encoded_bytes);

  Metadata md_new(kRedisNone, false, true); // decoding existing metadata with 64-bit feature activated
  md_new.Decode(encoded_bytes);
  EXPECT_FALSE(md_new.Is64BitEncoded());
  EXPECT_EQ(md_new.Type(), kRedisString);
  EXPECT_EQ(md_new.expire, expire_at*1000);
}
```

With the following error:

```
Expected equality of these values:
  md_new.expire
    Which is: 1679553000
  expire_at*1000
    Which is: 1679553143000

```

I'm just wondering is there an error in logic or if my test is somehow incorrect (typo/misunderstanding/etc.) :)
Could you please have a look at it?

PragmaTwice:
Hi @torwig, thanks for your testing. 

In this PR, the `metadata.expire` is now always in milliseconds (whatever it is in 64bit mode or not). And only when we encode it into some raw data in non-64bit mode, it will be converted to seconds.

This design is aimed to make the code simple. If the `metadata.expire` can be either in milliseconds or seconds, it can be a mess.

torwig:
@PragmaTwice I see. 
But if I use a new version of Kvrocks and I have existing data and their metadata contain expiration time in seconds. Then I use Decode on them and I expect that the new expiration time won't change, just seconds will be transformed into milliseconds. Am I right?

PragmaTwice:
> @PragmaTwice I see. But if I use a new version of Kvrocks and I have existing data and their metadata contain expiration time in seconds. Then I use Decode on them and I expect that the new expiration time won't change, just seconds will be transformed into milliseconds. Am I right?

Sure. The logic is in here:

https://github.com/apache/incubator-kvrocks/blob/11f2a80c1fa1fd17e8fa4a2c600a8e34eea1e3e0/src/storage/redis_metadata.cc#L264

torwig:
@PragmaTwice So there is a mistake in my test?

PragmaTwice:
> @PragmaTwice So there is a mistake in my test?

The correct version is like this:
```c++
TEST(Metadata, MetadataDecodingBackwardCompatibleSimpleKey) {
  auto expire_at = (Util::GetTimeStamp() + 10) * 1000;
  Metadata md_old(kRedisString, true, false);
  EXPECT_FALSE(md_old.Is64BitEncoded());
  md_old.expire = expire_at;
  std::string encoded_bytes;
  md_old.Encode(&encoded_bytes);

  Metadata md_new(kRedisNone, false, true); // decoding existing metadata with 64-bit feature activated
  md_new.Decode(encoded_bytes);
  EXPECT_FALSE(md_new.Is64BitEncoded());
  EXPECT_EQ(md_new.Type(), kRedisString);
  EXPECT_EQ(md_new.expire, expire_at);
}
```

You need to use milliseconds for `metadata.expire`.

torwig:
@PragmaTwice I intentionally use seconds and explicit `false` as the last parameter of a constructor to model the current encoding of metadata. And then I'm decoding that metadata with the new version by passing explicit `true` as the last parameter to the constructor.

PragmaTwice:
> I intentionally use seconds and explicit false as the last parameter of a constructor to model the current encoding of metadata. And then I'm decoding that metadata with the new version by passing explicit true as the last parameter to the constructor.

`metadata.expire` now cannot be seconds. It must be milliseconds, and when it is encoded to some raw data, it will be converted to seconds for `use_64bit_common_field=false`.

Your test is added to the codebase, thanks! https://github.com/apache/incubator-kvrocks/pull/1342/commits/558bece6d972a1468405790d38a512a71f5e27e9

torwig:
@PragmaTwice For testing new features I have the following tests, you can add them:

```
TEST(Metadata, Metadata64bitExpiration) {
  auto expire_at = Util::GetTimeStampMS() + 1000;
  Metadata md_src(kRedisString, true, true);
  EXPECT_TRUE(md_src.Is64BitEncoded());
  md_src.expire = expire_at;
  std::string encoded_bytes;
  md_src.Encode(&encoded_bytes);

  Metadata md_decoded(kRedisNone, false, true);
  md_decoded.Decode(encoded_bytes);
  EXPECT_TRUE(md_decoded.Is64BitEncoded());
  EXPECT_EQ(md_decoded.Type(), kRedisString);
  EXPECT_EQ(md_decoded.expire, expire_at);
}

TEST(Metadata, Metadata64bitSize) {
  uint64_t big_size = 100000000000;
  Metadata md_src(kRedisHash, true, true);
  EXPECT_TRUE(md_src.Is64BitEncoded());
  md_src.size = big_size;
  std::string encoded_bytes;
  md_src.Encode(&encoded_bytes);

  Metadata md_decoded(kRedisNone, false, true);
  md_decoded.Decode(encoded_bytes);
  EXPECT_TRUE(md_decoded.Is64BitEncoded());
  EXPECT_EQ(md_decoded.Type(), kRedisHash);
  EXPECT_EQ(md_decoded.size, big_size);
}
```

To test backward compatibility, I have the following two ones, but they are failing:

```
TEST(Metadata, MetadataDecodingBackwardCompatibleSimpleKey) {
  auto expire_at = Util::GetTimeStamp() + 10; // <-- assign seconds here
  Metadata md_old(kRedisString, true, false);
  EXPECT_FALSE(md_old.Is64BitEncoded());
  md_old.expire = expire_at;
  std::string encoded_bytes;
  md_old.Encode(&encoded_bytes);

  Metadata md_new(kRedisNone, false, true);
  md_new.Decode(encoded_bytes);
  EXPECT_FALSE(md_new.Is64BitEncoded());
  EXPECT_EQ(md_new.Type(), kRedisString);
  EXPECT_EQ(md_new.expire, expire_at*1000);
}

TEST(Metadata, MetadataDecodingBackwardCompatibleComplexKey) {
  auto expire_at = Util::GetTimeStamp() + 100; // <-- assign seconds here
  uint32_t size = 1000000000;
  Metadata md_old(kRedisHash, true, false);
  EXPECT_FALSE(md_old.Is64BitEncoded());
  md_old.expire = expire_at;
  md_old.size = size;
  std::string encoded_bytes;
  md_old.Encode(&encoded_bytes);

  Metadata md_new(kRedisHash, false, true);
  md_new.Decode(encoded_bytes);
  EXPECT_FALSE(md_new.Is64BitEncoded());
  EXPECT_EQ(md_new.Type(), kRedisHash);
  EXPECT_EQ(md_new.expire, expire_at*1000);
  EXPECT_EQ(md_new.size, size);
}
```

But fixing them by assigning milliseconds instead of seconds won't work because in this case, they would not test backward compatibility anymore. So these two tests can be discarded.

PragmaTwice:
@torwig Thanks! Added.

torwig (approved):
LGTM.
@PragmaTwice Now I realized what was wrong with my failing tests: I set the expiration timestamp in seconds while 64-bit encoding was false, trying to emulate "old" behavior, and PutExpire made a division by 1000, and the result confused me :)

PragmaTwice:
Thanks @torwig and @git-hulk. Merging...

mapleFU:
Can some stale versions of kvrocks recognize this format? If user ENABLE it, and find other bug, and revert kvrocks to a old version. Would user get segment fault on these keys?

PragmaTwice:
> Can some stale versions of kvrocks recognize this format? If user ENABLE it, and find other bug, and revert kvrocks to a old version. Would user get segment fault on these keys?

Firstly, we currently do not enable this new encoding by default.

And if users write some data via the new encoding, they cannot revert their kvrocks version to 2.3.0 or earlier. But if we release 2.4.0 and keep `ENABLE_NEW_ENCODING=OFF`, and then release 2.5.0 to turn on it as default, the users can revert their version from 2.5.0 to 2.4.0 without any data breaking.

BTW, I think any encoding changes may be a pain to users, which is not specially related to this PR.

src/storage/redis_metadata.cc
```diff
@@ -145,33 +147,39 @@ void ComposeSlotKeyPrefix(const Slice &ns, int slotid, std::string *output) {
   PutFixed16(output, static_cast<uint16_t>(slotid));
 }
 
-Metadata::Metadata(RedisType type, bool generate_version)
-    : flags((uint8_t)0x0f & type), expire(0), version(0), size(0) {
-  if (generate_version) version = generateVersion();
-}
+Metadata::Metadata(RedisType type, bool generate_version, bool use_64bit_common_field)
+    : flags((use_64bit_common_field ? METADATA_64BIT_ENCODING_MASK : 0) | (METADATA_TYPE_MASK & type)),
```
mapleFU, line 151: Hi twice, when decoding, should we force get `use_64bit` from field?

PragmaTwice: sure.

src/storage/redis_db.cc
```diff
@@ -155,6 +158,7 @@ rocksdb::Status Database::TTL(const Slice &user_key, int *ttl) {
   Metadata metadata(kRedisNone, false);
   metadata.Decode(value);
```
mapleFU, line 159: When Decode happens here, would it have the right data when a 32bit size instance read 64bits data?

PragmaTwice: Sure. You can read the implementation of `Decode`.

mapleFU: oops, seems `Decode` reset it flags rather than default argument. Seems ok to me

PR #1343, opened 2023-03-22
Clean up files before and after running go tests

enjoy-binbin:
We currently do not clean up temporary files generated
by tests, This causes the workspace directory to become
larger and larger until `No space left` appears, like:
```
--- FAIL: TestString (11.14s)
    --- FAIL: TestString/SETBIT/GETBIT/BITCOUNT/BITPOS_boundary_check_(type_string) (5.71s)
        strings_test.go:360:
                Error Trace:    /root/kvrocks/incubator-kvrocks/tests/gocase/unit/type/strings/strings_test.go:360
                Error:          Received unexpected error:
                                ERR IO error: No space left on device: While appending to file: /root/xxx/gocase/workspace/TestString-xxx/db/000010.log: No space left on device
```

An example of consumption:
```
[root@ incubator-kvrocks]# du -sh ./tests/gocase/workspace/ -s
6.3G    ./tests/gocase/workspace/
```

This PR adds the cleanup_go_tests function, which will clean
up the `Test*` in the workspace directory before and after the
go test. Also added a try-finally to prevent go tests from failing
midway (e.g. KeyboardInterrupt).

Logs will be output before and after cleaning, examples are:
```
Cleanup go test workspace: /xxx/incubator-kvrocks/tests/gocase/workspace, it may take some time...
Cleanup go test workspace: /xxx/incubator-kvrocks/tests/gocase/workspace, done.
```

### DIFF
--- a/x.py
+++ b/x.py
@@ -17,9 +17,10 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import os
+import shutil
 from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter, REMAINDER
 from glob import glob
-from os import makedirs
 from pathlib import Path
 import re
 from subprocess import Popen, PIPE
@@ -104,7 +105,7 @@ def build(dir: str, jobs: Optional[int], ghproxy: bool, ninja: bool, unittest: b
     cmake_version = output.read().strip()
     check_version(cmake_version, CMAKE_REQUIRE_VERSION, "CMake")
 
-    makedirs(dir, exist_ok=True)
+    os.makedirs(dir, exist_ok=True)
 
     cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo"]
     if ghproxy:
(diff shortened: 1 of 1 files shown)


tisonkun (changes requested):
gocase has a `deleteOnExit` flag you can pass to clean up files.

tisonkun:
```
./x.py test go build -deleteOnExit true
```

enjoy-binbin:
ohh, thanks, i didn't notice it before. (We may need to document this in the docs)...
If you press ctrl + c in the middle, will it be cleaned up? I tried it and it doesn’t seem to work, if there is an error or ctrl+c in the middle

PragmaTwice:
Thanks for your contributing. 

As mentioned by @tisonkun , we already have an option for testing without temp files. And these files are useful when some tests are failed (or when the program crashed).

But feel free to do other improvements : )

tisonkun:
> We may need to document this in the docs

You're welcome to add a section or new page in https://kvrocks.apache.org/community/contributing/, following https://pingcap.github.io/tidb-dev-guide/get-started/write-and-run-unit-tests.html or other practice.

PragmaTwice:
> If you press ctrl + c in the middle, will it be cleaned up? I tried it and it doesn’t seem to work, if there is an error or ctrl+c in the middle.

Currently we need to cleanup these files manually (if we really do not need them to debug), but you are welcome to improve it (with regarding to the `deleteOnExit` option, althought I think it is a bit tricky, and maybe not a vital thing).

enjoy-binbin:
thanks all. i am new to kvrocks, sorry for the NOP PR.

about the `write-and-run-unit-tests` part, i guess i will skip it now (prevent writing bad things), i need some times to get used to the process here.  a simple flag explanation in the README is probably enough for now since we do have a `Running test cases` in this repo's README

agree with points. one thought, is that we might need to set deleteOnExit to true by default? most users do not need to go to debug. Like me, i am new, and just try to play it a while, and it cost a lot of spaces.

thank you for your time.

PR #1345, opened 2023-03-22
Fix missing stream increment batch in cluster migration

git-hulk:
Currently, the cluster slot migration will split into two stages: send snapshots and sync WAL if the snapshot was sent. For the stream, we only parsed the KVs from the snapshot and missed the WAL part.

### DIFF
--- a/src/cluster/slot_migrate.cc
+++ b/src/cluster/slot_migrate.cc
@@ -764,23 +764,10 @@ Status SlotMigrate::MigrateStream(const Slice &key, const StreamMetadata &metada
       break;
     }
 
-    // Parse values of the complex key
-    // InternalKey is adopted to get complex key's value from the formatted key returned by iterator of rocksdb
-    InternalKey inkey(iter->key(), true);
-    std::vector<std::string> values;
-    auto s = Redis::DecodeRawStreamEntryValue(iter->value().ToString(), &values);
+    auto s = WriteBatchExtractor::ExtractStreamAddCommand(true, iter->key(), iter->value(), &user_cmd);
     if (!s.IsOK()) {
-      return {Status::NotOK, fmt::format("failed to decode stream values: {}", s.Msg())};
+      return s;
     }
-
-    Slice encoded_id = inkey.GetSubKey();
-    Redis::StreamEntryID entry_id;
-    GetFixed64(&encoded_id, &entry_id.ms);
-    GetFixed64(&encoded_id, &entry_id.seq);
-
-    user_cmd.emplace_back(entry_id.ToString());
-    user_cmd.insert(user_cmd.end(), values.begin(), values.end());
-
     *restore_cmds += Redis::MultiBulkString(user_cmd, false);
     current_pipeline_size_++;
 
--- a/src/storage/batch_extractor.cc
+++ b/src/storage/batch_extractor.cc
@@ -81,9 +81,7 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
     }
 
     return rocksdb::Status::OK();
-  }
-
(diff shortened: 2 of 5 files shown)


git-hulk:
@torwig I'm not quite familiar with the XSETID command, so I didn't handle it now.

torwig:
@git-hulk Thanks for your work. `SETID` is intended for things like replication (I can say it is an internal command) to set the values of last_added_id, entries_added, max_deleted_id. I used it as the last command in slot migration to update the stream's metadata, because with the `XADD` you can't migrate last_added_id, entries_added and max_deleted_id, because it actually just adds entries but not deletes them.

git-hulk:
@torwig Thank you. So for the XSETID command, I should parse the last_added_id / entries_added / max_deleted_id, then translate it into the XSETID. Is it right?

ShooterIT:
BTW, do we implement https://github.com/apache/incubator-kvrocks/pull/577#issuecomment-1128483342?

git-hulk:
@ShooterIT Yes, it's finished but missed the WAL's increment sync.

torwig:
@git-hulk Yes, absolutely.

src/cluster/slot_migrate.cc
```diff
@@ -765,23 +765,10 @@ Status SlotMigrate::MigrateStream(const Slice &key, const StreamMetadata &metada
...
 
-    // Parse values of the complex key
-    // InternalKey is adopted to get complex key's value from the formatted key returned by iterator of rocksdb
-    InternalKey inkey(iter->key(), true);
-    std::vector<std::string> values;
-    auto s = Redis::DecodeRawStreamEntryValue(iter->value().ToString(), &values);
+    auto s = WriteBatchExtractor::ExtractStreamAddCommand(true, iter->key(), iter->value(), &user_cmd);
     if (!s.IsOK()) {
-      return {Status::NotOK, fmt::format("failed to decode stream values: {}", s.Msg())};
+      return s;
```
PragmaTwice, line 770:
```suggestion
      return s.Prefixed("failed to decode stream values");
```

src/storage/batch_extractor.cc
```diff
@@ -291,3 +302,23 @@ rocksdb::Status WriteBatchExtractor::DeleteRangeCF(uint32_t column_family_id, co
...
+
+Status WriteBatchExtractor::ExtractStreamAddCommand(bool is_slotid_encoded, const Slice &subkey, const Slice &value,
+                                                    std::vector<std::string> *command_args) {
+  InternalKey ikey(subkey, is_slotid_encoded);
+  std::string user_key = ikey.GetKey().ToString();
+  *command_args = {"XADD", user_key};
+  std::vector<std::string> values;
+  auto s = Redis::DecodeRawStreamEntryValue(value.ToString(), &values);
+  if (!s.IsOK()) {
+    return {Status::NotOK, fmt::format("failed to decode stream values: {}", s.Msg())};
```
PragmaTwice:
```suggestion
    return s.Prefixed("failed to decode stream values");
```

torwig (dismissed):

tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -542,6 +542,45 @@ func TestSlotMigrateDataType(t *testing.T) {
 		}
 	})
 
+	t.Run("MIGRATE - increment sync stream from WAL", func(t *testing.T) {
+		slot := 3
```
git-hulk:
```suggestion
		slot := 40
```

PragmaTwice (dismissed):

torwig (approved):

git-hulk:
Thanks all, merging..

PR #1358, opened 2023-03-27
Rewrite encoding functions

PragmaTwice:
We use `if constexpr`, `__builtin_bswapN` and `__builtin_memcpy` for better performance and maintainability.

image


git-hulk (approved):
Cool, LGTM.

I saw the clang replace the memcpy with __builtin_memcpy in the underlying implementation but didn't check the GCC.

```
__DEVICE__ void *memcpy(void *__a, const void *__b, size_t __c) {
  return __builtin_memcpy(__a, __b, __c);
}
```

src/common/encoding.cc
```diff
@@ -217,50 +206,40 @@ bool GetDouble(rocksdb::Slice *input, double *value) {
...
-    return ((static_cast<uint32_t>(static_cast<uint8_t>(ptr[3]))) |
-            (static_cast<uint32_t>(static_cast<uint8_t>(ptr[2])) << 8) |
-            (static_cast<uint32_t>(static_cast<uint8_t>(ptr[1])) << 16) |
-            (static_cast<uint32_t>(static_cast<uint8_t>(ptr[0])) << 24));
-  }
+  uint32_t value = 0;
+
+  __builtin_memcpy(&value, ptr, sizeof(value));
+
+  return BYTE_ORDER == LITTLE_ENDIAN ? __builtin_bswap32(value) : value;
```
torwig, line 221: @PragmaTwice Is there a possibility to use `if constexpr` here with the condition? Or it wouldn't change anything?

PragmaTwice: Yes, `if constexpr` is just a constraint in this case. Compiler will eliminate constant conditions.

PR #1367, opened 2023-04-04
Use prefix extractor to improve performance

git-hulk:
- [x] Add benchmark result to confirm that can improve the performance

Update:

The RocksDB prefix extractor looks like that can bring significant performance improvement, but NOT for Kvrocks from my side. Welcome to have a try and correct me if I was wrong.

### References

- [RocksDB Prefix Seek API](https://github.com/EighteenZi/rocksdb_wiki/blob/master/Prefix-Seek-API-Changes.md)
- [Adopting RocksDB within Manhattan](https://blog.twitter.com/engineering/en_us/topics/infrastructure/2021/adopting-rocksdb-within-manhattan)
- [RocksDB in TiKV](https://tikv.org/blog/rocksdb-in-tikv/)
- [How We Use RocksDB at Rockset](https://rockset.com/blog/how-we-use-rocksdb-at-rockset/)
* [Tuning RocksDB - Prefix Extractor](https://www.jianshu.com/p/26214e45fd4a)

### DIFF
--- a/src/storage/prefix_extractor.h
+++ b/src/storage/prefix_extractor.h
@@ -50,11 +50,7 @@ class SubkeyPrefixExtractor : public rocksdb::SliceTransform {
     // return the origin key if it's too short
     if (ptr > key.data() + key.size()) return key;
 
-    size_t prefix_len = 1 /* ns encoding size */ + ns_size + 4 /* encoding key size */ + key_size;
-    if (cluster_enabled_) {
-      prefix_len += 2;  // add slot id encoding size if the cluster mode is enabled
-    }
-    return {key.data(), prefix_len};
+    return {key.data(), ptr - key.data()};
   }
 
   bool InDomain(const rocksdb::Slice &key) const override {
--- a/src/storage/prefix_extractor.h
+++ b/src/storage/prefix_extractor.h
@@ -36,7 +36,7 @@ class SubkeyPrefixExtractor : public rocksdb::SliceTransform {
   rocksdb::Slice Transform(const rocksdb::Slice &key) const override {
     const char *ptr = key.data();
     uint8_t ns_size = ptr[0];
-    ptr += 1 + ns_size;  // drop ns encoding size and ns
+    ptr += 1 + ns_size;  // add the ns encoding size and ns
     if (cluster_enabled_) {
       // add the cluster slot size if it's enabled
       ptr += 2;
@@ -46,11 +46,12 @@ class SubkeyPrefixExtractor : public rocksdb::SliceTransform {
     if (ptr > key.data() + key.size()) return key;
     uint32_t key_size = DecodeFixed32(ptr);
 
-    ptr += 4 + key_size;  // drop key encoding size and key
+    ptr += 4 + key_size;  // add the key encoding size and key
+    ptr += 8;             // add the version encoding size
     // return the origin key if it's too short
     if (ptr > key.data() + key.size()) return key;
 
-    return {key.data(), ptr - key.data()};
+    return {key.data(), static_cast<size_t>(ptr - key.data())};
   }
 
   bool InDomain(const rocksdb::Slice &key) const override {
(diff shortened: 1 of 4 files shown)


src/storage/prefix_extractor.h
```diff
@@ -0,0 +1,67 @@
...
+
+    ptr += 4 + key_size;  // drop key encoding size and key
+    // return the origin key if it's too short
+    if (ptr > key.data() + key.size()) return key;
+
+    size_t prefix_len = 1 /* ns encoding size */ + ns_size + 4 /* encoding key size */ + key_size;
+    if (cluster_enabled_) {
+      prefix_len += 2;  // add slot id encoding size if the cluster mode is enabled
+    }
+    return {key.data(), prefix_len};
```
PragmaTwice:
```suggestion
    return {key.data(), ptr - key.data()};
```

git-hulk:
We didn't add the version into the prefix, since it may cause bugs in some test cases like [hash_test.cc#L203](https://github.com/apache/incubator-kvrocks/blob/unstable/tests/cppunit/types/hash_test.cc#L203):
```C++
/Users/hulk/code/cxx/incubator-kvrocks/tests/cppunit/types/hash_test.cc:315: Failure
Expected equality of these values:
  4 + 26
    Which is: 30
  result.size()
    Which is: 0
```

And found it was caused by the upper bound in `SeekToLast` after diving into the reason. The root cause is the `Seek` API will also use the bloom filter to probe the existence of the prefix when enabling the prefix extractor, then it will be false if we added the version into the prefix when seeking with the upper bound(which is the version + 1).

```C++
void DBIter::SeekToLast() {
  if (iterate_upper_bound_ != nullptr) {
    // Seek to last key strictly less than ReadOptions.iterate_upper_bound.
    SeekForPrev(*iterate_upper_bound_);
    const bool is_ikey = (timestamp_size_ > 0 && timestamp_lb_ != nullptr);
    Slice k = Valid() ? key() : Slice();
    if (is_ikey && Valid()) {
      k.remove_suffix(kNumInternalBytes + timestamp_size_);
    }
```

We can work around this issue by removing the version like the current implementation, but adding the version would help a lot if there have many deleted tombstones. So another way is to avoid using the upper bound when using `SeekToLast`.

git-hulk:
Update: I have tested if the prefix extractor can bring better performance when seeking, but the result doesn't have any difference while comparing to adding the version into the prefix. Also, avoiding using the upper bound would introduce some complexity in the codebase, so let's remove the version from the prefix extractor.

The performance benefit of the prefix extractor, I tested with below random keys on my MacBook, also can't see the improvement:

```
# Create Hash keys with hset 

❯ redis-benchmark -p 6666 -n 1000000 -r 50000 hset key__rand_int__ field__rand_int__ value

# Benchmark with HRANGEBYLEX

❯ redis-benchmark -p 6666 -n 1000000 -r 50000 HRANGEBYLEX key__rand_int__ "[field000000000000" "[field000000005000"

```

So we can keep this PR in draft mode, and users can have a test if they want to try.

mapleFU:
Do we have some benchmark or micro-benchmark here? Seems MyRocks requires that, so RocksDB implements it. Can we take some benchmark from MyRocks?

git-hulk:
Thanks @mapleFU, do you have any references?

mapleFU:
1. MyRocks Paper, Section 3.2.2.1: https://vldb.org/pvldb/vol13/p3217-matsunobu.pdf
2. https://www.slideshare.net/matsunobu/myrocks-deep-dive/58
3. https://www.cidrdb.org/cidr2017/papers/p82-dong-cidr17.pdf

> Preﬁx Bloom ﬁlters. Bloom ﬁlters do not help with range queries. We have developed a preﬁx Bloom ﬁlter that helps with range queries, based on the observation that many range queries are often over a preﬁx; e.g., the userid part of a (userid,timestamp) key or postid of a (postid,likerid) key. We allow users to deﬁne preﬁx extractors to deterministically extract a preﬁx part of the key from which we construct a Bloom ﬁlter. When querying a range, the user can specify that the query is on a deﬁned preﬁx. We have found this optimization reduces read ampliﬁcation (and attendant CPU overheads) on range queries by up to 64% on our systems.

git-hulk:
Close this PR since we cannot find any performance improvements. Can reopen it if anyone is interested in this.

PR #1373, opened 2023-04-06
Minor refactoring of slot migration source code

torwig:
Unused variables were removed.
Some variables and methods were renamed.
Destination socket descriptor wrapped by `UniqueFD` and moved outside `SlotMigrationJob` (now it contains only migration parameters ).
Enum classes were used instead of plain enums.
Move the logic of setting the forbidden slot into `SetForbiddenSlot` and call it explicitly between two calls of syncing from WAL.

### DIFF
--- a/src/cluster/slot_migrate.h
+++ b/src/cluster/slot_migrate.h
@@ -128,8 +128,8 @@ class SlotMigrator : public Redis::Database {
   void ApplyMigrationSpeedLimit() const;
   Status GenerateCmdsFromBatch(rocksdb::BatchResult *batch, std::string *commands);
   Status MigrateIncrementData(std::unique_ptr<rocksdb::TransactionLogIterator> *iter, uint64_t end_seq);
-  Status SyncWalBeforeForbiddingSlot();
-  Status SyncWalAfterForbiddingSlot();
+  Status SyncWALBeforeForbiddingSlot();
+  Status SyncWALAfterForbiddingSlot();
   void SetForbiddenSlot(int16_t slot);
 
   enum class ParserState { ArrayLen, BulkLen, BulkData, OneRspEnd };
--- a/src/cluster/slot_migrate.h
+++ b/src/cluster/slot_migrate.h
@@ -34,6 +34,7 @@
 #include <utility>
 #include <vector>
 
+#include "common/fd_util.h"
 #include "config.h"
 #include "encoding.h"
 #include "parse_util.h"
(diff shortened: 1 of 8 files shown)


src/cluster/slot_migrate.h
```diff
@@ -44,86 +45,75 @@
...
+  void RunMigrationProcess();
+  bool IsTerminated() { return thread_state_ == ThreadState::Terminated; }
+  Status StartMigration();
   Status SendSnapshot();
-  Status SyncWal();
-  Status Success();
-  Status Fail();
+  Status SyncWAL();
+  Status FinishSuccessfulMigration();
+  Status FinishFailedMigration();
```
tisonkun, line 110: These two method name changes seem wordy.

src/server/server.h
```diff
@@ -111,7 +111,7 @@ class ServerLogData {
 };
 
 class SlotImport;
-class SlotMigrate;
+class SlotMigrator;
```
tisonkun, line 114: If we just name `SlotImport` and `SlotMigrate`, let's keep it as is.

tisonkun (commented):
I'll suggest you separate different refactor aspects in different commits the next time. Otherwise, it's hard to review..

src/server/server.h
torwig: I got your point. The reason is that I didn't touch `slot_import.h/.cc` in this PR.

src/cluster/slot_migrate.h
torwig: I thought that new names clearly describe what they do.

torwig:
@tisonkun You are absolutely right: it's hard to review because of so many changes.

src/cluster/slot_migrate.h
```diff
@@ -133,52 +123,51 @@ class SlotMigrate : public Redis::Database {
...
   Status MigrateBitmapKey(const InternalKey &inkey, std::unique_ptr<rocksdb::Iterator> *iter,
                           std::vector<std::string> *user_cmd, std::string *restore_cmds);
+
   Status SendCmdsPipelineIfNeed(std::string *commands, bool need);
   void ApplyMigrationSpeedLimit() const;
   Status GenerateCmdsFromBatch(rocksdb::BatchResult *batch, std::string *commands);
   Status MigrateIncrementData(std::unique_ptr<rocksdb::TransactionLogIterator> *iter, uint64_t end_seq);
-  Status SyncWalBeforeForbidSlot();
-  Status SyncWalAfterForbidSlot();
+  Status SyncWalBeforeForbiddingSlot();
```
git-hulk: Should we also change `Wal` to `WAL` for keeping consistent with previous function `SyncWAL`

git-hulk (commented):
General LGTM, left one comment.

src/cluster/slot_migrate.h
torwig: Done

git-hulk (approved):

torwig:
@PragmaTwice Could you please have a look at my PR? As @tisonkun advised, I can split it into several distinct PRs to make the review more convenient. What do you think?

PragmaTwice (approved):
Looks fine to me, expect `SlotImportor` like `SlotMigrator` though.

torwig:
@PragmaTwice I'll handle `SlotImport` in the next PRs.

git-hulk:
Thanks all, merging...

PR #1375, opened 2023-04-08
Add MEMORY USAGE command as an alias to existing DISK USAGE command

torwig:
Since Redis has the `MEMORY` command, tools like `RedisInsight` use it to show information to the user.
Currently, `RedisInsight` can't show memory usage by keys because `Kvrocks` doesn't support the `MEMORY USAGE` command.
Solution: make the `MEMORY USAGE` command an alias to the `DISK USAGE` command. 
Currently, the `MEMORY USAGE` command doesn't support the optional `SAMPLES` option.
The integration test basically checks that `Kvrocks` support the new command and the response is an integer value conforming to the Redis spec.

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -296,6 +296,8 @@ class CommandDisk : public Commander {
   }
 };
 
+class CommandMemory : public CommandDisk {};
+
 class CommandRole : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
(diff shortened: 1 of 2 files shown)


aleksraiden:
Good, thanks! Now I use a redisInsight tool im work and we have two separated service, redis and kvroks, and this features very good to me!

git-hulk (approved):

PragmaTwice (approved):

git-hulk:
Thanks all, merging...

PR #1380, opened 2023-04-12
Fix disordered stack pop in replyToRedisReply

PragmaTwice:
In some code path, `replyToRedisReply` does not pop the value after reply, which is expected to pop.

In this PR, we do not pop the value in any path of `replyToRedisReply` intead, making it more friendly to use.


git-hulk (approved):
Good catch

torwig (approved):

Issue #1381, opened 2023-04-12
SetClusterNodes  allow to match localhost and real ip when listening on 0.0.0.0


git-hulk:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

Currently, `SetClusterNodes` can't match the node if the instance is listening on 0.0.0.0 instead of the real IP. 

For example, the cluster topology is: `lohMbHeln658feDcRVKYb4tT7XClAm3Zr4awKuoK 127.0.0.1 6666 master -  0-16383`, but the listen address is 0.0.0.0, then it won't match the cluster node since the listen IP is not strictly equal to cluster's node IP 127.0.0.1(as well as not localhost IP like 172.14.1.x).

This would confuse users because the cluster didn't work as their expected.

### Solution

Check the IP address is matched by the real listening IP addresses instead of the bind address directly when using 0.0.0.0.

https://github.com/apache/incubator-kvrocks/blob/unstable/src/cluster/cluster.cc#L211

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

zevin02:
please assign to me

PragmaTwice:
> please assign to me

Great! Assigned.

PR #1384, opened 2023-04-13
Fix missing to set the cluster node id when loading

git-hulk:
we supported loading the cluster nodes from the local file, but we just ignored the cluster node id after parsing. So it can't match the cluster node since the local node id is empty, and it will fall back to check if the listening IP and port are matched.

For example, we create a cluster with a single node and listen on 0.0.0.0, then we would get the below output after restarting:

```shell
$ 127.0.0.1:6666> cluster nodes
3dQzcqeiTpSqPstxtD1utGuU52fPkGnoIHPO8KgL 127.0.0.1:6666@16666 master - 1681396166414 1681396166415 0 connected 0-16383
```
it missed a `myself` field because didn't use the node id from the local file, then it will fall back to IP/Port, but the listening IP 0.0.0.0 is unmatched with 127.0.0.1, this issue was submitted in #1381.

After this PR, we can get the correct result:

```shell
$ 127.0.0.1:6666> cluster nodes
3dQzcqeiTpSqPstxtD1utGuU52fPkGnoIHPO8KgL 127.0.0.1:6666@16666 myself,master - 1681396232426 1681396232427 0 connected 0-16383
```

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -643,6 +643,8 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
       return {Status::NotOK, fmt::format("unknown key: {}", key)};
     }
   }
+
+  myid_ = id;
   return SetClusterNodes(nodes_info, version, false);
 }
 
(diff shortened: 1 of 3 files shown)


torwig (dismissed):
LGTM

PragmaTwice (dismissed):

PragmaTwice (approved):

torwig (approved):

git-hulk:
Thanks all, merging...

PR #1385, opened 2023-04-13
Fix member name convension for readability-identifier-naming

PragmaTwice:
- rename all class field member names to something like `lower_case_`
- add a `--fix` option for tidy to `x.py`


torwig (approved):
LGTM

git-hulk (approved):

PR #1392, opened 2023-04-17
Convert 0.0.0.0 to local ip addresses when  matching the cluster node

### LINKED ISSUE #1381: SetClusterNodes  allow to match localhost and real ip when listening on 0.0.0.0

zevin02:
fixes: #1381

- Add a function to get all local IP addresses

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -20,9 +20,11 @@
 
 #include "cluster.h"
 
+#include <arpa/inet.h>
 #include <config/config_util.h>
+#include <ifaddrs.h>
+#include <sys/types.h>
 
-#include <algorithm>
 #include <cstring>
 #include <fstream>
 #include <memory>
@@ -157,6 +159,33 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
 
   return Status::OK();
 }
+// Get all local IP addresses.
+std::vector<std::string> GetLocalIpAddresses() {
+  std::vector<std::string> ip_addresses;
+
+  struct ifaddrs *if_addr_struct = nullptr;
+  struct ifaddrs *ifa = nullptr;
+  void *tmp_addr_ptr = nullptr;
+
+  getifaddrs(&if_addr_struct);
+
+  for (ifa = if_addr_struct; ifa; ifa = ifa->ifa_next) {
+    if (!ifa->ifa_addr) {
+      continue;
+    }
+    if (ifa->ifa_addr->sa_family == AF_INET) {
+      // check it is IP4 and not a loopback address
+      tmp_addr_ptr = &((struct sockaddr_in *)ifa->ifa_addr)->sin_addr;
+      char address_buffer[INET_ADDRSTRLEN];
+      inet_ntop(AF_INET, tmp_addr_ptr, address_buffer, INET_ADDRSTRLEN);
+      ip_addresses.emplace_back(std::string(address_buffer));
+    }
+  }
+
+  if (if_addr_struct) freeifaddrs(if_addr_struct);
+
+  return ip_addresses;
+}
 
 // cluster setnodes $all_nodes_info $version $force
 // one line of $all_nodes: $node_id $host $port $role $master_node_id $slot_range
@@ -199,16 +228,33 @@ Status Cluster::SetClusterNodes(const std::string &nodes_str, int64_t version, b
       size_++;
     }
   }
+  // Used to check if listening on "0.0.0.0“
+  bool is_listen_all_ip = false;
 
-  // Find myself
+  std::unique_ptr<std::vector<std::string>> local_hosts(nullptr);
+  if (std::find(binds_.begin(), binds_.end(), "0.0.0.0") != binds_.end()) {
+    is_listen_all_ip = true;
+    local_hosts = std::make_unique<std::vector<std::string>>(GetLocalIpAddresses());
+  }
+
   if (myid_.empty() || force) {
     for (auto &n : nodes_) {
-      if (n.second->port == port_ && std::find(binds_.begin(), binds_.end(), n.second->host) != binds_.end()) {
-        myid_ = n.first;
-        break;
+      if (is_listen_all_ip) {
+        // If the IP being listened on is 0.0.0.0, we need to compare the node's IP address with all local IP addresses.
+        if (n.second->port == port_ &&
+            std::find(local_hosts->begin(), local_hosts->end(), n.second->host) != local_hosts->end()) {
+          myid_ = n.first;
+          break;
+        }
+      } else {
+        if (n.second->port == port_ && std::find(binds_.begin(), binds_.end(), n.second->host) != binds_.end()) {
+          myid_ = n.first;
+          break;
+        }
       }
     }
   }
+
   myself_ = nullptr;
   if (!myid_.empty() && nodes_.find(myid_) != nodes_.end()) {
     myself_ = nodes_[myid_];
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -20,18 +20,15 @@
 
 #include "cluster.h"
 
-#include <arpa/inet.h>
 #include <config/config_util.h>
-#include <ifaddrs.h>
-#include <sys/types.h>
 
 #include <cstring>
 #include <fstream>
 #include <memory>
 
 #include "commands/commander.h"
+#include "common/io_util.h"
 #include "fmt/format.h"
-#include "io_util.h"
 #include "parse_util.h"
 #include "replication.h"
 #include "server/server.h"
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -159,33 +159,7 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
 
   return Status::OK();
 }
-// Get all local IP addresses.
-std::vector<std::string> GetLocalIpAddresses() {
-  std::vector<std::string> ip_addresses;
 
-  struct ifaddrs *if_addr_struct = nullptr;
-  struct ifaddrs *ifa = nullptr;
-  void *tmp_addr_ptr = nullptr;
-
-  getifaddrs(&if_addr_struct);
-
-  for (ifa = if_addr_struct; ifa; ifa = ifa->ifa_next) {
-    if (!ifa->ifa_addr) {
-      continue;
-    }
-    if (ifa->ifa_addr->sa_family == AF_INET) {
-      // check it is IP4 and not a loopback address
-      tmp_addr_ptr = &((struct sockaddr_in *)ifa->ifa_addr)->sin_addr;
-      char address_buffer[INET_ADDRSTRLEN];
-      inet_ntop(AF_INET, tmp_addr_ptr, address_buffer, INET_ADDRSTRLEN);
-      ip_addresses.emplace_back(std::string(address_buffer));
-    }
-  }
-
-  if (if_addr_struct) freeifaddrs(if_addr_struct);
-
-  return ip_addresses;
-}
 
 // cluster setnodes $all_nodes_info $version $force
 // one line of $all_nodes: $node_id $host $port $role $master_node_id $slot_range
@@ -230,11 +204,12 @@ Status Cluster::SetClusterNodes(const std::string &nodes_str, int64_t version, b
   }
   // Used to check if listening on "0.0.0.0“
   bool is_listen_all_ip = false;
-
+  // Store all the local ip
   std::unique_ptr<std::vector<std::string>> local_hosts(nullptr);
+
   if (std::find(binds_.begin(), binds_.end(), "0.0.0.0") != binds_.end()) {
     is_listen_all_ip = true;
-    local_hosts = std::make_unique<std::vector<std::string>>(GetLocalIpAddresses());
+    local_hosts = std::make_unique<std::vector<std::string>>(util::GetLocalIpAddresses());
   }
 
   if (myid_.empty() || force) {
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -160,7 +160,6 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
   return Status::OK();
 }
 
-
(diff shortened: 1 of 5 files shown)


src/cluster/cluster.cc
```diff
@@ -157,7 +159,34 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
 
   return Status::OK();
 }
+// Get all local IP addresses.
+std::vector<std::string> getLocalIPAddresses() {
+  std::vector<std::string> ipAddresses;
```
torwig: Hi @zevin02 , thank you for your contribution.
Please, rename the function's local variables like `ip_addresses`.
Also, you can use the `nullptr` keyword instead of `NULL` and simplify checks like `if (pointer != NULL)` by using `if (pointer)`

src/cluster/cluster.cc
```diff
@@ -157,7 +159,34 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
...
+    if (!ifa->ifa_addr) {
+      continue;
+    }
+    if (ifa->ifa_addr->sa_family == AF_INET) {
+      // check it is IP4 and not a loopback address
+      tmpAddrPtr = &((struct sockaddr_in *)ifa->ifa_addr)->sin_addr;
+      char addressBuffer[INET_ADDRSTRLEN];
+      inet_ntop(AF_INET, tmpAddrPtr, addressBuffer, INET_ADDRSTRLEN);
+      std::string ipAddress(addressBuffer);
+      ipAddresses.push_back(move(ipAddress));
```
torwig: I would add `std::` prefix to the `move` function.

torwig:
@zevin02 It's very handy to run `./x.py format` before committing the changes - it will format the C++ source code automatically so the CI's linter will be happy.

PragmaTwice:
Hi @zevin02, thanks for your contribution!

We setup lots of code checks in our CI, which requires you to pass for better code quality.

For example, from a glance to your code, maybe these checks will fail: 
- local variables must be lower_case with underscore,
- NULL must be replaced with nullptr

You can check the CI log for more details and for how to modify your code to pass the CI.

It is recommended to try some commands to solve these problems, but maybe not all problems:
- `./x.py format`
- `./x.py check tidy --fix`

Another thing is that, it would be better to have some unit tests for this function before it is merged : )

src/cluster/cluster.cc
git-hulk: It looks unnecessary to get IP addresses, we can do it once if finds the listening address was on 0.0.0.0 when starting.

zevin02: Can I understand that if the server listens to 0.0.0.0, any cluster node with any ip on the same port needs to be matched, or what?

git-hulk: No actually, we can only match the IP addresses that the current node is listening. And 0.0.0.0 is a special case since it represents listening on all IP addresses in this node, we need to convert to the real IP(like 192.168.1.4) + loopback IP(127.0.0.1) before matching.

zevin02: So is there a problem with my above operation? If it is 0.0.0.0, get all local ip and cluster node ip for matching

git-hulk: The implementation of `getLocalIPAddresses` looks good. My point is that we should use all IP addresses to match only if the listening address is `0.0.0.0`.

zevin02: Got it, thanks

src/cluster/cluster.cc
```diff
@@ -201,14 +230,20 @@ Status Cluster::SetClusterNodes(const std::string &nodes_str, int64_t version, b
   }
 
   // Find myself
+  // Get all local IP addresses.
+  std::vector<std::string> localhost = getLocalIPAddresses();
   if (myid_.empty() || force) {
     for (auto &n : nodes_) {
-      if (n.second->port == port_ && std::find(binds_.begin(), binds_.end(), n.second->host) != binds_.end()) {
-        myid_ = n.first;
+      //If the IP being listened on is 0.0.0.0, we need to compare the node's IP address with all local IP addresses.
```
enjoy-binbin:
```suggestion
      // If the IP being listened on is 0.0.0.0, we need to compare the node's IP address with all local IP addresses.
```

src/cluster/cluster.cc
```diff
@@ -157,6 +159,33 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
 
   return Status::OK();
 }
+// Get all local IP addresses.
+std::vector<std::string> GetLocalIpAddresses() {
```
PragmaTwice: I think it would be better to put this function in `common/io_util.h/cc`.

src/cluster/cluster.cc
```diff
@@ -157,6 +159,33 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
...
+    if (!ifa->ifa_addr) {
+      continue;
+    }
+    if (ifa->ifa_addr->sa_family == AF_INET) {
+      // check it is IP4 and not a loopback address
+      tmp_addr_ptr = &((struct sockaddr_in *)ifa->ifa_addr)->sin_addr;
+      char address_buffer[INET_ADDRSTRLEN];
+      inet_ntop(AF_INET, tmp_addr_ptr, address_buffer, INET_ADDRSTRLEN);
+      ip_addresses.emplace_back(std::string(address_buffer));
+    }
```
PragmaTwice: It seems the IPv6 part is missing, which is supported by kvrocks in other components.

src/cluster/cluster.cc
```diff
@@ -157,6 +159,33 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
 
   return Status::OK();
 }
+// Get all local IP addresses.
+std::vector<std::string> GetLocalIpAddresses() {
+  std::vector<std::string> ip_addresses;
+
+  struct ifaddrs *if_addr_struct = nullptr;
```
PragmaTwice: Seems the `struct` prefix is useless here.

src/cluster/cluster.cc
```diff
@@ -157,6 +159,33 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
...
+  struct ifaddrs *ifa = nullptr;
+  void *tmp_addr_ptr = nullptr;
```
PragmaTwice: Please define these variables closer to where they are used.

src/cluster/cluster.cc
```diff
@@ -157,6 +159,33 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
...
+  for (ifa = if_addr_struct; ifa; ifa = ifa->ifa_next) {
+    if (!ifa->ifa_addr) {
+      continue;
+    }
+    if (ifa->ifa_addr->sa_family == AF_INET) {
+      // check it is IP4 and not a loopback address
+      tmp_addr_ptr = &((struct sockaddr_in *)ifa->ifa_addr)->sin_addr;
+      char address_buffer[INET_ADDRSTRLEN];
+      inet_ntop(AF_INET, tmp_addr_ptr, address_buffer, INET_ADDRSTRLEN);
+      ip_addresses.emplace_back(std::string(address_buffer));
```
```suggestion
      ip_addresses.emplace_back(address_buffer);
```

src/cluster/cluster.cc
```diff
@@ -157,6 +159,33 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
...
+    }
+  }
+
+  if (if_addr_struct) freeifaddrs(if_addr_struct);
```
PragmaTwice: You can try to use `std::unique_ptr` or `ScopeExit` for `if_addr_struct` to avoid manually free.

zevin02: got it

git-hulk: @zevin02 can just append fix commits directly without rebasing, then reviewers can review the new changes only.

zevin02: OK, I will submit PR like this in the future

zevin02: I tried a bit using smart pointers to avoid manual freeing, it seems that there is some difficulty for me, can you give me some advice

zevin02:
@PragmaTwice I think I solved all the problems, can you give me a review

PragmaTwice:
> @PragmaTwice I think I solved all the problems, can you give me a review

Sure! But as I mentioned in https://github.com/apache/incubator-kvrocks/pull/1392#issuecomment-1511603694, you need first to make the CI pass, which currently failed.

```
./x.py format
```

src/cluster/cluster.cc
```diff
@@ -20,9 +20,11 @@
 
 #include "cluster.h"
 
+#include <arpa/inet.h>
 #include <config/config_util.h>
+#include <ifaddrs.h>
+#include <sys/types.h>
```
PragmaTwice: Seems these newly added headers are no longer needed in this file.

src/common/io_util.h
```diff
@@ -20,7 +20,9 @@
 
 #pragma once
 
+#include <ifaddrs.h>
 #include <netinet/in.h>
+#include <sys/types.h>
```
PragmaTwice: These headers can be put into `.cc`, rather than `.h`.

src/cluster/cluster.h
```diff
@@ -30,6 +30,7 @@
 #include <vector>
 
 #include "commands/commander.h"
+#include "common/io_util.h"
```
PragmaTwice, line 33: ditto

src/cluster/cluster.cc
```diff
@@ -199,16 +201,34 @@ Status Cluster::SetClusterNodes(const std::string &nodes_str, int64_t version, b
       size_++;
     }
   }
+  // Used to check if listening on "0.0.0.0“
+  bool is_listen_all_ip = false;
+  // Store all the local ip
+  std::unique_ptr<std::vector<std::string>> local_hosts(nullptr);
+
+  if (std::find(binds_.begin(), binds_.end(), "0.0.0.0") != binds_.end()) {
```
PragmaTwice: Seems the ipv6 address INADDR6_ANY (`::`) is missing.

src/cluster/cluster.cc
```diff
@@ -199,16 +201,34 @@ Status Cluster::SetClusterNodes(const std::string &nodes_str, int64_t version, b
       size_++;
     }
   }
+  // Used to check if listening on "0.0.0.0“
+  bool is_listen_all_ip = false;
+  // Store all the local ip
+  std::unique_ptr<std::vector<std::string>> local_hosts(nullptr);
```
```suggestion
  std::unique_ptr<std::vector<std::string>> local_hosts;
```

PragmaTwice (commented):
The rest looks good to me. cc @git-hulk

git-hulk:
> The rest looks good to me. cc @git-hulk

@PragmaTwice I do a minor refactor, can help to take a look again.

src/common/io_util.cc
```diff
@@ -312,6 +314,53 @@ int AeWait(int fd, int mask, int timeout) {
...
+
+  // If binds contains 0.0.0.0, we should resolve ip addresses and check it
+  if (std::find(binds.begin(), binds.end(), "0.0.0.0") != binds.end()) {
+    auto local_ip_addresses = GetLocalIPAddresses();
+    return std::find(local_ip_addresses.begin(), local_ip_addresses.end(), ip) != local_ip_addresses.end();
+  }
+  return false;
+}
+
+// Get all local IP addresses.
```
torwig: I believe the function's name is so descriptive that we can omit this comment.

src/common/io_util.cc
```diff
@@ -312,6 +314,53 @@ int AeWait(int fd, int mask, int timeout) {
...
+std::vector<std::string> GetLocalIPAddresses() {
+  std::vector<std::string> ip_addresses;
+  ifaddrs *if_addr_struct = nullptr;
+  // Use unique_ptr for if_addr_struct to avoid manually free
```
torwig: Also here we can omit the comment in my opinion.

src/common/io_util.cc
```diff
@@ -312,6 +314,53 @@ int AeWait(int fd, int mask, int timeout) {
...
+  }
+  ifaddrs_ptr.reset(if_addr_struct);
+
+  for (ifaddrs *ifa = if_addr_struct; ifa; ifa = ifa->ifa_next) {
+    if (!ifa->ifa_addr) {
+      continue;
+    }
+    void *tmp_addr_ptr = nullptr;
+    if (ifa->ifa_addr->sa_family == AF_INET) {
+      // check it is IP4 and not a loopback address
```
torwig: The comment states that there is a check that it's not a loopback address, but I don't see any checks. Is everything OK here?
The same for the `else` branch.

torwig (commented):
Generally, LGTM, but I have a few comments.

src/common/io_util.cc
zevin02: I rechecked this code, I didn't check the local loopback address, I think it's ok, so I should delete the extra comment after that, right? or should I check if it's the local loopback address?

git-hulk: Yes, the comment should be removed if we didn't check that. It looks no necessary to check if it's a loopback address.

PragmaTwice: The comment should be what the code does, otherwise it must be `TODO` or `FIXME`.

torwig (approved):
LGTM

git-hulk (approved):

PragmaTwice (approved):

PragmaTwice:

PR #1394, opened 2023-04-18
Add TIME command as Redis

### LINKED ISSUE #1393: Implement a TIME command from Redis
### Search before asking

### Motivation

For more compatibility with original Redis spec and because a some other command like XADD etc. use internal time as identificatory of items. 

Redis spec are here - [https://redis.io/commands/time/](https://redis.io/commands/time/)  

### Solution

_No response_

enjoy-binbin:
This PR adds the TIME command. The TIME command returns the current
server time as a two items lists: a Unix timestamp and the amount of
microseconds already elapsed in the current second.

Example:
```
127.0.0.1:6666> time
1) "1681828105"
2) "810037"
```

This closes #1393

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -957,7 +957,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandAuth>("auth", 2, "read-only ok-loadin
                         MakeCmdAttr<CommandDebug>("debug", -2, "read-only exclusive", 0, 0, 0),
                         MakeCmdAttr<CommandCommand>("command", -1, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandEcho>("echo", 2, "read-only", 0, 0, 0),
-                        MakeCmdAttr<CommandTime>("time", 1, "ok-loading", 0, 0, 0),
+                        MakeCmdAttr<CommandTime>("time", 1, "read-only ok-loading", 0, 0, 0),
                         MakeCmdAttr<CommandDisk>("disk", 3, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandMemory>("memory", 3, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandHello>("hello", -1, "read-only ok-loading", 0, 0, 0),
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -646,7 +646,7 @@ class CommandTime : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     uint64_t now = util::GetTimeStampUS();
-    uint64_t s = now / 1000 / 1000;  // unix time in seconds.
+    uint64_t s = now / 1000 / 1000;         // unix time in seconds.
     uint64_t us = now - (s * 1000 * 1000);  // microseconds.
 
     *output = redis::MultiLen(2);
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -25,6 +25,7 @@
 #include "server/redis_connection.h"
 #include "server/server.h"
 #include "stats/disk_stats.h"
+#include "time_util.h"
 
 namespace redis {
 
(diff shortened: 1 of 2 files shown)


enjoy-binbin:
having trouble with how to add a time test case...

git-hulk:
> having trouble with how to add a time test case...

@enjoy-binbin Do you mean that you don't know how to check if the time command is correct? If yes, we can only check if the seconds field is in the range of [now-2, now+2] by using `require.WithinRange`.

And if you don't know where to add the test case, can just imitate the [ping command test case](https://github.com/apache/incubator-kvrocks/tree/unstable/tests/gocase/unit/ping).

src/commands/cmd_server.cc
```diff
@@ -941,6 +957,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandAuth>("auth", 2, "read-only ok-loadin
                         MakeCmdAttr<CommandDebug>("debug", -2, "read-only exclusive", 0, 0, 0),
                         MakeCmdAttr<CommandCommand>("command", -1, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandEcho>("echo", 2, "read-only", 0, 0, 0),
+                        MakeCmdAttr<CommandTime>("time", 1, "ok-loading", 0, 0, 0),
```
torwig: I think you can add `read-only` to the attributes of the `TIME` command.

enjoy-binbin:
@git-hulk thanks. i got it now

src/commands/cmd_server.cc
enjoy-binbin: updated, thanks for the catch

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -40,6 +40,24 @@ func TestIntrospection(t *testing.T) {
 	rdb := srv.NewClient()
 	defer func() { require.NoError(t, rdb.Close()) }()
 
+	t.Run("TIME", func(t *testing.T) {
+		now_ms := int(time.Now().Unix())
```
torwig: @enjoy-binbin Just a little fix - in Go, the snakeCase is preferred, so you can rename the variable to `now`/`nowMs`.
P.S. According to the docs, `time.Now().Unix()` returns seconds not milliseconds. `UnixMilli()` returns milliseconds if you need milliseconds. Otherwise, the variable that holds seconds shouldn't hold seconds.

enjoy-binbin: oh right, bad habit... thanks a lot.

src/commands/cmd_server.cc
```diff
@@ -641,6 +642,21 @@ class CommandEcho : public Commander {
   }
 };
 
+class CommandTime : public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    uint64_t now = util::GetTimeStampUS();
+    uint64_t s = now / 1000 / 1000;  // unix time in seconds.
```
```suggestion
    uint64_t s = now / 1000 / 1000;         // unix time in seconds.
```

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -40,6 +40,24 @@ func TestIntrospection(t *testing.T) {
...
+		nowUnix := int(time.Now().Unix())
+
+		r := rdb.Do(ctx, "TIME")
+		vs, err := r.Slice()
+		require.NoError(t, err)
+
+		s, err := strconv.Atoi(vs[0].(string))
+		require.NoError(t, err)
+		require.Greater(t, s, nowUnix - 2)
+		require.Less(t, s, nowUnix + 2)
```
git-hulk: No worry about the Go fmt, we can auto-fix it in the future.

```suggestion
		require.Greater(t, s, nowUnix-2)
		require.Less(t, s, nowUnix+2)
```

torwig (approved):
LGTM

git-hulk (approved):

git-hulk:
Thanks all, merging...

PR #1397, opened 2023-04-20
Fix CONFIG REWRITE append duplicate newlines

### LINKED ISSUE #1396: config rewrite bug
### Search before asking

### Version

2.3.0

### Minimal reproduce step

namespace add xxx
config rewrite

no extra blank lines in kvrocks.conf

### What did you see instead?

First create 1w namespace, then config rewrite.
There are 1w extra blank lines after config rewrite

### Anything Else?

_No response_

enjoy-binbin:
In the old code, everytime we perform a config rewrite, an
additional newline is appended, see #1396

Because iostream::eof will only return true after reading the
end of the stream. It does not indicate, that the next read will
be the end of the stream.

So everytime config rewrite is performed, the old loop will
append a newline in `lines`, and then we will append it to the
conf file.

In addition, the same modification has been made to other similar
places. This PR fixes #1396

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -726,7 +726,7 @@ Status Config::Load(const CLIOptions &opts) {
 
     std::string line;
     int line_num = 1;
-    while (in.good() && std::getline(*in, line)) {
+    while (in->good() && std::getline(*in, line)) {
       if (auto s = parseConfigFromString(line, line_num); !s.IsOK()) {
         return s.Prefixed(fmt::format("at line #L{}", line_num));
       }
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -726,8 +726,7 @@ Status Config::Load(const CLIOptions &opts) {
 
     std::string line;
     int line_num = 1;
-    while (!in->eof()) {
-      std::getline(*in, line);
+    while (in.good() && std::getline(*in, line)) {
       if (auto s = parseConfigFromString(line, line_num); !s.IsOK()) {
         return s.Prefixed(fmt::format("at line #L{}", line_num));
       }
@@ -829,8 +828,7 @@ Status Config::Rewrite() {
   std::ifstream file(path_);
   if (file.is_open()) {
     std::string raw_line;
-    while (std::getline(file, raw_line)) {
-      if (file.eof()) break;
+    while (file.good() && std::getline(file, raw_line)) {
       auto parsed = ParseConfigLine(raw_line);
       if (!parsed || parsed->first.empty()) {
         lines.emplace_back(raw_line);
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -829,8 +829,8 @@ Status Config::Rewrite() {
   std::ifstream file(path_);
   if (file.is_open()) {
     std::string raw_line;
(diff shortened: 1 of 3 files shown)


enjoy-binbin:
I tested it locally (with the CONFIG REWRITE)
and the ref: https://stackoverflow.com/questions/5605125/why-is-iostreameof-inside-a-loop-condition-i-e-while-stream-eof-cons

git-hulk:
@enjoy-binbin Thanks for your efforts. ❤️

git-hulk (dismissed):

torwig (dismissed):
LGTM

src/config/config.cc
```diff
@@ -829,8 +829,8 @@ Status Config::Rewrite() {
   std::ifstream file(path_);
   if (file.is_open()) {
     std::string raw_line;
-    while (!file.eof()) {
-      std::getline(file, raw_line);
+    while (std::getline(file, raw_line)) {
+      if (file.eof()) break;
```
PragmaTwice: If the text file has no newline after the last line, this line will be ignored.

e.g.
```python
file.write("a\nb\nc"); # the `c` line will be ignored
```

Hence I think it is not a good change.

PragmaTwice (changes requested):
Thanks for your patch, but seems changes are required.

PragmaTwice:
Here is a simple test to demonstrate this point:
```
❯ cat y.cpp
#include <fstream>
#include <iostream>
#include <string>

int main() {
std::ifstream file("hello.txt");
std::string line;
while (std::getline(file, line)) {
    if (file.eof()) break;
    std::cout << "> " << line << std::endl;
}
}

❯ python -c "file=open('hello.txt', 'w');file.write('a\nb\nc');file.close()"

❯ cat hello.txt
a
b
c%

❯ ./y
> a
> b

❯
```

enjoy-binbin:
@PragmaTwice thanks for the detailed explanation! do you have any suggestions? i may need some time to figure it out
do you think if i put the `if (file.eof()) break;` in the last will help?

PragmaTwice:
> @PragmaTwice thanks for the detailed explanation! do you have any suggestions? i may need some time to figure it out do you think if i put the `if (file.eof()) break;` in the last will help?

Not sure the eof check is necessary, maybe something like `while(file.good() && getline(...))` is ok.

The call to `getline` will return the stream itself, so `operator bool` of the stream is called, and it is equal to `!stream.fail()`.

enjoy-binbin:
i change to use `while(file.good() && getline(...))` for now, i also tested it in local, it is fine

src/config/config.cc
```diff
@@ -726,8 +726,7 @@ Status Config::Load(const CLIOptions &opts) {
 
     std::string line;
     int line_num = 1;
-    while (!in->eof()) {
-      std::getline(*in, line);
+    while (in.good() && std::getline(*in, line)) {
```
```suggestion
    while (in->good() && std::getline(*in, line)) {
```

PragmaTwice (approved):

torwig (approved):

git-hulk (approved):

git-hulk:
Thanks all, merging...

PR #1401, opened 2023-04-23
Add a particular test case for HGETALL with async_io=true

torwig:
The bug was reproducible with RocksDB v8.0.0, large values in a hash set (>512 bytes). 
After the COMPACT command was executed, HGETALL/HVALS/HKEYS couldn't "see" all the elements of the hash set. However, HGET could find those values. 
It looked like the iterator via the snapshot didn't see them for some reason. 
Since the version of the RocksDB was bumped to v8.0.1, the bug disappeared.

Adding the test case to prevent facing this bug in the future.

This is the output of the failed test:

```
--- FAIL: TestHashWithAsyncIOEnabled (6.10s)
    --- FAIL: TestHashWithAsyncIOEnabled/Test_rare_bug_with_large_value_after_compaction (5.03s)
        hash_test.go:1587: 
            	Error Trace:	/home/yaroslav/programming/kvrocks/tests/gocase/unit/type/hash/hash_test.go:1587
            	Error:      	Not equal: 
            	            	expected: 50
            	            	actual  : 2
            	Test:       	TestHashWithAsyncIOEnabled/Test_rare_bug_with_large_value_after_compaction
FAIL
exit status 1
```

### DIFF
--- a/tests/gocase/unit/type/hash/hash_test.go
+++ b/tests/gocase/unit/type/hash/hash_test.go
@@ -23,6 +23,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math/rand"
 	"sort"
 	"strconv"
 	"strings"
@@ -340,12 +341,15 @@ func TestHash(t *testing.T) {
 	})
 
 	t.Run("HVALS - field with empty string as a value", func(t *testing.T) {
-		require.NoError(t, rdb.HSet(ctx, "test-hash-1", "field1", "some-value").Err())
-		require.NoError(t, rdb.HSet(ctx, "test-hash-1", "field2", "").Err())
+		testKey := "test-hash-1"
+		require.NoError(t, rdb.Del(ctx, testKey).Err())
 
-		require.Equal(t, []string{"some-value", ""}, rdb.HVals(ctx, "test-hash-1").Val())
+		require.NoError(t, rdb.HSet(ctx, testKey, "field1", "some-value").Err())
+		require.NoError(t, rdb.HSet(ctx, testKey, "field2", "").Err())
 
-		require.NoError(t, rdb.Del(ctx, "test-hash-1").Err())
+		require.Equal(t, []string{"some-value", ""}, rdb.HVals(ctx, testKey).Val())
+
+		require.NoError(t, rdb.Del(ctx, testKey).Err())
 	})
 
 	t.Run("HGETALL - small hash}", func(t *testing.T) {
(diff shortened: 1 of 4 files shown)


aleksraiden:
@torwig thanks for help in discovering this bug

git-hulk (dismissed):
LGTM

tisonkun (dismissed):
Thank you!

tests/gocase/unit/type/hash/hash_test.go
```diff
@@ -762,12 +769,54 @@ func TestHash(t *testing.T) {
...
+		dd := make([]byte, 5000)
+		for i := 1; i <= 50; i++ {
+			for j := range dd {
+				dd[j] = byte(src.Int63())
+			}
+			key := util.RandString(10, 20, util.Alpha)
+			require.NoError(t, rdb.HSet(ctx, testKey, key, string(dd)).Err())
+		}
+
+		require.Equal(t, int64(50), rdb.HLen(ctx, testKey).Val())
```
tisonkun: Perhaps `EqualValues` to reduce explicit type conversion.

tests/gocase/unit/type/hash/hash_test.go
```diff
@@ -762,12 +769,54 @@ func TestHash(t *testing.T) {
...
+		require.Equal(t, 50, len(rdb.HGetAll(ctx, testKey).Val()))
```
tisonkun: Perhaps `require.Len` to reduce explicit `len(...)`

tests/gocase/unit/type/hash/hash_test.go
```diff
@@ -762,12 +769,54 @@ func TestHash(t *testing.T) {
...
+		require.Equal(t, 50, len(rdb.HKeys(ctx, testKey).Val()))
+		require.Equal(t, 50, len(rdb.HVals(ctx, testKey).Val()))
+
+		require.NoError(t, rdb.Do(ctx, "COMPACT").Err())
+
+		time.Sleep(5 * time.Second)
```
tisonkun, line 815: Is it possible to be unstable?

torwig: What do you mean by `unstable`? Since the `COMPACT` command is asynchronous, I add the `time.Sleep` call here.

tests/gocase/unit/type/hash/hash_test.go
torwig: Done

tests/gocase/unit/type/hash/hash_test.go
torwig: Done. Thank you for your suggestions.

torwig:
Ooops, not the correct branch. I'll re-push now.

torwig:
Now it's OK.

tests/gocase/unit/type/hash/hash_test.go
tisonkun: Thanks for your explanation :)

Generally, sleep is unpredictable, so it may sleep too short and miss the correct condition. I asked here for more background. If we cannot have a determinate latch for COMPACT completion, I'm OK with current way.

tisonkun (approved):

git-hulk (approved):

git-hulk:
Thanks all, merging...

PR #1405, opened 2023-04-27
Add the ability to use the '*-123' pattern to specify stream entry ID via XADD

torwig:
Currently, Redis Stream has 3 strategies to form the new entry's ID (provided with the XADD command):
- autogenerated: the `*` pattern 
- fully specified by the client: the `12345-6789` pattern
- only the sequence number is auto-generated, and the timestamp is specified by the client: the `12345-*` pattern.

This PR introduces an additional pattern `*-12345` when a client is able to specify a sequence number and a timestamp will be generated by the Kvrocks server.
The new pattern allows clients to use their internal IDs as part of an entry ID. 
Kvrocks generates timestamps and order incoming entries by the receiving time.


aleksraiden:
Awesome feature, we are waiting for this, lot of thanks! For example, we can use this streamID for checking latency for each message (if consumer after processing, store an result with a same ID, but new autogenerated part of time).

aleksraiden:
And some of futures - this feature will be pretty use with XMOVE/XCOPY next command planned too... ;)

src/types/redis_stream_base.cc
```diff
@@ -85,39 +79,48 @@ Status ParseStreamEntryID(const std::string &input, StreamEntryID *id) {
   return Status::OK();
 }
 
-Status ParseNewStreamEntryID(const std::string &input, NewStreamEntryID *id) {
+std::unique_ptr<NextStreamEntryIDGenerationStrategy> ParseNextStreamEntryIDStrategy(const std::string &input) {
```
PragmaTwice: From my side, I recommend you to use something like `StatusOr<std::unique_ptr<NextStreamEntryIDGenerationStrategy>>` instead of exceptions, e.g.
```c++
StatusOr<std::unique_ptr<NextStreamEntryIDGenerationStrategy>> ParseNextStreamEntryIDStrategy(const std::string &input) {
  ...
  if (...) {
    return {Status::NotOk, "some reason..."};
  } else {
    return std::make_unique<SomeDerivedClass>(...);
  }
  ...
}
```
And use it as usual:
```c++
auto res = ParseNextStreamEntryIDStrategy(...);
if (res.IsOK()) {
  (*res)->AnyMethod(); // or `std::unique_ptr x = std::move(*res);`, whatever
} else {
  // print or forward `res.Msg()`, or `return res` directly
}
```

PragmaTwice (commented):
LGTM in general.

src/types/redis_stream_base.cc
torwig: Done

PragmaTwice (approved):
LGTM. Awesome feature!

git-hulk (approved):

PragmaTwice:
Merging...

PR #1406, opened 2023-04-28
Allow default username and fix case sensitive check in HELLO

### LINKED ISSUE #1403: Syntax of HELLO is different from Redis
### Search before asking

### Version

2.3.0

### Minimal reproduce step

1. Set `requirepass mypass`
2. Send command: `HELLO 2 AUTH default mypass` after connected.

mypass should be accepted.

### What did you see instead?

`invalid password` error.

### Anything Else?

Hi, 

According to https://redis.io/commands/hello, the syntax of `HELLO` should be 
```
HELLO [protover [AUTH username password] [SETNAME clientname]]
```
and the default username should be `default`.

However, kvrocks' syntax seems to lack the username part:
https://github.com/apache/incubator-kvrocks/blob/6889b5916ab560ecd5eed79bb9786e36342f0e17/src/commands/cmd_server.cc#L660

Not sure if this is intended? But I think it will be great if matching redis' syntax.

rueian:
Hi, following the conversation https://github.com/apache/incubator-kvrocks/issues/1403

* This closes #1403.

This PR makes the `HELLO` command:
1.  accept the `default` username as well.
2. use case insensitive check on the `AUTH` and `SETNAME`. Clients may send these options in lower or upper cases. For example, go-redis sends them in lower cases.

I think this PR won't break any existing Kvrocks clients but will remove barriers for other clients to adapt Kvrocks by aligning HELLO syntax with Redis.

Please let me know if these changes make sense.


PragmaTwice (dismissed):
LGTM, thanks for your contribution!

torwig (dismissed):

src/commands/cmd_server.cc
```diff
@@ -657,7 +657,7 @@ class CommandTime : public Commander {
   }
 };
 
-/* HELLO [<protocol-version> [AUTH <password>] [SETNAME <name>] ] */
+/* HELLO [<protocol-version> [AUTH [<password>|<username> <password>]] [SETNAME <name>] ] */
```
tisonkun: Perhaps add a comment that `username` should be always `default` otherwise it fails. And such syntax is for compatible only.

tisonkun (dismissed):
Thank you! One comment to convey the background discussion into code.

src/commands/cmd_server.cc
rueian: Sure!

git-hulk (approved):
LGTM

tisonkun (approved):

PragmaTwice:
Merging. Thanks for your contribution.

mapleFU:
Seems that only "hello" support authenticate with user, should  https://redis.io/commands/auth/ support that?

PragmaTwice:
> Seems that only "hello" support authenticate with user, should https://redis.io/commands/auth/ support that?

I think it is not urgent as HELLO, since HELLO in redis just does not allow the username to be omitted, but `AUTH password` is valid in redis.

mapleFU:
Ok, reasonable. LGTM!

PR #1408, opened 2023-04-29
Rename internal macro ENABLE_NEW_ENCODING to METADATA_ENCODING_VERSION

PragmaTwice:
As described in [the document](https://kvrocks.apache.org/community/data-structure-on-rocksdb), `KVROCKS_ENCODING_VERSION` seems to be a more scalable name than `ENABLE_NEW_ENCODING`, since the NEW encoding can become an OLD encoding, but the encoding version can be a definite number.

BTW, I move two test file from `cppunit/types` to `cppunit/`, since they are both not related to types.


tisonkun (dismissed):
LGTM. Thanks!

mapleFU:
Seems ok, RocksDB also use way like that https://rocksdb.org/blog/2019/03/08/format-version-4.html
Should we mark extract for "metadata encoding version"? Or just "encoding version" is ok?

PragmaTwice:
> Seems ok, RocksDB also use way like that https://rocksdb.org/blog/2019/03/08/format-version-4.html Should we mark extract for "metadata encoding version"? Or just "encoding version" is ok?

Done.

mapleFU (approved):

torwig (approved):

PR #1414, opened 2023-05-01
support slot batch for CLUSTERX SETSLOT

### LINKED ISSUE #1412: Support slot batch operation for CLUSTERX SETSLOT (REVIVED)
### Search before asking

### Motivation

Since it seems important to support slot batch operation for CLUSTERX SETSLOT, we want to revive this abandoned PR #529.

From #529:
> Support for #510 
>
> It is an enhancement to the original implement of command `CLUSTERX SETSLOT`  #463 
>
> **Command formart**
>
> `CLUSTERX SETSLOT "$slot_batch" NODE $node_id $new_version`
>
> assign batch slots to the node if new version is current version+1. 
> It is compatible with the original command syntax.
>
> **Parameter Changes**
>
> > $slot_batch: batch slots which we want to assign. It can be given by single slotid or slot batch.  Slot batch can be given by continuous slot range or discrete slot list with spaces as separators, e.g. "3 5 9 10-20".
>
> **Example**
> `CLUSTERX SETSLOT 1 NODE $nodeid1 1`
> `CLUSTERX SETSLOT "3 5 9 20-30" NODE $nodeid1 2`

### Solution

See #529

infdahai:
closes #1412. refer to #529

1. add stdint.h for uint_t type
 ref(https://godbolt.org/z/5q6zj5W9s)
2. add cluster_setslot_batch feature.
3. tests for parsedSlotRange(cpp unit and go test)
4. move parse func to commandX

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -26,6 +26,7 @@
 #include <fstream>
 #include <memory>
 
+#include "cluster/cluster_defs.h"
 #include "commands/commander.h"
 #include "common/io_util.h"
 #include "fmt/format.h"
@@ -34,16 +35,6 @@
 #include "server/server.h"
 #include "string_util.h"
 #include "time_util.h"
-
-const char *errInvalidNodeID = "Invalid cluster node id";
-const char *errInvalidSlotID = "Invalid slot id";
-const char *errSlotOutOfRange = "Slot is out of range";
-const char *errInvalidClusterVersion = "Invalid cluster version";
-const char *errSlotOverlapped = "Slot distribution is overlapped";
-const char *errNoMasterNode = "The node isn't a master";
-const char *errClusterNoInitialized = "CLUSTERDOWN The cluster is not initialized";
-const char *errInvalidClusterNodeInfo = "Invalid cluster nodes info";
-const char *errInvalidImportState = "Invalid import state";
 
 ClusterNode::ClusterNode(std::string id, std::string host, int port, int role, std::string master_id,
                          std::bitset<kClusterSlots> slots)
@@ -93,10 +84,6 @@ Status Cluster::SetNodeId(const std::string &node_id) {
   return Status::OK();
 }
 
-// Set the slot to the node if new version is current version +1. It is useful
-// when we scale cluster avoid too many big messages, since we only update one
-// slot distribution and there are 16384 slot in our design.
-//
 // The reason why the new version MUST be +1 of current version is that,
 // the command changes topology based on specific topology (also means specific
 // version), we must guarantee current topology is exactly expected, otherwise,
@@ -104,21 +91,17 @@ Status Cluster::SetNodeId(const std::string &node_id) {
 // This is different with CLUSTERX SETNODES commands because it uses new version
 // topology to cover current version, it allows kvrocks nodes lost some topology
 // updates since of network failure, it is state instead of operation.
-Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_version) {
-  // Parameters check
+Status Cluster::SetSlotRanges(const std::vector<SlotRange> &slot_ranges, const std::string &node_id,
+                              int64_t new_version) {
   if (new_version <= 0 || new_version != version_ + 1) {
     return {Status::NotOK, errInvalidClusterVersion};
-  }
-
-  if (!IsValidSlot(slot)) {
-    return {Status::NotOK, errInvalidSlotID};
   }
 
   if (node_id.size() != kClusterNodeIdLen) {
     return {Status::NotOK, errInvalidNodeID};
   }
 
-  // Get the node which we want to assign a slot into it
+  // Get the node which we want to assign slots into it
   std::shared_ptr<ClusterNode> to_assign_node = nodes_[node_id];
   if (to_assign_node == nullptr) {
     return {Status::NotOK, "No this node in the cluster"};
@@ -135,23 +118,29 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
   //  1. Remove the slot from old node if existing
   //  2. Add the slot into to-assign node
   //  3. Update the map of slots to nodes.
-  std::shared_ptr<ClusterNode> old_node = slots_nodes_[slot];
-  if (old_node != nullptr) {
-    old_node->slots[slot] = false;
-  }
-  to_assign_node->slots[slot] = true;
-  slots_nodes_[slot] = to_assign_node;
+  // remember: The atomicity of the process is based on
+  // the transactionality of ClearKeysOfSlot().
+  for (auto [s_start, s_end] : slot_ranges) {
+    for (int slot = s_start; slot <= s_end; slot++) {
+      std::shared_ptr<ClusterNode> old_node = slots_nodes_[slot];
+      if (old_node != nullptr) {
+        old_node->slots[slot] = false;
+      }
+      to_assign_node->slots[slot] = true;
+      slots_nodes_[slot] = to_assign_node;
 
-  // Clear data of migrated slot or record of imported slot
-  if (old_node == myself_ && old_node != to_assign_node) {
-    // If slot is migrated from this node
-    if (migrated_slots_.count(slot) > 0) {
-      svr_->slot_migrator->ClearKeysOfSlot(kDefaultNamespace, slot);
-      migrated_slots_.erase(slot);
-    }
-    // If slot is imported into this node
-    if (imported_slots_.count(slot) > 0) {
-      imported_slots_.erase(slot);
+      // Clear data of migrated slot or record of imported slot
+      if (old_node == myself_ && old_node != to_assign_node) {
+        // If slot is migrated from this node
+        if (migrated_slots_.count(slot) > 0) {
+          svr_->slot_migrator->ClearKeysOfSlot(kDefaultNamespace, slot);
+          migrated_slots_.erase(slot);
+        }
+        // If slot is imported into this node
+        if (imported_slots_.count(slot) > 0) {
+          imported_slots_.erase(slot);
+        }
+      }
     }
   }
 
--- a/src/cluster/cluster_defs.h
+++ b/src/cluster/cluster_defs.h
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#pragma once
+
+#include "cluster/redis_slot.h"
+
+enum {
+  kClusterMaster = 1,
+  kClusterSlave = 2,
+  kClusterNodeIdLen = 40,
+  kClusterPortIncr = 10000,
+  kClusterSlots = HASH_SLOTS_SIZE,
+};
+
+inline constexpr const char *errInvalidNodeID = "Invalid cluster node id";
+inline constexpr const char *errInvalidSlotID = "Invalid slot id";
+inline constexpr const char *errSlotOutOfRange = "Slot is out of range";
+inline constexpr const char *errInvalidClusterVersion = "Invalid cluster version";
+inline constexpr const char *errSlotOverlapped = "Slot distribution is overlapped";
+inline constexpr const char *errNoMasterNode = "The node isn't a master";
+inline constexpr const char *errClusterNoInitialized = "CLUSTERDOWN The cluster is not initialized";
+inline constexpr const char *errInvalidClusterNodeInfo = "Invalid cluster nodes info";
+inline constexpr const char *errInvalidImportState = "Invalid import state";
+
+using SlotRange = std::pair<int, int>;
--- a/src/cluster/redis_slot.h
+++ b/src/cluster/redis_slot.h
@@ -19,6 +19,7 @@
  */
 
 #pragma once
+#include <cstdint>
 #include <string>
 
 // crc16
--- a/src/commands/cmd_cluster.cc
+++ b/src/commands/cmd_cluster.cc
@@ -18,6 +18,7 @@
  *
  */
 
+#include "cluster/cluster_defs.h"
 #include "cluster/slot_import.h"
 #include "commander.h"
 #include "error_constants.h"
@@ -156,15 +157,9 @@ class CommandClusterX : public Commander {
 
     // CLUSTERX SETSLOT $SLOT_ID NODE $NODE_ID $VERSION
     if (subcommand_ == "setslot" && args_.size() == 6) {
-      auto parse_id = ParseInt<int>(args[2], 10);
-      if (!parse_id) {
-        return {Status::RedisParseErr, errValueNotInteger};
-      }
-
-      slot_id_ = *parse_id;
-
-      if (!Cluster::IsValidSlot(slot_id_)) {
-        return {Status::RedisParseErr, "Invalid slot id"};
+      Status s = CommanderHelper::ParseSlotRanges(args_[2], slot_ranges_);
+      if (!s.IsOK()) {
+        return s;
       }
 
       if (strcasecmp(args_[3].c_str(), "node") != 0) {
@@ -219,7 +214,7 @@ class CommandClusterX : public Commander {
         *output = redis::Error(s.Msg());
       }
     } else if (subcommand_ == "setslot") {
-      Status s = svr->cluster->SetSlot(slot_id_, args_[4], set_version_);
+      Status s = svr->cluster->SetSlotRanges(slot_ranges_, args_[4], set_version_);
       if (s.IsOK()) {
         need_persist_nodes_info = true;
         *output = redis::SimpleString("OK");
@@ -251,7 +246,7 @@ class CommandClusterX : public Commander {
   std::string dst_node_id_;
   int64_t set_version_ = 0;
   int64_t slot_ = -1;
-  int slot_id_ = -1;
+  std::vector<SlotRange> slot_ranges_;
   bool force_ = false;
 };
 
--- a/src/common/encoding.h
+++ b/src/common/encoding.h
@@ -23,6 +23,7 @@
 #include <rocksdb/slice.h>
 #include <unistd.h>
 
+#include <cstdint>
 #include <string>
 
 enum class Endian {
--- a/src/cluster/cluster.h
+++ b/src/cluster/cluster.h
@@ -29,19 +29,12 @@
 #include <unordered_map>
 #include <vector>
 
+#include "cluster/cluster_defs.h"
 #include "commands/commander.h"
 #include "common/io_util.h"
 #include "redis_slot.h"
 #include "server/redis_connection.h"
 #include "status.h"
-
-enum {
-  kClusterMaster = 1,
-  kClusterSlave = 2,
-  kClusterNodeIdLen = 40,
-  kClusterPortIncr = 10000,
-  kClusterSlots = HASH_SLOTS_SIZE,
-};
 
 class ClusterNode {
  public:
@@ -79,7 +72,7 @@ class Cluster {
   Status SetClusterNodes(const std::string &nodes_str, int64_t version, bool force);
   Status GetClusterNodes(std::string *nodes_str);
   Status SetNodeId(const std::string &node_id);
-  Status SetSlot(int slot, const std::string &node_id, int64_t version);
+  Status SetSlotRanges(const std::vector<SlotRange> &slot_ranges, const std::string &node_id, int64_t version);
   Status SetSlotMigrated(int slot, const std::string &ip_port);
   Status SetSlotImported(int slot);
   Status GetSlotsInfo(std::vector<SlotInfo> *slot_infos);
--- a/src/commands/commander.cc
+++ b/src/commands/commander.cc
@@ -20,6 +20,8 @@
 
 #include "commander.h"
 
+#include "cluster/cluster_defs.h"
+
 namespace redis {
 
 RegisterToCommandTable::RegisterToCommandTable(std::initializer_list<CommandAttributes> list) {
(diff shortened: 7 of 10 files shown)


src/cluster/cluster.cc
```diff
@@ -93,32 +93,27 @@ Status Cluster::SetNodeId(const std::string &node_id) {
   return Status::OK();
 }
 
-// Set the slot to the node if new version is current version +1. It is useful
-// when we scale cluster avoid too many big messages, since we only update one
-// slot distribution and there are 16384 slot in our design.
-//
+// Set the slots to the node if new version is current version +1. It is useful
+// when we scale cluster avoid too many big messages. By th way,there are 16384 slots
```
torwig: Typo in the phrase "By the way". You can omit it entirely because it doesn't add value to the sentence or explanation.

src/cluster/cluster.cc
```diff
@@ -128,30 +123,43 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
     return {Status::NotOK, errNoMasterNode};
   }
 
+  std::vector<std::pair<int, int>> slots;
+  Status s = parseSlotRange(slots_str, slots);
```
torwig: Is there always just one range of slots, according to the name of the function?

src/cluster/cluster.cc
```diff
@@ -641,6 +649,54 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
   return SetClusterNodes(nodes_info, version, false);
 }
 
+// TODO: maybe it needs to use a more precise error type to represent `NotOk`.
+Status Cluster::parseSlotRange(const std::string &slots_str, std::vector<std::pair<int, int>> &slots) {
+  if (slots_str.empty()) {
+    return {Status::NotOK, "Don't use empty slots."};
```
torwig: Here you can simply say "No slots to parse" or something like that.

src/cluster/cluster.cc
```diff
@@ -641,6 +649,54 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+      }
+    }
+    return true;
+#endif
+  };
+
+  // Parse all slots(include slot ranges)
+  for (const auto &slot_range : slot_ranges) {
+    if (is_number(slot_range)) {
+      int s_start = stoi(slot_range);
```
torwig: Here you can use the `ParseInt` function from `parse_util.h`.

src/cluster/cluster.cc
```diff
@@ -641,6 +649,54 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+    }
+
+    // parse slot range: "int1-int2" (satisfy: int1 <= int2 )
+    assert(slot_range.back() != '-');
+    std::vector<std::string> fields = util::Split(slot_range, "-");
+    assert(fields.size() == 2);
+    if (is_number(fields[0]) && is_number(fields[1])) {
+      int s_start = stoi(fields[0]), s_end = stoi(fields[1]);
+      assert((s_start <= s_end) && IsValidSlot(s_end));
+      slots.emplace_back(std::make_pair(s_start, s_end));
```
torwig: Did you forget to insert `continue` after `emplace_back`?

src/cluster/cluster.cc
```diff
@@ -641,6 +649,54 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+        return false;
+      }
+    }
+    return true;
+#endif
+  };
+
+  // Parse all slots(include slot ranges)
+  for (const auto &slot_range : slot_ranges) {
+    if (is_number(slot_range)) {
```
torwig: Personally, here I would split the input with `-` and then decide if it's a range or just a single slot number by checking the result.
Because right now you are checking all the characters, then splitting the string by `-` and checking all the characters of each substring.

src/cluster/cluster.cc
```diff
@@ -641,6 +649,54 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+  if (slots_str.empty()) {
+    return {Status::NotOK, "Don't use empty slots."};
+  }
+  std::vector<std::string> slot_ranges = util::Split(slots_str, " ");
+
+  if (slot_ranges.empty()) {
+    return {Status::NotOK, fmt::format("Invalid slots: {}. Please use ' ' to space slots.", slots_str)};
+  }
+
+  auto is_number = [&](const std::string &s) {
```
mapleFU: Why not use `src/common/parse_util.h`?

src/cluster/cluster.cc
```diff
@@ -641,6 +649,54 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+      int s_start = stoi(slot_range);
+      assert(IsValidSlot(s_start));
+      slots.emplace_back(std::make_pair(s_start, s_start));
+      continue;
+    }
+
+    // parse slot range: "int1-int2" (satisfy: int1 <= int2 )
+    assert(slot_range.back() != '-');
+    std::vector<std::string> fields = util::Split(slot_range, "-");
+    assert(fields.size() == 2);
```
torwig: Correct me if I'm wrong. `assert` works only in DEBUG build. Thus, in release build, if there is no `-` inside the string, the next line will try to access `fields[`]` and the service will crash. 
In my opinion, such cases should be handled explicitly and return Status::NotOK instead of excessively using `assert`.
However, I would like to hear @PragmaTwice on that as an expert.

src/cluster/cluster.cc
```diff
@@ -641,6 +649,54 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+  for (const auto &slot_range : slot_ranges) {
+    if (is_number(slot_range)) {
+      int s_start = stoi(slot_range);
+      assert(IsValidSlot(s_start));
+      slots.emplace_back(std::make_pair(s_start, s_start));
+      continue;
+    }
+
+    // parse slot range: "int1-int2" (satisfy: int1 <= int2 )
+    assert(slot_range.back() != '-');
```
mapleFU: If user pass a ambigious string, would this causing fatal?

src/cluster/cluster.cc
mapleFU: ditto

src/commands/cmd_cluster.cc
```diff
@@ -219,7 +210,7 @@ class CommandClusterX : public Commander {
         *output = redis::Error(s.Msg());
       }
     } else if (subcommand_ == "setslot") {
-      Status s = svr->cluster->SetSlot(slot_id_, args_[4], set_version_);
+      Status s = svr->cluster->SetSlot(slots_str_, args_[4], set_version_);
```
torwig: Maybe, it would be better to parse slot ranges inside `CommandClusterX::Parse` and then pass already parsed (and valid) numbers to the `SetSlot` function.

src/cluster/cluster.cc
infdahai: you're right. I will remove `assert`.

src/commands/cmd_cluster.cc
infdahai: I'm not sure to do this. `CommandClusterX::Parse` is only used to find some keywords and do the simple check now.
If necessary, I can migrate `parse` functions in `cluster.h` to `CommandClusterX`.

src/cluster/cluster.cc
infdahai: yeah, I change the logic.

src/common/encoding.h
```diff
@@ -23,6 +23,7 @@
 #include <rocksdb/slice.h>
 #include <unistd.h>
 
+#include <cstdint>
```
PragmaTwice, line 26:
```suggestion
```

CMakeLists.txt
```diff
@@ -164,6 +166,12 @@ if ((PROJECT_VERSION STREQUAL "unstable") AND (GIT_SHA STREQUAL ""))
 endif ()
 configure_file(src/version.h.in ${PROJECT_BINARY_DIR}/version.h)
 
+if (NOT CMAKE_BUILD_TYPE OR CMAKE_BUILD_TYPE STREQUAL "None")
+    set (CMAKE_BUILD_TYPE "RelWithDebInfo")
+    message (STATUS "CMAKE_BUILD_TYPE is not set, set to default = ${CMAKE_BUILD_TYPE}")
+endif ()
+message (STATUS "CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
```
PragmaTwice: I think change the value of `CMAKE_BUILD_TYPE` after the `project()` command (and even after all dependency fetching and configuration here) can be a totally mess (the dependency and many target can be mis-configured).

CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ option(PORTABLE "build a portable binary (disable arch-specific optimizations)"
 # TODO: set ENABLE_NEW_ENCODING to ON when we are ready
 option(ENABLE_NEW_ENCODING "enable new encoding (#1033) for storing 64bit size and expire time in milliseconds" OFF)
 
+set(CMAKE_LINK_DEPENDS_NO_SHARED 1)
```
```suggestion
set(CMAKE_LINK_DEPENDS_NO_SHARED ON)
```

And, could you share some reason for this change?

x.py
```diff
@@ -106,7 +106,7 @@ def build(dir: str, jobs: Optional[int], ghproxy: bool, ninja: bool, unittest: b
 
     makedirs(dir, exist_ok=True)
 
-    cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo"]
+    cmake_options = []
```
PragmaTwice: related to the change to CMakeLists.txt

src/cluster/redis_slot.h
```diff
@@ -19,6 +19,7 @@
  */
 
 #pragma once
+#include <cstdint>
```
PragmaTwice, line 22:
```suggestion
```

src/cluster/cluster.cc
```diff
@@ -641,6 +648,56 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+    return {Status::NotOK, "No slots to parse."};
+  }
+  std::vector<std::string> slot_ranges = util::Split(slots_str, " ");
+
+  if (slot_ranges.empty()) {
+    return {Status::NotOK, fmt::format("Invalid slots: {}. Please use ' ' to space slots.", slots_str)};
+  }
+
+  auto valid_range = NumericRange<int>{0, kClusterSlots - 1};
+  // Parse all slots(include slot ranges)
```
```suggestion
  // Parse all slots (include slot ranges)
```

src/cluster/cluster.cc
```diff
@@ -641,6 +648,56 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+    auto parse_end = ParseInt<int>(fields[1], valid_range, 10);
+    if (!parse_start || !parse_end || *parse_start > *parse_end) {
+      return {Status::NotOK,
+              fmt::format(
+                  "Invalid slot range: {}. The slot range `int1-int2` needs to satisfy the condition (int1 <= int2).",
+                  slot_range)};
+    }
+    int slot_start = *parse_start;
+    int slot_end = *parse_end;
+    slots.emplace_back(std::make_pair(slot_start, slot_end));
```
```suggestion
    slots.emplace_back(*parse_start, *parse_end);
```

src/cluster/cluster.cc
```diff
@@ -641,6 +648,56 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+      int slot_start = *parse_result;
+      slots.emplace_back(std::make_pair(slot_start, slot_start));
+      continue;
+    }
+
+    // parse slot range: "int1-int2" (satisfy: int1 <= int2 )
+    if (slot_range.back() == '-') {
+      return {Status::NotOK,
+              fmt::format("Invalid slot range: {}. The character '-' can't appear in the last position.", slot_range)};
+    }
```
PragmaTwice: It seems to be unnecessary since the code below (split according to `-` and expect `size() == 2`) already did this check.

src/cluster/cluster.cc
```diff
@@ -128,30 +122,43 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
...
-      svr_->slot_migrator->ClearKeysOfSlot(kDefaultNamespace, slot);
-      migrated_slots_.erase(slot);
-    }
-    // If slot is imported into this node
-    if (imported_slots_.count(slot) > 0) {
-      imported_slots_.erase(slot);
+  // remember: The atomicity of the process is based on
+  // the transactionality of ClearKeysOfSlot().
+  for (const auto &slot_range : slots) {
+    int s_start = slot_range.first, s_end = slot_range.second;
```
```suggestion
  for (auto [s_start, s_end] : slots) {
```

PragmaTwice (commented):
Thanks for your contribution!

You can run `./x.py format` to format the code and let the CI proceed.

CMakeLists.txt
infdahai: just avoid relinking all depended targets on .so. It's not necessary now, so I delete it.

src/cluster/redis_slot.h
infdahai: Please see the result in https://godbolt.org/z/5q6zj5W9s . So I think the header is due.

src/cluster/cluster.cc
infdahai: in util/Split(),this process is below.
```cpp
  size_t begin = 0, end = in.find_first_of(delim);
  do {
    std::string elem = in.substr(begin, end - begin);
    if (!elem.empty()) out.push_back(std::move(elem));
    if (end == std::string::npos) break;
    begin = end + 1;
    end = in.find_first_of(delim, begin);
  } while (true);
```
when I use the range such as `3-4-`, I also get a valid result. Because the last `-` just ignores and breaks in this func.

CMakeLists.txt
infdahai: yes, you are right. I just want to make `debug`,so shall I do this check in `x.py build`?

infdahai:
> 
> You can run `./x.py format` to format the code and let the CI proceed.

If I exec `./x.py format`, I get a irrelevant result.(Not important though, just ignored).
image

CMakeLists.txt
PragmaTwice: IIRC `x.py` already has a default build type.

PragmaTwice:
> If I exec `./x.py format`, I get a irrelevant result.(Not important though, just ignored). image

Could you provide the version of clang-format you are using?

Anyway, I think you can firstly manually ignore this unrelated change as a workaround.

PragmaTwice:
Sorry for the wrong close-button click. :rofl:

src/cluster/cluster.cc
PragmaTwice: Thanks. Got your point.

infdahai:
> > If I exec `./x.py format`, I get a irrelevant result.(Not important though, just ignored). image
> 
> Could you provide the version of clang-format you are using?
> 
> Anyway, I think you can firstly manually ignore this unrelated change as a workaround.

yes, I will propose a new pr for this `clang-tidy`.

src/cluster/redis_slot.h
PragmaTwice: Got it. Thanks for your demo.

PragmaTwice:
CI seems still blocked in format. Not sure your clang-format version, but you can try clang-format 12, which is used in CI.

https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.1

And I can confirm that clang-format 14 on my local has also no issue.

```
/home/runner/work/incubator-kvrocks/incubator-kvrocks/src/cluster/cluster.cc:474:39: error: code should be clang-formatted [-Wclang-format-violations]
  for (const auto &id : n->replicas) {      // replicas
                                      ^
```

infdahai:
> CI seems still blocked in format. Not sure your clang-format version, but you can try clang-format 12, which is used in CI.
> 
> https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.1
> 
> And I can confirm that clang-format 14 on my local has also no issue.
> 
> ```
> /home/runner/work/incubator-kvrocks/incubator-kvrocks/src/cluster/cluster.cc:474:39: error: code should be clang-formatted [-Wclang-format-violations]
>   for (const auto &id : n->replicas) {      // replicas
>                                       ^
> ```

Sorry for this problem. My llvm toolchain version is not unified and the `clang-format` version is 16.

You're too kind. Ignore this problem.

CMakeLists.txt
infdahai: OK, I will remove this part.

src/cluster/cluster.cc
```diff
@@ -93,32 +93,26 @@ Status Cluster::SetNodeId(const std::string &node_id) {
...
 // the command changes topology based on specific topology (also means specific
 // version), we must guarantee current topology is exactly expected, otherwise,
 // this update may make topology corrupt, so base topology version is very important.
 // This is different with CLUSTERX SETNODES commands because it uses new version
 // topology to cover current version, it allows kvrocks nodes lost some topology
 // updates since of network failure, it is state instead of operation.
-Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_version) {
-  // Parameters check
+Status Cluster::SetSlot(const std::string &slots_str, const std::string &node_id, int64_t new_version) {
+  // the parsing process takes some time, so let's check other arguments first.
```
git-hulk: Can remove meaningless comment line
```suggestion
```

src/cluster/cluster.cc
```diff
@@ -641,6 +647,53 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
   return SetClusterNodes(nodes_info, version, false);
 }
 
+// TODO: maybe it needs to use a more precise error type to represent `NotOk`.
+Status Cluster::parseSlotRanges(const std::string &slots_str, std::vector<std::pair<int, int>> &slots) {
```
git-hulk: Could you add some test cases for parseSlotRanges?

src/cluster/cluster.cc
```diff
@@ -641,6 +647,53 @@ Status Cluster::LoadClusterNodes(const std::string &file_path) {
...
+  if (slots_str.empty()) {
+    return {Status::NotOK, "No slots to parse."};
+  }
+  std::vector<std::string> slot_ranges = util::Split(slots_str, " ");
+
+  if (slot_ranges.empty()) {
+    return {Status::NotOK, fmt::format("Invalid slots: {}. Please use ' ' to space slots.", slots_str)};
```
```suggestion
    return {Status::NotOK, fmt::format("Invalid slots: {}. Please use spaces to separate slots.", slots_str)};
```

PragmaTwice (commented):
LGTM generally, it would be better if you add several golang tests to tests/gocase/integration/cluster/cluster_test.go .

src/commands/cmd_cluster.cc
caipengbo: I also think it would be better to do this in command parsing. Only valid slotranges is concerned in `cluster.cc` and these parsing aspects are not concerned.

src/cluster/cluster.cc
```diff
@@ -93,32 +93,26 @@ Status Cluster::SetNodeId(const std::string &node_id) {
...
 // The reason why the new version MUST be +1 of current version is that,
 // the command changes topology based on specific topology (also means specific
 // version), we must guarantee current topology is exactly expected, otherwise,
 // this update may make topology corrupt, so base topology version is very important.
 // This is different with CLUSTERX SETNODES commands because it uses new version
 // topology to cover current version, it allows kvrocks nodes lost some topology
 // updates since of network failure, it is state instead of operation.
-Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_version) {
-  // Parameters check
+Status Cluster::SetSlot(const std::string &slots_str, const std::string &node_id, int64_t new_version) {
```
caipengbo: Just pass in the valid slotrange and hand the parsing and validation to the outer layer. You can rename this function to `SetSlots` or `SetSlotRanges`.

src/cluster/cluster.cc
```diff
@@ -93,32 +83,27 @@ Status Cluster::SetNodeId(const std::string &node_id) {
...
 // The reason why the new version MUST be +1 of current version is that,
 // the command changes topology based on specific topology (also means specific
 // version), we must guarantee current topology is exactly expected, otherwise,
 // this update may make topology corrupt, so base topology version is very important.
 // This is different with CLUSTERX SETNODES commands because it uses new version
 // topology to cover current version, it allows kvrocks nodes lost some topology
 // updates since of network failure, it is state instead of operation.
-Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_version) {
-  // Parameters check
+Status Cluster::SetSlotRanges(const std::vector<std::pair<int, int>> &slot_ranges, const std::string &node_id,
```
caipengbo: Personally, I think it's better to use alias declaration(Like `using SlotRange = std::pair<int, int>` ) or struct, because I see `std::pair<int, int>` appear 6 times.

caipengbo (dismissed):

src/cluster/cluster.cc
infdahai: OK, I use SlotRange instead of std::pair<int,int>.
And it lefts the `start` and `end` integers in SlotInfo struct. Maybe we think about it later. 

```cpp
// TODO: replace integer `start`, `end` with SlotRange
struct SlotInfo {
  int start;
  int end;
  struct NodeInfo {
    std::string host;
    int port;
    std::string id;
  };
  std::vector<NodeInfo> nodes;
};
```

src/cluster/cluster_defs.h
```diff
@@ -0,0 +1,43 @@
...
+inline constexpr const char *errInvalidSlotID = "Invalid slot id";
+inline constexpr const char *errSlotOutOfRange = "Slot is out of range";
+inline constexpr const char *errInvalidClusterVersion = "Invalid cluster version";
+inline constexpr const char *errSlotOverlapped = "Slot distribution is overlapped";
+inline constexpr const char *errNoMasterNode = "The node isn't a master";
+inline constexpr const char *errClusterNoInitialized = "CLUSTERDOWN The cluster is not initialized";
+inline constexpr const char *errInvalidClusterNodeInfo = "Invalid cluster nodes info";
+inline constexpr const char *errInvalidImportState = "Invalid import state";
+
+using SlotRange = std::pair<int, int>;
```
PragmaTwice:
```suggestion
using SlotRange = std::pair<int, int>;

```
better to have a newline in the end of the file.

src/cluster/cluster_defs.h
```diff
@@ -0,0 +1,43 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#pragma once
+
+#include "cluster/redis_slot.h"
+
+enum {
```
infdahai, line 25: I move the enum struct and strings defined in `cluster` class.
It reduces dependencies and we can add this header in another module, such as `commander.h`.
Perhaps we can move many strings like this if we want to use these in other classs.

src/cluster/cluster.cc
```diff
@@ -135,23 +120,29 @@ Status Cluster::SetSlot(int slot, const std::string &node_id, int64_t new_versio
...
+        old_node->slots[slot] = false;
+      }
+      to_assign_node->slots[slot] = true;
+      slots_nodes_[slot] = to_assign_node;
+
+      // Clear data of migrated slot or record of imported slot
+      if (old_node == myself_ && old_node != to_assign_node) {
+        // If slot is migrated from this node
+        if (migrated_slots_.count(slot) > 0) {
+          svr_->slot_migrator->ClearKeysOfSlot(kDefaultNamespace, slot);
```
infdahai, line 136: One `ClearKeyOfSlot` call one writebatch of rocksdb.
So this process exists consistency issues if `slot` in the middle is down.
I don't know what to do with it yet

caipengbo: There is indeed this problem and we can persist it to Propagate CF.

caipengbo: After we merge this PR, you can ask an issue and I will solve this problem, which I have solved in production environment.

src/cluster/cluster_defs.h
infdahai: I add "\n" to the line in local env but it doesn't seem to see it in github. If I add two `\n` with increasing two line nums and then I get the above situation, but `clang-format` will fix this.

for example,
https://github.com/apache/incubator-kvrocks/blob/30e9fd788a4acb585566a008f4d5fff65ee06ce3/src/cluster/slot_import.h#L53-L59

src/cluster/cluster.cc
```diff
@@ -93,32 +84,27 @@ Status Cluster::SetNodeId(const std::string &node_id) {
   return Status::OK();
 }
 
-// Set the slot to the node if new version is current version +1. It is useful
-// when we scale cluster avoid too many big messages, since we only update one
-// slot distribution and there are 16384 slot in our design.
-//
+// Set the slots to the node if new version is current version +1. It is useful
+// when we scale cluster avoid too many big messages. By th way,there are 16384 slots
+// in our design.
```
```suggestion
```

git-hulk:
LGTM, I think we can remove TODO comments, then submit an issue if you feel it's necessary.

infdahai:
> LGTM, I think we can remove TODO comments, then submit an issue if you feel it's necessary.

OK

caipengbo (approved):
LGTM

torwig (approved):
LGTM

git-hulk (approved):

git-hulk (approved):

git-hulk (approved):

git-hulk:
Thanks all, merging...

Issue #1376, opened 2023-04-09
Is it possible to add redis-cli to the docker image so that it can be downloaded and used immediately.


KeeProMise:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

I downloaded docker Mirror, found that you need to install redis-cli yourself,s it possible to add redis-cli to the docker image so that it can be downloaded and used immediately.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

git-hulk:
This sounds good to me. cc @torwig @PragmaTwice @ShooterIT

PragmaTwice:
I do not think it is necessary. The docker image we provide is only for starting a kvrocks service, not a redis client.

You can build a docker image only for redis-cli, find one in the docker hub, or use redis-cli in your host environment.

BTW, we do not encourage anyone to modify the entrypoint of the docker image and use other binaries in the image. It is literally only for one binary, i.e. `bin/kvrocks`.

KeeProMise:
> We don't need to modify the entrypoint, I think redis-cli can be added to the bin directory, so that for some people who are new to the database, it is very convenient (enter the docker container to run bin/redis-cli) to execute some redis commands to interact with the database. Moreover, the same is true for mysql, redis, iotdb, etc. that I know.

PragmaTwice:
When you "enter" the docker container via shell (`docker exec`), you changed the entrypoint of the docker image from `bin/kvrocks` to `bash` (or the posix shell, whatever). 

We do not expect users to do this.

Another thing is that, seems there is no package in ubuntu for a single redis-cli.

SpecLad:
Seems like this was fixed in #1415.

PR #1415, opened 2023-05-02
Add redis cli to nightly image

ColinChamber:
solve #1376

Hi @PragmaTwice 
We have been working on improving kvrocks-operator recently and have found that many redis-operators use redis-cli for Kubernetes liveness probes. Although we can use commands like 'echo "PING" | nc localhost 6666', it can be a bit tricky. Additionally, redis-cli is more user-friendly for operations and makes it convenient to query service information within the container. In complex cloud network environments with audit policies, it is not possible for redis-cli from any node to connect to the redis-server

**I pushed the image to my own repository and ran some tests.**
image
image


Dockerfile
```diff
@@ -37,6 +37,10 @@ WORKDIR /kvrocks
 
 COPY --from=build /kvrocks/build/kvrocks ./bin/
 
+ARG TARGETARCH
+COPY tools/redis-cli-${TARGETARCH} /usr/bin/redis-cli
+RUN chmod a+x /usr/bin/redis-cli
```
PragmaTwice: IMHO if a user run `docker build` towards this dockerfile in him/her device instead of CI, a fatal error will appear since there is no `tools` dir.

So I think It is not a reasonable change.

PragmaTwice: And I want to emphasize my point: the dockerfile cannot have different behavior between built by a user and triggered in the CI.

PragmaTwice: I think these conditional branches related to arch in the current changes is not necessary. You can just build redis-cli in the dockerfile.

ColinChamber: done

Dockerfile
```diff
@@ -37,6 +44,9 @@ WORKDIR /kvrocks
 
 COPY --from=build /kvrocks/build/kvrocks ./bin/
 
+COPY --from=build /kvrocks/tools/redis-cli ./bin/
+RUN ln -s /kvrocks/bin/redis-cli /usr/bin/redis-cli
```
PragmaTwice: Is it necessary to add this symlink?

ColinChamber: /kvrocks/bin is not in the PATH environment variable, so I created a symlink. Are you suggesting that I should either copy it directly to /usr/bin or add /kvrocks/bin to the PATH?

PragmaTwice: I think the latter is better.

PragmaTwice (approved):
LGTM

tisonkun (approved):

PragmaTwice:
Thanks @ColinChamber , merging...

PR #1418, opened 2023-05-04
Support blocking migration for the cluster migrate command

### LINKED ISSUE #1350: Add the support of  the blocking migration for the cluster migrate command
### Search before asking

### Motivation

For now the command to migrate slot `clusterx migrate` works asynchronously. It will return "OK" right after the migration has started. If the client wants to know the result of the migration, it has to constantly use "cluster info" to get the state.

However, once the migration has been completed, the target slot will be forbidden, and the client (cluster manager) has to actively change the topology as soon as possible, in order to make the cluster continue to work.

In summary, the current `clusterx migrate` has two shortcuts: 
1. The client cannot get the result soon enough.
2. The "polling" approach to get the migration result is not efficient enough.

### Solution

I propose to leave the old `clusterx migrate` command intact, and add a new `clusterx bmirate` command, which works like other blocking commands (e.g. `blpop`). The server will block the connection and return "OK" (if succeed) or other error message (if failed) to the client only after the migration has completed.

The command shall be like:
```shell
clusterx bmirate ${SLOT} ${TARGET_NODEID} ${TIMEOUT}
```

ellutionist:
This PR closes [1350](https://github.com/apache/incubator-kvrocks/issues/1350).


ellutionist:
@git-hulk @xiaobiaozhao @PragmaTwice @torwig @ShooterIT

src/commands/cmd_cluster.cc
```diff
@@ -253,6 +278,10 @@ class CommandClusterX : public Commander {
   int64_t slot_ = -1;
   int slot_id_ = -1;
   bool force_ = false;
+
+  bool sync_migrate_ = false;
+  int sync_migrate_timeout_ = 0;
+  std::shared_ptr<SyncMigrateContext> sync_migrate_ctx_ = nullptr;
```
torwig: @ellutionist Great job!
I just thinking if we need to have `sync_migrate_ctx_` inside CommandClusterX because you just create it and pass it to the `MigrateSlot` function without further use of this object.

PragmaTwice: Could you share some reason to use `std::shared_ptr`? Seems except here (not necessary in my opinion) I cannot find where the object is "shared".

ellutionist:
@torwig @PragmaTwice The reason why I use `shared_ptr` of `SyncMigrateContext` as a field of  `CommandClusterX` is that I have to keep the lifetime of `SyncMigrateContext` as long as the command object.

When the `slot_migrator` wake up the `SyncMigrateContext` with a migration result status, it has to release the pointer so that no context will "occupy" the migrator any more (see [here](https://github.com/apache/incubator-kvrocks/pull/1418/files#diff-6a88559b7b94f9543cde3449bda77668eba37d8696153da4f244b250054e809aR1113)). 
However, at this time point, the `WriteCB` have not been triggered by the event loop yet, thus the context cannot be deconstructed yet. Using a shared pointer will bind the life cycle of the `SyncMigrateContext` with the `CommandClusterX` object, which will solve this issue.

PragmaTwice:
@ellutionist Thanks for your explanation.

I have not dived into these changes deeply, but if you just want to make lifetime of SyncMigrateContext as long as the command object, you can store an instance of SyncMigrateContext in CommandClusterX directly (`class CommandClusterX { SyncMigrateContext ctx_; ... }`), and pass some non-owning pointer (like raw pointer, no ownership) to the migrator.

src/commands/cmd_cluster.cc
```diff
@@ -126,11 +127,28 @@ class CommandClusterX : public Commander {
     if (subcommand_ == "setnodeid" && args_.size() == 3 && args_[2].size() == kClusterNodeIdLen) return Status::OK();
 
     if (subcommand_ == "migrate") {
-      if (args.size() != 4) return {Status::RedisParseErr, errWrongNumOfArguments};
+      if (args.size() < 4) return {Status::RedisParseErr, errWrongNumOfArguments};
```
xiaobiaozhao: I think we can do a strong check here
eg : args.size() must be 5/6

src/commands/cmd_cluster.cc
```diff
@@ -126,11 +127,28 @@ class CommandClusterX : public Commander {
...
+
+          if (args.size() == 6) {
+            sync_migrate_timeout_ = GET_OR_RET(ParseInt<int64_t>(args[5], 10));
+          } else if (args.size() > 6) {
+            return {Status::RedisParseErr, "Wrong number of arguments for MIGRATE SYNC option"};
+          }
+        } else {
+          return {Status::RedisParseErr, "Invalid sync flag"};
+        }
+      }
```
xiaobiaozhao: I think it's a little bit clearer

if (args.size() == 5) {
   ...
}

if (args.size() == 6) {
   ...
}

src/cluster/slot_migrate.h
```diff
@@ -80,8 +82,8 @@ class SlotMigrator : public redis::Database {
   ~SlotMigrator();
 
   Status CreateMigrationThread();
-  Status PerformSlotMigration(const std::string &node_id, std::string &dst_ip, int dst_port, int slot_id, int speed,
-                              int pipeline_size, int seq_gap);
+  Status PerformSlotMigration(const std::string &node_id, std::string &dst_ip, int dst_port, int slot_id,
+                              const std::shared_ptr<SyncMigrateContext> &blocking_ctx = nullptr);
```
xiaobiaozhao: I think std::shared_ptr<T> blocking_ctx is enough

xiaobiaozhao:
> @ellutionist Thanks for your explanation.
> 
> I have not dived into these changes deeply, but if you just want to make lifetime of SyncMigrateContext as long as the command object, you can store an instance of SyncMigrateContext in CommandClusterX directly (`class CommandClusterX { SyncMigrateContext ctx_; ... }`), and pass some non-owning pointer (like raw pointer, no ownership) to the migrator.

I agree with that

ellutionist:
> @ellutionist Thanks for your explanation.
> 
> I have not dived into these changes deeply, but if you just want to make lifetime of SyncMigrateContext as long as the command object, you can store an instance of SyncMigrateContext in CommandClusterX directly (`class CommandClusterX { SyncMigrateContext ctx_; ... }`), and pass some non-owning pointer (like raw pointer, no ownership) to the migrator.

@PragmaTwice @xiaobiaozhao Thank you for the advice. This is another reasonable solution. However, there are two things to note here:
1. The `CommandClusterX` does not always have a `SyncMigrateContext` (only when migrating with a "sync" flag). Making the context a non-pointer member of the `CommandClusterX` object may not make good sense here. If we decide to do this, I suggest to adopt `std::optional<SyncMigrateContext>`.
2. Due to the concern of memory safety, I am not a fan of raw pointers. Of course, in this PR we could use it, but we may set up some potential risk for the future. Because not every maintainer could get to know  that "the`CommandClusterX` object must live long enough to complete the migration". If some change is made to "how the command objects work" in the future, it may lead to crash.

PragmaTwice:
> The CommandClusterX does not always have a SyncMigrateContext (only when migrating with a "sync" flag). Making the context a non-pointer member of the CommandClusterX object may not make good sense here. If we decide to do this, I suggest to adopt std::optional<SyncMigrateContext>.

Sure, you can use something like `std::optional` or `std::unique_ptr`.

> Due to the concern of memory safety, I am not a fan of raw pointers. Of course, in this PR we could use it, but we may set up some potential risk for the future. Because not every maintainer could get to know that "theCommandClusterX object must live long enough to complete the migration". If some change is made to "how the command objects work" in the future, it may lead to crash.

I think in kvrock, we use `std::shared_ptr` if and only if the object need to be **"shared"** between multiple references, and the lifetime of this object cannot be determined **statically** (e.g. given an object `o` and two reference `a` and `b`, the lifetime of `o` is equal to the maximum of lifetime of `a` and `b`, which need to be determined **dynamically**).

So currently, just *"in the future ..."* seems not to be an accredited reason to me, since if this reason holds, we may need to change many raw pointer in the codebase to `shared_ptr`. If you just need a non-owning pointer (or called an observer/view/span, in modern C++), raw pointer is currently the available choice, since [`std::observer_ptr`](https://en.cppreference.com/w/cpp/experimental/observer_ptr) or `std::optional<T&>` (optional with reference type) is not in the standard.

ellutionist:
> > The CommandClusterX does not always have a SyncMigrateContext (only when migrating with a "sync" flag). Making the context a non-pointer member of the CommandClusterX object may not make good sense here. If we decide to do this, I suggest to adopt std::optional.
> 
> Sure, you can use something like `std::optional` or `std::unique_ptr`.
> 
> > Due to the concern of memory safety, I am not a fan of raw pointers. Of course, in this PR we could use it, but we may set up some potential risk for the future. Because not every maintainer could get to know that "theCommandClusterX object must live long enough to complete the migration". If some change is made to "how the command objects work" in the future, it may lead to crash.
> 
> I think in kvrock, we use `std::shared_ptr` if and only if the object need to be **"shared"** between multiple references, and the lifetime of this object cannot be determined **statically** (e.g. given an object `o` and two reference `a` and `b`, the lifetime of `o` is equal to the maximum of lifetime of `a` and `b`, which need to be determined **dynamically**).
> 
> So currently, just _"in the future ..."_ seems not to be an accredited reason to me, since if this reason holds, we may need to change many raw pointer in the codebase to `shared_ptr`. If you just need a non-owning pointer (or called an observer/view/span, in modern C++), raw pointer is currently the available choice, since [`std::observer_ptr`](https://en.cppreference.com/w/cpp/experimental/observer_ptr) or `std::optional<T&>` (optional with reference type) is not in the standard.

Thank you. I got the point of the difference between dynamic and static lifetime. From this point of view the `share_ptr` is not proper in this case. Perhaps the raw pointer is the best solution here.

src/commands/cmd_cluster.cc
ellutionist: size should be between 4 to 6, will fix this.

src/commands/cmd_cluster.cc
ellutionist: There will be some duplicate code checking the `sync_flag` like:
```c++
if (args.size() == 5) {
 // check sync_flag = args[4]
}

if (args.size() == 6) {
 // check sync_flag = args[4]
}
```

src/cluster/slot_migrate.h
ellutionist: Changed to raw pointer.

git-hulk:
Hi @ellutionist Need to use `./x.py` format to correct the code format

src/cluster/sync_migrate_context.cc
```diff
@@ -0,0 +1,100 @@
...
+  auto s = conn_->Owner()->EnableWriteEvent(conn_->GetFD());
+  if (!s.IsOK()) {
+    LOG(ERROR) << "[server] Failed to enable write event on the sync migrate connection " << conn_->GetFD() << ": "
+               << s.Msg();
+  }
+}
+
+void SyncMigrateContext::EventCB(bufferevent *bev, int16_t events, void *ctx) {
+  auto self = reinterpret_cast<SyncMigrateContext *>(ctx);
+  auto &&slot_migrator = self->svr_->slot_migrator;
```
mapleFU: Why using `auto &&` here?

src/cluster/sync_migrate_context.cc
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    slot_migrator->CancelBlocking();
+  }
+  redis::Connection::OnEvent(bev, events, self->conn_);
+}
+
+void SyncMigrateContext::TimerCB(int, int16_t events, void *ctx) {
+  auto self = reinterpret_cast<SyncMigrateContext *>(ctx);
+  auto &&slot_migrator = self->svr_->slot_migrator;
```
mapleFU: ditto

mapleFU (commented):
Not familiar with networking logic in kvrocks, so just some style comments

src/cluster/sync_migrate_context.cc
ellutionist: This is a reference to `std::unique_ptr<SlotMigrator>`, which cannot be copied.

Technically `auto&` would suffice, but I personally prefer `auto&&` since it accepts more kinds of value (e.g. `const`, not this case though).

Of course, it is also okay to use `auto slot_migrator = self->svr_->slot_migrator.get()` (raw pointer) here. Just a choice of style.

ellutionist:
> Hi @ellutionist Need to use `./x.py` format to correct the code format

> Hi @ellutionist Need to use `./x.py` format to correct the code format

Done. It took me a while to adapt to the recent changes from [#1420]. 

Also some more go test cases were added.

git-hulk:
The implementation generally looks good to me, left a few comments:

- The blockingLock function looks unnecessary, it's fine to use `std::unique_lock<std::mutex>lock(blocking_mutex_);` directly
- For the naming issue, SyncMigrateContext::(StartBlock|Wakeup) is a bit inconsistent with Migrator::(StartBlocking|WakeupBlocking|CancelBlocking). And for the function name, how about using `Suspend|Resume|Cancel`?
- Those functions in SyncMigrateContext can be private:
  void OnWrite(bufferevent *bev); => void onWrite(bufferevent *bev);
  void OnEvent(bufferevent *bev, int16_t events); => void onEvent(bufferevent *bev, int16_t events);
  void TimerCB(int, int16_t events); => void timerCB(int, int16_t events);

PragmaTwice:
Seems need run `./x.py check golangci-lint`.

> Those functions in SyncMigrateContext can be private:

Currently we need to make them public for the CRTP base classes to use them.

git-hulk:
> Seems need run `./x.py check golangci-lint`.
> 
> > Those functions in SyncMigrateContext can be private:
> 
> Currently we need to make them public for the CRTP base classes to use them.

OK, I didn't notice this limitation.

git-hulk:
Looks like the migration test case has a segment fault. https://github.com/apache/incubator-kvrocks/actions/runs/4951001502/jobs/8855626634?pr=1418

ellutionist:
> Looks like the migration test case has a segment fault. https://github.com/apache/incubator-kvrocks/actions/runs/4951001502/jobs/8855626634?pr=1418

I think I found the reason. When building with ASAN the pointer of context won't initialize with `nullptr` by default. Should explicitly assign it a default value.

git-hulk (approved):

xiaobiaozhao (approved):

PragmaTwice (approved):
LGTM

torwig (approved):

PragmaTwice:
Merging... Thanks for your contribution!

PR #1428, opened 2023-05-07
feat: Catch up `ZRANGE` options

tisonkun:
This closes #1117.

~~I'll add tests later.~~

PR #1429, opened 2023-05-07
refactor: zset ranges

tisonkun:
This is a follow-up to #1428 for cleaning up some code.

cc @mapleFU

### DIFF
--- a/src/commands/cmd_zset.cc
+++ b/src/commands/cmd_zset.cc
@@ -502,7 +502,7 @@ class CommandZRemRangeByRank : public Commander {
     redis::ZSet zset_db(svr->storage, conn->GetNamespace());
 
     int cnt = 0;
-    spec_.removed = true;
+    spec_.with_deletion = true;
 
     auto s = zset_db.RangeByRank(args_[1], spec_, nullptr, &cnt);
     if (!s.ok()) {
@@ -531,7 +531,7 @@ class CommandZRemRangeByScore : public Commander {
     redis::ZSet zset_db(svr->storage, conn->GetNamespace());
 
     int cnt = 0;
-    spec_.removed = true;
+    spec_.with_deletion = true;
 
     auto s = zset_db.RangeByScore(args_[1], spec_, nullptr, &cnt);
     if (!s.ok()) {
@@ -560,7 +560,7 @@ class CommandZRemRangeByLex : public Commander {
     redis::ZSet zset_db(svr->storage, conn->GetNamespace());
 
     int cnt = 0;
-    spec_.removed = true;
+    spec_.with_deletion = true;
 
     auto s = zset_db.RangeByLex(args_[1], spec_, nullptr, &cnt);
     if (!s.ok()) {
--- a/src/commands/cmd_zset.cc
+++ b/src/commands/cmd_zset.cc
@@ -146,7 +146,7 @@ class CommandZCount : public Commander {
   }
 
  private:
-  CommonRangeScoreSpec spec_;
+  RangeScoreSpec spec_;
 };
 
 class CommandZCard : public Commander {
@@ -215,7 +215,7 @@ class CommandZLexCount : public Commander {
   }
 
  private:
-  CommonRangeLexSpec spec_;
+  RangeLexSpec spec_;
 };
 
 class CommandZPop : public Commander {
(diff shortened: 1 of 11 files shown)


src/commands/cmd_zset.cc
```diff
@@ -356,59 +356,49 @@ class CommandZRangeGeneric : public Commander {
...
+        s = zset_db.RangeByLex(key_, lex_spec_, &members, nullptr);
+        break;
+    }
+
+    if (!s.ok()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
 
-      *output = redis::MultiBulkString(members, false);
+    if (!members.empty()) {
```
git-hulk: It wouldn't reply to the request if the members and member_scores are empty lists, since the output is empty.

We still need to use MultiBulkString to reply the empty string array even if all of them are empty.

git-hulk: Below commands can reproduce this issue:

```
127.0.0.1:6666> zadd z1 1 a
(integer) 1
127.0.0.1:6666> zrange z1 2 3
```

The latter zrange command will block forever.

tisonkun: Make sense. Let me think of how we can do this polymorphic thing properly...

tisonkun: See https://github.com/apache/incubator-kvrocks/pull/1429/commits/1fffca8771c9d4e2da07f817dfa86fbfed8f746e

src/common/range_spec.h
```diff
@@ -24,36 +24,36 @@
...
-  explicit CommonRangeLexSpec() = default;
+  explicit RangeLexSpec() = default;
 };
 
-Status ParseRangeLexSpec(const std::string &min, const std::string &max, CommonRangeLexSpec *spec);
+Status ParseRangeLexSpec(const std::string &min, const std::string &max, RangeLexSpec *spec);
 
-struct CommonRangeRankSpec {
+struct RangeRankSpec {
   int start, stop;
```
mapleFU: Why start and stop doesn't have default value?

tisonkun: It's somehow always set. But we can provide one?

If you take a closer look, `min` and `max` in the other two specs don't have default value also.

src/types/redis_zset.cc
```diff
@@ -220,9 +219,12 @@ rocksdb::Status ZSet::Pop(const Slice &user_key, int count, bool min, std::vecto
...
 }
 
-rocksdb::Status ZSet::RangeByRank(const Slice &user_key, const CommonRangeRankSpec &spec,
-                                  std::vector<MemberScore> *mscores) {
-  mscores->clear();
+rocksdb::Status ZSet::RangeByRank(const Slice &user_key, const RangeRankSpec &spec, MemberScores *mscores, int *ret) {
+  if (mscores) mscores->clear();
+
+  int cnt = 0;
+  if (!ret) ret = &cnt;
```
torwig, line 226: In this case, you will return the address of the local variable `cnt`? I'm not sure that it is safe to do that in C++. In Go - no problem :)

src/commands/cmd_zset.cc
```diff
@@ -356,58 +356,41 @@ class CommandZRangeGeneric : public Commander {
...
 
-      return Status::OK();
+    switch (range_type_) {
+      case kZRangeLex:
+        output->append(redis::MultiBulkString(members, false));
+        return Status::OK();
+      case kZRangeAuto:
+      case kZRangeRank:
+      case kZRangeScore:
+        output->append(redis::MultiLen(member_scores.size() * (1 << with_scores_)));
```
torwig: Clever trick to shift a boolean variable - one line instead of 3. But maybe just use the ternary operator `? :` here?

src/commands/cmd_zset.cc
```diff
@@ -543,19 +528,22 @@ class CommandZRemRangeByScore : public Commander {
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    int size = 0;
     redis::ZSet zset_db(svr->storage, conn->GetNamespace());
-    auto s = zset_db.RemoveRangeByScore(args_[1], spec_, &size);
+
+    int ret = 0;
```
torwig: I'd prefer to have `size` or `count`/`cnt` instead of `ret` because the variable `ret` doesn't describe anything. I another functions as well.

tests/cppunit/types/zset_test.cc
```diff
@@ -325,9 +328,18 @@ TEST_F(RedisZSetTest, RemoveRangeByRank) {
   }
   zset_->Add(key_, ZAddFlags::Default(), &mscores, &ret);
   EXPECT_EQ(fields_.size(), ret);
-  zset_->RemoveRangeByRank(key_, 0, static_cast<int>(fields_.size() - 2), &ret);
+
+  RangeRankSpec spec;
+  spec.removed = true;
```
torwig: I'm trying to think about a better name for `removed` but can't come out with something short. All I have is `with_deletion`/`need_remove`. In my opinion `removed` describes something that has been already removed, and the condition `if (removed)` we can read as `if something was removed`. However, we need something like `if (need_deletion/with_deletion/should_remove)`.

src/common/range_spec.h
mapleFU: Seems that you set default min max for `RangeScoreSpec` in previous patch.

`RangeLexSpec`'s min max is `std::string`, they have default construct value "empty string".

Not setting is ok, but I wonder if start/stop be an undefined value here.

PragmaTwice: We'd better to give default values to scalar types, otherwise the value is unspecified.

src/types/redis_zset.cc
tisonkun: No. Here we change the value of local (param) variable `ret` when `nullptr` is passed. No point to local var is "returned". Just a trick to reduce `if (ret) ...` everywhere.

src/commands/cmd_zset.cc
tisonkun: Make sense. I don't have strong preference here and forget why I write in this way :D

src/common/range_spec.h
tisonkun: OK. Let me set it to 0 and -1, correspondingly?

src/commands/cmd_zset.cc
tisonkun: Let's use `cnt` here.

tests/cppunit/types/zset_test.cc
tisonkun: Make sense. I'll use `with_deletion` now.

src/common/range_spec.h
tisonkun: Updated.

git-hulk (approved):

src/types/redis_zset.cc
torwig: Right, now I see.

torwig (approved):

tisonkun:
Merging...

PR #1430, opened 2023-05-07
Return error from CommandZRangeGeneric in case of unexpected range_type

torwig:
During the build process compiler raised a warning about the absence of a `return` statement in a non-void function.


PragmaTwice (approved):

tisonkun (approved):
This should be logically exhausted though 🤣

PR #1434, opened 2023-05-08
Docker image based on Alpine Linux

aleksraiden:
In this PR we are completely rewrite a Docker image:

- A build stage and runtime are based on Alpine Linux (now on 3.16 because we need an libexecinfo, that's no port into latest) 
- Improve dockerignore file to minimize context transferring
- Timezone change into UTC 
- Build now produce an architect-optimize and Release version
- Utilize an all system cores in build process 
- For best security we are using an dedicated group and user (no login, no root privilegies, no home dir etc.)
- Add native healthcheck into docker container using HEALTHCHECK command and redis-cli 

In a result:

- Image size reduce 10х - from 400+ Mb in default to 40Mb
- Build now up to 300 sec (more core - less time)
- Healthchecks
- More security

P.S. Thanks for @torwig about an help and support me

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,18 +23,18 @@ ARG MORE_BUILD_ARGS
 ENV TZ=Etc/UTC
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
-RUN apk update && apk add git gcc g++ make cmake ninja autoconf automake libtool python3 linux-headers curl openssl3-dev libexecinfo-dev redis
+RUN apk update && apk add git gcc g++ make cmake ninja autoconf automake libtool python3 linux-headers curl openssl-dev libexecinfo-dev redis
 WORKDIR /kvrocks
 
 COPY . .
 RUN ./x.py build -DENABLE_OPENSSL=ON -DCMAKE_BUILD_TYPE=Release -j $(nproc) $MORE_BUILD_ARGS
 
-FROM alpine:3.18
+FROM alpine:3.16
 
 ENV TZ=Etc/UTC
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
-RUN apk upgrade && apk add libexecinfo --update-cache --repository http://dl-cdn.alpinelinux.org/alpine/v3.16/main/ --allow-untrusted
+RUN apk upgrade && apk add libexecinfo
 
 WORKDIR /kvrocks
 
@@ -49,7 +49,7 @@ USER kvrocks
 COPY --from=build /kvrocks/build/kvrocks ./bin/
 COPY --from=build /usr/bin/redis-cli ./bin/
 
-HEALTHCHECK --interval=5s --timeout=1s --start-period=120s --retries=3 CMD ./bin/redis-cli -p 6666 PING | grep PONG || exit 1
+HEALTHCHECK --interval=5s --timeout=1s --start-period=5s --retries=3 CMD ./bin/redis-cli -p 6666 PING | grep PONG || exit 1
 
 ENV PATH="$PATH:/kvrocks/bin"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,7 +49,7 @@ USER kvrocks
 COPY --from=build /kvrocks/build/kvrocks ./bin/
 COPY --from=build /usr/bin/redis-cli ./bin/
 
-HEALTHCHECK --interval=5s --timeout=1s --start-period=5s --retries=3 CMD ./bin/redis-cli -p 6666 PING | grep -E '(PONG|NOAUTH)' || exit 1
+HEALTHCHECK --interval=10s --timeout=1s --start-period=30s --retries=3 CMD ./bin/redis-cli -p 6666 PING | grep -E '(PONG|NOAUTH)' || exit 1
 
 ENV PATH="$PATH:/kvrocks/bin"
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,43 +15,46 @@
 # specific language governing permissions and limitations
 # under the License.
 
-FROM ubuntu:focal as build
+FROM alpine:3.16 as build
 
 ARG MORE_BUILD_ARGS
 
 # workaround tzdata install hanging
-ENV TZ=Asia/Shanghai
+ENV TZ=Etc/UTC
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
-RUN apt update && apt install -y git gcc g++ make cmake autoconf automake libtool python3 libssl-dev curl
+RUN apk update && apk add git gcc g++ make cmake autoconf automake libtool python3 linux-headers curl openssl openssl-dev libexecinfo-dev
 WORKDIR /kvrocks
 
 COPY . .
-RUN ./x.py build -DENABLE_OPENSSL=ON -DPORTABLE=ON $MORE_BUILD_ARGS
+RUN ./x.py build -DENABLE_OPENSSL=ON -DCMAKE_BUILD_TYPE=Release -j $(nproc) $MORE_BUILD_ARGS
 
-RUN curl -O https://download.redis.io/releases/redis-6.2.7.tar.gz && \
-    tar -xzvf redis-6.2.7.tar.gz && \
-    mkdir tools && \
-    cd redis-6.2.7 && \
-    make redis-cli && \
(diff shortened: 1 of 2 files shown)


tisonkun:
For less dependencies and small image, what about using [distroless/cc](https://github.com/GoogleContainerTools/distroless/blob/main/cc/README.md) as base for the final image?

aleksraiden:
> For less dependencies and small image, what about using [distroless/cc](https://github.com/GoogleContainerTools/distroless/blob/main/cc/README.md) as base for the final image?

It's a good idea, lets work on this in next iteration, thanks!

PragmaTwice:
Could you provide a preview image on docker hub for quick review?

aleksraiden:
> Could you provide a preview image on docker hub for quick review?

https://hub.docker.com/repository/docker/aleksraiden/kvrocks/general

aleksraiden:
I have work on a few patch to improve docker with new alpine image, please waiting before merge

aleksraiden:
So, new version are ready:

- Use redis from official alpine repo instead of manual build (stable 7.х version - 7.0.11)
- Add ninja for build dependency (if you want to use it)
- Use UTC timezone as default in run image too
- Change healthcheck start delay to 5 sec.

In result we have:

- ~10% less container size (34.9 Mb vs. 40 Mb)
- 30% faster container build

torwig:
@aleksraiden Great result!

aleksraiden:
Pushed to Docker Hub - https://hub.docker.com/layers/aleksraiden/kvrocks/alpine-latest-v2/images/sha256-8a59b0ab7c79bf11375a4e41919b3ecfefb6c41d8f36536cb12e9b5c37cb3a4e?context=explore

aleksraiden:
Some note - we are using alpine-3.16 instead of latest 3.18 because a oneTBB library can't build in GCC 13. Other point is using redis from repo instead of manual build. In Alpine 3.17+ a default openssl is 3.х - so we need install openssl1.1-compat to use with repo-based redis or build manual for native openssl3 support. 

So, my  point is a waiting for TBB release with build fixes and check how we use an a latest Alpine image.

Dockerfile
```diff
@@ -15,43 +15,52 @@
...
+RUN addgroup -S kvrocks && adduser -D -H -S -G kvrocks kvrocks
+
+RUN chown kvrocks:kvrocks /var/run/kvrocks && chown kvrocks:kvrocks /var/lib/kvrocks
+
+USER kvrocks
+
 COPY --from=build /kvrocks/build/kvrocks ./bin/
+COPY --from=build /usr/bin/redis-cli ./bin/
+
+HEALTHCHECK --interval=5s --timeout=1s --start-period=5s --retries=3 CMD ./bin/redis-cli -p 6666 PING | grep PONG || exit 1
```
PragmaTwice: I am wondering whether it will work well if users set `requirepass` in config

Dockerfile
```diff
@@ -15,43 +15,52 @@
...
+ENV TZ=Etc/UTC
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
-RUN apt update && apt install -y git gcc g++ make cmake autoconf automake libtool python3 libssl-dev curl
+RUN apk update && apk add git gcc g++ make cmake ninja autoconf automake libtool python3 linux-headers curl openssl-dev libexecinfo-dev redis
 WORKDIR /kvrocks
 
 COPY . .
-RUN ./x.py build -DENABLE_OPENSSL=ON -DPORTABLE=ON $MORE_BUILD_ARGS
+RUN ./x.py build -DENABLE_OPENSSL=ON -DCMAKE_BUILD_TYPE=Release -j $(nproc) $MORE_BUILD_ARGS
```
PragmaTwice: I think `-DPORTABLE=ON` cannot be removed, since the build machine will have different arch version compared to the user machine (e.g. if avx512 in the build machine but not in the user machine, users will get "illegal instruction" crash).
And I think we use `RelWithDebInfo` instead of `Release` since we want to keep the debug info in the binary (which may make the binary more large). These options, like `-DCMAKE_BUILD_TYPE=Release` and `-j $(nproc)`, can also be passed from `MORE_BUILD_ARGS`, e.g. `docker build --build-arg MORE_BUILD_ARGS=...` (maybe, not confirmed yet) if you want these options.

PragmaTwice (commented):
Generally looks good to me. A little comments.

Dockerfile
aleksraiden: Oh, thanks! I have a hard thinking for this issue. So, a way with a require password in params are so unclean and not like to me. 

We have a two way for this:

1. Changing PING command for using without AUTH (Better for me, but need to patch kvrocks)
2. Changing an healthcheck to this "./bin/redis-cli -p 6666 PING |  grep -E '(PONG|NOAUTH)'  || exit 1" - so if server require password, an NOAUTH response also tell us that's server are alive.

I think, second scenario is simplest way at now.

Dockerfile
aleksraiden: A lot of thanks about detailed comment. Yes, reason for -DPORTABLE=ON sounds good, I reverting this (in a quick test, performance aren't changed a lot of) . 

But for the rest, I think, we may use an default docker build for build production-ready system (a most usages way, as I think), and just add a documentation with build instruction for produce more optimize (-DPORTABLE=OFF) or more debug-ready system (-DCMAKE_BUILD_TYPE=RelWithDebInfo). 

As I think, in my mind, a lot of people just use an default docker image, so we can build them as better performance as we can and just provide a documentation about other options for use.

torwig (dismissed):
LGTM

Dockerfile
```diff
@@ -15,43 +15,52 @@
...
-    make redis-cli && \
-    mv src/redis-cli /kvrocks/tools/redis-cli
+FROM alpine:3.16
 
-FROM ubuntu:focal
+ENV TZ=Etc/UTC
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
 
-RUN apt update && apt install -y libssl-dev
+RUN apk upgrade && apk add libexecinfo
```
PragmaTwice, line 37: I noticed that libopenssl is missing in the docker image. Will it lead the kvrocks to crash?

Dockerfile
```diff
@@ -15,43 +15,52 @@
...
+RUN addgroup -S kvrocks && adduser -D -H -S -G kvrocks kvrocks
+
+RUN chown kvrocks:kvrocks /var/run/kvrocks && chown kvrocks:kvrocks /var/lib/kvrocks
+
+USER kvrocks
+
 COPY --from=build /kvrocks/build/kvrocks ./bin/
+COPY --from=build /usr/bin/redis-cli ./bin/
+
+HEALTHCHECK --interval=5s --timeout=1s --start-period=5s --retries=3 CMD ./bin/redis-cli -p 6666 PING | grep -E '(PONG|NOAUTH)' || exit 1
```
PragmaTwice: I am not sure whether the interval and start-period value (both 5s) is well adjusted, since I am not familiar with them. cc @git-hulk

aleksraiden: start-period I can change to something like 30s - because if we have loading a big database, server after starting need a time to initialize.

Dockerfile
aleksraiden: In base image of alpine we have openssl already installed, so we don't need a manual add for it

Dockerfile
git-hulk: I think it's unnecessary to do the health check in the Kvrocks container since it doesn't depend on other resources except for the FS.

aleksraiden: This is need for a swarm / compose scenario and provide a very basic layer for liveness container with a very low overhead, so we can some change a timeouts only.

git-hulk: OK, I am not familiar with Swarm, but it makes sense.

aleksraiden: Change intervals to 10sec and delay 30 sec on start

tisonkun: > compose scenario and provide a very basic layer for liveness container

FWIW, IIRC health check can be added in the compose yaml file also. For example: [docker-compose.e2e.yml](https://github.com/apache/skywalking-rust/blob/30c8d8c10c3ab684e37a1339e7ba2c054f370e1f/docker-compose.e2e.yml#L25)

xiaobiaozhao (approved):
LGTM

torwig (approved):

xiaobiaozhao (approved):

tisonkun:
I try out to work with distroless + glibc. The result image size is still 40MB. No need to invest more time since that base image is good for small size only - it lacks of most functions.

PR #1444, opened 2023-05-18
Add support of the new command SINTERCARD(Redis 7)

### LINKED ISSUE #1083: [NEW] Add `SINTERCARD` command aligned Redis

infdahai:
closes #1083
closes #1116 (replace)

This PR introduces a `sintercard` command to `kvrocks` from redis 7.0.

~BTW, the `go-redis` package's version is `v9.0.0-beta.2` in the project so it doesn't support this command(more infomation in https://github.com/redis/go-redis/pull/2291). And I just write a bare test in `go-test` now.~

~Perhaps we can bump `go-redis` to a newer version.~

### DIFF
--- a/src/commands/cmd_set.cc
+++ b/src/commands/cmd_set.cc
@@ -284,8 +284,7 @@ class CommandSInter : public Commander {
 
     std::vector<std::string> members;
     redis::Set set_db(svr->storage, conn->GetNamespace());
-    uint64_t cnt = 0;
-    auto s = set_db.Inter(keys, &members, 0, &cnt);
+    auto s = set_db.Inter(keys, &members);
     if (!s.ok()) {
       return {Status::RedisExecErr, s.ToString()};
     }
--- a/src/commands/cmd_set.cc
+++ b/src/commands/cmd_set.cc
@@ -18,6 +18,8 @@
  *
  */
 
+#include <cstdint>
+
 #include "commander.h"
 #include "commands/scan_base.h"
 #include "error_constants.h"
@@ -292,6 +294,72 @@ class CommandSInter : public Commander {
   }
 };
 
+/*
+ * description:
+ *    syntax:   `SINTERCARD numkeys key [key ...] [LIMIT limit]`
+ *
+ *    limit:    the valid limit is an non-negative integer.
+ */
+class CommandSInterCard : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto parse_numkey = ParseInt<int>(args[1], 10);
+    if (!parse_numkey) {
+      return {Status::RedisParseErr, errValueNotInteger};
+    }
+
+    if (*parse_numkey <= 0) {
+      return {Status::RedisParseErr, errValueMustBePositive};
+    }
+    numkeys_ = *parse_numkey;
+
+    // command: for example, SINTERCARD 2 key1 key2 LIMIT 1
+    auto arg_sz = args.size();
+    if (arg_sz == numkeys_ + 4 && util::ToLower(args[numkeys_ + 2]) == "limit") {
+      auto parse_limit = ParseInt<int>(args[numkeys_ + 3], 10);
+      if (!parse_limit) {
+        return {Status::RedisParseErr, errValueNotInteger};
+      }
+      if (*parse_limit < 0) {
+        return {Status::RedisParseErr, errLimitIsNegative};
+      }
+      limit_ = *parse_limit;
+      return Commander::Parse(args);
+    }
+
+    if (arg_sz != numkeys_ + 2) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+    return Commander::Parse(args);
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    std::vector<Slice> keys;
+    for (size_t i = 2; i < numkeys_ + 2; i++) {
+      keys.emplace_back(args_[i]);
+    }
+
+    redis::Set set_db(svr->storage, conn->GetNamespace());
+    uint64_t ret = 0;
+    auto s = set_db.InterCard(keys, limit_, &ret);
+    if (!s.ok()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    *output = redis::Integer(ret);
+    return Status::OK();
+  }
+
+  static CommandKeyRange Range(const std::vector<std::string> &args) {
+    int num_key = *ParseInt<int>(args[1], 10);
+    return {2, 1 + num_key, 1};
+  }
+
+ private:
+  uint64_t numkeys_ = 0;
+  uint64_t limit_ = 0;
+};
+
 class CommandSDiffStore : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
@@ -380,6 +448,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandSAdd>("sadd", -3, "write", 1, 1, 1),
                         MakeCmdAttr<CommandSDiff>("sdiff", -2, "read-only", 1, -1, 1),
                         MakeCmdAttr<CommandSUnion>("sunion", -2, "read-only", 1, -1, 1),
                         MakeCmdAttr<CommandSInter>("sinter", -2, "read-only", 1, -1, 1),
+                        MakeCmdAttr<CommandSInterCard>("sintercard", -3, "read-only", CommandSInterCard::Range),
                         MakeCmdAttr<CommandSDiffStore>("sdiffstore", -3, "write", 1, -1, 1),
                         MakeCmdAttr<CommandSUnionStore>("sunionstore", -3, "write", 1, -1, 1),
                         MakeCmdAttr<CommandSInterStore>("sinterstore", -3, "write", 1, -1, 1),
--- a/src/commands/cmd_set.cc
+++ b/src/commands/cmd_set.cc
@@ -284,7 +284,8 @@ class CommandSInter : public Commander {
 
     std::vector<std::string> members;
     redis::Set set_db(svr->storage, conn->GetNamespace());
-    auto s = set_db.Inter(keys, &members);
+    uint64_t cnt = 0;
+    auto s = set_db.Inter(keys, &members, 0, &cnt);
     if (!s.ok()) {
       return {Status::RedisExecErr, s.ToString()};
     }
--- a/src/types/redis_set.cc
+++ b/src/types/redis_set.cc
@@ -332,41 +332,66 @@ rocksdb::Status Set::Union(const std::vector<Slice> &keys, std::vector<std::stri
  * key3 = {a,c,e}
  * INTER key1 key2 key3 = {c}
  */
-rocksdb::Status Set::Inter(const std::vector<Slice> &keys, std::vector<std::string> *members, uint64_t limit,
-                           uint64_t *cnt) {
+rocksdb::Status Set::Inter(const std::vector<Slice> &keys, std::vector<std::string> *members) {
   members->clear();
 
   std::map<std::string, size_t> member_counters;
   std::vector<std::string> target_members;
   auto s = Members(keys[0], &target_members);
   if (!s.ok() || target_members.empty()) return s;
+  for (const auto &member : target_members) {
+    member_counters[member] = 1;
+  }
+  for (size_t i = 1; i < keys.size(); i++) {
+    s = Members(keys[i], &target_members);
+    if (!s.ok() || target_members.empty()) return s;
+    for (const auto &member : target_members) {
+      if (member_counters.find(member) == member_counters.end()) continue;
+      member_counters[member]++;
+    }
+  }
+  for (const auto &iter : member_counters) {
+    if (iter.second == keys.size()) {  // all the sets contain this member
+      members->emplace_back(iter.first);
+    }
+  }
+  return rocksdb::Status::OK();
+}
 
-  bool has_limit = limit != 0;
-  bool limit_reached = false;
+rocksdb::Status Set::InterCard(const std::vector<Slice> &keys, uint64_t limit, uint64_t *cnt) {
+  *cnt = 0;
+
+  std::map<std::string, size_t> member_counters;
+  std::vector<std::string> target_members;
+
+  auto s = Members(keys[0], &target_members);
+  if (!s.ok() || target_members.empty()) return s;
+  for (const auto &member : target_members) {
+    member_counters[member] = 1;
+  }
+  if (limit == 0) {
+    limit = target_members.size();
+  }
+
   size_t keys_size = keys.size();
-
-  if (keys_size == 1 && has_limit) {
+  if (keys_size == 1) {
     *cnt = std::min(static_cast<uint64_t>(target_members.size()), limit);
     return rocksdb::Status::OK();
   }
 
-  for (const auto &member : target_members) {
-    member_counters[member] = 1;
-  }
-
+  int cardinality = 0;
+  bool limit_reached = false;
   for (size_t i = 1; i < keys_size; i++) {
     s = Members(keys[i], &target_members);
     if (!s.ok() || target_members.empty()) {
-      *cnt = 0;
       return s;
     }
 
     for (const auto &member : target_members) {
       auto iter = member_counters.find(member);
       if (iter == member_counters.end()) continue;
-
-      if (++iter->second == keys_size && has_limit) {
-        members->emplace_back(member);
+      if (++iter->second == keys_size) {
+        cardinality++;
         if (--limit == 0) {
           limit_reached = true;
           break;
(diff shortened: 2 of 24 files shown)


src/types/redis_set.cc
```diff
@@ -338,19 +338,42 @@ rocksdb::Status Set::Inter(const std::vector<Slice> &keys, std::vector<std::stri
...
     s = Members(keys[i], &target_members);
     if (!s.ok() || target_members.empty()) return s;
     for (const auto &member : target_members) {
-      if (member_counters.find(member) == member_counters.end()) continue;
+      if (member_counters.count(member) == 0) continue;
       member_counters[member]++;
+      if (has_limit && member_counters[member] == keys.size()) {
+        members->emplace_back(member);
+        if (--limit == 0) {
+          limit = -1;
```
torwig: The variable `limit` is of type `uint64_t`. How it could be assigned to `-1` and later checked via the condition `limit < 0`?

torwig: Maybe the following code should be considered as a loop break:

```
if (has_limit && member_counters[member] == keys.size()) {
      members->emplace_back(member);
      if (members->size() >= limit) {
        break;
      }
}
```

tests/cppunit/types/set_test.cc
```diff
@@ -191,6 +191,23 @@ TEST_F(RedisSetTest, Inter) {
   set_->Del(k3);
 }
 
+TEST_F(RedisSetTest, InterCard) {
```
torwig, line 205: Additionally, you can add a test for a corner/edge case when one of the sets is empty. And one for the case when there are no common elements in both sets.

src/types/redis_set.cc
infdahai: oh, I forgot this type.

tests/cppunit/types/set_test.cc
infdahai: I plan to add these in `inter` tests later beacause `intercard` is a subset of calling `inter` .

src/commands/cmd_set.cc
```diff
@@ -283,6 +285,46 @@ class CommandSInter : public Commander {
...
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    std::vector<Slice> keys;
+    for (size_t i = 2; i < numkeys_ + 2; i++) {
+      keys.emplace_back(args_[i]);
+    }
+
+    redis::Set set_db(svr->storage, conn->GetNamespace());
+    int ret = 0;
```
torwig: You don't use this `ret` variable, you don't send any response to the command. Please, add some integration tests in Go to check the proper responses of the newly created command.

infdahai: > You don't use this ret variable

yeah, I add it now.

>  the go-redis package's version is v9.0.0-beta.2 in the project so it doesn't support this command

So we should wait for updating the go-redis package and then write go tests.

torwig: You can just use `Do()` function of the client to send raw request instead of the command, for example: https://github.com/apache/incubator-kvrocks/blob/unstable/tests/gocase/unit/type/stream/stream_test.go#L45. Or update the package.

infdahai: OK, I get it.

infdahai:
@git-hulk I haven't finished go tests because of waiting for https://github.com/apache/incubator-kvrocks/pull/1446. If I write it done, I will ask for reviewers.

src/types/redis_set.cc
```diff
@@ -338,19 +338,43 @@ rocksdb::Status Set::Inter(const std::vector<Slice> &keys, std::vector<std::stri
...
+  bool has_limit = limit != 0;
+  bool limited = false;
   for (size_t i = 1; i < keys.size(); i++) {
     s = Members(keys[i], &target_members);
     if (!s.ok() || target_members.empty()) return s;
     for (const auto &member : target_members) {
-      if (member_counters.find(member) == member_counters.end()) continue;
+      if (member_counters.count(member) == 0) continue;
       member_counters[member]++;
+      if (has_limit && member_counters[member] == keys.size()) {
```
PragmaTwice: Why do we search for the `member` element three times in `member_counters` continuously? It looks so bad to me.

```c++
auto iter = find(member); // find it
use1(iter); // use it
use2(iter); // use it again

// **NOT**:
use1(find(member));
use2(find(member));
use3(find(member));
```

src/types/redis_set.cc
```diff
@@ -338,19 +338,43 @@ rocksdb::Status Set::Inter(const std::vector<Slice> &keys, std::vector<std::stri
...
+      if (iter.second == keys.size()) {  // all the sets contain this member
+        members->emplace_back(iter.first);
+      }
     }
   }
+
+  return rocksdb::Status::OK();
+}
+
+rocksdb::Status Set::InterCard(const std::vector<Slice> &keys, uint64_t limit, int *ret) {
```
PragmaTwice: Maybe not related to this PR, but there are two points:
- `ret` is not a descriptive word
- `int` type is not fit here: the return variable `members.size()` is typed `size_t`, not `int`

tests/gocase/unit/type/set/set_test.go
```diff
@@ -282,6 +284,8 @@ func TestSet(t *testing.T) {
 			}
 		})
 
+		t.Run("SINTERCARD against three sets - "+dstype, func(t *testing.T) {})
```
PragmaTwice: maybe we can add more tests here.

src/types/redis_set.cc
infdahai: We find a few functions with `int *ret` in [redis_set.cc](https://github.com/apache/incubator-kvrocks/blob/unstable/src/types/redis_set.cc). However the result of `ret` is unsinged long(size_t). I will check the correct size type in multiple files.

infdahai: @PragmaTwice I exec the statement `*output = redis::Integer(ret);` to return the result to the client in cmd_set.cc.
https://github.com/apache/incubator-kvrocks/blob/fb0e3d42031a6e0691c7a736711d0a01d72bdb1a/src/server/redis_reply.h#L38-L41

Add the `Integer` func checks whether compatibility with `int` is required(Of course, `size_t` is right). I see src/types/*.h and many funcs with this situation use `int`.

src/commands/cmd_set.cc
```diff
@@ -283,6 +285,55 @@ class CommandSInter : public Commander {
   }
 };
 
+/*
+ * description:
+ *
+ *    syntax: SINTERCARD numkeys key [key ...] [LIMIT limit]
+ *
+ *    params: the checked cardinality is unlimited (limit <= 0).
```
torwig: Redis prohibits the negative `LIMIT` parameter:

```
127.0.0.1:6379> sintercard 2 set1 set2 limit -5
(error) ERR LIMIT can't be negative
127.0.0.1:6379> 
```

BTW: right now the phrase `the checked cardinality is unlimited` sounds weird to me. 
Since `Kvrocks` supports `Redis` protocol, if you want to add some documentation, you can just copy-paste sentences directly from `Redis` docs.

infdahai: Thanks, I got it. I just read the docs about this command so I make a mistake.

git-hulk:
The flaky test still exists after applying #1449 and it should be caused by the retry in Go redis client.

src/commands/cmd_set.cc
```diff
@@ -371,6 +418,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandSAdd>("sadd", -3, "write", 1, 1, 1),
                         MakeCmdAttr<CommandSDiff>("sdiff", -2, "read-only", 1, -1, 1),
                         MakeCmdAttr<CommandSUnion>("sunion", -2, "read-only", 1, -1, 1),
                         MakeCmdAttr<CommandSInter>("sinter", -2, "read-only", 1, -1, 1),
+                        MakeCmdAttr<CommandSInterCard>("sintercard", -3, "read-only", 1, -1, 1),
```
infdahai: @PragmaTwice Would it be better to use this change based on the syntax `SINTERCARD numkeys key [key ...] [LIMIT limit]` ? It seems this range is related to `watch`
```suggestion
                        MakeCmdAttr<CommandSInterCard>("sintercard", -3, "read-only", 2, -1, 1),
```

PragmaTwice: The best solution here is write a `CommandKeyRangeGen` function, since we have to determine the last position of the key sequence. 

And of course, the position of the first key is 2.

infdahai: Thanks. I've got it.

infdahai:
ready for reviews.

src/commands/cmd_set.cc
```diff
@@ -283,6 +285,65 @@ class CommandSInter : public Commander {
...
+        return {Status::RedisParseErr, errLimitIsNegative};
+      }
+    }
+    return Commander::Parse(args);
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    std::vector<Slice> keys;
+    for (size_t i = 2; i < numkeys_ + 2; i++) {
+      keys.emplace_back(args_[i]);
```
PragmaTwice, line 339: It seems there is no check for the boundary of `args_`, e.g. for `SINTERCARD 99 a b` will the `Execute` function just crash on this line since `args_[4]` and so on do not exist?

infdahai: Thank you so much. So many vulnerabilities. But I have a question about this, `go-redis` compute the `num_key` value and pass the all arguments like `redis` commands to the rpc.

Shall we do this check?

PragmaTwice: Any user input via redis protocol cannot just make kvrocks crash. That is the **principal commitment** of development.

And you cannot enforce users to always use go-redis wrapped command functions.

infdahai: OK, I understand this and thanks again.

git-hulk:
The flaky test should be caused by timeout, but didn't check the error in this case: https://github.com/apache/incubator-kvrocks/actions/runs/5070341119/jobs/9105869084#step:12:101

infdahai:
The PR doesn't have complete tests for wrong parsing msgs  based on the latest updating currently. Please wait me for a few days(Sorry, I am moving home).

git-hulk:
@infdahai Thanks, no hurry. Can ping back while your PR is ready to review.

infdahai:
@git-hulk I was resurrected. Ready to review.

tests/cppunit/types/set_test.cc
```diff
@@ -188,9 +188,41 @@ TEST_F(RedisSetTest, Inter) {
...
+  set_->InterCard({k1, k2}, 3, &ret);
+  EXPECT_EQ(ret, 2);
+  set_->InterCard({k2, k3}, 1, &ret);
+  EXPECT_EQ(ret, 1);
+  set_->InterCard({k1, k3}, 5, &ret);
+  EXPECT_EQ(ret, 0);
+  set_->InterCard({k1, k4}, 5, &ret);
+  EXPECT_EQ(ret, 0);
+  set_->InterCard({k1}, 0, &ret);
+  EXPECT_EQ(ret, 4);
```
torwig, line 227: If I add the following test:

```
set_->InterCard({k1}, 2, &ret);
EXPECT_EQ(ret, 2);
```

it will fail.

I tested it here: https://redis.io/commands/sintercard/

```
redis> sintercard 1 key1 limit 0
(integer) 4
redis> sintercard 1 key1 limit 2
(integer) 2
```

infdahai: OK, I'll add more tests later in corner cases. The test for a single key has been updated. Thanks a lot.

src/types/redis_set.cc
```diff
@@ -332,29 +332,66 @@ rocksdb::Status Set::Union(const std::vector<Slice> &keys, std::vector<std::stri
...
   members->clear();
 
   std::map<std::string, size_t> member_counters;
   std::vector<std::string> target_members;
   auto s = Members(keys[0], &target_members);
   if (!s.ok() || target_members.empty()) return s;
+
+  uint64_t limit = cnt ? *cnt : 0;
+  bool has_limit = limit != 0;
+  bool limited = false;
```
torwig: I think you can rename this variable to `limit_reached`.

torwig (commented):
In general, it seems like the functionality is correct.

However, I don't like the implementation because (if I understood properly) the `cnt` parameter of the `Set::Inter` function serves as the sign that this function calculates only cardinality AND passes the limit (which has only the `SINTERCARD` command) AND simultaneously serves as the return value. The idea to introduce a separate function directly is bothering me every time I review this PR.
The disadvantage is obviously the additional function in this case.

I would like to hear @git-hulk , @PragmaTwice on this.

src/types/redis_set.cc
infdahai: Yeah, I should split the func.

src/types/redis_set.cc
```diff
@@ -332,21 +332,21 @@ rocksdb::Status Set::Union(const std::vector<Slice> &keys, std::vector<std::stri
...
 
-  uint64_t limit = cnt ? *cnt : 0;
   bool has_limit = limit != 0;
-  bool limited = false;
+  bool limit_reached = false;
   size_t keys_size = keys.size();
 
   if (keys_size == 1 && has_limit) {
-    if (limit > target_members.size()) *cnt = target_members.size();
+    *cnt = limit > target_members.size() ? target_members.size() : limit;
```
PragmaTwice: Here maybe a nullptr check is needed for `cnt`. But IMHO, we can assume that `cnt` can never be nullptr, remove all these check and the default argument of `cnt` (`nullptr`) in the function declaration.

infdahai:
https://github.com/apache/incubator-kvrocks/actions/runs/5234808336/jobs/9452175246

Do we need to ignore this?

PragmaTwice:
> https://github.com/apache/incubator-kvrocks/actions/runs/5234808336/jobs/9452175246
> 
> Do we need to ignore this?

Nope. The build procedure need to succeed in Darwin. It cannot be ignored.

You can add some type cast before the std::min call to make it compile successfully.

torwig:
@infdahai You didn't actually split the `Inter` function into two separate :) I meant that the `Inter` function should calculate the intersection between sets and return members while `InterCard` should only calculate the cardinality and return the computed value. The command `Inter` will call `Set::Inter` and the command `InterCard` will call the `Set::InterCard` (currently the `Set::InterCard` function calls `Set::Inter` and all the complexity is still in the `Set::Inter` function). In this case, you will get rid of some of the `if` statements and simplify the code.

infdahai:
> @infdahai You didn't actually split the `Inter` function into two separate :) I meant that the `Inter` function should calculate the intersection between sets and return members while `InterCard` should only calculate the cardinality and return the computed value. The command `Inter` will call `Set::Inter` and the command `InterCard` will call the `Set::InterCard` (currently the `Set::InterCard` function calls `Set::Inter` and all the complexity is still in the `Set::Inter` function). In this case, you will get rid of some of the `if` statements and simplify the code.

@torwig I just thought if I split the functions, these two functions exist duplicate codes before and we add the limit `0` to `Set::Inter` to unify the two functions currently. And I also find the generic function is integrated in redis. https://github.com/redis/redis/blob/f228ec1ea5ef8f3734f1dc0297c5181a651aba1d/src/t_set.c#L1276

I remove the if statements but use duplicate `for` statements. Of course, I will do as you say.

src/types/redis_set.cc
```diff
@@ -358,6 +358,54 @@ rocksdb::Status Set::Inter(const std::vector<Slice> &keys, std::vector<std::stri
...
+    s = Members(keys[i], &target_members);
+    if (!s.ok() || target_members.empty()) {
+      return s;
+    }
+
+    for (const auto &member : target_members) {
+      auto iter = member_counters.find(member);
+      if (iter == member_counters.end()) continue;
+      if (++iter->second == keys_size) {
+        cardinality++;
```
torwig: Is there any reason to introduce a new variable `cardinality` and later increment it instead of incrementing the `cnt` variable?

infdahai: ignore this. I don't like accessing pointers all the time.

torwig (dismissed):
Generally, LGTM.

src/types/redis_set.cc
```diff
@@ -358,6 +358,52 @@ rocksdb::Status Set::Inter(const std::vector<Slice> &keys, std::vector<std::stri
...
+    s = Members(keys[i], &target_members);
+    if (!s.ok() || target_members.empty()) {
+      return s;
+    }
+
+    for (const auto &member : target_members) {
+      auto iter = member_counters.find(member);
+      if (iter == member_counters.end()) continue;
+      if (++iter->second == keys_size) {
+        *cardinality++;
```
infdahai:
```suggestion
        *cardinality += 1;
```

git-hulk (approved):

torwig (approved):

PragmaTwice (approved):

git-hulk:
Thanks all, merging...

git-hulk:
@infdahai Can add the SINTERCARD command to https://github.com/apache/incubator-kvrocks-website/blob/main/docs/supported-commands.md

PR #1449, opened 2023-05-19
Fix flaky test case in Go unit/type/list.go

git-hulk:
Currently, we use the list key 'l' in multiple test cases and they may interact with each other. So we fix it by using a dedicated key for each test case.

For failure logs can see: https://github.com/apache/incubator-kvrocks/actions/runs/5014908898/jobs/8991680649#step:12:104


torwig (approved):

tisonkun (approved):

git-hulk:
Thanks all, merging...

PR #1463, opened 2023-05-22
Don't use auto-retry in unit/type/list_test.go

git-hulk:
Currently, go-redis will auto-retry if the request was timeout and it may cause the duplicate push in the list test case. So we need to manually disable it to reduce the possibility of the CI test.

This flaky test wasn't caused by using the same key in multiple test cases, which was resolved in #1449 and still failed frequently. After investigating the test case, the possible reason may be caused by the go-redis auto-retry strategy.


torwig (approved):

git-hulk:
Thanks @torwig, merging..

PR #1464, opened 2023-05-22
Fix flaky test in migration timeout

### LINKED ISSUE #1452: migrate timeout in Go Integration Cases
### Search before asking

### Version

unstable

### Minimal reproduce step

the ref to CI jobs: https://github.com/infdahai/incubator-kvrocks/actions/runs/5032023303/jobs/9025509337

The change is on https://github.com/infdahai/incubator-kvrocks/compare/fb0e3d4...7a5a2c2. I only run 10  tests in a loop.

All tests pass， especially the `Migrate_sync_with_(or_without)_all_kinds_of_timeouts` test.

### What did you see instead?

  --- FAIL: TestSlotMigrateSync/MIGRATE_-_Migrate_sync_with_(or_without)_all_kinds_of_timeouts#09 (0.01s)
        slotmigrate_test.go:426: 
            	Error Trace:	/home/runner/work/incubator-kvrocks/incubator-kvrocks/tests/gocase/integration/slotmigrate/slotmigrate_test.go:426
            	Error:      	Not equal: 
            	            	expected: string("OK")
            	            	actual  : <nil>(<nil>)
            	Test:       	TestSlotMigrateSync/MIGRATE_-_Migrate_sync_with_(or_without)_all_kinds_of_timeouts#09

### Anything Else?

_No response_

git-hulk:
This closes #1452

### DIFF
--- a/src/cluster/sync_migrate_context.h
+++ b/src/cluster/sync_migrate_context.h
@@ -25,7 +25,7 @@
 class SyncMigrateContext : private EvbufCallbackBase<SyncMigrateContext, false>,
                            private EventCallbackBase<SyncMigrateContext> {
  public:
-  SyncMigrateContext(Server *svr, redis::Connection *conn, float timeout) : svr_(svr), conn_(conn), timeout_(timeout){};
+  SyncMigrateContext(Server *svr, redis::Connection *conn, int timeout) : svr_(svr), conn_(conn), timeout_(timeout){};
 
   void Suspend();
   void Resume(const Status &migrate_result);
@@ -36,7 +36,7 @@ class SyncMigrateContext : private EvbufCallbackBase<SyncMigrateContext, false>,
  private:
   Server *svr_;
   redis::Connection *conn_;
-  float timeout_ = 0;
+  int timeout_ = 0;
   UniqueEvent timer_;
 
   Status migrate_result_;
--- a/src/commands/cmd_cluster.cc
+++ b/src/commands/cmd_cluster.cc
@@ -151,7 +151,7 @@ class CommandClusterX : public Commander {
               return {Status::RedisParseErr, "timeout is not an integer or out of range"};
             }
             if (*parse_result < 0) {
-              return {Status::RedisParseErr, "timeout should not be negative"};
+              return {Status::RedisParseErr, errTimeoutIsNegative};
             }
             sync_migrate_timeout_ = *parse_result;
           }
--- a/src/commands/cmd_cluster.cc
+++ b/src/commands/cmd_cluster.cc
@@ -146,7 +146,14 @@ class CommandClusterX : public Commander {
           sync_migrate_ = true;
 
           if (args.size() == 6) {
-            sync_migrate_timeout_ = GET_OR_RET(ParseFloat<float>(args[5]));
+            auto parse_result = ParseInt<int>(args[5], 10);
+            if (!parse_result) {
+              return {Status::RedisParseErr, "timeout is not an integer or out of range"};
+            }
+            if (*parse_result < 0) {
+              return {Status::RedisParseErr, "timeout should not be negative"};
+            }
+            sync_migrate_timeout_ = *parse_result;
           }
         } else {
           return {Status::RedisParseErr, "Invalid sync flag"};
@@ -277,7 +284,7 @@ class CommandClusterX : public Commander {
   bool force_ = false;
 
   bool sync_migrate_ = false;
-  float sync_migrate_timeout_ = 0;
+  int sync_migrate_timeout_ = 0;
   std::unique_ptr<SyncMigrateContext> sync_migrate_ctx_ = nullptr;
 };
 
--- a/tests/gocase/integration/slotmigrate/slotmigrate_test.go
+++ b/tests/gocase/integration/slotmigrate/slotmigrate_test.go
@@ -413,22 +413,18 @@ func TestSlotMigrateSync(t *testing.T) {
 
 	t.Run("MIGRATE - Migrate sync with (or without) all kinds of timeouts", func(t *testing.T) {
 		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync").Val())
+		// migrate sync without explicitly specify the timeout
+		result := rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync")
+		require.NoError(t, result.Err())
+		require.Equal(t, "OK", result.Val())
 
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", -1).Val())
-
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", 0).Val())
-
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", 10).Val())
-
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", 0.5).Val())
-
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", -3.14).Val())
+		timeouts := []interface{}{-1, 0, 10, 0.5, -3.14}
+		for _, timeout := range timeouts {
+			slot++
+			result := rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", timeout)
+			require.NoError(t, result.Err())
+			require.Equal(t, "OK", result.Val())
+		}
 	})
 
 	t.Run("MIGRATE - Migrate sync timeout", func(t *testing.T) {
--- a/tests/gocase/integration/slotmigrate/slotmigrate_test.go
+++ b/tests/gocase/integration/slotmigrate/slotmigrate_test.go
@@ -435,17 +435,15 @@ func TestSlotMigrateSync(t *testing.T) {
 	})
 
 	t.Run("MIGRATE - Migrate sync timeout", func(t *testing.T) {
-		cnt := 200000
 		slot++
+		cnt := 200000
 		for i := 0; i < cnt; i++ {
 			require.NoError(t, rdb0.LPush(ctx, util.SlotTable[slot], i).Err())
 		}
-		timeout := 0.001
 
-		require.Nil(t, rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", timeout).Val())
-
-		// check the following command on the same connection
-		require.Equal(t, "PONG", rdb0.Ping(ctx).Val())
+		timeout := 1
+		result := rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", timeout)
+		require.Nil(t, result.Val())
 	})
 }
 
--- a/tests/gocase/integration/slotmigrate/slotmigrate_test.go
+++ b/tests/gocase/integration/slotmigrate/slotmigrate_test.go
@@ -418,7 +418,7 @@ func TestSlotMigrateSync(t *testing.T) {
 		require.NoError(t, result.Err())
 		require.Equal(t, "OK", result.Val())
 
-		invalidTimeouts := []interface{}{-2, -1, "abc"}
+		invalidTimeouts := []interface{}{-2, -1, 1.2, "abc"}
 		for _, timeout := range invalidTimeouts {
 			slot++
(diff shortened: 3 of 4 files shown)


git-hulk:
We have reproduced this flaky error: https://github.com/apache/incubator-kvrocks/actions/runs/5044728692/jobs/9055654883. It's caused by the slot conflict when running the migration, let's increase the start slot to avoid this.

tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -412,23 +412,19 @@ func TestSlotMigrateSync(t *testing.T) {
...
-
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", 10).Val())
-
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", 0.5).Val())
-
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", -3.14).Val())
+		slot := 1024
```
tisonkun: Can we add a comment for this magic number? From the description it's to avoid slot num conflict.

git-hulk: Sure. To be honest, we didn't take care of those slot distributions when writing test cases, so it's a bit confusing for using which slot now.

infdahai: The reason for this error is that the sync time is too small such as `0.5s` and CI often runs failed while local machines run successful. And the commit https://github.com/apache/incubator-kvrocks/commit/635dd311e5d07007826ae5338eae4cefb88ccc55 also improves read success rate.

infdahai: So when I use a very small time parameter causing a failure, the program calls `resume` func and returns `nil` value. Maybe we need the following test. But the following statements may fail from my tests with throwing `There is already a migrating slot`.

```go
				// for result.Val() == nil && timeout < 15 {
				// 	timeout += 5
				// 	result := rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", timeout)
				// 	require.NoError(t, result.Err())
				// }
				// 	require.Equal(t, "OK", result.Val())

```

git-hulk: Yes, [8c76f33](https://github.com/apache/incubator-kvrocks/pull/1464/commits/8c76f3339b344b13a9964a075d08737c9e95bb56) this commit has increased the timeout.

infdahai (dismissed):

torwig (dismissed):

tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -422,7 +422,7 @@ func TestSlotMigrateSync(t *testing.T) {
 		for _, timeout := range timeouts {
 			slot++
 			result := rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", timeout)
-			require.NoError(t, result.Err())
+			require.NoError(t, result.Err(), "without timeout: %v", timeout)
```
ellutionist: Should the message be "with timeout"?

tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -422,7 +422,7 @@ func TestSlotMigrateSync(t *testing.T) {
...
 			require.Equal(t, "OK", result.Val())
```
ellutionist: Perhaps add "with timeout" message here also?

tests/gocase/integration/slotmigrate/slotmigrate_test.go
git-hulk: Yes, it's a typo.

tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -413,22 +413,18 @@ func TestSlotMigrateSync(t *testing.T) {
...
+		result := rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync")
+		require.NoError(t, result.Err())
+		require.Equal(t, "OK", result.Val())
+
+		timeouts := []interface{}{-1, 0, 20, 10.5, -3.14}
+		for _, timeout := range timeouts {
+			slot++
+			result := rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", timeout)
+			require.NoError(t, result.Err(), "without timeout: %v", timeout)
+			require.Equal(t, "OK", result.Val())
```
git-hulk:
```suggestion
			require.NoError(t, result.Err(), "with timeout: %v", timeout)
			require.Equal(t, "OK", result.Val(), "with timeout: %v", timeout)
```

tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -413,22 +413,18 @@ func TestSlotMigrateSync(t *testing.T) {
...
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", 0.5).Val())
-
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", -3.14).Val())
+		// migrate sync without explicitly specify the timeout
+		result := rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync")
+		require.NoError(t, result.Err())
+		require.Equal(t, "OK", result.Val())
+
+		timeouts := []interface{}{-1, 0, 20, 10.5, -3.14}
```
torwig: Could we use `[]float64` here instead of `[]interface{}`?
If the timeout is a negative value, should it be an error?

torwig: I see that right now timeout < 0 doesn't produce an error. Should we introduce an error in this case? Or just treat it as an absence of a timeout?

git-hulk: For the negative timeout, I also prefer retrieving an error instead of regarding it as 0. But I think we can submit another PR to fix this.

> Could we use []float64 here instead of []interface{}?

It's intent to use the interface{} since I wanna make sure that the int and float types are working well.

torwig: @git-hulk Do we ever need floats there? I see that float is cast to int in the code, so I think we should consider timeout as an integer number of seconds and prohibit float and negative values.

git-hulk: Good catch, I didn't dive into the implementation. For the server side, it allows the float timeout now, but I think it's unnecessary to support that.

git-hulk: I will append a commit to resolve the above comments.

ellutionist: @torwig @git-hulk The original reason to support float timeout is that in the go tests I want to "deliberately let the sync migration timeout" (see [here](https://github.com/apache/incubator-kvrocks/blob/unstable/tests/gocase/integration/slotmigrate/slotmigrate_test.go#L440)). And the minimal integer value of "1" is too large and cannot ensure the timeout happen...

Also I think supporting float timeout is not harmful but provides more precise control to the migration. What do you think?

As for the negative value, returning an error when the server receives an negative value seems acceptable to me. No problem.

torwig: @ellutionist Timeout is in seconds as far as I see so I don't think that making the timeout less than 1 second is reasonable (it's about migrating slot with data from one node to another via network).

git-hulk: @ellutionist Another concern is float timeout may be inconsistent with other commands.

torwig (dismissed):
LGTM

git-hulk:
@torwig I fixed the sync timeout test case as well in the last commit: https://github.com/apache/incubator-kvrocks/pull/1464/commits/bcc7b8509ab69f432f42c5cdeda43855ea066f4f, please take a look again. After changing the timeout to int, then use the float timeout would return an error.

tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -435,17 +435,15 @@ func TestSlotMigrateSync(t *testing.T) {
...
+		cnt := 200000
 		for i := 0; i < cnt; i++ {
 			require.NoError(t, rdb0.LPush(ctx, util.SlotTable[slot], i).Err())
 		}
-		timeout := 0.001
-
-		require.Nil(t, rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", timeout).Val())
 
-		// check the following command on the same connection
-		require.Equal(t, "PONG", rdb0.Ping(ctx).Val())
```
git-hulk, line 445: This line won' take effect since we're using the connection pool.

torwig: @git-hulk Yes, absolutely.

src/commands/cmd_cluster.cc
```diff
@@ -146,7 +146,14 @@ class CommandClusterX : public Commander {
           sync_migrate_ = true;
 
           if (args.size() == 6) {
-            sync_migrate_timeout_ = GET_OR_RET(ParseFloat<float>(args[5]));
+            auto parse_result = ParseInt<int>(args[5], 10);
+            if (!parse_result) {
+              return {Status::RedisParseErr, "timeout is not an integer or out of range"};
+            }
+            if (*parse_result < 0) {
+              return {Status::RedisParseErr, "timeout should not be negative"};
```
infdahai:
```suggestion
              return {Status::RedisParseErr, errTimeoutIsNegative};
```

tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -418,12 +418,19 @@ func TestSlotMigrateSync(t *testing.T) {
 		require.NoError(t, result.Err())
 		require.Equal(t, "OK", result.Val())
 
-		timeouts := []interface{}{-1, 0, 20, 10.5, -3.14}
-		for _, timeout := range timeouts {
+		invalidTimeouts := []interface{}{-2, -1, "abc"}
```
infdahai: Does it need to add a float argument to show the above dicussion?

infdahai (changes requested):

src/cluster/sync_migrate_context.h
```diff
@@ -25,7 +25,7 @@
 class SyncMigrateContext : private EvbufCallbackBase<SyncMigrateContext, false>,
                            private EventCallbackBase<SyncMigrateContext> {
  public:
-  SyncMigrateContext(Server *svr, redis::Connection *conn, float timeout) : svr_(svr), conn_(conn), timeout_(timeout){};
+  SyncMigrateContext(Server *svr, redis::Connection *conn, int timeout) : svr_(svr), conn_(conn), timeout_(timeout){};
```
infdahai, line 28: BTW, shall we add a float test to gtests with improving code coverage?

tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -413,36 +413,37 @@ func TestSlotMigrateSync(t *testing.T) {
...
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", 10).Val())
-
-		slot++
-		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync", 0.5).Val())
+		// migrate sync without explicitly specify the timeout
+		result := rdb0.Do(ctx, "clusterx", "migrate", slot, id1, "sync")
+		require.NoError(t, result.Err())
+		require.Equal(t, "OK", result.Val())
+
+		invalidTimeouts := []interface{}{-2, -1, "abc"}
```
git-hulk:
```suggestion
		invalidTimeouts := []interface{}{-2, -1, 1.2, "abc"}
```

src/cluster/sync_migrate_context.h
git-hulk: We should do it in util::ParseXXX test cases instead of here.

infdahai: we only use `Parsefloat` now in kvrocks to parse `geo` positions, hash computing, some command with concrete needs, such as `ZADD`, `INCRBYFLOAT`.

torwig (approved):

git-hulk:
Thanks all, merging...

PR #1465, opened 2023-05-22
[WIP] feat: persist the cluster-enabled status in rocksdb

### LINKED ISSUE #1410: Persist the cluster-enabled status in RocksDB
### Search before asking

### Motivation

Currently, we encode the slot id as the prefix of the key if the cluster mode was enabled according to Kvrocks's configuration. But it would go wrong if we start the Kvrocks with `cluster-enabled yes`, but change it to `cluster-enabled no` after restarting.

For the key encoding can see: https://kvrocks.apache.org/community/data-structure-on-rocksdb

### Solution

Persist the `cluster-enabled` configuration in the RocksDB(can use `propagate column family`) when starting the server, then check if the Kvrocks's cluster-enabled configuration matched the database status.

infdahai:
closes #1410


git-hulk:
Done by #2020

PR #1470, opened 2023-05-24
refactor:  register commands based on the order of redis commands

infdahai:
I sort the register commands based on `https://redis.io/commands` to help we find the missing impl commands from redis and develop it.


git-hulk:
How can we tell if a command is in its right place?

infdahai:
I find the command list in`https://kvrocks.apache.org/docs/supported-commands/` is ordered with `redis commands`. So the PR is needless and I close this. @git-hulk

PR #1474, opened 2023-05-25
docker can use env conf

xxaier:
* docker can use env conf
    * add docker/build.sh for build docker in local
    * add docker-compose.yml

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,10 +57,9 @@ VOLUME /var/lib/kvrocks
 
 RUN chown kvrocks:kvrocks /var/lib/kvrocks
 
-COPY ./LICENSE ./NOTICE ./DISCLAIMER ./
+COPY ./docker/run.sh ./LICENSE ./NOTICE ./DISCLAIMER ./
 COPY ./licenses ./licenses
 COPY ./kvrocks.conf /var/lib/kvrocks/
-
 EXPOSE 6666:6666
 
-ENTRYPOINT ["./bin/kvrocks", "-c", "/var/lib/kvrocks/kvrocks.conf", "--dir", "/var/lib/kvrocks", "--pidfile", "/var/run/kvrocks/kvrocks.pid"]
+ENTRYPOINT ["./run.sh"]
--- a/README.md
+++ b/README.md
@@ -103,6 +103,8 @@ $ docker run -it -p 6666:6666 apache/kvrocks
 $ docker run -it -p 6666:6666 apache/kvrocks:nightly
 ```
 
+Can set environment variables `KVROCKS_ARGS` for configuration, see [docker/docker-compose.yml](./docker/docker-compose.yml) for more .
+
 ### Connect Kvrocks service
 
 ```
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+DIR=$(dirname $(realpath "$0"))
+cd $DIR/..
+set -ex
+
+IMG=kvrocks
+
+if [ -v DOCKER_ORG ]; then
+  IMG=$DOCKER_ORG/$IMG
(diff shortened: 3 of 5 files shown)


PragmaTwice (changes requested):
Hi @xxaier, thanks for your contribution.

However, I do not think it is necessary to pass environment variables as config options to kvrocks via some ugly text substitution. If you want to pass some options to kvrocks in docker, you can just try `docker run apache/kvrocks --some-key some-value --other-key other-value ...`.

Besides, we do not prefer to add more shell scripts, which can hardly be general enough.

xxaier:
docker run apache/kvrocks --some-key some-value --other-key other-value ....

will run , but if you write docker-compose.yml and need read env for a .env, how to write ? change cmd ? Such an operation is very unintuitive.

These bash scripts can be run on all platforms with docker and bash installed.

Configuring docker through environment variables I think is a very common practice

for exampe nginx : https://hub.docker.com/_/nginx

image

for example postgres : https://hub.docker.com/_/postgres
image

for example redis : https://hub.docker.com/_/redis
image

PragmaTwice:
I acknowledge your idea to pass env vars as options, but IMHO the shell script composed by `sed` and some other text substitution tools looks terrible to me since it is more like workaround than a real solution.

BTW, shell scripts are hard to maintain and be general-purpose enough. Our development scripts are all written in python, refer to `x.py`.

For [docker/build.sh](https://github.com/apache/incubator-kvrocks/pull/1474/files#diff-efe74abc7213e95246105b4a7b029557ef16ff997e9cc13eadcfffa18e6757c0), I think we do not need the script. You can learn how  our community build and release docker images from https://kvrocks.apache.org/community/create-a-release .

And for [docker/docker-compose.yml](https://github.com/apache/incubator-kvrocks/pull/1474/files#diff-423deb13b7c401b1a7f41ee91c77f722e11d2f317d6a66b546524e8a04cc8b03), I think we do not need an example compose file, and I really do not know why these arguments are chosen and are suitable for various kvrocks users, e.g. `mem_limit: 2G`  `memswap_limit: -1`.

xxaier:
> I acknowledge your idea to pass env vars as options, but IMHO the shell script composed by `sed` and some other text substitution tools looks terrible to me since it is more like workaround than a real solution.
> 
> BTW, shell scripts are hard to maintain and be general-purpose enough. Our development scripts are all written in python, refer to `x.py`.
> 
> For [docker/build.sh](https://github.com/apache/incubator-kvrocks/pull/1474/files#diff-efe74abc7213e95246105b4a7b029557ef16ff997e9cc13eadcfffa18e6757c0), I think we do not need the script. You can learn how our community build and release docker images from https://kvrocks.apache.org/community/create-a-release .
> 
> And for [docker/docker-compose.yml](https://github.com/apache/incubator-kvrocks/pull/1474/files#diff-423deb13b7c401b1a7f41ee91c77f722e11d2f317d6a66b546524e8a04cc8b03), I think we do not need an example compose file, and I really do not know why these arguments are chosen and are suitable for various kvrocks users, e.g. `mem_limit: 2G` `memswap_limit: -1`.

The reason why I don't use python is that in order to keep the size of the final docker container from growing, the final docker image will not contain the python runtime.

The reason I added build.sh is that when developing and debugging docker, there will be a need for local construction, such as building a database based on kvrocks+topling https://github.com/apache/incubator-kvrocks/pull/1206.

I think local development and building docker debugging is also a common scenario, and it is very troublesome to manually enter commands every time.

torwig:
@xxaier Maybe, we can implement a feature to pass options via an env variable similar to `REDIS_ARGS`. However, personally, I don't like the trick with a bash script to 'translate` env variables to Kvrocks options.

PragmaTwice:
> @xxaier Maybe, we can implement a feature to pass options via an env variable similar to `REDIS_ARGS`. However, personally, I don't like the trick with a bash script to 'translate` env variables to Kvrocks options.

+1. `REDIS_ARGS` looks fine to me.

xxaier:
image
I changed code, read command line args from KVROCKS_ARGS

Dockerfile
```diff
@@ -57,10 +57,9 @@ VOLUME /var/lib/kvrocks
...
 
-COPY ./LICENSE ./NOTICE ./DISCLAIMER ./
+COPY ./docker/run.sh ./LICENSE ./NOTICE ./DISCLAIMER ./
 COPY ./licenses ./licenses
 COPY ./kvrocks.conf /var/lib/kvrocks/
-
 EXPOSE 6666:6666
 
-ENTRYPOINT ["./bin/kvrocks", "-c", "/var/lib/kvrocks/kvrocks.conf", "--dir", "/var/lib/kvrocks", "--pidfile", "/var/run/kvrocks/kvrocks.pid"]
+ENTRYPOINT ["./run.sh"]
```
git-hulk, line 65: Can we only add the KVROCKS_ARGS for the custom configuration and leave the others as they should(remove the run.sh and docker-compose.yml)?

xxaier: see https://stackoverflow.com/questions/37904682/how-do-i-use-docker-environment-variable-in-entrypoint-array
hard to do this

xxaier: see https://stackoverflow.com/questions/37904682/how-do-i-use-docker-environment-variable-in-entrypoint-array
hard to do this

git-hulk:
Close the PR due to inactive, feel free to reopen if you like.

PR #1476, opened 2023-05-27
Escape the special chars in the monitor command output

w4096:
## fix the format error in monitor response

If the payload of a request has `\r\n`, the response forwarded to monitoring client is in bad format since simple string can't contains CRLF in RESP.

I found this issue when I execute monitor command in redis-cli and set content of a  `.c` file as value in another redis-cli:

```c
$ redis-cli -p 6666
127.0.0.1:6666> monitor
OK
1685165694.988910 [__namespace 127.0.0.1:55614] "set" "a" "#include <stdio.h>
Error: Protocol error, got "\r" as reply type byte
```

```
$ redis-cli -p 6666 -x set a < ./main.c
```

In the redis code the monitor response sending to client has been escaped by `sdscatrepr`:

https://github.com/redis/redis/blob/e775b34e813654ead5be899faa065f1c31753040/src/replication.c#L560
https://github.com/redis/redis/blob/e775b34e813654ead5be899faa065f1c31753040/src/sds.c#L986

I implement a function `StringRepr` in `string_util` to do the escap which did the same escape with `sdscatrepr`. 

## improve the performance of monitor

the content of monitor response is formated in worker, if more than one worker exists, the content will be formated several times.

We can format the response in `Server::FeedMonitorConns` and pass the response to `Worker::FeedMonitorConns`.

### DIFF
--- a/src/common/string_util.cc
+++ b/src/common/string_util.cc
@@ -20,8 +20,9 @@
 
 #include "string_util.h"
 
-#include <string>
 #include <fmt/format.h>
+
+#include <string>
 
 #include "parse_util.h"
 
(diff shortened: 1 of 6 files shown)


git-hulk:
@wy-ei Cool, thanks

PragmaTwice:
IMHO I think maybe we should just use bulk string reply rather than simple string to solve this problem, and for me these string escaping in this patch is just like a workaround.

PragmaTwice:
https://redis.io/commands/monitor/

Seems redis itself already has some bad practice in solving these escaping problem, which is sad since we need to be compatible to redis.

For me, escaping is not need at all, we can just send an arrays with bulk strings, all in RESP.

w4096:
@PragmaTwice  If redis use bulk string as the monitor response format, this issue does't exists any more. But the simple string is used and the issue is exists. This change is a workaround and it has fix the bad case and make it compatible with redis.

It's easy to change the format as bulk string or array, but client SDK may not work.

For example, in redis-py, unescaping is applyed on monitor response:

https://github.com/redis/redis-py/blob/2d9b5ac6fe03fdc572b8ca47f7134082bae2a5e2/redis/client.py#L1337

src/common/string_util.cc
```diff
@@ -312,4 +314,47 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value) {
   return tokens;
 }
 
+/* convert string to an escaped representation where
+ * all the non-printable characters (tested with isprint()) are turned into
+ * escapes in the form "\n\r\a...." or "\x<hex-number>". */
+std::string StringRepr(const std::string &s) {
```
git-hulk: StringRepr => EscapeString

BTW, can you add the cpp test case for this function?

src/common/string_util.cc
```diff
@@ -312,4 +314,47 @@ std::vector<std::string> TokenizeRedisProtocol(const std::string &value) {
...
+  std::string str;
+  str.reserve(s.size());
+
+  for (auto ch : s) {
+    switch (ch) {
+      case '\\':
+      case '"':
+        str += "\\";
```
mapleFU, line 328: Would '\' become "\\\"?

src/common/string_util.cc
w4096: @git-hulk  function name was changed, and test case was added.

git-hulk: Thanks

git-hulk (dismissed):

PragmaTwice (dismissed):

torwig (approved):
LGTM

git-hulk (approved):

git-hulk:
Thanks all, merging...

PR #1479, opened 2023-05-30
Fix all weird out parameters `int *ret` in the codebase 

jihuayu:
I have followed the rules below and completed all the modifications.

https://github.com/apache/incubator-kvrocks/issues/1451#issuecomment-1565481854

> ### The files match `types/redis_*.h`
> 
> - For the function similar to [`Set#IsMember`](https://github.com/apache/incubator-kvrocks/blob/unstable/src/types/redis_set.h#L36) or [`String#SetNX`](https://github.com/apache/incubator-kvrocks/blob/unstable/src/types/redis_string.h#L47) that `ret` means the flag, I will change it to `bool *flag`
> 
> - For the function similar to [`Set#Remove`](https://github.com/apache/incubator-kvrocks/blob/unstable/src/types/redis_set.h#L39) that `ret` means the number of removed elements, I will change it to `uint64_t *removed_cnt`
> 
> - For the function similar to [`List#Size`](https://github.com/apache/incubator-kvrocks/blob/unstable/src/types/redis_list.h#L36) that `ret` means the size of list, I will change it to `uint64_t *size`
> 
> - For the function similar to [`List#Size`](https://github.com/apache/incubator-kvrocks/blob/unstable/src/types/redis_list.h#L36) that `ret` means the size of list, I will change it to `uint64_t *size`
> 
> - For the function similar to [`String#IncrBy`](https://github.com/apache/incubator-kvrocks/blob/unstable/src/types/redis_string.h#L50) that `ret` means the new value, I will change it to `int64_t *new_value`
> 
> - In cases where `ret` may be negative, I will set the type to `int*`.
> 
> ### The files match `commands/cmd_*.cc`
> 
> I will change the type of `int ret` to the correct type, but I will **not change the variable name** because it represents the return value of the Redis command.
> 
> ### The test files
> 
> I will change the type of `int ret` to the correct type, and I will decide whether to modify the variable name based on the situation.
> 
> 
> ### All files
> 
> I will remove all unnecessary `static_cast` statements.
> 
> **I will use `uint64_t`([`Metadata#size`](https://github.com/apache/incubator-kvrocks/blob/unstable/src/storage/redis_metadata.h#L121) type) as the type for length everywhere, instead of the `size_t`.**
>

### DIFF
--- a/src/commands/cmd_geo.cc
+++ b/src/commands/cmd_geo.cc
@@ -106,7 +106,7 @@ class CommandGeoAdd : public CommandGeoBase {
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    int ret = 0;
+    uint64_t ret = 0;
     redis::Geo geo_db(svr->storage, conn->GetNamespace());
     auto s = geo_db.Add(args_[1], &geo_points_, &ret);
     if (!s.ok()) {
--- a/src/commands/cmd_hash.cc
+++ b/src/commands/cmd_hash.cc
@@ -55,7 +55,7 @@ class CommandHSetNX : public Commander {
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    int ret = 0;
+    uint64_t ret = 0;
     redis::Hash hash_db(svr->storage, conn->GetNamespace());
     auto s = hash_db.MSet(args_[1], field_values_, true, &ret);
     if (!s.ok()) {
@@ -123,7 +123,7 @@ class CommandHExists : public Commander {
 class CommandHLen : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    uint32_t count = 0;
+    uint64_t count = 0;
     redis::Hash hash_db(svr->storage, conn->GetNamespace());
     auto s = hash_db.Size(args_[1], &count);
     if (!s.ok() && !s.IsNotFound()) {
@@ -229,7 +229,7 @@ class CommandHMSet : public Commander {
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    int ret = 0;
+    uint64_t ret = 0;
     redis::Hash hash_db(svr->storage, conn->GetNamespace());
     auto s = hash_db.MSet(args_[1], field_values_, false, &ret);
     if (!s.ok()) {
--- a/src/commands/cmd_hash.cc
+++ b/src/commands/cmd_hash.cc
@@ -93,7 +93,7 @@ class CommandHDel : public Commander {
       fields.emplace_back(args_[i]);
     }
 
-    int ret = 0;
+    uint64_t ret = 0;
     redis::Hash hash_db(svr->storage, conn->GetNamespace());
     auto s = hash_db.Delete(args_[1], fields, &ret);
     if (!s.ok()) {
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -36,7 +36,7 @@ class CommandPush : public Commander {
       elems.emplace_back(args_[i]);
     }
 
-    int ret = 0;
+    uint64_t ret = 0;
     rocksdb::Status s;
     redis::List list_db(svr->storage, conn->GetNamespace());
     if (create_if_missing_) {
@@ -320,7 +320,7 @@ class CommandLRem : public Commander {
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    int ret = 0;
+    uint64_t ret = 0;
     redis::List list_db(svr->storage, conn->GetNamespace());
     auto s = list_db.Rem(args_[1], count_, args_[3], &ret);
     if (!s.ok() && !s.IsNotFound()) {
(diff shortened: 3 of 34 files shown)


torwig:
@jihuayu In `CommandSIsMember` you changed `int` to `bool` which is good. However, the line `*output = redis::Integer(ret);` means the boolean value is implicitly converted to an integer value. Am I right?

torwig:
@jihuayu Please, have a look at `String::CAD` where `flag` is of type `*int` and inside a function, you assign boolean and integer values to the `flag` variable. The first issue is that it's not consistent. The second one is, again, an implicit conversion between an integer value and a boolean value.

jihuayu:
> @jihuayu Please, have a look at `String::CAD` where `flag` is of type `*int` and inside a function, you assign boolean and integer values to the `flag` variable. The first issue is that it's not consistent. The second one is, again, an implicit conversion between an integer value and a boolean value.

I will fix it.


> @jihuayu In `CommandSIsMember` you changed `int` to `bool` which is good. However, the line `*output = redis::Integer(ret);` means the boolean value is implicitly converted to an integer value. Am I right?

@torwig Thank you for bringing this to my attention. While it is true that `bool` is a [integral types](https://en.cppreference.com/w/cpp/types/is_integral) that can be passed to the `Integer` function, there is no `std::to_string` function for `bool` parameters. This could potentially lead to implicit type conversion.

There are two possible solutions to the current issue:
1. Change `*output = redis::Integer(ret)` to `*output = redis::Integer(static_cast<int>(ret))` in order to prevent implicit type conversion.
2. Add a new `Integer` function to handle cases where the parameter type is `bool`. 
```c++
std::string Integer(bool data) { return Integer(data ? 1 : 0); }
```

Personally, I am inclined towards the second solution.  Do you have any other suggestions?

git-hulk:
@jihuayu How about adding the `redis::Bool` function to convert the boolean value to RESP?

jihuayu:
@git-hulk 
I don't think using `redis::Bool` is appropriate.
The meaning of `redis::Integer` function is to return a Integer RESP, not process a Integer. 
Even though we are passing in a `bool` here, we need to return a Integer RESP, so I think `redis::Integer` function is more appropriate.

torwig:
@jihuayu I think instead of adding a new function we can simply write `*output = redis::Integer(ret ? 1 : 0);` where `ret` is boolean.

jihuayu:
@torwig Sounds good to me too.
There are only 4 places where `ret` is `boolean`. Maybe we don't need to add an new function for these 4 calls. Keeping it simple is also a good idea.
I will change it to `*output = redis::Integer(ret ? 1 : 0);`

jihuayu:
But now there's another issue. If someone passes a bool type arguments to the `redis::Integer` function, our clang-tidy tool won't be able to find it. 
Shouldn't we modify `redis::Integer` function to prevent others from passing bool type arguments?

torwig:
@jihuayu Perhaps, we can add `readability-implicit-bool-conversion` to the `clang-tidy` list of checks with the `AllowPointerConditions` options set to `true`. However, it may be the topic of another PR if other contributors think it's a good idea because adding this check needs refactoring some code.

torwig (approved):
Generally, LGTM.
Good job, @jihuayu !

git-hulk:
> @git-hulk I don't think using `redis::Bool` is appropriate. The meaning of `redis::Integer` function is to return a Integer RESP, not process a Integer. Even though we are passing in a `bool` here, we need to return a Integer RESP, so I think `redis::Integer` function is more appropriate.


@jihuayu We already have `redis::Integer`  function now, what I mean is to add redis::Bool based on the Integer function like below:

```
std::string Bool(b bool) {
  return redis::Integer(b ? 1 : 0)
}
```

PragmaTwice:
> > @git-hulk I don't think using `redis::Bool` is appropriate. The meaning of `redis::Integer` function is to return a Integer RESP, not process a Integer. Even though we are passing in a `bool` here, we need to return a Integer RESP, so I think `redis::Integer` function is more appropriate.
> 
> @jihuayu We already have `redis::Integer` function now, what I mean is to add redis::Bool based on the Integer function like below:
> 
> ```
> std::string Bool(b bool) {
>   return redis::Integer(b ? 1 : 0)
> }
> ```

I think function name `Bool` is not concrete here. The returned string is an integer string, like `1`, `0` rather than `true` `false`.
The current solution LGTM.

git-hulk:
> > > @git-hulk I don't think using `redis::Bool` is appropriate. The meaning of `redis::Integer` function is to return a Integer RESP, not process a Integer. Even though we are passing in a `bool` here, we need to return a Integer RESP, so I think `redis::Integer` function is more appropriate.
> > 
> > 
> > @jihuayu We already have `redis::Integer` function now, what I mean is to add redis::Bool based on the Integer function like below:
> > ```
> > std::string Bool(b bool) {
> >   return redis::Integer(b ? 1 : 0)
> > }
> > ```
> 
> I think function name `Bool` is not concrete here. The returned string is an integer string, like `1`, `0` rather than `true` `false`. The current solution LGTM.

The current solution is good for me. My point is from the reply API, we indeed have the boolean type though it will convert to the integer string in the underly.

jihuayu:
Hello, I have fixed all mentioned above. Is there anything else that needs to be corrected?

git-hulk (approved):

git-hulk:
@jihuayu Thanks for your contribution.

PR #1481, opened 2023-06-01
Fix crash when hitting the perflog command rule

### LINKED ISSUE #1480: process crash when running perflog on macos
### Search before asking

### Version

OS - macos 13.3.1 (a) 
Kvrocks - 2.4.0

### Minimal reproduce step

1.config set profiling-sample-commands set
2.config set profiling-sample-ratio 100
3.config set profiling-sample-record-threshold-ms 0
4.set a 1

Then the process crashed.

shows 'OK'

### What did you see instead?

process crashed

### Anything Else?

_No response_

git-hulk:
The root cause is that we didn't allow the entry when creating 
the unique ptr. So it will be a null pointer when the command 
was hitting the perflog rule.

This closes #1480

### DIFF
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -281,7 +281,7 @@ void Connection::RecordProfilingSampleIfNeed(const std::string &cmd, uint64_t du
   rocksdb::SetPerfLevel(rocksdb::PerfLevel::kDisable);
   if (perf_context.empty()) return;  // request without db operation
 
-  auto entry = std::unique_ptr<PerfEntry>();
+  auto entry = std::make_unique<PerfEntry>();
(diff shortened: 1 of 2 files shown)


PragmaTwice:
Good catch! My bad.

torwig (approved):

torwig:
@git-hulk Thanks for reproducing and adding the appropriate test.

PragmaTwice (approved):

git-hulk:
> Good catch! My bad.

No problem. I also didn't catch this when reviewing. The main issue is that we didn't have any test cases for this command.

git-hulk:
Thanks all, merging...

PR #1486, opened 2023-06-06
Remove global server ptr getter used in scripting

PragmaTwice:
The `GetServer` function is only used in scripting, and actually can be removed since we can store the ptr in lua runtime via lightuserdata.  Hence we remove it in this PR to eliminate the number of (accessible) global variables.


git-hulk (approved):

torwig (approved):

PragmaTwice:
Thanks all. Merging...

PR #1489, opened 2023-06-10
Add Redis-compatible cursors for `SCAN` commands

### LINKED ISSUE #877: bigkeys analyze support
### Search before asking

### Motivation

We all know redis support --bigkeys & --memkeys commands, which can help us analyze bigkeys conveniently. Or we can use rdb generated by redis to parse bigkeys.
```
$ redis-cli --bigkeys

# Scanning the entire keyspace to find biggest keys as well as
# average sizes per key type.  You can use -i 0.01 to sleep 0.01 sec
# per SCAN command (not usually needed).

[00.00%] Biggest string found so far 'key-419' with 3 bytes
[05.14%] Biggest list   found so far 'mylist' with 100004 items
[35.77%] Biggest string found so far 'counter:__rand_int__' with 6 bytes
[73.91%] Biggest hash   found so far 'myobject' with 3 fields
```
However, according to my knowledge, kvrocks does not support --bigkeys or --memkeys, because of its different iteration way and other reasons, wshich makes it hard to analyze bigkeys.

### Solution

Maybe you guys can add a --bigkeys(--memkeys) feature like redis or generate some kind of rdb-like file for us to analyze? thanks ; ).

### LINKED ISSUE #1402: `scan` and `scan_iter` stuck to first iteration
### Search before asking

### Version

2.3.0

### Minimal reproduce step

1. Start Kvrocks Docker container:

`docker run -it -p 6666:6666 apache/kvrocks`

2. Populate the database with 200k keys with Python 3.9.13 and the forked version of Redis-py available here : https://github.com/hashlookup/redis-py

```bash
pip install git+https://github.com/hashlookup/redis-py@master
```

```python3
import redis

REDIS = redis.Redis(
    host='0.0.0.0',
    port=6666,
    password='',
    db=0,
    )

# Populate Kvrocks with 200k keys
pipe = REDIS.pipeline()

for i in range(200_000):
    pipe.set(f'dba_{i}', 1)

pipe.execute()

# Test "scan_iter"
for key in scan_iter('dba_*'):
    print(key)


# Test "scan"
cursor = '0'

while True:
    cursor, keys = REDIS.scan(
        cursor=cursor,
        match='dba_*',
    )
    for key in keys:
        print(key)

    if not keys:
        break
```

I expected both commands to iterate through every `dba_*` key.

### What did you see instead?

Instead, both commands are stuck on the first iteration of default size and keep printing the same batch of keys every time.

Exemple with `scan_iter`:

```
b'dba_0'
b'dba_1'
b'dba_10'
b'dba_100'
b'dba_1000'
b'dba_10000'
b'dba_100000'
b'dba_100001'
b'dba_100002'
b'dba_100003'
b'dba_100004'
b'dba_100005'
b'dba_100006'
b'dba_100007'
b'dba_100008'
b'dba_100009'
b'dba_10001'
b'dba_100010'
b'dba_100011'
b'dba_100012'
b'dba_1'
b'dba_10'
b'dba_100'
b'dba_1000'
b'dba_10000'
b'dba_100000'
b'dba_100001'
b'dba_100002'
b'dba_100003'
b'dba_100004'
b'dba_100005'
b'dba_100006'
b'dba_100007'
b'dba_100008'
b'dba_100009'
b'dba_10001'
b'dba_100010'
b'dba_100011'
b'dba_100012'
b'dba_100013'
b'dba_1'
b'dba_10'
b'dba_100'
b'dba_1000'
b'dba_10000'
b'dba_100000'
b'dba_100001'
b'dba_100002'
b'dba_100003'
b'dba_100004'
b'dba_100005'
b'dba_100006'
b'dba_100007'
b'dba_100008'
b'dba_100009'
b'dba_10001'
b'dba_100010'
b'dba_100011'
b'dba_100012'
b'dba_100013'
```

etc.

### Anything Else?

While being inefficient, `keys` doesn't suffer from this problem and return every key matching the pattern:

```python3
len(REDIS.keys('dba_*'))

200000
```

jihuayu:
Hello everyone, I have completed this PR and below is my explanation of the changes made.

We have added the following steps to the processing of the `SCAN`, `ZSCAN`, `SSCAN`, `HSCAN` commands:
- Before processing the command, convert the `numeric cursor` sent by the client back to the `keyname cursor`.
- After processing the command, convert the `keyname cursor` back to a `numeric cursor` and return it to the client, And store the conversion dictionary in the `Server#cursor_dict_`.

Since those steps is non-intrusive to the internal implementation of the `SCAN` command, we can ensure that the behavior of commands such as `ZSCAN`, `SSCAN`, and `HSCAN` is consistent with that of the `SCAN` command.
In the following, I will only use the `SCAN` command as an example to describe the new design.

### Cursor design

We call the new cursor the `numeric cursor` and the old one the `keyname cursor`. 
The numeric cursor is composed of 3 parts: 
- 1-16 bit is counter，and the first bit always 1
- 17-32 bit is timestamp
- 33-64 bit is hash of keyname

The `counter` is a 16-bit unsigned integer that is incremented by 1 every time. When the `counter` overflows, it returns to 0 because it is an unsigned number. Since our `cursor_dict_size` is a power of 2, the `counter` is continuous mod `cursor_dict_size`.
`timestap` is a 16-bit timestamp in seconds, which can store up to 9 hours.
`hash` is a 32-bit hash value of the `keyname`.

### Cursor dictionary(`Server#cursor_dict_`)

The cursor dictionary is an array with a length of 16384(1024*16), which is determined at compile time, and occupies about 640KB of memory. Including the length of the referenced keyname strings, its size is about 1-2M.

During convert the `keyname cursor` back to a `numeric cursor`, a new cursor is generated based on the above rules, and the index for storing the dictionary is determined based on the counter value (index = counter % DICT_SIZE). 

During convert the `numeric cursor` back to a `keyname cursor` , we get the counter from the cursor, and calculate the index of the cursor in the `cursor_dict_` based on the counter. We only need to compare the cursor value of the item at that index with the input cursor value to determine if they are the same.

### Other information about the cursor
This design guarantees the validity of the latest 16384(1024*16) cursors, while cursors that are older or not generated in our system are considered invalid cursors. For invalid cursors, we treat them as a 0 cursor, which means we will start iterating over the collection from the beginning.

Our cursor is globally visible, and we store index information in the cursor. As long as the cursor remains valid, using the same cursor in different connections will produce the same results.

We prevent other users from guessing the data traversed by adjacent cursors by adding the hash value of the `keyname` to the cursor. If a user tries to obtain adjacent cursor information by traversing the hash, the cursor will become an invalid cursor before the traversal is complete because the size of the 32-bit space is much larger than the length of the `cursor_dict_`.

We add a timestamp to the cursor to ensure that the same cursor does not appear within a short period of time before and after a restart.

Other behaviors are consistent with the original SCAN implementation.

### Configuration file
Added `redis-cursor-compatible` configuration item.
If enabled, the cursor will be an unsigned 64-bit integer.
If disabled, the cursor will be a string.

### Test file
Added tests for `redis-cli --bigkey` and `redis-cli --memkeys` commands. We only need to ensure that these commands run correctly, because their correctness is guaranteed by `redis-cli` on the premise that we ensure the correctness of the `scan` command.
Modified the scan command test to test for the cases where `redis-cursor-compatible` is set to `yes` or `no`.

### Other changes:
Fixed a bug where the cursor did not return 0 when `SCAN` commands return less than number elements.

fixed #1402 
fixed #877


git-hulk:
Hi @jihuayu, Thanks for your great efforts. 

Can you explain a bit about the cursor number design and what if we use the wrong cursor for the hash/set/zset scan(should we store and compare the prefix as well when getting it by the cursor)?  And if it's good to use a random or timestamp to avoid conflicts when restarting the server.

PragmaTwice:
Thanks for your contribution!

Several points after a short glance:
- seems the ring buffer is shared between all connections, so just be curious that, will the functionality of `SCAN` get break if we have lots of client connections simultaneously? And could we make it duplicate per connections or able to adjust the buffer size dynamically?
- if some keys are deleted while the `SCAN` operation is executed, will the key properties of `SCAN` be preserved? (refer to "Scan guarantees" in https://redis.io/commands/scan/#scan%20guarantees)
- self-increase cursor may cause informantion leaks: one client can retrieve some informantion of the current executed commands from another client (partially same as the first point)
- seems the mutex can be replaced by rwlock or atomic variables

cc @mapleFU @torwig

git-hulk:
> seems the ring buffer is shared between all connections, so just be curious that, will the functionality of SCAN get break if we have lots of client connections simultaneously? And could we make it duplicate per connections or able to adjust the buffer size dynamically?

@PragmaTwice It cannot be per connection since we need to allow using the cursor in different connections. That's to say, the connection A got the cursor should can be visible to connection B from the client side.

jihuayu:
I will redesign the cursor. 
I'm very busy these days, so I will come back in a few days. Wait for me.

PragmaTwice:
> > seems the ring buffer is shared between all connections, so just be curious that, will the functionality of SCAN get break if we have lots of client connections simultaneously? And could we make it duplicate per connections or able to adjust the buffer size dynamically?
> 
> @PragmaTwice It cannot be per connection since we need to allow using the cursor in different connections. That's to say, the connection A got the cursor should can be visible to connection B from the client side.

Is this property used by some client libraries?

Actually I cannot find how and when these cursors become **invalid** (e.g. Will different connections lead to invalid cursor? Will adding and deleting elements lead to invalid cursor?) in the official document, which seems they just do not give any guarentee about this. It is sad.

But I find such a statement.
> Multiple parallel iterations
> It is possible for an infinite number of clients to iterate the same collection at the same time, as the full state of the iterator is in the cursor, that is obtained and returned to the client at every call. No server side state is taken at all.

It may make the implementation of redis cursor harder, if we need to follow it.

git-hulk:
> Is this property used by some client libraries?

Most libraries should use the same connection to scan, but the cursor needs to be still valid if the connection was reconnected. And as the scan command is a long-time operation, it'd be better to allow reconnecting behavior.

Yes, agreed that it's impossible to follow the Redis exactly since the underlying storage is different. What we can do is to keep compatible as possible.

jihuayu:
@PragmaTwice  @git-hulk  If the user uses `redis-cli scan 0` and `redis-cli scan 10` to scan, the two `scan` will be in different connections. I think this usage may be common, so it is necessary to ensure that the cursor is available in different connection .

We can make a promise to how long or how many recent cursors are guaranteed to be valid. For example, cursors created within the last hour or the last 100 cursors.

In redis doc:
> A full iteration always retrieves all the elements that were present in the collection from the start to the end of a full iteration. This means that if a given element is inside the collection when an iteration is started, and is still there when an iteration terminates, then at some point SCAN returned it to the user.
A full iteration never returns any element that was NOT present in the collection from the start to the end of a full iteration. So if an element was removed before the start of an iteration, and is never added back to the collection for all the time an iteration lasts, SCAN ensures that this element will never be returned.

I think this is necessary, while the others are optional. Our current implementation can ensure these two points.


> Yes, agreed that it's impossible to follow the Redis exactly since the underlying storage is different. What we can do is to keep compatible as possible.

I agree as well. We just need to follow redis in common behavior.

git-hulk:
Thanks, @jihuayu.

Keeping the recent N cursors solution is good for me since those cursors won't occupy too much memory. To see if @PragmaTwice has any comments.

jihuayu:
@git-hulk We may need to know how the N value is determined in actual project groups. If N=1024 and the average key name length is 40, the memory usage would be 1024*(40+40)=80KB.

git-hulk:
> @git-hulk We may need to know how the N value is determined in actual project groups. If N=1024 and the average key name length is 40, the memory usage would be 1024*(40+40)=80KB.

It's hard to determine which value is suitable, but I think we can give a relatively large one like 10240 or even more. It's fine since the usage of memory is a fixed number.

jihuayu:
> It's hard to determine which value is suitable, but I think we can give a relatively large one like 10240 or even more. It's fine since the usage of memory is a fixed number.

Do we need to provide a configuration option for this value?

git-hulk:
> > It's hard to determine which value is suitable, but I think we can give a relatively large one like 10240 or even more. It's fine since the usage of memory is a fixed number.
> 
> Do we need to provide a configuration option for this value?

No from my side, the reason is that users cannot understand how it works.

jihuayu:
@git-hulk @PragmaTwice 
Hello, this PR is ready for review. I have change [top comment](https://github.com/apache/incubator-kvrocks/pull/1489#issue-1750786287) for explanation of the changes . Thanks

kvrocks.conf
```diff
@@ -296,6 +296,12 @@ max-backup-keep-hours 24
 # Default: 16
 max-bitmap-to-string-mb 16
 
+# Whether to enable SCAN, HSCAN, SSCAN, ZSCAN cursor redis compatibility mode. 
+# If enabled, the cursor will be unsigned 64-bit integers.
+# If disable, the cursor will be string.
+# Default: yes
+redis-cursor-compatible yes
```
```suggestion
# If enabled, the cursor will be unsigned 64-bit integers.
# If disabled, the cursor will be a string.
# Default: no
redis-cursor-compatible no
```

src/config/config.cc
```diff
@@ -165,6 +165,7 @@ Config::Config() {
       {"unixsocketperm", true, new OctalField(&unixsocketperm, 0777, 1, INT_MAX)},
       {"log-retention-days", false, new IntField(&log_retention_days, -1, -1, INT_MAX)},
       {"persist-cluster-nodes-enabled", false, new YesNoField(&persist_cluster_nodes_enabled, true)},
+      {"redis-cursor-compatible", false, new YesNoField(&redis_cursor_compatible, true)},
```
git-hulk: The default value should be false to avoid breaking the old behavior.

src/server/server.cc
```diff
@@ -1753,3 +1755,50 @@ std::list<std::pair<std::string, uint32_t>> Server::GetSlaveHostAndPort() {
...
+  auto s = ParseInt<uint64_t>(cursor, 10);
+  // When Cursor 0 or not a Integer return empty string.
+  // Although the parameter 'cursor' is not expected to be 0, we still added a check for 0 to increase the robustness of
+  // the code.
+  if (!s.IsOK() || *s == 0) {
+    return {};
+  }
+  auto cursor_num = *s;
+  // Because the index information is fully stored in the cursor, we can directly obtain the index from the cursor.
+  auto index = (cursor_num >> 48) % CURSOR_DICT_SIZE;
```
mapleFU: Can we encapsulate bits operation to function or a `cursor_id` class?

src/server/server.cc
```diff
@@ -1753,3 +1755,50 @@ std::list<std::pair<std::string, uint32_t>> Server::GetSlaveHostAndPort() {
...
+  return static_cast<uint64_t>(hash) | static_cast<uint64_t>(time_stamp) << 32 | static_cast<uint64_t>(counter) << 48 |
+         static_cast<uint64_t>(1) << 63;
+}
+
+std::string Server::GenerateCursorFromKeyName(const std::string &key_name, const char *prefix) {
+  if (!config_->redis_cursor_compatible) {
+    // add prefix for SCAN
+    return prefix + key_name;
+  }
+  auto counter = cursor_counter_.fetch_add(1);
```
mapleFU, line 1783: So prefix name will not be added in new-cursor?

mapleFU (commented):
Just curious why a hash is required?

git-hulk:
@jihuayu Overall looks good to me. Adding scope for the cursor mapping would be better, then we can check if it's the right type to use the cursor's key name.

For example, we can store the scope = hash in the cursor element, it cannot use this cursor's key if the operation is not HSCAN, and vice versa.

src/server/server.cc
jihuayu: [Prefix will be auto remove in ParseCursor.](https://github.com/apache/incubator-kvrocks/blob/unstable/src/commands/scan_base.h#L57-L64
)  If we add prefix, we will need to remove it again in the `Execute` function, which I think is redundant.
I think prefix is just a readability identifier.

jihuayu:
@mapleFU We add hash to avoid informantion leaks

> We prevent other users from guessing the data traversed by adjacent cursors by adding the hash value of the keyname to the cursor. If a user tries to obtain adjacent cursor information by traversing the hash, the cursor will become an invalid cursor before the traversal is complete because the size of the 32-bit space is much larger than the length of the cursor_dict_.

src/server/server.cc
mapleFU: Currently, `CURSOR_DICT_SIZE` is `2**14`, so highest bit doesn't matter?

jihuayu: CURSOR_DICT_SIZE must less than 2**15, I add `static_assert` in .h file.
https://github.com/apache/incubator-kvrocks/pull/1489/files#diff-a57adfc622093e1a4bedd3c0d7727e6c42b126a015fd5f2c166aed385c2557b3R94

kvrocks.conf
```diff
@@ -296,6 +296,12 @@ max-backup-keep-hours 24
 # Default: 16
 max-bitmap-to-string-mb 16
 
+# Whether to enable SCAN, HSCAN, SSCAN, ZSCAN cursor redis compatibility mode.
```
```suggestion
# Whether to enable SCAN-like cursor compatible with Redis. 
```

src/server/server.h
```diff
@@ -84,6 +88,16 @@ struct ChannelSubscribeNum {
   size_t subscribe_num;
 };
 
+// CURSOR_DICT_SIZE must 2^n and n < 16
```
```suggestion
// CURSOR_DICT_SIZE must be 2^n where n < 16
```

src/server/server.h
```diff
@@ -84,6 +88,16 @@ struct ChannelSubscribeNum {
...
+constexpr const size_t CURSOR_DICT_SIZE = 1024 * 16;
+static_assert((CURSOR_DICT_SIZE & (CURSOR_DICT_SIZE - 1)) == 0, "CURSOR_DICT_SIZE must 2^n and n < 16");
+static_assert(CURSOR_DICT_SIZE < (1 << 15), "CURSOR_DICT_SIZE must 2^n and n < 16");
```
```suggestion
static_assert((CURSOR_DICT_SIZE & (CURSOR_DICT_SIZE - 1)) == 0, "CURSOR_DICT_SIZE must be 2^n");
static_assert(CURSOR_DICT_SIZE < (1 << 15), "CURSOR_DICT_SIZE must be less than 2^16");
```

src/server/server.h
```diff
@@ -319,4 +336,8 @@ class Server {
   std::atomic<size_t> watched_key_size_ = 0;
   std::map<std::string, std::set<redis::Connection *>> watched_key_map_;
   std::shared_mutex watched_key_mutex_;
+
+  // SCAN ring buffer
+  std::atomic<uint16_t> cursor_counter_ = {0};
+  std::array<struct CursorDictElement, CURSOR_DICT_SIZE> cursor_dict_;
```
```suggestion
  std::array<CursorDictElement, CURSOR_DICT_SIZE> cursor_dict_;
```

src/config/config.cc
jihuayu: Wow, I got it. We don't need to make the default configuration file consistent with the default values.

tests/gocase/unit/scan/scan_test.go
```diff
@@ -52,13 +52,27 @@ func TestScanEmptyKey(t *testing.T) {
...
 }
 
-func TestScan(t *testing.T) {
+func TestScanWithNumberCursor(t *testing.T) {
 	srv := util.StartServer(t, map[string]string{})
 	defer srv.Close()
+	ctx := context.Background()
+	rdb := srv.NewClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
+	require.NoError(t, rdb.ConfigSet(ctx, "redis-cursor-compatible", "yes").Err())
```
PragmaTwice, line 61: I think `redis-cursor-compatible` set to `yes` is a default behavior. Could you add test for  `redis-cursor-compatible` set to `no`? (e.g. ScanTest works well and `--bigkeys` fails.)

jihuayu: I don't think the failure of `--bigkeys`  is something that needs to be tested. We don't care about whether  `--bigkeys` 
 runs successfully or not when `redis-cursor-compatible` is set to `no`. Success is best, but failure is okay too.



The function [`TestScanWithStringCursor`](https://github.com/apache/incubator-kvrocks/pull/1489/files/a36c1412fb25dc283589bcc2c1169f90b9724b35#diff-07aeba1404d8a74294e03c47a257b71cf973c8e8aad8950f2f630e14b5cf1455R65-R73) shows when `redis-cursor-compatible` set to `no` ScanTest works well

I think I can remove `require.NoError(t, rdb.ConfigSet(ctx, "redis-cursor-compatible", "yes").Err())`.

PragmaTwice: Got it.

src/config/config.cc
git-hulk: Yes, it will break the old users if they upgraded the Kvrocks without adding the new configuration.

src/server/server.cc
jihuayu: > Can we encapsulate bits operation to function or a `cursor_id` class?

It is a good idea. I add a new function `GetIndexFromNumberCursor` which can convert `number_cursor` to `index`

src/server/server.cc
```diff
@@ -1753,3 +1755,52 @@ std::list<std::pair<std::string, uint32_t>> Server::GetSlaveHostAndPort() {
...
+// prevent the generation of the same cursor in the extremely short period before and after a restart.
+static uint64_t GetNumberCursor(const std::string &key_name, uint16_t counter) {
+  auto hash = static_cast<uint32_t>(std::hash<std::string>{}(key_name));
+  auto time_stamp = static_cast<uint16_t>(util::GetTimeStamp());
+  // set first bit 1, so number cursor not be 0
+  return static_cast<uint64_t>(hash) | static_cast<uint64_t>(time_stamp) << 32 | static_cast<uint64_t>(counter) << 48 |
+         static_cast<uint64_t>(1) << 63;
+}
+
+static size_t GetIndexFromNumberCursor(uint64_t number_cursor) { return (number_cursor >> 48) % CURSOR_DICT_SIZE; }
```
infdahai: hi @jihuayu can I ask why the counter puts the high bit? I mean maybe we don't need to move the bit here.

src/server/server.cc
```diff
@@ -1753,3 +1755,52 @@ std::list<std::pair<std::string, uint32_t>> Server::GetSlaveHostAndPort() {
...
+
+// The numeric cursor consists of a 32-bit hash, a 16-bit time stamp, and a 16-bit counter, with the highest bit set to
+// 1 to prevent a zero cursor from occurring. The hash is used to prevent information leakage. The time_stamp is used to
+// prevent the generation of the same cursor in the extremely short period before and after a restart.
+static uint64_t GetNumberCursor(const std::string &key_name, uint16_t counter) {
+  auto hash = static_cast<uint32_t>(std::hash<std::string>{}(key_name));
+  auto time_stamp = static_cast<uint16_t>(util::GetTimeStamp());
+  // set first bit 1, so number cursor not be 0
+  return static_cast<uint64_t>(hash) | static_cast<uint64_t>(time_stamp) << 32 | static_cast<uint64_t>(counter) << 48 |
+         static_cast<uint64_t>(1) << 63;
```
infdahai: And can we set the first bit to 1? Can we skip this if the counter is zero ?

src/server/server.cc
jihuayu: You are right, it's better to put it in the lower bit.

src/server/server.cc
jihuayu: I don't like skip zero counter, as it will make the code more complicated.
Maybe we can set the first bit of the `hash` to 1 instead of `counter`.

infdahai: 👌I get it. Just asking.

infdahai: I'm just wondering you use GetIndexFromNumberCursor() to generate the index of the array `cursor_dict_`. It seems we will ignore the high bit of the index because we module the num by `CURSOR_DICT_SIZE`. So the key point is that we need the `1` to block the `numeric cursor` from 0 and we will reduce the chance of confilcts if we put the bit to the counter. Am I right?

jihuayu: That's right. However, in reality, the difference is not very big, and each has its own advantages and disadvantages. 
For me, it makes no difference where `1` is placed.

infdahai: I get it.

src/storage/redis_db.cc
```diff
@@ -277,9 +277,8 @@ rocksdb::Status Database::Scan(const std::string &cursor, uint64_t limit, const
       keys->emplace_back(user_key);
       cnt++;
     }
-
     if (!storage_->IsSlotIdEncoded() || prefix.empty()) {
-      if (!keys->empty()) {
+      if (!keys->empty() && cnt >= limit) {
```
mapleFU, line 281: I don't fully understand why a `cnt >= limit` is required?

src/server/server.cc
mapleFU: Using counter is zero to marking means that when it overflow, the number will back to 1, I guess when ring size is 2**14, there are only slightly differences...

src/storage/redis_db.cc
jihuayu: This code fix other bug. 

> Fixed a bug where the cursor did not return 0 when SCAN commands return less than number elements.

jihuayu:
I modified the format of the cursor.  https://github.com/apache/incubator-kvrocks/pull/1489/commits/47fb25a63b4f2c207f9544cb0089e35477778d93
- Placed the `counter` in the lowest bit 
- set the highest bit of the `hash` to `1`. 
- Modified the content of the `static_assert` from `static_assert(CURSOR_DICT_SIZE < (1 << 15),...);` to `static_assert(CURSOR_DICT_SIZE <= (1 << 15),...);`


@infdahai @mapleFU

src/server/server.h
```diff
@@ -84,6 +88,24 @@ struct ChannelSubscribeNum {
...
+enum CursorType {
+  kTypeBase,  // cursor for SCAN
+  kTypeHash,  // cursor for HSCAN
+  kTypeSet,   // cursor for SSCAN
+  kTypeZSet,  // cursor for ZSCAN
+};
+
+struct CursorDictElement {
+  uint64_t cursor;
+  CursorType cursor_type;
```
PragmaTwice: Could this also be encoded to the cursor? Seems we can take about 4 bit out of 32bit hash for it.

src/storage/redis_db.cc
mapleFU: Great! Would you mind add a testing for this?

mapleFU:
So after that the hashing is 31bits?

PragmaTwice:
> So after that the hashing is 31bits?

Seems right, and [with my idea](https://github.com/apache/incubator-kvrocks/pull/1489#discussion_r1234790169) hash can be 27bit.

jihuayu:
> So after that the hashing is 31bits?

@mapleFU 
Yes, now hashing is 31bits.

> Could this also be encoded to the cursor? Seems we can take about 4 bit out of 32bit hash for it.

If we encoded `cursor_type` to the cursor, hash will only have 27 bits left.

Poor hash, it's being squeezed for space by everyone.

PragmaTwice:
> If we encoded cursor_type to the cursor, hash will only have 27 bits left.

But I think the hash value here is not an important thing, it is just used as [check digits](https://en.wikipedia.org/wiki/Check_digit), and actually the command type can also be regarded as check digits.

Seems a 16bit hash is already enough here for me.

mapleFU (commented):
This LGTM. Some advices:
1. Can `item.cursor == number_cursor && item.cursor_type == cursor_type` be a "match" function in `CursorDictElement`?
2. Can cursor be encapsulated to avoid explicit bit-ops outside?
3. Do we need test bugfix?

And notice that `cursor_dict_`'s cursor and other element is not atomic, if there are huge number of scan, it might produce bad result or concurrent read/write `key_name`. But maybe it's ok not to handle this?

jihuayu:
> And notice that cursor_dict_'s cursor and other element is not atomic, if there are huge number of scan, it might produce bad result or concurrent read/write key_name. But maybe it's ok not to handle this?

This will not cause errors. Firstly, the assignment operation is atomic, so there will be no problem with the correspondence between the cursor and the key name. Secondly, the network transmission speed is much slower than the synchronization of cache data, so there will be no inconsistency in the cache.

> Can cursor be encapsulated to avoid explicit bit-ops outside?

Our cursor is becoming increasingly complex, and it can indeed be encapsulated separately.

> Do we need test bugfix?

If anyone feels that additional testing is needed, they can raise it. I will go to add it.

mapleFU:
> This will not cause errors. Firstly, the assignment operation is atomic, so there will be no problem with the correspondence between the cursor and the key name. 

Assume that scan1 allocate at slot 1000, and there are 2^14 scans in-flight, scan 2^14 will also allocate 1000. But I guess that's not problem

jihuayu:
> Assume that scan1 allocate at slot 1000, and there are 2^14 scans in-flight, scan 2^14 will also allocate 1000. But I guess that's not problem

I also thinks it is not problem.

git-hulk:
This PR generally looks good to me, to see if others have further comments.

torwig (commented):
@jihuayu Thank you for the great explanation.
Generally, LGTM.

src/server/server.h
```diff
@@ -319,4 +359,8 @@ class Server {
   std::atomic<size_t> watched_key_size_ = 0;
   std::map<std::string, std::set<redis::Connection *>> watched_key_map_;
   std::shared_mutex watched_key_mutex_;
+
+  // SCAN ring buffer
+  std::atomic<uint16_t> cursor_counter_ = {0};
+  std::array<CursorDictElement, CURSOR_DICT_SIZE> cursor_dict_;
```
mapleFU: @PragmaTwice would a `sizeof(CursorDictElement) * CURSOR_DICT_SIZE` dict in `Server` makes it too large? And If `Server` is large, would it matters?

src/server/server.h
```diff
@@ -84,6 +88,39 @@ struct ChannelSubscribeNum {
...
+};
+struct CursorDictElement;
+
+class NumberCursor {
+ public:
+  NumberCursor() = default;
+  explicit NumberCursor(CursorType cursor_type, uint16_t counter, const std::string &key_name);
+  explicit NumberCursor(uint64_t number_cursor) : cursor_(number_cursor) {}
+  size_t GetIndex() { return cursor_ % CURSOR_DICT_SIZE; }
+  bool IsMatch(const CursorDictElement &element, CursorType cursor_type);
```
mapleFU: better mark these as `const`?

src/server/server.h
```diff
@@ -84,6 +88,39 @@ struct ChannelSubscribeNum {
   size_t subscribe_num;
 };
 
+// CURSOR_DICT_SIZE must be 2^n where n <= 16
+constexpr const size_t CURSOR_DICT_SIZE = 1024 * 16;
+static_assert((CURSOR_DICT_SIZE & (CURSOR_DICT_SIZE - 1)) == 0, "CURSOR_DICT_SIZE must be 2^n");
+static_assert(CURSOR_DICT_SIZE <= (1 << 15), "CURSOR_DICT_SIZE must be less than or equal to 2^16");
+
+enum CursorType {
```
mapleFU:
```c++
enum class CursorType: uint8_t {};
```

would `enum class` helps?

src/server/server.h
```diff
@@ -84,6 +88,39 @@ struct ChannelSubscribeNum {
   size_t subscribe_num;
 };
 
+// CURSOR_DICT_SIZE must be 2^n where n <= 16
+constexpr const size_t CURSOR_DICT_SIZE = 1024 * 16;
+static_assert((CURSOR_DICT_SIZE & (CURSOR_DICT_SIZE - 1)) == 0, "CURSOR_DICT_SIZE must be 2^n");
+static_assert(CURSOR_DICT_SIZE <= (1 << 15), "CURSOR_DICT_SIZE must be less than or equal to 2^16");
```
mapleFU: Emm is the comment mismatches the code?

src/server/server.h
jihuayu: Good advice!

src/server/server.h
jihuayu: maybe not? `1 << 15` = `2^16`

src/server/server.h
jihuayu: `sizeof(cursor_dict_)` =  40 * 1024 * 16 = 640KB.
Are there any issues with a large `Server`?

infdahai: use a unordered_map?

jihuayu: 640KB is small enough for several GB servers, and we don't really need to save this memory. 
The biggest advantage of `arrays` is that they are simple enough to use in multi-threaded environments.

mapleFU: @jihuayu I guess 640KB is small, using an sequence of memory is ok. But I don't know do we need to copy or do something with `Server`, if we need, the `Server` would become an large object. So I ask for Twice's idea

src/server/server.h
mapleFU: Emmm, did I make it wrong?

```python
>>> 1 << 15
32768
>>> 2**15
32768
>>> 1 << 16
65536
```

src/server/server.h
PragmaTwice: Seems the global server object is allocated directly on the stack rather than heap, so I think it is better to put the array on heap, e.g. `std::vector` or `std::unique_ptr<T[]>`. I think we need to prevent from taking too many space of the stack.

src/server/server.h
jihuayu: Oh, you're right. 
I confused myself.

src/server/server.h
```diff
@@ -84,6 +88,39 @@ struct ChannelSubscribeNum {
   size_t subscribe_num;
 };
 
+// CURSOR_DICT_SIZE must be 2^n where n < 16
```
mapleFU: This is conflict with line 94?

jihuayu: Sorry, I've confused the maximum values of uint16_t, index, and size again.
CURSOR_DICT_SIZE is the size, which can be larger than the maximum value of uint16_t.

Thank you for your patience.

src/server/server.cc
```diff
@@ -1753,3 +1758,51 @@ std::list<std::pair<std::string, uint32_t>> Server::GetSlaveHostAndPort() {
...
+// the extremely short period before and after a restart.
+NumberCursor::NumberCursor(CursorType cursor_type, uint16_t counter, const std::string &key_name) {
+  auto hash = static_cast<uint32_t>(std::hash<std::string>{}(key_name));
+  auto time_stamp = static_cast<uint16_t>(util::GetTimeStamp());
+  cursor_ = static_cast<uint64_t>(counter) | static_cast<uint64_t>(time_stamp) << 16 |
+            static_cast<uint64_t>(hash) << 32 | static_cast<uint64_t>(cursor_type) << 61;
+}
+
+bool NumberCursor::IsMatch(const CursorDictElement &element, CursorType cursor_type) const {
+  return cursor_ == element.cursor.cursor_ && cursor_type == getCursorType();
```
mapleFU, line 1775: I don't fully understand, isn't `cursorType` already part of `cursor_`?

jihuayu: We need to ensure that the `cursor` generated by the `SCAN` command can only be used by the `SCAN` command and cannot be used by the `HSCAN` or `ZSCAN` commands.

mapleFU: So this prevent from ZSCAN using a SCAN iterator?

jihuayu: yes

mapleFU: (I guess this can be separate into a:
* Scan verifies that the scan type has a right cursor
* Matches only cursor

But this code also LGTM)

mapleFU (approved):
LGTM

src/server/server.h
jihuayu: OK, I have changed `std::array` to `std::unique_ptr<std::array>`

jihuayu:
Thanks for the review. Does anyone have any other suggestions?

tests/gocase/unit/scan/scan_test.go
```diff
@@ -52,13 +52,26 @@ func TestScanEmptyKey(t *testing.T) {
...
 
+func TestScanWithStringCursor(t *testing.T) {
+	srv := util.StartServer(t, map[string]string{})
+	defer srv.Close()
 	ctx := context.Background()
 	rdb := srv.NewClient()
 	defer func() { require.NoError(t, rdb.Close()) }()
+	require.NoError(t, rdb.ConfigSet(ctx, "redis-cursor-compatible", "no").Err())
+	ScanTest(t, rdb, ctx)
+}
```
PragmaTwice: The default value of `redis-cursor-compatible` is `no`. I think the `TestScanWithNumberCursor` does not work as expected. You need to enable `redis-cursor-compatible` before the `ScanTest` call.

jihuayu: Oh, I changed the default value of `redis-cursor-compatible` but forgot to update the test.

git-hulk (dismissed):
LGTM, thanks for your contribution and patience.

git-hulk:
@torwig @PragmaTwice Can you take a look again while you're free.

src/config/config.h
```diff
@@ -144,6 +144,7 @@ struct Config {
   int pipeline_size;
   int sequence_gap;
 
+  bool redis_cursor_compatible = true;
```
torwig: I just wondering why the initial value is `true`? As I see in the `.cc` file we have the default value as `false` for this variable.
Other LGTM.

torwig (dismissed):

git-hulk (approved):

src/config/config.h
jihuayu: Fixed

torwig (approved):

git-hulk:
Thanks all, merging...

qingping209:
very neat design !

PR #1490, opened 2023-06-10
Add the support of the BZMPOP command

### LINKED ISSUE #1445: Add the support of the `BZMPOP` command
### Search before asking

### Motivation

 This issue considers to support [`bzmpop`](https://redis.io/commands/bzmpop/), `bzpopmax`, `bzpopmin` commands.

And the implementation should process a `timeout` argument. 

### Solution

- [ ] bzmpop
- [ ] bzpopmax
- [ ] bzpopmin

Yangsx-1:
Close #1445 
This PR add supports of [bzmpop](https://redis.io/commands/bzmpop/), [bzpopmax](https://redis.io/commands/bzpopmax/), [bzpopmin](https://redis.io/commands/bzpopmin/) command like Redis.
And I also add some tests for these three commands.
(It's my first PR for kvrocks, there maybe something i didn't recognize. I'm very glad to revise it if there is something wrong.)


torwig:
@Yangsx-1 You can use the `./x.py format` command before committing the code to avoid errors during CI.

Yangsx-1:
> @Yangsx-1 You can use the `./x.py format` command before committing the code to avoid errors during CI.

Thanks, fixed.

Yangsx-1:
It seems like other things wrong? And it didn't happen in my forked repo's [workflow](https://github.com/Yangsx-1/incubator-kvrocks/actions/runs/5234167251).

PragmaTwice (commented):
Thanks for your contribution!

I notice there are some duplicated code segments, could you try to refactor to avoid these 
trivially repeated code?

Yangsx-1:
> 
> I notice there are some duplicated code segments, could you try to refactor to avoid these trivially repeated code?

OK, i'll try, but it may need some time, i'm a little busy these days.

PragmaTwice:
>  It seems like other things wrong? And it didn't happen in my forked repo's [workflow](https://github.com/Yangsx-1/incubator-kvrocks/actions/runs/5234167251).

It looks to be a false positive of ThreadSanitizer about TBB. Dont worry.

Yangsx-1:
> 
> I notice there are some duplicated code segments, could you try to refactor to avoid these trivially repeated code?

Refactored.

src/commands/cmd_zset.cc
```diff
@@ -220,64 +222,211 @@ class CommandZLexCount : public Commander {
...
+                    private EventCallbackBase<CommandMPop> {
+ public:
+  explicit CommandMPop(bool block) : block_(block) {}
 
   Status Parse(const std::vector<std::string> &args) override {
     CommandParser parser(args, 1);
+    if (block_) {
+      timeout_ = GET_OR_RET(parser.TakeInt<int>(NumericRange<int>{0, std::numeric_limits<int>::max()}));
+      if (timeout_ < 0) {
+        return {Status::RedisParseErr, "timeout should not be negative"};
```
torwig: It's not critical, but in this case `Redis` returns `timeout is negative`. 
I prefer to launch `redis-server` and `redis-cli` locally to test such cases and see the exact error messages.

src/commands/cmd_zset.cc
```diff
@@ -301,40 +450,149 @@ class CommandZMPop : public Commander {
...
-    redis::ZSet zset_db(svr->storage, conn->GetNamespace());
+    svr_ = svr;
+    conn_ = conn;
+
+    if (!block_) {
+      auto s = TryPopFromZset();
+      return Status::OK();
+    }
+
+    auto bev = conn->GetBufferEvent();
```
torwig: You can move this line closer to the place where `bev` is used, e.g. before the `SetCB(bev);` line.

src/commands/cmd_zset.cc
```diff
@@ -301,40 +450,149 @@ class CommandZMPop : public Commander {
...
+    if (output.empty() && !block_) {
+      output = redis::NilString();
+    }
+    conn_->Reply(output);
+    return s;
+  }
+
+  void OnWrite(bufferevent *bev) {
+    auto s = TryPopFromZset();
+    if (!s.ok() || !reply_flag_) {
```
torwig: In the case of an error inside `TryPopFromZset`, the function by itself will send an error and then the connection allegedly will stay blocked. Am I right?

src/commands/cmd_zset.cc
```diff
@@ -301,40 +450,149 @@ class CommandZMPop : public Commander {
...
+    if (timeout_) {
+      timer_.reset(NewTimer(bufferevent_get_base(bev)));
+      timeval tm = {timeout_, 0};
+      evtimer_add(timer_.get(), &tm);
+    }
+
+    return {Status::BlockingCmd};
+  }
+
+  rocksdb::Status TryPopFromZset() {
```
torwig: Actually, the function pops from multiple sorted sets which is not expressed by its name. And, except for popping elements, it sends a reply. It's better to separate concerns: pop and return elements in one function and then analyze (if something was popped/if the mode is blocking/etc) and send the response in another function. Let the function do just one thing.

src/commands/cmd_zset.cc
```diff
@@ -301,40 +450,149 @@ class CommandZMPop : public Commander {
...
+    redis::ZSet zset_db(svr_->storage, conn_->GetNamespace());
+    rocksdb::Status s;
+    std::string output;
     for (auto &user_key : keys_) {
       std::vector<MemberScore> member_scores;
-      auto s = zset_db.Pop(user_key, count_, flag_ == ZSET_MIN, &member_scores);
+      s = zset_db.Pop(user_key, count_, flag_ == ZSET_MIN, &member_scores);
       if (!s.ok()) {
-        return {Status::RedisExecErr, s.ToString()};
+        conn_->Reply(redis::Error("ERR " + s.ToString()));
```
torwig: In case of an error inside `Pop` you will reply with the error message and then just `break` which only interrupts the `for` loop. And later there is a chance that you will reply with a nil-string or some data that will be in the `output`.

src/commands/cmd_zset.cc
```diff
@@ -220,64 +222,211 @@ class CommandZLexCount : public Commander {
...
+        auto parse_result = ParseInt<int>(args[2], 10);
+        if (!parse_result) {
+          return {Status::RedisParseErr, errValueNotInteger};
+        }
+
+        count_ = *parse_result;
+      }
+      keys_.push_back(args[1]);
+      return Commander::Parse(args);
+    } else {
```
torwig: Here you can write without `else` since you have `return` in the `if` part - it saves some indentation.

src/commands/cmd_zset.cc
```diff
@@ -220,64 +222,211 @@ class CommandZLexCount : public Commander {
...
-    redis::ZSet zset_db(svr->storage, conn->GetNamespace());
-    std::vector<MemberScore> member_scores;
-    auto s = zset_db.Pop(args_[1], count_, min_, &member_scores);
-    if (!s.ok()) {
-      return {Status::RedisExecErr, s.ToString()};
+    svr_ = svr;
+    conn_ = conn;
+
+    if (!block_) {
+      auto s = TryPopFromZset();
```
torwig: If `!block_` you do `TryPopFromZset` and if `block_`  actually you do `TryPopFromZset` as well (identical actions inside the `if` and the `else` branches). 
I'm sure you will refactor this after you refactor/split the `TryPopFromZset` function.

src/commands/cmd_zset.cc
```diff
@@ -220,64 +222,211 @@ class CommandZLexCount : public Commander {
...
+    if (timeout_) {
+      timer_.reset(NewTimer(bufferevent_get_base(bev)));
+      timeval tm = {timeout_, 0};
+      evtimer_add(timer_.get(), &tm);
+    }
+
+    return {Status::BlockingCmd};
+  }
+
+  rocksdb::Status TryPopFromZset() {
```
torwig: I saw this function twice in this file. Perhaps, something could be done to remove/minimize duplication.

src/commands/cmd_zset.cc
Yangsx-1: Yes, you are right. There shouldn't be a `!s.ok()`. I forgot to delete it.

src/commands/cmd_zset.cc
Yangsx-1: I can split the func `TryPopFromZset`, but it is a little difficult or tedious to merge `BZMPOP` and `BZPOPMIN/MAX` together. The ways they parse arguments and reply messages are so different. The only same thing for them in the func `TryPopFromZset` is they pop from the database, although the name are same, maybe it's a bad name. :(

torwig: @Yangsx-1 Yes, you can have different functions if the logic inside is different. But try to separate retrieving the data from the storage and sending the response - maybe it reveals some possibilities of deduplication.

src/commands/cmd_zset.cc
infdahai: You can use `errTimeoutIsNegative` in error_constants.h

src/commands/cmd_zset.cc
Yangsx-1: Is there anything i need to fix? :)

torwig: @Yangsx-1 Thank you for the changes you've made. I'll review them later.

src/commands/cmd_zset.cc
```diff
@@ -220,64 +222,218 @@ class CommandZLexCount : public Commander {
...
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    redis::ZSet zset_db(svr->storage, conn->GetNamespace());
-    std::vector<MemberScore> member_scores;
-    auto s = zset_db.Pop(args_[1], count_, min_, &member_scores);
+    svr_ = svr;
+    conn_ = conn;
+
+    auto s = TryPopFromMultiZset();
```
torwig: Could you simplify the logic like the following:

```
redis::ZSet zset_db(svr_->storage, conn_->GetNamespace());

for (auto &key : keys_) {
  std::vector<MemberScore> member_scores;
  auto s = zset_db.Pop(key, count_, flag_ == ZSET_MIN, &member_scores);
  if (!s.ok()) {
  }

 if (!member_scores.empty()) {
    ReplyWithMemberScores(member_scores); // construct output and send it to conn_
    return Status::OK();
  }
}

// no elements found in all keys
if (!block_ || conn->IsInExec()) {
    // reply with nil array
    return Status::OK();
}

for (const auto &key : keys_) {
  svr_->BlockOnKey(key, conn_);
}
    
 // the rest of the `Execute` code with `bev` and setting a timer.
```

src/commands/cmd_zset.cc
```diff
@@ -220,64 +222,218 @@ class CommandZLexCount : public Commander {
...
  private:
   bool min_;
+  bool block_;
   int count_ = 1;
+  int timeout_;
+  std::vector<std::string> keys_;
+  Server *svr_ = nullptr;
+  Connection *conn_ = nullptr;
+  UniqueEvent timer_;
+  std::string user_key_;
```
torwig: And you can delete `user_key_` and `member_scores_` as the command class members.

src/commands/cmd_zset.cc
```diff
@@ -220,64 +222,218 @@ class CommandZLexCount : public Commander {
   RangeLexSpec spec_;
 };
 
-class CommandZPop : public Commander {
+class CommandZPop : public Commander,
+                    private EvbufCallbackBase<CommandZPop, false>,
+                    private EventCallbackBase<CommandZPop> {
  public:
-  explicit CommandZPop(bool min) : min_(min) {}
+  explicit CommandZPop(bool min, bool block) : min_(min), block_(block) {}
```
torwig: BTW, did you add also `BZPOPMAX` and `BZPOPMIN`?

torwig (changes requested):

src/commands/cmd_zset.cc
Yangsx-1: The task i do is to implement bzpopmax/bzpopmin, in order to deduplicate, i merge the code of zpopmax/zpopmin together.

src/commands/cmd_zset.cc
Yangsx-1: The function `TryPopFromMultiZset()` is also used in function `OnWrite` later, i think expanding the code here rather than using a function will cause duplication.

Yangsx-1: There may be some complexity because i merged zpopmin/zpopmax/bzpopmin/bzpopmax commands together.

torwig: @Yangsx-1 I appreciate your effort to remove duplication - it's the right way to write clean code. However, right now, as I mentioned before, the `TryPopFromMultiZset` pops and sends only an error reply and stores popped elements in a class member variable - a lot of actions just for a single function. Maybe, at least, it can return status and then you will check it and take appropriate actions?

Yangsx-1: Ok, i understand your opinion. I'll revise it.

torwig: @Yangsx-1 Personally, I don't like functions with the `Try` prefix. 
Does the function `ReplyForBZPop` have something specific for blocking pop from a sorted set that you named it in this way?

Yangsx-1: I just refer to BLPOP/BRPOP in file cmd_list.cc, so i used the Try prefix. Is there a better way to name it?
The reply of BZPOPMIN/BZPOPMAX has an additional 'key' compared to ZPOPMIN/ZPOPMAX. I think i can merge them together so that they can use the same function.

torwig: @Yangsx-1 Now I see what "inspired" you :) 
You can name something like `TryPopFromMultiZset` -> `PopFromMultipleZsets`, `ReplyForBZPop` -> `SendMembersWithScores`. It would be great if you don't send an error response inside the `PopFromMultipleZsets` function.

src/commands/cmd_zset.cc
```diff
@@ -221,64 +223,217 @@ class CommandZLexCount : public Commander {
...
```
PragmaTwice: Sorry but, I actually can hardly find the benefit of merging the code for `BZPOP` and `ZPOP`. Most of I can see are lots of `if (block_)` checks to implement different logic separately, which seems just make the code more unreadable.

If you want to merge some thing to prevent duplication, I think you need to find some **common logic** so that these code can be treated uniformly rather than just put more `if`s, e.g.

```
f() { block a ... }
g() { block b ... }

// that is NOT a deduplication
h() { if(...) { block a ...} else { block b ... } }
```

Yangsx-1: The differences between zpopmin/max and bzpopmin/max is that: 1.bzpopmin/max have an additional 'timeout' input parameter 2.bzpopmin/max have an additional 'key' output. The inner processing logic is the same.
In my first edition of the implementation, i implement the bzpopmin/max independent of zpopmin/max. But i received a comment that it has something duplicate. Maybe i misunderstood it. 
But i think it's Ok? There are also some 'block' in the implementation of bzmpop and zmpop.

src/commands/cmd_zset.cc
```diff
@@ -221,64 +223,217 @@ class CommandZLexCount : public Commander {
...
     }
+
+    if (*parse_result < 0) {
+      return {Status::RedisParseErr, errTimeoutIsNegative};
+    }
+
+    timeout_ = *parse_result;
+
+    keys_ = std::vector<std::string>(args.begin() + 1, args.end() - 1);
     return Commander::Parse(args);
```
PragmaTwice: For example, in the `Parse` function you write, it is just two code block separated by `if (block_)`, i.e.
```
// the original code
ZPOP::Parse() {
  parse ZPOP...
}

BZPOP::Parse() {
  parse BZPOP...
}

// your improved code
Parse() {
  if(!block_) { 
    parse ZPOP ... 
  } 
  parse BZPOP ... 
}
``` 

IMHO it is not related to deduplication, i.e. the duplicated code seems not eliminated at all. 
And the readabilitiy of the code get even worse in my view.

Yangsx-1: Ok, i'll split zpopmin/max, do i also need to split zmpop and bzmpop? 
I think there are many same codes in zmpop and bzmpop, so the two commands can merge together.

torwig: @Yangsx-1 You can move the same code into functions (and call them from both commands) but still have two different commands.

PragmaTwice (dismissed):
Looks good to me with a glance. See if @torwig has more detailed review : )

Yangsx-1:
@torwig Is there anything i need to modify? :)

torwig:
@Yangsx-1 Right now I don't know :) I'll have a look at the code later.

tests/gocase/unit/type/zset/zset_test.go
```diff
@@ -318,6 +318,62 @@ func basicTests(t *testing.T, rdb *redis.Client, ctx context.Context, encoding s
...
+		require.Equal(t, redis.Z{Score: 2, Member: "e"}, resultz)
+		resultz = rdb.BZPopMin(ctx, 0, "zseta", "zsetb").Val().Z
+		require.Equal(t, redis.Z{Score: 3, Member: "c"}, resultz)
+		var err = rdb.BZPopMin(ctx, time.Millisecond*1000, "zseta", "zsetb").Err()
+		require.Equal(t, redis.Nil, err)
+
+		rd := srv.NewTCPClient()
+		defer func() { require.NoError(t, rd.Close()) }()
+		require.NoError(t, rd.WriteArgs("bzpopmin", "zseta", "0"))
+		time.Sleep(time.Millisecond * 1000)
```
torwig: @Yangsx-1 Why do we need the call to `Sleep` here?

Yangsx-1: I think we need to wait for the client `rd` actually blocked. 
The code has two client, if `rdb` add the key-value before `rd` actually blocked, it will not test out whether the key-value is poped after another client add.
Other go tests for command with block also have the same `sleep` operation.

torwig:
@Yangsx-1 I made a PR with my changes to your branch. You can approve it and merge so they should be visible here.

Yangsx-1:
> @Yangsx-1 I made a PR with my changes to your branch. You can approve it and merge so they should be visible here.

Thanks for your patience and modification!

tests/gocase/unit/type/zset/zset_test.go
torwig: I thought the line `require.NoError(t, rd.WriteArgs("bzpopmin", "zseta", "0"))` is blocking by itself.

Yangsx-1: You are right. I test it without sleep. There is no need to sleep here.

torwig (approved):
LGTM

PragmaTwice (approved):

git-hulk:
Thanks all, merging...

git-hulk:
Thanks for @Yangsx-1 contribution, you can add the command in https://github.com/apache/incubator-kvrocks-website/blob/main/docs/supported-commands.md if you like. Also thanks for the great review comments from @torwig and @PragmaTwice.

PR #1491, opened 2023-06-11
Automatically download the golangci-lint if not exists

infdahai:
I use `dpkg` to install golanglint-ci packags in sys_path because the visit to raw.github.com for me is unstable. So I think we can support this way about system paths.

```shell
user$ which golangci-lint
/usr/bin/golangci-lint
```

If we find the package exists in system path, we will do check-version function and the program raises an error to tell users to update system packages if the check fails.
If we don't find it, we get the package from GoPATH or download it from github.

These two ways are from https://golangci-lint.run/usage/install/.


x.py
```diff
@@ -192,15 +193,24 @@ def clang_tidy(dir: str, jobs: Optional[int], clang_tidy_path: str, run_clang_ti
...
-    gopath = run_pipe(go, 'env', 'GOPATH').read().strip()
-    bindir = Path(gopath).absolute() / 'bin'
-    binpath = bindir / 'golangci-lint'
-    if not binpath.exists():
-        output = run_pipe('curl', '-sfL', 'https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh',
-                          verbose=True)
-        run('sh', '-s', '--', '-b', str(bindir), 'v1.49.0', verbose=True, stdin=output)
+def golangci_lint(golangci_lint_path: str) -> None:
+    binpath = ""
+    golangci_command = find_command(golangci_lint_path, msg="golangci-lint is required")
```
PragmaTwice: It seems if it is not found in `$PATH` (according to the default argument), the script will abort. However, I think we have no such requirement in the original code.

x.py
```diff
@@ -192,15 +193,24 @@ def clang_tidy(dir: str, jobs: Optional[int], clang_tidy_path: str, run_clang_ti
...
-    gopath = run_pipe(go, 'env', 'GOPATH').read().strip()
-    bindir = Path(gopath).absolute() / 'bin'
-    binpath = bindir / 'golangci-lint'
-    if not binpath.exists():
-        output = run_pipe('curl', '-sfL', 'https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh',
-                          verbose=True)
-        run('sh', '-s', '--', '-b', str(bindir), 'v1.49.0', verbose=True, stdin=output)
+def golangci_lint(golangci_lint_path: str) -> None:
+    binpath = ""
+    if system('which '+ golangci_lint_path)  == 0:
```
PragmaTwice: `shutil.which` seems to be a better choice here.

tisonkun (changes requested):
1. Please check the path to golangci-lint in one step, and download in another step. You currently mix up all these logics.
2. Please build the download URL with the predefined `GOLANGCI_LINT_REQUIRED_VERSION` so that we don't miss update multiple place.

x.py
```diff
@@ -192,15 +194,36 @@ def clang_tidy(dir: str, jobs: Optional[int], clang_tidy_path: str, run_clang_ti
...
         output = run_pipe('curl', '-sfL', 'https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh',
-                          verbose=True)
-        run('sh', '-s', '--', '-b', str(bindir), 'v1.49.0', verbose=True, stdin=output)
+                            verbose=True)
+        version_str = 'v' + '.'.join(map(str, GOLANGCI_LINT_REQUIRED_VERSION))
+        run('sh', '-s', '--', '-b', bindir, version_str, verbose=True, stdin=output)
+
+    binpath = ""
+    if which(golangci_lint_path) is None:
+        bindir, binpath = get_gopath()
```
infdahai, line 218: `binpath` -> Type error.

PragmaTwice (approved):
LGTM. cc @tisonkun

tisonkun (approved):
Thank you!

PR #1493, opened 2023-06-11
Fix data race when joining the task runner

### LINKED ISSUE #1448: Data race in TBB::concurrent_queue when running integrate tests
### Search before asking

### Version

unstable

### Minimal reproduce step

detail logs: https://github.com/apache/incubator-kvrocks/actions/runs/5015903443/jobs/8992316856?pr=1446

all tests are passed

### What did you see instead?

one of them crashed when running

### Anything Else?

_No response_

git-hulk:
This may close #1448 

Currently, server->Join will join the task runner before the cron thread, so it may have the data race since the cron thread may access the task runner as well. We can avoid this by removing the task runner join which will do in the task runner destructor.

### DIFF
--- a/src/common/task_runner.h
+++ b/src/common/task_runner.h
@@ -42,7 +42,13 @@ class TaskRunner {
 
   TaskRunner(const TaskRunner&) = delete;
   TaskRunner& operator=(const TaskRunner&) = delete;
-  ~TaskRunner() = default;
+
+  ~TaskRunner() {
+    if (state_ != Stopped) {
+      Cancel();
+      auto _ = Join();
+    }
+  }
 
   template <typename T>
   void Publish(T&& task) {
--- a/src/common/task_runner.h
+++ b/src/common/task_runner.h
@@ -42,13 +42,7 @@ class TaskRunner {
 
   TaskRunner(const TaskRunner&) = delete;
   TaskRunner& operator=(const TaskRunner&) = delete;
-
-  ~TaskRunner() {
-    if (state_ != Stopped) {
-      Cancel();
-      auto _ = Join();
-    }
-  }
+  ~TaskRunner() = default;
 
   template <typename T>
   void Publish(T&& task) {
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -236,6 +236,9 @@ void Server::Join() {
   if (auto s = util::ThreadJoin(compaction_checker_thread_); !s) {
     LOG(WARNING) << "Compaction checker thread operation failed: " << s.Msg();
   }
+  if (auto s = task_runner_.Join(); !s) {
+    LOG(WARNING) << "Task runner thread operation failed: " << s.Msg();
+  }
 }
 
 Status Server::AddMaster(const std::string &host, uint32_t port, bool force_reconnect) {
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -230,9 +230,6 @@ void Server::Join() {
     worker->Join();
   }
 
-  if (auto s = task_runner_.Join(); !s) {
-    LOG(WARNING) << s.Msg();
-  }
   if (auto s = util::ThreadJoin(cron_thread_); !s) {
     LOG(WARNING) << "Cron thread operation failed: " << s.Msg();
   }
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -237,7 +237,7 @@ void Server::Join() {
     LOG(WARNING) << "Compaction checker thread operation failed: " << s.Msg();
   }
   if (auto s = task_runner_.Join(); !s) {
-    LOG(WARNING) << "Task runner thread operation failed: " << s.Msg();
+    LOG(WARNING) << s.Msg();
   }
 }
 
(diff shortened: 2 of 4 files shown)


torwig (dismissed):
Good catch.
Sounds legitimate to me.

git-hulk:
@torwig Thanks for your quick review.

PragmaTwice:
Could you specify the location of another Join call of the task runner?

git-hulk:
> Could you specify the location of another Join call of the task runner?

Sure, the TaskRunner destructor will join as well: https://github.com/apache/incubator-kvrocks/blob/unstable/src/common/task_runner.h#L49

PragmaTwice (dismissed):
LGTM. Although a dtor with a long time waiting may cause some misunderstanding.

git-hulk:
> LGTM. Although a dtor with a long time waiting may cause some misunderstanding.

Yes, I think we can remove the Cancel and Join in dtor since the Server::Stop will cancel the task runner first. Can take a look again. @PragmaTwice @torwig

torwig:
@git-hulk Yes, `Server::Stop` will cancel and `Server::Join` will join all threads.
Additionally, here https://github.com/apache/incubator-kvrocks/blob/unstable/src/common/task_runner.cc#L48 should we stop joining threads when the first call to `util::ThreadJoin` or just log an error and continue joining other threads?

git-hulk:
> @git-hulk Yes, `Server::Stop` will cancel and `Server::Join` will join all threads. Additionally, here https://github.com/apache/incubator-kvrocks/blob/unstable/src/common/task_runner.cc#L48 should we stop joining threads when the first call to `util::ThreadJoin` or just log an error and continue joining other threads?

oooh, my bad. I didn't see that, updated.

src/server/server.cc
```diff
@@ -230,15 +230,15 @@ void Server::Join() {
...
-    LOG(WARNING) << s.Msg();
-  }
   if (auto s = util::ThreadJoin(cron_thread_); !s) {
     LOG(WARNING) << "Cron thread operation failed: " << s.Msg();
   }
   if (auto s = util::ThreadJoin(compaction_checker_thread_); !s) {
     LOG(WARNING) << "Compaction checker thread operation failed: " << s.Msg();
   }
+  if (auto s = task_runner_.Join(); !s) {
+    LOG(WARNING) << "Task runner thread operation failed: " << s.Msg();
```
git-hulk:
```suggestion
    LOG(WARNING) << s.Msg();
```

torwig:
@git-hulk My idea was to change the loop to the following:

```
for (auto &thread : threads_) {
  if (auto s = util::ThreadJoin(thread); !s) {
    LOG(WARNING) << "Failed to join thread: " << s.Msg();
    continue;
  }
}
```

What do you think about that? @PragmaTwice What is your opinion on that?

git-hulk:
> LOG(WARNING) << "Failed to join thread: " << s.Msg();

Yes, it's good to only log an error when failing to join since other places like the worker thread only prints an error.

git-hulk:
> @git-hulk My idea was to change the loop to the following:
> 
> ```
> for (auto &thread : threads_) {
>   if (auto s = util::ThreadJoin(thread); !s) {
>     LOG(WARNING) << "Failed to join thread: " << s.Msg();
>     continue;
>   }
> }
> ```
> 
> What do you think about that? @PragmaTwice What is your opinion on that?

Yes, it's good to only log an error when failing to join since other places like the worker thread only print an error.

PragmaTwice:
> @git-hulk My idea was to change the loop to the following:
> 
> ```
> for (auto &thread : threads_) {
>   if (auto s = util::ThreadJoin(thread); !s) {
>     LOG(WARNING) << "Failed to join thread: " << s.Msg();
>     continue;
>   }
> }
> ```
> 
> What do you think about that? @PragmaTwice What is your opinion on that?

Seems good to me.

torwig (dismissed):

src/common/task_runner.h
```diff
@@ -42,13 +42,7 @@ class TaskRunner {
...
   TaskRunner(const TaskRunner&) = delete;
   TaskRunner& operator=(const TaskRunner&) = delete;
-
-  ~TaskRunner() {
-    if (state_ != Stopped) {
-      Cancel();
-      auto _ = Join();
-    }
-  }
+  ~TaskRunner() = default;
```
PragmaTwice: Seems we need some guarantee for this object? i.e. the behavior will be **undefined** while the taskrunner is destructed and the attached threads are still running, which is dangerous.

Can refer to std::jthread (https://en.cppreference.com/w/cpp/thread/jthread/~jthread) and std::thread (https://en.cppreference.com/w/cpp/thread/thread/~thread).

The previous dtor build such a guarantee like std::jthread, but I think we can call `Join` explicitly as far as possible to prevent misunderstanding. Removing such dtor cannot help the situation.

git-hulk: The data race may be caused by the task runner's `Join` function will clear the concurrent queue, so it could be fixed after moving to the last one to join. It's good for me to keep the Cancel and Join in dctor.

PragmaTwice (approved):

torwig (approved):

git-hulk:
Thanks all, merging...

PR #1498, opened 2023-06-16
Warning message for invalid key in configuration file

### LINKED ISSUE #1497: Print warning messages for invalid config options
### Search before asking

### Motivation

Currently kvrocks just ignore these invalid config options, like nonexistent config keys, e.g.
```
# kvrocks config file
some-config-key-do-not-exist value
```

No warning or error message is printed, which makes configration debug harder.

We want to print some warning messages (rather than a fatal error for compatibility) when facing inexistent keys.

### Solution

_No response_

gloof11:
This is a fix that I was able to come up with!
It's hacky, but it works.


infdahai:
Hi @gloof11 can you write a cpp unit test? (probably in https://github.com/apache/incubator-kvrocks/blob/unstable/tests/cppunit/config_test.cc)

src/config/config.cc
```diff
@@ -659,6 +659,12 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
     tokens[input.second] = input.first.substr(ns_str_size);
   }
   auto iter = fields_.find(field_key);
+
+  // Check if a configuration key is a valid key
```
infdahai: Can you explain why does it work here?

We write configs to `fields_`.
https://github.com/apache/incubator-kvrocks/blob/46e7eaf428c3e4935cc73f70be3f8c9066d5cb3c/src/config/config.cc#L226-L230

src/config/config.cc
```diff
@@ -659,6 +659,12 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
...
+  if (iter->second == 0x0) {
```
torwig: What is the magic number `0x0`?

src/config/config.cc
PragmaTwice: Actually I wonder why not check `iter == fields_.end()` directly : )

PragmaTwice:
Thanks for your contribution!

The code need to be formatted to pass the CI, refer to https://kvrocks.apache.org/community/contributing#code-style.

src/config/config.cc
gloof11: I realized that "iter" after the find operations returns an object of all possible keys. I summarized that "iter->second" points to the name of the key, so if you were to pass a key that cannot exist, it would return zero. Therefore, if it finds a key that doesn't exist, it will return 0x0.

src/config/config.cc
gloof11: I was wondering what the properties of "fields" was. This was a very hacky fix, so I will dig deeper to see if I can make the code more "understandable" and less "magic" lol.

src/config/config.cc
infdahai: we don't use the magic number, because it's undefined based on https://eel.is/c++draft/iterator.requirements#general-7.

PragmaTwice:
It seems the code is still not formatted properly.

And I think you need to give a reason why you do not just check `iter == fields_.end()`. TBH I cannot understand your code.

gloof11:
Just made a commit. I read up on how maps **actually** work in Cpp. In the refactor, if the key that was passed to "iter" was not a valid key via "find()", then it would be equivalent to "past-the-end". It simply checks if both values are considered "past-the-end", and throws an error message if that's the case. Also, I'm not at all familiar with writing unit tests for C++. Sorry :(

gloof11:
Also, I'm not familiar at all with writing unit tests in C++. I'm sorry :(

src/config/config.cc
```diff
@@ -658,7 +658,13 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
     field_key = input.first;
     tokens[input.second] = input.first.substr(ns_str_size);
   }
+
   auto iter = fields_.find(field_key);
+
+  if (iter == fields_.end()) {
```
torwig: Would you like to move the warning message inside `else` branch of `if (iter != fields_.end()) {` check which is located later? 
P.S. I think you can omit a lot of `!` inside the message, just capitalize the word 'warning' as `WARNING:...`

gloof11: Sure can

torwig (dismissed):
LGTM

src/common/status.h
```diff
@@ -174,7 +174,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> :
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
-      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
+      : BaseType((typename StringInStatusOr<U>::BaseType&&)(std::move(v))) {}
```
infdahai:
```suggestion
      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
```

Don't change this line. Just like this.

infdahai:
I don't know whether it needs a test because we can't get a log message in th test. I think the code is obvious.

src/common/status.h
gloof11: Strange, I never touched this file. When I change it to how its supposed to be formatted and run the format script, it appends the && back to BaseType.

torwig: @gloof11 The change was made by `./x.py format`. On my Fedora 38, I also have this behavior.

torwig:
@infdahai Yes, I also think that a test, in this case, is too fanatic.

src/common/status.h
torwig: @gloof11 You can undo the change manually and commit + push again.

infdahai: `./x.py format --clang-format-path clang-format-14`

I guess you're using clang-format-16.

torwig: @infdahai Right. Before I upgraded Fedora from 37 to 38, the command `./x.py format` did its job the right way :)

PragmaTwice:
You need to revert the change of `src/common/status.h` to make the CI pass.

image

src/config/config.cc
```diff
@@ -658,12 +658,16 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
...
+
   auto iter = fields_.find(field_key);
+
   if (iter != fields_.end()) {
     auto &field = iter->second;
     field->line_number = line_number;
     auto s = field->Set(input.second);
     if (!s.IsOK()) return s.Prefixed(fmt::format("failed to set value of field '{}'", field_key));
+  } else {
+    std::cout << fmt::format("WARNING: '{}' is not a valid configuration key!", field_key) << std::endl;
```
```suggestion
    std::cout << fmt::format("WARNING: '{}' at line {} is not a valid configuration key.", field_key, line_number) << std::endl;
```

PragmaTwice:
Hi @gloof11 , any progress or problem here?

tisonkun (commented):
Generally looks good. Please revert unrelated whitespace changes.

torwig (dismissed):

torwig:
@PragmaTwice You slashed the clang linter with the long line :)

src/config/config.cc
```diff
@@ -659,12 +659,16 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
     field_key = input.first;
     tokens[input.second] = input.first.substr(ns_str_size);
   }
+
   auto iter = fields_.find(field_key);
+
```
```suggestion
```

git-hulk (dismissed):

torwig (dismissed):

src/config/config.cc
```diff
@@ -659,12 +659,15 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
...
   }
+
   auto iter = fields_.find(field_key);
   if (iter != fields_.end()) {
     auto &field = iter->second;
     field->line_number = line_number;
     auto s = field->Set(input.second);
     if (!s.IsOK()) return s.Prefixed(fmt::format("failed to set value of field '{}'", field_key));
+  } else {
+    std::cout << fmt::format("WARNING: '{}' at line {} is not a valid configuration key.", field_key, line_number) << std::endl;
```
```suggestion
    std::cout << fmt::format("WARNING: '{}' at line {} is not a valid configuration key.", field_key, line_number)
              << std::endl;
```
```suggestion
    std::cout << fmt::format("WARNING: '{}' at line {} is not a valid configuration key.", field_key, line_number)
              << std::endl;
```

git-hulk (approved):

torwig (approved):

tisonkun (approved):
Thank you!

Merging...

PR #1499, opened 2023-06-16
Remove ConnContext pointers to prevent manual allocation

PragmaTwice:
We do not need to make heap pointers for ConnContext, since it is only a light POD type.


torwig (approved):
LGTM

git-hulk (approved):

PR #1502, opened 2023-06-17
Add support of the new command ZUNION

### LINKED ISSUE #1461: Add the support of the ZUNION command
### Search before asking

### Motivation

This issue considers to support the [zunion](https://redis.io/commands/zunion/) command.

### Solution

_No response_

uds5501:
Fixes : #1461

### DIFF
--- a/src/commands/cmd_zset.cc
+++ b/src/commands/cmd_zset.cc
@@ -805,6 +805,71 @@ class CommandZMScore : public Commander {
   }
 };
 
+class CommandZUnion : public Commander {
+  public:
+    Status Parse(const std::vector<std::string> &args) override {
+      auto parse_result = ParseInt<int>(args[1], 10);
+      if (!parse_result) {
+        return {Status::RedisParseErr, errValueNotInteger};
+      }
+
+      numkeys_ = *parse_result;
+      if (numkeys_ > args.size() - 2) {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+
+      size_t key_iterator = 0;
+      while (key_iterator < numkeys_) {
+        keys_weights_.emplace_back(KeyWeight{args[key_iterator + 2], 1});
+        key_iterator++;
+      }
+
+      size_t option_iterator = 2 + numkeys_;
+      while (option_iterator < args.size()) {
+        if (util::ToLower(args[option_iterator]) == "aggregate" && option_iterator + 1 < args.size()) {
+          if (util::ToLower(args[option_iterator + 1]) == "sum") {
+            aggregate_method_ = kAggregateSum;
+          } else if (util::ToLower(args[option_iterator + 1]) == "min") {
+            aggregate_method_ = kAggregateMin;
+          } else if (util::ToLower(args[option_iterator + 1]) == "max") {
+            aggregate_method_ = kAggregateMax;
+          } else {
+            return {Status::RedisParseErr, "aggregate param error"};
+          }
+          option_iterator += 2;
+        } else if (util::ToLower(args[option_iterator]) == "weights" && option_iterator + numkeys_ < args.size()) {
+          size_t k = 0;
+          while (k < numkeys_) {
+            auto weight = ParseFloat(args[option_iterator + k + 1]);
+            if (!weight || std::isnan(*weight)) {
+              return {Status::RedisParseErr, "weight is not a double or out of range"};
+            }
+            keys_weights_[k].weight = *weight;
+            k++;
+          }
+          option_iterator += numkeys_ + 1;
+        } else if (util::ToLower(args[option_iterator] == "withscores")) {
+          with_scores_ = true;
+        } else {
+          return {Status::RedisParseErr, errInvalidSyntax};
+        }
+      }
+      return Commander::Parse(args);
+    }
+
+    Status Execute(Server *svr, Connection *conn, std::string *output) override {
+      
+    }
+  
+  protected:
+    size_t numkeys_ = 0;
+    bool with_scores_ = false;
+    std::vector<KeyWeight> keys_weights_;
+    AggregateMethod aggregate_method_ = kAggregateSum;
+}
+
+// zunionstore destination numkeys key [key ...] [WEIGHTS weight [weight ...]] [AGGREGATE SUM|MIN|MAX]
+// zunion numkeys key [key ...] [WEIGHTS weight [weight ...]] [AGGREGATE SUM|MIN|MAX] [WITHSCORES]
 class CommandZUnionStore : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
--- a/src/commands/cmd_zset.cc
+++ b/src/commands/cmd_zset.cc
@@ -1028,6 +1028,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandZAdd>("zadd", -4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandZScore>("zscore", 3, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandZMScore>("zmscore", -3, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandZScan>("zscan", -3, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandZUnionStore>("zunionstore", -4, "write", 1, 1, 1), )
+                        MakeCmdAttr<CommandZUnionStore>("zunionstore", -4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandZUnion>("zunion", -4, "read-only", 1, 1, 1), )
 
 }  // namespace redis
--- a/src/commands/cmd_zset.cc
+++ b/src/commands/cmd_zset.cc
@@ -864,6 +864,7 @@ class CommandZUnion : public Commander {
         option_iterator += numkeys_ + 1;
       } else if (util::ToLower(args[option_iterator]) == "withscores") {
         with_scores_ = true;
+        option_iterator++;
       } else {
         return {Status::RedisParseErr, errInvalidSyntax};
       }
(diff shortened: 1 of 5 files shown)


src/commands/cmd_zset.cc
```diff
@@ -819,6 +819,81 @@ class CommandZMScore : public Commander {
...
+          auto weight = ParseFloat(args[option_iterator + k + 1]);
+          if (!weight || std::isnan(*weight)) {
+            return {Status::RedisParseErr, "weight is not a double or out of range"};
+          }
+          keys_weights_[k].weight = *weight;
+          k++;
+        }
+        option_iterator += numkeys_ + 1;
+      } else if (util::ToLower(args[option_iterator]) == "withscores") {
+        with_scores_ = true;
```
git-hulk, line 1183: It would go into the dead loop if passed the WITHSCORES flag here. And for the flag parse should take a look at the implementation like: https://github.com/apache/incubator-kvrocks/blob/91d9abb6c1169db7b91409456b8bf4bd478309dc/src/commands/cmd_zset.cc#L287

src/commands/cmd_zset.cc
```diff
@@ -819,6 +819,81 @@ class CommandZMScore : public Commander {
...
+      }
+    }
+    return Commander::Parse(args);
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::ZSet zset_db(svr->storage, conn->GetNamespace());
+    std::vector<MemberScore> member_scores;
+    uint64_t size = 0;
+    auto s = zset_db.Union(keys_weights_, aggregate_method_, &size, &member_scores);
```
torwig: Since you don't use the `size` variable you could remove it and pass a `nullptr` to the `Union` function.

tests/gocase/unit/type/zset/zset_test.go
```diff
@@ -956,6 +956,50 @@ func basicTests(t *testing.T, rdb *redis.Client, ctx context.Context, encoding s
...
+		zsetInt := []redis.Z{
+			{1, "a"},
+			{3, "b"},
+			{3, "d"},
+			{5, "c"}}
+		require.Equal(t, zsetInt, rdb.ZUnionWithScores(ctx, redis.ZStore{Keys: []string{"zseta", "zsetb"}}).Val())
+	})
+
+	t.Run(fmt.Sprintf("ZUNION with weights - %s", encoding), func(t *testing.T) {
+		zsetInt := []redis.Z{
```
torwig, line 1080: I wouldn't recommend relying on the order of test execution. Instead, delete the keys you need at the beginning of the test, populate those keys with values and then test required assertions. In this particular case, if you need sorted sets defined in another test, you can extract dedicated functions that will prepare the sorted sets for you. 
On top of that, you can check the `createZset` functions in this file and use it at the beginning of each (if needed) test. This function does exactly what I described.

src/commands/cmd_zset.cc
uds5501: Makes sense.
Btw for parsing, imo, a simple `option_iterator++` would also work, wdyt?

src/commands/cmd_zset.cc
uds5501: ack, incorporating this suggestion.

tests/gocase/unit/type/zset/zset_test.go
uds5501: updated.

src/types/redis_zset.cc
```diff
@@ -744,16 +754,15 @@ rocksdb::Status ZSet::UnionStore(const Slice &dst, const std::vector<KeyWeight>
       }
     }
   }
-  if (!dst_zset.empty()) {
-    std::vector<MemberScore> mscores;
-    mscores.reserve(dst_zset.size());
+  auto compare_score = [](const MemberScore &score1, const MemberScore &score2) { return score1.score < score2.score; };
```
uds5501: @torwig / @git-hulk needed an opinion on this. If the scores of both the members are same during sorting, do we want to also lexicographically check the members as tie breakers or are we okay with them being in the order which they were originally present in?

torwig: @uds5501 If I understand correctly, the initial order of how members are stored is already lexicographically sorted and retrieved in that order. @git-hulk Fix my thought if it is incorrect.

git-hulk: > @uds5501 If I understand correctly, the initial order of how members are stored is already lexicographically sorted and retrieved in that order. @git-hulk Fix my thought if it is incorrect.

Yes, that's right.

src/commands/cmd_zset.cc
git-hulk: I think it'd be better to use the common parser like other commands to make the codebase more coherent.

src/types/redis_zset.cc
uds5501: Understood, so I shold lexicographically sort the final result as well , right?

git-hulk: Yes, exactly.

src/commands/cmd_zset.cc
uds5501: Incorporated this change, thanks

src/types/redis_zset.cc
uds5501: Incorporated this sorting logic, thanks.

src/types/redis_zset.cc
```diff
@@ -744,16 +754,20 @@ rocksdb::Status ZSet::UnionStore(const Slice &dst, const std::vector<KeyWeight>
...
+  if (members && !dst_zset.empty()) {
+    members->reserve(dst_zset.size());
     for (const auto &iter : dst_zset) {
-      mscores.emplace_back(MemberScore{iter.first, iter.second});
+      members->emplace_back(MemberScore{iter.first, iter.second});
     }
-    if (saved_cnt) *saved_cnt = mscores.size();
-    Overwrite(dst, mscores);
+    if (saved_cnt) *saved_cnt = members->size();
+    std::sort(members->begin(), members->end(), compare_score);
```
git-hulk: Sorry, maybe I misunderstand your previous comment. We don't need to sort members here, because the underlying storage will sort them.

tests/cppunit/command_parser_test.cc
```diff
@@ -26,9 +26,10 @@ TEST(CommandParser, Parse) {
   // [ HELLO i1 v1 | HI v2 ] [X i2 | Y]
   std::vector<std::string> c1{"hello", "1", "a"}, c2{"hi", "b"}, c3{"hi", "c", "x", "2"}, c4{"hello", "3", "d", "y"},
       c5{"hi", "e", "y"}, c6{"y"}, d1{"hello"}, d2{"hi"}, d3{"hello", "no-int"}, d4{"x", "1", "y"},
-      d5{"hello", "1", "v", "hi", "v"}, d6{"hello", "1"};
+      d5{"hello", "1", "v", "hi", "v"}, d6{"hello", "1"}, d7{"float_test", "1.1", "c"};
```
git-hulk:
```suggestion
      d5{"hello", "1", "v", "hi", "v"}, d6{"hello", "1"}, d7{"float_test", "1.1"};
```

Adding an extra "c" will cause invalid syntax in the parser since it cannot be recognized.

src/commands/cmd_zset.cc
```diff
@@ -1285,6 +1345,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandZAdd>("zadd", -4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandZScore>("zscore", 3, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandZMScore>("zmscore", -3, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandZScan>("zscan", -3, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandZUnionStore>("zunionstore", -4, "write", 1, 1, 1), )
+                        MakeCmdAttr<CommandZUnionStore>("zunionstore", -4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandZUnion>("zunion", -4, "read-only", 1, 1, 1), )
```
git-hulk: Looks like the ZUNIONSTORE command wrongly passed the key position. For the ZUNIONSTORE/ZUNION command, its key positions need to calculate by the [numkeys] argument. Can refer to the `CommandZMPop::Range` implementation.

src/types/redis_zset.cc
uds5501: Underlying storage will sort them for ZUNIONSTORE, but we don’t use the storage for ZUNION, so i think to keep the response consistent, we’ll need to sort it at this level (atleast for ZUNION) command

git-hulk: @uds5501 OK, then it's good for me.  How about moving the sort logic to the ZUNION command since we don't need to sort for the union store?

uds5501: Makes sense @git-hulk.

src/commands/cmd_zset.cc
PragmaTwice: > Btw for parsing, imo, a simple option_iterator++ would also work, wdyt?

It is firstly errorphone, e.g. you just do not realize your dead loop here.

src/commands/cmd_zset.cc
uds5501: Updated both ZUNION and ZUNIONSTORE

src/types/redis_zset.cc
uds5501: Incorporated this change.

src/commands/cmd_zset.cc
```diff
@@ -1148,6 +1148,78 @@ class CommandZMScore : public Commander {
...
+    for (const auto &ms : member_scores) {
+      output->append(redis::BulkString(ms.member));
+      if (with_scores_) output->append(redis::BulkString(util::Float2String(ms.score)));
+    }
+    return Status::OK();
+  }
+
+  static CommandKeyRange Range(const std::vector<std::string> &args) {
+    int num_key = *ParseInt<int>(args[1], 10);
+    return {2, 1 + num_key, 1};
```
torwig, line 1213: I'm not sure, but since `ZUnion` and `ZUnionStore` differs that `ZunionStore` has the `destination` argument, should their `Range` functions return different values (at least in one position)? @git-hulk @PragmaTwice

uds5501: Ah yes, missed that, will update this.

uds5501: updated.

torwig (approved):
LGTM

git-hulk (approved):

PR #1507, opened 2023-06-21
Fix incorrect mount data in docker volume

aleksraiden:
This patch fix an issue when we use Kvrocks image into Docker Compose and will need a mount host directory as a persistent data volume. In a default, inside the container we use kvrocks:kvrocks user/group for running kvrocks in non-root mode. But we don't need a use this for data directory, because if docker compose mount host directory, a server can't use it ( and we have an startup error: Failed to open: failed to create column families: IO error: While open a file for appending: /var/lib/kvrocks/db/LOG: Permission denied


torwig (approved):
LGTM

git-hulk (approved):

git-hulk:
Thanks for @aleksraiden taking care of this, merging...

PR #1513, opened 2023-06-22
Remove user and group from Docker

aleksraiden:
Addition fix - remove a dedicated user/group kvrocks and disable manual creating a working dir instead of VOLUME directive. Because in real usages, we still have an IO error if work from non-root user and mount a working dir from host machine


tisonkun (approved):

git-hulk (approved):
LGTM, thanks for @aleksraiden taking care of this.

PR #1516, opened 2023-06-24
build(deps): Bump rocksdb v8.3.2

aleksraiden:
Bump Rocksdb v8.3.2

In this release:

- (perf) Improved the I/O efficiency of prefetching SST metadata by recording more information in the DB manifest
- (perf) CPU features are no longer detected at runtime nor in build scripts, but in source code using common preprocessor defines
- (bug) Delete an empty WAL file on DB open
- (bug) Delete temp OPTIONS file on DB open
- (bug) Fixed higher read QPS during DB::Open()
- (bug) Reduced cases of illegally using Env::Default()
- (feat) New statistics tickers
- (feat) Add new API DB::ClipColumnFamily to clip the key in CF to a certain range. It will physically deletes all keys outside the range including tombstones.
- (feat) Add MakeSharedCache() construction functions to various cache Options objects
- (feat) Changed the meaning of various Bloom filter stats

Full release doc: https://github.com/facebook/rocksdb/releases/tag/v8.3.2

### DIFF
--- a/x.py
+++ b/x.py
@@ -106,7 +106,7 @@ def build(dir: str, jobs: Optional[int], ghproxy: bool, ninja: bool, unittest: b
 
     makedirs(dir, exist_ok=True)
 
-    cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo"]
+    cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo", "-DPORTABLE=0"]
     if ghproxy:
         cmake_options.append("-DDEPS_FETCH_PROXY=https://ghproxy.com/")
     if ninja:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ option(USE_LUAJIT "use luaJIT instead of lua" ON)
 option(ENABLE_OPENSSL "enable openssl to support tls connection" OFF)
 option(ENABLE_IPO "enable interprocedural optimization" ON)
 option(ENABLE_UNWIND "enable libunwind in glog" ON)
-option(PORTABLE "build a portable binary (disable arch-specific optimizations)" OFF)
+option(PORTABLE "build a portable binary (disable arch-specific optimizations)" 0)
 # TODO: set ENABLE_NEW_ENCODING to ON when we are ready
 option(ENABLE_NEW_ENCODING "enable new encoding (#1033) for storing 64bit size and expire time in milliseconds" OFF)
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ RUN apk update && apk add git gcc g++ make cmake ninja autoconf automake libtool
 WORKDIR /kvrocks
 
 COPY . .
-RUN ./x.py build -DENABLE_OPENSSL=ON -DPORTABLE=ON -DCMAKE_BUILD_TYPE=Release -j $(nproc) $MORE_BUILD_ARGS
+RUN ./x.py build -DENABLE_OPENSSL=ON -DCMAKE_BUILD_TYPE=Release -j $(nproc) $MORE_BUILD_ARGS
 
 FROM alpine:3.16
 
--- a/cmake/rocksdb.cmake
+++ b/cmake/rocksdb.cmake
@@ -26,8 +26,8 @@ endif()
 include(cmake/utils.cmake)
 
 FetchContent_DeclareGitHubWithMirror(rocksdb
-  facebook/rocksdb v8.1.1
-  MD5=b362246096dbd15839749da37d3ccda9
+  facebook/rocksdb v8.3.2
+  MD5=0a3251e94df18e06711fc7d4c5fed9cf
 )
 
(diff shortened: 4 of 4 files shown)


aleksraiden:
In this version of rocksdb changes a compiler flag PORTABLE, so I changes it into Cmake and Docker for default value 0. In this case all of features (related to CPU) are in runtime. 

So, now PORTABLE option must be:

- 0 and compiler work as best as it can
- 1 usually nothing to do; compiler default is typically the most general
- <ARCH> and we can set a CPU family for build best optimal code, e.g.: nocona core2 nehalem corei7 westmere sandybridge corei7-avx ivybridge core-avx-i haswell core-avx2 broadwell skylake skylake-avx512 cannonlake icelake-client rocketlake icelake-server cascadelake tigerlake cooperlake sapphirerapids alderlake bonnell atom silvermont slm goldmont goldmont-plus tremont knl knm x86-64 x86-64-v2 x86-64-v3 x86-64-v4 eden-x2 nano nano-1000 nano-2000 nano-3000 nano-x2 eden-x4 nano-x4 k8 k8-sse3 opteron opteron-sse3 athlon64 athlon64-sse3 athlon-fx amdfam10 barcelona bdver1 bdver2 bdver3 bdver4 znver1 znver2 znver3 btver1 btver2 native

Detailed, we can see in this code diff: https://github.com/facebook/rocksdb/pull/11419/commits/5cb625142c77c84111f1f65b6f35f2318c602802

x.py
```diff
@@ -106,7 +106,7 @@ def build(dir: str, jobs: Optional[int], ghproxy: bool, ninja: bool, unittest: b
 
     makedirs(dir, exist_ok=True)
 
-    cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo"]
+    cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo", "-DPORTABLE=0"]
```
git-hulk, line 109: Why do we need to add `PORTABLE=0` here? I saw the default value for the PORTABLE option is 0 now.

aleksraiden: I have no idea, why, but build are Ok only when we manual add -DPORTABLE=0. In all other cases:
cc1plus: error: bad value ‘OFF’ for ‘-march=’ switch

aleksraiden: As I see a cmake docs: "If no initial <value> is provided, boolean OFF is the default value", so we need a manual add 0 as value

git-hulk: It should be broken by this PR: https://github.com/facebook/rocksdb/pull/11419. But I think we can remove `"-DPORTABLE=0"` since the default value has been corrected to 0 in the previous line: https://github.com/apache/kvrocks/pull/1516/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR32

aleksraiden: If we add 0 or 1 as value (in option) cmake rewrite it into ON/OFF and only manual setup are valid

aleksraiden: > It should be broken by this PR: [facebook/rocksdb#11419](https://github.com/facebook/rocksdb/pull/11419). But I think we can remove `"-DPORTABLE=0"` since the default value has been corrected to 0 in the previous line: https://github.com/apache/kvrocks/pull/1516/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR32

No, default value, if we set 0 or 1, cmake rewrite into own default ON/OFF

git-hulk: oh, I got your point. Would it be better to remove the PORTABLE option in CMakeList.txt? since the rocksdb will set the default value to 0 if the PORTABLE was not passed. cc @PragmaTwice

aleksraiden: And this way not a good point - now we can manual add cpu arch for better optimization (much better as default) and if completely remove options, there no way to do this anymore

git-hulk: What I mean is only to remove the option from the Kvrocks CMakeList.txt to prevent automatically rewriting it into ON/OFF, users still can pass the PORTABLE via the CMake arguments. Did I miss anything?

aleksraiden: Yes, but we need manual setup always, because in rocksdb.cmake we use it.

So, only one right way in now - always manual set PORTABLE as a Cmake flag on build

torwig (approved):
LGTM.
@git-hulk I also had that question about the `PORTABLE` flag :+1:

git-hulk (approved):

x.py
tisonkun: Do we report this case to the upstream? I'm OK with this patch but we should at least report the inconvenience for possible improvement from upstream.

tisonkun (approved):

x.py
aleksraiden: Yes, i send issue for rocksdb team

tisonkun: @aleksraiden Is it public and we can cross-post the link here?

tisonkun:
Merging...

Thank you!

Issue #1457, opened 2023-05-21
Add the support of the ZINTER, ZINTERCARD commands


infdahai:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

This issue considers to support [zinter](https://redis.io/commands/zinter/), zintercard commands.

### Solution

- [ ] zinter
- [ ] zintercard

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

IoCing:
i want to do this

git-hulk:
@IoCing Assigned, thanks

jihuayu:
Has been done in https://github.com/apache/kvrocks/issues/1983, close.

PR #1517, opened 2023-06-24
Add support of new command: ZINTER / ZINTERCARD

IoCing:
#1457


src/commands/cmd_zset.cc
```diff
@@ -1311,6 +1311,133 @@ class CommandZInterStore : public CommandZUnionStore {
...
+    }
+
+    size_t j = 0;
+    while (j < numkeys_) {
+      keys_weights_.emplace_back(KeyWeight{args[j + 2], 1});
+      j++;
+    }
+
+    size_t i = 2 + numkeys_;
+    while (i < args.size()) {
```
torwig, line 1410: What if there will be more than 1 `LIMIT` parameter? I mean, why there is a loop here? Should we just check the next element after the last key?
`ZINTERCARD 3 key1 key2 key3 LIMIT 10 LIMIT 15 LIMIT 20`

IoCing: Thanks,this is  a problem. and i think some parameter Parse have the same problem.

src/types/redis_zset.cc
```diff
@@ -707,6 +707,112 @@ rocksdb::Status ZSet::InterStore(const Slice &dst, const std::vector<KeyWeight>
...
+            dst_zset[ms.member] = score;
+          }
+          break;
+      }
+    }
+  }
+  if (!mscores->empty()) {
+    mscores->clear();
+  }
+  if (!dst_zset.empty()) {
```
torwig: You can omit to check the `!mscores->empty()` condition here because later you use range for-loop.

src/types/redis_zset.cc
```diff
@@ -707,6 +707,112 @@ rocksdb::Status ZSet::InterStore(const Slice &dst, const std::vector<KeyWeight>
...
+    double score = ms.score * keys_weights[0].weight;
+    if (std::isnan(score)) score = 0;
+    dst_zset[ms.member] = score;
+    member_counters[ms.member] = 1;
+  }
+
+  for (size_t i = 1; i < keys_weights.size(); i++) {
+    s = RangeByScore(keys_weights[i].key, spec, &target_mscores, &target_size);
+    if (!s.ok() || target_mscores.empty()) return s;
+    bool flag = false;
```
torwig: Give a more descriptive name to a boolean variable instead of `flag`.

src/types/redis_zset.cc
```diff
@@ -707,6 +707,112 @@ rocksdb::Status ZSet::InterStore(const Slice &dst, const std::vector<KeyWeight>
...
+        flag = true;
+      }
+    }
+    if (!flag) {
+      *saved_cnt = 0;
+      return rocksdb::Status::OK();
+    }
+  }
+  uint64_t count = 0;
+  if (!dst_zset.empty()) {
```
torwig: Same as in the previous function.

src/types/redis_zset.cc
```diff
@@ -707,6 +707,112 @@ rocksdb::Status ZSet::InterStore(const Slice &dst, const std::vector<KeyWeight>
...
+    for (const auto &iter : dst_zset) {
+      if (member_counters[iter.first] != keys_weights.size()) continue;
+      count++;
+      LOG(INFO) << "limit : " << limit << " count : " << count;
```
torwig: I guess you added this line just for debugging purposes. Now you can remove it.

torwig (changes requested):

src/types/redis_zset.cc
IoCing: Thanks for your comment, I have made the edits @torwig

src/types/redis_zset.cc
```diff
@@ -707,6 +707,109 @@ rocksdb::Status ZSet::InterStore(const Slice &dst, const std::vector<KeyWeight>
...
+
+    for (const auto &ms : target_mscores) {
+      if (dst_zset.find(ms.member) == dst_zset.end()) continue;
+      member_counters[ms.member]++;
+      double score = ms.score * keys_weights[i].weight;
+      if (std::isnan(score)) score = 0;
+      switch (aggregate_method) {
+        case kAggregateSum:
+          dst_zset[ms.member] += score;
+          if (std::isnan(dst_zset[ms.member])) {
```
torwig, line 742: Just curiosity: in what circumstances we possibly can get `NaN` here?

src/types/redis_zset.cc
torwig: @IoCing Not via **comment** but use a proper **name** which will describe the **purpose** of the variable, what the variable **does**. For example, `have_intersection`.

src/types/redis_zset.cc
```diff
@@ -707,6 +707,109 @@ rocksdb::Status ZSet::InterStore(const Slice &dst, const std::vector<KeyWeight>
...
+    member_counters[ms.member] = 1;
+  }
+
+  for (size_t i = 1; i < keys_weights.size(); i++) {
+    s = RangeByScore(keys_weights[i].key, spec, &target_mscores, &target_size);
+    if (!s.ok() || target_mscores.empty()) return s;
+    // Judging whether this cycle can find the intersection
+    bool flag = false;
+    for (const auto &ms : target_mscores) {
+      if (dst_zset.find(ms.member) == dst_zset.end()) continue;
```
torwig: I'd recommend you to have a look at the implementation of `SINTERCARD` here https://github.com/apache/kvrocks/blob/unstable/src/types/redis_set.cc#L390 . Specifically, I want to stress using an iterator and then incrementing the value of the counter using the iterator.

torwig (changes requested):

src/types/redis_zset.cc
IoCing: Sorry, I don't know. I just copy the implement of interstore

src/types/redis_zset.cc
IoCing: thanks,I have modified the code

src/types/redis_zset.cc
```diff
@@ -707,6 +707,104 @@ rocksdb::Status ZSet::InterStore(const Slice &dst, const std::vector<KeyWeight>
...
+  if (!s.ok() || target_mscores.empty()) return s;
+
+  for (const auto &ms : target_mscores) {
+    member_counters[ms.member] = 1;
+  }
+
+  for (size_t i = 1; i < keys_weights.size(); i++) {
+    s = RangeByScore(keys_weights[i].key, spec, &target_mscores, &target_size);
+    if (!s.ok() || target_mscores.empty()) return s;
+    // Judging whether this cycle can find the intersection
```
torwig: @IoCing Now you can safely remove this comment and another one at line 792.

src/commands/cmd_zset.cc
```diff
@@ -1340,6 +1469,8 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandZAdd>("zadd", -4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandZCount>("zcount", 4, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandZIncrBy>("zincrby", 4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandZInterStore>("zinterstore", -4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandZInter>("zinter", -3, "read-only", 1, 1, 1),
```
infdahai, line 1472: It seems the keyrange should be processed with `numkeys_`.

src/types/redis_zset.cc
```diff
@@ -707,6 +707,102 @@ rocksdb::Status ZSet::InterStore(const Slice &dst, const std::vector<KeyWeight>
...
+  }
+
+  for (const auto &iter : dst_zset) {
+    if (member_counters[iter.first] != keys_weights.size()) continue;
+    mscores->emplace_back(MemberScore{iter.first, iter.second});
+  }
+  return rocksdb::Status::OK();
+}
+
+rocksdb::Status ZSet::InterCard(const std::vector<KeyWeight> &keys_weights, uint64_t limit, uint64_t *saved_cnt) {
```
torwig, line 767: To be completely sure, I'd also check if the `keys_weights` is empty and make an early return because later we use the `keys_weights[0].key` value and it will lead to a runtime error. The same could be applied to the `Inter` function.

IoCing: ok，i will check it

src/commands/cmd_zset.cc
```diff
@@ -1311,6 +1311,135 @@ class CommandZInterStore : public CommandZUnionStore {
...
+        }
+        i += numkeys_ + 1;
+      } else if (util::ToLower(args[i]) == "withscores") {
+        with_scores_ = true;
+        i++;
+      } else {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+    }
+    return Commander::Parse(args);
```
PragmaTwice, line 1364:
```suggestion
    CommandParser parser(args, 1);
    numkeys_ = GET_OR_RET(parser.TakeInt<int>());

    for (size_t i = 0; i < numkeys_; ++i) {
      keys_weights_.emplace_back(GET_OR_RET(parser.TakeStr()), 1);
    }

    while (parser.Good()) {
      if (parser.EatEqICase("aggregate")) {
        if (parser.EatEqICase("sum")) {
          aggregate_method_ = kAggregateSum;
        } else if (parser.EatEqICase("min")) {
          aggregate_method_ = kAggregateMin;
        } else if (parser.EatEqICase("max")) {
          aggregate_method_ = kAggregateMax;
        } else {
          return {Status::RedisParseErr, "aggregate param error"};
        }
      } else if (parser.EatEqICase("weights")) {
        for (size_t i = 0; i < numkeys_; ++i) {
          auto weight = parser.TakeFloat();
          if (!weight || std::isnan(*weight)) {
            return {Status::RedisParseErr, "weight is not a double or out of range"};
          }
          keys_weights_[i].weight = *weight;
        }
      } else if (parser.EatEqICase("withscores")) {
        with_scores_ = true;
      } else {
        return {Status::RedisParseErr, errInvalidSyntax};
      }
    }
    return Status::OK();
```

git-hulk:
Has been done in #1983, close this PR.

PR #1518, opened 2023-06-24
Parse the timeout parameter as double in the blocking command 

### LINKED ISSUE #1503: Convert the timeout parameter from an integer to a double value
### Search before asking

### Motivation

In redis, the timeout parameter of commands like [BLPOP](https://redis.io/commands/blpop/) is a double value, but in kvrocks, the parameter must be an integer.

### Solution

We can change `parseint` in the Parse function into `parsefloat`.

Yangsx-1:
Close #1503 
Convert timeout parameter in commands blpop/brpop, bzpopmin/max and bzmpop from int to double.

### DIFF
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -204,7 +204,7 @@ class CommandBPop : public Commander,
     if (timeout_) {
       timer_.reset(NewTimer(bufferevent_get_base(bev)));
       int64_t timeout_second = timeout_ / 1000 / 1000;
-      int64_t timeout_microsecond = timeout_ - timeout_second * 1000 * 1000;
+      int64_t timeout_microsecond = timeout_ % (1000 * 1000);
       timeval tm = {timeout_second, static_cast<int>(timeout_microsecond)};
       evtimer_add(timer_.get(), &tm);
     }
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -174,7 +174,7 @@ class CommandBPop : public Commander,
       return {Status::RedisParseErr, "timeout should not be negative"};
     }
 
-    timeout_ = static_cast<int64_t>(*parse_result * 1000 * 1000);  // microsecond
+    timeout_ = static_cast<int64_t>(*parse_result * 1000 * 1000);
 
     keys_ = std::vector<std::string>(args.begin() + 1, args.end() - 1);
     return Commander::Parse(args);
@@ -205,7 +205,7 @@ class CommandBPop : public Commander,
       timer_.reset(NewTimer(bufferevent_get_base(bev)));
       int64_t timeout_second = timeout_ / 1000 / 1000;
       int64_t timeout_microsecond = timeout_ - timeout_second * 1000 * 1000;
-      timeval tm = {timeout_second, timeout_microsecond};
+      timeval tm = {timeout_second, static_cast<int>(timeout_microsecond)};
       evtimer_add(timer_.get(), &tm);
     }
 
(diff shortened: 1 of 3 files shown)


src/commands/cmd_list.cc
```diff
@@ -284,7 +286,7 @@ class CommandBPop : public Commander,
 
  private:
   bool left_ = false;
-  int timeout_ = 0;  // seconds
+  int64_t timeout_ = 0;  // microsecond
```
torwig: `microsecond` => `microseconds` with `s` and in similar places.

src/commands/cmd_list.cc
```diff
@@ -165,16 +165,16 @@ class CommandBPop : public Commander,
...
-      return {Status::RedisParseErr, "timeout is not an integer or out of range"};
+      return {Status::RedisParseErr, errTimeoutIsNotFloat};
     }
 
     if (*parse_result < 0) {
       return {Status::RedisParseErr, "timeout should not be negative"};
     }
 
-    timeout_ = *parse_result;
+    timeout_ = static_cast<int64_t>(*parse_result * 1000 * 1000);  // microsecond
```
torwig: I think here you can remove the comment. In same places as well.

Yangsx-1:
@torwig There is something wrong in CI with Darwin Clang, what can i do to deal with it?

torwig:
@Yangsx-1 You can try casting `timeout_microsecond` to `int` by using static_cast<int>(timeout_microsecond) because according to the warning, the second argument should be `int` not `int64`.

Yangsx-1:
> @Yangsx-1 You can try casting `timeout_microsecond` to `int` by using static_cast(timeout_microsecond) because according to the warning, the second argument should be `int` not `int64`.

The range of int is a little small, i think using millisecond is better than microsecond.

torwig:
@Yangsx-1 According to this:

```
struct timeval {
   time_t      tv_sec;   // Number of whole seconds of elapsed time
   long int    tv_usec;  // Number of microseconds of rest of elapsed time minus tv_sec. Always less than one million
};
```

The number of microseconds is always less than one million so `int` is OK. Or what is your concern?

Yangsx-1:
> @Yangsx-1 You can try casting `timeout_microsecond` to `int` by using static_cast(timeout_microsecond) because according to the warning, the second argument should be `int` not `int64`.

The range of int is a little small, there may be a chance to overflow for microsecond, but it seems like we can only use int.

Yangsx-1:
> @Yangsx-1 According to this:
> 
> ```
> struct timeval {
>    time_t      tv_sec;   // Number of whole seconds of elapsed time
>    long int    tv_usec;  // Number of microseconds of rest of elapsed time minus tv_sec. Always less than one million
> };
> ```
> 
> The number of microseconds is always less than one million so `int` is OK. Or what is your concern?

Oh, i understand. There is no need to worry about overflow!

torwig (dismissed):
LGTM.
@Yangsx-1 Thank you for your contribution.

git-hulk (dismissed):

src/commands/cmd_list.cc
```diff
@@ -203,7 +203,9 @@ class CommandBPop : public Commander,
 
     if (timeout_) {
       timer_.reset(NewTimer(bufferevent_get_base(bev)));
-      timeval tm = {timeout_, 0};
+      int64_t timeout_second = timeout_ / 1000 / 1000;
+      int64_t timeout_microsecond = timeout_ - timeout_second * 1000 * 1000;
```
PragmaTwice:
```suggestion
      int64_t timeout_microsecond = timeout_second % (1000 * 1000);
```

torwig (approved):

git-hulk (approved):

git-hulk:
Thanks all, merging...

PR #1533, opened 2023-06-28
Add support for GEOSEARCH and GEOSEARCHSTORE

### LINKED ISSUE #1511: Add the support of the GEOSEARCH, GEOSEARCHSTORE comands
### Search before asking

### Motivation

This issue considers to support [geosearch](https://redis.io/commands/geosearch/), [geosearchstore](https://redis.io/commands/geosearchstore/) commands.

### Solution

_No response_

uds5501:
Closes #1511

### DIFF
--- a/src/commands/cmd_geo.cc
+++ b/src/commands/cmd_geo.cc
@@ -438,7 +438,7 @@ class CommandGeoSearch : public CommandGeoBase {
     if (!s.ok()) {
       return {Status::RedisExecErr, s.ToString()};
     }
-    *output = GenerateOutput(geo_points);
+    *output = generateOutput(geo_points);
     // storing comes later.
     return Status::OK();
   }
@@ -496,7 +496,7 @@ class CommandGeoSearch : public CommandGeoBase {
     return Status::OK();
   }
 
-  std::string GenerateOutput(const std::vector<GeoPoint> &geo_points) {
+  std::string generateOutput(const std::vector<GeoPoint> &geo_points) {
     int result_length = static_cast<int>(geo_points.size());
     int returned_items_count = (count_ == 0 || result_length < count_) ? result_length : count_;
     std::vector<std::string> list;
@@ -531,8 +531,6 @@ class CommandGeoSearch : public CommandGeoBase {
 
 class CommandGeoSearchStore : public CommandGeoSearch {
  public:
-  CommandGeoSearchStore() : CommandGeoSearch() {}
-
   Status Parse(const std::vector<std::string> &args) override {
     CommandParser parser(args, 1);
     store_key_ = GET_OR_RET(parser.TakeStr());
--- a/src/commands/cmd_geo.cc
+++ b/src/commands/cmd_geo.cc
@@ -387,14 +387,14 @@ class CommandGeoSearch : public CommandGeoBase {
         s = ValidateLongLat(&longitude_, &latitude_);
         if (!s.IsOK()) return s;
       } else if (parser.EatEqICase("byradius")) {
-        auto s = setShapeType(CIRCULAR);
+        auto s = setShapeType(kGeoShapeTypeCircular);
         if (!s.IsOK()) return s;
         radius_ = GET_OR_RET(parser.TakeFloat());
         std::string distance_raw = GET_OR_RET(parser.TakeStr());
         s = ParseDistanceUnit(distance_raw);
         if (!s.IsOK()) return s;
       } else if (parser.EatEqICase("bybox")) {
-        auto s = setShapeType(RECTANGULAR);
+        auto s = setShapeType(kGeoShapeTypeRectangular);
         if (!s.IsOK()) return s;
         width_ = GET_OR_RET(parser.TakeFloat());
         height_ = GET_OR_RET(parser.TakeFloat());
@@ -454,12 +454,12 @@ class CommandGeoSearch : public CommandGeoBase {
   std::string member_;
   std::string key_;
   DistanceSort sort_ = kSortNone;
-  GeoShapeType shape_type_ = NONE;
+  GeoShapeType shape_type_ = kGeoShapeTypeNone;
   OriginPointType origin_point_type_ = kNone;
   GeoShape geo_shape_;
 
   Status setShapeType(GeoShapeType shape_type) {
-    if (shape_type_ != NONE) {
+    if (shape_type_ != kGeoShapeTypeNone) {
       return {Status::RedisParseErr, "please use only one of BYBOX or BYRADIUS"};
     }
     shape_type_ = shape_type;
@@ -475,13 +475,13 @@ class CommandGeoSearch : public CommandGeoBase {
   }
 
   Status createGeoShape() {
-    if (shape_type_ == NONE) {
+    if (shape_type_ == kGeoShapeTypeNone) {
       return {Status::RedisParseErr, "please use BYBOX or BYRADIUS"};
     }
     geo_shape_.type = shape_type_;
     geo_shape_.conversion = GetUnitConversion();
 
-    if (shape_type_ == CIRCULAR) {
+    if (shape_type_ == kGeoShapeTypeCircular) {
       geo_shape_.radius = radius_;
     } else {
       geo_shape_.width = width_;
@@ -503,7 +503,7 @@ class CommandGeoSearch : public CommandGeoBase {
     for (int i = 0; i < returned_items_count; i++) {
       auto geo_point = geo_points[i];
       if (!with_coord_ && !with_hash_ && !with_dist_) {
-        output[i] = redis::BulkString(geo_point.member);
+        output.emplace_back(redis::BulkString(geo_point.member));
       } else {
         std::vector<std::string> one;
         one.emplace_back(redis::BulkString(geo_point.member));
@@ -517,7 +517,7 @@ class CommandGeoSearch : public CommandGeoBase {
           one.emplace_back(redis::MultiBulkString(
               {util::Float2String(geo_point.longitude), util::Float2String(geo_point.latitude)}));
         }
-        output[i] = redis::Array(one);
+        output.emplace_back(redis::Array(one));
       }
     }
     return redis::Array(output);
@@ -550,14 +550,14 @@ class CommandGeoSearchStore : public CommandGeoSearch {
         s = ValidateLongLat(&longitude_, &latitude_);
         if (!s.IsOK()) return s;
       } else if (parser.EatEqICase("byradius")) {
-        auto s = setShapeType(CIRCULAR);
+        auto s = setShapeType(kGeoShapeTypeCircular);
         if (!s.IsOK()) return s;
         radius_ = GET_OR_RET(parser.TakeFloat());
         std::string distance_raw = GET_OR_RET(parser.TakeStr());
         s = ParseDistanceUnit(distance_raw);
         if (!s.IsOK()) return s;
       } else if (parser.EatEqICase("bybox")) {
-        auto s = setShapeType(RECTANGULAR);
+        auto s = setShapeType(kGeoShapeTypeRectangular);
         if (!s.IsOK()) return s;
         width_ = GET_OR_RET(parser.TakeFloat());
         height_ = GET_OR_RET(parser.TakeFloat());
--- a/src/commands/cmd_geo.cc
+++ b/src/commands/cmd_geo.cc
@@ -19,6 +19,7 @@
  */
 
 #include "commander.h"
+#include "command_parser.h"
 #include "error_constants.h"
 #include "server/server.h"
 #include "types/redis_geo.h"
@@ -52,11 +53,16 @@ class CommandGeoBase : public Commander {
     *longitude = *long_stat;
     *latitude = *lat_stat;
 
-    if (*longitude < GEO_LONG_MIN || *longitude > GEO_LONG_MAX || *latitude < GEO_LAT_MIN || *latitude > GEO_LAT_MAX) {
-      return {Status::RedisParseErr, "invalid longitude,latitude pair " + longitude_para + "," + latitude_para};
-    }
+    auto s = ValdiateLongLat(longitude, latitude);
+    if (!s.OK()) return s;
 
     return Status::OK();
+  }
+
+  static Status ValdiateLongLat(double *longitude, double *latitude) {
+    if (*longitude < GEO_LONG_MIN || *longitude > GEO_LONG_MAX || *latitude < GEO_LAT_MIN || *latitude > GEO_LAT_MAX) {
+      return {Status::RedisParseErr, "invalid longitude,latitude pair"};
+    }
   }
 
   double GetDistanceByUnit(double distance) { return distance / GetUnitConversion(); }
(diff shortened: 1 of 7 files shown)


uds5501:
[WIP] Completed the BYBOX implementation.

TODO

- [x] Deprecate old code which uses RADIUS-specific functions, instead use by shape alternatives and verify that tests don't break
- [x] Perform a lint.
- [x] Write tests for GeoSearch

uds5501:
[WIP] Bug fixed in the radius section of RADIUS searches.

- [x] Work on GEOSEARCH BYBOX tests
- [x] Find the bug and check if expected implementation work similar to redis search.
- [x] Implement GEOSEARCHSTORE.

src/types/redis_geo.cc
```diff
@@ -85,11 +85,18 @@ rocksdb::Status Geo::Radius(const Slice &user_key, double longitude, double lati
...
+  GeoShape geo_shape;
+  geo_shape.type = CIRCULAR;
+  geo_shape.xy[0] = longitude;
+  geo_shape.xy[1] = latitude;
+  geo_shape.radius = radius_meters;
+  geo_shape.conversion = 1;
+
   /* Get all neighbor geohash boxes for our radius search */
-  GeoHashRadius georadius = GeoHashHelper::GetAreasByRadiusWGS84(longitude, latitude, radius_meters);
+  GeoHashRadius georadius = GeoHashHelper::GetAreasByShapeWGS84(&geo_shape);
```
uds5501: @infdahai , @git-hulk I wanted an opinion on this.

Should I change the existing georadius functions to use the new shape specific functions or should I let them be for now and they will be taken as part of a different refactor?

If taken as different refactor, I'll revert this piece of code to use `GetAreasByRadiusWGS84`.

uds5501: Ps: This new implementation is different, hence breaking some tests. (TestGeo/GEORANGE_STORE_option:_plain_usage)

PragmaTwice: I think it is ok to refactor them if it is convenient.

uds5501:
@git-hulk @PragmaTwice @infdahai , this MR is ready for review, thanks for your patience :D

PragmaTwice:
> @git-hulk @PragmaTwice @infdahai , this MR is ready for review, thanks for your patience :D

Congrats! It seems you need firstly resolve the conflict in geo_test.go.

uds5501:
@PragmaTwice done.

src/types/geohash.cc
```diff
@@ -329,47 +329,47 @@ uint8_t GeoHashHelper::EstimateStepsByRadius(double range_meters, double lat) {
...
- * However, for instance, a min_lon of 7.680495 is not correct, because the
- * point -1.27579540014266968 61.33421815228281559 is at less than 7000
- * kilometers away.
- *
- * Since this function is currently only used as an optimization, the
- * optimization is not used for very big radiuses, however the function
- * should be fixed. */
-int GeoHashHelper::BoundingBox(double longitude, double latitude, double radius_meters, double *bounds) {
+ * maximum latitude. */
+int GeoHashHelper::BoundingBox(GeoShape *geo_shape, double *bounds) {
```
PragmaTwice: I think we'd better pass `const GeoShape&` here if the `geo_shape` cannot be null and has not been modifed in the function.

src/types/geohash.cc
```diff
@@ -329,47 +329,47 @@ uint8_t GeoHashHelper::EstimateStepsByRadius(double range_meters, double lat) {
...
+  bounds[2] = is_in_southern_hemisphere ? longitude + long_delta_bottom : longitude + long_delta_top;
+  bounds[1] = latitude - lat_delta;
+  bounds[3] = latitude + lat_delta;
   return 1;
 }
 
-/* Return a set of areas (center + 8) that are able to cover a range query
- * for the specified position and radius. */
-GeoHashRadius GeoHashHelper::GetAreasByRadius(double longitude, double latitude, double radius_meters) {
+GeoHashRadius GeoHashHelper::GetAreasByShapeWGS84(GeoShape *geo_shape) {
```
PragmaTwice: ditto

src/types/redis_geo.h
```diff
@@ -68,24 +70,27 @@ class Geo : public ZSet {
...
+                          GeoShape &geo_shape);
   static void scoresOfGeoHashBox(GeoHashBits hash, GeoHashFix52Bits *min, GeoHashFix52Bits *max);
-  int getPointsInRange(const Slice &user_key, double min, double max, double lon, double lat, double radius,
+  int getPointsInRange(const Slice &user_key, double min, double max, GeoShape &geo_shape,
                        std::vector<GeoPoint> *geo_points);
   static bool appendIfWithinRadius(std::vector<GeoPoint> *geo_points, double lon, double lat, double radius,
                                    double score, const std::string &member);
-
+  static bool appendIfWithinShape(std::vector<GeoPoint> *geo_points, GeoShape &geo_shape, double score,
+                                  const std::string &member);
```
PragmaTwice: ditto

src/types/geohash.cc
uds5501: The `*bounds` is  Geoshape's bound itself, hence geo_shape is being changed.
I should probably refactor this function to reflect this.

src/types/geohash.cc
uds5501: ack, addressing this.

git-hulk:
Sorry for pending so long, will review in a few days.

torwig:
@uds5501 Could you please run the `./x.py format` command to format the code with `clang`?

git-hulk:
@uds5501 PR generally looks good to me, can help to format like what @torwig mentioned.

uds5501:
@git-hulk / @torwig will update this today.

uds5501:
Apologies for the delay, @git-hulk @torwig could you please take a look now?

uds5501:
Rebased branch, @git-hulk / @torwig / @PragmaTwice please take a look whenever you get time, thank you!

src/commands/cmd_geo.cc
```diff
@@ -52,10 +54,16 @@ class CommandGeoBase : public Commander {
...
+    auto s = ValdiateLongLat(longitude, latitude);
+    if (!s.OK()) return s;
+
+    return Status::OK();
+  }
+
+  static Status ValdiateLongLat(double *longitude, double *latitude) {
     if (*longitude < GEO_LONG_MIN || *longitude > GEO_LONG_MAX || *latitude < GEO_LAT_MIN || *latitude > GEO_LAT_MAX) {
-      return {Status::RedisParseErr, "invalid longitude,latitude pair " + longitude_para + "," + latitude_para};
+      return {Status::RedisParseErr, "invalid longitude,latitude pair"};
```
torwig: I think it's a good idea to include longitude and latitude values in the error message.

src/commands/cmd_geo.cc
```diff
@@ -52,10 +54,16 @@ class CommandGeoBase : public Commander {
     *longitude = *long_stat;
     *latitude = *lat_stat;
 
+    auto s = ValdiateLongLat(longitude, latitude);
+    if (!s.OK()) return s;
+
+    return Status::OK();
+  }
+
+  static Status ValdiateLongLat(double *longitude, double *latitude) {
```
torwig: Typo (between I and D) in the function name. Should be `ValidateLongLat`.

git-hulk (dismissed):

src/commands/cmd_geo.cc
```diff
@@ -355,6 +363,249 @@ class CommandGeoRadius : public CommandGeoBase {
...
+  double longitude_ = 0;
+  double latitude_ = 0;
+  std::string member_;
+  std::string key_;
+  DistanceSort sort_ = kSortNone;
+  GeoShapeType shape_type_ = NONE;
+  OriginPointType origin_point_type_ = kNone;
+  GeoShape geo_shape_;
+  std::string store_key_;
+  bool store_distance_ = false;
```
torwig: This variable isn't used in the `GEOSEARCH` command.

src/commands/cmd_geo.cc
```diff
@@ -355,6 +363,249 @@ class CommandGeoRadius : public CommandGeoBase {
...
+    redis::Geo geo_db(svr->storage, conn->GetNamespace());
+
+    auto s = geo_db.Search(args_[1], geo_shape_, origin_point_type_, member_, count_, sort_, store_key_, false,
+                           GetUnitConversion(), &geo_points);
+
+    if (!s.ok()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+    *output = generateOutput(geo_points);
+    // storing comes later.
```
torwig: I think this comment is useless.

src/commands/cmd_geo.cc
```diff
@@ -355,6 +363,249 @@ class CommandGeoRadius : public CommandGeoBase {
...
+  std::string key_;
+  DistanceSort sort_ = kSortNone;
+  GeoShapeType shape_type_ = NONE;
+  OriginPointType origin_point_type_ = kNone;
+  GeoShape geo_shape_;
+  std::string store_key_;
+  bool store_distance_ = false;
+
+  Status setShapeType(GeoShapeType shape_type) {
+    if (shape_type_ != NONE) {
```
torwig: I just scratch my head when I see this condition. Perhaps you wanted to check the `if (shape_type == NONE)` here? Because currently you check the already set value of the `shape_type_` variable (pay attention to the `_` character at the end of the name).

torwig (changes requested):

src/commands/cmd_geo.cc
torwig: Or perhaps the error message isn't correct. If you want to check whether the `shape_type_` variable was already set and this is a repetitive attempt, then the error message should be appropriate.

src/commands/cmd_geo.cc
```diff
@@ -355,6 +363,249 @@ class CommandGeoRadius : public CommandGeoBase {
...
+      return {Status::RedisParseErr, "please use only one of BYBOX or BYRADIUS"};
+    }
+    shape_type_ = shape_type;
+    return Status::OK();
+  }
+
+  Status setOriginType(OriginPointType origin_point_type) {
+    if (origin_point_type_ != kNone) {
```
torwig, line 467: The same consideration as with the `setShapeType` function.

src/commands/cmd_geo.cc
```diff
@@ -355,6 +363,249 @@ class CommandGeoRadius : public CommandGeoBase {
...
+      geo_shape_.xy[0] = longitude_;
+      geo_shape_.xy[1] = latitude_;
+    }
+    return Status::OK();
+  }
+
+  std::string generateOutput(const std::vector<GeoPoint> &geo_points) {
+    int result_length = static_cast<int>(geo_points.size());
+    int returned_items_count = (count_ == 0 || result_length < count_) ? result_length : count_;
+    std::vector<std::string> list;
```
torwig: Could we call the `reserve` function on the `list` variable with the `returned_items_count` parameter here?
Also, you can rename the `list` variable to something like `output`/`response`.

src/commands/cmd_geo.cc
torwig: Actually, you don't use the `store_key_` variable in the `GEOSEARCH` command so you can delete it and use just `""` as a function parameter later. But this leads us to the situation when the `Geo::Search` function in the case of the `GEOSEARCH` command doesn't use 2 input parameters so maybe we can have two separate functions instead of one `Geo::Search`.

uds5501: Agreed, removing

src/commands/cmd_geo.cc
uds5501: The thought process here is that only one of BYBOX or BYRADIUS should be present.
Initially the internal `shape_type_` is none, 
- suppose I parse `BYBOX` first, I'll verify first if the shape is not already set to something else, if not then I can set it to RECTANGULAR.
- Now suppose I parse `BYRADIUS` in the command, I'll try overriding the shape to CIRCULAR which shouldn't be allowed. Hence sending the mentioned error message.

src/commands/cmd_geo.cc
uds5501: Similar thought process to differentiate between a longlat type vs member type instruction.

src/commands/cmd_geo.cc
uds5501: Hmmm, agreed. One could be a simple search and one could be search store. Refactoring this.

uds5501: Hmmm, agreed. One could be a simple search and one could be search store. Refactoring this.

uds5501:
@torwig thanks for taking out time to review, I have made changes and replied to a couple of comments, kindly check.

src/commands/cmd_geo.cc
torwig: Now I got your point.

src/commands/cmd_geo.cc
```diff
@@ -355,6 +364,249 @@ class CommandGeoRadius : public CommandGeoBase {
...
+
+  std::string generateOutput(const std::vector<GeoPoint> &geo_points) {
+    int result_length = static_cast<int>(geo_points.size());
+    int returned_items_count = (count_ == 0 || result_length < count_) ? result_length : count_;
+    std::vector<std::string> output;
+    output.reserve(returned_items_count);
+    for (int i = 0; i < returned_items_count; i++) {
+      auto geo_point = geo_points[i];
+      if (!with_coord_ && !with_hash_ && !with_dist_) {
+        output[i] = redis::BulkString(geo_point.member);
```
torwig: I'm afraid this line will crash because the `reserve` function doesn't change the vector's size, it only reserves the so-called capacity.

torwig: I think you can leave `push_back/emplace_back` here.

src/types/geohash.h
```diff
@@ -59,6 +59,8 @@ enum GeoDirection {
   GEOHASH_NORT_EAST
 };
 
+enum GeoShapeType { NONE = 0, CIRCULAR, RECTANGULAR };
```
torwig: You can name enum values using the `kGeoShapeType` prefix: `kGeoShapeTypeNone`, `kGeoShapeTypeCircular`, `kGeoShapeTypeRectangular`. Or use an enum class (for example, `GeoShapeType::kNone` etc.)

src/commands/cmd_geo.cc
uds5501: ah my bad, fixing this.

src/types/geohash.h
uds5501: ack, renaming it

uds5501:
@torwig updated the MR with suggestions, please check.

torwig (dismissed):
Generally LGTM.

src/commands/cmd_geo.cc
```diff
@@ -52,10 +54,17 @@ class CommandGeoBase : public Commander {
     *longitude = *long_stat;
     *latitude = *lat_stat;
 
+    auto s = ValidateLongLat(longitude, latitude);
+    if (!s.OK()) return s;
+
+    return Status::OK();
```
git-hulk:
```suggestion
    return ValidateLongLat(longitude, latitude);
```

git-hulk (approved):

torwig (approved):

git-hulk:
Thanks all, merging...

git-hulk:
@uds5501 Thanks for your support again!

Issue #1223, opened 2023-01-11
Improve slot migration speed and resource consumption using raw key values


caipengbo:
### Motivation

Currently, kvrocks performs slot migration by converting the data into RESP and then sending it to the destination(#430).
We made a little optimizations in #904 to speed up migration.

However, there are still some problems with command-based slot migration:
- we will determine the expire time during the data migration, which makes the data inconsistent (mentioned in the comments of #906)
- for data with an expiration time, if the expiration occurs during the migration process, then we cannot migrate the data and the entire migration process will fail.

### Solution

We can use **rawkv-based** data migration, where we directly send rocksdb's key values data to the target.

With rawkv-based migration, neither the source nor the destination need to judge expire time, which completely solves the data inconsistency problem. It is equivalent to translating the raw key values data from one instance to another. 

It can save a lot of CPU by eliminating the need to convert Key-Value to the RESP. 

In my tests, it can:
- save up to **2x** CPU on the target side
- for small values(100byte), the rawkv-based migration is 2.75 times faster than the command-based migration

### Plan

I decomposed this issues into those tasks:

- [x] #1982
- [x] #1989
- [x] #2007
- [x] #2008
- [x] #2009

git-hulk:
@caipengbo Do you mean that we can send the write batch directly instead of the RESP format？

caipengbo:
> Do you mean that we can send the write batch directly instead of the RESP format？

@git-hulk Yup, we send the write batch directly, and the target calls the `rocksdb::Write()` interface directly

git-hulk:
> > Do you mean that we can send the write batch directly instead of the RESP format？
> 
> @git-hulk Yup, we send the write batch directly, and the target calls the `rocksdb::Write()` interface directly

Got it, thanks a lot

ZENOTME:
I'm interested in this issue, but I'm confused about 

> With rawkv-based migration, neither the source nor the destination need to judge expire time
>> Do you mean that we can send the write batch directly instead of the RESP format？

Does it means that: for a key-value pair with expire (k,v,expire), just send the (k,v) and ignore the expire?

caipengbo:
> Does it means that: for a key-value pair with expire (k,v,expire), just send the (k,v) and ignore the expire?

Yes, it ignore the expire. Let the target determine the expire.

ZENOTME:
> Let the target determine the expire.

Do you means the case like:
1. migrate (k,v) to target, and now (k,v) don't have expire
[...]

PR #1534, opened 2023-06-30
Support rapid slot migration based on raw key value

caipengbo:
This PR introduces a new migration type called **rawkv-based migration**(in code named `raw_key_value`), which directly sends  rocksdb's key values to the target.

As mentioned in in #1223, rawkv-based migration does not need to determine the expiration time during migration, which solves the data consistency problem and avoids the migration failure problem that frequently occurs when migrating data to be expired.

It can save a lot of CPU by eliminating the need to convert Key-Value to the RESP. In my tests, it can save up to **2x** CPU on the target side; for small values(100byte), the rawkv-based migration is 2.75 times faster than the command-based migration

For the stability and code review simplicity, I did not combine the two migration types in this CR. Instead, I used two different if else branches. I've included the `hiredis` library for interacting with the destination, and in the future we can use `hiredis` to streamline command-based migrations as well.

Here is the basic idea of this pull:
1. Directly migrating the underlying data between different machines may run into clock rollback issues, which may introduce the problem of duplicate versions in complex data types, so before each migration, I use the `TIME` command to check the machine clock gap.
2. In order to avoid incomplete data caused by the expiration of the destination data during the migration process, we will prohibit the expiration of the slot during the migration
3. Instead of using a separate thread to transfer raw key value batches, I used the `BATCHSET` command so that the outside could use this command to inject data directly(may saving some CPU). `BATCHSET` command returns an integer, currently is the return data size, the reason for returning an integer is that I plan to make adaptive throttling.
4. I have introduced an iterator (`MigrateIterator`) and a sender (`BatchSender`). This decouples the traversal and sending, which makes the code clearer.


git-hulk:
cool, thanks for your great efforts!

kvrocks.conf
```diff
@@ -501,7 +501,17 @@ compaction-checker-range 0-7
 #
 # rename-command KEYS ""
 
-################################ MIGRATE #####################################
+################################ MIGRATION #####################################
+# Slot migration supports two ways:
+# - redis_command: Migrate data by redis serialization protocol(RESP).
+# - raw_key_value: Migrate the raw key value data of the storage engine directly.
```
PragmaTwice: `redis-command` and `raw-key-value`?

src/cluster/slot_migrate.h
```diff
@@ -169,6 +195,7 @@ class SlotMigrator : public redis::Database {
   std::string dst_ip_;
   int dst_port_ = -1;
   UniqueFD dst_fd_;
+  redisContext *dst_redis_context_ = nullptr;
```
PragmaTwice, line 198: It seems the redisFree call for deallocating dst_redis_context_ is missing?

src/cluster/batch_sender.cc
```diff
@@ -0,0 +1,109 @@
...
+  if (redis_context == nullptr) {
+    return {Status::NotOK, "redis context is null"};
+  }
+
+  auto *reply = static_cast<redisReply *>(
+      redisCommand(redis_context, "BATCHSET %d %b", slot, write_batch.Data().c_str(), write_batch.GetDataSize()));
+  auto exit = MakeScopeExit([reply] {
+    if (reply != nullptr) {
+      freeReplyObject(reply);
+    }
```
PragmaTwice, line 93: nullptr check seems unnecessary here

PragmaTwice (commented):
I recommend you to add unique_ptr wrappers for redisContext and redisReply pointers. You can refer to [event_util.h](https://github.com/apache/kvrocks/blob/unstable/src/common/event_util.h).

kvrocks.conf
caipengbo: I want to distinguish the item value from the item name

PragmaTwice: I mean dash instead of underlines

PragmaTwice:
It seems some memory leaks are found by ASAN.

https://github.com/apache/kvrocks/actions/runs/5527836049/jobs/10085132766?pr=1534

caipengbo:
> I recommend you to add unique_ptr wrappers for redisContext and redisReply pointers. You can refer to [event_util.h](https://github.com/apache/kvrocks/blob/unstable/src/common/event_util.h).

@PragmaTwice I wanted to use unique_ptr to encapsulate `redisContext` and `redisReply`, but I found that it is only used for free. Currently hiredis does not use much, maybe simply handle it. We can better encapsulate hiredis when we use it more in the future.

src/cluster/batch_sender.cc
```diff
@@ -0,0 +1,109 @@
...
+  });
+
+  if (redis_context->err != 0) {
+    return {Status::NotOK, std::string(redis_context->errstr)};
```
PragmaTwice, line 97:
```suggestion
    return {Status::NotOK, redis_context->errstr};
```

src/cluster/batch_sender.cc
PragmaTwice: seems the check still exists here

src/cluster/batch_sender.cc
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  if (reply == nullptr) {
+    return {Status::NotOK, "get null reply"};
+  }
+
+  if (reply->type == REDIS_REPLY_ERROR) {
+    auto error_str = std::string(reply->str);
+    return {Status::NotOK, error_str};
```
PragmaTwice, line 106:
```suggestion
    return {Status::NotOK, reply->str};
```

PragmaTwice: actually i do not understand why construct a temp string here

src/cluster/slot_migrate.cc
```diff
@@ -1110,3 +1162,224 @@ void SlotMigrator::resumeSyncCtx(const Status &migrate_result) {
...
+  if (reply->type == REDIS_REPLY_ERROR) {
+    auto error_str = std::string(reply->str);
+    return {Status::NotOK, error_str};
+  }
+
+  if (reply->elements != 2) {
+    return {Status::NotOK, "get invalid reply from TIME command"};
+  }
+
+  uint64_t dst_timestamp = std::stoul(reply->element[0]->str) * 1000000 + std::stoul(reply->element[1]->str);
```
PragmaTwice, line 1199: stoul will throws exceptions while the parsing fails, so i wonder if the parsing will always succeed here?

caipengbo: Yes, normal `TIME` command will parse successfully.

src/common/io_util.cc
```diff
@@ -404,4 +406,23 @@ Status Write(int fd, const std::string &data) { return WriteImpl<write>(fd, data
...
+    redisFree(*redis_context);
+    return {Status::NotOK, error_str};
+  }
+
+  if (redisSetTimeout(*redis_context, timeval{timeout, 0}) != REDIS_OK) {
+    redisFree(*redis_context);
+    return {Status::NotOK, "set timeout failed"};
+  }
+  return Status::OK();
+}
```
PragmaTwice, line 426:
```suggestion
using StaticRedisFree = StaticFunction<decltype(redisFree), redisFree>;

struct UniqueRedisContext : std::unique_ptr<redisContext, StaticRedisFree> {
  using BaseType = std::unique_ptr<redisContext, StaticRedisFree>;
  using BaseType::BaseType;
};

StatusOr<UniqueRedisContext> CreateRedisContextFromConnectedFd(int fd, int timeout) {
  auto redis_context = UniqueRedisContext{redisConnectFd(fd)};
  if (!redis_context) {
    return {Status::NotOK, "init failed"};
  }

  if (redis_context->err != 0) {
    return {Status::NotOK, redis_context->errstr};
  }

  if (redisSetTimeout(redis_context.get(), timeval{timeout, 0}) != REDIS_OK) {
    return {Status::NotOK, "set timeout failed"};
  }
  return redis_context;
}
```

src/cluster/batch_sender.cc
```diff
@@ -0,0 +1,109 @@
...
+ *
+ */
+
+#include "batch_sender.h"
+
+#include "hiredis.h"
+#include "scope_exit.h"
+
+Status BatchSender::Put(rocksdb::ColumnFamilyHandle *cf, const rocksdb::Slice &key, const rocksdb::Slice &value) {
+  if (pending_entries_ == 0 && !prefix_logdata_.empty()) {
```
git-hulk, line 27: Why only put the log data for the first entry?

src/cluster/migrate_iterator.cc
```diff
@@ -0,0 +1,159 @@
...
+  assert(valid_);
+  return items_;
+}
+
+std::string MigrateIterator::GetLogData() const {
+  assert(valid_);
+  return log_data_;
+}
+
+void MigrateIterator::findMetaData() {
```
git-hulk, line 95: findMetaData => findMetadata

src/cluster/migrate_iterator.cc
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  log_data_ = log_data.Encode();
+
+  items_.push_back(MigrateItem{metadata_cf_, metadata_iter_->key().ToString(), metadata_iter_->value().ToString()});
+
+  if (redis_type == RedisType::kRedisStream) {
+    stream_iter_->Reset();
+  } else {
+    subdata_iter_->Reset();
```
git-hulk, line 114: subdata_iter_ => subkey_iter_ to keep consistent with the column family name

src/cluster/migrate_iterator.cc
```diff
@@ -0,0 +1,159 @@
...
+  redis::WriteBatchLogData log_data(redis_type);
+  if (redis_type == RedisType::kRedisList) {
+    log_data.SetArguments({std::to_string(RedisCommand::kRedisCmdRPush)});
+  }
+
+  log_data_ = log_data.Encode();
+
+  items_.push_back(MigrateItem{metadata_cf_, metadata_iter_->key().ToString(), metadata_iter_->value().ToString()});
+
+  if (redis_type == RedisType::kRedisStream) {
```
git-hulk, line 111: If we can move this reset inside the initSubData?

git-hulk (commented):
It's a bit more complex than my imagination, maybe it needs some time to test on the local.

src/cluster/batch_sender.cc
caipengbo: If the data is too large to fit in one `MigrateBatch`, it needs to be split into multiple `MigrateBatches`, for each `MigrateBatch`, it automatically appends the log data

When adding metadata, we actively call `BatchSender::PutLogData` to add log data.

src/cluster/migrate_iterator.cc
caipengbo: Yes, it looks fine

PragmaTwice:
For introducing hiredis, here are something need to be clarified:
- how is TLS support of hiredis? can we directly pass SSL_CTX and ssl_st to it?
- as hiredis is a c library, we need a c++ wrapper. **kvrocks is not written in C**, and we need to follow modern C++ principles. raw C style is not welcomed.
- will the existing code (like, in replication) be replaced with hiredis? or more concrete, how hiredis can help with the client connection reading/writing?

caipengbo:
`hiredis` isn't used much at the moment, so I've included a lightweight `hiredis`.

There is no TLS requirement, we can simply wrapper `hiredis` now.

In the future, if we need more, we can introduce some more complex clients, such as [redis-plus-plus](https://github.com/sewenew/redis-plus-plus), for now I think a lightweight library `hiredis` will do.

PragmaTwice:
> There is no TLS requirement

TLS support for replication is an ongoing effort, and we will support TLS for slot migration in the future. So we need to consider these problems now, to prevent TLS being more hard to support.

caipengbo:
In the future, if there is a need, we can import other libraries and use this library on other existing code. For now, we can start with a simple library.

PragmaTwice:
> In the future, if there is a need, we can import other libraries and use this library on other existing code.

We cannot introduce libraries so casually.

When introducing a library, we need to make sure that it can indeed solve the requirements/simplify the code and is easy to maintain. We will not randomly introduce a library in the early stages and then introduce other similar libraries later when problems arise: this is not something a well-managed project should do, as it would make the code worse by having multiple libraries doing the same thing. We need to carefully consider before introducing any library.

I am not against introducing hiredis, but I believe that these issues should be considered clearly when introducing it. **Instead of "randomly introducing a simple library and considering what comes next later."**

[hiredis seems to support TLS](https://github.com/redis/hiredis/blob/master/hiredis_ssl.h), but I'm not sure if it can reuse our SSL_CTX instead of reconstructing it.

cc @git-hulk

caipengbo:
We certainly can't just arbitrarily import a library, so I've introduced the **most official,** and simplest library — `hiredis`.

git-hulk:
Yes, need to think through before introducing the new library. Kvrocks's dependencies are more and more, we need to take care of this situation to avoid being out of control.

PragmaTwice:
> We certainly can't just arbitrarily import a library, so I've introduced the most official, and simplest library — hiredis.

I know that hiredis is the official library, but it doesn't seem to answer our question.

I would like to provide more specific explanations and understand the context of introducing hiredis:

In the current codebase, such as in replication, when we need to construct a client connection, we directly build a raw TCP connection and then use our own redis protocol tools (such as redis::SimpleString and other protocol encoding/decoding functions).

These processes are not well encapsulated and there are even some manual protocol constructions, such as `SockSend(..., "+OK\r\n")` (although it is not in a client), which is an existing issue.

Of course, we can introduce a library to solve this problem. However, it is important to consider the aforementioned points when introducing a library:
- in #1630, we add TLS support for client connection. how can we rewrite them in hiredis? (as we do not want to have multiple ways to construct client connection)
- some code style problem as hiredis is a c library
- how to replace the existing code (like, in replication) with hiredis

caipengbo:
I understand what you mean, you want to use a client library to do all the things you mentioned above.

I introduced the `hiredis` just to make it easier to send and receive responses and I've also privately written a multi-threaded version of the kvrocks migration tool that also uses `hiredis`. I'm not thinking that we have to use hiredis to do everything related to the redis protocol and redis parsing in the existing code.

These are the things you mention that I probably don't have the answers to yet, and I don't have the answers yet to figure out how to do them, so introducing a simple, easily replaceable library is the best option for now. If we come up with a better way to handle the things you mentioned above, we can easily replace `hiredis`(reuse is better).

git-hulk:
I'm fine to import the hiredis which can make the read/write operation easier, but for other libraries like redis-plus-plus, it needs the solid reason why we need it.

PR #1537, opened 2023-07-04
Fix stdin usage text in PrintUsage

enjoy-binbin:


torwig (approved):

git-hulk (approved):

src/main.cc
```diff
@@ -144,7 +144,7 @@ static void PrintUsage(const char *program) {
   std::cout << program << " implements the Redis protocol based on rocksdb" << std::endl
             << "Usage:" << std::endl
             << std::left << new_opt << "-c, --config <filename>"
-            << "set config file to <filename>, or `-' for stdin" << std::endl
```
tisonkun, line 147: This is actually some old-fashion style back to the age of UNIX commands or POSIX commands manual.

I'm OK with this change but notice here that it's not a "mistake" to fix :D

git-hulk: 👍 I have never noticed this.

enjoy-binbin: kind of interested in this, do you happen to have a link to it? thanks for pointing it out btw

PR #1539, opened 2023-07-04
Fix the inconsistent response behavior with Redis in memory usage  command

### LINKED ISSUE #1538: Running the `memory usage` command against a non-existent key resulted in an error
### Search before asking

### Version

kvrocks 2.4.0

### Minimal reproduce step

```sh
memory usage key-not-exist
```

return `nil`

### What did you see instead?

```sh
(error) ERROR NotFound: Not found : key-not-exist
```

### Anything Else?

This is not compatible with redis, which return `nil` when the key does not exist, and will cause [asynqmon](https://github.com/hibiken/asynqmon) failed sometimes when using `memory usage` command.

git-hulk:
This fixes #1538

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -289,16 +289,16 @@ class CommandDisk : public Commander {
     auto s = disk_db.Type(args_[2], &type);
     if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
 
-    // Redis returns the Nil string when the key does not exist,
-    // so we need to check the type here to check if the key is not exists.
-    if (type == RedisType::kRedisNone) {
-      *output = redis::NilString();
-      return Status::OK();
-    }
-
     uint64_t result = 0;
     s = disk_db.GetKeySize(args_[2], type, &result);
-    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+    if (!s.ok()) {
+      // Redis returns the Nil string when the key does not exist
+      if (s.IsNotFound()) {
+        *output = redis::NilString();
+        return Status::OK();
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
 
     *output = redis::Integer(result);
     return Status::OK();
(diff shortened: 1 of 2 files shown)


mapleFU:
Seems that when key not found or expired, `Database::Type(const Slice &user_key, RedisType *type)` would return `kRedisNone`, and get `Disk::GetKeySize` with `kRedisNone` would raise not found rather than return `Nil`?

(It's interesting that "type" fixed it in another patch: https://github.com/apache/kvrocks/commit/3cea04515a3d9cf15cd9a968b38cd919f3c5ef8e )

mapleFU (approved):
The patch LGTM. I also checked all place using `Database::Type`, they're:

1. `Set::Move`
2. `List::RPopLPush`
3. `CommandDisk::Execute`
4. `CommandType::Execute`

Seems they have all been handled after this patch

PragmaTwice (dismissed):

src/commands/cmd_server.cc
```diff
@@ -289,6 +289,13 @@ class CommandDisk : public Commander {
...
+    // Redis returns the Nil string when the key does not exist,
+    // so we need to check the type here to check if the key is not exists.
+    if (type == RedisType::kRedisNone) {
+      *output = redis::NilString();
+      return Status::OK();
+    }
+
     uint64_t result = 0;
     s = disk_db.GetKeySize(args_[2], type, &result);
     if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
```
torwig: I would recommend checking the returned status after the call to the `GetKeySize` as well (or only after this call) because, in theory, something can happen between the call to `Type` and `GetKeySize` (key deletion, expiration, etc.).

torwig: @git-hulk Thank you for quick fix :+1:

mapleFU: Sounds ok, RocksDB Snapshot might helps for this, but I after `GetKeySize` would be more easily

git-hulk: > I would recommend checking the returned status after the call to the `GetKeySize` as well (or only after this call) because, in theory, something can happen between the call to `Type` and `GetKeySize` (key deletion, expiration, etc.).

good suggestion.

mapleFU (approved):

enjoy-binbin (approved):

PragmaTwice (approved):

tisonkun (approved):

torwig (approved):

tisonkun:
Merging...

Thanks for the quick fix :D

PR #1541, opened 2023-07-04
Add dynamic key range generator with multiple range output

### LINKED ISSUE #1520: add the skip points for KeyRange
### Search before asking

### Motivation

Do we need to add a position to ignore the numkey integer because we only support a consistent range? For example, we add a wrong key currently in the `watch` command in [zdiffstore](https://redis.io/commands/zdiffstore/) and [zinterstore](https://redis.io/commands/zinterstore/).

### Solution

_No response_

infdahai:
close #1520

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -1692,8 +1692,8 @@ void Server::UpdateWatchedKeysFromArgs(const std::vector<std::string> &args, con
     } else if (attr.key_range.first_key == -2) {
       std::vector<redis::CommandKeyRange> vec_range = attr.key_range_vec_gen(args);
 
-      if (vec_range.front().first_key > 0) {
-        for (const auto &range : vec_range) {
+      for (const auto &range : vec_range) {
+        if (range.first_key > 0) {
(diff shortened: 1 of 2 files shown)


src/server/server.cc
```diff
@@ -1683,12 +1683,21 @@ void Server::UpdateWatchedKeysFromArgs(const std::vector<std::string> &args, con
...
+    } else if (attr.key_range.first_key == -1) {
       redis::CommandKeyRange range = attr.key_range_gen(args);
 
       if (range.first_key > 0) {
         updateWatchedKeysFromRange(args, range);
       }
+    } else if (attr.key_range.first_key == -2) {
+      std::vector<redis::CommandKeyRange> vec_range = attr.key_range_vec_gen(args);
+
+      if (vec_range.front().first_key > 0) {
```
PragmaTwice: The program will abort if `vec_range.size() == 0`. We can check every range in the for loop instead.

PragmaTwice (approved):
LGTM

PragmaTwice:
Thanks for your contribution!

PR #1542, opened 2023-07-05
Fix compile error after upgrading the rocksdb

git-hulk:
It got the error: `clang: error: the clang compiler does not support '-march=OFF'` after upgrading the rocksdb in macOS. It should be related to #1516


git-hulk:
cc @aleksraiden

torwig (approved):

aleksraiden:
LGTM, thanks!

PR #1546, opened 2023-07-05
Fix GETEX not checking wrong type error causing key overwriting

enjoy-binbin:
This bug causes keys to be overwritten:
```
127.0.0.1:6666> lpush foo bar
(integer) 1
127.0.0.1:6666> type foo
list
127.0.0.1:6666> getex foo
""
127.0.0.1:6666> type foo
string
```

The reason is that we did not return early for wrong type
error and then the code overwrites it as a string key.
We should throw a wrong type error in this case:
```
127.0.0.1:6666> lpush foo bar
(integer) 1
127.0.0.1:6666> getex foo
(error) ERR Invalid argument: WRONGTYPE Operation against a key holding the wrong kind of value
127.0.0.1:6666> type foo
list
```

GETEX was added in #961, look like this bug has existed since
the command was added.

### DIFF
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -163,7 +163,7 @@ rocksdb::Status String::GetEx(const std::string &user_key, std::string *value, u
 
   LockGuard guard(storage_->GetLockManager(), ns_key);
   rocksdb::Status s = getValue(ns_key, value);
-  if (!s.ok() && s.IsNotFound()) return s;
+  if (!s.ok()) return s;
 
   std::string raw_data;
   Metadata metadata(kRedisString, false);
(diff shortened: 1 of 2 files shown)


tisonkun (approved):
Thank you!

git-hulk (approved):
LGTM, good catch

torwig (approved):
@enjoy-binbin Good catch!

xiaobiaozhao (approved):

enjoy-binbin:
Hi guys, i see we are doing squash merge but the commit message is empty.
Personally, I value the commit message. To be precise, i value the git blame log.
It is very helpful for us to track the PR changes, the change details etc.
My suggestion is to add appropriate message to express a certain commit when the code is merged.

In Redis community, we usually maintain the top comment of the PR. Both the proposer of the PR and the maintainer of the project have the right to modify the top comment (to reflect the latest status of PR). and when the time is come, maintainers can directly use top comment as commit message and do the squash merge. We can simply edit the top comment, copy those plain text and then do the squash merge

PS. it was discussed in https://github.com/apache/kvrocks/discussions/1553

PR #1549, opened 2023-07-05
Reduce the block cache configurations into a single one

### LINKED ISSUE #1495: [proposal] Reduce the block cache configurations into a single one
### Search before asking

### Motivation

Historically, we have three configurations to control the block cache behavior:

- [rocksdb.metadata_block_cache_size](https://github.com/apache/incubator-kvrocks/blob/unstable/kvrocks.conf#LL530C1-L530C34) which will determine the block cache size of the metadata column family
- [rocksdb.subkey_block_cache_size](https://github.com/apache/incubator-kvrocks/blob/unstable/kvrocks.conf#LL535C1-L535C32) which will determine the block cache size of other column families
- [rocksdb.share_metadata_and_subkey_block_cache](https://github.com/apache/incubator-kvrocks/blob/unstable/kvrocks.conf#L542) which is used to determine whether to share the same block cache between column families.

But most users don't understand why we need to configure two block cache sizes without investigating the implementation. And it also didn't make sense to separate them since the user data may be changed as time elapsed.

### Solution

We can reduce them into a single one like `rocksdb.block_cache_size` and deprecate old configurations, and also remove them from the kvrocks.conf.

1. If the `rocksdb.block_cache_size` was set, then use it as the block cache and ignore old configurations.

2. If the `rocksdb.block_cache_size` was NOT and `rocksdb.share_metadata_and_subkey_block_cache` was enabled, then calculate the `rocksdb.block_cache_size` from `rocksdb.metadata_block_cache_size` and `rocksdb.share_metadata_and_subkey_block_cache`

Yangsx-1:
Close #1495


kvrocks.conf
```diff
@@ -530,22 +530,36 @@ migrate-sequence-gap 10000
...
+# If this configuration is set to no, the following block cache settings
+# should be set. (deprecated)
+# Default: yes
+rocksdb.set_block_cache_size yes
+
 # Specify the capacity of metadata column family block cache. A larger block cache
 # may make requests faster while more keys would be cached. Max Size is 200*1024.
 # Default: 2048MB
-rocksdb.metadata_block_cache_size 2048
+#rocksdb.metadata_block_cache_size 2048 (deprecated)
```
PragmaTwice: It is not good to write `(deprecated)` here. Since it is not a vaild config item then.

src/config/config.cc
```diff
@@ -186,10 +186,12 @@ Config::Config() {
       {"rocksdb.enable_pipelined_write", true, new YesNoField(&rocks_db.enable_pipelined_write, false)},
       {"rocksdb.stats_dump_period_sec", false, new IntField(&rocks_db.stats_dump_period_sec, 0, 0, INT_MAX)},
       {"rocksdb.cache_index_and_filter_blocks", true, new YesNoField(&rocks_db.cache_index_and_filter_blocks, false)},
+      {"rocksdb.block_cache_size", true, new IntField(&rocks_db.block_cache_size, 4096, 0, INT_MAX)},
+      {"rocksdb.set_block_cache_size", true, new YesNoField(&rocks_db.set_block_cache_size, true)},
```
git-hulk: We can remove this configuration and set the `block_cache_size` default value to 0 in the previous line, then can check if `block_cache_size` was enabled or not.

If not set, calculate the value by `rocks_db.subkey_block_cache_size+rocks_db.metadata_block_cache_size`.

src/storage/storage.cc
```diff
@@ -235,6 +235,7 @@ Status Storage::Open(bool read_only) {
   db_closing_ = false;
 
   bool cache_index_and_filter_blocks = config_->rocks_db.cache_index_and_filter_blocks;
+  size_t block_cache_size = config_->rocks_db.block_cache_size * MiB;
```
git-hulk, line 238: Correct the block_cache_size value to metadata_block_cache_size + subkey_block_cache_size if it's zero.

src/storage/storage.cc
```diff
@@ -244,7 +245,9 @@ Status Storage::Open(bool read_only) {
   }
 
   std::shared_ptr<rocksdb::Cache> shared_block_cache;
-  if (config_->rocks_db.share_metadata_and_subkey_block_cache) {
+  if (config_->rocks_db.set_block_cache_size) {
+    shared_block_cache = rocksdb::NewLRUCache(block_cache_size, -1, false, 0.75);
+  } else if (config_->rocks_db.share_metadata_and_subkey_block_cache) {
     size_t shared_block_cache_size = metadata_block_cache_size + subkey_block_cache_size;
     shared_block_cache = rocksdb::NewLRUCache(shared_block_cache_size, -1, false, 0.75);
   }
```
git-hulk: always use the block_cache_size to create the lru cache

git-hulk:
@Yangsx-1 Another point is that needs to ignore the below configurations when rewriting:

- rocksdb.share_metadata_and_subkey_block_cache
- rocksdb.metadata_block_cache_size
- rocksdb.subkey_block_cache_size

to make sure they won't appear in the config file.

kvrocks.conf
```diff
@@ -530,22 +530,36 @@ migrate-sequence-gap 10000
...
+# If this configuration is set to no, the following block cache settings
+# should be set.
+# Default: yes
+rocksdb.set_block_cache_size yes
+
 # Specify the capacity of metadata column family block cache. A larger block cache
 # may make requests faster while more keys would be cached. Max Size is 200*1024.
 # Default: 2048MB
-rocksdb.metadata_block_cache_size 2048
+#rocksdb.metadata_block_cache_size 2048
```
git-hulk: Remove those configurations from the kvrocks.conf:

rocksdb.share_metadata_and_subkey_block_cache
rocksdb.metadata_block_cache_size
rocksdb.subkey_block_cache_size

Yangsx-1: Oh! I thought these parameters can be kept to be compatible with previous versions, maybe it's not necessary. :)

git-hulk: @Yangsx-1 We just remove them from the configuration file and it won't affect the compatibility instead of removing them from the config.cc.

Yangsx-1:
> 

This means that these configurations shouldn't be in the config file? If it appears, there may be some warning messages or error?

Yangsx-1:
> @Yangsx-1 Another point is that needs to ignore the below configurations when rewriting:
> 
> * rocksdb.share_metadata_and_subkey_block_cache
> * rocksdb.metadata_block_cache_size
> * rocksdb.subkey_block_cache_size
> 
> to make sure they won't appear in the config file.

This means that these configurations shouldn't be in the config file? If it appears, there may be some warning messages or error?

git-hulk:
> > @Yangsx-1 Another point is that needs to ignore the below configurations when rewriting:
> > 
> > * rocksdb.share_metadata_and_subkey_block_cache
> > * rocksdb.metadata_block_cache_size
> > * rocksdb.subkey_block_cache_size
> > 
> > to make sure they won't appear in the config file.
> 
> This means that these configurations shouldn't be in the config file? If it appears, there may be some warning messages or error?

What I mean is to remove them from the `kvrocks.conf` and keep the `config.cc` as it is, since the rocksdb.block_cache_size is expected to calculate from them if it's not set.

Yangsx-1:
> > > @Yangsx-1 Another point is that needs to ignore the below configurations when rewriting:
> > > 
> > > * rocksdb.share_metadata_and_subkey_block_cache
> > > * rocksdb.metadata_block_cache_size
> > > * rocksdb.subkey_block_cache_size
> > > 
> > > to make sure they won't appear in the config file.
> > 
> > 
> > This means that these configurations shouldn't be in the config file? If it appears, there may be some warning messages or error?
> 
> What I mean is to remove them from the `kvrocks.conf` and keep the config.cc as it is, since the rocksdb.block_cache_size is expected to calculate from them if it's not set.

OK

src/storage/storage.cc
```diff
@@ -245,8 +249,7 @@ Status Storage::Open(bool read_only) {
 
   std::shared_ptr<rocksdb::Cache> shared_block_cache;
   if (config_->rocks_db.share_metadata_and_subkey_block_cache) {
```
git-hulk: @Yangsx-1 We can remove the condition check here since it makes sense to separate the cache by the column family.

Yangsx-1: I think it's necessary. If the user manually set `rocks_db.share_metadata_and_subkey_block_cache` to `no`, the metatable cache size should be `metadata_block_cache_size`, which is assigned later. The same as subkey cache.

git-hulk: Yes, but it makes no sense for users to separate the column family block cache, so it's good to ignore the share block cache condition. I prefer to remove those old configurations and only keep them in config.cc for compatible reasons.

Yangsx-1: OK.  :D

src/storage/storage.cc
```diff
@@ -235,19 +235,20 @@ Status Storage::Open(bool read_only) {
...
 
   std::shared_ptr<rocksdb::Cache> shared_block_cache;
-  if (config_->rocks_db.share_metadata_and_subkey_block_cache) {
-    size_t shared_block_cache_size = metadata_block_cache_size + subkey_block_cache_size;
-    shared_block_cache = rocksdb::NewLRUCache(shared_block_cache_size, -1, false, 0.75);
-  }
+  shared_block_cache = rocksdb::NewLRUCache(block_cache_size, -1, false, 0.75);
 
   rocksdb::BlockBasedTableOptions metadata_table_opts = InitTableOptions();
   metadata_table_opts.block_cache =
```
git-hulk: Can use shared_block_cache directly, also other column family table's block cache options can remove the NewLRU condition.

git-hulk:
Others are generally good for me, thanks for your efforts.

git-hulk (approved):

PragmaTwice:
@git-hulk `rocksdb.share_metadata_and_subkey_block_cache` is completely ignored after the change (it will always be ON, despite whether the option is ON or OFF). Is it OK?

git-hulk:
> @git-hulk `rocksdb.share_metadata_and_subkey_block_cache` is completely ignored after the change (it will always be ON, despite whether the option is ON or OFF). Is it OK?

Yes, it's intentional. And it's a historically wrong decision to separate the block cache between column families. From the user side, it's really hard to balance the cache capacity between column families, the best way is to let LRU determine. To avoid the change being too aggressive, @ShooterIT added this option to allow users to use the old way. But from the performance, sharing the block cache is always better than the separate, so I think it's time to ignore this configuration and remove it in next major version.

PragmaTwice:
Thanks for your contribution!

PR #1550, opened 2023-07-05
[feat] diable no-tls ports

### LINKED ISSUE #1545: Disable "plain" server when using (m)TLS
### Search before asking

### Motivation

Usually, when we want to enable (m)TLS connections on Redis, we first disable the `port` parameter by setting it to 0 in order to have only one "endpoint", the TLS one:

> The `tls-port` configuration directive enables accepting SSL/TLS connections on the specified port. This is in addition to listening on port for TCP connections, so it is possible to access Redis on different ports using TLS and non-TLS connections simultaneously.
> You may specify `port 0` to disable the non-TLS port completely.

Source : https://redis.io/docs/management/security/encryption/

On Kvrocks though, it seems impossible to do the same. When we enable TLS through the various settings such as `tls-port`,    using version 2.4.0, it is not possible to disable the non TLS endpoint. When setting `port` to 0, we get an error "out of numeric range".  When we set `port` to the same value as `tls-port`, we can only reach the non-TLS version etc.

### Solution

It would be nice if we could disable the non-TLS endpoint when enabling TLS.

As I can see in the code, right now the `port` must be set between 1 and PORT_MAX - 1.  We could check:

* If `tls-port` is enabled ;
* And `port` is set to 0 ;

Then disable the non-TLS endpoint.

infdahai:
close #1545

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -664,8 +664,16 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
   if (iter != fields_.end()) {
     auto &field = iter->second;
     field->line_number = line_number;
+    if (field_key == "tls-port" && !close_notls_port && input.second == fillds_["port"]) {
+      close_notls_port = true;
+    }
     auto s = field->Set(input.second);
-    if (!s.IsOK()) return s.Prefixed(fmt::format("failed to set value of field '{}'", field_key));
+    if (!s.IsOK()) {
+      if (field_key == "port" && input.second == "0") {
+        close_notls_port = true;
+      } else
+        return s.Prefixed(fmt::format("failed to set value of field '{}'", field_key));
+    }
   } else {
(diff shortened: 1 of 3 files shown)


src/config/config.cc
```diff
@@ -664,8 +664,16 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
...
+      close_notls_port = true;
+    }
     auto s = field->Set(input.second);
-    if (!s.IsOK()) return s.Prefixed(fmt::format("failed to set value of field '{}'", field_key));
+    if (!s.IsOK()) {
+      if (field_key == "port" && input.second == "0") {
+        close_notls_port = true;
+      } else
+        return s.Prefixed(fmt::format("failed to set value of field '{}'", field_key));
+    }
```
PragmaTwice, line 676: Please do not just put these special code in a general function, which is just like a kind of "pollution".

Please use the config field mechanism and callback functions.

git-hulk:
Close this PR due to lack of activity.

PR #1556, opened 2023-07-06
Return an error when Stream::Range fails and remove some debug server logs

enjoy-binbin:
This PR has two parts.

The first one, internally, both XREAD and XRANGE call the Stream::Range
function. When Range fails, we should return an error.

The second one is to remove the debug logs. These logs were there in the
earliest days, i guess they are for debug, remove them now to avoid polluting
the server logs.

The blpop one is easy to reproduce:
```
Failed to execute redis command: blpop, err: Invalid argument: WRONGTYPE
Operation against a key holding the wrong kind of value
```

Checked the relevant LOG calls and fix a typo.

### DIFF
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -742,8 +742,6 @@ class CommandXRead : public Commander,
       auto s = stream_db.Range(streams_[i], options, &result);
       if (!s.ok()) {
         conn_->Reply(redis::MultiLen(-1));
-        LOG(ERROR) << "ERR executing XRANGE for stream " << streams_[i] << " from " << ids_[i].ToString() << " to "
-                   << options.end.ToString() << " with count " << count_ << ": " << s.ToString();
       }
 
       if (result.size() > 0) {
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -740,8 +740,8 @@ class CommandXRead : public Commander,
 
       std::vector<StreamEntry> result;
       auto s = stream_db.Range(streams_[i], options, &result);
-      if (!s.ok()) {
-        conn_->Reply(redis::MultiLen(-1));
+      if (!s.ok() && !s.IsNotFound()) {
+        return {Status::RedisExecErr, s.ToString()};
       }
 
       if (result.size() > 0) {
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -741,7 +741,8 @@ class CommandXRead : public Commander,
       std::vector<StreamEntry> result;
       auto s = stream_db.Range(streams_[i], options, &result);
       if (!s.ok() && !s.IsNotFound()) {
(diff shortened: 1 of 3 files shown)


src/commands/cmd_stream.cc
```diff
@@ -742,8 +742,6 @@ class CommandXRead : public Commander,
       auto s = stream_db.Range(streams_[i], options, &result);
       if (!s.ok()) {
         conn_->Reply(redis::MultiLen(-1));
-        LOG(ERROR) << "ERR executing XRANGE for stream " << streams_[i] << " from " << ids_[i].ToString() << " to "
```
enjoy-binbin, line 745: i am not sure how to trigger it yet, but look at the code, look like it works in XREAD instead of XRANGE.
i belive it is also a debug log?

git-hulk: @torwig Could you take a look at this line?

PragmaTwice:
@xiaobiaozhao Why force push cf80ff7d443e42c08f73c1f89caf8996965132bf?

xiaobiaozhao:
> @xiaobiaozhao Why force push [cf80ff7](https://github.com/apache/kvrocks/commit/cf80ff7d443e42c08f73c1f89caf8996965132bf)?

rebase ？

PragmaTwice:
> > @xiaobiaozhao Why force push [cf80ff7](https://github.com/apache/kvrocks/commit/cf80ff7d443e42c08f73c1f89caf8996965132bf)?
> 
> rebase ？

It is very weird. DO NOT force push to other contributor's branch without notification.

If you want to merge unstable to it, just click the "Update Branch" button, it will append a new commit.

PragmaTwice:
@xiaobiaozhao And I would like to state that I believe this is a very serious violation of the "unspoken agreement" among committers in git operations.

git-hulk:
Yes, you can append a new commit and SHOULD NOT rebase or force push to other contributors' PR.

xiaobiaozhao:
> Yes, you can append a new commit and SHOULD NOT rebase or force push to other contributors' PR.

OK, got it

src/commands/cmd_stream.cc
torwig: @enjoy-binbin Internally, both `XREAD` and `XRANGE` call the `Stream::Range` function. 
I think we can respond with an error here (as we do in line 634) in case of an error:

```
if (!s.ok() && !s.IsNotFound()) {
} 
```

enjoy-binbin: thanks. I will update it soon

enjoy-binbin:
i doing a force-push since it is a trivial diff.
For me, reset --hard unstable + cherry-pick + a new commit is clearer (and easier)

git-hulk (dismissed):

torwig (dismissed):
LGTM

git-hulk (approved):

git-hulk:
Thanks all, merging...

PR #1558, opened 2023-07-06
Avoid printing the invalid configuration key when loading namespaces

git-hulk:
For the namespace/token pair, it shouldn't go through the configuration key check process, but we forgot to return when it's the namespace key. And it will print an invalid configuration key warning message after #1498 which may confuse users.


PragmaTwice (approved):
LGTM. BTW it also need to cherry-pick to release 2.5.0.

PR #1559, opened 2023-07-06
Fix ZUNION hang in syntax error

enjoy-binbin:
Bug was introduced in #1502, without this return, when syntax
error happen, the code will stuck in the while loop and
result a hang.


git-hulk (approved):
LGTM, good catch! 👍

enjoy-binbin:
this could be a security issue if there is a release that includes this command
i see 2.5 RC include it, please make sure patch it

git-hulk:
> this could be a security issue if there is a release that includes this command i see 2.5 RC include it, please make sure patch it

Yes, 2.5.0 is voting now, but I'm not sure if we should cancel the vote. This bug would cause the server to be trapped in the dead loop. cc @tisonkun @torwig @torwig @ShooterIT @ColinChamber

PragmaTwice (approved):

git-hulk:
Thanks all, merging...

Issue #1510, opened 2023-06-22
Add the support of the HRANDFIELD comand


infdahai:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

This issue considers to support the [hrandfield](https://redis.io/commands/hrandfield/) command.

After this, we will support all the commands in the hash type from redis.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

infdahai:
Please add a good-first-issue label.

zevin02:
please assign to me

torwig:
@zevin02 Done. Thank you!

PR #1565, opened 2023-07-08
Add the support of the HRANDFIELD command

zevin02:
Close :#1510


src/commands/cmd_hash.cc
```diff
@@ -376,6 +376,52 @@ class CommandHScan : public CommandSubkeyScanBase {
...
+      if (!parse_result) {
+        return {Status::RedisParseErr, errValueNotInteger};
+      }
+      if (*parse_result >= 0) {
+        count_ = *parse_result;
+      } else {
+        count_ = -*parse_result;
+        uniq_ = false;
+      }
+      //如果有withvalue这个参数，就一定有前面的count
```
PragmaTwice: Please comment in English :)

src/commands/cmd_hash.cc
```diff
@@ -376,6 +376,52 @@ class CommandHScan : public CommandSubkeyScanBase {
...
+      }
+    }
+    return Commander::Parse(args);
+  }
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::Hash hash_db(svr->storage, conn->GetNamespace());
+    std::vector<FieldValue> field_values;
+    auto s = hash_db.RandField(args_[1], &field_values, count_, uniq_,
+                               withvalues_ ? HashFetchType::kAll : HashFetchType::kOnlyKey);
+    if (!s.ok()) {
```
Yangsx-1: It should return `nil` if the key is not exist.

src/types/redis_hash.cc
```diff
@@ -386,5 +386,46 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+        field_values->emplace_back(ikey.GetSubKey().ToString(), "");
+      } else {
+        InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+        field_values->emplace_back(ikey.GetSubKey().ToString(), iter->value().ToString());
+      }
+    }
+  }
+  /* case3:The number of elements inside the hash is not greater than the number of
+    elements inside the hash*/
+  else {
```
mapleFU: is this still wip?

zevin02: I have just completed one case, and I am still working on implementing the other two cases.

src/types/redis_hash.cc
```diff
@@ -386,5 +387,78 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+    if (type == HashFetchType::kAll) {
+      values.emplace_back(iter->value().ToString());
+    }
+  }
+
+  // Case 1: Negative count, randomly select elements
+  if (!uniq) {
+    field_values->reserve(count);
+    for (uint64_t i = 0; i < count; i++) {
+      uint64_t index = (std::rand() % size);
```
Yangsx-1: It's better to add a rand seed before `rand()` and use `std::mt19937` the same as below.

src/commands/cmd_hash.cc
```diff
@@ -376,6 +376,52 @@ class CommandHScan : public CommandSubkeyScanBase {
...
+      if (withvalues_) result_entries.emplace_back(p.value);
+    }
+    *output = s.IsNotFound() ? redis::NilString() : redis::MultiBulkString(result_entries, false);
+    return Status::OK();
+  }
+
+ private:
+  bool withvalues_ = false;
+  bool uniq_ = true;
+  uint64_t count_ = 1;
```
Yangsx-1: Maybe `int64_t` is better.

src/types/redis_hash.cc
```diff
@@ -25,11 +25,12 @@
 #include <algorithm>
 #include <cctype>
 #include <cmath>
+#include <random>
 #include <utility>
 
 #include "db_util.h"
 #include "parse_util.h"
-
+#include "time_util.h"
```
Yangsx-1: Leave a blank line here, the same as the end of namespace and between functions.   :)

src/types/redis_hash.cc
```diff
@@ -386,5 +387,78 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+  std::vector<std::string> fields;
+  std::vector<std::string> values;
+  auto iter = util::UniqueIterator(storage_, read_options);
+  for (iter->Seek(prefix_key); iter->Valid() && iter->key().starts_with(prefix_key); iter->Next()) {
+    InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+    fields.emplace_back(ikey.GetSubKey().ToString());
+    if (type == HashFetchType::kAll) {
+      values.emplace_back(iter->value().ToString());
+    }
+  }
```
Yangsx-1: You can use `GetAll` function, it's almost the same.

Yangsx-1:
You can also add some go tests.

src/commands/cmd_hash.cc
```diff
@@ -376,6 +376,52 @@ class CommandHScan : public CommandSubkeyScanBase {
...
+    }
+    return Commander::Parse(args);
+  }
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::Hash hash_db(svr->storage, conn->GetNamespace());
+    std::vector<FieldValue> field_values;
+
+    auto s = hash_db.RandField(args_[1], &field_values, count_, uniq_,
+                               withvalues_ ? HashFetchType::kAll : HashFetchType::kOnlyKey);
+    if (!s.ok() && !s.IsNotFound()) {
```
Yangsx-1, line 406: Sorry, in redis, if without the additional count argument, the command returns a Bulk Reply with the randomly selected field, or nil when key does not exist. When the additional count argument is passed, the command returns an array of fields, or an empty array when key does not exist. If the WITHVALUES modifier is used, the reply is a list fields and their values from the hash. See details in https://redis.io/commands/hrandfield/

src/types/redis_hash.cc
```diff
@@ -386,5 +387,78 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+  InternalKey(ns_key, "", metadata.version + 1, storage_->IsSlotIdEncoded()).Encode(&next_version_prefix_key);
 
+  rocksdb::ReadOptions read_options;
+  LatestSnapShot ss(storage_);
+  read_options.snapshot = ss.GetSnapShot();
+  rocksdb::Slice upper_bound(next_version_prefix_key);
+  read_options.iterate_upper_bound = &upper_bound;
+  storage_->SetReadOptions(read_options);
+
+  uint64_t size = metadata.size;
```
mapleFU, line 399: Getting all values in Hash might be heavy, could we lazy load these values later, or just leave it here and add a TODO for further optimizations?

src/types/redis_hash.cc
```diff
@@ -386,5 +387,78 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+  std::vector<std::string> values;
+  auto iter = util::UniqueIterator(storage_, read_options);
+  for (iter->Seek(prefix_key); iter->Valid() && iter->key().starts_with(prefix_key); iter->Next()) {
+    InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+    fields.emplace_back(ikey.GetSubKey().ToString());
+    if (type == HashFetchType::kAll) {
+      values.emplace_back(iter->value().ToString());
+    }
+  }
+
```
mapleFU: Should we optimize the case when `count == 0`, when `count == 0`, seems that no values need to be loaded from storage

src/types/redis_hash.cc
mapleFU: Seems that we have a `src/common/rand.h`, could we use that @PragmaTwice

src/types/redis_hash.cc
```diff
@@ -386,5 +387,78 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+        field_values->emplace_back(fields[index], "");
+      }
+    }
+  }
+
+  // Case 2: Requested count is greater than or equal to the number of elements inside the hash
+  else if (size <= count) {
+    field_values->reserve(size);
+    for (uint64_t i = 0; i < size; i++) {
+      if (type == HashFetchType::kAll) {
```
mapleFU:
```c++
      if (type == HashFetchType::kAll) {
        field_values->emplace_back(fields[i], values[i]);
      } else {
        field_values->emplace_back(fields[i], "");
      }
```

This part occurs three times here, would you mind encapsulate it to a lambda?

mapleFU (commented):
I didn't review the `Hash::RandField` implements, just some style nits

src/types/redis_hash.cc
Yangsx-1: Yes, it's a good optimization.  :)

src/types/redis_hash.cc
zevin02: I can add a TODO for further optimizations

src/types/redis_hash.cc
```diff
@@ -386,5 +387,78 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+        field_values->emplace_back(fields[index], values[index]);
+      } else {
+        field_values->emplace_back(fields[index], "");
+      }
+    }
+  }
+
+  // Case 2: Requested count is greater than or equal to the number of elements inside the hash
+  else if (size <= count) {
+    field_values->reserve(size);
```
PragmaTwice: no need to shuffle here? what's redis corresponding behavior?

zevin02: In the redis implementation, when the requested data is larger than the data stored in the hash, the data in the hash is returned in order

src/types/redis_hash.cc
```diff
@@ -386,5 +387,78 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+    std::shuffle(indices.begin(), indices.end(), gen);  // use Fisher-Yates shuffle algorithm to randomize the order
+    for (uint64_t i = 0; i < count; i++) {
+      uint64_t index = indices[i];
+      if (type == HashFetchType::kAll) {
+        field_values->emplace_back(fields[index], values[index]);
+      } else {
+        field_values->emplace_back(fields[index], "");
+      }
+    }
+  }
```
PragmaTwice: seems the output will always be a random order of index {0,...`count`-1}, which cannot be called "rand field" IMHO.

PragmaTwice: if you want to do the shuffle, you need to shuffle [0, `hash_elem_size`) and cut the first `count` numbers as indices.

zevin02: > if you want to do the shuffle, you need to shuffle [0, `hash_elem_size`) and cut the first `count` numbers as indices.

should I modify it like this, or what should I do
~~~c
std::shuffle(indices.begin(), indices.begin() + count, gen);  
std::sort(indices.begin(), indices.begin() + count);
~~~

zevin02: > if you want to do the shuffle, you need to shuffle [0, `hash_elem_size`) and cut the first `count` numbers as indices.
should I modify like this,or anything else
~~~cpp
indices.resize(count);
~~~

src/types/redis_hash.cc
PragmaTwice: oh got it

src/types/redis_hash.cc
zevin02: > if you want to do the shuffle, you need to shuffle [0, `hash_elem_size`) and cut the first `count` numbers as indices.

should I modify like this,or anything else
~~~cpp
indices.resize(count);
~~~

PragmaTwice: oh I may mistake `std::vector indices(size)` as `indices(count)`, sorry for that. the implementation looks fine to me.

PragmaTwice: I think you can just use random_device as the URBG without mt19937. e.g 

```
shuffle(begin, end, std::random_device{})
```

Yangsx-1: A little reference for [why not just use random_device](https://stackoverflow.com/questions/39288595/why-not-just-use-random-device).  :)

mapleFU: usally random device can be used to generate a mt19937 seed

zevin02:
> You can also add some go tests.

I will do it later

src/types/redis_hash.cc
PragmaTwice: > A little reference for [why not just use random_device](https://stackoverflow.com/questions/39288595/why-not-just-use-random-device). :)

@maplefu @Yangsx-1  I am not sure whether the common way to implement std::random_device is still `/dev/urandom`, the answer seems to be out-dated.

Usually, e.g. in x86, the `rdrand` instruction is a far better way to implement std::random_device with great performance.

PragmaTwice: > usally random device can be used to generate a mt19937 seed

mt19937 is just a good PRNG algorithm. it is not required to use if your device can generate **real random numbers**.

PragmaTwice: @Yangsx-1 I have checked the latest version of libstdc++, which are using `rdrand` when possible.

https://github.com/gcc-mirror/gcc/blob/63ae6bc60c0f67fb2791991bf4b6e7e0a907d420/libstdc%2B%2B-v3/src/c%2B%2B11/random.cc#L34

src/types/redis_hash.cc
git-hulk: Why not use `GetAll` to get all fields for field_values directly?

src/commands/cmd_hash.cc
```diff
@@ -376,6 +376,54 @@ class CommandHScan : public CommandSubkeyScanBase {
...
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() >= 3) {
+      noParameters_ = false;
+      auto parse_result = ParseInt<int64_t>(args[2], 10);
+      if (!parse_result) {
+        return {Status::RedisParseErr, errValueNotInteger};
+      }
+      l_ = *parse_result;
+
+      if (args.size() > 4 || (args.size() == 4 && strcasecmp(args[3].c_str(), "withvalues"))) {
```
mapleFU:
```suggestion
      if (args.size() > 4 || (args.size() == 4 && !util::EqualICase(args[3], "withvalues"))) {
```

src/commands/cmd_hash.cc
```diff
@@ -376,6 +376,54 @@ class CommandHScan : public CommandSubkeyScanBase {
...
+    if (noParameters_)
+      *output = s.IsNotFound() ? redis::NilString() : redis::BulkString(result_entries[0]);
+    else
+      *output = redis::MultiBulkString(result_entries, false);
+    return Status::OK();
+  }
+
+ private:
+  bool withvalues_ = false;
+  int64_t l_ = 0;
```
mapleFU: `l_` is a confusing name, I guess `count_` could be much better

src/types/redis_hash.cc
```diff
@@ -387,4 +388,60 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
 }
 
+rocksdb::Status Hash::RandField(const Slice &user_key, std::vector<FieldValue> *field_values, int64_t l,
+                                bool noparmeter, HashFetchType type) {
+  uint64_t count = (l >= 0) ? static_cast<uint64_t>(l) : static_cast<uint64_t>(-l);
+  bool unique = (l >= 0);
+
+  std::string ns_key;
+  AppendNamespacePrefix(user_key, &ns_key);
+  HashMetadata metadata(false);
```
mapleFU:
```suggestion
  HashMetadata metadata(/*generate_version=*/false);
```

src/types/redis_hash.cc
```diff
@@ -387,4 +388,60 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+  rocksdb::Status s = GetMetadata(ns_key, &metadata);
+  if (!s.ok()) return s;
+
+  uint64_t size = metadata.size;
+  std::vector<FieldValue> samples;
+  // TODO: Getting all values in Hash might be heavy, consider lazy-loading these values later
+
```
mapleFU: Can just return when `l == 0`?

src/types/redis_hash.cc
```diff
@@ -387,4 +388,60 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+  GetAll(user_key, &samples, type);
+
+  auto processField = [field_values, &samples, type](uint64_t index) {
```
mapleFU: would `appendFieldWithIndex` or others be better?

src/types/redis_hash.cc
```diff
@@ -387,4 +388,60 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+    if (type == HashFetchType::kAll) {
+      field_values->emplace_back(samples[index].field, samples[index].value);
+    } else {
+      field_values->emplace_back(samples[index].field, "");
+    }
```
mapleFU, line 410: By the way, could we reserve the `field_values` as `min(size, count)` in this case after `GetAll` called?

src/types/redis_hash.cc
```diff
@@ -387,4 +388,60 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+  };
+
+  if (!unique) {
```
mapleFU: would it ok to use

```
if (!unique || count == 1)
```

like in redis? I believe it also handles the `noparameter` case

mapleFU (commented):
Looks much better now!

src/types/redis_hash.cc
mapleFU:
```
/* If our unsigned long type can store a 64 bit number, use a 64 bit PRNG. */
#if ULONG_MAX >= 0xffffffffffffffff
#define randomULong() ((unsigned long) genrand64_int64())
#else
#define randomULong() random()
#endif
```

@PragmaTwice after check in redis, I guess it would uses `randomULong()`, and it wraps a PRNG algorithm. I guess mt19937 would be ok

src/commands/cmd_hash.cc
zevin02: `count` is also used in the `RandField` , and `l` is also used in parameter extraction in the implementation of redis, so I also use `l`. Is there any better suggestion?

mapleFU: Hmm I guess Redis Command level uses name `COUNT`. Maybe using `command_count` or other is ok?

src/types/redis_hash.cc
```diff
@@ -387,4 +388,58 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+  std::string ns_key;
+  AppendNamespacePrefix(user_key, &ns_key);
+  HashMetadata metadata(/*generate_version=*/false);
+  rocksdb::Status s = GetMetadata(ns_key, &metadata);
+  if (!s.ok()) return s;
+
+  uint64_t size = metadata.size;
+  std::vector<FieldValue> samples;
+  // TODO: Getting all values in Hash might be heavy, consider lazy-loading these values later
+  if (count == 0 && !noparmeter) return rocksdb::Status::OK();
```
mapleFU: Would we remove `noparameter` in `Hash::RandField` by default set `command_count_ == 1`?

src/types/redis_hash.cc
```diff
@@ -387,4 +388,58 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+    }
+  } else if (size <= count) {
+    // Case 2: Requested count is greater than or equal to the number of elements inside the hash
+    for (uint64_t i = 0; i < size; i++) {
+      appendFieldWithIndex(i);
+    }
+  } else {
+    // Case 3: Requested count is less than the number of elements inside the hash
+    std::vector<uint64_t> indices(size);
+    std::iota(indices.begin(), indices.end(), 0);
```
mapleFU, line 430: Seems C++17 provides a `sample` ( https://en.cppreference.com/w/cpp/algorithm/sample ) , would it be better ? @PragmaTwice

src/types/redis_hash.h
```diff
@@ -61,6 +61,8 @@ class Hash : public SubKeyScanner {
   rocksdb::Status Scan(const Slice &user_key, const std::string &cursor, uint64_t limit,
                        const std::string &field_prefix, std::vector<std::string> *fields,
                        std::vector<std::string> *values = nullptr);
+  rocksdb::Status RandField(const Slice &user_key, std::vector<FieldValue> *field_values, int64_t l, bool noparmeter,
```
mapleFU:
```suggestion
  rocksdb::Status RandField(const Slice &user_key, int64_t command_count, bool noparmeter, std::vector<FieldValue> *field_values, 
```

Seems that input field should better before output fields

src/types/redis_hash.cc
```diff
@@ -387,4 +388,58 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+  AppendNamespacePrefix(user_key, &ns_key);
+  HashMetadata metadata(/*generate_version=*/false);
+  rocksdb::Status s = GetMetadata(ns_key, &metadata);
+  if (!s.ok()) return s;
+
+  uint64_t size = metadata.size;
+  std::vector<FieldValue> samples;
+  // TODO: Getting all values in Hash might be heavy, consider lazy-loading these values later
+  if (count == 0 && !noparmeter) return rocksdb::Status::OK();
+  GetAll(user_key, &samples, type);
```
mapleFU: `GetAll` returns a `Status`, seems that we'd better check it.

mapleFU (commented):
Looks much better! I didn't check the tests carefully

PragmaTwice:
Seems here are some clang-tidy reports that need to be fixed.

https://github.com/apache/kvrocks/actions/runs/5542464594/jobs/10117205769?pr=1565

src/commands/cmd_hash.cc
```diff
@@ -376,6 +376,54 @@ class CommandHScan : public CommandSubkeyScanBase {
...
+    auto s = hash_db.RandField(args_[1], command_count_, &field_values,
+                               withvalues_ ? HashFetchType::kAll : HashFetchType::kOnlyKey);
+    if (!s.ok() && !s.IsNotFound()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    std::vector<std::string> result_entries;
+    result_entries.reserve(field_values.size());
+    for (const auto &p : field_values) {
+      result_entries.emplace_back(p.field);
```
mapleFU, line 413: would `std::move(p.field)` be better here?

mapleFU (commented):
Rest LGTM!

By the way, I noticed that you have written:

```
// TODO: Add test to verify randomness of the selected random fields
```

Would this be added in this patch or later?

zevin02:
> 

This patch will not implement that

zevin02:
> Rest LGTM!
> 
> By the way, I noticed that you have written:
> 
> ```
> // TODO: Add test to verify randomness of the selected random fields
> ```
> 
> Would this be added in this patch or later?

This patch will not implement that

tests/gocase/unit/type/hash/hash_test.go
```diff
@@ -783,6 +783,53 @@ func TestHash(t *testing.T) {
...
+			testKey := "test-hash-1"
+			require.NoError(t, rdb.Del(ctx, testKey).Err())
+			require.NoError(t, rdb.HSet(ctx, testKey, "key1", "value1", "key2", "value2", "key3", "value3").Err())
+			result1, err1 := rdb.HRandField(ctx, testKey, 5).Result()
+			require.NoError(t, err1)
+			require.Len(t, result1, 3)
+			require.Equal(t, []string{"key1", "key2", "key3"}, result1)
+			result3, err3 := rdb.HRandField(ctx, testKey, 2).Result()
+			require.NoError(t, err3)
+			require.Len(t, result3, 2)
```
mapleFU: would you mind check the value is in `{"key1", "value1", "key2", "value2", "key3", "value3"}`?

tests/gocase/unit/type/hash/hash_test.go
```diff
@@ -783,6 +783,53 @@ func TestHash(t *testing.T) {
 
 			require.Equal(t, []interface{}{"field1", "some-value", "field2", ""}, rdb.Do(ctx, "HrangeByLex", testKey, "[a", "[z").Val())
 		})
+
+		t.Run("HRandField count is positive", func(t *testing.T) {
+			testKey := "test-hash-1"
+			require.NoError(t, rdb.Del(ctx, testKey).Err())
+			require.NoError(t, rdb.HSet(ctx, testKey, "key1", "value1", "key2", "value2", "key3", "value3").Err())
+			result1, err1 := rdb.HRandField(ctx, testKey, 5).Result()
```
mapleFU: can you use same err and result to avoid too many variable names?

zevin02: Ok, is there anything else that needs to be improved?

mapleFU: I guess no, if these are finished I'll approve this pull request. Thanks for your working

mapleFU (approved):
cc @PragmaTwice @Yangsx-1 @git-hulk

tests/cppunit/types/hash_test.cc
```diff
@@ -331,3 +331,34 @@ TEST_F(RedisHashTest, HRangeByLexNonExistingKey) {
...
+  s = hash_->RandField(key_, size - 1, &fvs);
+  EXPECT_TRUE(s.ok() && fvs.size() == fields_.size() - 1);
+
+  // hrandfield key 0
+  fvs.clear();
+  s = hash_->RandField(key_, 0, &fvs);
+  EXPECT_TRUE(s.ok() && fvs.size() == 0);
+
+  hash_->Del(key_);
+}
```
PragmaTwice:
```suggestion
}

```

src/commands/cmd_hash.cc
```diff
@@ -376,6 +376,54 @@ class CommandHScan : public CommandSubkeyScanBase {
     return Status::OK();
   }
 };
+class CommandHRandField : public Commander {
```
Yangsx-1, line 380:
```suggestion

class CommandHRandField : public Commander {
```

src/commands/cmd_hash.cc
```diff
@@ -376,6 +376,54 @@ class CommandHScan : public CommandSubkeyScanBase {
...
+      command_count_ = *parse_result;
+
+      if (args.size() > 4 || (args.size() == 4 && !util::EqualICase(args[3], "withvalues"))) {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      } else if (args.size() == 4) {
+        withvalues_ = true;
+      }
+    }
+    return Commander::Parse(args);
+  }
```
Yangsx-1, line 398:
```suggestion
  }
  
```

src/types/redis_hash.cc
```diff
@@ -387,4 +388,57 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+    for (uint64_t i = 0; i < size; i++) {
+      append_field_with_index(i);
+    }
+  } else {
+    // Case 3: Requested count is less than the number of elements inside the hash
+    std::vector<uint64_t> indices(size);
+    std::iota(indices.begin(), indices.end(), 0);
+    std::shuffle(indices.begin(), indices.end(),
+                 std::random_device{});  // use Fisher-Yates shuffle algorithm to randomize the order
+    indices.resize(count);
```
Yangsx-1: This resize may be not necessary.

src/types/redis_hash.cc
```diff
@@ -387,4 +388,57 @@ rocksdb::Status Hash::Scan(const Slice &user_key, const std::string &cursor, uin
...
+    for (uint64_t i = 0; i < count; i++) {
+      uint64_t index = indices[i];
+      append_field_with_index(index);
+    }
+  }
+  return rocksdb::Status::OK();
+}
```
Yangsx-1, line 439:
```suggestion
}

```

Yangsx-1 (commented):
Others are good to me.  :)

Yangsx-1:
> fixes :#1510
> 
> * Add parse for hrandfield
> * Implemented the case where the requested data is a positive number greater than the size of the hash.

Close #1510 is better.

Yangsx-1 (approved):

PragmaTwice (dismissed):

git-hulk (approved):

git-hulk:
@zevin02 Thanks for your contribution and patience and greater thanks to reviewers: @mapleFU @PragmaTwice @Yangsx-1

PR #1576, opened 2023-07-10
Fix config set compression type didn't take effect

### LINKED ISSUE #1560: Dynamic modification of the parameter rocksdb.compression to zstd in version 2.4 may not take effect immediately on newly created files
### Search before asking

### Version

2.4

### Minimal reproduce step

1.Modify the compression parameters.
2.FlushDB and compact
3.Rewrite the data, but it appears that zstd compression was not applied based on the logs. The output still shows that the compression used is lz4.
"[CompactOnExpiredCollector]", "compression": "LZ4", "compression_options": "window_bits=-14; level=32767; strategy=0; max_dict_bytes=0; zstd_max_train_bytes=0; enabled=0; max_dict_buffer_bytes=0; use_zstd_dict_trainer=1; "

if reboot kvrocks then use zstd

2023/07/05-13:22:43.170822 1        Options.write_buffer_size: 67108864
2023/07/05-13:22:43.170824 1  Options.max_write_buffer_number: 10
2023/07/05-13:22:43.170826 1        Options.compression[0]: NoCompression
2023/07/05-13:22:43.170827 1        Options.compression[1]: NoCompression
2023/07/05-13:22:43.170829 1        Options.compression[2]: ZSTD
2023/07/05-13:22:43.170830 1        Options.compression[3]: ZSTD
2023/07/05-13:22:43.170832 1        Options.compression[4]: ZSTD
2023/07/05-13:22:43.170833 1        Options.compression[5]: ZSTD
2023/07/05-13:22:43.170835 1        Options.compression[6]: ZSTD

"[CompactOnExpiredCollector]", "compression": "ZSTD", "compression_options": "window_bits=-14; level=32767; strategy=0; max_dict_bytes=0; zstd_max_train_bytes=0; enabled=0; max_dict_buffer_bytes=0; use_zstd_dict_trainer=1; "

Test passed

### What did you see instead?

The compression parameters only take effect after a restart

### Anything Else?

_No response_

xiaobiaozhao:
In this PR, we also refactor the c-style ConfigEnum array into the std::vector, the main reason is to make it easier to adopt dynamic config enums like the compression types. And for the test case, it's a bit hard to test if the compression type takes effect in underlying storage because we only enabled it when the levels >= 2. But will check it manually to make sure if works.

This closes https://github.com/apache/kvrocks/issues/1560

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -604,7 +604,7 @@ void Config::initFieldCallback() {
       {"rocksdb.compression",
        [](Server *srv, const std::string &k, const std::string &v) -> Status {
          if (!srv) return Status::OK();
-         return srv->storage->SetOptionForAllColumnFamilies(TrimRocksDbPrefix(k), v);
+         return srv->storage->SetCompressionOption(TrimRocksDbPrefix(k), v);
        }},
 #ifdef ENABLE_OPENSSL
       {"tls-cert-file", set_tls_option},
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -601,6 +601,11 @@ void Config::initFieldCallback() {
       {"rocksdb.level0_slowdown_writes_trigger", set_cf_option_cb},
       {"rocksdb.level0_stop_writes_trigger", set_cf_option_cb},
       {"rocksdb.level0_file_num_compaction_trigger", set_cf_option_cb},
+      {"rocksdb.compression",
+       [](Server *srv, const std::string &k, const std::string &v) -> Status {
+         if (!srv) return Status::OK();
+         return srv->storage->SetOptionForAllColumnFamilies(TrimRocksDbPrefix(k), v);
+       }},
 #ifdef ENABLE_OPENSSL
       {"tls-cert-file", set_tls_option},
       {"tls-key-file", set_tls_option},
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -45,40 +45,35 @@ constexpr const char *errBlobDbNotEnabled = "Must set rocksdb.enable_blob_files 
 constexpr const char *errLevelCompactionDynamicLevelBytesNotSet =
     "Must set rocksdb.level_compaction_dynamic_level_bytes yes first.";
 
-ConfigEnum compression_types[] = {
-    {"no", rocksdb::CompressionType::kNoCompression},     {"snappy", rocksdb::CompressionType::kSnappyCompression},
-    {"lz4", rocksdb::CompressionType::kLZ4Compression},   {"zstd", rocksdb::CompressionType::kZSTD},
-    {"zlib", rocksdb::CompressionType::kZlibCompression}, {nullptr, 0}};
+std::vector<ConfigEnum> supervised_modes{
+    {"no", kSupervisedNone},
+    {"auto", kSupervisedAutoDetect},
+    {"upstart", kSupervisedUpStart},
+    {"systemd", kSupervisedSystemd},
+};
 
-ConfigEnum supervised_modes[] = {{"no", kSupervisedNone},
-                                 {"auto", kSupervisedAutoDetect},
-                                 {"upstart", kSupervisedUpStart},
-                                 {"systemd", kSupervisedSystemd},
-                                 {nullptr, 0}};
-
-ConfigEnum log_levels[] = {{"info", google::INFO},
-                           {"warning", google::WARNING},
-                           {"error", google::ERROR},
-                           {"fatal", google::FATAL},
-                           {nullptr, 0}};
+std::vector<ConfigEnum> log_levels{
+    {"info", google::INFO},
+    {"warning", google::WARNING},
+    {"error", google::ERROR},
+    {"fatal", google::FATAL},
+};
 
 std::string TrimRocksDbPrefix(std::string s) {
   if (strncasecmp(s.data(), "rocksdb.", 8) != 0) return s;
   return s.substr(8, s.size() - 8);
 }
 
-int ConfigEnumGetValue(ConfigEnum *ce, const char *name) {
-  while (ce->name != nullptr) {
-    if (strcasecmp(ce->name, name) == 0) return ce->val;
-    ce++;
+int ConfigEnumGetValue(const std::vector<ConfigEnum> &enums, const char *name) {
+  for (const auto &e : enums) {
+    if (strcasecmp(e.name, name) == 0) return e.val;
   }
   return INT_MIN;
 }
 
-const char *ConfigEnumGetName(ConfigEnum *ce, int val) {
-  while (ce->name != nullptr) {
-    if (ce->val == val) return ce->name;
-    ce++;
+const char *ConfigEnumGetName(const std::vector<ConfigEnum> &enums, int val) {
+  for (const auto &e : enums) {
+    if (e.val == val) return e.name;
   }
   return nullptr;
 }
@@ -92,6 +87,11 @@ Config::Config() {
     FieldWrapper(std::string name, bool readonly, ConfigField *field)
         : name(std::move(name)), readonly(readonly), field(field) {}
   };
+
+  std::vector<ConfigEnum> compression_types;
+  for (const auto &e : *engine::GetCompressionOptions()) {
+    compression_types.push_back({e.name, e.type});
+  }
   FieldWrapper fields[] = {
       {"daemonize", true, new YesNoField(&daemonize, false)},
       {"bind", true, new StringField(&binds_str_, "")},
@@ -327,6 +327,30 @@ void Config::initFieldCallback() {
   auto set_cf_option_cb = [](Server *srv, const std::string &k, const std::string &v) -> Status {
     if (!srv) return Status::OK();  // srv is nullptr when load config from file
     return srv->storage->SetOptionForAllColumnFamilies(TrimRocksDbPrefix(k), v);
+  };
+  auto set_compression_type_cb = [](Server *srv, const std::string &k, const std::string &v) -> Status {
+    if (!srv) return Status::OK();
+
+    std::string compression_option = "";
+    for (auto &option : *engine::GetCompressionOptions()) {
+      if (option.name == v) {
+        compression_option = option.val;
+        break;
+      }
+    }
+    if (compression_option.empty()) {
+      return {Status::NotOK, "Invalid compression type"};
+    }
+
+    // For the first two levels, it may contain the frequently accessed data,
+    // so it'd be better to use uncompressed data to save the CPU.
+    std::string compression_levels = "kNoCompression:kNoCompression";
+    auto db = srv->storage->GetDB();
+    for (size_t i = 2; i < db->GetOptions().compression_per_level.size(); i++) {
+      compression_levels += ":";
+      compression_levels += compression_option;
+    }
+    return srv->storage->SetOptionForAllColumnFamilies("compression_per_level", compression_levels);
   };
 #ifdef ENABLE_OPENSSL
   auto set_tls_option = [](Server *srv, const std::string &k, const std::string &v) {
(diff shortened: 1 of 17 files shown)


src/storage/storage.cc
```diff
@@ -180,6 +180,29 @@ rocksdb::Options Storage::InitRocksDBOptions() {
 }
 
 Status Storage::SetOptionForAllColumnFamilies(const std::string &key, const std::string &value) {
+  if ("compression" == key) {
```
git-hulk: I think we can take this into a single function like `SetCompressionOption` since it's a special setting option.

src/storage/storage.cc
```diff
@@ -183,6 +183,29 @@ rocksdb::Options Storage::InitRocksDBOptions() {
...
+    std::unordered_map<std::string, std::string> compression_map = {{"no", "kNoCompression"},
```
git-hulk: It'd be better to put this text in compression_types, to make it more easy to maintain. Or it needs to take care of many places if we wanna add a new compression type.

xiaobiaozhao: > It'd be better to put this text in compression_types, to make it more easy to maintain. Or it needs to take care of many places if we wanna add a new compression type.

done

src/storage/storage.cc
xiaobiaozhao: done

src/config/config.cc
```diff
@@ -601,6 +601,11 @@ void Config::initFieldCallback() {
       {"rocksdb.level0_slowdown_writes_trigger", set_cf_option_cb},
       {"rocksdb.level0_stop_writes_trigger", set_cf_option_cb},
       {"rocksdb.level0_file_num_compaction_trigger", set_cf_option_cb},
+      {"rocksdb.compression",
+       [](Server *srv, const std::string &k, const std::string &v) -> Status {
+         if (!srv) return Status::OK();
+         return srv->storage->SetOptionForAllColumnFamilies(TrimRocksDbPrefix(k), v);
+       }},
```
PragmaTwice: Why can't we simply write this compression option manipulation code here instead of `Storage::SetOptionForAllColumnFamilies`?

src/config/config.h
```diff
@@ -39,6 +39,7 @@ namespace engine {
 class Storage;
 }
 
+extern ConfigEnum compression_types[];
```
PragmaTwice: I cannot find the reason to add an extern variable here.

src/storage/storage.cc
```diff
@@ -182,7 +182,61 @@ rocksdb::Options Storage::InitRocksDBOptions() {
...
+    auto s = db_->SetOptions(cf_handle, {{"compression_per_level", compression_levels}});
+    if (!s.ok()) return {Status::NotOK, s.ToString()};
+  }
+  return Status::OK();
+}
+
 Status Storage::SetOptionForAllColumnFamilies(const std::string &key, const std::string &value) {
+  if ("compression" == key) {
+    return SetCompressionOption(key, value);
+  }
```
PragmaTwice: IMHO we do not want special logic here, please state the reason.

src/storage/storage.cc
```diff
@@ -182,7 +182,61 @@ rocksdb::Options Storage::InitRocksDBOptions() {
...
+  return iter->second;
+}
+
+std::string Storage::CompressString2CompressionString(const std::string &type) {
+  std::unordered_map<std::string, std::string> compression_map = {{"no", "kNoCompression"},
+                                                                  {"snappy", "kSnappyCompression"},
+                                                                  {"lz4", "kLZ4Compression"},
+                                                                  {"zstd", "kZSTD"},
+                                                                  {"zlib", "kZlibCompression"}};
```
PragmaTwice: please extract these two maps as **static storage-duration**, **const** global variables.

src/storage/storage.cc
```diff
@@ -182,7 +182,61 @@ rocksdb::Options Storage::InitRocksDBOptions() {
...
+  }
+  return iter->second;
+}
+
+Status Storage::SetCompressionOption(const std::string &key, const std::string &value) {
+  std::string compression_string = CompressString2CompressionString(value);
+  if ("unknown" == compression_string) {
+    return {Status::NotOK, "Invalid compression type"};
+  }
+  std::string compression_levels = "kNoCompression:kNoCompression";
```
PragmaTwice: please write some comment here to state why there is no compression in first two levels.

PragmaTwice (changes requested):
the code does not look well to me, some fixes need to be pushed

tests/gocase/unit/config/config_test.go
```diff
@@ -108,3 +108,33 @@ func TestSetConfigBackupDir(t *testing.T) {
...
+	rList = r.Val().([]interface{})
+	require.EqualValues(t, rList[1], "lz4")
+
+	require.NoError(t, rdb.Do(ctx, "CONFIG", "SET", "rocksdb.compression", "zstd").Err())
+	require.Equal(t, "bar", rdb.Get(ctx, "foo").Val())
+	r = rdb.Do(ctx, "CONFIG", "GET", "rocksdb.compression")
+	rList = r.Val().([]interface{})
+	require.EqualValues(t, rList[1], "zstd")
+
+	require.ErrorContains(t, rdb.Do(ctx, "CONFIG", "SET", "rocksdb.compression", "undefine").Err(), "invalid enum option")
```
PragmaTwice: I wonder if these test code can trigger a compaction?

src/storage/storage.cc
```diff
@@ -182,6 +199,41 @@ rocksdb::Options Storage::InitRocksDBOptions() {
...
 
+std::string Storage::CompressType2String(const rocksdb::CompressionType &type) {
+  auto iter = compression_type_string_map.find(type);
+  if (iter == compression_type_string_map.end()) {
+    return "unknown";
+  }
+  return iter->second;
+}
+
+std::string Storage::CompressString2CompressionString(const std::string &type) {
```
PragmaTwice: i cannot understand the function name `CompressString2CompressionString`.

src/storage/storage.cc
```diff
@@ -56,6 +56,23 @@ const int64_t kIORateLimitMaxMb = 1024000;
 
 using rocksdb::Slice;
 
+const static std::map<std::string, std::string> compression_sort_long_string_map = {{"no", "kNoCompression"},
+                                                                                    {"snappy", "kSnappyCompression"},
+                                                                                    {"lz4", "kLZ4Compression"},
+                                                                                    {"zstd", "kZSTD"},
+                                                                                    {"zlib", "kZlibCompression"}};
+const static std::map<rocksdb::CompressionType, std::string> compression_type_string_map = {
```
```suggestion
const static std::map<rocksdb::CompressionType, std::string> compression_type_to_string_map = {
```

src/storage/storage.cc
```diff
@@ -56,6 +56,23 @@ const int64_t kIORateLimitMaxMb = 1024000;
 
 using rocksdb::Slice;
 
+const static std::map<std::string, std::string> compression_sort_long_string_map = {{"no", "kNoCompression"},
```
```suggestion
const static std::map<std::string, std::string> compression_string_to_rocksdb_option_map = {{"no", "kNoCompression"},
```

caipengbo: I think the following map is sufficient. Why do we need `compression_sort_long_string_map`?

src/storage/storage.cc
```diff
@@ -182,6 +199,41 @@ rocksdb::Options Storage::InitRocksDBOptions() {
...
+  for (size_t i = 2; i < db_->GetOptions().compression_per_level.size(); i++) {
+    compression_levels += ":";
+    compression_levels += compression_string;
+  }
+
+  for (auto &cf_handle : cf_handles_) {
+    auto s = db_->SetOptions(cf_handle, {{"compression_per_level", compression_levels}});
+    if (!s.ok()) return {Status::NotOK, s.ToString()};
+  }
+  return Status::OK();
```
```suggestion
  return SetOptionForAllColumnFamilies("compression_per_level", compression_levels);
```

src/storage/storage.cc
datavisorxiaobiaozhao: setOptions requires parameters of type string

caipengbo:
I think the `SetCompressionOption` parameter value is the name of the compression algorithm. We get `rocksdb::CompressionType` by `map[name]`(name->CompressionType), and then pass it to `SetOptions`, in `SetOptions` static_cast it to `rocksdb::CompressionType`.

git-hulk:
@PragmaTwice @xiaobiaozhao @caipengbo I minor refactor the PR, PTAL again.

git-hulk:
I have tested if the config set command works on my side:

Set the compression to LZ4:
```
2023/07/14-20:45:08.518174 6158610432 [db/db_impl/db_impl.cc:1195] SetOptions() on column family [default], inputs:
2023/07/14-20:45:08.518210 6158610432 [db/db_impl/db_impl.cc:1198] compression_per_level: kNoCompression:kNoCompression:kLZ4Compression:kLZ4Compression:kLZ4Compression:kLZ4Compression:kLZ4Compression
```

then use `redis-benchmark -p 6666  -n 1000000 -r 5000000 -t set -d 1024` to write the data, and check the db LOG to see if the compression was used:

```
2023/07/14-20:50:15.181736 6156890112 (Original Log Time 2023/07/14-20:50:15.181594) [db/compaction/compaction_job.cc:888] [metadata] compacted to: files[0 1 13 0 0 0 0] 
max score 0.93, MB/sec: 39.1 rd, 4.5 wr, level 2, files in(1, 1) out(1 +0 blob) MB 
in(58.0, 2.6 +0.0 blob) out(7.1 +0.0 blob), read-write-amplify(1.2) write-amplify(0.1) OK, 
records in: 89927, records dropped: 3779 output_compression: LZ4
```

src/config/config_type.h
```diff
@@ -186,15 +186,17 @@ class YesNoField : public ConfigField {
 
 class EnumField : public ConfigField {
  public:
-  EnumField(int *receiver, ConfigEnum *enums, int e) : receiver_(receiver), enums_(enums) { *receiver_ = e; }
+  EnumField(int *receiver, std::vector<ConfigEnum> &enums, int e) : receiver_(receiver), enums_(std::move(enums)) {
```
PragmaTwice:
```suggestion
  EnumField(int *receiver, std::vector<ConfigEnum> enums, int e) : receiver_(receiver), enums_(std::move(enums)) {
```

PragmaTwice: otherwise the passed `enums` will be invalid after the ctor call, which is very counterintuitive since it is passed as lvalue reference.

src/config/config.cc
```diff
@@ -45,42 +45,37 @@ constexpr const char *errBlobDbNotEnabled = "Must set rocksdb.enable_blob_files
...
 std::string TrimRocksDbPrefix(std::string s) {
   if (strncasecmp(s.data(), "rocksdb.", 8) != 0) return s;
   return s.substr(8, s.size() - 8);
 }
 
-int ConfigEnumGetValue(ConfigEnum *ce, const char *name) {
-  while (ce->name != nullptr) {
-    if (strcasecmp(ce->name, name) == 0) return ce->val;
-    ce++;
+int ConfigEnumGetValue(const std::vector<ConfigEnum> &enums, const std::string &name) {
```
PragmaTwice: I think this and the next function can be methods of `ConfigEnum`.

src/storage/storage.cc
```diff
@@ -54,8 +54,18 @@ constexpr const char *kReplicationIdKey = "replication_id_";
...
+    {rocksdb::kNoCompression, "no", "kNoCompression"},
+    {rocksdb::kSnappyCompression, "snappy", "kSnappyCompression"},
+    {rocksdb::kZlibCompression, "zlib", "kZlibCompression"},
+    {rocksdb::kLZ4Compression, "lz4", "kLZ4Compression"},
+    {rocksdb::kZSTD, "zstd", "kZSTD"},
+};
+
 using rocksdb::Slice;
 
+std::vector<CompressionOption> GetCompressionOptions() { return compressionOptions; }
```
```suggestion
const std::vector<CompressionOption>& GetCompressionOptions() { return compressionOptions; }
```

PragmaTwice: we do not need to construct a vector every time when we call the function, which is expensive. BTW I think it is no difference than just an `inline const` vector in header file.

src/config/config_type.h
git-hulk: Thanks @PragmaTwice, it makes a lot of sense

src/storage/storage.cc
git-hulk: Yes, it's better to put this in the header file.

src/config/config.cc
git-hulk: Good point, will do it soon.

git-hulk: I have done this by removing this function since only one place used them.

caipengbo (dismissed):

caipengbo (approved):
LGTM

PragmaTwice (approved):

PR #1577, opened 2023-07-11
Avoid Write calls when SIADD added_cnt is 0

enjoy-binbin:
This may improve performance by 40% in certain case.

Simple benchmarks:
```
src/redis-benchmark -p 6666 -P 100 -n 100000 siadd key 1 2 3
```

Before is 95057.03 requests per second.
After is 135135.14 requests per second.


git-hulk (approved):

xiaobiaozhao (approved):

xiaobiaozhao:
Why such a big performance boost? I don't get that from the logic of the code

PragmaTwice (approved):

PR #1596, opened 2023-07-18
Use LMOVE logic to handle RPOPLPUSH

enjoy-binbin:
RPOPLPUSH is actually LMOVE RIGHT LEFT, we can use
list_db.LMove to replace List::RPopLPush. In this way
we can enjoy the multi locks guard and operate multi
keys atomically.

### DIFF
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -508,7 +508,7 @@ class CommandRPopLPUSH : public Commander {
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     redis::List list_db(svr->storage, conn->GetNamespace());
     std::string elem;
-    auto s = list_db.LMove(args_[1], args_[2], false, true, &elem);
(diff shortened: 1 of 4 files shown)


git-hulk (dismissed):
Cool!

src/commands/cmd_list.cc
```diff
@@ -508,7 +508,7 @@ class CommandRPopLPUSH : public Commander {
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     redis::List list_db(svr->storage, conn->GetNamespace());
     std::string elem;
-    auto s = list_db.RPopLPush(args_[1], args_[2], &elem);
+    auto s = list_db.LMove(args_[1], args_[2], false, true, &elem);
```
mapleFU:
```suggestion
    auto s = list_db.LMove(args_[1], args_[2], /*src_left=*/false, /*dst_left=*/true, &elem);
```

mapleFU (approved):

torwig (approved):

PR #1597, opened 2023-07-19
Update error message and add the ERR error code

enjoy-binbin:
We now have some error messages that do not return
an ERR error code:
```
127.0.0.1:6666> cluster nodes
Cluster mode is not enabled
```

As far as i know this doesn't affect the client libs,
and doesn't affect the users. But I think it's right
to add the error code.

Note that Redis will actually count error messages and
standardize the error codes, like:
```
127.0.0.1:6379> info Errorstats
 # Errorstats
errorstat_ERR:count=2
errorstat_EXECABORT:count=1
errorstat_WRONGTYPE:count=1
```

We mixed using the status message and error output in
some places, this PR checks all redis::Error calls and
use error status to replace it. (Only the ERR case is
handled, others are not handled yet.)

We like to have the "ERR" + some_error_message construction
only in one place as an implementation detail of the Redis
protocol. Maybe, inside redis::Error or so.

We will leave it for a future PR as a cleanup. (pull out 
the error code and error message)

### DIFF
--- a/src/commands/cmd_cluster.cc
+++ b/src/commands/cmd_cluster.cc
@@ -52,13 +52,11 @@ class CommandCluster : public Commander {
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     if (!svr->GetConfig()->cluster_enabled) {
-      *output = redis::Error("Cluster mode is not enabled");
-      return Status::OK();
+      return {Status::RedisExecErr, "Cluster mode is not enabled"};
     }
 
     if (!conn->IsAdmin()) {
-      *output = redis::Error(errAdministorPermissionRequired);
-      return Status::OK();
+      return {Status::RedisExecErr, errAdministorPermissionRequired};
     }
 
     if (subcommand_ == "keyslot") {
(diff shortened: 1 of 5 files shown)


torwig:
It's not a direct objective of this PR, but I'd like to have the `"ERR" + some_error_message` construction **only** in one place as an implementation detail of the Redis protocol. Maybe, inside `redis::Error` or so.

torwig (approved):
Generally LGTM.

git-hulk (approved):

enjoy-binbin:
> It's not a direct objective of this PR, but I'd like to have the "ERR" + some_error_message construction only in one place as an implementation detail of the Redis protocol. Maybe, inside redis::Error or so.

@torwig thanks for the review. agree, that is a good suggestion.
we still have some places that mixed using the status message and error output. (e.g. NOSCRIPT)
i think we can leave it for a future PR as a cleanup. (pull out the error code and error message)

xiaobiaozhao (approved):

PR #1602, opened 2023-07-21
Fix: Enable portable flag for Docker container builds

ovaistariq:
Enable portable flag for Docker container builds

Fixes issue #1146

### DIFF
--- a/x.py
+++ b/x.py
@@ -108,7 +108,7 @@ def build(dir: str, jobs: Optional[int], ghproxy: bool, ninja: bool, unittest: b
 
     makedirs(dir, exist_ok=True)
 
-    cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo", "-DPORTABLE=0"]
+    cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo"]
     if ghproxy:
         cmake_options.append("-DDEPS_FETCH_PROXY=https://ghproxy.com/")
     if ninja:
@@ -119,6 +119,11 @@ def build(dir: str, jobs: Optional[int], ghproxy: bool, ninja: bool, unittest: b
         cmake_options += ["-DCMAKE_C_COMPILER=clang", "-DCMAKE_CXX_COMPILER=clang++"]
     if D:
         cmake_options += [f"-D{o}" for o in D]
+
+    portable_flag_enabled = any("DPORTABLE" in o for o in cmake_options)
+    if not portable_flag_enabled:
+        cmake_options.append("-DPORTABLE=0")
+
     run(cmake, str(basedir), *cmake_options, verbose=True, cwd=dir)
 
     if skip_build:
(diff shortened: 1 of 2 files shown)


git-hulk (approved):

x.py
```diff
@@ -119,6 +119,11 @@ def build(dir: str, jobs: Optional[int], ghproxy: bool, ninja: bool, unittest: b
         cmake_options += ["-DCMAKE_C_COMPILER=clang", "-DCMAKE_CXX_COMPILER=clang++"]
     if D:
         cmake_options += [f"-D{o}" for o in D]
+
+    portable_flag_enabled = any("DPORTABLE" in o for o in cmake_options)
```
mapleFU, line 123: would we require `portable_flag_enabled` be decided by `-DPORTABLE=1` or `-DPORTABLE=0`? Would

```
portable_flag_enabled = any("DPORTABLE" in o for o in cmake_options)
```

might cause `-DPORTABLE=1` disable portable?

aleksraiden:
Thanks for investigation, @ovaistariq! In my mind, options 1 and 0 are similar (with some case) - the compiler do as best as it can. But we need to can manual add this variable to compiler process. So, in case of #1146 a best way is change PORTABLE=znver1 (if CPU are a first family of EPYC arch). Change a default PORTABLE to 1 maybe a wrong case - as  describe in rocksdb source - _usually nothing to do; compiler default is typically the most general_ - but for hi-load system it's a wrong way, because we cant use an advanced CPU features and acceleration, 1 is a simple baseline CPU.

Maybe we can detect a right CPU platform directly in x.py Python script and use it?

aleksraiden (approved):

mapleFU (approved):

ovaistariq:
> Thanks for investigation, @ovaistariq! In my mind, options 1 and 0 are similar (with some case) - the compiler do as best as it can. But we need to can manual add this variable to compiler process. So, in case of #1146 a best way is change PORTABLE=znver1 (if CPU are a first family of EPYC arch). Change a default PORTABLE to 1 maybe a wrong case - as  describe in rocksdb source - _usually nothing to do; compiler default is typically the most general_ - but for hi-load system it's a wrong way, because we cant use an advanced CPU features and acceleration, 1 is a simple baseline CPU.
> 
> 
> 
> Maybe we can detect a right CPU platform directly in x.py Python script and use it?  

In my opinion the Docker images are supposed to be portable for the platform they are built on. So an image built for the amd64 platform should be able to run on all amd64 machines.

Secondly, the Docker image may be built on a machine with a different hardware spec then the machine the container is going to run on. For example in this case, the image is built on GitHub runners which are running Intel CPUs while the container is running on a machine with AMD CPUs.

git-hulk:
@aleksraiden This PR only changes the docker image PORTABLE value to 1. I think this makes a lot of sense since we cannot promise users are always running on the intel CPU like @ovaistariq mentioned.

PR #1611, opened 2023-07-26
Support change log-level in runtime

### LINKED ISSUE #1610: Please support change log-level in runtime
### Search before asking

### Motivation

version 2.5.0
error: Unsupported CONFIG parameter: log-level

### Solution

_No response_

enjoy-binbin:
Now we support config set log-level xxx to modify log-level
in runtime.

This closes #1610.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -121,7 +121,7 @@ Config::Config() {
       {"dir", true, new StringField(&dir, "/tmp/kvrocks")},
       {"backup-dir", false, new StringField(&backup_dir, "")},
       {"log-dir", true, new StringField(&log_dir, "")},
-      {"log-level", true, new EnumField(&log_level, log_levels, google::INFO)},
+      {"log-level", false, new EnumField(&log_level, log_levels, google::INFO)},
       {"pidfile", true, new StringField(&pidfile, "")},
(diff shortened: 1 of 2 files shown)


git-hulk (approved):

src/config/config.cc
```diff
@@ -478,6 +478,12 @@ void Config::initFieldCallback() {
          if (cluster_enabled) srv->slot_migrator->SetSequenceGapLimit(sequence_gap);
          return Status::OK();
        }},
+      {"log-level",
+       [this](Server *srv, const std::string &k, const std::string &v) -> Status {
+         if (!srv) return Status::OK();
+         FLAGS_minloglevel = log_level;
```
mapleFU, line 484: just a nit, should we not change it when `FLAGS_minloglevel` and `log_level` is equal?

enjoy-binbin: i am not very familiar with this (the details), if it's the same, is there any side effect of changing it?

mapleFU: Not so important, it will write a global variable, causing some cache invalid. But it's not important

enjoy-binbin: ohh, i got it, this is new to me, thanks for the explanation.

PR #1613, opened 2023-07-27
Fix -Wreturn-std-move warning

enjoy-binbin:
When building, the following warning appears:
```
warning: local variable 's' will be copied despite being returned by name [-Wreturn-std-move]
note: call 'std::move' explicitly to avoid copying
```

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -68,7 +68,7 @@ Status FeedSlaveThread::Start() {
     conn_ = nullptr;  // prevent connection was freed when failed to start the thread
   }
 
-  return s;
+  return std::move(s);
 }
 
 void FeedSlaveThread::Stop() {
--- a/src/config/config_type.h
+++ b/src/config/config_type.h
@@ -124,7 +124,7 @@ class IntegerField : public ConfigField {
   }
   Status Set(const std::string &v) override {
     auto s = ParseInt<IntegerType>(v, {min_, max_});
-    if (!s.IsOK()) return s;
+    if (!s.IsOK()) return std::move(s);
     *receiver_ = s.GetValue();
     return Status::OK();
   }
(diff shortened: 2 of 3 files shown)


enjoy-binbin:
i am not very familiar with it, so hopefully it is a right fix

torwig (approved):

PragmaTwice (changes requested):
Move a local variable to return is not a good way in C++ (especially modern C++), since:
1. the move in return is already implicit in c++ standard
2. it will prevent NRVO (named return value optimization).

refer to https://stackoverflow.com/questions/14856344/when-should-stdmove-be-used-on-a-function-return-value.

enjoy-binbin:
thanks for the link. so these warnings are false positives? do we have any other option to ignore it?

mapleFU:
I guess this would be fixed with a higher version of compiler or clang-tidy. https://en.cppreference.com/w/cpp/language/copy_elision C++17 has some gurantees

image

Even if NRVO not work well, the compiler will not copy. By the way, would you mind print the version of clang and clang-tidy? Maybe it would make things clear

enjoy-binbin:
> By the way, would you mind print the version of clang and clang-tidy?

sure, i will take a look when i get home. (i am using different PC when i am working)

enjoy-binbin:
```
$ clang -v
Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

PragmaTwice:
Seems your clang version is too old. Apple Clang 12 is based on LLVM/Clang 10, but the latest version of LLVM is 16.

BTW I recommend that you use LLVM/Clang rather than Apple Clang.

mapleFU:
Can this patch be closed? Since it's no problem in our current C++ version?

enjoy-binbin:
ohh, right, since there is no real issue, i am closing it.
I forgot to sync, i am having some env trouble in my laptop so i haven't found the time to verify it

PR #1615, opened 2023-07-28
Fix data race when updating the slots_info string

### LINKED ISSUE #1598: Kvrocks crashed in Cluster::updateSlotsInfo()
### Search before asking

### Version

- CentOS Linux release 7.9.2009 (Core)
- kvrocks 2.4.0

```
ldd /usr/local/bin/kvrocks2.4.0 
        linux-vdso.so.1 =>  (0x00007fff09cb9000)
        librt.so.1 => /lib64/librt.so.1 (0x00007f98c9f80000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f98c9d7c000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f98c9b60000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f98c985e000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f98c9490000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f98ca188000)

yum list glibc
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
Installed Packages
glibc.x86_64                                                                                                   2.17-326.el7_9                                                                                                   @Nexus-Update
Available Packages
glibc.i686                                                                                                     2.17-326.el7_9                                                                                                   Nexus-Update
```

### Minimal reproduce step

kernel: worker[45424]: segfault at 7fbea660c1 ip 0000000000aac777 sp 00007fbe9cdf27f0 error 4 in kvrocks2.4.0[406000+8cb000]

```
GNU gdb (GDB) Red Hat Enterprise Linux 7.6.1-120.el7
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>...
Reading symbols from /usr/local/bin/kvrocks2.4.0...done.
[New LWP 45424]
[New LWP 48250]
[New LWP 45381]
[New LWP 45421]
[New LWP 45382]
[New LWP 45418]
[New LWP 45426]
[New LWP 45422]
[New LWP 45414]
[New LWP 45417]
[New LWP 45420]
[New LWP 45427]
[New LWP 45383]
[New LWP 45425]
[New LWP 45423]
[New LWP 45380]
[New LWP 45419]
[New LWP 45372]
[New LWP 45416]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
Core was generated by `kvrocks2.4.0 -c /data/kvrocks/conf/kvrocks-cluster-6669.conf'.
Program terminated with signal 11, Segmentation fault.
#0  0x0000000000aac777 in evmap_io_active_ (base=base@entry=0x7fbeaa921200, fd=131, events=34) at /root/kvrocks-2.4.0/build/_deps/libevent-src/evmap.c:434

warning: Source file is more recent than executable.
434             LIST_FOREACH(ev, &ctx->events, ev_io_next) {
Missing separate debuginfos, use: debuginfo-install glibc-2.17-326.el7_9.x86_64
```

### What did you see instead?

I want it to run stably
But I can't solve this problem

### Anything Else?

_No response_

git-hulk:
Currently, we cached the slots_info inside the cluster node and update the slots_info every time before getting the node description. But we didn't use the exclusive lock except the cluster set command, so it may have data race between the read and write operation that causes the crash. This fix avoids this by removing the unnecessary cache field `slots_info` from the cluster node.

This fixes #1598

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -507,7 +507,7 @@ std::string Cluster::genNodesDescription() {
   return nodes_desc;
 }
 
-std::map<std::string, std::string> Cluster::getClusterNodeSlots() {
+const std::map<std::string, std::string> Cluster::getClusterNodeSlots() {
   int start = -1;
   // node id => slots info string
   std::map<std::string, std::string> slots_infos;
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -507,7 +507,7 @@ std::string Cluster::genNodesDescription() {
   return nodes_desc;
 }
 
-const std::map<std::string, std::string> Cluster::getClusterNodeSlots() {
+std::map<std::string, std::string> Cluster::getClusterNodeSlots() const {
   int start = -1;
   // node id => slots info string
   std::map<std::string, std::string> slots_infos;
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -472,7 +472,7 @@ Status Cluster::GetClusterNodes(std::string *nodes_str) {
 }
 
 std::string Cluster::genNodesDescription() {
-  updateSlotsInfo();
+  auto slots_infos = getClusterNodeSlots();
 
   auto now = util::GetTimeStampMS();
   std::string nodes_desc;
@@ -495,8 +495,11 @@ std::string Cluster::genNodesDescription() {
     // Ping sent, pong received, config epoch, link status
     node_str.append(fmt::format("{} {} {} connected", now - 1, now, version_));
 
-    if (n->role == kClusterMaster && n->slots_info.size() > 0) {
-      node_str.append(" " + n->slots_info);
+    if (n->role == kClusterMaster) {
+      auto iter = slots_infos.find(n->id);
+      if (iter != slots_infos.end() && iter->second.size() > 0) {
+        node_str.append(" " + iter->second);
+      }
     }
 
     nodes_desc.append(node_str + "\n");
@@ -504,13 +507,10 @@ std::string Cluster::genNodesDescription() {
   return nodes_desc;
 }
 
-void Cluster::updateSlotsInfo() {
+std::map<std::string, std::string> Cluster::getClusterNodeSlots() {
   int start = -1;
-  // reset the previous slots info
-  for (const auto &item : nodes_) {
-    const std::shared_ptr<ClusterNode> &n = item.second;
-    n->slots_info.clear();
-  }
+  // node id => slots info string
+  std::map<std::string, std::string> slots_infos;
 
   std::shared_ptr<ClusterNode> n = nullptr;
   for (int i = 0; i <= kClusterSlots; i++) {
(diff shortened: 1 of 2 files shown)


PragmaTwice (dismissed):

torwig:
@git-hulk @PragmaTwice With these changes, does the data race still exist? I mean the `Cluster::slots_nodes_` variable. Do the read operations obtain an exclusive lock when reading from that variable?

src/cluster/cluster.cc
```diff
@@ -524,24 +524,24 @@ void Cluster::updateSlotsInfo() {
...
       n = slots_nodes_[i];
       start = i;
     }
   }
 
-  for (const auto &item : nodes_) {
-    const std::shared_ptr<ClusterNode> n = item.second;
-    if (n->slots_info.size() > 0) n->slots_info.pop_back();  // Remove last space
+  for (const auto &iter : slots_infos) {
+    if (slots_infos[iter.first].size() > 0) slots_infos[iter.first].pop_back();  // Remove last space
```
```suggestion
  for (auto &[_, info] : slots_infos) {
    if (info.size() > 0) info.pop_back();  // Remove last space
```

src/cluster/cluster.cc
```diff
@@ -495,22 +495,22 @@ std::string Cluster::genNodesDescription() {
...
+      }
     }
 
     nodes_desc.append(node_str + "\n");
   }
   return nodes_desc;
 }
 
-void Cluster::updateSlotsInfo() {
+std::map<std::string, std::string> Cluster::getClusterNodeSlots() {
```
PragmaTwice: could it be marked to `const`?

git-hulk: sure, won't expect to change outside

git-hulk:
> e changes, does the data race still exist? I mean the `Cluster::slots_nodes_` variable. Do the read operations obtain an exclusive lock when reading from that variable?

Some of custerx/custer subcommands will be executed in exclusive mode like clusterx setnodes, so it won't have the data race in Cluster::slots_nodes_.

src/cluster/cluster.cc
```diff
@@ -495,22 +495,22 @@ std::string Cluster::genNodesDescription() {
...
+      }
     }
 
     nodes_desc.append(node_str + "\n");
   }
   return nodes_desc;
 }
 
-void Cluster::updateSlotsInfo() {
+const std::map<std::string, std::string> Cluster::getClusterNodeSlots() {
```
```suggestion
std::map<std::string, std::string> Cluster::getClusterNodeSlots() const {
```

caipengbo (approved):

PR #1622, opened 2023-07-30
Release 2.5.1

git-hulk:
**Highlights:** #1568 may the server crash when passing numkeys is -1 in Lua script, and #1615 is caused by the data race when updating the slots info string in cluster mode. That means it won't affect you if you're not using the Lua script or cluster mode.

## Bug Fixes
- #1568 
- #1567
- #1562 
- #1581 
- #1589 
- #1591 
- #1593
- #1576 
- #1556 
- #1595
- #1602 
- #1604 
- #1609 
- #1605 
- #1615

PR #1238, opened 2023-01-30
Disable the rebase-merge button

PragmaTwice:
As informally discussed in Slack, we disable the rebase-merge button, and just leave the squash-merge button to ensure that one commit (in the unstable branch) corresponds to one PR, so that the cherry-pick procedure can be more simple during releasing new versions.

PR #1624, opened 2023-07-31
Enable rebase merge in enabled_merge_buttons

enjoy-binbin:
In some cases, such as a release version, we will pick some
commits (note that doing a cherry-pick will change the commit date
but will not change the origin author date), then do a PR review,
and then merge the commits into the branch. In this case, we want to
use rebase-merge to keep the info of each commit instead of using squash-merge.
Note that we disable rebase-merge in #1238, now we are enabling it.

Or in some refactoring PRs, we can separate code moves (rename)
and actual changes into different commits. These single commits
we want to keep the commit message and doing a plain merge. And
then doing a plain merge can create a merge-commit that associate
the target PR and allow we writing a nice commit body describing
the PR (we use this as the merge-commit message). In this way,
when tracking changes, we can only focus on actual changes without
being affected by meaningless diffs. We don't enable plain-merge now.

When enabled, we can control it, the choice of squash or rebase-merge
is the responsibility of the person for merging the PR.

See #1622 for more discussion on release part.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -31,7 +31,7 @@ github:
     - redis-cluster
   enabled_merge_buttons:
     squash:  true
-    merge:   false
+    merge:   true
     rebase:  false
   protected_branches:
     unstable:
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -31,8 +31,8 @@ github:
     - redis-cluster
   enabled_merge_buttons:
     squash:  true
-    merge:   true
-    rebase:  false
+    merge:   false
+    rebase:  true
   protected_branches:
     unstable:
       required_pull_request_reviews:


enjoy-binbin:
hope to get your approval or comments :)

git-hulk (dismissed):

mapleFU (dismissed):

torwig (dismissed):

git-hulk:
@tisonkun @PragmaTwice Does this sound good to you? with the merge button would also make the release process more smooth.

tisonkun (changes requested):
No. In this case we use rebase and merge.

enjoy-binbin:
if I remember correctly, rebase merge in github will change the date of those commits to the current date? if that's the case, i would like to use merge to keep those commits with the truest commit time

or maybe i am wrong, i am ok with rebase-merge if this is not the case

tisonkun:
> if I remember correctly, rebase merge in github will change the date of those commits to the current date? if that's the case, i would like to use merge to keep those commits with the truest commit time

When you do cherry-pick, the date already becomes the pick date instead of the original commit. I don't think the date is significant.

And I check that rebase and merge won't change the commit date. (But it's the picked date, not the original commit date)

enjoy-binbin:
ok, as long as the author date is not changing, i am ok with that.

you are right, when doing pick, the commit date is changing (but the author time will not changing, i am seeing the author date in the default format in my PC, and the github seem to show the commit date by default, so it cause me some confusions)

i want mention that the another (or the only one) benefit of plain merge is that we can have a merge commit that associate the target PR, but it doesn't seem to matter in the release part. (But it does have a use case for sure)

i am changing the PR to enable the rebase-merge, for the release part.
thanks for the confirming btw

.asf.yaml
```diff
@@ -31,7 +31,7 @@ github:
     - redis-cluster
   enabled_merge_buttons:
     squash:  true
-    merge:   false
+    merge:   true
     rebase:  false
```
enjoy-binbin:
```suggestion
    merge:   false
    rebase:  true
```

git-hulk (approved):

tisonkun (approved):

PR #1625, opened 2023-07-31
Fix GEOSEARCH/GEOSEARCHSTORE FROMMEMBER against non existing src key reply

enjoy-binbin:
The new GEOSEARCH and GEOSEARCHSTORE commands was added
in #1533. When typing FROMMEMBER option against non-existing
src key, an IsNotFound error is return and resulting in the
following inconsistency:
```
127.0.0.1:6666> GEOSEARCH src FROMMEMBER Shenzhen BYBOX 88 88 m
(error) ERR NotFound:
127.0.0.1:6666> GEOSEARCHSTORE dst src FROMMEMBER Shenzhen BYBOX 88 88 m
(error) ERR NotFound:

127.0.0.1:6379> GEOSEARCH src FROMMEMBER Shenzhen BYBOX 88 88 m
(empty array)
127.0.0.1:6379> GEOSEARCHSTORE dst src FROMMEMBER Shenzhen BYBOX 88 88 m
(integer) 0
```

### DIFF
--- a/tests/gocase/unit/geo/geo_test.go
+++ b/tests/gocase/unit/geo/geo_test.go
@@ -139,6 +139,11 @@ func TestGeo(t *testing.T) {
 		require.EqualValues(t, []interface{}{nil, nil, nil}, rdb.Do(ctx, "GEOHASH", "points", "a", "b", "c").Val())
 	})
 
+	t.Run("GEOSEARCH against non existing src key", func(t *testing.T) {
+		require.NoError(t, rdb.Del(ctx, "points").Err())
+		require.EqualValues(t, []interface{}([]interface {}{}), rdb.Do(ctx, "GEOSEARCH", "src", "FROMMEMBER", "Shenzhen", "BYBOX", 88, 88, "m").Val())
+	})
+
 	t.Run("GEOSEARCH simple", func(t *testing.T) {
 		require.NoError(t, rdb.Del(ctx, "points").Err())
 		require.NoError(t, rdb.GeoAdd(ctx, "points",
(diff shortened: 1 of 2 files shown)


torwig (dismissed):
@enjoy-binbin Good catch!

mapleFU (dismissed):
Seems that geo not handle the return value of `ZSet::Del(user_key)`, would this also need to be handled?

enjoy-binbin:
> Seems that geo not handle the return value of ZSet::Del(user_key), would this also need to be handled?

i haven't looked at the store section below yet, just a quick look, looks like it doesn't need it.

tests/gocase/unit/geo/geo_test.go
```diff
@@ -139,6 +139,11 @@ func TestGeo(t *testing.T) {
 		require.EqualValues(t, []interface{}{nil, nil, nil}, rdb.Do(ctx, "GEOHASH", "points", "a", "b", "c").Val())
 	})
 
+	t.Run("GEOSEARCH against non existing src key", func(t *testing.T) {
+		require.NoError(t, rdb.Del(ctx, "points").Err())
+		require.EqualValues(t, []interface{}([]interface {}{}), rdb.Do(ctx, "GEOSEARCH", "src", "FROMMEMBER", "Shenzhen", "BYBOX", 88, 88, "m").Val())
```
enjoy-binbin, line 144:
```suggestion
		require.EqualValues(t, []interface{}([]interface{}{}), rdb.Do(ctx, "GEOSEARCH", "src", "FROMMEMBER", "Shenzhen", "BYBOX", 88, 88, "m").Val())
```

git-hulk (approved):

torwig (approved):

PragmaTwice (approved):

PR #1627, opened 2023-08-01
Fix PORTABLE variable in CMake

PragmaTwice:
PORTABLE need to be a string variable rather than an option.

BTW, USE_LUAJIT is renamed to ENABLE_LUAJIT for better aligning.


CMakeLists.txt
```diff
@@ -25,11 +25,11 @@ option(ENABLE_ASAN "enable address sanitizer" OFF)
 option(ENABLE_TSAN "enable thread sanitizer" OFF)
 option(ASAN_WITH_LSAN "enable leak sanitizer while address sanitizer is enabled" ON)
 option(ENABLE_STATIC_LIBSTDCXX "link kvrocks with static library of libstd++ instead of shared library" ON)
-option(USE_LUAJIT "use luaJIT instead of lua" ON)
+option(ENABLE_LUAJIT "enable use of luaJIT instead of lua" ON)
 option(ENABLE_OPENSSL "enable openssl to support tls connection" OFF)
 option(ENABLE_IPO "enable interprocedural optimization" ON)
 option(ENABLE_UNWIND "enable libunwind in glog" ON)
-option(PORTABLE "build a portable binary (disable arch-specific optimizations)" 0)
+set(PORTABLE 0 CACHE STRING "build a portable binary (disable arch-specific optimizations)")
```
git-hulk, line 32: So this default value didn't take effect is caused by the wrong type?

PragmaTwice: yeah. option can only be ON or OFF, cannot be a string.

git-hulk: cool, thanks for your confirmation.

git-hulk (approved):

PR #1631, opened 2023-08-03
Fix MSETNX not allow overriding the same key

enjoy-binbin:
The changes in #337 causing this issue:
```
127.0.0.1:6379> flushall
OK
127.0.0.1:6379> msetnx k v1 k v2
(integer) 1
127.0.0.1:6379> get k
"v2

127.0.0.1:6666> flushall
OK
127.0.0.1:6666> msetnx k v1 k v2
(integer) 0
127.0.0.1:6666> get k
"v1"
```

Redis allow we overriding the same key but kvrocks will
fail in this case. The way to handle it is to revert the
changes in #337.

So this PR is based on that, after reverting the changes
of #337, we can reuse the logic of MSet. And hulk mentions
that MSETNX is an exclusive command and we better to lock
those keys and then remove the exclusive flag from the command.

So we use MultiLockGuard before Exists call to lock multi
keys. And change MSet to support non-lock calls. And also
remove the exclusive flag from MSETNX command.

Co-authored-by: hulk <hulk.website@gmail.com>

### DIFF
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -24,6 +24,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <limits>
+#include <optional>
 #include <string>
 
 #include "parse_util.h"
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -213,12 +213,12 @@ rocksdb::Status String::GetDel(const std::string &user_key, std::string *value) 
 
 rocksdb::Status String::Set(const std::string &user_key, const std::string &value) {
   std::vector<StringPair> pairs{StringPair{user_key, value}};
-  return MSet(pairs, 0);
+  return MSet(pairs, 0, true);
 }
 
 rocksdb::Status String::SetEX(const std::string &user_key, const std::string &value, uint64_t ttl) {
   std::vector<StringPair> pairs{StringPair{user_key, value}};
-  return MSet(pairs, ttl);
+  return MSet(pairs, ttl, true);
 }
 
 rocksdb::Status String::SetNX(const std::string &user_key, const std::string &value, uint64_t ttl, bool *flag) {
@@ -363,7 +363,7 @@ rocksdb::Status String::IncrByFloat(const std::string &user_key, double incremen
   return updateRawValue(ns_key, raw_value);
 }
 
-rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl) {
+rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl, bool lock) {
   uint64_t expire = 0;
   if (ttl > 0) {
     uint64_t now = util::GetTimeStampMS();
@@ -384,7 +384,7 @@ rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl)
     batch->PutLogData(log_data.Encode());
     AppendNamespacePrefix(pair.key, &ns_key);
     batch->Put(metadata_cf_handle_, ns_key, bytes);
-    LockGuard guard(storage_->GetLockManager(), ns_key);
+    if (lock) LockGuard guard(storage_->GetLockManager(), ns_key);
     auto s = storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
     if (!s.ok()) return s;
   }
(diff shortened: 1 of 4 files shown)


src/types/redis_string.cc
```diff
@@ -213,12 +213,12 @@ rocksdb::Status String::GetDel(const std::string &user_key, std::string *value)
 
 rocksdb::Status String::Set(const std::string &user_key, const std::string &value) {
   std::vector<StringPair> pairs{StringPair{user_key, value}};
-  return MSet(pairs, 0);
+  return MSet(pairs, 0, true);
```
mapleFU: We'd better mark the argument:
```suggestion
  return MSet(pairs, /*ttl=*/0, /*lock=*/true);
```

src/types/redis_string.cc
```diff
@@ -384,7 +384,7 @@ rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl)
     batch->PutLogData(log_data.Encode());
     AppendNamespacePrefix(pair.key, &ns_key);
     batch->Put(metadata_cf_handle_, ns_key, bytes);
-    LockGuard guard(storage_->GetLockManager(), ns_key);
+    if (lock) LockGuard guard(storage_->GetLockManager(), ns_key);
```
mapleFU: Would lifetime of LockGuard be a temperal?

```
std::optional<LockGuard> guard;
if (lock) {
  guard = ...
}
```

What about this?

mapleFU (commented):
Didn't review it carefully, just a tiny round.

src/types/redis_string.cc
git-hulk: Yes, this lock wouldn't work since it will release the lock after exiting the if block. It also has a problem with the previous implementation which wrote the batch per key, we could change it to a single batch in this PR.

enjoy-binbin: > It also has a problem with the previous implementation which wrote the batch per key, we could change it to a single batch in this PR.

i want to put this in another PR, just to make the PRs clear (the diff should be small and easy to follow), WDYT?

git-hulk: @enjoy-binbin It makes a lot of sense.

git-hulk (dismissed):

src/types/redis_string.cc
```diff
@@ -384,7 +384,10 @@ rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl)
     batch->PutLogData(log_data.Encode());
     AppendNamespacePrefix(pair.key, &ns_key);
     batch->Put(metadata_cf_handle_, ns_key, bytes);
-    LockGuard guard(storage_->GetLockManager(), ns_key);
+    std::optional<LockGuard> guard;
+    if (lock) {
+      guard(storage_->GetLockManager(), ns_key);
+    }
```
PragmaTwice: TBH I cannot understand the C++ code at line 389. You can try something like:
```c++
if (lock) storage_->GetLockManager().Lock(...);
...
if (lock) storage_->GetLockManager().Unlock(...);
```

Or, if you want a more decent way:
```c++
std::optional<LockGuard> guard;
if (lock) {
  guard.emplace(storage_->GetLockManager(), ns_key);
}
```

Since *move ctor* and *move assignment* for `LockGuard` is missing, the assignment `guard = LockGuard(storage_->GetLockManager(), ns_key)` after the ctor call of `std::optional<LockGuard>` may be not possible.

Fore more details, I think you can check [*C++ Primer*](https://www.amazon.co.jp/-/en/Stanley-Lippman/dp/0321714113).

mapleFU: Sorry I didn't catch catch `LockGuard` miss move ctor. Sorry for being misleading, since `std::unique_lock` and `std::lock_guard` has it.

enjoy-binbin: sorry. i am not very familiar with C++ (i will find some times to catch it up), thanks a lot for the review (suggestion)

src/types/redis_string.cc
```diff
@@ -394,41 +398,28 @@ rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl)
...
-    uint64_t now = util::GetTimeStampMS();
-    expire = now + ttl;
-  }
-
   int exists = 0;
+  std::string ns_key;
+  std::vector<std::string> lock_keys;
   std::vector<Slice> keys;
   keys.reserve(pairs.size());
+
```
mapleFU, line 407: `lock_keys.reserve(pairs.size();` ?

src/types/redis_string.cc
```diff
@@ -394,41 +398,28 @@ rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl)
...
-    metadata.Encode(&bytes);
-    bytes.append(pair.value.data(), pair.value.size());
-    auto batch = storage_->GetWriteBatchBase();
-    WriteBatchLogData log_data(kRedisString);
-    batch->PutLogData(log_data.Encode());
-    batch->Put(metadata_cf_handle_, ns_key, bytes);
-    auto s = storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
-    if (!s.ok()) return s;
-  }
+  rocksdb::Status s = MSet(pairs, ttl, false);
```
```
MSet(pairs, /*ttl=*/ttl, /*lock=*/false);
```

mapleFU:
After a round of review I get to understand this patch. There is a problem:

```c++
class MultiLockGuard {
 public:
  explicit MultiLockGuard(LockManager *lock_mgr, const std::vector<std::string> &keys) : lock_mgr_(lock_mgr) {
    locks_ = lock_mgr_->MultiGet(keys);
    for (const auto &iter : locks_) {
      iter->lock();
    }
  }

  ~MultiLockGuard() {
    // Lock with order `A B C` and unlock should be `C B A`
    for (auto iter = locks_.rbegin(); iter != locks_.rend(); ++iter) {
      (*iter)->unlock();
    }
  }

  MultiLockGuard(const MultiLockGuard &) = delete;
  MultiLockGuard &operator=(const MultiLockGuard &) = delete;

 private:
  LockManager *lock_mgr_ = nullptr;
  std::vector<std::mutex *> locks_;
};
```

Previously, the lock happens per-key. It was a bit wierd but would not introduce problem. Now, `exclusive` is removed, so it might introduce the deadlock problem.

A proposal is that using std::lock( https://en.cppreference.com/w/cpp/thread/lock ) instead of loop in `MultiLockGuard`.

@PragmaTwice would that be possible?

git-hulk:
@mapleFU I think it won't have the deadlock for the multi-keys lock since it will sort keys first: https://github.com/apache/kvrocks/blob/unstable/src/storage/lock_manager.cc#L43

mapleFU:
Nice I didn't find out that.

mapleFU (approved):

git-hulk (approved):

PR #1632, opened 2023-08-03
[WIP] BloomFilter command Draft and implemention

### LINKED ISSUE #1500: Add support of RedisBloom
This issue is part of a project from the Open Source Promotion Plan (OSPP, https://summer-ospp.ac.cn/org/prodetail/239430147), which intent to port BloomFilter and Cuckoo Filter to kvrocks, like RedisBloom(https://redis.io/docs/stack/bloom/).

Deliverables:
- The code and testing for BloomFilter and Cuckoo Filter implementions
- Performance benchmark for them

cc @mapleFU

zncleon:
This is part of Redis BloomFilter Implemention.

- [x] BF.ADD, BF.MADD basic impl
- [x] BF.EXIST, BF.MEXIST basic impl
- [x] BF.RESERVE basic impl
- [x] BF.INSERT basic impl
- [x] BF.CARD basic impl
- [x] BF.INFO basic impl
- [ ] which cf to save bfmetadata need to think
- [ ] some "todo" need to modify 
- [ ] more testcases

This closes https://github.com/apache/kvrocks/issues/1500


src/types/redis_bloomfilter.h
```diff
@@ -0,0 +1,88 @@
...
+const double kBFDefaultErrorRate = 0.01;
+const uint16_t kBFDefaultExpansion = 2;
+const double kErrorTighteningRatio = 0.5;
+
+struct BloomHashval {
+  uint64_t a;
+  uint64_t b;
+};
+
+struct BFInsertOptions {
```
mapleFU, line 37:
```
struct BFInsertOptions {
  uint64_t capacity = kBFDefaultInitCapacity;
  double error_rate = kBFDefaultErrorRate;
 ...
```

Nit: I guess provide default values would be better here. C might not have default argument. But we can make full use of it.

src/common/murmurhash2.h
```diff
@@ -0,0 +1,22 @@
+//-----------------------------------------------------------------------------
+// MurmurHash2 was written by Austin Appleby, and is placed in the public
+// domain. The author hereby disclaims copyright to this source code.
```
mapleFU: This is nice, but should we vendor this in thirdparty library or vendor at somewhere? @PragmaTwice

src/commands/cmd_bloomfilter.cc
```diff
@@ -0,0 +1,284 @@
...
+    if (!parse_capacity) {
+      return {Status::RedisParseErr, errValueNotInteger};
+    }
+    capacity_ = *parse_capacity;
+    if (capacity_ <= 0) {
+      return {Status::RedisParseErr, "capacity should be larger than 0"};
+    }
+
+    size_t expect_size = 4;
+    auto iter = find(args.begin(), args.end(), "NONSCALING");
```
mapleFU, line 50:
```
    while (parser.Good()) {
     if (parser.EatEqICaseFlag("NONSCALING", ...)) {
        ...
      } else {
      }
    }
```

Using parser like this might helps

mapleFU (commented):
Nice! I'll take time to dive and give a detail review tonight

src/common/murmurhash2.h
PragmaTwice: I think currently it is ok to put it in `common/`, but we may move it to a vendor directory later.

src/server/redis_reply.h
```diff
@@ -40,6 +40,15 @@ std::string Integer(T data) {
   return ":" + std::to_string(data) + CRLF;
 }
 
+template <typename T, std::enable_if_t<std::is_integral_v<T>, int> = 0>
+std::string MultiInteger(std::vector<T> &multi_data) {
```
mapleFU, line 44:
```suggestion
std::string MultiInteger(const std::vector<T> &multi_data) {
```

src/types/redis_bloomfilter.cc
```diff
@@ -0,0 +1,378 @@
...
+
+void BloomFilter ::appendBFMetaSuffix(const Slice &bf_key, std::string *output) {
+  output->clear();
+
+  output->append(bf_key.data(), bf_key.size());
+  output->append(kBloomFilterSeparator);
+  output->append("meta");
+}
+
+void BloomFilter::bfInit(BFMetadata *bf_metadata, uint64_t entries, double error, unsigned int options) {
```
mapleFU: `unsigned int` would better be `uint32_t`

src/types/redis_bloomfilter.cc
```diff
@@ -0,0 +1,378 @@
...
+  rocksdb::Status s = getSBChainMetadata(ns_key, &sb_chain_metadata);
+  if (!s.ok() && !s.IsNotFound()) return s;
+  if (!s.IsNotFound()) {
+    return rocksdb::Status::InvalidArgument("item exists");
+  }
+
+  return createSBChain(ns_key, error_rate, capacity, expansion, scaling, sb_chain_metadata);
+}
+
+rocksdb::Status BloomFilter::Add(const Slice &user_key, const Slice &item, int &ret) {
```
mapleFU:
```suggestion
rocksdb::Status BloomFilter::Add(const Slice &user_key, const Slice &item, int* ret) {
```

Personally, I perfer `pointer` for output argument.

src/storage/redis_metadata.h
```diff
@@ -197,3 +199,33 @@ class StreamMetadata : public Metadata {
   void Encode(std::string *dst) override;
   rocksdb::Status Decode(const std::string &bytes) override;
 };
+
+class SBChainMetadata : public Metadata {
+ public:
+  uint16_t n_filters;
+  uint16_t options;
```
mapleFU, line 207: RedisBloom uses `options`. it use `BLOOM_OPT_FORCE64` and `BLOOM_OPT_NO_SCALING` etc. We can not compability with them. And if we'd like to keep compatible, maybe we need some comments here.

src/storage/redis_metadata.h
```diff
@@ -143,6 +144,7 @@ class Metadata {
   virtual void Encode(std::string *dst);
   virtual rocksdb::Status Decode(const std::string &bytes);
   bool operator==(const Metadata &that) const;
+  virtual ~Metadata() = default;
```
mapleFU, line 147: Nice catch! cc @git-hulk seems we have this for a long time?

src/storage/redis_metadata.h
```diff
@@ -197,3 +199,33 @@ class StreamMetadata : public Metadata {
...
+  uint8_t n2;
+  uint64_t entries;
+
+  double error;
+  double bpe;
+
+  uint64_t bf_bits;
+  uint64_t bf_bytes;
+
+  explicit BFMetadata(bool generate_version = true) : Metadata(kRedisBloomFilter, generate_version) {}
```
mapleFU, line 222: Should this better to be a String or something?

src/storage/redis_metadata.h
mapleFU: Oh previously this might work, because seems we didn't `delete Metadata` when subclass like `ListMetadata` is used. However I think we'd better has this.

src/commands/cmd_bloomfilter.cc
```diff
@@ -0,0 +1,284 @@
...
+
+    *output = redis::MultiInteger(rets);
+    return Status::OK();
+  }
+
+ private:
+  std::vector<Slice> items_;
+};
+
+class CommandBFINSERT : public Commander {
```
mapleFU, line 132: Maybe `CommandBFInsert` would be ok here?

src/types/redis_bloomfilter.cc
```diff
@@ -0,0 +1,378 @@
...
+  BFInsertOptions insert_options = {.capacity = kBFDefaultInitCapacity,
+                                    .error_rate = kBFDefaultErrorRate,
+                                    .autocreate = 1,
+                                    .expansion = kBFDefaultExpansion,
+                                    .scaling = 1};
+  return InsertCommon(user_key, items, &insert_options, rets);
+}
+
+rocksdb::Status BloomFilter::InsertCommon(const Slice &user_key, const std::vector<Slice> &items,
+                                          const BFInsertOptions *insert_options, std::vector<int> &rets) {
```
mapleFU:
```suggestion
                                          const BFInsertOptions& insert_options, std::vector<int>* rets) {
```

I think it's better to mark `insert_options` as const reference here.

src/types/redis_bloomfilter.cc
```diff
@@ -0,0 +1,378 @@
...
+  std::string ns_key;
+  AppendNamespacePrefix(user_key, &ns_key);
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+  SBChainMetadata sb_chain_metadata;
+  rocksdb::Status s = getSBChainMetadata(ns_key, &sb_chain_metadata);
+  if (!s.ok()) return s;
+
+  std::vector<BFMetadata> bf_metadata_list;
+  for (u_int16_t i = 0; i < sb_chain_metadata.n_filters; ++i) {
```
mapleFU:
```suggestion
  for (uint16_t i = 0; i < sb_chain_metadata.n_filters; ++i) {
```

It's this a typo?

src/types/redis_bloomfilter.cc
```diff
@@ -0,0 +1,378 @@
...
+rocksdb::Status BloomFilter::MExist(const Slice &user_key, const std::vector<Slice> &items, std::vector<int> &rets) {
+  std::string ns_key;
+  AppendNamespacePrefix(user_key, &ns_key);
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+  SBChainMetadata sb_chain_metadata;
+  rocksdb::Status s = getSBChainMetadata(ns_key, &sb_chain_metadata);
+  if (!s.ok()) return s;
+
+  std::vector<BFMetadata> bf_metadata_list;
```
mapleFU, line 272: Can be
```suggestion
  std::vector<BFMetadata> bf_metadata_list;
  bf_metadata_list.reserve(sb_chain_metadata.n_filters);
```

src/types/redis_bloomfilter.cc
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  for (size_t i = 0; i < items.size(); ++i) {
+    const auto &item = items[i];
+    BloomHashval h = bloomCalcHash64(item.data(), static_cast<int>(item.size()));
+    // check
+    auto batch = storage_->GetWriteBatchBase();  // todo: as placeholder, change the function to delete this
+    for (int ii = sb_chain_metadata.n_filters - 1; ii >= 0; --ii) {
+      std::string bf_key;
+      appendBFSuffix(ns_key, ii, &bf_key);
```
mapleFU, line 289: Can we avoid re-caculate this by `bf_metadata_list`? Seems it's calculated once

src/storage/redis_metadata.h
mapleFU: After reviewing RedisBloom's code.

```C
/**
 * Common function for adding one or more items to a bloom filter.
 * capacity and error rate must not be 0.
 */
static SBChain *bfCreateChain(RedisModuleKey *key, double error_rate, size_t capacity,
                              unsigned expansion, unsigned scaling) {
    SBChain *sb = SB_NewChain(capacity, error_rate, BLOOM_OPT_FORCE64 | scaling | BLOOM_OPT_NOROUND,
                              expansion);
    if (sb != NULL) {
        RedisModule_ModuleTypeSetValue(key, BFType, sb);
    }
    return sb;
}
```

It uses `options` for compability. I guess we don't need to keep all same as them(we don't have existing data). We can just simplify this options and bloom filter.

src/storage/redis_metadata.h
```diff
@@ -197,3 +199,33 @@ class StreamMetadata : public Metadata {
...
+  uint16_t growth;
+
+  explicit SBChainMetadata(bool generate_version = true) : Metadata(kRedisBloomFilter, generate_version) {}
+
+  void Encode(std::string *dst) override;
+  rocksdb::Status Decode(const std::string &bytes) override;
+};
+
+class BFMetadata : public Metadata {
+ public:
```
mapleFU, line 217: As a simple implemention. We can use an aligned bloom filter. And keep only:

```
class BFMetadata : public Metadata {
 public:
  uint32_t hashes;
  uint64_t entries;

  double error;
  double bpe;

  uint64_t bf_bytes;
};
```

mapleFU, line 217: As a simple implemention. We can use an aligned bloom filter. And keep only:

```C
class BFMetadata : public Metadata {
 public:
  uint32_t hashes;
  uint64_t entries;

  double error;
  double bpe;

  uint64_t bf_bytes;
};
```

We can even implement it easier using a BlockSplitBloomFilter.

src/storage/redis_metadata.cc
```diff
@@ -401,3 +401,85 @@ rocksdb::Status StreamMetadata::Decode(const std::string &bytes) {
...
+  }
+
+  if (input.size() < 8 + CommonEncodedSize()) {
+    return rocksdb::Status::InvalidArgument(kErrMetadataTooShort);
+  }
+
+  GetFixed64(&input, &version);
+  GetFixedCommon(&input, &size);
+
+  if (input.size() < 45) {
```
mapleFU: Prefer constexpr than magic numbers.

src/types/redis_bloomfilter.cc
```diff
@@ -0,0 +1,378 @@
...
+    if (byte_index >= value.size()) {  // expand the bloomfilter
+      size_t expand_size = 0;
+      if (byte_index >= value.size() * 2) {
+        expand_size = byte_index - value.size() + 1;
+      } else if (value.size() * 2 > kBloomFilterSegmentBytes) {
+        expand_size = kBloomFilterSegmentBytes - value.size();
+      } else {
+        expand_size = value.size();
+      }
+      value.append(expand_size, 0);
```
mapleFU: So, currently, we'll try our best to saving memories, avoid too large value?

src/storage/redis_metadata.h
```diff
@@ -197,3 +199,33 @@ class StreamMetadata : public Metadata {
   void Encode(std::string *dst) override;
   rocksdb::Status Decode(const std::string &bytes) override;
 };
+
+class SBChainMetadata : public Metadata {
```
mapleFU, line 203: We'd better add an version to metadata to support future evolution

mapleFU:
This commit ( https://github.com/apache/kvrocks/pull/1632/commits/acf50104fbbecc84e363dfd88f750bf52c46f0e7 ) looks ok, but would you mind create an single patch for it?

You can also add an issue as a tracker for BloomFilter

src/storage/redis_metadata.h
git-hulk: Not sure what's SB short for which words, Scalable BloomFilter?

zncleon: Sure, SB is short for Scalable BloomFilter as RedisBloom do.

mapleFU:
Close as doing in other patch

PR #1634, opened 2023-08-04
MSET change it to batch write instead of batch per key

enjoy-binbin:
The previous implementation which wrote the batch per key,
we should change it to a single batch. Also use the multi-key
lock before the write operations instead of inside the loop.

This is mention in #1631, found by hulk.


git-hulk (dismissed):

src/types/redis_string.cc
```diff
@@ -373,22 +373,20 @@ rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl)
...
     Metadata metadata(kRedisString, false);
     metadata.expire = expire;
     metadata.Encode(&bytes);
     bytes.append(pair.value.data(), pair.value.size());
-    auto batch = storage_->GetWriteBatchBase();
-    WriteBatchLogData log_data(kRedisString);
-    batch->PutLogData(log_data.Encode());
     AppendNamespacePrefix(pair.key, &ns_key);
     batch->Put(metadata_cf_handle_, ns_key, bytes);
     LockGuard guard(storage_->GetLockManager(), ns_key);
```
mapleFU: what does this guard for?

git-hulk: It uses to tell tools like `kvrocks2redis` about how to process the current batch.

mapleFU: Previously, the code is

```
LockGuard guard(storage_->GetLockManager(), ns_key);
auto s = storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
if (!s.ok()) return s;
```

but now `storage_->Write` is removed, so I wonder if LockGuard is a nop?

git-hulk: I misunderstood your point. We need to use multi-lock guard here.

mapleFU: So lets wait for another patch merged first

src/types/redis_string.cc
```diff
@@ -374,25 +374,23 @@ rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl,
...
     Metadata metadata(kRedisString, false);
     metadata.expire = expire;
     metadata.Encode(&bytes);
     bytes.append(pair.value.data(), pair.value.size());
-    auto batch = storage_->GetWriteBatchBase();
-    WriteBatchLogData log_data(kRedisString);
-    batch->PutLogData(log_data.Encode());
     AppendNamespacePrefix(pair.key, &ns_key);
     batch->Put(metadata_cf_handle_, ns_key, bytes);
     std::optional<LockGuard> guard;
```
git-hulk: @enjoy-binbin We need to use the multi-key lock before the write operation instead of inside the loop. Because the for loop only inserts key-values into the write batch and it won't affect the DB before the write operation.

enjoy-binbin: thanks, i overlook the comment, and just doing a plain merge to update the branch.

git-hulk (dismissed):

xiaobiaozhao (dismissed):

src/types/redis_string.cc
```diff
@@ -371,28 +371,33 @@ rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl,
...
 
-  // Data race, key string maybe overwrite by other key while didn't lock the key here,
+  // Data race, key string maybe overwrite by other key while didn't lock the keys here,
   // to improve the set performance
   std::string ns_key;
+  std::optional<MultiLockGuard> guard;
+  if (lock) {
+    std::vector<std::string> lock_keys;
+    lock_keys.reserve(pairs.size());
+    for (StringPair pair : pairs) {
```
```suggestion
    for (const StringPair& pair : pairs) {
```

In previous patch, pair need to be copied, but here we doesn't need that.

src/types/redis_string.cc
```diff
@@ -371,28 +371,33 @@ rocksdb::Status String::MSet(const std::vector<StringPair> &pairs, uint64_t ttl,
     expire = now + ttl;
   }
 
-  // Data race, key string maybe overwrite by other key while didn't lock the key here,
+  // Data race, key string maybe overwrite by other key while didn't lock the keys here,
   // to improve the set performance
   std::string ns_key;
```
mapleFU, line 376: Can we move `ns_key` in the loop?

```c++
      std::string ns_key
      AppendNamespacePrefix(pair.key, &ns_key);
    }

**Review by mapleFU — COMMENTED** [review]
Rest LGTM

File: src/types/redis_string.cc (hunk 4 above)
Comment (enjoy-binbin) [reply src/types/redis_string.cc:376]: In the code below, we also use `ns_key`
```
  WriteBatchLogData log_data(kRedisString);
  for (const auto &pair : pairs) {
    std::string bytes;
    metadata.expire = expire;
    metadata.Encode(&bytes);
    bytes.append(pair.value.data(), pair.value.size());
    AppendNamespacePrefix(pair.key, &ns_key);
    batch->Put(metadata_cf_handle_, ns_key, bytes);
  }
```

File: src/types/redis_string.cc (hunk 3 above)
Comment (enjoy-binbin) [reply src/types/redis_string.cc]: thanks.

File: src/types/redis_string.cc (hunk 4 above)
Comment (mapleFU) [reply src/types/redis_string.cc:376]: Could it be `move` here?

`ns_key` is built and used instantly, seems that it would not keep it

Comment (enjoy-binbin) [reply src/types/redis_string.cc:376]: can you describe it a bit more or insert a suggestion? 
i am not very familiar with this. (not a pure C++ coder)

Comment (mapleFU) [reply src/types/redis_string.cc:376]:
```c++
    for (const StringPair &pair : pairs) {
      std::string ns_key;
      AppendNamespacePrefix(pair.key, &ns_key);
    }
```

And 

```c++
  for (const auto &pair : pairs) {
    std::string bytes;
    metadata.expire = expire;
    metadata.Encode(&bytes);
    bytes.append(pair.value.data(), pair.value.size());
    std::string ns_key;
    std::optional<MultiLockGuard> guard;
  if (lock) {
    std::vector<std::string> lock_keys;
    lock_keys.reserve(pairs.size());
    for (const StringPair &pair : pairs) {
      AppendNamespacePrefix(pair.key, &ns_key);
      lock_keys.emplace_back(ns_key);
    }
    guard.emplace(storage_->GetLockManager(), lock_keys);
  }
```

I suggest this one. But it's ok to keep the current implemention

mapleFU (approved):

src/types/redis_string.cc
enjoy-binbin: i see in the second code block, the lock part is inside the for loop? is that a copy-paste error?

mapleFU: Yeah just an example..

enjoy-binbin: ohh, ok i got your point (finally!), thanks for the review.

git-hulk (approved):

PR #1652, opened 2023-08-08
Fix add/delete repeated members error in zset/set/hash types

### LINKED ISSUE #1651: Add/delete repeated members will result in incorrect metadata in zset/set/hash types
### Search before asking

### Version

OS: macOS 12.1
kvrocks branch: unstable
commitID: 60bcc3fac32d0df45e9a06ca0616d9d8d36fa579

### Minimal reproduce step

1. Execute `sadd` with multiple members including repetition. 
`sadd skey m1 m1 m2 m3`
2. Execute `srem` on the same way. 
`srem skey m1 m1`
3. Check the result of `sadd/srem` and check metadata of key `skey` with `scard`. 
`scard skey`

The result that execute `sadd` on redis
image

The result that execute `srem` on redis
image

### What did you see instead?

Add members with repetition into a set key
image
image

Remove members with repetition
image

As shown above, the repetitions members can't be filtered in `sadd/srem` commands, which will result in wrong metadata.

### Anything Else?

This bug is also existing in `zrem/hset/hmset/hdel`.
Using `zrem` as an example, if `zrem` is executed with repeated members, `zcard/zcount` will get inconsistent results.
image

ChrisZMF:
Closes #1651 
Fix that zrem/sadd/srem/hset/hmset/hdel commands get incorrect results when repeated members are passed.
Use `sadd/srem` as example
```
127.0.0.1:6666> sadd skey m1 m1 m2 m3
(integer) 4
127.0.0.1:6666> scard skey
(integer) 4
127.0.0.1:6666> srem skey m1 m1
(integer) 2
127.0.0.1:6666> scard skey
(integer) 2
```
The right results should be
```
127.0.0.1:6666> sadd skey m1 m1 m2 m3
(integer) 3
127.0.0.1:6666>  scard skey
(integer) 3
127.0.0.1:6666> srem skey m1 m1
(integer) 1
127.0.0.1:6666> scard skey
(integer) 2
```
This bug will result in incorrect metadata of  zset/set/hash keys, consequently leading to wrong results while accessing keys related to the corresponding types.  More details in #1651

### DIFF
--- a/src/types/redis_hash.cc
+++ b/src/types/redis_hash.cc
@@ -217,7 +217,11 @@ rocksdb::Status Hash::Delete(const Slice &user_key, const std::vector<Slice> &fi
   if (!s.ok()) return s.IsNotFound() ? rocksdb::Status::OK() : s;
 
   std::string sub_key, value;
+  std::unordered_set<std::string> field_set;
   for (const auto &field : fields) {
+    if (!field_set.emplace(field.ToString()).second) {
+      continue;
+    }
     InternalKey(ns_key, field, metadata.version, storage_->IsSlotIdEncoded()).Encode(&sub_key);
     s = storage_->Get(rocksdb::ReadOptions(), sub_key, &value);
     if (s.ok()) {
@@ -250,10 +254,14 @@ rocksdb::Status Hash::MSet(const Slice &user_key, const std::vector<FieldValue> 
   auto batch = storage_->GetWriteBatchBase();
   WriteBatchLogData log_data(kRedisHash);
   batch->PutLogData(log_data.Encode());
+  std::unordered_set<std::string> field_set;
+  for (int32_t i = static_cast<int32_t>(field_values.size() - 1); i >= 0; i--) {
+    const auto& fv = field_values[i];
+    if (!field_set.insert(fv.field).second) {
+      continue;
+    }
 
-  for (const auto &fv : field_values) {
     bool exists = false;
-
     std::string sub_key;
     InternalKey(ns_key, fv.field, metadata.version, storage_->IsSlotIdEncoded()).Encode(&sub_key);
 
--- a/src/types/redis_hash.cc
+++ b/src/types/redis_hash.cc
@@ -217,7 +217,7 @@ rocksdb::Status Hash::Delete(const Slice &user_key, const std::vector<Slice> &fi
   if (!s.ok()) return s.IsNotFound() ? rocksdb::Status::OK() : s;
 
   std::string sub_key, value;
-  std::unordered_set<std::string> field_set;
+  std::unordered_set<std::string_view> field_set;
   for (const auto &field : fields) {
     if (!field_set.emplace(field.ToString()).second) {
       continue;
@@ -254,7 +254,7 @@ rocksdb::Status Hash::MSet(const Slice &user_key, const std::vector<FieldValue> 
   auto batch = storage_->GetWriteBatchBase();
   WriteBatchLogData log_data(kRedisHash);
   batch->PutLogData(log_data.Encode());
-  std::unordered_set<std::string> field_set;
+  std::unordered_set<std::string_view> field_set;
   for (auto it = field_values.rbegin(); it != field_values.rend(); it++) {
     if (!field_set.insert(it->field).second) {
       continue;
--- a/src/types/redis_hash.cc
+++ b/src/types/redis_hash.cc
@@ -255,15 +255,14 @@ rocksdb::Status Hash::MSet(const Slice &user_key, const std::vector<FieldValue> 
   WriteBatchLogData log_data(kRedisHash);
   batch->PutLogData(log_data.Encode());
   std::unordered_set<std::string> field_set;
-  for (int32_t i = static_cast<int32_t>(field_values.size() - 1); i >= 0; i--) {
-    const auto& fv = field_values[i];
-    if (!field_set.insert(fv.field).second) {
+  for (auto it = field_values.rbegin(); it != field_values.rend(); it++) {
+    if (!field_set.insert(it->field).second) {
       continue;
     }
 
     bool exists = false;
     std::string sub_key;
-    InternalKey(ns_key, fv.field, metadata.version, storage_->IsSlotIdEncoded()).Encode(&sub_key);
+    InternalKey(ns_key, it->field, metadata.version, storage_->IsSlotIdEncoded()).Encode(&sub_key);
 
     if (metadata.size > 0) {
       std::string field_value;
@@ -271,7 +270,7 @@ rocksdb::Status Hash::MSet(const Slice &user_key, const std::vector<FieldValue> 
       if (!s.ok() && !s.IsNotFound()) return s;
 
       if (s.ok()) {
-        if (nx || field_value == fv.value) continue;
+        if (nx || field_value == it->value) continue;
 
         exists = true;
       }
@@ -279,7 +278,7 @@ rocksdb::Status Hash::MSet(const Slice &user_key, const std::vector<FieldValue> 
 
     if (!exists) added++;
 
-    batch->Put(sub_key, fv.value);
+    batch->Put(sub_key, it->value);
   }
 
   if (added > 0) {
--- a/src/types/redis_hash.cc
+++ b/src/types/redis_hash.cc
@@ -219,7 +219,7 @@ rocksdb::Status Hash::Delete(const Slice &user_key, const std::vector<Slice> &fi
   std::string sub_key, value;
   std::unordered_set<std::string_view> field_set;
   for (const auto &field : fields) {
-    if (!field_set.emplace(field.ToString()).second) {
+    if (!field_set.emplace(field.ToStringView()).second) {
       continue;
     }
     InternalKey(ns_key, field, metadata.version, storage_->IsSlotIdEncoded()).Encode(&sub_key);
--- a/src/types/redis_set.cc
+++ b/src/types/redis_set.cc
@@ -70,7 +70,11 @@ rocksdb::Status Set::Add(const Slice &user_key, const std::vector<Slice> &member
   WriteBatchLogData log_data(kRedisSet);
   batch->PutLogData(log_data.Encode());
   std::string sub_key;
+  std::unordered_set<std::string> mset;
   for (const auto &member : members) {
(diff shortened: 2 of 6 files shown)


git-hulk:
@ChrisZMF Very happy to see you back again!

src/types/redis_set.cc
```diff
@@ -70,7 +70,11 @@ rocksdb::Status Set::Add(const Slice &user_key, const std::vector<Slice> &member
   WriteBatchLogData log_data(kRedisSet);
   batch->PutLogData(log_data.Encode());
   std::string sub_key;
+  std::unordered_set<std::string> mset;
```
mapleFU: Would `set` or `unordered_set` for `Slice` be ok here?

mapleFU (commented):
So, this is because the input is first input in batch, and then, write to store. Should we insert to set only when the key is not existed, and should be insert to batch?

src/types/redis_hash.cc
```diff
@@ -250,10 +254,14 @@ rocksdb::Status Hash::MSet(const Slice &user_key, const std::vector<FieldValue>
   auto batch = storage_->GetWriteBatchBase();
   WriteBatchLogData log_data(kRedisHash);
   batch->PutLogData(log_data.Encode());
+  std::unordered_set<std::string> field_set;
+  for (int32_t i = static_cast<int32_t>(field_values.size() - 1); i >= 0; i--) {
```
torwig: Could we use `size_t` instead of `int32_t` here and omit the `static_cast` usage?

git-hulk:
> So, this is because the input is first input in batch, and then, write to store. Should we insert to set only when the key is not existed, and should be insert to batch?

It may have a partial commit if the instance is broken in the middle of the operation.

src/types/redis_hash.cc
mapleFU: `std::unordered_set<std::string> field_set(field_values.begin(), field_values.end());`

I think this is ok.

enjoy-binbin:
IMO, if you already have a fix, there is no need to open an issue here.
Although there is already an issue associated with the PR, please also update the top comment (it can be copied from issue, pictures are not so convenient, those can actually be copied with block text)

Good catch btw, did not CR

src/types/redis_hash.cc
caipengbo: Can use `for (auto it = field_values.rbegin(); it != field_values.rend(); it++)` ?

src/types/redis_set.cc
ChrisZMF: `rocksdb::Slice` is not ok, it can't pass the compiling.

src/types/redis_hash.cc
ChrisZMF: `size_t` is not ok, it is unsigned type,  always `>= 0`.

`for (auto it = field_values.rbegin(); it != field_values.rend(); it++) ` could be a good choice.

ChrisZMF:
> IMO, if you already have a fix, there is no need to open an issue here. Although there is already an issue associated with the PR, please also update the top comment (it can be copied from issue, pictures are not so convenient, those can actually be copied with block text)
> 
> Good catch btw, did not CR

Issue is used to record and describe the details of the problem. It's more convenient to get and check it in the future.
Top comment should be updated, thx.

enjoy-binbin:
> Issue is used to record and describe the details of the problem. It's more convenient to get and check it in the future.

yean, i get your point.

PR also can record the details (in the top comment) and then merge them in as a squash commit message.
it (commit or PR or the details) also can be got and checked in the future.
In other words, a PR can actually do everything that a issue can do.

src/types/redis_set.cc
mapleFU: Yeah, seems it doesn't implement `std::hash`. Lets leave it for future optimizations

mapleFU:
This code LGTM. Would you mind format the code using `./x.py format`? Seems that CI failed here.

src/types/redis_hash.cc
```diff
@@ -217,7 +217,11 @@ rocksdb::Status Hash::Delete(const Slice &user_key, const std::vector<Slice> &fi
   if (!s.ok()) return s.IsNotFound() ? rocksdb::Status::OK() : s;
 
   std::string sub_key, value;
+  std::unordered_set<std::string> field_set;
```
PragmaTwice: Why do we store std::string here rather than Slice? I think it is unnecessary to store a copy of string.

src/types/redis_hash.cc
```diff
@@ -250,28 +254,31 @@ rocksdb::Status Hash::MSet(const Slice &user_key, const std::vector<FieldValue>
   auto batch = storage_->GetWriteBatchBase();
   WriteBatchLogData log_data(kRedisHash);
   batch->PutLogData(log_data.Encode());
+  std::unordered_set<std::string> field_set;
+  for (auto it = field_values.rbegin(); it != field_values.rend(); it++) {
```
PragmaTwice, line 258: Any reason to iterator the set in reverse order?

ChrisZMF:
> > Issue is used to record and describe the details of the problem. It's more convenient to get and check it in the future.
> 
> yean, i get your point.
> 
> PR also can record the details (in the top comment) and then merge them in as a squash commit message. it (commit or PR or the details) also can be got and checked in the future. In other words, a PR can actually do everything that a issue can do.

ChrisZMF:
> > Issue is used to record and describe the details of the problem. It's more convenient to get and check it in the future.
> 
> yean, i get your point.
> 
> PR also can record the details (in the top comment) and then merge them in as a squash commit message. it (commit or PR or the details) also can be got and checked in the future. In other words, a PR can actually do everything that a issue can do.

OK

src/types/redis_hash.cc
ChrisZMF: Slice is not supported to use as type in std::unordered_set or std::set. Any suggestions to solve it?

src/types/redis_hash.cc
ChrisZMF: If duplicates appear,  we have to use the last one of same member. Like in `hset hkey f1 v1 f1 v11`, 'f1 v11' should be kept.

src/types/redis_hash.cc
mapleFU: Personally I think `std::set` might works. `std::unordered_set` requires a `std::hash` or self-defined hash function.

PragmaTwice: You can use std::string_view instead of Slice.

src/types/redis_hash.cc
PragmaTwice: Thanks. Got it.

src/types/redis_hash.cc
ChrisZMF: std::set is not ok either. std::string_view is a good choice, thx

caipengbo (dismissed):

src/types/redis_set.cc
```diff
@@ -70,7 +70,11 @@ rocksdb::Status Set::Add(const Slice &user_key, const std::vector<Slice> &member
   WriteBatchLogData log_data(kRedisSet);
   batch->PutLogData(log_data.Encode());
   std::string sub_key;
+  std::unordered_set<std::string_view> mset;
   for (const auto &member : members) {
+    if (!mset.insert(member.ToString()).second) {
```
mapleFU: better remove `ToString()`, `rocksdb::Slice` might has `ToStringView()`, or you can use `emplace({member.data(), member.size()})`

src/types/redis_set.cc
```diff
@@ -101,7 +105,11 @@ rocksdb::Status Set::Remove(const Slice &user_key, const std::vector<Slice> &mem
   auto batch = storage_->GetWriteBatchBase();
   WriteBatchLogData log_data(kRedisSet);
   batch->PutLogData(log_data.Encode());
+  std::unordered_set<std::string_view> mset;
   for (const auto &member : members) {
+    if (!mset.insert(member.ToString()).second) {
```
mapleFU: ditto

src/types/redis_zset.cc
```diff
@@ -545,7 +543,11 @@ rocksdb::Status ZSet::Remove(const Slice &user_key, const std::vector<Slice> &me
   batch->PutLogData(log_data.Encode());
   int removed = 0;
   std::string member_key, score_key;
+  std::unordered_set<std::string_view> mset;
   for (const auto &member : members) {
+    if (!mset.insert(member.ToString()).second) {
```
mapleFU: ditto

mapleFU (commented):
Rest LGTM, thanks!

src/types/redis_set.cc
PragmaTwice: Good catch. ToString here will result in dangling reference (you cannot first construct a temporary string and then store its reference). So I think it is not "better to remove" but "need to remove".

Dangling reference is UB and may cause program to crash. And I think that is why the CI failed.

src/types/redis_zset.cc
```diff
@@ -52,24 +52,12 @@ rocksdb::Status ZSet::Add(const Slice &user_key, ZAddFlags flags, MemberScores *
...
-    // column family many times and cause problems in the ZRANGE command.
-    //
-    // For example, we add members with `ZADD mykey 1 a 2 a` and `ZRANGE mykey 0 1`
-    // return only one member(`a`) was expected but got the member `a` twice now.
-    //
-    // The root cause of this issue was the score key was composed by member and score,
-    // so the last one can't overwrite the previous when the score was different.
-    // A simple workaround was add those members with reversed order and skip the member if has added.
-    if (added_member_keys.find(member_key) != added_member_keys.end()) {
+  std::unordered_set<std::string_view> added_member_keys;
```
mapleFU, line 55: So previously, `added_member_keys` is a "decoded" key, now, it becomes a "user-key"?

mapleFU (approved):
Thanks!

torwig (approved):

PragmaTwice (approved):

src/types/redis_zset.cc
ChrisZMF: Previously, the repeated members are filtered by encoded member_key, but using the original member_key can achieve the same effect.

ChrisZMF:
Thanks @mapleFU @PragmaTwice

PR #1655, opened 2023-08-09
Add dynamic flag generator to replace adhoc logic in ExecuteCommands

PragmaTwice:
We add dynamic flag generator regarding to command arguments to replace adhoc exclusive check logic in Connection::ExecuteCommands().

And we also fix and refactor GetKeysFromCommand().

### DIFF
--- a/src/commands/commander.cc
+++ b/src/commands/commander.cc
@@ -46,7 +46,7 @@ std::string GetCommandInfo(const CommandAttributes *command_attributes) {
   command.append(redis::BulkString(command_attributes->name));
   command.append(redis::Integer(command_attributes->arity));
   command_flags.append(redis::MultiLen(1));
-  command_flags.append(redis::BulkString(command_attributes->IsWrite() ? "write" : "readonly"));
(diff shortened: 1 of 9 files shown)


src/commands/commander.cc
```diff
@@ -77,30 +77,55 @@ void GetCommandsInfo(std::string *info, const std::vector<std::string> &cmd_name
...
   }
 
-  auto last = command_attribute->key_range.last_key;
-  if (last < 0) last = argc + last;
+  DumpKeyRange(*keys_index, argc, attributes->key_range);
+
+  return Status::OK();
+}
+
+Status GetKeysFromCommand(const CommandAttributes *attributes, const std::vector<std::string> &cmd_tokens,
```
git-hulk: Looks like `command getkeys` can also use this to get keys from the command.

PragmaTwice: there is no cmd_tokens here

PragmaTwice: done

git-hulk (approved):

PR #1657, opened 2023-08-09
Add new MPUBLISH command

torwig:
The new `MPUBLISH` command allows publishing one or more messages to a channel.
Syntax: `MPUBLISH channel-name message1 message2 ... messageN`

### DIFF
--- a/src/commands/cmd_pubsub.cc
+++ b/src/commands/cmd_pubsub.cc
@@ -27,12 +27,13 @@ namespace redis {
 
 class CommandPublish : public Commander {
  public:
-  // mark is_write as false here because slave should be able to execute publish command
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     if (!svr->IsSlave()) {
-      // Compromise: can't replicate message to sub-replicas in a cascading-like structure.
-      // Replication relies on WAL seq, increase the seq on slave will break the replication, hence the compromise
+      // Compromise: can't replicate a message to sub-replicas in a cascading-like structure.
+      // Replication relies on WAL seq; increasing the seq on a replica will break the replication process,
+      // hence the compromise solution
       redis::PubSub pubsub_db(svr->storage);
+
       auto s = pubsub_db.Publish(args_[1], args_[2]);
       if (!s.ok()) {
         return {Status::RedisExecErr, s.ToString()};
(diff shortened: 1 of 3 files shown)


enjoy-binbin (commented):
I'll add some refs first.
redis have this pending issue: [[QUESTION] Why not allow PUBLISH to publish multiple messages?](https://github.com/redis/redis/issues/12213)
and this pending PR: [How MPUBLISH could work #12267](https://github.com/redis/redis/pull/12267)

torwig:
@enjoy-binbin Thank you for the links.

torwig:
I guess I should consider introducing a new command for this instead of changing the `PUBLISH` command.

caipengbo:
> I guess I should consider introducing a new command for this instead of changing the PUBLISH command.

Agree with it

enjoy-binbin:
Do you already want to use this feature? One case is we can pipeline it in this monment (Waiting for Redis to actually implement it if not in a hurry).

or implementing `MPUBLISH channel message1 message2 ...`. This should be possible since it should be implemented in Redis 8 (i see two Redis core-team member agree with it)

aleksraiden:
Thanks, @enjoy-binbin @torwig 

Yes, variant `MPUBLISH channel message1 message2` will be awesome

torwig:
Implemented the feature as a separate command `MPUBLISH`.

PragmaTwice (approved):

src/commands/cmd_pubsub.cc
```diff
@@ -27,20 +27,48 @@ namespace redis {
...
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    int total_receivers = 0;
+
+    for (size_t i = 2; i < args_.size(); i++) {
+      if (!svr->IsSlave()) {
+        redis::PubSub pubsub_db(svr->storage);
+
+        auto s = pubsub_db.Publish(args_[1], args_[i]);
+        if (!s.ok()) {
```
mapleFU, line 61: Seems that `Publish` will be likely to success?

src/commands/cmd_pubsub.cc
```diff
@@ -27,20 +27,48 @@ namespace redis {
...
+          return {Status::RedisExecErr, s.ToString()};
+        }
+      }
+
+      int receivers = svr->PublishMessage(args_[1], args_[i]);
+      total_receivers += receivers;
+    }
+
+    *output = redis::Integer(total_receivers);
```
mapleFU, line 70: Would an array of int be better here?

torwig: @mapleFU Possibly, yes. However, according to this PR (https://github.com/redis/redis/pull/12267) it's likely that `MPUBLISH` will have a single integer as a response.

caipengbo (approved):
LGTM

src/commands/cmd_pubsub.cc
mapleFU: Oh this LGTM. Seems this is not stable. Maybe we can mark "mpublish" as "experimental" or something, and make it able to change the syntax of it.

torwig: In general, I think it's not easy to utilize the return value of the `PUBLISH` command, except to check whether it is 0 or more. The actual goal of Pub/Sub pattern is to decouple publishers from subscribers and if I want to check the number of receivers I should know the exact number of them. And for example, if I know that there should be 5 subscribers but I received only 4 in response to the `PUBLISH` command, what should I do?

mapleFU (approved):
This patch general LGTM. I've some tiny problems.

Seems `pubsub_db.Publish` different messages separately. It's ok if all write works well. However, could user find how many message is successed when `Publish` failed?

Another is that this calls a look for each `arg`. Could we make `mpublish` faster, and batch the notifying later? Would this save some time?

src/commands/cmd_pubsub.cc
mapleFU: Sounds ok. It's ok to leave it like this

git-hulk (approved):

PR #1591, opened 2023-07-13
Fix version typo in variable name

enjoy-binbin:

PR #1660, opened 2023-08-10
Add a spell check ci that will fail future PRs

enjoy-binbin:
This PR adds a spell checker CI action that will fail
future PRs if they introduce typos and spelling mistakes.

This spell checker is based on blacklist of common spelling
mistakes, so it will not catch everything, but at least it
is also unlikely to cause false positives.

closing in favor of #1663


enjoy-binbin:
I am not sure whether the apache license is in conflict with this (https://github.com/codespell-project/codespell), please help me to check.
If conflict, i will remove the CI and refresh the PR with only the typo fix.

here is am example: https://github.com/apache/kvrocks/actions/runs/5818954638/job/15776396946?pr=1660#step:5:7
```
Used config files:
    1: ./.codespell/.codespellrc
./kvrocks.conf:569: Leve ==> Level, Levee
```

more info:
```
# A CI action that using codespell to check spell.
# .codespell/.codespellrc is a config file.
# .codespell/wordlist.txt is a list of words that will ignore word checks.
# More details please check the following link:
# https://github.com/codespell-project/codespell
```

used to mention in https://github.com/apache/kvrocks/pull/1591#issuecomment-1633629161

src/types/redis_geo.cc
```diff
@@ -320,7 +320,7 @@ void Geo::scoresOfGeoHashBox(GeoHashBits hash, GeoHashFix52Bits *min, GeoHashFix
  * 'max', appending them into the array of GeoPoint structures 'gparray'.
  * The command returns the number of elements added to the array.
  *
- * Elements which are farest than 'radius' from the specified 'x' and 'y'
+ * Elements which are farther than 'radius' from the specified 'x' and 'y'
```
PragmaTwice, line 323:
```suggestion
 * Elements which are farthest than 'radius' from the specified 'x' and 'y'
```

enjoy-binbin: since there is a `than`, i think it should be farther? (although i am not sure about it)

PragmaTwice:
> I am not sure whether the apache license is in conflict with this (https://github.com/codespell-project/codespell), please help me to check.

It is OK to use a GPL licensed project to do spell check, as we also use GCC in CI to build kvrocks : )

tisonkun:
I'd suggest we use https://github.com/crate-ci/typos which is easy to install and also licensed under MIT OR Apache-2.0

enjoy-binbin:
> It is OK to use a GPL licensed project to do spell check, as we also use GCC in CI to build kvrocks : )

good to know, thanks for the confirmation!

.codespell/.codespellrc
```diff
@@ -0,0 +1,4 @@
+[codespell]
```
PragmaTwice, line 1: I think we can move these files to `.github/config`.

enjoy-binbin:
closing in favor of #1663

Issue #1662, opened 2023-08-10
CMake compiler check random failure in GCC TSAN CI


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

https://github.com/apache/kvrocks/actions/runs/5819255897/job/15782518129?pr=1661

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PragmaTwice:
It is weird, since the compiler check procedure is not related to the kvrocks code.

PR #1663, opened 2023-08-10
ci: add typos check 

tisonkun:
This PR adds a typo checker CI action that will fail
future PRs if they introduce typos and spelling mistakes.

This supersedes and thus closes #1660.

PR #1668, opened 2023-08-11
Tentatively comment out GCC TSAN in CI to avoid blocking PRs

git-hulk:
Currently, GCC TSAN was broken but didn't identify the underlying reason,
so let's comment it out first to avoid blocking new PRs and add it back
after it was fixed.

#1662 is tracking the GCC TSAN issue.

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -134,12 +134,6 @@ jobs:
             with_sanitizer: -DENABLE_ASAN=ON
             without_jemalloc: -DDISABLE_JEMALLOC=ON
             compiler: clang
-          - name: Ubuntu GCC TSan
-            os: ubuntu-20.04
-            without_jemalloc: -DDISABLE_JEMALLOC=ON
-            with_sanitizer: -DENABLE_TSAN=ON
-            compiler: gcc
-            ignore_when_tsan: -tags="ignore_when_tsan"
           - name: Ubuntu Clang TSan
             os: ubuntu-20.04
             with_sanitizer: -DENABLE_TSAN=ON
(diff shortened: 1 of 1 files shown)


git-hulk:
#1663 has removed this unit, close this PR.

enjoy-binbin:
The topic of #1663 is not related to this, i think it can be kept

enjoy-binbin (dismissed):

tisonkun:
We should comment out and add an inline comment. Even we should open an issue to dig out, instead of remove the matrix item.

tisonkun:
This is the related issue - https://github.com/apache/kvrocks/issues/1662

enjoy-binbin:
why bring this into that typo ci? we can merge this one first and then refresh the typo ci one

tisonkun:
@enjoy-binbin you're right.

.github/workflows/kvrocks.yaml
```diff
@@ -134,12 +134,6 @@ jobs:
             with_sanitizer: -DENABLE_ASAN=ON
             without_jemalloc: -DDISABLE_JEMALLOC=ON
             compiler: clang
```
tisonkun, line 136:
```suggestion
            compiler: clang
          # FIXME - tsan.so not found https://github.com/apache/kvrocks/issues/1662
          # - name: Ubuntu GCC TSan
          #   os: ubuntu-20.04
          #   without_jemalloc: -DDISABLE_JEMALLOC=ON
          #   with_sanitizer: -DENABLE_TSAN=ON
          #   compiler: gcc
          #   ignore_when_tsan: -tags="ignore_when_tsan"
```

tisonkun (approved):

enjoy-binbin (approved):

PR #1669, opened 2023-08-11
Fix data dace when closing the server close in Go tests

### LINKED ISSUE #1647: Flaky test in string_test.go
### Search before asking

### Version

unstable

### Minimal reproduce step

None

go tests passed

### What did you see instead?

https://github.com/apache/kvrocks/actions/runs/5779002383/job/15660953839#step:12:114

```
--- FAIL: TestString (45.66s)
    server.go:109: 
        	Error Trace:	/Users/runner/work/kvrocks/kvrocks/tests/gocase/unit/type/strings/server.go:109
        	            				/Users/runner/work/kvrocks/kvrocks/tests/gocase/unit/type/strings/server.go:112
        	            				/Users/runner/work/kvrocks/kvrocks/tests/gocase/unit/type/strings/server.go:101
        	            				/Users/runner/work/kvrocks/kvrocks/tests/gocase/unit/type/strings/strings_test.go:757
        	Error:      	An error is expected but got nil.
        	Test:       	TestString
FAIL
exit status 1
FAIL	github.com/apache/kvrocks/tests/gocase/unit/type/strings	45.811s
```

### Anything Else?

_No response_

git-hulk:
Currently, the Kvrocks server close will wait for the process to be exited or until the grace period(30s). And there's a data race between the close function and the timer expired function. We need to wait for the timer's expired routine if it's fired.

This closes #1647

### DIFF
--- a/tests/gocase/util/server.go
+++ b/tests/gocase/util/server.go
@@ -28,6 +28,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"sync"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


tisonkun (approved):

mapleFU (commented):
So it's because the syntax of `Stop` ? ( https://pkg.go.dev/time#Timer.Stop )

What would happen if `Kill` never starts?

git-hulk:
> So it's because the syntax of `Stop` ? ( https://pkg.go.dev/time#Timer.Stop )
> 
> What would happen if `Kill` never starts?

There're two issues:

1. data race between the close routine and the time-expired routine, the close routine was finished but the timer function was fired, then will expect the kill operation succeeded.
2. The time-expired routine may be panic if the main routine didn't wait for it, refer to the comment: https://github.com/apache/kvrocks/issues/1647#issuecomment-1670595129


> What would happen if `Kill` never starts?

For the most of time, the kill operation shouldn't be started.

mapleFU:
Oh I get it. `Kill` is just for timeout handling. And most of time it will not happen.

When it start to `Kill`, we expect a graceful shutdown?

tisonkun:
> When it start to Kill, we expect a graceful shutdown?

No. It's a final guard to ensure we do not leak

mapleFU:
Oh got it! Thanks!

PR #1671, opened 2023-08-12
Return string instead of output pointer in metadata

PragmaTwice:
```diff
-void ExtractNamespaceKey(Slice ns_key, std::string *ns, std::string *key, bool slot_id_encoded);
-void ComposeNamespaceKey(const Slice &ns, const Slice &key, std::string *ns_key, bool slot_id_encoded);
-void ComposeSlotKeyPrefix(const Slice &ns, int slotid, std::string *output);
+[[nodiscard]] std::tuple<Slice, Slice> ExtractNamespaceKey(Slice ns_key, bool slot_id_encoded);
+[[nodiscard]] std::string ComposeNamespaceKey(const Slice &ns, const Slice &key, bool slot_id_encoded);
+[[nodiscard]] std::string ComposeSlotKeyPrefix(const Slice &ns, int slotid);
 
 class InternalKey {
  public:
   ...
-  void Encode(std::string *out);
+  [[nodiscard]] std::string Encode() const;
```

### DIFF
--- a/src/storage/redis_metadata.cc
+++ b/src/storage/redis_metadata.cc
@@ -109,10 +109,11 @@ bool InternalKey::operator==(const InternalKey &that) const {
   return version_ == that.version_;
 }
 
-std::tuple<Slice, Slice> ExtractNamespaceKey(Slice ns_key, bool slot_id_encoded) {
+template <typename T>
+std::tuple<T, T> ExtractNamespaceKey(Slice ns_key, bool slot_id_encoded) {
   uint8_t namespace_size = 0;
   GetFixed8(&ns_key, &namespace_size);
-  Slice ns(ns_key.data(), namespace_size);
+  T ns(ns_key.data(), namespace_size);
   ns_key.remove_prefix(namespace_size);
 
   if (slot_id_encoded) {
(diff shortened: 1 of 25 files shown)


src/storage/redis_metadata.cc
```diff
@@ -110,41 +109,46 @@ bool InternalKey::operator==(const InternalKey &that) const {
   return version_ == that.version_;
 }
 
-void ExtractNamespaceKey(Slice ns_key, std::string *ns, std::string *key, bool slot_id_encoded) {
+std::tuple<Slice, Slice> ExtractNamespaceKey(Slice ns_key, bool slot_id_encoded) {
   uint8_t namespace_size = 0;
   GetFixed8(&ns_key, &namespace_size);
-  *ns = ns_key.ToString().substr(0, namespace_size);
+  Slice ns(ns_key.data(), namespace_size);
```
mapleFU: So this is an optimization that return Slice rather than string?

mapleFU (dismissed):
The code LGTM, lets wait for CI and asan, since it might introduce memory problem...

src/storage/redis_metadata.cc
PragmaTwice: Yeah, here it is unnecessary to return string. And if it is required, users can construct string manually from slice.

torwig (approved):
LGTM.

:+1:

mapleFU (approved):
LGTM, though it increase the risk of lifetine handling, the patch would improve our style(and make some string copy less)

PR #1673, opened 2023-08-14
Rename cluster-enable in kvrocks2redis to cluster-enabled

enjoy-binbin:
Rename cluster-enable to cluster-enabled for keeping consistent
with Kvrocks, we keeping the old one for compatibility.


git-hulk (approved):
LGTM, thanks for your quick PR.

PR #1683, opened 2023-08-18
Refactor encoding via templates

PragmaTwice:
We refactor encoding utils via templates in this PR.
All existing interfaces remain unchanged, and some missing and generic interfaces are added.


git-hulk (approved):
LGTM, cool

PR #1684, opened 2023-08-19
Add the support of the restore command

### LINKED ISSUE #409: [NEW]Please support restore command
Migration tools like redisshake usually use restore command to realize migration from redis to kvrocks。
Please support restore command。

### LINKED ISSUE #1648: Add the support of the RESTORE command
### Search before asking

### Motivation

This issue considers to support the [DUMP](https://redis.io/commands/dump/) / [RESTORE](https://redis.io/commands/restore/) command.

### Solution

_No response_

git-hulk:
This PR will close #1648 which supports the restore command for making Redis migration tools happy. To achieve this feature, we need to implement all encoding types now supported by Redis from 2.x to 7.x which includes:

- All kinds of data objects like String/List/Set/Hash/Sorted Set
- Binary encoding types: ZipMap/ZipList/IntSet/ListPack

For test cases, I have dumped value for Redis according to encoding types and confirmed it works well with the RedisShake migration tool. Another thing to be noticed, we haven't supported the stream and module type yet, because our stream type is partially implemented, so we can go back to support it after all commands are supported.


JoverZhang:
Maybe I can do the "Add the support of the set type". Let me try it.

src/storage/rdb.cc
```diff
@@ -0,0 +1,320 @@
...
+
+  Status peekOK(size_t n) {
+    if (pos_ + n > input_.size()) {
+      return {Status::NotOK, "invalid list pack entry"};
+    }
+    return Status::OK();
+  }
+
+  StatusOr<std::string> next() {
+    RET_IF_ERROR(peekOK(1));
```
PragmaTwice: You can just use GET_OR_RET, which is supported to return void.

git-hulk: OK, thanks for your warm reminder, I think it must return a value.

git-hulk:
PR is ready to be reviewed, but I need some time to add test cases for all encoding types.

src/server/server.cc
```diff
@@ -63,6 +64,7 @@ Server::Server(engine::Storage *storage, Config *config)
 
   // init cursor_dict_
   cursor_dict_ = std::make_unique<CursorDictType>();
+  crc64_init();
```
PragmaTwice: Since it is not related to Server, I think it can be put into the main function.

src/storage/rdb.h
```diff
@@ -0,0 +1,100 @@
...
+constexpr const int RDBTypeSetListPack = 20;
+
+// Quick list node encoding
+constexpr const int QuickListNodeContainerPlain = 1;
+constexpr const int QuickListNodeContainerPacked = 2;
+
+class RDB {
+ public:
+  explicit RDB(engine::Storage *storage, std::string ns, const std::string &input)
+      : storage_(storage), ns_(std::move(ns)), input_(input){};
```
PragmaTwice:
```suggestion
  explicit RDB(engine::Storage *storage, std::string ns, std::string_view input)
      : storage_(storage), ns_(std::move(ns)), input_(input){};
```

src/vendor/lzf.h
```diff
@@ -0,0 +1,201 @@
...
+ * not being able to store offset above UINT32_MAX in 64bit. */
+#define LZF_USE_OFFSETS 0
+
+/*****************************************************************************/
+/* nothing should be changed below */
+
+#ifdef __cplusplus
+#include <climits>
+#include <cstring>
+using namespace std;
```
PragmaTwice: IMHO it is a very bad behavior to import std namepace to the global scope in a header file.

I think we need to change it, even it is from a dependency.

src/storage/rdb.cc
```diff
@@ -0,0 +1,444 @@
...
+  auto compression_len = GET_OR_RET(loadObjectLen(nullptr));
+  auto len = GET_OR_RET(loadObjectLen(nullptr));
+  GET_OR_RET(peekOk(static_cast<size_t>(compression_len)));
+
+  auto out_buf = make_unique<char[]>(len);
+  if (lzf_decompress(input_.data() + pos_, compression_len, out_buf.get(), len) != len) {
+    return {Status::NotOK, "Invalid LZF compressed string"};
+  }
+  pos_ += compression_len;
+  return std::string().assign(out_buf.get(), len);
```
PragmaTwice:
```suggestion
  std::string out_buf(len, 0);
  if (lzf_decompress(input_.data() + pos_, compression_len, out_buf.data(), len) != len) {
    return {Status::NotOK, "Invalid LZF compressed string"};
  }
  pos_ += compression_len;
  return out_buf;
```

src/vendor/lzf.h
git-hulk: It should be falsely imported by IDE when copying since C would never import this.

src/storage/rdb.cc
```diff
@@ -0,0 +1,444 @@
...
+  char buf[256];
+  GET_OR_RET(peekOk(1));
+  auto len = static_cast<uint8_t>(input_[pos_++]);
+  switch (len) {
+    case 255:
+      return -1.0 / 0.0; /* Negative inf */
+    case 254:
+      return 1.0 / 0.0; /* Positive inf */
+    case 253:
+      return 0.0 / 0.0; /* NaN */
```
PragmaTwice: Use macro `NAN`, `INFINITY` or `std::numeric_limits`?

src/storage/rdb_intset.cc
```diff
@@ -0,0 +1,86 @@
...
+        break;
+      }
+      case IntSetEncInt64: {
+        uint64_t v = 0;
+        memcpy(&v, input_.data() + pos_, sizeof(uint64_t));
+        pos_ += sizeof(uint64_t);
+        intrev64ifbe(&v);
+        entries.emplace_back(std::to_string(v));
+        break;
+      }
```
PragmaTwice, line 82: add a default branch?

PragmaTwice:
For compatibility testing, I think we can start a redis service in CI and do some mutual en/decoding? e.g. `KvrocksDecode(RedisEncode(x)) = x` or `RedisDecode(KvrocksEncode(y)) = y`

git-hulk:
> For compatibility testing, I think we can start a redis service in CI and do some mutual en/decoding? e.g. `KvrocksDecode(RedisEncode(x)) = x` or `RedisDecode(KvrocksEncode(y)) = y`

Aha, my test case values are dumped from the Redis, it should be difficult since those encoding types are located in many Redis versions, for example, the zip map exists in Redis 2.x.

git-hulk:
Sorry for the big PR but the good news is it only affects the restore command itself :)

torwig:
@git-hulk Yeah, a massive PR! But I believe we will handle it :)

git-hulk:
> @git-hulk Yeah, a massive PR! But I believe we will handle it :)

Thank you! That's a bit hard to review since most of them are Redis encoding-related implementations. All test values are dumped from the Redis from Redis 2.6(dump command was supported from this version), and now test cases can cover all sorts of encoding types to guarantee it works.

src/storage/rdb.h
```diff
@@ -0,0 +1,100 @@
...
+  std::string_view input_;
+  size_t pos_ = 0;
+
+  StatusOr<std::string> loadLzfString();
+  StatusOr<std::string> loadEncodedString();
+  StatusOr<uint64_t> loadObjectLen(bool *is_encoded);
+  Status peekOk(size_t n);
+  StatusOr<double> loadBinaryDouble();
+  StatusOr<double> loadDouble();
+};
```
PragmaTwice:
```suggestion
};

```

better to have a new line in the end, same as other files

src/storage/rdb.cc
```diff
@@ -0,0 +1,451 @@
...
+  if (!db_status.ok()) {
+    return {Status::RedisExecErr, db_status.ToString()};
+  }
+  // String type will use the SETEX, so just only set the ttl for other types
+  if (ttl_ms > 0 && type != RDBTypeString) {
+    redis::Database db(storage_, ns_);
+    db_status = db.Expire(key, ttl_ms + util::GetTimeStampMS());
+  }
+  return db_status.ok() ? Status::OK() : Status{Status::RedisExecErr, db_status.ToString()};
+}
```
PragmaTwice:
```suggestion
}

```

git-hulk:
I have tested the [RedisShake](https://github.com/tair-opensource/RedisShake) to migrate Redis data into Kvrocks and it works well now.

src/commands/cmd_server.cc
```diff
@@ -978,6 +980,70 @@ static uint64_t GenerateConfigFlag(const std::vector<std::string> &args) {
...
+    rocksdb::Status db_status;
+    redis::Database redis(svr->storage, conn->GetNamespace());
+    if (!replace_) {
+      int count = 0;
+      db_status = redis.Exists({args_[1]}, &count);
+      if (!db_status.ok()) {
+        return {Status::RedisExecErr, db_status.ToString()};
+      }
+      if (count > 0) {
+        return {Status::RedisExecErr, "target key name already exists."};
```
enjoy-binbin, line 1022: redis will throw a `BUSYKEY` error, ie. is actually a BUSYKEY error code. We will only return ERR here. but i guess it is ok, we need to find time to clean up and align some error codes
```
127.0.0.1:6379> restore busy_key 1111 "\x00\x01b\x0b\x00\x7f\xfb\x05\xe2[|\xfb\n" absttl
(error) BUSYKEY Target key name already exists.
```

src/commands/cmd_server.cc
```diff
@@ -978,6 +980,70 @@ static uint64_t GenerateConfigFlag(const std::vector<std::string> &args) {
...
+    ttl_ms_ = GET_OR_RET(ParseInt<uint64_t>(args[2], 10));
+    while (parser.Good()) {
+      if (parser.EatEqICase("replace")) {
+        replace_ = true;
+      } else if (parser.EatEqICase("absttl")) {
+        auto now = util::GetTimeStampMS();
+        if (ttl_ms_ <= static_cast<uint64_t>(now)) {
+          return {Status::RedisExecErr, "Invalid expire time"};
+        }
+        ttl_ms_ -= now;
```
enjoy-binbin: in this case, redis will implicitly expire keys instead of throwing an error.
```
127.0.0.1:6379> restore busy_key 1111 "\x00\x01b\x0b\x00\x7f\xfb\x05\xe2[|\xfb\n" absttl
OK
```

and if we want to throw the error, here i think we should use RedisParseErr.

src/commands/cmd_server.cc
```diff
@@ -978,6 +980,70 @@ static uint64_t GenerateConfigFlag(const std::vector<std::string> &args) {
...
+        return {Status::RedisExecErr, db_status.ToString()};
+      }
+      if (count > 0) {
+        return {Status::RedisExecErr, "target key name already exists."};
+      }
+    } else {
+      db_status = redis.Del(args_[1]);
+      if (!db_status.ok() && !db_status.IsNotFound()) {
+        return {Status::RedisExecErr, db_status.ToString()};
+      }
```
enjoy-binbin, line 1028: if key is not found, i think we should not return an error?
```
127.0.0.1:6379> del busy_key
(integer) 0
127.0.0.1:6379> restore busy_key 1111 "\x00\x01b\x0b\x00\x7f\xfb\x05\xe2[|\xfb\n" absttl replace
OK
```

enjoy-binbin (commented):
great jobs. i added some minor comments. 
did not review the RDB part since i'm not familiar with it, i am not able th give the advice.

src/storage/rdb_intset.cc
```diff
@@ -0,0 +1,88 @@
...
+      default:
+        return {Status::NotOK, "invalid intset encoding"};
+    }
+  }
+  return entries;
+}
```
PragmaTwice: There are also several other files without newline in the end. Maybe you can have a check?

git-hulk: ok, I will go through all new files.

src/commands/cmd_server.cc
git-hulk: For Redis, it will return ok if the key is not found even when the replace flag exists.

src/commands/cmd_server.cc
git-hulk: Yes, we need a dedicated PR to fix this.

src/commands/cmd_server.cc
git-hulk: Good catch, will fix it.

src/commands/cmd_server.cc
git-hulk: Yes, this code won't throw an error if the error is key not found.

PragmaTwice (approved):

git-hulk:
Thanks all, I will merge this PR and can submit an issue to followup if you found any issues.

PR #1687, opened 2023-08-21
Fix LPOS rank passing LLONG_MIN overflow issue

enjoy-binbin:
There is a minor overflow issue in RANK negation, passing LLONG_MIN
will overflow and is effectively be the same as passing -1.

This is the example before the fix:
```
127.0.0.1:6666> flushall
OK
127.0.0.1:6666> lpos mylist foo rank -9223372036854775808
(nil)
127.0.0.1:6666> lpush mylist foo foo foo foo foo
(integer) 5

127.0.0.1:6666> lpos mylist foo rank -1
(integer) 4
127.0.0.1:6666> lpos mylist foo rank -5
(integer) 0
127.0.0.1:6666> lpos mylist foo rank -6
(nil)
127.0.0.1:6666> lpos mylist foo rank -9223372036854775807
(nil)

-- this should return nil but it returned the last one because the overflow rank became -1
127.0.0.1:6666> lpos mylist foo rank -9223372036854775808
(integer) 4
```

Now we limit RANK to not be LLONG_MIN and will throw an error directly
(this is the behavior of Redis 7.2, but with different error words):
```
127.0.0.1:6666> lpos mylist foo rank -9223372036854775808
(error) ERR rank would overflow

127.0.0.1:6379> lpos mylist foo rank -9223372036854775808
(error) ERR value is out of range, value must between -9223372036854775807 and 9223372036854775807
```

Unrelated change: a small cleanup, return RedisParseErr instead of
RedisExecErr in Parse.

### DIFF
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -697,16 +697,19 @@ class CommandLPos : public Commander {
                   "RANK can't be zero: use 1 to start from "
                   "the first match, 2 from the second ... "
                   "or use negative to start from the end of the list"};
+        } else if (spec_.rank == LLONG_MIN) {
+          // Negating LLONG_MIN will cause an overflow, and is effectively be the same as passing -1.
+          return {Status::RedisParseErr, "rank would overflow"};
         }
       } else if (parser.EatEqICase("count")) {
         spec_.count = GET_OR_RET(parser.TakeInt());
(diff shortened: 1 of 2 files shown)


JoverZhang:
Thank you.

torwig (approved):

git-hulk (approved):

PR #1689, opened 2023-08-22
Move sha1 and redis rand sources to vendor directory

PragmaTwice:
We add a vendor directory for external headers and sources.

If an external source can be imported via CMake FetchContent without our modification, we just use FetchContent,
otherwise we need to put them to the vendor directory. 

Currently we put sha1.h/cc rand.h/cc into the directory.


git-hulk (approved):

enjoy-binbin (approved):

mapleFU (approved):

mapleFU:
cc @zncleon we can also use "vendor" directory to vendor the xxhash algorithm.

PR #1690, opened 2023-08-22
Remove valgrind config file

PragmaTwice:
Since the file (src/valgrind.sup) is outdated and out-of-maintain (and we currently use AddressSanitizer to detect memory bugs),
we remove it in this PR.


git-hulk (approved):

enjoy-binbin (approved):

tisonkun (approved):
Thank you!

Do we have other settings to use this script for debugging?

Or may you or @git-hulk can tell how we use this file previously?

git-hulk:
> Thank you!
> 
> Do we have other settings to use this script for debugging?
> 
> Or may you or @git-hulk can tell how we use this file previously?

We previously use Valgrind to test if any memory leaks on the local side, and I guess no one does it now since we have ASAN in CI.

PR #1691, opened 2023-08-23
Fix GEORADIUSBYMEMBER against non existing key reply

enjoy-binbin:
Current GEORADIUSBYMEMBER against non existing key will
a NotFound error, which is wrong:
```
127.0.0.1:6666> georadiusbymember non-existing member 1 km
(error) ERR NotFound:

127.0.0.1:6379> georadiusbymember non-existing member 1 km
(empty array)
```

georadiusbymember_ro is the same since they share the code.


torwig (approved):
LGTM. @enjoy-binbin, thank you for your rigorous checks.

PR #1693, opened 2023-08-23
Add splited bloom filter for RedisBloom support 

zncleon:
As RedisBloom do, we add splited_bloom_filter as sub_bloom_filter of the scalable bloom filter chain.


src/types/bloom_filter.h
```diff
@@ -0,0 +1,159 @@
...
+/// Putze et al.'s "Cache-,Hash- and Space-Efficient Bloom filters". The basic idea is to
+/// hash the item to a tiny Bloom filter which size fit a single cache line or smaller.
+///
+/// This implementation sets 8 bits in each tiny Bloom filter. Each tiny Bloom
+/// filter is 32 bytes to take advantage of 32-byte SIMD instructions.
+class BlockSplitBloomFilter {
+ public:
+  /// The constructor of BlockSplitBloomFilter. It uses XXH64 as hash function.
+  ///
+  /// \param pool memory pool to use.
```
mapleFU: Seems this line could be removed

src/types/bloom_filter.h
```diff
@@ -0,0 +1,159 @@
...
+  // of bit to set, one bit in each 32-bit word.
+  static constexpr uint32_t SALT[kBitsSetPerBlock] = {0x47b6137bU, 0x44974d91U, 0x8824ad5bU, 0xa2b7289dU,
+                                                      0x705495c7U, 0x2df1424bU, 0x9efc4947U, 0x5c6bfb31U};
+
+  // The underlying buffer of bitset.
+  std::string data_;
+
+  // The number of bytes of Bloom filter bitset.
+  uint32_t num_bytes_;
+};
```
mapleFU: fmt: add a line after here

src/types/bloom_filter.cc
```diff
@@ -0,0 +1,97 @@
...
+  auto* bitset32 = reinterpret_cast<uint32_t*>(data_.data());
+
+  for (int i = 0; i < kBitsSetPerBlock; i++) {
+    // Calculate mask for key in the given bitset.
+    const uint32_t mask = UINT32_C(0x1) << ((key * SALT[i]) >> 27);
+    bitset32[bucket_index * kBitsSetPerBlock + i] |= mask;
+  }
+}
+
+uint64_t BlockSplitBloomFilter::Hash(const char* data, size_t length) { return XXH64(data, length, /*seed=*/0); }
```
mapleFU: fmt: add a line after here

src/types/bloom_filter.h
```diff
@@ -0,0 +1,159 @@
...
+  /// Calculate optimal size according to the number of distinct values and false
+  /// positive probability.
+  ///
+  /// @param ndv The number of distinct values.
+  /// @param fpp The false positive probability.
+  /// @return it always return a value between kMinimumBloomFilterBytes and
+  /// kMaximumBloomFilterBytes, and the return value is always a power of 2
+  static uint32_t OptimalNumOfBytes(uint32_t ndv, double fpp) {
+    uint32_t optimal_num_of_bits = OptimalNumOfBits(ndv, fpp);
+    //    CHECK(IsMultipleOf8(optimal_num_of_bits));
```
mapleFU, line 99: @PragmaTwice can we use `CHECK` in glog? Or other function or macro can be used?

mapleFU (commented):
Background: This is part of BloomFilter support.

Original implemention uses self-written bloom filter with many arguments(like number of hash function, bloom filters). We port a BlockSplitBloomFilter[1] from arrow parquet as a storage. BlockSplitBloomFilter is widely used in Kudu, Impala, Parquet, and RocksDB Full Filter. And it can simplify the arguments to `{fpp(false-positive-rate), num-distinct-element}` (and deduce size from it)

The code is ported and pruned from https://github.com/apache/arrow/blob/main/cpp/src/parquet/bloom_filter.h 

cc @git-hulk @PragmaTwice 

[1] https://github.com/apache/parquet-format/blob/master/BloomFilter.md#technical-approach

git-hulk:
That's great, finally it's ready to be reviewed.

mapleFU:
@zncleon Seems that we need a `.clang-format-ignore` for `vendor` directory. clang-format or tidy is useful, but perhaps we don't need to apply it for vendor library

would you mind create a `.clang-format-ignore` file like:

```
src/vendor
```

Or @PragmaTwice do we have better way to do this?

PragmaTwice:
I think we can import xxhash via FetchContent instead of put them to vendor directory.

https://github.com/Cyan4973/xxHash

You can refer to our other dependences in `cmake/` directory, e.g. https://github.com/apache/kvrocks/blob/unstable/cmake/fmt.cmake.

Only if FetchContent does not apply, or we need to modify the external sources, we then put them to vendor.

src/types/bloom_filter.h
PragmaTwice: Yeah. But note that it need to be an assumption that cannot be faced in runtime if the function is used properly.

tests/cppunit/types/bloom_filter_test.cc
```diff
@@ -0,0 +1,184 @@
...
+
+  // Boundary check
+  testOptimalNumEstimation(4, 0.01, BlockSplitBloomFilter::kMinimumBloomFilterBytes * 8);
+  testOptimalNumEstimation(4, 0.25, BlockSplitBloomFilter::kMinimumBloomFilterBytes * 8);
+
+  testOptimalNumEstimation(std::numeric_limits<uint32_t>::max(), 0.01, kMaximumBloomFilterBytes * 8);
+  testOptimalNumEstimation(std::numeric_limits<uint32_t>::max(), 0.25, kMaximumBloomFilterBytes * 8);
+}
+
+}  // namespace test
```
PragmaTwice:
```suggestion
}  // namespace test

```

it is better to have a newline in the end of file.

PragmaTwice (commented):
Other looks fine to me, except the dep issue.

PragmaTwice:
If you do not familiar with cmake, I can help you to import xxhash. @zncleon

PragmaTwice:
I have submit a patch #1694 to import xxhash.

You can just use xxhash without any step after it is merged.

PragmaTwice:
It is merged. You can now just use xxhash without bringing any sources or changing cmake code.

src/types/bloom_filter.cc
```diff
@@ -0,0 +1,97 @@
...
+    num_bytes = static_cast<uint32_t>(NextPower2(num_bytes));
+  }
+
+  if (num_bytes > kMaximumBloomFilterBytes) {
+    num_bytes = kMaximumBloomFilterBytes;
+  }
+
+  num_bytes_ = num_bytes;
+  data_.resize(num_bytes_, 0);
+  return true;
```
git-hulk: This function always returns true, we can remove it if it has no possible false condition.

src/types/bloom_filter.cc
```diff
@@ -0,0 +1,97 @@
...
+      return false;
+    }
+  }
+  return true;
+}
+
+void BlockSplitBloomFilter::InsertHash(uint64_t hash) {
+  const auto bucket_index = static_cast<uint32_t>(((hash >> 32) * (num_bytes_ / kBytesPerFilterBlock)) >> 32);
+  const auto key = static_cast<uint32_t>(hash);
+  auto* bitset32 = reinterpret_cast<uint32_t*>(data_.data());
```
git-hulk, line 87: Should we check if the index is out of the range?

PragmaTwice:
Oh, since it is ported from Apache Parquet, I am not sure whether we need to put a statement to NOTICE file for license issue. cc @tisonkun

mapleFU:
>  ported from Apache Parquet

Since C++ Parquet is owned by Apache Arrow, maybe we require a notice lincense for apache arrow

mapleFU:
```
/home/runner/work/kvrocks/kvrocks/src/types/bloom_filter.h:113:14: error: variable 'num_bits' is not initialized [cppcoreguidelines-init-variables,-warnings-as-errors]
    uint32_t num_bits;
             ^
                      = 0
/home/runner/work/kvrocks/kvrocks/tests/cppunit/types/bloom_filter_test.cc:154:8: error: invalid case style for local variable 'testOptimalNumEstimation' [readability-identifier-naming,-warnings-as-errors]
  auto testOptimalNumEstimation = [](uint32_t ndv, double fpp, uint32_t num_bits) {
       ^~~~~~~~~~~~~~~~~~~~~~~~
       test_optimal_num_estimation
```

Two lint require to be fixed here

PragmaTwice (approved):

mapleFU (approved):

src/types/bloom_filter.cc
mapleFU: I guess it's safe currently because `bucket_index` is counted by `num_bytes_`

git-hulk (approved):

mapleFU:
Thanks all, merging...

PR #1695, opened 2023-08-24
Disable building of the unused shared library for jemalloc

jonahgao:
```shell
$ ls kvrocks/build/_deps/jemalloc-build/lib
libjemalloc.a  libjemalloc_pic.a  libjemalloc.so  libjemalloc.so.2
```
It appears that the shared library is not being utilized.


git-hulk:
@jonahgao Thanks for your good catch. That's right that we don't use the shared library.

git-hulk (approved):

PragmaTwice (approved):
LGTM. Thanks for your contribution!

PR #1699, opened 2023-08-24
Add the support of basic scalable bloom

zncleon:
This PR introduce the structure, metadata and basic functions(e.g. `RESERVE`, `ADD`, `EXIST`) of the Scalable BloomFilter. 

According to the RedisBloom do, we design metadata and subkey for Scalable BloomFilter:

```

key:

        +----------+------------+-----------+-----------+-----------+-------------+-------------+-----------+-----------+
key =>  |  flags   |  expire    |  version  |  size     |  n_filters|  expansion  |base_capacity| error_rate | bloom_bytes     |
        | (1byte)  | (Ebyte)    |  (8byte)  | (Sbyte)   | (2byte)   | (2byte)     | (4byte)     | (8byte)   | (4byte)   |
        +----------+------------+-----------+-----------+-----------+-------------+-------------+-----------+-----------+

subkey:

                      +---------------+
key|version|filter => |     value     |
                      +---------------+

filter is an int16 as index in parent n_filter
```


zncleon:
For BfMetadata, I change it to the `default_cf`. If stored it in  `metadata_cf`, it will influence the `DEL` and` KEYS`. Should we design a new cf for it, or just place it in `default_cf`.

cc @mapleFU

src/types/redis_sb_chain.h
```diff
@@ -0,0 +1,59 @@
...
+
+ private:
+  rocksdb::Status getSBChainMetadata(const Slice &ns_key, SBChainMetadata *metadata);
+  rocksdb::Status getBFMetadata(const Slice &bf_meta_key, BFMetadata *metadata);
+  rocksdb::Status createSBChain(const Slice &ns_key, double error_rate, uint32_t capacity, uint16_t expansion,
+                                uint16_t scaling, SBChainMetadata &sb_chain_metadata);
+  rocksdb::Status bloomCheckAdd(const Slice &bf_key, const std::string &item, ReadWriteMode mode, int &ret);
+
+  static void appendBFSuffix(const Slice &ns_key, uint16_t filters_index, std::string *output);
+  static void appendBFMetaSuffix(const Slice &bf_key, std::string *output);
```
PragmaTwice:
```suggestion
  static std::string appendBFSuffix(const Slice &ns_key, uint16_t filters_index);
  static std::string appendBFMetaSuffix(const Slice &bf_key);
```

It is better to have the function signature like this.

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+  PutFixed16(output, filters_index);
+}
+
+void SBChain ::appendBFMetaSuffix(const Slice &bf_key, std::string *output) {
+  output->clear();
+
+  output->append(bf_key.data(), bf_key.size());
+  output->append(kBloomFilterSeparator);
+  output->append("meta");
+}
```
PragmaTwice: Could these field be encoded as subkeys? So that the compaction procedure can work well.

Refer to the key encoding https://kvrocks.apache.org/community/data-structure-on-rocksdb#user-key-encoding , it seems the version field is skiped?

src/types/redis_sb_chain.h
```diff
@@ -0,0 +1,59 @@
...
+
+const char kBloomFilterSeparator[] = ":";
+const uint32_t kBFDefaultInitCapacity = 100;
+const double kBFDefaultErrorRate = 0.01;
+const uint16_t kBFDefaultExpansion = 2;
+const double kErrorTighteningRatio = 0.5;
+
+enum class ReadWriteMode {
+  MODE_READ = 0,
+  MODE_WRITE = 1,
```
PragmaTwice: the `MODE_` prefix can be removed?

src/storage/redis_metadata.h
```diff
@@ -198,3 +200,27 @@ class StreamMetadata : public Metadata {
   void Encode(std::string *dst) override;
   rocksdb::Status Decode(const std::string &bytes) override;
 };
+
+class SBChainMetadata : public Metadata {
+ public:
+  uint16_t n_filters;
```
mapleFU, line 209: Though personally I understand what these arguments means. But I think we can add comments here.

We can copy paste some comments from https://redis.io/docs/data-types/probabilistic/bloom-filter/ . Would can make reviewer understand more about the syntax here. Like:

```c++
/// number of child Bloom Filters 
uint16_t n_filters;
/// Adding an element to a Bloom filter never fails due to the data structure "filling up". Instead the error rate 
/// starts to grow. To keep the error close to the one set on filter initialisation - the bloom filter will auto-scale, 
/// meaning when capacity is reached an additional sub-filter will be created.
///
/// The size of the new sub-filter is the size of the last sub-filter multiplied by EXPANSION.
///
/// The default expansion value is 2.
///
/// For non-scaling, the value should be...
uint16_t scaling;
uint16_t expansion;
```

And by the way, we should:
1. if no scaling, define a macro or constant for it, and define what `expansion` should be in this case
2. Try to simplify two argument into just one?

src/storage/redis_metadata.cc
```diff
@@ -401,3 +401,73 @@ rocksdb::Status StreamMetadata::Decode(const std::string &bytes) {
...
+  Metadata::Encode(dst);
+
+  PutFixed16(dst, n_filters);
+  PutFixed16(dst, scaling);
+  PutFixed16(dst, expansion);
+}
+
+rocksdb::Status SBChainMetadata::Decode(const std::string &bytes) {
+  Slice input(bytes);
+  if (!GetFixed8(&input, &flags)) {
```
mapleFU, line 415: @PragmaTwice Can it be like:

```c++
Status s = Metadata::Decode(bytes);
```

To avoid copy-paste original `Metadata::Decode`'s code?

mapleFU (commented):
Thansk! This looks much more clear than previous patch! I'll take a carefully review this afternoon!

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+  bf_metadata->entries = entries;
+  bf_metadata->error = error;
+  bf_metadata->size = 0;
+
+  uint32_t bf_bytes = BlockSplitBloomFilter::OptimalNumOfBytes(entries, error);
+  bf_metadata->bf_bytes = bf_bytes;
+
+  BlockSplitBloomFilter block_split_bloom_filter;
+  block_split_bloom_filter.Init(bf_bytes);
+  *bf_data = block_split_bloom_filter.GetData();
```
mapleFU: This would copy these data. I prefer a interface like:

```c++
void SBChain::bfInit(BFMetadata *bf_metadata, uint32_t entries, double error, std::unique_ptr<BlockSplitBloomFilter>) {
  bf_metadata->entries = entries;
  bf_metadata->error = error;
  bf_metadata->size = 0;

  uint32_t bf_bytes = BlockSplitBloomFilter::OptimalNumOfBytes(entries, error);
  bf_metadata->bf_bytes = bf_bytes;

  *block_split_bloom_filter = std::make_unique<BlockSplitBloomFilter>();
  block_split_bloom_filter->Init(bf_bytes);
}
```

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+
+rocksdb::Status SBChain::getBFMetadata(const Slice &bf_meta_key, BFMetadata *metadata) {
+  LatestSnapShot ss(storage_);
+  rocksdb::ReadOptions read_options;
+  read_options.snapshot = ss.GetSnapShot();
+  std::string bytes;
+  auto s = storage_->Get(read_options, bf_meta_key, &bytes);
+
+  std::string old_metadata;
+  metadata->Encode(&old_metadata);
```
mapleFU: Hmmm why same `BFMetadata` is beging `Encode` and `Decode` multiple times?

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+    rocksdb::Status create_s =
+        createSBChain(ns_key, kBFDefaultErrorRate, kBFDefaultInitCapacity, kBFDefaultExpansion, 0, sb_chain_metadata);
+    if (!create_s.ok()) return create_s;
+  } else if (!s.ok()) {
+    return s;
+  }
+
+  std::vector<std::string> bf_key_list;
+  std::vector<BFMetadata> bf_metadata_list;
+  for (u_int16_t i = 0; i < sb_chain_metadata.n_filters; ++i) {
```
mapleFU: We can extract this loop to a function.

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+    BFMetadata bf_metadata;
+    s = getBFMetadata(bf_meta_key, &bf_metadata);
+    if (!s.ok()) return s;
+    bf_key_list.push_back(bf_key);
+    bf_metadata_list.push_back(bf_metadata);
+  }
+
+  BFMetadata cur_bf_metadata = bf_metadata_list.back();
+  std::string cur_bf_key = bf_key_list.back();
+  if (cur_bf_metadata.size + 1 > cur_bf_metadata.entries) {
```
mapleFU: Should this line after "check" and before "insert"?

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+  SBChainMetadata sb_chain_metadata;
+  rocksdb::Status s = getSBChainMetadata(ns_key, &sb_chain_metadata);
+  if (s.IsNotFound()) return rocksdb::Status::NotFound("key is not found");
+  if (!s.ok()) return s;
+
+  std::vector<std::string> bf_key_list;
+  for (uint16_t i = 0; i < sb_chain_metadata.n_filters; ++i) {
+    std::string bf_key;
+    appendBFSuffix(ns_key, i, &bf_key);
+    bf_key_list.push_back(bf_key);
```
mapleFU:
```suggestion
    bf_key_list.push_back(std::move(bf_key));
```

src/types/redis_sb_chain.h
```diff
@@ -0,0 +1,59 @@
...
+};
+
+class SBChain : public Database {
```
mapleFU: We can naming it `BloomChain`, `SBChain` is same as `RedisBloom`, but the naming is a bit confusing...

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  const auto &item_string = item.ToString();
+  // check
+  for (int i = sb_chain_metadata.n_filters - 1; i >= 0; --i) {
```
mapleFU: (As for direction, now we and RedisBloom are searching from largest BloomFilter to smaller BloomFilter, but I wonder  if we can searching it from smallest to largest? We can leave a todo or note here.)

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+  if (cur_bf_metadata.size + 1 > cur_bf_metadata.entries) {
+    return rocksdb::Status::Aborted("filter is full");
+  }
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisBloomFilter, {"insert"});
+  batch->PutLogData(log_data.Encode());
+
+  bool item_exist = false;
+  const auto &item_string = item.ToString();
```
mapleFU:
```suggestion
  std::string item_string = item.ToString();
```

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+  batch->Put(bf_key, bf_data);
+
+  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+rocksdb::Status SBChain::bloomCheckAdd(const Slice &bf_key, const std::string &item, ReadWriteMode mode, int &ret) {
+  std::string bf_data;
+  rocksdb::Status s = storage_->Get(rocksdb::ReadOptions(), bf_key, &bf_data);
+  BlockSplitBloomFilter block_split_bloom_filter;
+  block_split_bloom_filter.Init(bf_data);
```
mapleFU:
```suggestion
  block_split_bloom_filter.Init(std::move(bf_data));
```

src/types/redis_sb_chain.cc
```diff
@@ -0,0 +1,240 @@
...
+
+  ret = 0;
+  if (found && mode == ReadWriteMode::MODE_READ) {
+    ret = 1;
+  } else if (!found && mode == ReadWriteMode::MODE_WRITE) {
+    auto batch = storage_->GetWriteBatchBase();
+    ret = 1;
+    block_split_bloom_filter.InsertHash(h);
+    bf_data = block_split_bloom_filter.GetData();
+    batch->Put(bf_key, bf_data);
```
mapleFU:
```suggestion
    batch->Put(bf_key, block_split_bloom_filter.GetData());
```

src/types/redis_sb_chain.cc
zncleon: Sure, these shoule be encoded as subkeys. I will fix it.

src/types/redis_sb_chain.cc
zncleon: It’s just like the function `getMetadata`. I think if run error(expired or other), metadata will just return the old_metadata

zncleon:
I change the metadata structure as #1701 solution 2.
Now, there is only one metadata:

BloomChainMetadata: (`num_filters`, `expansion`, `base_capacity`, `error_rate`, `bloom_bytes`).
and the `sum_num_distinct` use Metadata's `size`

src/storage/redis_metadata.h
```diff
@@ -198,3 +200,47 @@ class StreamMetadata : public Metadata {
...
+  /// The desired probability for false positives.
+  ///
+  /// The rate is a decimal value between 0 and 1. For example, for a desired false positive rate of 0.1% (1 in 1000),
+  /// error_rate should be set to 0.001.
+  ///
+  /// The default error_rate value is 0.01.
+  double error_rate;
+
+  /// The total number of bytes allocated for all sub-filters.
+  uint32_t bloom_bytes;
```
mapleFU, line 237: `uint32_t` is ok here, but we should implement size-checking carefully latter

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,185 @@
...
+  sub_key.append(kBloomFilterSeparator);
+  PutFixed16(&sub_key, filters_index);
+
+  std::string bf_key = InternalKey(ns_key, sub_key, metadata.version, storage_->IsSlotIdEncoded()).Encode();
+  return bf_key;
+}
+
+void BloomChain::getBFKeyList(const Slice &ns_key, const BloomChainMetadata &metadata,
+                              std::vector<std::string> &bf_key_list) {
+  for (u_int16_t i = 0; i < metadata.n_filters; ++i) {
```
mapleFU: You can `bf_key_list.reserve(metadata.n_filters);` first

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,185 @@
...
+  sb_chain_metadata.size = 0;
+
+  sb_chain_metadata.error_rate = error_rate;
+  sb_chain_metadata.base_capacity = capacity;
+  sb_chain_metadata.bloom_bytes = BlockSplitBloomFilter::OptimalNumOfBytes(capacity, error_rate);
+
+  std::string bf_data;
+  BlockSplitBloomFilter block_split_bloom_filter;
+  block_split_bloom_filter.Init(sb_chain_metadata.bloom_bytes);
+  bf_data = block_split_bloom_filter.GetData();
```
mapleFU: `GetData` returns a `const std::string&`, we need take it carefully and try our best not to copying these bytes:

```c++
 batch->Put(bf_key, block_split_bloom_filter.GetData());
```

The code above is ok in Line 70

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,185 @@
...
+    if (ret == 1) {
+      item_exist = true;
+      break;
+    }
+  }
+
+  // insert
+  if (!item_exist) {
+    if (sb_chain_metadata.size + 1 > sb_chain_metadata.GetCapacity()) {
+      return rocksdb::Status::Aborted("filter is full");
```
mapleFU, line 161: We can leave a todo here, since scaling would be supported later

mapleFU (commented):
Looks much clearer now!

src/storage/redis_metadata.h
zncleon: It seems `uint32_t` is big enough for bloom_bytes. What is size-checking for ?

src/storage/redis_metadata.cc
PragmaTwice: Sure.

src/storage/redis_metadata.h
```diff
@@ -200,3 +202,47 @@ class StreamMetadata : public Metadata {
...
+
+  explicit BloomChainMetadata(bool generate_version = true) : Metadata(kRedisBloomFilter, generate_version) {}
+
+  void Encode(std::string *dst) override;
+  rocksdb::Status Decode(const std::string &bytes) override;
```
```suggestion
  rocksdb::Status Decode(Slice bytes) override;
```

PragmaTwice (commented):
We made some small changes to Metadata, so you need to fix the interface.

Sorry for the inconvenience.

src/storage/redis_metadata.h
mapleFU: Ah yeah, currently `BlockSplitBloomFilter` has maximum size 128MB, so it's enough. We can report the error message to user if the allocation is too large. Let leave it in future optimizations

src/storage/redis_metadata.h
zncleon: no matter, I will fix it.

src/storage/redis_metadata.cc
PragmaTwice: I have looked into this and it seems currently it is hard to reuse Metadata::Decode since the output Slice is missing. You can just copy code here and I will fix it later after it is merged.

src/commands/cmd_bloom_filter.cc
```diff
@@ -0,0 +1,117 @@
...
+class CommandBFReserve : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto parse_error_rate = ParseFloat<double>(args[2]);
+    if (!parse_error_rate) {
+      return {Status::RedisParseErr, errValueIsNotFloat};
+    }
+    error_rate_ = *parse_error_rate;
+    if (error_rate_ >= 1 || error_rate_ <= 0) {
+      return {Status::RedisParseErr, "0 < error rate range < 1"};
```
git-hulk:
```suggestion
      return {Status::RedisParseErr, "error rate should be between 0 and 1"};
```

src/commands/cmd_bloom_filter.cc
```diff
@@ -0,0 +1,117 @@
...
+
+    *output = redis::Integer(ret);
+    return Status::OK();
+  }
+};
+
+REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandBFReserve>("bf.reserve", -4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandBFAdd>("bf.add", 3, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandBFExist>("bf.exist", 3, "read-only", 1, 1, 1), )
+}  // namespace redis
```
git-hulk:
```suggestion
}  // namespace redis

```

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,184 @@
...
+
+  std::string bf_key = getBFKey(ns_key, *metadata, metadata->n_filters - 1);
+  batch->Put(bf_key, block_split_bloom_filter.GetData());
+
+  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+rocksdb::Status BloomChain::bloomCheckAdd(const Slice &bf_key, const std::string &item, ReadWriteMode mode, int *ret) {
+  std::string bf_data;
+  rocksdb::Status s = storage_->Get(rocksdb::ReadOptions(), bf_key, &bf_data);
```
git-hulk, line 74: We should check the get result?

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  std::vector<std::string> bf_key_list;
+  getBFKeyList(ns_key, metadata, &bf_key_list);
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisBloomFilter, {"insert"});
+  batch->PutLogData(log_data.Encode());
+
+  bool item_exist = false;
```
git-hulk: Looks like we can check the add status by the index 'i'

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,184 @@
...
+  std::string sb_chain_metadata_bytes;
+  metadata.Encode(&sb_chain_metadata_bytes);
+  batch->Put(metadata_cf_handle_, ns_key, sb_chain_metadata_bytes);
+
+  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+rocksdb::Status BloomChain::Exist(const Slice &user_key, const Slice &item, int *ret) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+  LockGuard guard(storage_->GetLockManager(), ns_key);
```
git-hulk: I think we can split the check and add logic. For the existing operation, it's unnecessary to lock for adding the missing bloom filter, because the BloomFilter read operations would be much larger than the write operations in most of scenarios. So it'd be better to non-lock for that.

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,201 @@
...
+  BloomChainMetadata metadata;
+  rocksdb::Status s = getBloomChainMetadata(ns_key, &metadata);
+
+  if (s.IsNotFound()) {
+    rocksdb::Status create_s =
+        createBloomChain(ns_key, kBFDefaultErrorRate, kBFDefaultInitCapacity, kBFDefaultExpansion, &metadata);
+    if (!create_s.ok()) return create_s;
+  } else if (!s.ok()) {
+    return s;
+  }
```
```suggestion

  if (s.IsNotFound()) {
    s =
        createBloomChain(ns_key, kBFDefaultErrorRate, kBFDefaultInitCapacity, kBFDefaultExpansion, &metadata);
  }
  if (!s.ok()) {
    return s;
  }
```

Would this a bit more clear?

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,201 @@
...
+
+  metadata->error_rate = error_rate;
+  metadata->base_capacity = capacity;
+  metadata->bloom_bytes = BlockSplitBloomFilter::OptimalNumOfBytes(capacity, error_rate);
+
+  BlockSplitBloomFilter block_split_bloom_filter;
+  block_split_bloom_filter.Init(metadata->bloom_bytes);
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisBloomFilter, {"createSBChain"});
```
mapleFU, line 59: The logic here is related to `WriteBatchExtractor::PutCF`.

```c++
constexpr std::string_view CreateBloomChain = "...";
constexpr std::string_view InsertBloomFilter = "...";
```

And in `src/storage/batch_extractor.cc`, you'd better have the correspond behavior. Like:

```
case kRedisBloomFilter: {
  // ...
}
```

And follow the `kRedisBitmap` style here.

mapleFU (commented):
General look good to me now. Final we need fix the `WriteBatchLogData` and `kRedisBloomFilter` behavior, and add test for it.

test can be seen in `tests/gocase/integration/slotmigrate/slotmigrate_test.go`

src/types/redis_bloom_chain.cc
mapleFU: Actually I think we can delay locking when checking the "last" bloom-filter.

During insert:
1. Acquire n_filters
2. (without-lock)  check filter `1...n_filters-1`
3. (lock) check final filter. It might contains a "get metadata and check n_filters" to verify the result in (1) is still right, if not, we need to check the new inserted filter here..

This is possible because `1...n_filters-1` is immutable. This can be regard as a future optimization. At least our code is right currently.

src/commands/cmd_bloom_filter.cc
```diff
@@ -0,0 +1,117 @@
...
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = redis::Integer(ret);
+    return Status::OK();
+  }
+};
+
+REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandBFReserve>("bf.reserve", -4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandBFAdd>("bf.add", 3, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandBFExist>("bf.exist", 3, "read-only", 1, 1, 1), )
```
PragmaTwice: Seems it is EXISTS, not EXIST.

tests/gocase/unit/type/bloom/bloom_test.go
```diff
@@ -0,0 +1,82 @@
...
+		require.ErrorContains(t, rdb.Do(ctx, "bf.reserve", key, "0.01", "1000", "expansion", "1.5").Err(), "ERR encounter non-integer characters")
+		require.ErrorContains(t, rdb.Do(ctx, "bf.reserve", key, "0.01", "1000", "expansion", "123asd").Err(), "ERR encounter non-integer characters")
+	})
+
+	t.Run("Reserve a bloom filter with nonscaling and expansion", func(t *testing.T) {
+		require.NoError(t, rdb.Del(ctx, key).Err())
+		require.ErrorContains(t, rdb.Do(ctx, "bf.reserve", key, "0.01", "1000", "expansion", "1", "nonscaling").Err(), "ERR nonscaling filters cannot expand")
+		require.ErrorContains(t, rdb.Do(ctx, "bf.reserve", key, "0.01", "1000", "nonscaling", "expansion", "1").Err(), "ERR nonscaling filters cannot expand")
+	})
+}
```
PragmaTwice, line 119: It would be better if you add some test for ADD and EXISTS in go test.

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,201 @@
...
+ *
+ */
+
+#include "redis_bloom_chain.h"
+
+namespace redis {
+
+std::string BloomChain::getBFKey(const Slice &ns_key, const BloomChainMetadata &metadata, uint16_t filters_index) {
+  std::string sub_key;
+  sub_key.append(kBloomFilterSeparator);
```
PragmaTwice: Seems we can remove the separator? IMO it seems same with or without the seperator. cc @mapleFU

mapleFU: Yeah, actually we can. As https://kvrocks.apache.org/community/data-structure-on-rocksdb/#list-sub-keys-values says, other subkey doesn't have any `separator`. We can refer to List or zset, like

```c++
    PutFixed16(&buf, reversed ? --pivot_index : ++pivot_index);
    std::string to_update_key = InternalKey(ns_key, buf, metadata.version, storage_->IsSlotIdEncoded()).Encode();
    batch->Put(to_update_key, iter->value());
```

src/types/redis_bloom_chain.cc
mapleFU: Or you can leave a todo in `src/storage/batch_extractor.cc`, I can add an todo in issue tracker

zncleon: OK, I will fix it in the next patch.

mapleFU: You can fix it after all function is done :-) migration might be refactor later

src/types/redis_bloom_chain.cc
```diff
@@ -0,0 +1,198 @@
...
+  if (check_index < 0) {
+    if (metadata.size + 1 > metadata.GetCapacity()) {  // TODO: scaling would be supported later
+      return rocksdb::Status::Aborted("filter is full");
+    }
+    s = bloomAdd(bf_key_list.back(), item_string, ret);
+    if (!s.ok()) return s;
+    metadata.size += *ret;
+  }
+
+  std::string sb_chain_metadata_bytes;
```
mapleFU, line 168: Only need to `Put` when `check_index < 0`?

mapleFU (commented):
Rest LGTM

Also, I've updated the description of the patch.

src/types/redis_bloom_chain.cc
zncleon: Sure, only need to `Put` when `metadata` change. But if I reuse this functions in `MADD`, I think it would be better to leave the `Put` outside and only `Put` once when `metadata.size` change.

mapleFU: Ok, you can support `INFO` in next patch. It's easy and it can help us debug:
https://github.com/RedisBloom/RedisBloom/blob/21a2620e75873353fead8c5d70950d3791e36b18/docs/commands/bf.info.md#L4

mapleFU (approved):
LGTM, thanks!

git-hulk (approved):

PragmaTwice (approved):

mapleFU:
Thanks all, merging...

PR #1700, opened 2023-08-25
Use MultiLockGuard to guarantee atomicity for multiple keys commands

### LINKED ISSUE #1692: Kvrocks can't guarantee atomicity for multiple keys commands
### Search before asking

### Motivation

In kvrocks, for multiple keys commands, we may break atomicity.
such as `SDIFF、SUNIONSTORE、SINTERSTORE、SDIFFSTORE、ZUNIONSTORE、ZINTERSTORE`

For `SDIFFSTORE destination key1 [key2 ...]` command, kvrocks will read key1, key2 ... orderly without any lock, and then restore the diff into destination key. So maybe some keys may be changed when we read them orderly, that breaks atomicity but redis can guarantee atomicity.

Thank you @oranagra reports this.

### Solution

Now we have `MultiLockGuard`, i think it is easy to fix this.  We can lock these keys before we read or write, so it is impossible to change these keys.

enjoy-binbin:
In kvrocks, for multiple keys commands, we may break atomicity.
such as `ZUNION`, `ZUNIONSTORE`, `ZINTERSTORE`, `SUNION`, `SUNIONSTORE`,
`SINTER`, `SINTERSTORE`, `SDIFF` and `SDIFFSTORE`.

For `SDIFFSTORE destination key1 [key2 ...]` command, kvrocks will
read key1, key2 ... orderly without any lock, and then restore the
diff into destination key. So maybe some keys may be changed when
we read them orderly, that breaks atomicity but redis can guarantee
atomicity.

In this PR, we are using `MultiLockGuard` to lock these keys before
we read or write, so it is impossible to change these keys. This change
affects all commands listed previously.

Fixes #1692.

### DIFF
--- a/src/types/redis_set.cc
+++ b/src/types/redis_set.cc
@@ -272,6 +272,15 @@ rocksdb::Status Set::Scan(const Slice &user_key, const std::string &cursor, uint
  * DIFF key1 key2 key3 = {b,d}
  */
 rocksdb::Status Set::Diff(const std::vector<Slice> &keys, std::vector<std::string> *members) {
+  std::string ns_key;
+  std::vector<std::string> lock_keys;
+  lock_keys.reserve(keys.size());
+  for (const auto key : keys) {
+    ns_key = AppendNamespacePrefix(key);
+    lock_keys.emplace_back(ns_key);
+  }
+  MultiLockGuard guard(storage_->GetLockManager(), lock_keys);
+
   members->clear();
   std::vector<std::string> source_members;
   auto s = Members(keys[0], &source_members);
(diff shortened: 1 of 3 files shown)


torwig (dismissed):
LGTM

src/types/redis_set.cc
```diff
@@ -272,6 +272,15 @@ rocksdb::Status Set::Scan(const Slice &user_key, const std::string &cursor, uint
  * DIFF key1 key2 key3 = {b,d}
  */
 rocksdb::Status Set::Diff(const std::vector<Slice> &keys, std::vector<std::string> *members) {
+  std::string ns_key;
+  std::vector<std::string> lock_keys;
+  lock_keys.reserve(keys.size());
+  for (const auto key : keys) {
+    ns_key = AppendNamespacePrefix(key);
+    lock_keys.emplace_back(ns_key);
```
mapleFU:
```suggestion
    lock_keys.emplace_back(std::move(ns_key));
```

mapleFU (dismissed):
Rest looks good to me, except a nit. `lock_keys.emplace_back(ns_key);` can add a move here.

mapleFU (approved):
Thanks!

(by the way, the code below provide the same ability, you can also use this if you like)

```
lock_keys.emplace_back(AppendNamespacePrefix(key));
```

src/types/redis_set.cc
enjoy-binbin: thanks, updated.

PR #1702, opened 2023-08-25
Add LASTSAVE command

xq2010:
add redis LASTSAVE command, modify the return value of info.persistence.rdb_last_save_time to be consistent with Redis, return start_time_  when bgsave is not performed.

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -985,7 +985,7 @@ class CommandLastSave : public Commander {
       return {Status::RedisExecErr, errAdminPermissionRequired};
     }
 
-    int unix_sec = svr->GetLastBgsaveTime();
+    std::int64_t unix_sec = svr->GetLastBgsaveTime();
     *output = redis::Integer(unix_sec);
     return Status::OK();
   }
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -978,6 +978,19 @@ static uint64_t GenerateConfigFlag(const std::vector<std::string> &args) {
   return 0;
 }
 
+class CommandLastSave : public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    if (!conn->IsAdmin()) {
+      return {Status::RedisExecErr, errAdminPermissionRequired};
+    }
+
+    int unix_sec = svr->GetLastBgsaveTime();
+    *output = redis::Integer(unix_sec);
+    return Status::OK();
+  }
+};
+
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandAuth>("auth", 2, "read-only ok-loading", 0, 0, 0),
                         MakeCmdAttr<CommandPing>("ping", -1, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandSelect>("select", 2, "read-only", 0, 0, 0),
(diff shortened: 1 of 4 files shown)


git-hulk:
Hi @xq2010, thanks for your contribution. Could you please fix the issue in CI: https://github.com/apache/kvrocks/actions/runs/5974247406/job/16208064626#step:8:1361

src/server/server.h
```diff
@@ -211,6 +211,7 @@ class Server {
   void SetLastRandomKeyCursor(const std::string &cursor);
 
   static int GetCachedUnixTime();
+  int GetLastBgsaveTime();
```
```suggestion
  std::int64_t GetLastBgsaveTime();
```

Seems we can return int64 instead of int to avoid narrow conversion reported in CI.

PragmaTwice:
@xq2010 You need to modify the function signature not only in declaration but also in definition.

mapleFU:
The new implemention keeps consistent with Redis. But I think we need to fix the lint :-)

src/commands/cmd_server.cc
```diff
@@ -978,6 +978,19 @@ static uint64_t GenerateConfigFlag(const std::vector<std::string> &args) {
...
 }
 
+class CommandLastSave : public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    if (!conn->IsAdmin()) {
+      return {Status::RedisExecErr, errAdminPermissionRequired};
+    }
+
+    int unix_sec = svr->GetLastBgsaveTime();
```
enjoy-binbin:
```suggestion
    std::int64_t unix_sec = svr->GetLastBgsaveTime();
```

src/server/server.cc
```diff
@@ -990,6 +990,11 @@ int Server::GetCachedUnixTime() {
   return unix_time.load();
 }
 
+int Server::GetLastBgsaveTime() {
```
enjoy-binbin:
```suggestion
std::int64_t Server::GetLastBgsaveTime() {
```

xq2010: Thanks for the fix!

xq2010:
> 



> The new implemention keeps consistent with Redis. But I think we need to fix the lint :-)↳

Okay, I'll fix the lint and test error.

xq2010:
> @xq2010 You need to modify the function signature not only in declaration but also in definition.↳

thanks!

PragmaTwice:
@xq2010 The CI failed, and there are also conflicts against the unstable branch.

You need to solve the conflict and fix the code to make CI pass. Sorry but currently I have no time to help you check the root cause. You can check the CI log or just run go test in your local.

src/commands/cmd_server.cc
```diff
@@ -980,6 +980,18 @@ static uint64_t GenerateConfigFlag(const std::vector<std::string> &args) {
...
 }
 
+class CommandLastSave : public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    if (!conn->IsAdmin()) {
+      return {Status::RedisExecErr, errAdminPermissionRequired};
+    }
+
+    std::int64_t unix_sec = svr->GetLastBgsaveTime();
```
torwig: Could we remove the `std::` prefix here and in all the similar places across this PR?

xq2010: I will remove it later

torwig (approved):
LGTM

mapleFU (approved):

git-hulk (approved):

PR #1703, opened 2023-08-25
Fix wrong metadata parsing in CompactOnExpiredCollector

PragmaTwice:
The bug only affects users who enable the option ENABLE_NEW_ENCODING.

We fixed the metadata parsing (on expire time field) in CompactOnExpiredCollector.

Besides, we change the type of some timestamp from int to int64.

### DIFF
--- a/src/storage/table_properties_collector.cc
+++ b/src/storage/table_properties_collector.cc
@@ -53,7 +53,7 @@ rocksdb::Status CompactOnExpiredCollector::AddUserKey(const rocksdb::Slice &key,
   if (!s.ok()) return rocksdb::Status::OK();
 
   total_keys_ += metadata.size;
-  if (metadata.ExpireAt(Server::GetCachedUnixTime())) {
+  if (metadata.ExpireAt(Server::GetCachedUnixTime() * 1000)) {
     deleted_keys_ += metadata.size + 1;
   }
   return rocksdb::Status::OK();
--- a/src/storage/table_properties_collector.cc
+++ b/src/storage/table_properties_collector.cc
@@ -30,10 +30,6 @@
 rocksdb::Status CompactOnExpiredCollector::AddUserKey(const rocksdb::Slice &key, const rocksdb::Slice &value,
                                                       rocksdb::EntryType entry_type, rocksdb::SequenceNumber,
                                                       uint64_t) {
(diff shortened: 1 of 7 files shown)


src/storage/table_properties_collector.cc
```diff
@@ -47,23 +43,18 @@ rocksdb::Status CompactOnExpiredCollector::AddUserKey(const rocksdb::Slice &key,
...
-  total_keys_ += subkeys;
-  int now = Server::GetCachedUnixTime();
-  if ((expired > 0 && expired < static_cast<uint32_t>(now)) || (type != kRedisString && subkeys == 0)) {
-    deleted_keys_ += subkeys + 1;
+
+  Metadata metadata(RedisType::kRedisNone);
+  auto s = metadata.Decode(value);
+  if (!s.ok()) return rocksdb::Status::OK();
+
+  total_keys_ += metadata.size;
```
git-hulk, line 55: For string type, its size means the length of the string, so should only increase 1  for total and delete keys.

PragmaTwice: I think for string the metadata.size == 0, and you can check Metadata class code to confirm it.

git-hulk: My bad, you're right.

git-hulk (dismissed):

git-hulk (dismissed):

git-hulk (approved):

PragmaTwice:
Carry comment from @git-hulk in PragmaTwice/kvrocks#2 for better context:

> RocksDB compaction allows to use of trivial move mode(move file into the next level instead of using compaction filter to generate the output file) when the compaction is not triggered by the manual way.

> So after this fix, NeedCompact will return the correct result and then wait for the background compaction schedule(become no manual way so it allows the trivial move), so after we call the storage->Compact, it won't manually compact the moved file.

> I fixed this issue by using force bottom-level compaction when triggering the manual compaction, it's our expected behavior.

> See https://github.com/facebook/rocksdb/blob/ba597514309b686d8addb59616f067d5522186b7/db/db_impl/db_impl_compaction_flush.cc#L3358

torwig (approved):

PragmaTwice:
Thanks all!

PR #1704, opened 2023-08-25
Add the support of the XGROUP CREATE and DESTROY command

### LINKED ISSUE #1586: Add the support of the XGROUP CREATE and DESTROY command
This issue intend to implement [XGROUP CREATE](https://redis.io/commands/xgroup-create/) and [XGROUP DESTROY](https://redis.io/commands/xgroup-destroy/) like redis.

see #1654 for more discusstions

Yangsx-1:
Close #1586 
Implement discussion #1654 

the subkeys and values for consumer group:
```
1.consumer_group_metadata: key(stream_name) + version + sizeof(consumer_group_name) + $(consumer_group_name) + METADATA --> consumer_group_metadata
2.consumer_metadata: key(stream_name) + version + sizeof(consumer_group_name) + $(consumer_group_name) + sizeof(consumer_name) + $(consumer_name) + METADATA --> consumer_metadata
3.PEL entries: key(stream_name) + version + sizeof(consumer_group_name) + $(consumer_group_name) + $(stream_entryid) --> values
```

In this PR, we only focus on 1.consumer_group_metadata

### DIFF
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -261,7 +261,7 @@ class CommandXGroup : public Commander {
         return {Status::RedisExecErr, s.ToString()};
       }
 
-      *output = redis::BulkString("OK");
+      *output = redis::SimpleString("OK");
     }
 
     if (subcommand_ == "destroy") {
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -250,9 +250,14 @@ class CommandXGroup : public Commander {
     }
 
     if (subcommand_ == "destroy") {
+      if (args.size() != 4) {
+        return {Status::RedisParseErr, errWrongNumOfArguments};
+      }
+
+      return Status::OK();
     }
 
-    return {Status::RedisParseErr, "unknown subcommand" + subcommand_};
+    return {Status::RedisParseErr, "unknown subcommand"};
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
@@ -265,9 +270,23 @@ class CommandXGroup : public Commander {
       }
 
       *output = redis::BulkString("OK");
-
-      return Status::OK();
     }
+
+    if (subcommand_ == "destroy") {
+      uint64_t delete_cnt = 0;
+      auto s = stream_db.DestroyGroup(stream_name_, group_name_, &delete_cnt);
+      if (!s.ok()) {
+        return {Status::RedisExecErr, s.ToString()};
+      }
+
+      if (delete_cnt > 0) {
+        *output = redis::Integer(1);
+      } else {
+        *output = redis::Integer(0);
+      }
+    }
+
+    return Status::OK();
   }
 
  private:
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -22,6 +22,7 @@
 #include <stdexcept>
 
 #include "commander.h"
+#include "command_parser.h"
 #include "error_constants.h"
 #include "event_util.h"
 #include "server/server.h"
@@ -208,9 +209,10 @@ class CommandXDel : public Commander {
 class CommandXGroup : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    subcommand_ = util::ToLower(args[1]);
-    stream_name_ = args[2];
-    group_name_ = args[3];
+    CommandParser parser(args, 1);
+    subcommand_ = util::ToLower(GET_OR_RET(parser.TakeStr()));
+    stream_name_ = GET_OR_RET(parser.TakeStr());
+    group_name_ = GET_OR_RET(parser.TakeStr());
     if (std::isdigit(group_name_[0])) {
       return {Status::RedisParseErr, "group name cannot start with number"};
     }
@@ -220,29 +222,24 @@ class CommandXGroup : public Commander {
         return {Status::RedisParseErr, errWrongNumOfArguments};
       }
 
-      xgroup_create_options_.last_id = args[4];
+      xgroup_create_options_.last_id = GET_OR_RET(parser.TakeStr());
 
-      for (size_t i = 5; i < args.size(); ++i) {
-        auto val = util::ToLower(args[i]);
-        if (val == "mkstream") {
+      while (parser.Good()) {
+        if (parser.EatEqICase("mkstream")) {
           xgroup_create_options_.mkstream = true;
-          continue;
         }
-
-        if (val == "entriesread") {
-          if (i + 1 >= args.size()) {
+        else if (parser.EatEqICase("entriesread")) {
+          if (!parser.Good()) {
             return {Status::RedisParseErr, errUnknownSubcommandOrWrongArguments};
           }
-          auto parse_result = ParseInt<int64_t>(args[i + 1], 10);
-          if (!parse_result) {
+          auto parse_result = parser.TakeInt<int64_t>();
+          if (!parse_result.IsOK()) {
             return {Status::RedisParseErr, errValueNotInteger};
           }
-          if (*parse_result < 0 && *parse_result != -1) {
+          if (parse_result.GetValue() < 0 && parse_result.GetValue() != -1) {
             return {Status::RedisParseErr, "value for ENTRIESREAD must be positive or -1"};
           }
-          xgroup_create_options_.entries_read = *parse_result;
-          ++i;
-          continue;
+          xgroup_create_options_.entries_read = parse_result.GetValue();
         }
       }
 
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -205,6 +205,78 @@ class CommandXDel : public Commander {
   std::vector<redis::StreamEntryID> ids_;
 };
 
+class CommandXGroup : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    subcommand_ = util::ToLower(args[1]);
+    stream_name_ = args[2];
+    group_name_ = args[3];
+    if (std::isdigit(group_name_[0])) {
+      return {Status::RedisParseErr, "group name cannot start with number"};
+    }
+
+    if (subcommand_ == "create") {
+      if (args.size() < 5 || args.size() > 8) {
+        return {Status::RedisParseErr, errWrongNumOfArguments};
+      }
+
+      xgroup_create_options_.last_id = args[4];
+
+      for (size_t i = 5; i < args.size(); ++i) {
+        auto val = util::ToLower(args[i]);
+        if (val == "mkstream") {
+          xgroup_create_options_.mkstream = true;
+          continue;
+        }
+
+        if (val == "entriesread") {
+          if (i + 1 >= args.size()) {
+            return {Status::RedisParseErr, errUnknownSubcommandOrWrongArguments};
+          }
+          auto parse_result = ParseInt<int64_t>(args[i + 1], 10);
+          if (!parse_result) {
+            return {Status::RedisParseErr, errValueNotInteger};
+          }
+          if (*parse_result < 0 && *parse_result != -1) {
+            return {Status::RedisParseErr, "value for ENTRIESREAD must be positive or -1"};
+          }
+          xgroup_create_options_.entries_read = *parse_result;
+          ++i;
+          continue;
+        }
+      }
+
+      return Status::OK();
+    }
+
+    if (subcommand_ == "destroy") {
+    }
+
+    return {Status::RedisParseErr, "unknown subcommand" + subcommand_};
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::Stream stream_db(svr->storage, conn->GetNamespace());
+
+    if (subcommand_ == "create") {
+      auto s = stream_db.CreateGroup(stream_name_, xgroup_create_options_, group_name_);
+      if (!s.ok()) {
+        return {Status::RedisExecErr, s.ToString()};
+      }
+
+      *output = redis::BulkString("OK");
+
+      return Status::OK();
+    }
+  }
+
+ private:
+  std::string subcommand_;
+  std::string stream_name_;
+  std::string group_name_;
+  StreamXGroupCreateOptions xgroup_create_options_;
+};
+
 class CommandXLen : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
@@ -969,6 +1041,7 @@ class CommandXSetId : public Commander {
 
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandXAdd>("xadd", -5, "write", 1, 1, 1),
                         MakeCmdAttr<CommandXDel>("xdel", -3, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandXGroup>("xgroup", -4, "write", 2, 2, 1),
                         MakeCmdAttr<CommandXLen>("xlen", -2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandXInfo>("xinfo", -2, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandXRange>("xrange", -4, "read-only", 1, 1, 1),
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -213,9 +213,6 @@ class CommandXGroup : public Commander {
     subcommand_ = util::ToLower(GET_OR_RET(parser.TakeStr()));
     stream_name_ = GET_OR_RET(parser.TakeStr());
     group_name_ = GET_OR_RET(parser.TakeStr());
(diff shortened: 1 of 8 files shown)


Yangsx-1:
I'll add more test cases after finishing xinfo command.

src/storage/redis_metadata.cc
```diff
@@ -378,7 +379,7 @@ rocksdb::Status StreamMetadata::Decode(const std::string &bytes) {
   GetFixed64(&input, &version);
   GetFixedCommon(&input, &size);
 
-  if (input.size() < 88) {
+  if (input.size() < 96) {
```
PragmaTwice: It seems this line will break the compatibility of kvrocks in stream datatype, which is critical.

PragmaTwice: i.e., the old userdata of stream datatype will always not satisfy this condition.

Yangsx-1: Oh, i don't recognize it. The variable `group_number` is added to `StreamMetadata` and it serves for quickly getting the consumer group number which will be used in `XINFO STREAM`. I can delete it if necessary and the number of consumer groups can be counted through iteration.

PragmaTwice: I mean, you can add some new field to StreamMetadata, but you cannot make old data in user's storage inaccessible.

So if you want to add a new field, you need to consider to write a Decode function that can work well while facing old userdata without the field you just add.

Yangsx-1: Got it. :)

src/commands/cmd_stream.cc
```diff
@@ -205,6 +205,97 @@ class CommandXDel : public Commander {
...
+    }
+
+    if (subcommand_ == "create") {
+      if (args.size() < 5 || args.size() > 8) {
+        return {Status::RedisParseErr, errWrongNumOfArguments};
+      }
+
+      xgroup_create_options_.last_id = args[4];
+
+      for (size_t i = 5; i < args.size(); ++i) {
```
git-hulk: it'd be better to use `command_parser` to parse arguments.

src/storage/redis_metadata.cc
```diff
@@ -396,5 +397,9 @@ rocksdb::Status StreamMetadata::Decode(Slice input) {
 
   GetFixed64(&input, &entries_added);
 
+  if (input.size() != 0) {
```
```suggestion
  if (input.size() >= 8) {
```

src/commands/cmd_stream.cc
```diff
@@ -205,6 +206,92 @@ class CommandXDel : public Commander {
...
+class CommandXGroup : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 1);
+    subcommand_ = util::ToLower(GET_OR_RET(parser.TakeStr()));
+    stream_name_ = GET_OR_RET(parser.TakeStr());
+    group_name_ = GET_OR_RET(parser.TakeStr());
+    if (std::isdigit(group_name_[0])) {
+      return {Status::RedisParseErr, "group name cannot start with number"};
+    }
```
PragmaTwice: I think this piece of logic can be better in Stream::CreateGroup?

src/commands/cmd_stream.cc
```diff
@@ -205,6 +206,92 @@ class CommandXDel : public Commander {
...
+          }
+          auto parse_result = parser.TakeInt<int64_t>();
+          if (!parse_result.IsOK()) {
+            return {Status::RedisParseErr, errValueNotInteger};
+          }
+          if (parse_result.GetValue() < 0 && parse_result.GetValue() != -1) {
+            return {Status::RedisParseErr, "value for ENTRIESREAD must be positive or -1"};
+          }
+          xgroup_create_options_.entries_read = parse_result.GetValue();
+        }
```
PragmaTwice, line 238:
```suggestion
        } else {
          return parser.InvalidSyntax();
        }
```

src/commands/cmd_stream.cc
```diff
@@ -205,6 +206,92 @@ class CommandXDel : public Commander {
...
+      xgroup_create_options_.last_id = GET_OR_RET(parser.TakeStr());
+
+      while (parser.Good()) {
+        if (parser.EatEqICase("mkstream")) {
+          xgroup_create_options_.mkstream = true;
+        } else if (parser.EatEqICase("entriesread")) {
+          if (!parser.Good()) {
+            return {Status::RedisParseErr, errUnknownSubcommandOrWrongArguments};
+          }
+          auto parse_result = parser.TakeInt<int64_t>();
```
PragmaTwice:
```suggestion
          auto parse_result = parser.TakeInt<int64_t>();
```

src/commands/cmd_stream.cc
```diff
@@ -205,6 +206,92 @@ class CommandXDel : public Commander {
...
+          if (!parse_result.IsOK()) {
+            return {Status::RedisParseErr, errValueNotInteger};
+          }
+          if (parse_result.GetValue() < 0 && parse_result.GetValue() != -1) {
+            return {Status::RedisParseErr, "value for ENTRIESREAD must be positive or -1"};
```
PragmaTwice, line 233: Could the value be zero? The condition includes zero but the error message does not.

src/commands/cmd_stream.cc
Yangsx-1: Oh yes, it's better in Stream::CreateGroup.

src/commands/cmd_stream.cc
Yangsx-1: The value can be zero, and the error message is the same as redis.

git-hulk:
@Yangsx-1 Could you please add the Go tests for this command?

Yangsx-1:
> @Yangsx-1 Could you please add the Go tests for this command?

Of course.

src/types/redis_stream.cc
```diff
@@ -160,6 +163,142 @@ rocksdb::Status Stream::Add(const Slice &stream_name, const StreamAddOptions &op
...
+  StreamMetadata metadata;
+  rocksdb::Status s = GetMetadata(ns_key, &metadata);
+  if (!s.ok() && !s.IsNotFound()) {
+    return s;
+  }
+
+  if (s.IsNotFound() && !options.mkstream) {
+    return rocksdb::Status::InvalidArgument(errXGroupSubcommandRequiresKeyExist);
+  }
+
```
mapleFU, line 231: Is `mkstream` implemented? We'd better add a test for it

Yangsx-1: Of course, the code below will use the empty metadata for the key of stream if it doesn't exist.

Yangsx-1: I add xinfo to test if the stream has been created in go test.

git-hulk (dismissed):

src/commands/cmd_stream.cc
```diff
@@ -205,6 +206,88 @@ class CommandXDel : public Commander {
...
+    }
+
+    if (subcommand_ == "destroy") {
+      uint64_t delete_cnt = 0;
+      auto s = stream_db.DestroyGroup(stream_name_, group_name_, &delete_cnt);
+      if (!s.ok()) {
+        return {Status::RedisExecErr, s.ToString()};
+      }
+
+      if (delete_cnt > 0) {
```
torwig, line 274: Is there any circumstances when `delete_cnt` can be greater than 1? If not we can simplify this `if-else` statement and just write `*output = redis::Integer(delete_cnt);`

enjoy-binbin:
@Yangsx-1 Can you or someone extract some content from #1586 and #1654 and put it in the top comment?
since there are some comments in the discussion, that will be great if we have a final version in the top comment

src/types/redis_stream.cc
```diff
@@ -160,6 +163,142 @@ rocksdb::Status Stream::Add(const Slice &stream_name, const StreamAddOptions &op
...
+  rocksdb::ReadOptions read_options = storage_->DefaultScanOptions();
+  LatestSnapShot ss(storage_);
+  read_options.snapshot = ss.GetSnapShot();
+  rocksdb::Slice upper_bound(next_version_prefix_key);
+  read_options.iterate_upper_bound = &upper_bound;
+  rocksdb::Slice lower_bound(prefix_key);
+  read_options.iterate_lower_bound = &lower_bound;
+
+  auto iter = util::UniqueIterator(storage_, read_options, stream_cf_handle_);
```
torwig, line 300: What actually we are deleting here using the loop and the iterator? Just CG's metadata? PEL and all the consumers?

Yangsx-1:
> @Yangsx-1 Can you or someone extract some content from #1586 and #1654 and put it in the top comment? since there are some comments in the discussion, that will be great if we have a final version in the top comment

added

src/commands/cmd_stream.cc
Yangsx-1: It can only be 0 or 1.

src/types/redis_stream.cc
Yangsx-1: Since all the data(consumer group, consumers in the group and PEL) for a single consumer group has the same prefix in the subkey, so all the data of the consumer group will be deleted through the iteration.

src/commands/cmd_stream.cc
mapleFU: (So a explicit boolean might be better here?)

src/types/redis_stream.cc
```diff
@@ -160,6 +163,142 @@ rocksdb::Status Stream::Add(const Slice &stream_name, const StreamAddOptions &op
...
+                                             const std::string &group_name) const {
+  std::string sub_key;
+  PutFixed64(&sub_key, group_name.size());
+  sub_key += group_name;
+  sub_key += "METADATA";
+  std::string entry_key = InternalKey(ns_key, sub_key, metadata.version, storage_->IsSlotIdEncoded()).Encode();
+  return entry_key;
+}
+
+std::string Stream::groupNameFromInternalKey(const rocksdb::Slice &key) const {
```
mapleFU:
```suggestion
std::string Stream::groupNameFromInternalKey(rocksdb::Slice key) const {
```

src/types/redis_stream.cc
```diff
@@ -160,6 +163,142 @@ rocksdb::Status Stream::Add(const Slice &stream_name, const StreamAddOptions &op
...
+  InternalKey ikey(key, storage_->IsSlotIdEncoded());
+  Slice group_name_metadata = ikey.GetSubKey();
+  uint64_t len = 0;
+  GetFixed64(&group_name_metadata, &len);
+  std::string group_name = group_name_metadata.ToString().substr(0, len);
```
mapleFU: substr before to string?

mapleFU (commented):
The patch is overall LGTM, I wonder should we add a test that we parse from / mkstream to previous stream metadata without group_number

src/types/redis_stream.cc
mapleFU: Should we check that `delete_cnt == 1`? Since L292-L297 would `-=1` rather than `-=*delete_cnt`

src/types/redis_stream.cc
Yangsx-1: Is there any difference? And it seems that class slice has no method the same as the function of substr.

src/types/redis_stream.cc
Yangsx-1: I think it's ok, because there are something deleted if `delete_cnt != 0`. When `delete_cnt == 1`, the deleted thing can only be consumer group metadata in the design. So the `-= 1` in L293 only means a consumer group are deleted and the counter in stream metadata changes.

torwig: @Yangsx-1 But you actually don't check the prefix inside the loop. Potentially, if there is any other data (that has another prefix), it will be deleted as well.

src/types/redis_stream.cc
mapleFU: Nope I think it could Produce less copy ( slice substr don't produce any copying ). But it doesn't matter.

src/types/redis_stream.cc
```diff
@@ -160,6 +163,142 @@ rocksdb::Status Stream::Add(const Slice &stream_name, const StreamAddOptions &op
   return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
 }
 
+std::string Stream::internalKeyFromGroupName(const std::string &ns_key, const StreamMetadata &metadata,
+                                             const std::string &group_name) const {
+  std::string sub_key;
+  PutFixed64(&sub_key, group_name.size());
+  sub_key += group_name;
+  sub_key += "METADATA";
```
torwig: I think this "magic" string can be extracted as a constant value.

src/types/redis_stream.cc
Yangsx-1: @torwig  In L271-L284, i used the subkey prefix to scan the snapshot. The prefix include the consumer groups name size and its name, so the scan result only includes the subkey with the prefix and there is no chance to delete something doesn't belong to the group.

Yangsx-1:
> The patch is overall LGTM, I wonder should we add a test that we parse from / mkstream to previous stream metadata without group_number

Sorry, i'm not fully understand what you want to check. @mapleFU

src/types/redis_stream.cc
Yangsx-1: You mean something like `bool isMetadata` in value? But it is easier to extract metadata directly with the magic string.

src/types/redis_stream.cc
mapleFU: Another point is that do we need to check whether the `length` is too long(check the result of `GetFixed64` pretent from bad len, and check `len` is not overflow)?

src/types/redis_stream.cc
torwig: No, I mean `constexpr const char *consumerGroupMetadataDelimiter = "METADATA"`.

Yangsx-1: Oh, that would be nice.

src/types/redis_stream.cc
Yangsx-1: Do you have more suggestions? @torwig

src/types/redis_stream.cc
```diff
@@ -160,6 +163,142 @@ rocksdb::Status Stream::Add(const Slice &stream_name, const StreamAddOptions &op
...
+    }
+  }
+  consumer_group_metadata.entries_read = options.entries_read;
+  std::string entry_key = internalKeyFromGroupName(ns_key, metadata, group_name);
+  std::string entry_value = encodeStreamConsumerGroupMetadataValue(consumer_group_metadata);
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisStream);
+  batch->PutLogData(log_data.Encode());
+  batch->Put(stream_cf_handle_, entry_key, entry_value);
```
mapleFU, line 258: One other question is that, should we check if the `entry_key` existing before?

https://github.com/redis/redis/blob/e792653753dc62b5a00822121e585511542a024b/src/t_stream.c#L2484

mapleFU:
Sorry for misleading, I mean maybe we can have some "upgrading test", since now we evolution our Stream Metadata. Users in 2.5.1 using stream would have shorter metadata. Maybe we need to test create group on an existing stream without `group_number` (or just not write it if it's 0).

Or don't test that is also ok for me, since the logic here is not complex. I guess we might not make error on it.

src/types/redis_stream.cc
Yangsx-1: Good catch! I think it's necessary and i'll add it later.

mapleFU:
(Maybe not write `group_number` if it's 0 is not a good idea. Because if we support more config later, we will lack a flag shows that what's the logic content after 88 bytes.

It's ok to not testing it. If we want test it, just `Encode` and build a subslice for it. )

src/types/redis_stream.cc
Yangsx-1: Maybe `uint64_t` is big enough and we don't need to check it.

mapleFU: Oh I think we won't overflow `u64`. It's a kind of defensive programming, since maybe internal key passed to `groupNameFromInternalKey` is kind of rubbish, it might to short or given a bad `length`. And `substr` it might causing a out_of_range exception.

It might not happen if everything works well, and if something bad happens, we still suffer from error or bug. However I think adding a `Status` and checking here is no risky

mapleFU (approved):
I'm not so familiar with STREAM logic, might need @torwig 's review

Other LGTM

torwig:
@mapleFU Generally LGTM/ However, I think that CREATE and DESTROY without other CG commands are useless. Thus, I would like to see the implementation of READ, XCLAIM, XPENDING commands. Additionally, during the work on them, @Yangsx-1 may want to improve something in the CREATE command's implementation.

Yangsx-1:
> 

Yes, i'll participate in most implementation of the consumer group commands and improve the already finished commands when necessary.

Yangsx-1:
Can anyone help merge this pr?

git-hulk (approved):

PR #1705, opened 2023-08-28
Fix RESTORE to create a key without expire under the same ms

enjoy-binbin:
I am unable to reproduce it, but if the ttl_ms (absttl) passed in
happens to be in the same ms as GetTimeStampMS, the `ttl_ms < now`
will be false and `ttl_ms -= now` will make ttl_ms become 0 and then
we will create a key without any expire. It is off by one bug.


git-hulk (approved):
Good catch, thank you!

mapleFU (approved):
Great catch!

PR #1708, opened 2023-08-29
Remove the no-script flag in the LASTSAVE command

enjoy-binbin:
Redis allow run it in the script:
```
127.0.0.1:6379> eval "return redis.call('lastsave')" 0
(integer) 1693277322
```

Look like a copy-paste bug.


enjoy-binbin:
another question, now run the no-script flag command will response this error:
```
127.0.0.1:6666> eval "return redis.call('lastsave')" 0
(error) ERR running script (call to f_d096f7c74e67ca72b33cecd3eb313f2d7732c9e6): This Redis command is not allowed from scripts
```

the error message mention `This Redis command`, maybe in the future we need to change this. since we may add some kvrocks-only commands?

PragmaTwice (approved):

git-hulk (approved):

PR #1710, opened 2023-08-29
Add the support of the bloom BF.INFO command

zncleon:
This PR introduce the bloom BF.INFO command. 

[1] [https://redis.io/commands/bf.info/](https://redis.io/commands/bf.info/)

### DIFF
--- a/src/commands/cmd_bloom_filter.cc
+++ b/src/commands/cmd_bloom_filter.cc
@@ -119,15 +119,15 @@ class CommandBFInfo : public Commander {
     CommandParser parser(args, 2);
     while (parser.Good()) {
       if (parser.EatEqICase("capacity")) {
-        type_ = BloomInfoType::CAPACITY;
+        type_ = BloomInfoType::kCapacity;
       } else if (parser.EatEqICase("size")) {
-        type_ = BloomInfoType::SIZE;
+        type_ = BloomInfoType::kSize;
       } else if (parser.EatEqICase("filters")) {
-        type_ = BloomInfoType::FILTERS;
+        type_ = BloomInfoType::kFilters;
       } else if (parser.EatEqICase("items")) {
-        type_ = BloomInfoType::ITEMS;
+        type_ = BloomInfoType::kItems;
       } else if (parser.EatEqICase("expansion")) {
-        type_ = BloomInfoType::EXPANSION;
+        type_ = BloomInfoType::kExpansion;
       } else {
         return {Status::RedisParseErr, "Invalid info argument"};
       }
@@ -144,7 +144,7 @@ class CommandBFInfo : public Commander {
     if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
 
     switch (type_) {
-      case BloomInfoType::ALL:
+      case BloomInfoType::kAll:
         *output = "*" + std::to_string(2 * 5) + CRLF;
         *output += redis::SimpleString("Capacity");
         *output += redis::Integer(info.capacity);
@@ -157,19 +157,19 @@ class CommandBFInfo : public Commander {
         *output += redis::SimpleString("Expansion rate");
         *output += redis::Integer(info.expansion);
         break;
-      case BloomInfoType::CAPACITY:
+      case BloomInfoType::kCapacity:
         *output = redis::Integer(info.capacity);
         break;
-      case BloomInfoType::SIZE:
+      case BloomInfoType::kSize:
         *output = redis::Integer(info.bloom_bytes);
         break;
-      case BloomInfoType::FILTERS:
+      case BloomInfoType::kFilters:
         *output = redis::Integer(info.n_filters);
         break;
-      case BloomInfoType::ITEMS:
+      case BloomInfoType::kItems:
         *output = redis::Integer(info.size);
         break;
-      case BloomInfoType::EXPANSION:
+      case BloomInfoType::kExpansion:
         *output = redis::Integer(info.expansion);
         break;
       default:
@@ -180,7 +180,7 @@ class CommandBFInfo : public Commander {
   }
 
  private:
-  BloomInfoType type_ = BloomInfoType::ALL;
+  BloomInfoType type_ = BloomInfoType::kAll;
 };
 
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandBFReserve>("bf.reserve", -4, "write", 1, 1, 1),
--- a/src/commands/cmd_bloom_filter.cc
+++ b/src/commands/cmd_bloom_filter.cc
@@ -106,7 +106,6 @@ class CommandBFExists : public Commander {
     redis::BloomChain bloom_db(svr->storage, conn->GetNamespace());
     int ret = 0;
     auto s = bloom_db.Exist(args_[1], args_[2], &ret);
-    if (s.IsNotFound()) return {Status::RedisExecErr, "key is not found"};
     if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
 
     *output = redis::Integer(ret);
--- a/src/commands/cmd_bloom_filter.cc
+++ b/src/commands/cmd_bloom_filter.cc
@@ -118,15 +118,15 @@ class CommandBFInfo : public Commander {
     CommandParser parser(args, 2);
     while (parser.Good()) {
       if (parser.EatEqICase("capacity")) {
-        type_ = CAPACITY;
+        type_ = BloomInfoType::CAPACITY;
       } else if (parser.EatEqICase("size")) {
-        type_ = SIZE;
+        type_ = BloomInfoType::SIZE;
       } else if (parser.EatEqICase("filters")) {
-        type_ = FILTERS;
+        type_ = BloomInfoType::FILTERS;
       } else if (parser.EatEqICase("items")) {
-        type_ = ITEMS;
+        type_ = BloomInfoType::ITEMS;
       } else if (parser.EatEqICase("expansion")) {
-        type_ = EXPANSION;
+        type_ = BloomInfoType::EXPANSION;
       } else {
         return {Status::RedisParseErr, "Invalid info argument"};
       }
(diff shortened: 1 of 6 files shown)


src/types/redis_bloom_chain.h
```diff
@@ -30,12 +30,22 @@ const uint32_t kBFDefaultInitCapacity = 100;
 const double kBFDefaultErrorRate = 0.01;
 const uint16_t kBFDefaultExpansion = 2;
 
+enum BloomInfoType {
```
PragmaTwice:
```suggestion
enum class BloomInfoType {
```

src/types/redis_bloom_chain.cc
```diff
@@ -195,4 +195,44 @@ rocksdb::Status BloomChain::Exist(const Slice &user_key, const Slice &item, int
...
+      break;
+    case FILTERS:
+      rets->push_back(static_cast<int>(metadata.n_filters));
+      break;
+    case ITEMS:
+      rets->push_back(static_cast<int>(metadata.size));
+      break;
+    case EXPANSION:
+      rets->push_back(static_cast<int>(metadata.expansion));
+      break;
```
PragmaTwice: I think maybe we can just return a structure here, e.g.
```
struct BloomFilterInfo {
    cap, bytes, ...;
};

Info(..., BloomFilterInfo *info) {
   ...
   info->... = ...;
}
```

And move the switch case to the command execute function.

tests/gocase/unit/type/bloom/bloom_test.go
```diff
@@ -116,4 +116,44 @@ func TestBloom(t *testing.T) {
 		}
 		require.LessOrEqual(t, float64(exist), fpp*float64(totalCount))
 	})
+
```
mapleFU, line 147: Apart from this, we can use "INFO" to test the reserve and add

zncleon: Sure, there are some bugs about add.

src/types/redis_bloom_chain.cc
```diff
@@ -151,6 +151,7 @@ rocksdb::Status BloomChain::Add(const Slice &user_key, const Slice &item, int *r
     s = bloomCheck(bf_key_list[check_index], item_string, ret);
```
git-hulk: I think we should add another variable to identify the check result(e.g. exists or check_ret) instead of resuing the add result.

src/types/redis_bloom_chain.cc
```diff
@@ -143,21 +143,21 @@ rocksdb::Status BloomChain::Add(const Slice &user_key, const Slice &item, int *r
...
 
-  int check_index = metadata.n_filters - 1;
   std::string item_string = item.ToString();
 
   // check
-  for (; check_index >= 0; --check_index) {  // TODO: to test which direction for searching is better
-    s = bloomCheck(bf_key_list[check_index], item_string, ret);
+  int check_ret = 0;
+  for (int i = metadata.n_filters - 1; i >= 0; --i) {  // TODO: to test which direction for searching is better
+    s = bloomCheck(bf_key_list[i], item_string, &check_ret);
```
mapleFU: Actually I think a `Result<bool>` might be better here. Since "int" might be weird ( RedisBloom using C language, C Boolean is just a hacking )

src/types/redis_bloom_chain.cc
```diff
@@ -185,14 +178,35 @@ rocksdb::Status BloomChain::Exist(const Slice &user_key, const Slice &item, int
...
   return rocksdb::Status::OK();
 }
+
+rocksdb::Status BloomChain::Info(const Slice &user_key, BloomFilterInfo *info) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+
+  BloomChainMetadata metadata;
+  rocksdb::Status s = getBloomChainMetadata(ns_key, &metadata);
+  if (s.IsNotFound()) return rocksdb::Status::NotFound("key is not found");
```
torwig: Why not simply use `if (!s.ok()) return s;` that will return the `rocksdb::Status::NotFound` status as well? Then check on the caller side if the status is NotFound if necessary.

src/types/redis_bloom_chain.h
```diff
@@ -30,20 +30,38 @@ const uint32_t kBFDefaultInitCapacity = 100;
 const double kBFDefaultErrorRate = 0.01;
 const uint16_t kBFDefaultExpansion = 2;
 
+enum class BloomInfoType {
+  ALL,
```
torwig: You can format enum values as `kAll, kCapacity, etc` (see here: https://google.github.io/styleguide/cppguide.html#Enumerator_Names)

src/types/redis_bloom_chain.cc
zncleon: Sure, It's a mistake left by solution 1. I will fix it.

src/commands/cmd_bloom_filter.cc
```diff
@@ -104,14 +105,86 @@ class CommandBFExists : public Commander {
...
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::BloomChain bloom_db(svr->storage, conn->GetNamespace());
+    BloomFilterInfo info;
+    auto s = bloom_db.Info(args_[1], &info);
+    if (s.IsNotFound()) return {Status::RedisExecErr, "key is not found"};
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    switch (type_) {
+      case BloomInfoType::kAll:
+        *output = "*" + std::to_string(2 * 5) + CRLF;
```
PragmaTwice:
```suggestion
        *output = redis::MultiLen(2 * 5);
```

src/commands/cmd_bloom_filter.cc
```diff
@@ -104,14 +105,86 @@ class CommandBFExists : public Commander {
...
 
     *output = redis::Integer(ret);
     return Status::OK();
   }
 };
 
+class CommandBFInfo : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 2);
```
mapleFU, line 122: We'd better `args` should not greater than 3?

https://github.com/RedisBloom/RedisBloom/blob/master/src/rebloom.c#L949

src/commands/cmd_bloom_filter.cc
```diff
@@ -104,14 +105,86 @@ class CommandBFExists : public Commander {
...
+        *output += redis::SimpleString("Capacity");
+        *output += redis::Integer(info.capacity);
+        *output += redis::SimpleString("Size");
+        *output += redis::Integer(info.bloom_bytes);
+        *output += redis::SimpleString("Number of filters");
+        *output += redis::Integer(info.n_filters);
+        *output += redis::SimpleString("Number of items inserted");
+        *output += redis::Integer(info.size);
+        *output += redis::SimpleString("Expansion rate");
+        *output += redis::Integer(info.expansion);
```
mapleFU: https://github.com/RedisBloom/RedisBloom/blob/master/src/rebloom.c#L990

Since `info.expansion` might be null if it's 0?

src/commands/cmd_bloom_filter.cc
```diff
@@ -104,14 +105,86 @@ class CommandBFExists : public Commander {
...
     *output = redis::Integer(ret);
     return Status::OK();
   }
 };
 
+class CommandBFInfo : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 2);
+    while (parser.Good()) {
```
mapleFU:
```suggestion
    if (parser.Good()) {
```

src/commands/cmd_bloom_filter.cc
```diff
@@ -104,14 +105,86 @@ class CommandBFExists : public Commander {
...
+        *output = redis::Integer(info.n_filters);
+        break;
+      case BloomInfoType::kItems:
+        *output = redis::Integer(info.size);
+        break;
+      case BloomInfoType::kExpansion:
+        *output = redis::Integer(info.expansion);
+        break;
+      default:
+        LOG(ERROR) << "Failed to parse the type of BF.INFO command";
```
mapleFU: return error here?

src/types/redis_bloom_chain.cc
```diff
@@ -177,22 +168,41 @@ rocksdb::Status BloomChain::Exist(const Slice &user_key, const Slice &item, int
...
     }
   }
+  *ret = exist ? 1 : 0;
 
   return rocksdb::Status::OK();
 }
+
+rocksdb::Status BloomChain::Info(const Slice &user_key, BloomFilterInfo *info) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+  LockGuard guard(storage_->GetLockManager(), ns_key);
```
mapleFU: `Info` doesn't need to lock since it just atomicly get the metadata?

mapleFU (commented):
Rest LGTM

src/commands/cmd_bloom_filter.cc
zncleon: If I add a return here, lint would prompt an error: "Unreachable code ".

mapleFU: Ah yeah, we can just remove this branch.

mapleFU (dismissed):

torwig (dismissed):
LGTM

torwig:
@zncleon BTW, according to the Redis documentation, the BF.EXISTS command returns 0 if the key doesn't exist. But I see that currently, the code returns an error `{Status::RedisExecErr, "key is not found"}`. Is it intentional?

zncleon:
> @zncleon BTW, according to the Redis documentation, the BF.EXISTS command returns 0 if the key doesn't exist. But I see that currently, the code returns an error `{Status::RedisExecErr, "key is not found"}`. Is it intentional?

Sure, I think I made a mistake. It should return 0 in this case. I will fix it.

mapleFU (approved):

torwig (approved):

git-hulk (approved):

mapleFU:
Thanks all, merging

PR #1712, opened 2023-08-30
Introduce MDel do the batch delete and use MultiLockGuard guarantee atomicity

### LINKED ISSUE #1692: Kvrocks can't guarantee atomicity for multiple keys commands

enjoy-binbin:
Previously, the DEL/UNLINK commands looped through each key to
delete a single key. We need to change to batch delete (write),
and also the original implementation did not lock the keys and
will break atomicity.

This PR introduces a new Mdel function, which uses MultiLockGuard
to lock all the keys to guarantee atomicity, and use batch to do
the final delete (write).

Now it will be only used in DEL / UNLINKE commands. Noticed that
our DEL / UNLINK test coverage was not sufficient, added some simple
coverage tests to cover it.

Fixes for DEL / UNLINK issues in #1692.

unrelated change: in Exists for loop Get, return an execution error
when `!s.ok() && !s.IsNotFound()` without continuing the operation
(because it will be some internal RocksDB error).

### DIFF
--- a/src/commands/cmd_key.cc
+++ b/src/commands/cmd_key.cc
@@ -99,6 +99,7 @@ class CommandExists : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     std::vector<rocksdb::Slice> keys;
+    keys.reserve(args_.size() - 1);
     for (size_t i = 1; i < args_.size(); i++) {
       keys.emplace_back(args_[i]);
     }
@@ -237,6 +238,7 @@ class CommandDel : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     std::vector<rocksdb::Slice> keys;
+    keys.reserve(args_.size() - 1);
     for (size_t i = 1; i < args_.size(); i++) {
       keys.emplace_back(args_[i]);
     }
--- a/src/commands/cmd_key.cc
+++ b/src/commands/cmd_key.cc
@@ -45,6 +45,24 @@ class CommandType : public Commander {
   }
 };
 
+class CommandMove : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    GET_OR_RET(ParseInt<int64_t>(args[2], 10));
+    return Status::OK();
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    int count = 0;
+    redis::Database redis(svr->storage, conn->GetNamespace());
+    rocksdb::Status s = redis.Exists({args_[1]}, &count);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = count ? redis::Integer(1) : redis::Integer(0);
+    return Status::OK();
+  }
+};
+
 class CommandObject : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
@@ -256,6 +274,7 @@ class CommandDel : public Commander {
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandTTL>("ttl", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandPTTL>("pttl", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandType>("type", 2, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandMove>("move", 3, "write", 1, 1, 1),
                         MakeCmdAttr<CommandObject>("object", 3, "read-only", 2, 2, 1),
                         MakeCmdAttr<CommandExists>("exists", -2, "read-only", 1, -1, 1),
                         MakeCmdAttr<CommandPersist>("persist", 2, "write", 1, 1, 1),
--- a/src/commands/cmd_key.cc
+++ b/src/commands/cmd_key.cc
@@ -236,12 +236,16 @@ class CommandPersist : public Commander {
 class CommandDel : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    int cnt = 0;
-    redis::Database redis(svr->storage, conn->GetNamespace());
+    std::vector<rocksdb::Slice> keys;
     for (size_t i = 1; i < args_.size(); i++) {
-      auto s = redis.Del(args_[i]);
-      if (s.ok()) cnt++;
+      keys.emplace_back(args_[i]);
     }
+
+    uint64_t cnt = 0;
+    redis::Database redis(svr->storage, conn->GetNamespace());
+    auto s = redis.MDel(keys, &cnt);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
     *output = redis::Integer(cnt);
     return Status::OK();
   }
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -152,7 +152,7 @@ rocksdb::Status Database::MDel(const std::vector<Slice> &keys, uint64_t *deleted
 
   for (size_t i = 0; i < slice_keys.size(); i++) {
     if (!statuses[i].ok() && !statuses[i].IsNotFound()) return statuses[i];
-    if (!statuses[i].IsNotFound()) continue;
+    if (statuses[i].IsNotFound()) continue;
 
     Metadata metadata(kRedisNone, false);
     metadata.Decode(pin_values[i]);
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -136,16 +136,28 @@ rocksdb::Status Database::MDel(const std::vector<Slice> &keys, uint64_t *deleted
   WriteBatchLogData log_data(kRedisNone);
   batch->PutLogData(log_data.Encode());
 
+  std::vector<Slice> slice_keys;
+  slice_keys.reserve(lock_keys.size());
   for (const auto &ns_key : lock_keys) {
-    std::string value;
-    rocksdb::Status s = storage_->Get(rocksdb::ReadOptions(), metadata_cf_handle_, ns_key, &value);
-    if (!s.ok()) continue;
+    slice_keys.emplace_back(ns_key);
+  }
+
+  LatestSnapShot ss(storage_);
+  rocksdb::ReadOptions read_options = storage_->DefaultMultiGetOptions();
+  read_options.snapshot = ss.GetSnapShot();
+  std::vector<rocksdb::Status> statuses(slice_keys.size());
+  std::vector<rocksdb::PinnableSlice> pin_values(slice_keys.size());
+  storage_->MultiGet(read_options, metadata_cf_handle_, slice_keys.size(), slice_keys.data(), pin_values.data(),
+                     statuses.data());
+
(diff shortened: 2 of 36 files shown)


src/commands/cmd_key.cc
```diff
@@ -236,12 +236,16 @@ class CommandPersist : public Commander {
 class CommandDel : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
-    int cnt = 0;
-    redis::Database redis(svr->storage, conn->GetNamespace());
+    std::vector<rocksdb::Slice> keys;
```
torwig, line 258: Could we call `reserve` here for the vector before adding the known number of elements to it?

torwig: Others LGTM.

enjoy-binbin: thanks, added

torwig (dismissed):

git-hulk:
We may have the deadlock in ZSet::InterStore(as well as other similar APIs) if the dst key was inside the source key list since the OverWrite function will also lock the dst key: https://github.com/apache/kvrocks/blob/unstable/src/types/redis_zset.cc#L635 

This issue was found by TSAN in CI: https://github.com/apache/kvrocks/actions/runs/6022265431/job/16338859035?pr=1712

torwig:
@git-hulk Good point.

torwig:
I guess we should lock all the keys including the destination inside the `ZSet::InterStore` and similar methods (right now we lock the destination inside the `Overwite` method). Will it work?

git-hulk:
> I guess we should lock all the keys including the destination inside the `ZSet::InterStore` and similar methods (right now we lock the destination inside the `Overwite` method). Will it work?

Yes, I think it can resolve this issue.

git-hulk:
I submit another issue to track this since it's related but NOT introduced by this PR. https://github.com/apache/kvrocks/issues/1715

enjoy-binbin:
Or can it be ok without locking the dst key? since we are doing a overwrite

torwig:
@enjoy-binbin I think we need to serialize write operations on the key to avoid "lost updates". However, you made me rethink the previous sentence because we deal with overwriting :)

git-hulk:
Is it possible to lock the dst key only and use the read snapshot like SDIFFSTORE/SINTERSTORE?

enjoy-binbin:
> Is it possible to lock the dst key only and use the read snapshot like SDIFFSTORE/SINTERSTORE?

@git-hulk can you describe this a bit more? (for example posting a pseudo code)

git-hulk:
> > Is it possible to lock the dst key only and use the read snapshot like SDIFFSTORE/SINTERSTORE?
> 
> @git-hulk can you describe this a bit more? (for example posting a pseudo code)

Sure. For now, SDIFFSTORE is implemented in two steps:

- The first step is to calculate the diff results without locking source keys
- Then overwrite the dst key with locking the dst key

Please refer https://github.com/apache/kvrocks/blob/unstable/src/types/redis_set.cc#L433

enjoy-binbin:
actually now we will lock the source keys. it was added in #1700, maybe we should remove the source keys locks?

```
rocksdb::Status Set::Diff(const std::vector<Slice> &keys, std::vector<std::string> *members) {
  std::vector<std::string> lock_keys;
  lock_keys.reserve(keys.size());
  for (const auto key : keys) {
    std::string ns_key = AppendNamespacePrefix(key);
  }
  MultiLockGuard guard(storage_->GetLockManager(), lock_keys);
```

edit: ohh, i think i got your point. maybe we should split InterStore to two function, one do the inter and the other one do the store, just like the Set apis. I wanted to do it (adding a ZINTER command) before, but because we already had a pending #1517

git-hulk:
> > > Is it possible to lock the dst key only and use the read snapshot like SDIFFSTORE/SINTERSTORE?
> > 
> > 
> > @git-hulk can you describe this a bit more? (for example posting a pseudo code)
> 
> Sure. For now, SDIFFSTORE is implemented in two steps:
> 
> * The first step is to calculated the diff result without locking source keys
> * Then overwrite the dst key with locking the dst key



> actually now we will lock the source keys. it was added in #1700, maybe we should remove the source keys locks?
> 
> ```
> rocksdb::Status Set::Diff(const std::vector<Slice> &keys, std::vector<std::string> *members) {
>   std::vector<std::string> lock_keys;
>   lock_keys.reserve(keys.size());
>   for (const auto key : keys) {
>     std::string ns_key = AppendNamespacePrefix(key);
>     lock_keys.emplace_back(std::move(ns_key));
>   }
>   MultiLockGuard guard(storage_->GetLockManager(), lock_keys);
> ```
> 
> edit: ohh, i think i got your point. maybe we should split InterStore to two function, one do the inter and the other one do the store, just like the Set apis. I wanted to do it (adding a ZINTER command) before, but because we already had a pending #1517

Great, thank you!

enjoy-binbin:
i pushed a commit [c51975b](https://github.com/apache/kvrocks/pull/1712/commits/c51975b4e243705584f7ae1429d022aeee48c6b7)

split InterStore to two function, one do the inter and the other one do the store, just like the Set apis. to see if the CI can pass

enjoy-binbin:
i see the CI is passed, so the split change can fix the deadlock, i am going to submit another PR to make the change, am revert the change in this PR (after we merge the new PR see #1726, we can re-merge it)

PragmaTwice:
lock-order-inversion found in TSan

https://github.com/apache/kvrocks/actions/runs/6062050794/job/16447962572?pr=1712

git-hulk (dismissed):

src/storage/redis_db.cc
```diff
@@ -121,6 +121,39 @@ rocksdb::Status Database::Del(const Slice &user_key) {
...
+  MultiLockGuard guard(storage_->GetLockManager(), lock_keys);
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisNone);
+  batch->PutLogData(log_data.Encode());
+
+  for (const auto &ns_key : lock_keys) {
+    std::string value;
+    rocksdb::Status s = storage_->Get(rocksdb::ReadOptions(), metadata_cf_handle_, ns_key, &value);
+    if (!s.ok()) continue;
```
mapleFU: A bit confusion on this, what about the `!ok` and `!IsNotFound` status?

src/storage/redis_db.cc
```diff
@@ -121,6 +121,39 @@ rocksdb::Status Database::Del(const Slice &user_key) {
...
+  }
+  MultiLockGuard guard(storage_->GetLockManager(), lock_keys);
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisNone);
+  batch->PutLogData(log_data.Encode());
+
+  for (const auto &ns_key : lock_keys) {
+    std::string value;
+    rocksdb::Status s = storage_->Get(rocksdb::ReadOptions(), metadata_cf_handle_, ns_key, &value);
```
mapleFU: As an optimization, can `MultiGet` be used here? (maybe we can implement it in a separate patch)

mapleFU (dismissed):
General LGTM. Just some nits.

src/storage/redis_db.cc
torwig: I guess we can return an execution error in this case without continuing the operation (because it will be some internal RocksDB error).

src/storage/redis_db.cc
enjoy-binbin: look like a good idea, i take a shot, not sure i wrote it right, please take a look, thanks.

src/storage/redis_db.cc
```diff
@@ -121,6 +121,51 @@ rocksdb::Status Database::Del(const Slice &user_key) {
...
+  LatestSnapShot ss(storage_);
+  rocksdb::ReadOptions read_options = storage_->DefaultMultiGetOptions();
+  read_options.snapshot = ss.GetSnapShot();
+  std::vector<rocksdb::Status> statuses(slice_keys.size());
+  std::vector<rocksdb::PinnableSlice> pin_values(slice_keys.size());
+  storage_->MultiGet(read_options, metadata_cf_handle_, slice_keys.size(), slice_keys.data(), pin_values.data(),
+                     statuses.data());
+
+  for (size_t i = 0; i < slice_keys.size(); i++) {
+    if (!statuses[i].ok()) continue;
```
mapleFU: As mentioned previously, should we check the `!ok && !IsNotExists`

src/storage/redis_db.cc
```diff
@@ -121,6 +121,51 @@ rocksdb::Status Database::Del(const Slice &user_key) {
...
+
+    Metadata metadata(kRedisNone, false);
+    metadata.Decode(pin_values[i].ToString());
```
mapleFU: Not need `ToString`?

src/storage/redis_db.cc
enjoy-binbin: ohh, sorry, i missed this comment (thought it is ok to continue). yean, it is ok to return the error.

we do had some commands that will do the continue, like `EXISTS`. i will change it too.
```
rocksdb::Status Database::Exists(const std::vector<Slice> &keys, int *ret) {
  *ret = 0;
  LatestSnapShot ss(storage_);

  rocksdb::Status s;
  std::string value;
  for (const auto &key : keys) {
    std::string ns_key = AppendNamespacePrefix(key);
    s = storage_->Get(read_options, metadata_cf_handle_, ns_key, &value);
    if (s.ok()) {
      metadata.Decode(value);
      if (!metadata.Expired()) *ret += 1;
    }
  }
  return rocksdb::Status::OK();
}
```

src/storage/redis_db.cc
```diff
@@ -121,6 +121,52 @@ rocksdb::Status Database::Del(const Slice &user_key) {
...
+  rocksdb::ReadOptions read_options = storage_->DefaultMultiGetOptions();
+  read_options.snapshot = ss.GetSnapShot();
+  std::vector<rocksdb::Status> statuses(slice_keys.size());
+  std::vector<rocksdb::PinnableSlice> pin_values(slice_keys.size());
+  storage_->MultiGet(read_options, metadata_cf_handle_, slice_keys.size(), slice_keys.data(), pin_values.data(),
+                     statuses.data());
+
+  for (size_t i = 0; i < slice_keys.size(); i++) {
+    if (!statuses[i].ok() && !statuses[i].IsNotFound()) return statuses[i];
+    if (!statuses[i].IsNotFound()) continue;
```
mapleFU:
```suggestion
    if (statuses[i].IsNotFound()) continue;
```

mapleFU (approved):

torwig (approved):

git-hulk (approved):

PR #1716, opened 2023-08-31
Replace sha512sum with shasum

tisonkun:
sha512sum is part of GNU coreutils and shasum is a Perl script. macOS doesn't have sha512sum out of the box and shasum should have better portability.


git-hulk (approved):

tisonkun:
Or even we should depend on Python builtin so that it works for Windows as well..

I don't do an investigation, though.

PR #1721, opened 2023-08-31
Add the support of SCALING for bloom filter

zncleon:
This PR supports the "SCALING" of bloom filter. When the bloom filter is full, it will add a new bloom filter into bloom chain if is "SCALING"，otherwise it will return an error.

### DIFF
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -245,4 +245,9 @@ class BloomChainMetadata : public Metadata {
   ///
   /// @return the total capacity value
   uint32_t GetCapacity() const;
+
+  /// Check the bloom chain is scaling or not
+  ///
+  /// @return true if scaling, false if nonscaling
+  bool IsScaling() const;
 };
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -241,13 +241,7 @@ class BloomChainMetadata : public Metadata {
   void Encode(std::string *dst) override;
   rocksdb::Status Decode(Slice bytes) override;
 
-  /// Get the total capacity of the bloom chain (the sum capacity of all sub-filters)
-  ///
-  /// @return the total capacity value
   uint32_t GetCapacity() const;
 
-  /// Check the bloom chain is scaling or not
-  ///
-  /// @return true if scaling, false if nonscaling
-  bool IsScaling() const;
+  bool IsScaling() const { return expansion != 0; };
 };
--- a/src/types/bloom_filter.h
+++ b/src/types/bloom_filter.h
@@ -154,7 +154,9 @@ class BlockSplitBloomFilter {
   /// Get the plain bitset value from the Bloom filter bitset.
   ///
   /// @return bitset value;
-  const std::string& GetData() { return data_; }
+  const std::string& GetData() & { return data_; }
+
+  std::string GetData() && { return data_; }
 
   /// Compute hash for string value by using its plain encoding result.
   ///
--- a/src/types/bloom_filter.h
+++ b/src/types/bloom_filter.h
@@ -154,9 +154,9 @@ class BlockSplitBloomFilter {
   /// Get the plain bitset value from the Bloom filter bitset.
   ///
   /// @return bitset value;
-  const std::string& GetData() & { return data_; }
+  const std::string& GetData() const& { return data_; }
 
-  std::string GetData() && { return data_; }
+  std::string GetData() const&& { return data_; }
 
   /// Compute hash for string value by using its plain encoding result.
   ///
--- a/src/types/bloom_filter.h
+++ b/src/types/bloom_filter.h
@@ -156,7 +156,7 @@ class BlockSplitBloomFilter {
   /// @return bitset value;
   const std::string& GetData() const& { return data_; }
 
-  std::string GetData() const&& { return data_; }
+  std::string&& GetData() && { return std::move(data_); }
 
   /// Compute hash for string value by using its plain encoding result.
   ///
--- a/src/types/redis_bloom_chain.cc
+++ b/src/types/redis_bloom_chain.cc
@@ -79,7 +79,7 @@ void BloomChain::createBloomFilterInBatch(const Slice &ns_key, BloomChainMetadat
 
   BlockSplitBloomFilter block_split_bloom_filter;
   block_split_bloom_filter.Init(bloom_filter_bytes);
-  *bf_data = block_split_bloom_filter.GetData();
+  *bf_data = std::move(block_split_bloom_filter).GetData();
 
   std::string bloom_chain_meta_bytes;
   metadata->Encode(&bloom_chain_meta_bytes);
@@ -92,7 +92,7 @@ void BloomChain::bloomAdd(const std::string &item, std::string *bf_data) {
 
   uint64_t h = BlockSplitBloomFilter::Hash(item.data(), item.size());
   block_split_bloom_filter.InsertHash(h);
-  *bf_data = block_split_bloom_filter.GetData();
+  *bf_data = std::move(block_split_bloom_filter).GetData();
 }
 
 rocksdb::Status BloomChain::bloomCheck(const Slice &bf_key, const std::string &item, bool *exist) {
--- a/src/types/redis_bloom_chain.cc
+++ b/src/types/redis_bloom_chain.cc
@@ -174,6 +174,7 @@ rocksdb::Status BloomChain::Add(const Slice &user_key, const Slice &item, int *r
     if (metadata.size + 1 > metadata.GetCapacity()) {
       if (metadata.IsScaling()) {
         s = createBloomFilter(ns_key, &metadata);
+        if (!s.ok()) return s;
       } else {
         return rocksdb::Status::Aborted("filter is full and is nonscaling");
       }
--- a/src/types/redis_bloom_chain.cc
+++ b/src/types/redis_bloom_chain.cc
@@ -69,7 +69,8 @@ rocksdb::Status BloomChain::createBloomChain(const Slice &ns_key, double error_r
   return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
 }
 
-rocksdb::Status BloomChain::createBloomFilter(const Slice &ns_key, BloomChainMetadata *metadata) {
+void BloomChain::createBloomFilterInBatch(const Slice &ns_key, BloomChainMetadata *metadata,
+                                          ObserverOrUniquePtr<rocksdb::WriteBatchBase> &batch, std::string *bf_data) {
   uint32_t bloom_filter_bytes = BlockSplitBloomFilter::OptimalNumOfBytes(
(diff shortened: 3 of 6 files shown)


src/storage/redis_metadata.h
```diff
@@ -245,4 +245,9 @@ class BloomChainMetadata : public Metadata {
   ///
   /// @return the total capacity value
   uint32_t GetCapacity() const;
+
+  /// Check the bloom chain is scaling or not
```
torwig: I think it's needless to comment on simple methods such as `IsScaling` or `GetCapacity` - they have self-explanatory names. BTW, the implementation of `IsScaling` is so short/simple that you can write it in the header file.

src/types/redis_bloom_chain.cc
```diff
@@ -147,18 +171,22 @@ rocksdb::Status BloomChain::Add(const Slice &user_key, const Slice &item, int *r
 
   // insert
   if (!exist) {
-    if (metadata.size + 1 > metadata.GetCapacity()) {  // TODO: scaling would be supported later
-      return rocksdb::Status::Aborted("filter is full");
+    if (metadata.size + 1 > metadata.GetCapacity()) {
+      if (metadata.IsScaling()) {
+        s = createBloomFilter(ns_key, &metadata);
```
PragmaTwice: the status seems not checked

mapleFU: I'm not sure it can be done like this. Now we have holding a "lock" here. And `createBloomFilter` will directly call `Write` finally

I propose a `createBloomFilterInBatch` or other, the interface would just put the new bloomFilter in batch. The previous create can make use of this, and here we can just put it in batch first?

src/types/redis_bloom_chain.cc
```diff
@@ -147,18 +171,23 @@ rocksdb::Status BloomChain::Add(const Slice &user_key, const Slice &item, int *r
...
+        if (!s.ok()) return s;
+      } else {
+        return rocksdb::Status::Aborted("filter is full and is nonscaling");
+      }
     }
     s = bloomAdd(bf_key_list.back(), item_string);
```
mapleFU: Should we enlarge `bf_key_list` if scaling?

Directly insert is ok if size == 1, since it would be not matter to insert a key here. However I think when madd happens, would there insert too many keys? and make the "Last" BloomFilter Huge?

zncleon: Sure, it certainly needs to be enlarged. I think I mistakenly missed it.

src/types/redis_bloom_chain.cc
```diff
@@ -56,12 +56,36 @@ rocksdb::Status BloomChain::createBloomChain(const Slice &ns_key, double error_r
...
+
+  std::string bf_key = getBFKey(ns_key, *metadata, metadata->n_filters - 1);
+  batch->Put(bf_key, block_split_bloom_filter.GetData());
+
+  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+rocksdb::Status BloomChain::createBloomFilter(const Slice &ns_key, BloomChainMetadata *metadata) {
+  uint32_t bloom_filter_bytes = BlockSplitBloomFilter::OptimalNumOfBytes(
+      static_cast<uint32_t>(metadata->base_capacity * pow(metadata->expansion, metadata->n_filters)),
```
mapleFU, line 75: Might not strong related to this patch, should we enforce a "hardlimit" to avoid too large BloomFilter?

cc @PragmaTwice @git-hulk

PragmaTwice: Does Redis have some limits here?

mapleFU: RedisBloom has a maximum Cuckoo Filter `n_filters`. And don't have limit

I think our bitmap has some limits. And `BlockSplitBloomFilter` limit 128MB, perhaps if filter is too large, we need to reject and return to user?

git-hulk: 128MiB is also too large for one key, maybe we can improve this by separating them into multi subkeys in future.

mapleFU: Previously I talked with @zncleon . Split into subkeys doesn't work like bitmap, since bloom filter would tent to random access all bits in it's space. Maybe a maximum size would help?

zncleon:
If error and return after `CreateBloomfilter`, it will left zombie bloomfilter. So I Update the function interfaces about `CreateBloomfilterInBatch `and `BloomAdd`. Using only one write will also get high performance.

cc @mapleFU

src/storage/redis_metadata.h
```diff
@@ -248,8 +248,7 @@ class BloomChainMetadata : public Metadata {
   using Metadata::Decode;
   rocksdb::Status Decode(Slice *bytes) override;
 
-  /// Get the total capacity of the bloom chain (the sum capacity of all sub-filters)
-  ///
-  /// @return the total capacity value
   uint32_t GetCapacity() const;
+
+  bool IsScaling() const { return expansion != 0; };
```
mapleFU, line 255: cc @PragmaTwice should adding `noexcept` better here?

PragmaTwice: currently kvrocks does not care about noexpect, so it can be ignored

src/types/redis_bloom_chain.cc
```diff
@@ -56,31 +56,44 @@ rocksdb::Status BloomChain::createBloomChain(const Slice &ns_key, double error_r
...
+  BlockSplitBloomFilter block_split_bloom_filter;
+  std::string in_data = *bf_data;
+  block_split_bloom_filter.Init(std::move(in_data));
 
   uint64_t h = BlockSplitBloomFilter::Hash(item.data(), item.size());
   block_split_bloom_filter.InsertHash(h);
-  auto batch = storage_->GetWriteBatchBase();
-  batch->Put(bf_key, block_split_bloom_filter.GetData());
-  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+  *bf_data = block_split_bloom_filter.GetData();
```
mapleFU: This interface is ok, but what about adding a `MoveData` to avoid an around of copying?

src/types/redis_bloom_chain.cc
```diff
@@ -56,31 +56,44 @@ rocksdb::Status BloomChain::createBloomChain(const Slice &ns_key, double error_r
...
+
+  std::string bloom_chain_meta_bytes;
+  metadata->Encode(&bloom_chain_meta_bytes);
+  batch->Put(metadata_cf_handle_, ns_key, bloom_chain_meta_bytes);
+}
+
+void BloomChain::bloomAdd(const std::string &item, std::string *bf_data) {
+  BlockSplitBloomFilter block_split_bloom_filter;
+  std::string in_data = *bf_data;
+  block_split_bloom_filter.Init(std::move(in_data));
```
```suggestion
  block_split_bloom_filter.Init(std::move(*bf_data));
```

The previous implemention in L:91 does a copy, which should not be used

mapleFU (commented):
Rest LGTM, thanks!

mapleFU:
Oh seems Twice has refactor the storage, so you mind need to solve the conflicts.

src/types/redis_bloom_chain.cc
PragmaTwice: From a glance, I think it seems that we can just return a reference of the string rather than copy it.

PragmaTwice: From a glance, I think we just add a `string&& GetData() &&`.

src/types/redis_bloom_chain.cc
```diff
@@ -56,31 +56,43 @@ rocksdb::Status BloomChain::createBloomChain(const Slice &ns_key, double error_r
...
+void BloomChain::bloomAdd(const std::string &item, std::string *bf_data) {
+  BlockSplitBloomFilter block_split_bloom_filter;
+  block_split_bloom_filter.Init(std::move(*bf_data));
 
   uint64_t h = BlockSplitBloomFilter::Hash(item.data(), item.size());
   block_split_bloom_filter.InsertHash(h);
-  auto batch = storage_->GetWriteBatchBase();
-  batch->Put(bf_key, block_split_bloom_filter.GetData());
-  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+  *bf_data = block_split_bloom_filter.GetData();
```
PragmaTwice:
```suggestion
  *bf_data = std::move(block_split_bloom_filter).GetData();
```

And add a overload to `GetData`:

```c++
std::string&& GetData() && { return data_; }
```

PragmaTwice: cc @mapleFU that seems better than a `MoveData` method.

mapleFU: Ok, this also LGTM

src/types/redis_bloom_chain.h
```diff
@@ -61,7 +61,9 @@ class BloomChain : public Database {
   rocksdb::Status getBloomChainMetadata(const Slice &ns_key, BloomChainMetadata *metadata);
   rocksdb::Status createBloomChain(const Slice &ns_key, double error_rate, uint32_t capacity, uint16_t expansion,
                                    BloomChainMetadata *metadata);
-  rocksdb::Status bloomAdd(const Slice &bf_key, const std::string &item);
+  void createBloomFilterInBatch(const Slice &ns_key, BloomChainMetadata *metadata,
+                                ObserverOrUniquePtr<rocksdb::WriteBatchBase> &batch, std::string *bf_data);
+  static void bloomAdd(const std::string &item, std::string *bf_data);
```
mapleFU, line 66: Nit: when using pointer as an argument, it looks like an *output*. Maybe we can add a comment to mark this like:

```
/// bf_data: [in/out] The content string of bloomfilter.
```

mapleFU (dismissed):
LGTM

src/types/bloom_filter.h
```diff
@@ -154,7 +154,9 @@ class BlockSplitBloomFilter {
   /// Get the plain bitset value from the Bloom filter bitset.
   ///
   /// @return bitset value;
-  const std::string& GetData() { return data_; }
+  const std::string& GetData() & { return data_; }
```
PragmaTwice:
```suggestion
  const std::string& GetData() const & { return data_; }
```

src/types/bloom_filter.h
```diff
@@ -154,7 +154,9 @@ class BlockSplitBloomFilter {
   /// Get the plain bitset value from the Bloom filter bitset.
   ///
   /// @return bitset value;
-  const std::string& GetData() { return data_; }
+  const std::string& GetData() const& { return data_; }
+
+  std::string GetData() const&& { return data_; }
```
mapleFU: Emmm would this need a const? You can take a look at how `StatusOr` does this: https://github.com/apache/kvrocks/blob/2ecaef221c1300f14d0a8f03a51ddd5f1f48766c/src/common/status.h#L276

PragmaTwice:
```suggestion
  std::string&& GetData() && { return std::move(data_); }
```

mapleFU (approved):

PragmaTwice:
Thanks for your contribution!

PR #1723, opened 2023-09-01
Add MOVE command as Redis

enjoy-binbin:
Since kvrocks don't support Redis DBs, so this is a dummy command.
If key does not exist, the command returns 0, otherwise it will always
returns 1.

### DIFF
--- a/tests/gocase/unit/introspection/introspection_test.go
+++ b/tests/gocase/unit/introspection/introspection_test.go
@@ -203,6 +203,18 @@ func TestIntrospection(t *testing.T) {
 		require.NoError(t, rdb.Set(ctx, "a", "b", 0).Err())
 		require.GreaterOrEqual(t, time.Since(now).Seconds(), 2.0)
 	})
+
+	t.Run("MOVE dummy coverage", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


git-hulk (dismissed):

tests/gocase/unit/introspection/introspection_test.go
```diff
@@ -203,6 +203,18 @@ func TestIntrospection(t *testing.T) {
 		require.NoError(t, rdb.Set(ctx, "a", "b", 0).Err())
 		require.GreaterOrEqual(t, time.Since(now).Seconds(), 2.0)
 	})
+
+	t.Run("MOVE dummy coverage", func(t *testing.T) {
+		require.Error(t, rdb.Do(ctx, "MOVE", "key", "dbid").Err())
+
+		// key dose not exist, return 0
```
torwig: A tiny typo here: `dose` => `does`.

torwig (dismissed):

tests/gocase/unit/introspection/introspection_test.go
enjoy-binbin:
```suggestion
		// key does not exist, return 0
```

git-hulk (approved):

PR #1726, opened 2023-09-02
Split Inter in ZSet::InterStore into a separate function

### LINKED ISSUE #1715: We may have the deadlock in ZSet::InterStore(as well as other similar APIs)
We may have the deadlock in ZSet::InterStore(as well as other similar APIs) if the dst key was inside the source key list since the OverWrite function will also lock the dst key: https://github.com/apache/kvrocks/blob/unstable/src/types/redis_zset.cc#L635 

This issue was found by TSAN in CI: https://github.com/apache/kvrocks/actions/runs/6022265431/job/16338859035?pr=1712

_Originally posted by @git-hulk in https://github.com/apache/kvrocks/issues/1712#issuecomment-1698893367_

enjoy-binbin:
CI TSAN show ZINTERSTORE may has a deadlock after introducing
locks to DEL in #1712. In ZSet::InterStore if the dst key was
inside the source key list we may have a deadlock since the
OverWrite function will also lock the dst key.

In this PR, we split Inter in ZSet::InterStore into a separate
function, just like the Set apis.

After this PR, after the CI verification in #1712, it can pass
the CI verification now. Closes #1715

This PR also do a saved_cnt cleanup since it is same as members.size().

### DIFF
--- a/src/types/redis_zset.cc
+++ b/src/types/redis_zset.cc
@@ -626,6 +626,16 @@ rocksdb::Status ZSet::Overwrite(const Slice &user_key, const MemberScores &mscor
 
 rocksdb::Status ZSet::InterStore(const Slice &dst, const std::vector<KeyWeight> &keys_weights,
                                  AggregateMethod aggregate_method, uint64_t *saved_cnt) {
+  *saved_cnt = 0;
+  std::vector<MemberScore> members;
+  auto s = Inter(keys_weights, aggregate_method, saved_cnt, &members);
+  if (!s.ok()) return s;
+  *saved_cnt = members.size();
+  return Overwrite(dst, members);
+}
+
+rocksdb::Status ZSet::Inter(const std::vector<KeyWeight> &keys_weights, AggregateMethod aggregate_method,
+                            uint64_t *saved_cnt, std::vector<MemberScore> *members) {
   std::vector<std::string> lock_keys;
   lock_keys.reserve(keys_weights.size());
   for (const auto &key_weight : keys_weights) {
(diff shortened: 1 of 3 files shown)


git-hulk (dismissed):

src/types/redis_zset.cc
```diff
@@ -626,6 +626,16 @@ rocksdb::Status ZSet::Overwrite(const Slice &user_key, const MemberScores &mscor
...
+  *saved_cnt = 0;
+  std::vector<MemberScore> members;
+  auto s = Inter(keys_weights, aggregate_method, saved_cnt, &members);
+  if (!s.ok()) return s;
+  *saved_cnt = members.size();
+  return Overwrite(dst, members);
+}
+
+rocksdb::Status ZSet::Inter(const std::vector<KeyWeight> &keys_weights, AggregateMethod aggregate_method,
+                            uint64_t *saved_cnt, std::vector<MemberScore> *members) {
```
PragmaTwice:
```suggestion
rocksdb::Status ZSet::Inter(const std::vector<KeyWeight> &keys_weights, AggregateMethod aggregate_method,
                            std::vector<MemberScore> *members) {
```
We do not need saved_cnt here, which is same as members.size().

PragmaTwice: Same problem found in  ZSet::Union.

enjoy-binbin: yes, i noticed ZSet::Union before, updated, thanks

PragmaTwice (approved):

torwig (approved):

PR #1737, opened 2023-09-04
Add the support of LMPOP command

### LINKED ISSUE #1526: Add the support of the LMPOP command
### Search before asking

### Motivation

This issue considers to support the [lmpop](https://redis.io/commands/lmpop/) command.

### Solution

_No response_

HolyLow:
This PR adds support for LMPOP command.
Currently the test cases are not added yet, but will soon. This PR serves as an implementation preview.
Any suggestions are welcomed.

### DIFF
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -154,6 +154,92 @@ class CommandRPop : public CommandPop {
   CommandRPop() : CommandPop(false) {}
 };
 
+class CommandLMPop : public Commander {
+ public:
+  /// format: LMPOP #numkeys key0 [key1 ...] <LEFT | RIGHT> [COUNT count]
+  Status Parse(const std::vector<std::string> &args) override {
+    auto v = ParseInt<int32_t>(args[1], 10);
+    if (!v) {
+      return {Status::RedisParseErr, errValueNotInteger};
+    }
+    if (*v <= 0) {
+      return {Status::RedisParseErr, errValueMustBePositive};
+    }
+    num_keys_ = *v;
+
+    if ((args.size() != 3 + num_keys_) && (args.size() != 5 + num_keys_)) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+
+    std::string left_or_right = util::ToLower(args[1 + num_keys_]);
+    if (left_or_right == "left") {
+      left_ = true;
+    }
+    else if (left_or_right == "right") {
+      left_ = false;
+    }
+    else {
+      return {Status::RedisParseErr, errInvalidSyntax};
+    }
+
+    if (args.size() == 5 + num_keys_) {
+      if (util::ToLower(args[3 + num_keys_]) != "count") {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+      auto c = ParseInt<int32_t>(args[4 + num_keys_]);
+      if (!c) {
+        return {Status::RedisParseErr, errValueNotInteger};
+      }
+      if (*c < 0) {
+        return {Status::RedisParseErr, errValueMustBePositive};
+      }
+      count_ = *c;
+    }
+
+    return Status::OK();
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::List list_db(svr->storage, conn->GetNamespace());
+
+    std::vector<std::string> elems;
+    int key_id = 0;
+    for (; key_id < num_keys_; ++key_id) {
+      auto s = list_db.PopMulti(args_[key_id + 2], left_, count_, &elems);
+      if (!s.ok() && !s.IsNotFound()) {
+        return {Status::RedisExecErr, s.ToString()};
+      }
+      if (!elems.empty()) {
+        break;
+      }
+    }
+
+    if (elems.empty()) {
+      *output = redis::NilString();
+    }
+    else {
+      std::string elems_bulk = redis::MultiBulkString(elems);
+      *output = redis::MultiBulkString({args_[key_id + 2], elems_bulk});
+    }
+
+    return Status::OK();
+  }
+
+  static const inline CommandKeyRangeGen keyRangeGen = [](const std::vector<std::string> &args) {
+    CommandKeyRange range;
+    range.first_key = 2;
+    range.key_step = 1;
+    auto v = ParseInt<int32_t>(args[1], 10);
+    range.last_key = range.first_key + (*v) - 1;
+    return range;
+  };
+
+ private:
+  bool left_;
+  uint32_t num_keys_ = 1;
+  uint32_t count_ = 1;
+};
+
 class CommandBPop : public Commander,
                     private EvbufCallbackBase<CommandBPop, false>,
                     private EventCallbackBase<CommandBPop> {
@@ -766,6 +852,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandBLPop>("blpop", -3, "write no-script"
                         MakeCmdAttr<CommandLRem>("lrem", 4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandLSet>("lset", 4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandLTrim>("ltrim", 4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandLMPop>("lmpop", -4, "write", CommandLMPop::keyRangeGen),
                         MakeCmdAttr<CommandRPop>("rpop", -2, "write", 1, 1, 1),
                         MakeCmdAttr<CommandRPopLPUSH>("rpoplpush", 3, "write", 1, 2, 1),
                         MakeCmdAttr<CommandRPush>("rpush", -3, "write", 1, 1, 1),
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -179,8 +179,9 @@ class CommandLMPop : public Commander {
     if (parser.Good()) {
       if (parser.EatEqICase("count")) {
         count_ = GET_OR_RET(parser.TakeInt<uint32_t>());
-      } else
+      } else {
         return parser.InvalidSyntax();
+      }
     }
 
     return parser.Good() ? parser.InvalidSyntax() : Status::OK();
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -174,11 +174,9 @@ class CommandLMPop : public Commander {
     std::string left_or_right = util::ToLower(args[num_keys_ + 2]);
     if (left_or_right == "left") {
       left_ = true;
-    }
-    else if (left_or_right == "right") {
+    } else if (left_or_right == "right") {
       left_ = false;
-    }
-    else {
+    } else {
       return {Status::RedisParseErr, errInvalidSyntax};
     }
 
@@ -216,8 +214,7 @@ class CommandLMPop : public Commander {
 
     if (elems.empty()) {
       *output = redis::NilString();
-    }
-    else {
+    } else {
       std::string elems_bulk = redis::MultiBulkString(elems);
       *output = redis::Array({redis::BulkString(args_[key_id + 2]), elems_bulk});
     }
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -176,15 +176,18 @@ class CommandLMPop : public Commander {
       return {Status::RedisParseErr, errInvalidSyntax};
     }
 
-    if (parser.Good()) {
-      if (parser.EatEqICase("count")) {
+    while (parser.Good()) {
+      if (parser.EatEqICase("count") && count_ == -1) {
         count_ = GET_OR_RET(parser.TakeInt<uint32_t>());
       } else {
         return parser.InvalidSyntax();
       }
     }
+    if (count_ == -1) {
+      count_ = 1;
+    }
 
-    return parser.Good() ? parser.InvalidSyntax() : Status::OK();
+    return Status::OK();
   }
 
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
@@ -225,7 +228,7 @@ class CommandLMPop : public Commander {
 
  private:
   bool left_;
-  uint32_t count_ = 1;
+  uint32_t count_ = -1;
   std::vector<std::string> keys_;
 };
 
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -217,8 +217,9 @@ class CommandLMPop : public Commander {
     CommandKeyRange range;
     range.first_key = 2;
     range.key_step = 1;
-    auto v = ParseInt<int32_t>(args[1], 10);
-    range.last_key = range.first_key + (*v) - 1;
+    // This parsing would always succeed as this cmd has been parsed before.
+    auto num_key = *ParseInt<int32_t>(args[1], 10);
+    range.last_key = range.first_key + num_key - 1;
     return range;
   };
 
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -206,7 +206,7 @@ class CommandLMPop : public Commander {
       *output = redis::NilString();
     } else {
       std::string elems_bulk = redis::MultiBulkString(elems);
-      *output = redis::Array({redis::BulkString(chosen_key), elems_bulk});
+      *output = redis::Array({redis::BulkString(chosen_key), std::move(elems_bulk)});
     }
 
     return Status::OK();
--- a/src/commands/cmd_list.cc
+++ b/src/commands/cmd_list.cc
@@ -156,7 +156,7 @@ class CommandRPop : public CommandPop {
 
 class CommandLMPop : public Commander {
  public:
-  /// format: LMPOP #numkeys key0 [key1 ...] <LEFT | RIGHT> [COUNT count]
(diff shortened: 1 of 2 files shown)


PragmaTwice:
Thanks for your contribution!

Please refer to https://kvrocks.apache.org/community/contributing/ to format, lint and test your code : )

src/commands/cmd_list.cc
```diff
@@ -167,11 +167,11 @@ class CommandLMPop : public Commander {
     }
     num_keys_ = *v;
 
-    if ((args.size() != 3 + num_keys_) && (args.size() != 5 + num_keys_)) {
+    if ((args.size() != num_keys_ + 3) && (args.size() != num_keys_ + 5)) {
       return {Status::RedisParseErr, errWrongNumOfArguments};
     }
 
-    std::string left_or_right = util::ToLower(args[1 + num_keys_]);
+    std::string left_or_right = util::ToLower(args[num_keys_ + 2]);
```
git-hulk: @HolyLow Can use the CommandParser to parse arguments.

HolyLow: @git-hulk I have some questions on the usage of CommandParser.
The common usage of CommandParser is like:
```
while (cmdParser.isGood()) {
  if (eatCase0()) { doSomething(); }
  else if (eatCase1()) { doSomething(); }
  else { return ParseError(kInvalidSyntax); }
}
```
But this schema assumes no strict order constraint and no strict repeat constraint. For example, the standard LMPOP command is like: 

*LMPOP 2 key1 key2 LEFT COUNT 2*; 

but with the parsing schema above, this command is also considered as valid: 

*LMPOP 2 key1 key2 LEFT RIGHT COUNT 2 COUNT 4 RIGHT*. 

The repeat and the order of keywords are not constrained by the isGood() loop.  Even though we could constrain the repeat by checking args.size(), the order of keywords is not constrained.

So should I constrain the order of the keywords, or leave it loose? If the orders of the keywords SHOULD be constrained, the existing parsers using CommandParser might have parsing bugs, as the CommandParser has no order semantic by design.

git-hulk: We can align the behavior with Redis, I'm not sure if Redis allows the duplicate keyword like you mentioned.

> The repeat and the order of keywords are not constrained by the isGood() loop.

We can work around this by checking if the variable has already been set. For example:

```c
direction = DirectionNone
while (cmdParser.isGood()) {
  if (eatCase0()) { doSomething(); }
  else if (eatCase1("LEFT")) {
     if (direction != DirectionNone) {
        // throw error
     }
     direction = DirectionLeft
  } else { return ParseError(kInvalidSyntax); }
}
```

HolyLow: Sure, we could find out some way to handle the order/repeat issue, but I just think that these workarounds lose the elegancy brought by CommandParser. Maybe the better way is to refactor CommandParser to support order/repeat checking semantic (which is painful...). Well this might be another issue.

I would refactor the parsing with the CommandParser and check if redis cares about keyword repeat/order.

PragmaTwice: IMHO here we do not need a `while` loop.

```c++

auto num_keys = GET_OR_RET(parser.TakeInt<uint32_t>());
for (uint32_t i = 0; i < num_keys; ++i) keys_.emplace(GET_OR_RET(parser.TakeStr())); // or you can choose to not have a key vector.

auto left_or_right = GET_OR_RET(parser.TakeStr());
// do something about left_or_right 

if (parser.Good()) {
  if (parser.EatEqICase("count")) count_ = GET_OR_RET(parser.TakeInt<uint32_t>());
  else return parser.InvalidSyntax();
}

if (parser.Good()) return parser.InvalidSyntax();
```

To your questions:

> as the CommandParser has no order semantic by design.

Nope. Whether there is an order is **entirely up to you**, not the CommandParser.

Your code contains **so many magic index numbers**, which makes it very hard to maintain and scale.

git-hulk: Thank you! Redis should not care about the order in most commands if my memory works. cc @PragmaTwice

HolyLow: @PragmaTwice you are right about the order issue, it's the user's responsibility to charge for the order rather than CommandParser. Besides, the GET_OR_RET macro makes the err handling much easier.

@git-hulk I've refactored the code and tested the redis behavior. Seems for most commands (e.g. LPOS) the ordering of different keywords doesn't matter, but for LMPOP, the ordering must be fixed, otherwise redis reports "syntax error". So our discussion here is not for nothing.

src/commands/cmd_list.cc
```diff
@@ -154,6 +154,79 @@ class CommandRPop : public CommandPop {
...
+    }
+
+    return Status::OK();
+  }
+
+  static const inline CommandKeyRangeGen keyRangeGen = [](const std::vector<std::string> &args) {
+    CommandKeyRange range;
+    range.first_key = 2;
+    range.key_step = 1;
+    auto v = ParseInt<int32_t>(args[1], 10);
```
git-hulk: Guess we should check the range of `v` here?

HolyLow: There are two reasons I didn't do the checking here. 

Firstly, in the call chain, the command would be parsed before this keyRangeGen is invoked, so if the range is invalid then the parsing phase would report error. 

Secondly (and most importantly...), I am not aware how to handle the exception/error if there is something wrong here. The error handling style of this project is mostly via the returned error code, and the rangeGen interface has no such error return value. I am not sure what to do if there is an error here. Should I throw some exception?

git-hulk: ok, it makes sense, thank you.

PragmaTwice: Yeah, the parsing will be invoked before this method, so technically the ParseInt will always succeed.

git-hulk (dismissed):
LGTM

tests/gocase/unit/type/list/list_test.go
```diff
@@ -1000,4 +1000,100 @@ func TestList(t *testing.T) {
 			require.Empty(t, rdb.LPosCount(ctx, "mylist", "b", 10, redis.LPosArgs{Rank: 5}).Val())
 		})
 	}
+
```
mapleFU, line 1003:
```go
// LMPop Pops one or more elements from the first non-empty list key from the list of provided key names.
// direction: left or right, count: > 0
// example: client.LMPop(ctx, "left", 3, "key1", "key2")
func (c cmdable) LMPop(ctx context.Context, direction string, count int64, keys ...string) *KeyValuesCmd {
	args := make([]interface{}, 2+len(keys), 5+len(keys))
	args[0] = "lmpop"
	args[1] = len(keys)
	for i, key := range keys {
		args[2+i] = key
	}
	args = append(args, strings.ToLower(direction), "count", count)
	cmd := NewKeyValuesCmd(ctx, args...)
	_ = c(ctx, cmd)
	return cmd
}
```

Seems `go-redis` `LMPOP` will force a `COUNT` argument? should we try to test without `COUNT`?

src/commands/cmd_list.cc
```diff
@@ -154,6 +154,79 @@ class CommandRPop : public CommandPop {
...
+        chosen_key = key;
+        break;
+      }
+    }
+
+    if (elems.empty()) {
+      *output = redis::NilString();
+    } else {
+      std::string elems_bulk = redis::MultiBulkString(elems);
+      *output = redis::Array({redis::BulkString(chosen_key), elems_bulk});
```
mapleFU:
```suggestion
      *output = redis::Array({redis::BulkString(chosen_key), std::move(elems_bulk)});
```

mapleFU (commented):
General LGTM. The test is reach, thanks for the carefully testing!

src/commands/cmd_list.cc
```diff
@@ -154,6 +154,79 @@ class CommandRPop : public CommandPop {
...
 };
 
+class CommandLMPop : public Commander {
+ public:
+  /// format: LMPOP #numkeys key0 [key1 ...] <LEFT | RIGHT> [COUNT count]
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 1);
+
+    auto num_keys = GET_OR_RET(parser.TakeInt<uint32_t>());
+    keys_.clear();
```
torwig, line 164: Is this call redundant?

src/commands/cmd_list.cc
```diff
@@ -154,6 +154,79 @@ class CommandRPop : public CommandPop {
...
+    } else if (left_or_right == "right") {
+      left_ = false;
+    } else {
+      return {Status::RedisParseErr, errInvalidSyntax};
+    }
+
+    if (parser.Good()) {
+      if (parser.EatEqICase("count")) {
+        count_ = GET_OR_RET(parser.TakeInt<uint32_t>());
+      } else
```
torwig: I'd use curly braces with the `else` part of the statement.

src/commands/cmd_list.cc
```diff
@@ -154,6 +154,79 @@ class CommandRPop : public CommandPop {
...
+        return parser.InvalidSyntax();
+    }
+
+    return parser.Good() ? parser.InvalidSyntax() : Status::OK();
```
torwig: What does it mean to check `parser.Good()` here?

mapleFU: I guess it means to check "no argument left"

PragmaTwice: > I guess it means to check "no argument left"

Exactly.

tests/gocase/unit/type/list/list_test.go
HolyLow: Sure, I am not familiar with go-redis, so have no idea how to strip the 'count' argument. But now it occurs that I could simply copy-and-modify the go-redis's LMPOP function... Will do it, thank you.

src/commands/cmd_list.cc
HolyLow: This call is some kind of "defensive programming". If the Parse() function is called more than once, the keys_ vector would has too many args (accumulates a batch each time called). 
Besides, the following Execute() relies on keys_.size(), so we don't want the keys_ vector to be polluted. 
Lastly, this calling has almost no overhead. 
So I tend to leave it there.

src/commands/cmd_list.cc
HolyLow: Well, this code is refactored by the naughty clang-format with "./x.py format". I personally don't like this no-brace style as well. Will try to add it back, but if the clang-format still says no, I have to follow its instruction to pass the format checking.

tests/gocase/unit/type/list/list_test.go
git-hulk: @HolyLow Can use the go-redis client's Do function which can pass the command arguments instead of encapsulating one.

src/commands/cmd_list.cc
```diff
@@ -154,6 +154,79 @@ class CommandRPop : public CommandPop {
...
+    auto left_or_right = util::ToLower(GET_OR_RET(parser.TakeStr()));
+    if (left_or_right == "left") {
+      left_ = true;
+    } else if (left_or_right == "right") {
+      left_ = false;
+    } else {
+      return {Status::RedisParseErr, errInvalidSyntax};
+    }
+
+    if (parser.Good()) {
```
enjoy-binbin: i prefer using the while loop (though we now only have one arg), so that we can remove the parser.Good() in `return parser.Good() ? parser.InvalidSyntax() : Status::OK();`
```suggestion
    while (parser.Good()) {
```

src/commands/cmd_list.cc
```diff
@@ -154,6 +154,79 @@ class CommandRPop : public CommandPop {
   CommandRPop() : CommandPop(false) {}
 };
 
+class CommandLMPop : public Commander {
+ public:
+  /// format: LMPOP #numkeys key0 [key1 ...] <LEFT | RIGHT> [COUNT count]
```
enjoy-binbin: we did not use `///` in our codebase
```suggestion
  // format: LMPOP #numkeys key0 [key1 ...] <LEFT | RIGHT> [COUNT count]
```

src/commands/cmd_list.cc
```diff
@@ -154,6 +154,79 @@ class CommandRPop : public CommandPop {
...
+    }
+
+    return Status::OK();
+  }
+
+  static const inline CommandKeyRangeGen keyRangeGen = [](const std::vector<std::string> &args) {
+    CommandKeyRange range;
+    range.first_key = 2;
+    range.key_step = 1;
+    auto v = ParseInt<int32_t>(args[1], 10);
```
enjoy-binbin: Is there any special reason for writing this? i see all the range in our codebase is using this format.

Please consider consistent with the original code style, or at least, use a meaningful name instead of v
```suggestion
    int num_key = *ParseInt<int>(args[1], 10);
```

tests/gocase/unit/type/list/list_test.go
enjoy-binbin: you can do this (just issue the command with the plain Do):
```go
require.NoError(t, rdb.Do(ctx, "LMPOP", "1", "mykey", "left").Err())
```

tests/gocase/unit/type/list/list_test.go
```diff
@@ -1000,4 +1000,100 @@ func TestList(t *testing.T) {
...
+			resultKey, resultVal := result.Val()
+			require.NoError(t, result.Err())
+			require.EqualValues(t, key1, resultKey)
+			if direction == "LEFT" {
+				require.Equal(t, []string{"five"}, resultVal)
+			} else {
+				require.Equal(t, []string{"one"}, resultVal)
+			}
+		})
+		t.Run(fmt.Sprintf("LMPOP-test-oneKey-countMulti-%s", direction), func(t *testing.T) {
```
enjoy-binbin: please try to remove the `-` in the test name, also, please leave a newline between the each test
```suggestion

		t.Run(fmt.Sprintf("LMPOP test-oneKey-countMulti-%s", direction), func(t *testing.T) {
```

src/commands/cmd_list.cc
mapleFU: I guess we don't need `while` because if that we need to check the count of args later?

tests/gocase/unit/type/list/list_test.go
HolyLow: The LMPOP test cases with no COUNT keyword are added now.

src/commands/cmd_list.cc
HolyLow: Done.

src/commands/cmd_list.cc
HolyLow: Please refer to the discussion of bulk, pragmaTwice and me above. 

The while loop has no order semantic, and cannot guarantee the orders of keywords. However, as aligned to redis behavior, the LMPOP's keywords MUST have strict order. So I choose to not to adopt the CommandParser's while() loop schema.

src/commands/cmd_list.cc
HolyLow: I think this is just trying to remind the programmer and reviewer that this parsing might fail, thus we should always be careful about it.

tests/gocase/unit/type/list/list_test.go
HolyLow: Refactored to meet your suggestions.

src/commands/cmd_list.cc
HolyLow: Modified as requested.

src/commands/cmd_list.cc
enjoy-binbin: i did see the disscussion below

> LMPOP's keywords MUST have strict order

what do you mean that Redis's LMPOP must have the strict order? The COUNT is an optional arg, it dose not require the order. what happen if Redis's LMPOP have more optional args in the future?

enjoy-binbin: the keys and the left_or_right dose require the order (and you already handle it), but the rest of the args are the optional, although now it only have one COUNT arg

HolyLow: I could see your point, but the while loop here would bring another problem: what if the COUNT keyword is repeated?
For redis, the command "LMPOP 1 mylist RIGHT COUNT 1 COUNT 2" is considered invalid. If we use while loop to parse the COUNT keyword here, the repeated COUNT keywords would be accepted.
Sure we could use another flag like "is_count_parsed" to check if we have seen the COUNT before. But for what? These additional checking already brought more complexity than the "return isGood() ? caseA : caseB"!
I do try to modify the code as you suggest, but it really looks even uglier to me. If more optional keywords are added, then the parsing could be modified, but the repeat problem still exists.
I've tested other commands such as "LPOS", and their optional keywords could be repeated with only the last one works. But for LMPOP, the COUNT keyword could not be repeated. I don't know if this is a redis bug or redis feature.

enjoy-binbin: Redis's optional args for old commands are indeed, some can be repeated. But actually that is not very important, or the Redis community is not willing to spend energy to modify it, it will cost time and introduce the changes or break the git blame log

For the new commands, like LMPOP, it was introduced in Redis 7.0, or in fact, i am the one who introduced it. normally, a new `is_count_parsed` is indeed ugly. In redis, the count will be init to -1, and use something like `if (count == -1 && parser.EatEqICase("count"))` format to handle the repeat issue, and in the last, do a `if (count == -1) count = 1;` before the execute

HolyLow: Refactored as suggested.

git-hulk (approved):

mapleFU (approved):
Thanks!

torwig (approved):
LGTM

PragmaTwice (approved):

PragmaTwice:
Thanks for your contribution!

PR #1739, opened 2023-09-04
Fix error messages in parsing of bit commands

gogim1:
This PR makes the error message consistent with the error message in Redis when calling the bit commands.

### DIFF
--- a/src/commands/cmd_bit.cc
+++ b/src/commands/cmd_bit.cc
@@ -39,7 +39,7 @@ class CommandGetBit : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
     Status s = GetBitOffsetFromArgument(args[2], &offset_);
-    if (!s.IsOK()) return s;
+    if (!s.IsOK()) return {Status::RedisParseErr, "bit offset is not an integer or out of range"};;
 
     return Commander::Parse(args);
   }
@@ -157,7 +157,7 @@ class CommandBitPos : public Commander {
     } else if (args[2] == "1") {
       bit_ = true;
     } else {
-      return {Status::RedisParseErr, "bit should be 0 or 1"};
+      return {Status::RedisParseErr, errValueNotInteger};
     }
 
     return Commander::Parse(args);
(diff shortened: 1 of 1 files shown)


src/commands/cmd_bit.cc
```diff
@@ -157,7 +157,7 @@ class CommandBitPos : public Commander {
     } else if (args[2] == "1") {
       bit_ = true;
     } else {
-      return {Status::RedisParseErr, "bit should be 0 or 1"};
```
mapleFU, line 160: Seems that Redis would first check if it's integer ( via `getLongFromObjectOrReply`) and then check it's `1` or `0`

```c
    /* Parse the bit argument to understand what we are looking for, set
     * or clear bits. */
    if (getLongFromObjectOrReply(c,c->argv[2],&bit,NULL) != C_OK)
        return;
    if (bit != 0 && bit != 1) {
        addReplyError(c, "The bit argument must be 1 or 0.");
        return;
    }
```

Can we keep the same?

mapleFU (commented):
Nice catch!

src/commands/cmd_bit.cc
gogim1: done

mapleFU (dismissed):

mapleFU (dismissed):

src/commands/cmd_bit.cc
```diff
@@ -152,12 +152,16 @@ class CommandBitPos : public Commander {
       stop_ = *parse_stop;
     }
 
+    auto parse_arg = ParseInt<int64_t>(args[2], 10);
+    if (!parse_arg) {
+      return {Status::RedisParseErr, errValueNotInteger};
+    }
     if (args[2] == "0") {
```
PragmaTwice:
```suggestion
    if (*parse_arg == 0) {
```

src/commands/cmd_bit.cc
```diff
@@ -152,12 +152,16 @@ class CommandBitPos : public Commander {
...
       bit_ = false;
     } else if (args[2] == "1") {
```
PragmaTwice: ditto

gogim1: done

mapleFU:
Will merge if all CI passed

git-hulk (approved):

mapleFU (approved):

mapleFU:
Thank you for catching this @GoGim1

PragmaTwice:
I want to complain about the PR title, which is not formal and clear enough to me. A better message may be **"Fix error messages in parsing of bit commands"**.

@mapleFU You can adjust the commit message before you merge it, please do not miss the chance.

And please do not use spoken English and abbreviation, and avoid grammatical errors in command messages.

mapleFU:
Sorry for that, I'll wait for some time before merging. You can force push to overwrite the title @PragmaTwice

PragmaTwice:
It is forbidden by project settings to force push to the unstable branch.

Just be careful next time : )

PR #1741, opened 2023-09-05
Bump luaJIT to 53882a1

xiaobiaozhao:
Many fixes imported from Mike Pall's upstream https://github.com/luajit/luajit repository.

### DIFF
--- a/cmake/luajit.cmake
+++ b/cmake/luajit.cmake
@@ -40,8 +40,8 @@ if ((${CMAKE_SYSTEM_NAME} MATCHES "Darwin") AND (NOT CMAKE_OSX_DEPLOYMENT_TARGET
 endif ()
 
 FetchContent_DeclareGitHubWithMirror(luajit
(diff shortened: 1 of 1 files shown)


cmake/luajit.cmake
```diff
@@ -67,6 +67,6 @@ if (NOT lua_POPULATED)
 endif()
 
 add_library(luajit INTERFACE)
-target_include_directories(luajit INTERFACE ${luajit_BINARY_DIR}/include)
+target_include_directories(luajit INTERFACE ${luajit_BINARY_DIR}/include ${luajit_SOURCE_DIR}/src)
```
git-hulk, line 70: Why to include LuaJIT src here?

xiaobiaozhao: It appears to be luajit.h This automatically generated file is not found in the new version of luajit. I'll check it out when I have time.

PR #1743, opened 2023-09-06
Fix `GETEX` not working correctly on `Redis::Bitmap`

gogim1:
The expiration time of a `Redis::Bitmap` cannot be changed by `GETEX`, this PR solves the problem.

### DIFF
--- a/src/commands/cmd_string.cc
+++ b/src/commands/cmd_string.cc
@@ -88,13 +88,11 @@ class CommandGetEx : public Commander {
       redis::Bitmap bitmap_db(svr->storage, conn->GetNamespace());
       s = bitmap_db.GetString(args_[1], max_btos_size, &value);
       if (s.ok()) {
-        auto s = rocksdb::Status();
         if (ttl_ > 0) {
           s = bitmap_db.Expire(args_[1], ttl_ + util::GetTimeStampMS());
         } else if (persist_) {
           s = bitmap_db.Expire(args_[1], 0);
         }
-        if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
       }
     }
     if (!s.ok() && !s.IsNotFound()) {
--- a/src/commands/cmd_string.cc
+++ b/src/commands/cmd_string.cc
@@ -25,11 +25,11 @@
 #include "commands/command_parser.h"
 #include "error_constants.h"
 #include "server/server.h"
+#include "storage/redis_db.h"
 #include "time_util.h"
 #include "ttl_util.h"
 #include "types/redis_bitmap.h"
 #include "types/redis_string.h"
-
 namespace redis {
 
 class CommandGet : public Commander {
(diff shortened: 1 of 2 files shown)


git-hulk:
@GoGim1 Thanks for your good catch.

mapleFU:
@git-hulk Would the expired subkey finally cleared by compaction filter?

git-hulk:
> 



> @git-hulk Would the expired subkey finally cleared by compaction filter?

Yes, exactly.

src/commands/cmd_string.cc
```diff
@@ -25,11 +25,11 @@
 #include "commands/command_parser.h"
 #include "error_constants.h"
 #include "server/server.h"
+#include "storage/redis_db.h"
 #include "time_util.h"
 #include "ttl_util.h"
 #include "types/redis_bitmap.h"
 #include "types/redis_string.h"
-
```
PragmaTwice: Could you revert this line?

src/commands/cmd_string.cc
```diff
@@ -86,6 +86,8 @@ class CommandGetEx : public Commander {
       uint32_t max_btos_size = static_cast<uint32_t>(config->max_bitmap_to_string_mb) * MiB;
       redis::Bitmap bitmap_db(svr->storage, conn->GetNamespace());
       s = bitmap_db.GetString(args_[1], max_btos_size, &value);
+      if (ttl_ > 0) bitmap_db.Expire(args_[1], ttl_ + util::GetTimeStampMS());
+      if (persist_) bitmap_db.Expire(args_[1], 0);
```
git-hulk: It would be more consistent to use `if..else if` since it's exclusive between TTL and PERSIST.
```suggestion
      if (ttl_ > 0) {
        bitmap_db.Expire(args_[1], ttl_ + util::GetTimeStampMS());
      } else if (persist_) {
        bitmap_db.Expire(args_[1], 0);
      }
```

PragmaTwice: The returned status seems unchecked.

src/commands/cmd_string.cc
```diff
@@ -86,6 +86,8 @@ class CommandGetEx : public Commander {
       uint32_t max_btos_size = static_cast<uint32_t>(config->max_bitmap_to_string_mb) * MiB;
       redis::Bitmap bitmap_db(svr->storage, conn->GetNamespace());
       s = bitmap_db.GetString(args_[1], max_btos_size, &value);
```
mapleFU, line 89: Seems that L79 might find that `GetEx` is unexists, and then return. However when key is an expired Bitmap or having other error, we might not `Expire` it.

```
s = bitmap_db.GetString(args_[1], max_btos_size, &value);
if (s.ok()) {
  // ...
}
```

Better change to this?

gogim1:
Changes have been made in response to the comments above. PTAL. :)

src/commands/cmd_string.cc
```diff
@@ -86,6 +87,15 @@ class CommandGetEx : public Commander {
...
+      if (s.ok()) {
+        auto s = rocksdb::Status();
```
PragmaTwice:
```suggestion
```

src/commands/cmd_string.cc
```diff
@@ -86,6 +87,15 @@ class CommandGetEx : public Commander {
...
+        if (ttl_ > 0) {
+          s = bitmap_db.Expire(args_[1], ttl_ + util::GetTimeStampMS());
+        } else if (persist_) {
+          s = bitmap_db.Expire(args_[1], 0);
+        }
+        if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
```
PragmaTwice:
```suggestion
```

gogim1: done

mapleFU (approved):
LGTM

enjoy-binbin (approved):

git-hulk (approved):

PR #1747, opened 2023-09-08
Fix `TYPE` command with bloom filter

gogim1:
When #1699 introduced BloomFilter, RedisTypeNames was not modified
at the same time, causing the TYPE command return nothing in bf key.
Now it will return `MBbloom--` (keep names consistent with Redis).

### DIFF
--- a/src/commands/cmd_key.cc
+++ b/src/commands/cmd_key.cc
@@ -37,6 +37,7 @@ class CommandType : public Commander {
     RedisType type = kRedisNone;
     auto s = redis.Type(args_[1], &type);
     if (s.ok()) {
+      if (type >= RedisTypeNames.size()) return {Status::RedisExecErr, "Invalid type"};
       *output = redis::SimpleString(RedisTypeNames[type]);
       return Status::OK();
     }
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -61,7 +61,7 @@ enum RedisCommand {
   kRedisCmdLMove,
 };
 
-const std::vector<std::string> RedisTypeNames = {"none", "string", "hash",      "list",  "set",
+const std::vector<std::string> RedisTypeNames = {"none", "string", "hash",      "list",   "set",
                                                  "zset", "bitmap", "sortedint", "stream", "MBbloom--"};
 
 constexpr const char *kErrMsgWrongType = "WRONGTYPE Operation against a key holding the wrong kind of value";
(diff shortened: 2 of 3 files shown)


src/storage/redis_metadata.h
```diff
@@ -62,7 +62,7 @@ enum RedisCommand {
 };
 
 const std::vector<std::string> RedisTypeNames = {"none", "string", "hash",      "list",  "set",
-                                                 "zset", "bitmap", "sortedint", "stream"};
+                                                 "zset", "bitmap", "sortedint", "stream", "MBbloom--"};
```
PragmaTwice, line 67:
```suggestion
                                                 "zset", "bitmap", "sortedint", "stream", "bloomfilter"};
```

gogim1: > https://github.com/RedisBloom/RedisBloom/blob/master/src/rebloom.c#L1448

Redis seems to use `MBbloom--` as the type name of the bloom filter. Should we be consistent?

mapleFU: Lets keep consistent with Redis

mapleFU:
Also cc @zncleon

mapleFU:
This patch is LGTM. Would you mind add a tiny test in `tests/gocase/unit/type/bloom/bloom_test.go` for this?

gogim1:
> This patch is LGTM. Would you mind add a tiny test in `tests/gocase/unit/type/bloom/bloom_test.go` for this?

done

mapleFU (dismissed):

src/storage/redis_metadata.h
PragmaTwice: got it.

PragmaTwice (dismissed):

torwig (dismissed):
LGTM.
@GoGim1 Thank you for the contribution.

src/commands/cmd_key.cc
```diff
@@ -37,6 +37,7 @@ class CommandType : public Commander {
     RedisType type = kRedisNone;
     auto s = redis.Type(args_[1], &type);
     if (s.ok()) {
+      if (type >= RedisTypeNames.size()) return {Status::RedisExecErr, "Invalid type"};
```
enjoy-binbin, line 40: i think in this case we can trigger a crash instead of returning an error.

returning an error will make the user think he is doing something wrong,
when in fact this is a kvrocks internally bug (user can fire a issue and report this)

but i am also ok with it since mostly it will be a dead code.
(actually returning nothing seems better than returning an error)

please also add a comment near by RedisType and RedisTypeNames, mention that they need to stay in sync

gogim1: done

enjoy-binbin (approved):
thanks.

PR #1748, opened 2023-09-08
Support logging subcompaction job info in event listener

darionyaphet:


git-hulk:
@darionyaphet Thanks for your contribution.

src/storage/event_listener.cc
```diff
@@ -59,6 +59,79 @@ std::string CompressType2String(const rocksdb::CompressionType type) {
...
+      result = "RoundRobinTtl";
+      break;
+    case rocksdb::CompactionReason::kRefitLevel:
+      result = "RefitLevel";
+      break;
+    case rocksdb::CompactionReason::kNumOfReasons:
+      result = "NumOfReasons";
+      break;
+    default:
+      result = "UnSupport";
```
torwig: I'd change this to "UnexpectedReasonValue".
In general, LGTM.
@darionyaphet thank you for the contribution.

torwig:
@darionyaphet You can run `/.x.py format` to format the code.

src/storage/event_listener.cc
darionyaphet: ok & fixed

src/common/status.h
```diff
@@ -168,7 +168,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> :
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
-      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
+      : BaseType((typename StringInStatusOr<U>::BaseType&&)(std::move(v))) {}
```
torwig: I think you should manually revert the change (add space) in this particular place after the `format` command because CI will fail. I have the same issue with my Fedora 38.

darionyaphet: I saw the lint error because my clang format version was too high. I used 16 but it was 14 on github, so there was a difference.

src/storage/event_listener.cc
```diff
@@ -59,6 +59,79 @@ std::string CompressType2String(const rocksdb::CompressionType type) {
   return "unknown";
 }
 
+std::string CompactionReason2String(const rocksdb::CompactionReason compression_reason) {
```
caipengbo, line 62: We can use the same style as above. Put strings in `std::vector<std::string>` in order.

src/storage/event_listener.h
```diff
@@ -31,7 +31,10 @@ class EventListener : public rocksdb::EventListener {
   ~EventListener() override = default;
   void OnFlushBegin(rocksdb::DB *db, const rocksdb::FlushJobInfo &fi) override;
   void OnFlushCompleted(rocksdb::DB *db, const rocksdb::FlushJobInfo &fi) override;
+  void OnCompactionBegin(rocksdb::DB *db, const rocksdb::CompactionJobInfo &ci) override;
```
caipengbo, line 34: I was wondering if we should add functions about Begin, they don't seem to be useful, and with subcompaction-related information, these logs get a lot bigger and don't seem to be very helpful (in my opinion). Actually, we can just keep the functions about Completed, because it contains the `elapsed`, and we'll know when it happened.

torwig: @caipengbo One case that comes to my mind is a crash or long freeze/lock during the operation. In this case, we can see from logs that the operation had started but wasn't completed. However, I'm not sure if this particular case is common.

caipengbo: Maybe that's a case. But if you really need to look at this infomation, perhaps user has been affected (perhaps a write stall has occurred because the compaction is blocking), in which case it might be better to look at the engine log(which contain more information). 

But there's no problem with adding this logging, I just don't think it's useful.

src/storage/event_listener.cc
PragmaTwice, line 62:
```suggestion
std::string CompactionReason2String(rocksdb::CompactionReason compression_reason) {
```

src/storage/event_listener.h
darionyaphet: I added the Compact Begin log because compaction can take a long time to occur, and compaction can require compacts to confirm the Compact start time while compacts determine the current state.

git-hulk: Yes, I think it makes sense to add this for observing the compaction behavior.

git-hulk:
@darionyaphet Would you like to push forward this PR?

git-hulk:
This has been resolved in https://github.com/apache/kvrocks/pull/1962, close this PR.

PR #1750, opened 2023-09-09
Add nodiscard attribute to methods of Storage

PragmaTwice:
We add `[[nodiscard]]` attribute to methods of Storage to prevent missing handling of returned statuses.

For our `Status` and `StatusOr`, the `[[nodiscard]]` attribute is default, but for `rocksdb::Status`, it is missing.


torwig (approved):
LGTM

PR #1751, opened 2023-09-09
Add nodiscard to Metadata::Decode

PragmaTwice:
We add [[nodiscard]] attribute to Metadata::Decode to prevent missing handling of returned statuses.

For our Status and StatusOr, the [[nodiscard]] attribute is default, but for rocksdb::Status, it is missing.


torwig (approved):
LGTM

mapleFU (approved):

PR #1752, opened 2023-09-11
Add nodiscard to methods of redis::Database

PragmaTwice:
We add `[[nodiscard]]` attribute to methods of `redis::Database` to prevent missing handling of returned statuses.

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -127,7 +127,10 @@ Status Cluster::SetSlotRanges(const std::vector<SlotRange> &slot_ranges, const s
       if (old_node == myself_ && old_node != to_assign_node) {
         // If slot is migrated from this node
         if (migrated_slots_.count(slot) > 0) {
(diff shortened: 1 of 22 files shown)


mapleFU (approved):
Seems subclass of `Database` still suffer from not having `[[nodiscard]]`? Should we add thm later?

enjoy-binbin (approved):

git-hulk (approved):

torwig (approved):

PragmaTwice:
> Seems subclass of `Database` still suffer from not having `[[nodiscard]]`? Should we add thm later?

I actually want to replace all of them from rocksdb::Status to our Status or StatusOr, then we can just have implicit `[[nodiscard]]`s instead of adding them to all functions.

However, it requires additional effort, so I have not yet begun.

torwig:
@PragmaTwice It is a great idea to replace `rocksdb::Status` with `Status/StatusOr` in the `Database` class.

PR #1755, opened 2023-09-12
Fix GEO store commands not removing dst key when result set is empty

enjoy-binbin:
If dst exists, when using the store variant, we need to
delete the dst key when the result set is empty, like
we are overwriting the dst key with an empty result set.

This change covers the following commands and scenarios:
- GEOSEARCHSTORE FROMMEMBER against non-existing src key.
- GEOSEARCHSTORE FROMLONLAT against non-existing src key.
- GEOSEARCHSTORE FROMLONLAT the search result set is empty.
- GEORADIUS STORE Against non-existing src key.
- GEORADIUS STORE search result set is empty.
- GEORADIUSBYMEMBER STORE against non-existing src key.
- GEORADIUSBYMEMBER STORE search result set is empty.

While writing the test cases, we still have a issue, FROMMEMBER
against non-existing src key member:
```
127.0.0.1:6666> geoadd src 10 10 Shenzhen
(integer) 1
127.0.0.1:6666> GEOSEARCHSTORE dst src FROMMEMBER Shenzhen_2 BYBOX 88 88 m
(integer) 0

127.0.0.1:6379> GEOADD src 10 10 Shenzhen
(integer) 1
127.0.0.1:6379> GEOSEARCHSTORE dst src FROMMEMBER Shenzhen_2 BYBOX 88 88 m
(error) ERR could not decode requested zset member
```

Redis will return a specific error if the member does not exist,
but Kvrocks currently handles it as if the src key does not exist,
this will be addressed in a future PR since it require more works.

### DIFF
--- a/src/types/redis_geo.cc
+++ b/src/types/redis_geo.cc
@@ -115,7 +115,7 @@ rocksdb::Status Geo::SearchStore(const Slice &user_key, GeoShape geo_shape, Orig
   if (point_type == kMember) {
     GeoPoint geo_point;
     auto s = Get(user_key, member, &geo_point);
-    // store key is not emtpy, try to remove it before returning.
+    // store key is not empty, try to remove it before returning.
     if (!s.ok() && s.IsNotFound() && !store_key.empty()) {
       auto del_s = ZSet::Del(store_key);
       if (!del_s.ok()) return s;
@@ -129,7 +129,7 @@ rocksdb::Status Geo::SearchStore(const Slice &user_key, GeoShape geo_shape, Orig
   std::string ns_key = AppendNamespacePrefix(user_key);
   ZSetMetadata metadata(false);
   rocksdb::Status s = ZSet::GetMetadata(ns_key, &metadata);
-  // store key is not emtpy, try to remove it before returning.
+  // store key is not empty, try to remove it before returning.
   if (!s.ok() && s.IsNotFound() && !store_key.empty()) {
     auto del_s = ZSet::Del(store_key);
     if (!del_s.ok()) return s;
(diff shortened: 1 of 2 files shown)


src/types/redis_geo.cc
```diff
@@ -115,6 +115,11 @@ rocksdb::Status Geo::SearchStore(const Slice &user_key, GeoShape geo_shape, Orig
   if (point_type == kMember) {
     GeoPoint geo_point;
     auto s = Get(user_key, member, &geo_point);
+    // store key is not empty, try to remove it before returning.
+    if (!s.ok() && s.IsNotFound() && !store_key.empty()) {
+      auto del_s = ZSet::Del(store_key);
+      if (!del_s.ok()) return s;
```
torwig: Should we return `del_s` here?

src/types/redis_geo.cc
```diff
@@ -124,6 +129,11 @@ rocksdb::Status Geo::SearchStore(const Slice &user_key, GeoShape geo_shape, Orig
   std::string ns_key = AppendNamespacePrefix(user_key);
   ZSetMetadata metadata(false);
   rocksdb::Status s = ZSet::GetMetadata(ns_key, &metadata);
+  // store key is not empty, try to remove it before returning.
+  if (!s.ok() && s.IsNotFound() && !store_key.empty()) {
+    auto del_s = ZSet::Del(store_key);
+    if (!del_s.ok()) return s;
```
torwig: The same here.

src/types/redis_geo.cc
```diff
@@ -134,6 +144,11 @@ rocksdb::Status Geo::SearchStore(const Slice &user_key, GeoShape geo_shape, Orig
 
   // if no matching results, give empty reply
   if (geo_points->empty()) {
+    // store key is not empty, try to remove it before returning.
+    if (!store_key.empty()) {
+      auto del_s = ZSet::Del(store_key);
+      if (!del_s.ok()) return s;
```
torwig: And here.

src/types/redis_geo.cc
enjoy-binbin: right, originally i wrote s, then changed it to del_s but forgot to modify the latter part. thanks

git-hulk (approved):

torwig (approved):

PR #1758, opened 2023-09-15
Add the support of the BF.MADD command

zncleon:
This PR introduce the bloom BF.MADD command.

[1] [https://redis.io/commands/bf.madd/](https://redis.io/commands/bf.madd/)


zncleon:
I add the `getBFDataList` function to get all the bloomfilter data at once. And update the `bloomCheck` interface for only check the data.

zncleon:
About `BF.MADD` command, it will be likely to return error reply in a array. But I have no way to check it in gocase, for the string type is inconsistent to the error type. It seems have not enough API to transform it. How should I solve it ? 

cc @mapleFU

mapleFU:
Emmm nice catch. Since `MADD` in `RedisBloom` is not atomic, and just a helper...

You can try to use raw-text to check the test reponse in gocase, or cast it to any already-existing pattern( I think it's ok that not fully-consistent with RedisBloom)

src/commands/cmd_bloom_filter.cc
```diff
@@ -95,11 +95,46 @@ class CommandBFAdd : public Commander {
...
+    items_.reserve(args_.size() - 2);
+    for (size_t i = 2; i < args_.size(); ++i) {
+      items_.emplace_back(args_[i]);
+    }
+    return Commander::Parse(args);
+  }
+
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::BloomChain bloom_db(svr->storage, conn->GetNamespace());
+    std::vector<int> rets(items_.size(), 0);
```
mapleFU: Would something like `vector<bool>` better here?

src/types/redis_bloom_chain.cc
```diff
@@ -95,26 +111,12 @@ void BloomChain::bloomAdd(const Slice &item, std::string *bf_data) {
...
-  LatestSnapShot ss(storage_);
-  rocksdb::ReadOptions read_options;
-  read_options.snapshot = ss.GetSnapShot();
-  std::string bf_data;
-  rocksdb::Status s = storage_->Get(read_options, bf_key, &bf_data);
-  if (!s.ok()) return s;
+bool BloomChain::bloomCheck(const Slice &item, std::string &bf_data) {
   BlockSplitBloomFilter block_split_bloom_filter;
-  block_split_bloom_filter.Init(std::move(bf_data));
+  block_split_bloom_filter.Init(bf_data);
```
mapleFU: Maybe you can refactor `BlockSplitBloomFilter` and give it a non-owned slice version. Currently, `bloomCheck` will copy the whole bf_data, which could be cpu-consuming and waste

src/commands/cmd_bloom_filter.cc
zncleon: Sure, I think it's necessary to change it(and `Exists`) to the explicit type and name.

src/types/redis_bloom_chain.h
```diff
@@ -39,6 +39,12 @@ enum class BloomInfoType {
   kExpansion,
 };
 
+enum class AddRetType {
```
PragmaTwice:
```suggestion
enum class BloomAddResult {
```

src/types/redis_bloom_chain.h
```diff
@@ -39,6 +39,12 @@ enum class BloomInfoType {
...
+  kFailure,
+  kOk,
+  kFull,
```
PragmaTwice:
```suggestion
  kExist,
  kOk,
  kFull,
```

zncleon:
I refactor `BlockSplitBloomFilter` to a  non-owned slice version: `InsertHash(h, bitset)` and `FindHash(h, bitset)` . It's user's responsiblity to save the bloomfilter bitset. But:
1. Now `BlockSplitBloomFilter` only have static member. Maybe we can use `namespace` otherwise `class`.
2. It caller's responsibility to ensure the size of bitset. We should `Check` it first in `InsertHash` and `FindHash`. What methods do we have to check it ?

mapleFU:
You can maintain a protect/private constructor, and when build from non-owned, you can use a `unique_ptr<const BlockSplit>CreateNonOwned()`, and it can only call constant fn.

src/types/bloom_filter.h
```diff
@@ -84,6 +84,13 @@ class BlockSplitBloomFilter {
   /// @return false if the number of bytes of Bloom filter bitset is not a power of 2, and true means successfully init
   bool Init(std::string bitset);
 
+  /// Create the non-owned BlockSplitBloomFilter. It use the bitset as underlying bitset. It is the caller's
+  /// responsibility to ensure the bitset would not to change.
+  ///
+  /// @param bitset The given bitset for the Bloom filter underlying bitset.
+  /// @return the unique_ptr of the const non-owned BlockSplitBloomFilter
+  static std::unique_ptr<const BlockSplitBloomFilter> CreateNonOwned(const std::string& bitset);
```
PragmaTwice: why is it called "non-owned"?

zncleon: Because it create a non-owned slice version BloomFilter. The underlying bitset of BloomFilter is a reference string from caller. Maybe we should call it "ReadOnly" or others?

zncleon: Oh, I think it doesn‘t work as expected, I will fix it.

src/commands/cmd_bloom_filter.cc
```diff
@@ -91,24 +91,59 @@ class CommandBFAdd : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
     redis::BloomChain bloom_db(svr->storage, conn->GetNamespace());
-    int ret = 0;
+    BloomFilterAddResult ret = BloomFilterAddResult::kOk;
     auto s = bloom_db.Add(args_[1], args_[2], &ret);
     if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
 
-    *output = redis::Integer(ret);
+    if (ret == BloomFilterAddResult::kFull) {
```
mapleFU: Nit: we'd better use `switch` to avoid omit some enum

```
switch(ret) {
  case ..:
    output = ...
}
```

src/types/redis_bloom_chain.cc
```diff
@@ -142,79 +152,84 @@ rocksdb::Status BloomChain::Add(const Slice &user_key, const Slice &item, int *r
...
 
-  // check
-  std::vector<bool> exist{false};                      // TODO: to refine in BF.MADD
-  for (int i = metadata.n_filters - 1; i >= 0; --i) {  // TODO: to test which direction for searching is better
-    s = bloomCheck(bf_key_list[i], {item}, &exist);
-    if (!s.ok()) return s;
-    if (exist[0]) {
-      *ret = 0;
-      break;
+  for (size_t i = 0; i < items.size(); ++i) {
```
mapleFU, line 170: This part is right, but can we leave a todo here, and change the procedure to:
1. compute hashes for all insert keys ( we don't need these keys actually, hashes is computed multiple times)
2. for each `bf_data`, batch check the input hash if it doesn't exists

I guess this could be 2 times faster than current impl, and avoid multiple `bloomCheck`

mapleFU (commented):
At first glance this looks ok to me, but I think we can add some further optimizations

src/types/redis_bloom_chain.cc
zncleon: The suggestion 1 is great. But I think suggestion 2 would not work in `BF.MADD`, for the bf_data would be changed after an item is inserted and the next item maybe found as existed (bloomfilter false-positve).  
BTW, the suggestion 2 maybe work in` BF.MExists`.

mapleFU: Emmm I think you're right, but maybe this kind of conflict doesn't matter. 

Once this happened, it would report full or pre-reserve some bloom "slot", but the probability is low, and it merely meets hash collision, so I think it's ok to do this

mapleFU: The worst cast might be madd same key multiple times. Maybe we need to handle this case by dedup the hash or other. Here I just want to avoid get a BloomFilter key from storage for multiple times.

zncleon: But now, we just get bloomfilter data at once and put them into a data_list. It seems not to get from storage for multiple times.

src/commands/cmd_bloom_filter.cc
```diff
@@ -91,24 +91,69 @@ class CommandBFAdd : public Commander {
...
-    *output = redis::Integer(ret);
+    switch (ret) {
+      case BloomFilterAddResult::kOk:
+        *output = redis::Integer(1);
+        break;
+      case BloomFilterAddResult::kExist:
+        *output = redis::Integer(0);
+        break;
+      case BloomFilterAddResult::kFull:
+        *output = redis::Error("ERR nonscaling filter is full");
```
mapleFU, line 106: nit: adding a break here to prevent from forget

src/types/redis_bloom_chain.cc
```diff
@@ -85,32 +108,14 @@ void BloomChain::createBloomFilterInBatch(const Slice &ns_key, BloomChainMetadat
   batch->Put(metadata_cf_handle_, ns_key, bloom_chain_meta_bytes);
 }
 
-void BloomChain::bloomAdd(const Slice &item, std::string *bf_data) {
+void BloomChain::bloomAdd(const uint64_t &item_hash, std::string *bf_data) {
```
mapleFU: uint64_t is a small POD type, I think pass-by value for these small object is better

src/types/redis_bloom_chain.cc
```diff
@@ -85,32 +108,14 @@ void BloomChain::createBloomFilterInBatch(const Slice &ns_key, BloomChainMetadat
...
-  for (size_t i = 0; i < items.size(); ++i) {
-    // this item exists in other bloomfilter already, and it's not necessary to check in this bloomfilter.
-    if ((*exists)[i]) {
-      continue;
-    }
-    uint64_t h = BlockSplitBloomFilter::Hash(items[i].data(), items[i].size());
-    (*exists)[i] = block_split_bloom_filter.FindHash(h);
-  }
-  return rocksdb::Status::OK();
+bool BloomChain::bloomCheck(const uint64_t &item_hash, std::string &bf_data) {
```
mapleFU: ditto

src/types/redis_bloom_chain.cc
mapleFU: Oh I see, they're already in `bf_data_list`

src/types/redis_bloom_chain.cc
```diff
@@ -142,79 +155,90 @@ rocksdb::Status BloomChain::Add(const Slice &user_key, const Slice &item, int *r
...
-    batch->Put(bf_key_list.back(), bf_data);
-    *ret = 1;
-    metadata.size += 1;
   }
 
   std::string bloom_chain_metadata_bytes;
   metadata.Encode(&bloom_chain_metadata_bytes);
   batch->Put(metadata_cf_handle_, ns_key, bloom_chain_metadata_bytes);
-
+  batch->Put(bf_key_list.back(), bf_data_list.back());
```
mapleFU: We can add a "modified bf indices", rather than this. Assume we trigger "scaling", the logic here would be trickey, because it only put the final `bf_data_list`. And when no data has been added, here will also write a batch.

mapleFU (commented):
Rest LGTM

src/types/redis_bloom_chain.h
```diff
@@ -51,26 +57,27 @@ class BloomChain : public Database {
...
 
   /// bf_data: [in/out] The content string of bloomfilter.
-  static void bloomAdd(const Slice &item, std::string *bf_data);
+  static void bloomAdd(uint64_t item_hash, std::string *bf_data);
 
-  /// exists: [in/out] The items exist in bloomfilter already or not.
-  /// exists[i] is true means items[i] exists in other bloomfilter already, and it's not necessary to check in this
-  /// bloomfilter.
-  rocksdb::Status bloomCheck(const Slice &bf_key, const std::vector<Slice> &items, std::vector<bool> *exists);
+  static bool bloomCheck(uint64_t item_hash, std::string &bf_data);
```
PragmaTwice, line 81:
```suggestion
  static bool bloomCheck(uint64_t item_hash, std::string *bf_data);
```

we'd better make it more align with bloomAdd.

tests/gocase/unit/type/bloom/bloom_test.go
```diff
@@ -145,6 +146,82 @@ func TestBloom(t *testing.T) {
...
+			Add := rdb.Do(ctx, "bf.madd", key, buf, buf+"xx", buf+"yy")
+			require.NoError(t, Add.Err())
+		}
+		require.Equal(t, []interface{}{"Capacity", int64(30), "Size", int64(128), "Number of filters", int64(1), "Number of items inserted", int64(30), "Expansion rate", int64(2)}, rdb.Do(ctx, "bf.info", key).Val())
+
+		// bloom filter is full and scaling
+		require.NoError(t, rdb.Do(ctx, "bf.add", key, "xxx").Err())
+		require.Equal(t, []interface{}{"Capacity", int64(90), "Size", int64(384), "Number of filters", int64(2), "Number of items inserted", int64(31), "Expansion rate", int64(2)}, rdb.Do(ctx, "bf.info", key).Val())
+
+		// insert items, suppose false positives is 0
```
mapleFU: Would this produce some flacky test?

mapleFU (commented):
Overall LGTM, but I wonder would we produce some flacky test if random string is used?

tests/gocase/unit/type/bloom/bloom_test.go
zncleon: Sure, It's maybe flacky with a very small probability. I will fix it.

tests/gocase/unit/type/bloom/bloom_test.go
```diff
@@ -222,38 +304,43 @@ func TestBloom(t *testing.T) {
...
 
-		// insert items, suppose false positives is 0
-		for i := 0; i < 50; i++ {
+		// insert items
+		var insertNum int64 = 0
+		for insertNum < 50 {
 			buf := util.RandString(7, 8, util.Alpha)
 			Add := rdb.Do(ctx, "bf.add", key, buf)
 			require.NoError(t, Add.Err())
+			insertNum = rdb.Do(ctx, "bf.card", key).Val().(int64)
```
mapleFU, line 313: nice!

mapleFU (approved):

PragmaTwice:
Thanks for your contribution!

PR #1760, opened 2023-09-16
Fix pointer-bool-conversion warning in `GeoHashHelper::BoundingBox`

gogim1:
A warning is generated [here](https://github.com/apache/kvrocks/blob/83aaa7531fdfdd874a936287fe3f466b184c9e41/src/types/geohash.cc#L334):

> address of array 'geo_shape->bounds' will always evaluate to 'true' [-Wpointer-bool-conversion]GCC

This PR fixes it.


git-hulk (approved):
LGTM, thanks

torwig (approved):

PR #1762, opened 2023-09-17
Use SSL_sendfile in SockSendFile while available

PragmaTwice:
`SSL_sendfile` is available when OpenSSL version is no less than 3.0.


git-hulk (approved):
LGTM

torwig (approved):

PR #1763, opened 2023-09-18
Fix wrong replica's port number(=0) if the listening port is not set

xiaofan8421:
Background:
For some reason, I used Go to implement my slaves based on kvrocks. But by redis-cli to get replication-info from kvrocks(no matter what master/slave), the peer port is 0 which is wrong. 
Reproduce:
Not only localhost but also the cross-machine can reproduce the bug scene.

So, I fixed it. Maybe it is not your code standard, Please review it.

before:
img_v2_058c11fb-db25-4852-acc0-d3f0d757749h
after:
1b30b756-37a1-4846-a9d7-b1d585a53bf0

### DIFF
--- a/src/server/redis_connection.h
+++ b/src/server/redis_connection.h
@@ -94,7 +94,7 @@ class Connection : public EvbufCallbackBase<Connection> {
   void SetAnnounceIP(std::string ip) { announce_ip_ = std::move(ip); }
   std::string GetAnnounceIP() const { return !announce_ip_.empty() ? announce_ip_ : ip_; }
   uint32_t GetAnnouncePort() const { return listening_port_ != 0 ? listening_port_ : port_; }
-  std::string GetAnnounceAddr() const { return GetAnnounceIP() + ":" + std::to_string(listening_port_); }
+  std::string GetAnnounceAddr() const { return GetAnnounceIP() + ":" + std::to_string(GetAnnouncePort()); }
   uint64_t GetClientType() const;
   Server *GetServer() { return svr_; }
 
--- a/src/server/redis_connection.h
+++ b/src/server/redis_connection.h
@@ -93,6 +93,7 @@ class Connection : public EvbufCallbackBase<Connection> {
   int GetListeningPort() const { return listening_port_; }
   void SetAnnounceIP(std::string ip) { announce_ip_ = std::move(ip); }
   std::string GetAnnounceIP() const { return !announce_ip_.empty() ? announce_ip_ : ip_; }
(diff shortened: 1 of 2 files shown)


src/server/redis_connection.h
```diff
@@ -93,6 +93,7 @@ class Connection : public EvbufCallbackBase<Connection> {
   int GetListeningPort() const { return listening_port_; }
   void SetAnnounceIP(std::string ip) { announce_ip_ = std::move(ip); }
   std::string GetAnnounceIP() const { return !announce_ip_.empty() ? announce_ip_ : ip_; }
+  uint32_t GetAnnouncePort() const { return listening_port_ != 0 ? listening_port_ : port_; }
   std::string GetAnnounceAddr() const { return GetAnnounceIP() + ":" + std::to_string(listening_port_); }
```
git-hulk: We should fix the `GetAnnounceAddr` as well.

git-hulk:
@xiaofan8421 Thanks for your good catch! Leave a comment inline.

src/server/redis_connection.h
xiaofan8421: I had the same idea two days ago, but I am not very familiar with Kvrcoks source code. I have been new to Kvrocks for almost 2 months. I'm not sure if I fix GetAnnounceAddr, other calls will have a side effect or not. So I only added GetAnnouncePort for securely in case of bad import.

git-hulk: @xiaofan8421 That's fine, can fix GetAnnounceAddr's listening port as well.

PragmaTwice:
@xiaofan8421  Hi, would you like to finish the final modification according to @git-hulk before the 2.6.0 cherry-picking period? (refer to https://github.com/apache/kvrocks/discussions/1767)

If it can be merged before the deadline, it can be shipped to kvrocks 2.6.0 smoothly : )

If you are busy, we can help you to do that.

xiaofan8421:
> @xiaofan8421 Hi, would you like to finish the final modification according to @git-hulk before the 2.6.0 cherry-picking period? (refer to #1767)
> 
> If it can be merged before the deadline, it can be shipped to kvrocks 2.6.0 smoothly : )
> 
> If you are busy, we can help you to do that.

Thanks for your mention. Update PR. Please review it.

src/server/redis_connection.h
xiaofan8421: ok, done. Please review it.
Moreover, maybe GetAnnouncePort interface should be reserved for replica-announce-port or something else in the future.

git-hulk (approved):

torwig (approved):

PR #1764, opened 2023-09-18
Make BlockSplitBloomFilter a view

PragmaTwice:
Besides, we supplement licenses and introduce span-lite to use std::span in c++20.


src/types/bloom_filter.h
```diff
@@ -46,6 +49,40 @@ constexpr bool IsMultipleOf8(int64_t n) { return (n & 7) == 0; }
...
+/// Initialize the BlockSplitBloomFilter. The range of num_bytes should be within
+/// [kMinimumBloomFilterBytes, kMaximumBloomFilterBytes], it will be
+/// rounded up/down to lower/upper bound if num_bytes is out of range and also
+/// will be rounded up to a power of 2.
+///
+/// @param num_bytes The number of bytes to store Bloom filter bitset.
+OwnedBlockSplitBloomFilter CreateBlockSplitBloomFilter(uint32_t num_bytes);
+
+/// Initialize the BlockSplitBloomFilter. It copies the bitset as underlying
+/// bitset because the given bitset may not satisfy the 32-byte alignment requirement
```
mapleFU:
```suggestion
/// bitset when the given bitset may not satisfy the 32-byte alignment requirement
```

Would this be better?

mapleFU (dismissed):
+1

src/types/bloom_filter.h
mapleFU: Oh, seems this bloom never copy, and when copy it's happening, it doesn't have owned-`std::string` as return value?

PragmaTwice: it is just a view, does not own anything. If you want to a owned filter, you can have a std::tuple<BloomFilter, std::string>.

src/types/bloom_filter.h
```diff
@@ -46,6 +49,40 @@ constexpr bool IsMultipleOf8(int64_t n) { return (n & 7) == 0; }
...
+/// rounded up/down to lower/upper bound if num_bytes is out of range and also
+/// will be rounded up to a power of 2.
+///
+/// @param num_bytes The number of bytes to store Bloom filter bitset.
+OwnedBlockSplitBloomFilter CreateBlockSplitBloomFilter(uint32_t num_bytes);
+
+/// Initialize the BlockSplitBloomFilter. It copies the bitset as underlying
+/// bitset when the given bitset may not satisfy the 32-byte alignment requirement
+/// which may lead to segfault when performing SIMD instructions. It is the caller's
+/// responsibility to free the bitset passed in.
```
zncleon, line 70: Maybe we should modify some comments like this?
```suggestion
/// responsibility to ensure the bitset would not be changed when using the bloomfilter.
```

PragmaTwice: the bitset can be changed via `InsertHash`.

mapleFU: I guess @zncleon means the mutability. Sometime we may need `const` ctor

PragmaTwice: it seems not related to this function

mapleFU: Yeah, I mean can we provide something like:

```
StatusOr<unique_ptr<BlockSplitBloomFilter>> CreateBlockSplitBloomFilter(uint8_t* bitset, uint32_t num_bytes);
StatusOr<unique_ptr<BlockSplitBloomFilter>> CreateBlockSplitBloomFilter(std::string& bitset);
StatusOr<unique_ptr<const BlockSplitBloomFilter>> CreateBlockSplitBloomFilter(const uint8_t* bitset, uint32_t num_bytes);
StatusOr<unique_ptr<const BlockSplitBloomFilter>> CreateBlockSplitBloomFilter(const std::string& bitset);
```

PragmaTwice: if a readonly version is needed, you can have a `const BlockSplitBloomFilter`.

PragmaTwice: Got your point. I think you can add it when you need it.

mapleFU (approved):

PR #1768, opened 2023-09-19
Add the support of the BF.INSERT command

zncleon:
This PR:
1. Introduce the bloom BF.INSERT command
2. Change the parameter parser error return to unified and explicit msg (e.g. `"Bad capacity"`)

[1] [https://redis.io/commands/bf.insert/](https://redis.io/commands/bf.insert/)

### DIFF
--- a/src/commands/cmd_bloom_filter.cc
+++ b/src/commands/cmd_bloom_filter.cc
@@ -31,7 +31,7 @@ class CommandBFReserve : public Commander {
   Status Parse(const std::vector<std::string> &args) override {
     auto parse_error_rate = ParseFloat<double>(args[2]);
     if (!parse_error_rate) {
-      return {Status::RedisParseErr, errValueIsNotFloat};
+      return {Status::RedisParseErr, "Bad error rate"};
     }
     error_rate_ = *parse_error_rate;
     if (error_rate_ >= 1 || error_rate_ <= 0) {
@@ -40,7 +40,7 @@ class CommandBFReserve : public Commander {
 
     auto parse_capacity = ParseInt<uint32_t>(args[3], 10);
     if (!parse_capacity) {
-      return {Status::RedisParseErr, errValueNotInteger};
+      return {Status::RedisParseErr, "Bad capacity"};
     }
     capacity_ = *parse_capacity;
     if (capacity_ <= 0) {
@@ -56,7 +56,11 @@ class CommandBFReserve : public Commander {
         expansion_ = 0;
       } else if (parser.EatEqICase("expansion")) {
         has_expansion = true;
-        expansion_ = GET_OR_RET(parser.TakeInt<uint16_t>());
+        auto parse_expansion = parser.TakeInt<uint16_t>();
+        if (!parse_expansion.IsOK()) {
+          return {Status::RedisParseErr, "Bad expansion"};
+        }
+        expansion_ = parse_expansion.GetValue();
         if (expansion_ < 1) {
           return {Status::RedisParseErr, "expansion should be greater or equal to 1"};
         }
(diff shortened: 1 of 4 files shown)


zncleon:
The reasons Why I change the msg are:
1. If I direct parse (e.g. capacity) in `BF.Reserve` and use parser's `take` and `GET_OR_RET` in `BF.INSERT`, the error msg would be differentiated.
2. In this patch, the parse error msg is the same as the RedisBloom.

src/commands/cmd_bloom_filter.cc
```diff
@@ -31,7 +31,7 @@ class CommandBFReserve : public Commander {
   Status Parse(const std::vector<std::string> &args) override {
     auto parse_error_rate = ParseFloat<double>(args[2]);
     if (!parse_error_rate) {
-      return {Status::RedisParseErr, errValueIsNotFloat};
+      return {Status::RedisParseErr, "Bad error rate"};
```
mapleFU: We can use some constant here, e.g., in `cmd_bloom_filter.cc`

```
namespace {
errBadErrorRate = "Bad error rate";
...
}
```

mapleFU (commented):
I'm a little busy today, I'll take a look carefully tonight

src/commands/cmd_bloom_filter.cc
```diff
@@ -147,6 +151,103 @@ class CommandBFMAdd : public Commander {
...
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 2);
+    bool is_nonscaling = false;
+    bool has_expansion = false;
+    while (parser.Good()) {
+      if (parser.EatEqICase("capacity")) {
+        auto parse_capacity = parser.TakeInt<uint32_t>();
+        if (!parse_capacity.IsOK()) {
+          return {Status::RedisParseErr, "Bad capacity"};
```
mapleFU: ditto, this can be add in a anonymous namespace as constants

src/commands/cmd_bloom_filter.cc
```diff
@@ -147,6 +151,103 @@ class CommandBFMAdd : public Commander {
...
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+    }
+
+    if (is_nonscaling && has_expansion) {
+      return {Status::RedisParseErr, "nonscaling filters cannot expand"};
+    }
+
+    auto items_begin =
+        find_if(args.begin(), args.end(), [](std::string_view lhs) { return lhs == "items" || lhs == "ITEMS"; });
```
mapleFU: Can we use `toLower` and check `lower(lhs) == "items"` here?

src/commands/cmd_bloom_filter.cc
```diff
@@ -147,6 +151,103 @@ class CommandBFMAdd : public Commander {
...
+
+    if (items_begin == args.end()) {
```
mapleFU: Hmmm 

```suggestion
    if (items_begin == args.end() || items_begin + 1 == args.end()) {
```

?

mapleFU (commented):
Looks ok, just some advices

src/commands/cmd_bloom_filter.cc
PragmaTwice: `util::EqualICase` is a better choice.

src/commands/cmd_bloom_filter.cc
```diff
@@ -147,6 +151,103 @@ class CommandBFMAdd : public Commander {
...
+      return {Status::RedisParseErr, errInvalidSyntax};
+    }
+    items_.reserve(args_.end() - items_begin);
+    for (auto item_iter = items_begin + 1; item_iter < args.end(); ++item_iter) {
+      items_.emplace_back(*item_iter);
+    }
```
PragmaTwice:
```suggestion
      } else if (parser.EatEqICase("items")) {
        has_items = true;
        break;
      } else {
        return {Status::RedisParseErr, errInvalidSyntax};
      }
    }

    if (is_nonscaling && has_expansion) {
      return {Status::RedisParseErr, "nonscaling filters cannot expand"};
    }

    if (not has_items) {
      return {Status::RedisParseErr, errInvalidSyntax};
    }
    while (parser.Good()) {
      items_.emplace_back(GET_OR_RET(parser.TakeStr()));
    }
```

zncleon: It's great. But it will have some bugs for the `items_` is `vector<Slice>` and `Slice` will not maintain the data . Whether the parser can return the reference of `args_` (or we should do this) ?

PragmaTwice: If you want a most intuitive way, you can let args_ typed `vector<string>`.

Another way is to use RawTake().

zncleon: Sure, I think vector<string> would be more intutive

src/commands/cmd_bloom_filter.cc
```diff
@@ -136,15 +152,113 @@ class CommandBFMAdd : public Commander {
...
+
+    if (is_nonscaling && has_expansion) {
+      return {Status::RedisParseErr, errNonscalingButExpand};
+    }
+
+    if (not has_items) {
+      return {Status::RedisParseErr, errInvalidSyntax};
+    }
+
+    items_.reserve(args_.size() - 3);
```
mapleFU: Emmm just `emplace_back` is ok here...

mapleFU (commented):
I catch a cold these two days, sorry for late reply...

Other looks ok to me

src/types/redis_bloom_chain.h
```diff
@@ -57,18 +64,21 @@ class BloomChain : public Database {
...
   BloomChain(engine::Storage *storage, const std::string &ns) : Database(storage, ns) {}
   rocksdb::Status Reserve(const Slice &user_key, uint32_t capacity, double error_rate, uint16_t expansion);
-  rocksdb::Status Add(const Slice &user_key, const Slice &item, BloomFilterAddResult *ret);
-  rocksdb::Status MAdd(const Slice &user_key, const std::vector<Slice> &items, std::vector<BloomFilterAddResult> *rets);
-  rocksdb::Status Exists(const Slice &user_key, const Slice &item, bool *exist);
-  rocksdb::Status MExists(const Slice &user_key, const std::vector<Slice> &items, std::vector<bool> *exists);
+  rocksdb::Status Add(const Slice &user_key, const std::string &item, BloomFilterAddResult *ret);
+  rocksdb::Status MAdd(const Slice &user_key, const std::vector<std::string> &items,
+                       std::vector<BloomFilterAddResult> *rets);
+  rocksdb::Status InsertCommon(const Slice &user_key, const std::vector<std::string> &items,
```
PragmaTwice, line 70: I wonder why it is called `InsertCommon` instead of `Insert`?

zncleon: It's named as RedisBloom do. I think it indicates not only `Insert` but `Madd` share this function.

mapleFU: I guess this is extended from RedisBloom, lol

We can separate underlying impl and interface @zncleon

zncleon: But the main logic of these two function is same(only some judge differ, in other words, `Madd` is a special case of `Insert`). If we separate them, we would have many redundant codes. @mapleFU

PragmaTwice: Got it.

PragmaTwice (approved):

PragmaTwice:

PR #1776, opened 2023-09-24
Add the support of the namespace replication

### LINKED ISSUE #1720: Allow to propagate namespace and tokens from master to slaves
## Background

Currently, Kvrocks won't propagate the namespace and token between master and slaves, and it may cause trouble if users add them in master but missing in slaves. So it would make this process easier if we could propagate the namespace/token as well.

## Design

To implement this feature, we need to consider the following situations:

- How to replicate namespaces between the master and slaves
- How to be compatible with old versions

**For the replication:**

We now have the propagate column family to propagate commands between the master and slaves, so we can use it to replicate namespaces as well. The steps are like below:

1. Modify namespace/token in master(forbid changing in slaves)
2. Propagate the namespace event to notify slaves like the Lua script
3. Slaves replay namespace/token changes

This works well for new instances, but what if there are different namespaces between the master and slaves in old instances?  There are three possible conditions:

1. Namespace/token exists in master but NOT in slave
2. Namespace exists in slave but NOT in master
3. Namespace exists both in master and slave, but the token is different

For condition 1, we can add the namespace/token to the slave as well. For condition 2, it makes sense to keep the namespace NOT existing in the master since data are from it, so we can remove the namespace/token. The most troublesome condition is 3 since users may have already used the token to access the data, changing the token would cause an invalid password error. **To make this simple, I prefer also overwriting the token.** And offer the configuration to allow users to enable or disable this feature, they MUST understand the risks before enabling.

To be noticed, we don't expect to propagate the default namespace since it may affect the replication part, so we won't persist the default namespace.

**For the compatible issue:**

Currently, namespaces are kept in the configuration file, so we need to read from the rocksdb first(store in propagate column like Lua script) when starting:

- If the key exists in rocksdb, load namespace/token from rocksdb and check if any namespace/token is inside the configuration file. If yes, refuse to start the server and ask users to remove them and add them via the `namespace add` command.
- If the key does NOT exist in rocksdb, read namespace/token from the configuration file and write them into rocksdb. To be noticed, we need to add an empty value for the key even though no namespace/token.

Users can only modify the namespace via command if this feature is enabled.

git-hulk:
This PR implements the namespace replication between master and slaves, the detailed design is inside #1720. We added a new configuration `repl-namespace-enabled` to turn on/off this feature and it's disabled by default. 
 
- If `repl-namespace-enabled` is set to `no`, it behaves the same as the previous. All namespace configurations are stored in the configuration file.

- if `repl-namespace-enabled` is set to `yes`, all namespace configurations will be removed from the configuration file, stored in DB, and then propagated to slaves. If the namespace has already existed in the slave, it will be overwritten by the master. Another significant change is that slaves cannot change namespaces if the replication is enabled.

This closes #1720

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -51,6 +51,14 @@ daemonize no
 
 cluster-enabled no
 
+# By default, namespaces are stored in the configuration file and won't be replicated
+# to replicas. This option allows to change this behavior, so that namespaces are also
+# propagated to slaves. Note that:
+# 1) it won't replicate the 'masterauth' to prevent breaking master/replica replication
+# 2) it will overwrite replica's namespace with master's namespace, so be careful of in-using namespaces.
+#
+# Default: no
+repl-namespace-enabled no
 
 # Persist the cluster nodes topology in local file($dir/nodes.conf). This configuration
 # takes effect only if the cluster mode was enabled.
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -55,7 +55,8 @@ cluster-enabled no
 # to replicas. This option allows to change this behavior, so that namespaces are also
 # propagated to slaves. Note that:
 # 1) it won't replicate the 'masterauth' to prevent breaking master/replica replication
-# 2) it will overwrite replica's namespace with master's namespace, so be careful of in-using namespaces.
+# 2) it will overwrite replica's namespace with master's namespace, so be careful of in-using namespaces
+# 3) cannot switch off the namespace replication once it's enabled
 #
 # Default: no
 repl-namespace-enabled no
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -159,6 +159,7 @@ Config::Config() {
       {"log-retention-days", false, new IntField(&log_retention_days, -1, -1, INT_MAX)},
       {"persist-cluster-nodes-enabled", false, new YesNoField(&persist_cluster_nodes_enabled, true)},
       {"redis-cursor-compatible", false, new YesNoField(&redis_cursor_compatible, false)},
+      {"repl-namespace-enabled", false, new YesNoField(&repl_namespace_enabled, false)},
 
       /* rocksdb options */
       {"rocksdb.compression", false,
@@ -233,17 +234,17 @@ void Config::initFieldValidator() {
   std::map<std::string, ValidateFn> validators = {
       {"requirepass",
        [this](const std::string &k, const std::string &v) -> Status {
-         if (v.empty() && !tokens.empty()) {
+         if (v.empty() && !load_tokens.empty()) {
            return {Status::NotOK, "requirepass empty not allowed while the namespace exists"};
          }
-         if (tokens.find(v) != tokens.end()) {
+         if (load_tokens.find(v) != load_tokens.end()) {
            return {Status::NotOK, "requirepass is duplicated with namespace tokens"};
          }
          return Status::OK();
        }},
       {"masterauth",
        [this](const std::string &k, const std::string &v) -> Status {
-         if (tokens.find(v) != tokens.end()) {
+         if (load_tokens.find(v) != load_tokens.end()) {
            return {Status::NotOK, "masterauth is duplicated with namespace tokens"};
          }
          return Status::OK();
@@ -515,6 +516,12 @@ void Config::initFieldCallback() {
          remove(nodes_file_path.data());
          return Status::OK();
        }},
+      {"repl-namespace-enabled",
+       [this](Server *srv, const std::string &k, const std::string &v) -> Status {
+         if (!srv || !cluster_enabled) return Status::OK();
+
+         return Status::OK();
+       }},
       {"rocksdb.target_file_size_base",
        [this](Server *srv, const std::string &k, const std::string &v) -> Status {
          if (!srv) return Status::OK();
@@ -682,7 +689,7 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
   if (strncasecmp(input.first.data(), ns_str, ns_str_size) == 0) {
     // namespace should keep key case-sensitive
     field_key = input.first;
-    tokens[input.second] = input.first.substr(ns_str_size);
+    load_tokens[input.second] = input.first.substr(ns_str_size);
     return Status::OK();
   }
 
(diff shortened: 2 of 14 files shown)


src/server/namespace.cc
```diff
@@ -0,0 +1,175 @@
...
+
+  // We would like to load namespaces from db even if repl_namespace_enabled is false,
+  // this can avoid missing some namespaces when turn on/off repl_namespace_enabled.
+  std::string value;
+  auto s = storage_->Get(rocksdb::ReadOptions(), cf_, kNamespaceDBKey, &value);
+  if (!s.ok()) {
+    if (s.IsNotFound()) return Status::OK();
+    return {Status::NotOK, s.ToString()};
+  }
+  jsoncons::json j = jsoncons::json::parse(value);
```
PragmaTwice: Maybe the json will be large and time-consuming while there are many namespaces.

I think here maybe we can store some key-values instead?
- `kNamespaceDBKey + token` -> `ns`
- `kNamespaceDBKey2 + ns` -> `token`

src/server/namespace.h
```diff
@@ -0,0 +1,51 @@
...
+  StatusOr<std::string> GetByToken(const std::string &token) const;
+  Status Set(const std::string &ns, const std::string &token);
+  Status Add(const std::string &ns, const std::string &token);
+  Status Del(const std::string &ns);
+  const std::map<std::string, std::string> &List() const { return tokens_; }
+
+ private:
+  engine::Storage *storage_;
+  rocksdb::ColumnFamilyHandle *cf_ = nullptr;
+  std::map<std::string, std::string> tokens_;
```
PragmaTwice, line 49:
```suggestion
  // from token to namespace name
  std::map<std::string, std::string> tokens_;
```

src/config/config.cc
```diff
@@ -159,6 +159,7 @@ Config::Config() {
       {"log-retention-days", false, new IntField(&log_retention_days, -1, -1, INT_MAX)},
       {"persist-cluster-nodes-enabled", false, new YesNoField(&persist_cluster_nodes_enabled, true)},
       {"redis-cursor-compatible", false, new YesNoField(&redis_cursor_compatible, false)},
+      {"repl-namespace-enabled", false, new YesNoField(&repl_namespace_enabled, false)},
```
PragmaTwice, line 162: What will happen if the master enables it but the replica does not, or vice versa?

src/server/server.cc
```diff
@@ -131,12 +131,16 @@ Server::~Server() {
 //     - feed-replica-data-info: generate checkpoint and send files list when full sync
 //     - feed-replica-file: send SST files when slaves ask for full sync
 Status Server::Start() {
+  auto s = namespace_.Load();
```
PragmaTwice: It is loaded despite the value of `repl-namespace-enabled`?

src/server/namespace.cc
```diff
@@ -0,0 +1,175 @@
...
+  for (const auto& iter : j.object_range()) {
+    if (tokens_.find(iter.key()) == tokens_.end()) {
+      // merge the namespace from db
+      tokens_[iter.key()] = iter.value().as<std::string>();
+    }
+  }
+  return rewriteOrWriteDB();
+}
+
+StatusOr<std::string> Namespace::Get(const std::string& ns) const {
```
PragmaTwice, line 76: I think if the call frequency of `Get`, `Set` ... is not so high, we can just get or set it directly in the storage via rocksdb api to avoid `Load` and `rewriteOrWriteDB` in many places? (after using a schema like https://github.com/apache/kvrocks/pull/1776/files#r1342042828)

kvrocks.conf
```diff
@@ -51,6 +51,14 @@ daemonize no
...
 cluster-enabled no
 
+# By default, namespaces are stored in the configuration file and won't be replicated
+# to replicas. This option allows to change this behavior, so that namespaces are also
+# propagated to slaves. Note that:
+# 1) it won't replicate the 'masterauth' to prevent breaking master/replica replication
+# 2) it will overwrite replica's namespace with master's namespace, so be careful of in-using namespaces.
+#
+# Default: no
+repl-namespace-enabled no
```
PragmaTwice, line 62: After a glance, it seems that even it is set to no, the namespace will also be written to the storage?

Correct me if I'm wrong.

git-hulk: It will not write to DB if this option is disabled, it returns in line: https://github.com/apache/kvrocks/blob/60d607b15fe663fb1791b784ece652238badc9d1/src/server/namespace.cc#L167

src/server/namespace.cc
git-hulk: Yes, this is a good point. I also thought about this solution, but I finally chose to write them in one key due to I would like to overwrite the slave's namespace/token if they do not exist in the master. And it's not a good practice to keep too many namespaces in one instance, maybe we can limit the max namespace number.

src/config/config.cc
git-hulk: This option is only to control if the current instance needs to replicate its namespace, so if the master enables it but slaves do not, then namespaces will be still replicated to slaves. And vice versa, all namespaces won't be replicated in master and slaves, since we don't allow changing namespace if the option is enabled and its role is slave.

src/server/server.cc
git-hulk: Yes, it will only load tokens from the configuration file if the repl-namespace-enabled is disabled.

src/server/namespace.cc
git-hulk: We need to check if the namespace/token is correct when authenticating, so we need to keep them in memory.

src/config/config.cc
PragmaTwice: got it.

src/server/namespace.cc
PragmaTwice: Got it. So you mean that it will affect the performance of the authentication if it is not in memory?

kvrocks.conf
PragmaTwice: Oh I see.

src/server/server.cc
PragmaTwice: https://github.com/apache/kvrocks/pull/1776/files#diff-e7439881acdda2d7b193156d6255f5691a76bd6b1bc69ddd46432a6c3903ad68R54 Here we load tokens from storage despite this value?

PragmaTwice:
It just come to me that if we firstly set `repl-namespace-enabled` to yes, use kvrocks in a period of time, and then change it to no, will it process correctly? e.g. will these tokens in the storage be removed?

src/server/namespace.cc
git-hulk: Yes, I guess it will affect the performance a lot if the users connect-write-close every time without keeping the connection.

src/server/server.cc
git-hulk: Yes, we still need to load tokens from DB even though it is disabled, because it may lose the db token if users switch from yes to no when restarting the server.

git-hulk:
> It just come to me that if we firstly set `repl-namespace-enabled` to yes, use kvrocks in a period of time, and then change it to no, will it process correctly? e.g. will these tokens in the storage be removed?

Good point, we should remove the DB key if this option switches from yes to no. But this may introduce another issue if the slave enables this configuration. So I think we can disallow changing it to `no` after turning it to yes.

PragmaTwice (approved):

PR #1787, opened 2023-09-28
Split functions in main.cc to multiple headers

PragmaTwice:
After the split, we can reuse these functions in other binaries, like kvrocks2redis.


git-hulk (approved):

torwig (approved):
LGTM.

PR #1790, opened 2023-10-01
Fix data race in instant metrics

### LINKED ISSUE #1789: Data race detected when recording metrics
### Search before asking

### Version

unstable

### Minimal reproduce step

https://github.com/apache/kvrocks/actions/runs/6363892540/job/17279898076?pr=1776

CI passed

### What did you see instead?

CI failed

### Anything Else?

_No response_

git-hulk:
Currently, only the server's cron thread will write the instant metrics
and those metrics are accessed by worker threads. So it may have a data race
between the server's cron and workers.

This closes #1789


PragmaTwice:
I am not sure if we can just disable these two function in TSAN since maybe we can bear some data race here.

git-hulk:
> I am not sure if we can just disable these two function in TSAN since maybe we can bear some data race here.

Yes, that's a nice solution as well. `GetInstantaneousMetric` only be used in the info command and those values are inside memory, so I guess it'd be good to have a read lock here.

PragmaTwice:
> Yes, that's a nice solution as well. `GetInstantaneousMetric` only be used in the info command and those values are inside memory, so I guess it'd be good to have a read lock here.

It seems TrackInstantaneousMetric is called in a server cron (every 100ms?), so I am wondering if it can downgrade the performance of the cron cycle.

git-hulk:
> > Yes, that's a nice solution as well. `GetInstantaneousMetric` only be used in the info command and those values are inside memory, so I guess it'd be good to have a read lock here.
> 
> It seems TrackInstantaneousMetric is called in a server cron (every 100ms?), so I am wondering if it can downgrade the performance of the cron cycle.

Yes, I feel good about this. How about every second?

PragmaTwice (approved):

torwig (approved):
Data race fix LGTM.

git-hulk:
Thanks all, merging...

Issue #1753, opened 2023-09-12
Add the support of loading data from RDB file


git-hulk:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

#1684 supports parsing serialized binary formats from Redis
which is also the key implementation of loading the Redis RDB file.
So after that, I think we can support loading data from the RDB file directly.

### Solution

Implement other parts of parsing the RDB file.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

xq2010:
I'm willing to submit a PR and will finish it next month.

git-hulk:
@xq2010 Assigned, feel free to ask any questions if you need.

xq2010:
@git-hulk  There is a commit in this pr:https://github.com/apache/kvrocks/pull/1798

git-hulk:
> @git-hulk There is a commit in this pr:#1798

Thanks for your efforts!

PR #1798, opened 2023-10-09
Add the support of the load RDB command

xq2010:
This PR is for https://github.com/apache/kvrocks/issues/1753


src/storage/rdb.h
```diff
@@ -47,26 +53,31 @@ constexpr const int RDBTypeZSetListPack = 17;
...
+class RdbStream;
+
+using RedisObjValue =
+    std::variant<std::string, std::vector<std::string>, std::vector<MemberScore>, std::map<std::string, std::string>>;
+
 class RDB {
  public:
-  explicit RDB(engine::Storage *storage, std::string ns, std::string_view input)
-      : storage_(storage), ns_(std::move(ns)), input_(input){};
+  explicit RDB(engine::Storage *storage, Config *config, std::string ns, std::shared_ptr<RdbStream> stream)
```
git-hulk: Can get config via storage's GetConfig.

git-hulk:
Hi @xq2010 Thanks for your excellent contribution first!

About how to load the RDB file, I'm wondering if it's better to export a command to do this. For example, we can support a new command like `load $type $path` which allows loading RDB file from local.

```shell
$ redis-cli 6666> load RDB /path/to/your/rdb_file
```

So that we don't need to introduce a new tool. How do you feel about this? cc @apache/kvrocks-committers

xq2010:
> Hi @xq2010 Thanks for your excellent contribution first!↳
> 
> About how to load the RDB file, I'm wondering if it's better to export a command to do this. For example, we can support a new command like `load $type $path` which allows loading RDB file from local.↳
> 
> ```shell
> $ redis-cli 6666> load RDB /path/to/your/rdb_file
> ```
> 
> So that we don't need to introduce a new tool. How do you feel about this? cc @apache/kvrocks-committers↳


 If we add the 'load rdb' command, we need to determine how to process an existing key that's present in both kvrocks and rdb, particularly when the key's type is a string in rdb. It will overwrite the existing key in kvrocks, which may be of a different type."

git-hulk:
> If we add the 'load rdb' command, we need to determine how to process an existing key that's present in both kvrocks and rdb, particularly when the key's type is a string in rdb. It will overwrite the existing key in kvrocks, which may be of a different type."

Yes, it's a good concern. I prefer overwriting by default if it exists and we can offer an option NX like the set command to let users determine in the future. This is my personal perspective, would like to hear more voices about the loading way design.

xq2010:
> Yes, it's a good concern. I prefer overwriting by default if it exists and we can offer an option NX like the set command to let users determine in the future. This is my personal perspective, would like to hear more voices about the loading way design.↳

Using NX is a good idea, I would also like to hear more opinions.

tests/testdata/corrupt_empty_keys.rdb
tisonkun, line 1: Generally, Apache release doesn't allow ship binary files. May you describe how these files are generated? Is it possible to generate test data file on the fly instead of committing them?

xq2010: These rdb files are copied from 'Redis's tests/asset/*.rdb'. If this is not allowed, I can generate the test data files on the fly

PragmaTwice:
Thanks for your contribution!

Agree with @git-hulk, I think it would be better to just implement it like a command inside the kvrocks process instead of a new program.

xq2010:
@git-hulk @PragmaTwice Okay, I will add a command to load the RDB file. The command might be rdb load \<path\> [namespace]. Do you have any suggestions or advice?

git-hulk:
> @git-hulk @PragmaTwice Okay, I will add a command to load the RDB file. The command might be rdb load <path> [namespace]. Do you have any suggestions or advice?

Thank you! Sounds good to me, but [namespace] seems unnecessary since it knows the current namespace once it enters the command.

xq2010:
> Thank you! Sounds good to me, but [namespace] seems unnecessary since it knows the current namespace once it enters the command.↳

Yes, it's unnecessary to use a namespace.

If the RDB file contains multiple db, the current implementation sets the namespace of each db to \<ns\>_\<db number\>. The password will be the same as the namespace. Only db 0 will use the current namespace. 
Perhaps using a parameter to specify which db to load would be better. If so, the command would be rdb load \<path\> [database number (default 0)].

git-hulk:
> Perhaps using a parameter to specify which db to load would be better. If so, the command would be rdb load <path> [database number (default 0)].

@xq2010 Yes, I think this is a good solution.

tests/testdata/corrupt_empty_keys.rdb
tisonkun: Yes please.

git-hulk:
@xq2010 Cool! The overall design is nice. I'll take another pass to take a look at detail implementation tomorrow.

CMakeLists.txt
```diff
@@ -253,4 +253,4 @@ file(GLOB_RECURSE TESTS_SRCS tests/cppunit/*.cc)
 add_executable(unittest ${TESTS_SRCS})
 target_include_directories(unittest PRIVATE tests/cppunit)
 
-target_link_libraries(unittest PRIVATE kvrocks_objs gtest_main ${EXTERNAL_LIBS})
```
tisonkun: Please revert this change.

tisonkun (commented):
Generally looks good. But I don't have time to check in details for giving an approval on the implementation 🤣

src/storage/rdb.cc
```diff
@@ -356,17 +378,28 @@ StatusOr<std::vector<MemberScore>> RDB::LoadZSetWithZipList() {
...
 
   // Check the checksum of the payload
-  GET_OR_RET(VerifyPayloadChecksum());
+  GET_OR_RET(VerifyPayloadChecksum(payload));
 
   auto type = GET_OR_RET(LoadObjectType());
+
+  auto value = GET_OR_RET(loadRdbObject(type, key));
+
+  return saveRdbObject(type, key, value, ttl_ms);  // NOLINT
```
git-hulk, line 393: Why do we need to add NOLINT for this line?

src/storage/rdb.cc
```diff
@@ -426,6 +443,50 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
       elements = GET_OR_RET(LoadListWithQuickList(type));
     }
+    return elements;
+  } else {
+    return {Status::RedisParseErr, fmt::format("unsupported type: {}", type)};
+  }
+
+  // can't be here
+  return Status::OK();
+}
```
git-hulk:
```suggestion
  }

  return {Status::RedisParseErr, fmt::format("unsupported type: {}", type)};
}
```

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+
+// Load RDB file: copy from redis/src/rdb.c:branch 7.0, 76b9c13d.
+Status RDB::LoadRdb(uint32_t db_index, bool is_nx) {
+  char buf[1024] = {0};
+  GET_OR_RETWITHLOG(stream_->Read(buf, 9));
+  buf[9] = '\0';
+
+  if (memcmp(buf, "REDIS", 5) != 0) {
+    LOG(WARNING) << "Wrong signature trying to load DB from file";
+    return {Status::NotOK};
```
git-hulk: Can we bring the error message into Status?

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+  }
+
+  auto rdb_ver = std::atoi(buf + 5);
+  if (rdb_ver < 1 || rdb_ver > RDBVersion) {
+    LOG(WARNING) << "Can't handle RDB format version " << rdb_ver;
+    return {Status::NotOK};
```
git-hulk: The same as the above

src/storage/rdb.h
```diff
@@ -87,16 +97,29 @@ class RDB {
...
-  Status peekOk(size_t n);
   StatusOr<double> loadBinaryDouble();
   StatusOr<double> loadDouble();
+
+  StatusOr<int> loadRdbType();
+  StatusOr<RedisObjValue> loadRdbObject(int rdbtype, const std::string &key);
+  Status saveRdbObject(int type, const std::string &key, const RedisObjValue &obj, uint64_t ttl_ms);
+  StatusOr<uint32_t> loadTime();
+  StatusOr<uint64_t> loadMillisecondTime(int rdb_version);
+  rocksdb::Status exist(const std::string &key);
```
git-hulk: exist => isKeyExisted

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+      LOG(WARNING) << "RDB file was saved with checksum disabled: no check performed.";
+    } else if (chk_sum != expected) {
+      LOG(WARNING) << "Wrong RDB checksum expected: " << chk_sum << " got: " << expected;
+      return {Status::NotOK, "Wrong RDB checksum"};
+    }
+  }
+
+  std::string skip_info = (is_nx ? ", exist keys skipped: " + std::to_string(skip_exist_keys) : "");
+
+  if (empty_keys_skipped > 0) {
```
git-hulk: I think we can print the empty keys info, no matter it's zero or not.

tests/cppunit/rdb_test.cc
```diff
@@ -0,0 +1,322 @@
...
+  // zset
+  s = exists("zset");
+  ASSERT_TRUE(s.IsNotFound());
+
+  s = exists("zset_ziplist");
+  ASSERT_TRUE(s.IsNotFound());
+
+  s = exists("zset_listpack");
+  ASSERT_TRUE(s.IsNotFound());
+}
```
git-hulk, line 312:
```suggestion
}

```

tests/cppunit/rdb_util.h
```diff
@@ -0,0 +1,203 @@
...
+  }
+
+  ScopedTestRDBFile(const ScopedTestRDBFile &) = delete;
+  ScopedTestRDBFile &operator=(const ScopedTestRDBFile &) = delete;
+
+  ~ScopedTestRDBFile() { std::remove(name_.c_str()); }
+
+ private:
+  std::string name_;
+};
```
git-hulk, line 203:
```suggestion
};

```

tests/cppunit/rdb_stream_test.cc
```diff
@@ -0,0 +1,90 @@
...
+
+  while (size >= len) {
+    ASSERT_EQ(reader.Read(buf, len).GetValue(), len);
+    size -= len;
+  }
+
+  if (size > 0) {
+    ASSERT_EQ(reader.Read(buf, size).GetValue(), size);
+  }
+}
```
git-hulk:
```suggestion
}

```

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+  redis::Database redis(storage_, ns_);
+  auto s = redis.Exists(keys, &cnt);
+  if (!s.ok()) {
+    return s;
+  }
+  if (cnt == 0) {
+    return rocksdb::Status::NotFound();
+  }
+  return rocksdb::Status::OK();
+}
```
git-hulk:
```suggestion
}

```

src/storage/rdb.cc
```diff
@@ -39,53 +43,69 @@
...
+#define GET_OR_RETWITHLOG(...)                                                                         \
+  ({                                                                                                   \
+    auto &&status = (__VA_ARGS__);                                                                     \
+    if (!status) {                                                                                     \
+      LOG(WARNING) << "Short read or unsupported type loading DB. Unrecoverable error, aborting now."; \
+      LOG(ERROR) << "Unexpected EOF reading RDB file";                                                 \
+      return std::forward<decltype(status)>(status);                                                   \
+    }                                                                                                  \
+    std::forward<decltype(status)>(status);                                                            \
+  }).GetValue()
```
git-hulk: @PragmaTwice What do you think about this implementation.

src/commands/cmd_server.cc
```diff
@@ -1072,6 +1074,46 @@ class CommandRestore : public Commander {
...
+
+    RDB rdb(svr->storage, conn->GetNamespace(), stream_ptr);
+    GET_OR_RET(rdb.LoadRdb(db_index_, is_nx_));
+
+    *output = redis::SimpleString("OK");
+    return Status::OK();
+  }
+
+ private:
+  bool is_nx_ = false;
```
torwig: I'm wondering if we can rename this parameter to `overwrite_existing` that will be `true` by default if I understood the meaning of the `is_nx` parameter.

src/common/rdb_stream.h
```diff
@@ -0,0 +1,80 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#pragma once
+
+#include <stdint.h>
```
torwig: I guess we can use `<cstdint>` here as a C++-header instead of C-like `<stdint.h>`.

src/storage/rdb.cc
PragmaTwice: I think it can just be replaced by something like `void logWhenError(v) { if (!v) LOG... }`.

We'd better to keep as less macros as we can.

src/commands/cmd_server.cc
```diff
@@ -1059,8 +1060,9 @@ class CommandRestore : public Commander {
       ttl_ms_ -= now;
     }
 
-    RDB rdb(svr->storage, conn->GetNamespace(), args_[3]);
-    auto s = rdb.Restore(args_[1], ttl_ms_);
+    auto stream_ptr = std::make_shared<RdbStringStream>(args_[3]);
```
PragmaTwice: Could you make a little explanation here about where this pointer need to be "share" for undetermined lifetime? Otherwise we'd better to use `unique_ptr` : )

src/commands/cmd_server.cc
```diff
@@ -1072,6 +1074,46 @@ class CommandRestore : public Commander {
   uint64_t ttl_ms_ = 0;
 };
 
+// command format: rdb load <path> [NX]  [DB index]
+class CommandRdb : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 3);
+    std::string_view ttl_flag, set_flag;
```
PragmaTwice: Seems these variables are useless?

src/common/rdb_stream.cc
```diff
@@ -0,0 +1,70 @@
...
+    return {Status::NotOK, fmt::format("failed to open rdb file: '{}': {}", file_name_, strerror(errno))};
+  }
+
+  return Status::OK();
+}
+
+StatusOr<size_t> RdbFileStream::Read(char *buf, size_t len) {
+  size_t n = 0;
+  while (len) {
+    size_t read_bytes = max_read_chunk_size_ < len ? max_read_chunk_size_ : len;
```
torwig: Perhaps, we can shorten this line by using `std::min(max_read_chunk_size_, len)`.

src/storage/rdb.cc
```diff
@@ -39,53 +43,69 @@
...
+
+Status RDB::VerifyPayloadChecksum(const std::string_view &payload) {
+  if (payload.size() < 10) {
     return {Status::NotOK, "invalid payload length"};
   }
-  auto footer = input_.substr(input_.size() - 10);
+  auto footer = payload.substr(payload.size() - 10);
   auto rdb_version = (footer[1] << 8) | footer[0];
   // For now, the max redis rdb version is 11
   if (rdb_version > 11) {
```
torwig: I think we can make the value `11` a constant.

src/storage/rdb.cc
```diff
@@ -39,53 +43,69 @@
...
     return {Status::NotOK, fmt::format("invalid or unsupported rdb version: {}", rdb_version)};
   }
-  uint64_t crc = crc64(0, reinterpret_cast<const unsigned char *>(input_.data()), input_.size() - 8);
-  memrev64ifbe(&crc);
+  auto crc = GET_OR_RET(stream_->GetCheckSum());
   if (memcmp(&crc, footer.data() + 2, 8)) {
```
torwig: Perhaps here we also can make `2` and `8` constants (to describe that something is a checksum padding and something is a checksum length).

src/storage/rdb.cc
```diff
@@ -95,20 +115,17 @@ StatusOr<uint64_t> RDB::loadObjectLen(bool *is_encoded) {
...
-    case RDB64BitLen:
-      GET_OR_RET(peekOk(8));
-      __builtin_memcpy(&len, input_.data() + pos_, 8);
-      pos_ += 8;
-      return len;
+      return (len << 8) | GET_OR_RET(stream_->ReadByte());
     default:
-      return {Status::NotOK, fmt::format("Unknown length encoding {} in loadObjectLen()", type)};
+      if (c == RDB32BitLen) {
+        GET_OR_RET(stream_->Read(reinterpret_cast<char *>(&len), 4));
```
torwig: The same is related to `4` here and `8`  down the code.

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
   }
   return db_status.ok() ? Status::OK() : Status{Status::RedisExecErr, db_status.ToString()};
 }
+
+StatusOr<uint32_t> RDB::loadTime() {
```
torwig: You can name the function more descriptive like `loadExpirationTimeSeconds`.

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+  uint32_t t32 = 0;
+  GET_OR_RET(stream_->Read(reinterpret_cast<char *>(&t32), 4));
+  return t32;
+}
+
+StatusOr<uint64_t> RDB::loadMillisecondTime(int rdb_version) {
```
torwig: You can name the function more descriptive like `loadExpirationTimeMilliseconds`.

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+    return {Status::NotOK};
+  }
+
+  auto rdb_ver = std::atoi(buf + 5);
+  if (rdb_ver < 1 || rdb_ver > RDBVersion) {
+    LOG(WARNING) << "Can't handle RDB format version " << rdb_ver;
+    return {Status::NotOK};
+  }
+
+  uint64_t expire_time = 0;
```
torwig, line 563:
```
uint64_t expire_time = 0;
int64_t expire_keys = 0;
int64_t load_keys = 0;
int64_t empty_keys_skipped = 0;
uint64_t skip_exist_keys = 0;
```

Minor renaming to the following:
```
uint64_t expiration_time = 0;
int64_t expired_keys = 0;
int64_t loaded_keys = 0;
int64_t skipped_empty_keys = 0;
uint64_t skipped_existing_keys = 0;
```

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+    auto ret = saveRdbObject(type, key, value, expire_time);
+    if (!ret.IsOK()) {
+      LOG(WARNING) << "save rdb object key " << key << " failed: " << ret.Msg();
+    } else {
+      load_keys++;
+    }
+  }
+
+  // Verify the checksum if RDB version is >= 5
+  if (rdb_ver >= 5) {
```
torwig: Make `5` as a constant (something like `kMinRdbVersionToVerifyChecksum` or something shorter).

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+  GET_OR_RETWITHLOG(stream_->Read(buf, 9));
+  buf[9] = '\0';
+
+  if (memcmp(buf, "REDIS", 5) != 0) {
+    LOG(WARNING) << "Wrong signature trying to load DB from file";
+    return {Status::NotOK};
+  }
+
+  auto rdb_ver = std::atoi(buf + 5);
+  if (rdb_ver < 1 || rdb_ver > RDBVersion) {
```
torwig: Transform `1` into the `kMinRdbVersion` constant.

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+  // Verify the checksum if RDB version is >= 5
+  if (rdb_ver >= 5) {
+    uint64_t chk_sum = 0;
+    auto expected = GET_OR_RETWITHLOG(stream_->GetCheckSum());
+    GET_OR_RETWITHLOG(stream_->Read(reinterpret_cast<char *>(&chk_sum), 8));
+    if (chk_sum == 0) {
+      LOG(WARNING) << "RDB file was saved with checksum disabled: no check performed.";
+    } else if (chk_sum != expected) {
+      LOG(WARNING) << "Wrong RDB checksum expected: " << chk_sum << " got: " << expected;
+      return {Status::NotOK, "Wrong RDB checksum"};
```
torwig: So we check the checksum to return an error? What would be the result? As I understand from the code, all objects will be loaded and the function `RDB::LoadRdb` will return an error without any rollback. Can we write something like "All objects were processed and loaded but the checksum is unexpected" in the response to the client?

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+    LOG(INFO) << "Done loading RDB,  keys loaded: " << load_keys << ", keys expired:" << expire_keys
+              << ", empty keys skipped: " << empty_keys_skipped << skip_info;
+  } else {
+    LOG(INFO) << "Done loading RDB,  keys loaded: " << load_keys << ", keys expired:" << expire_keys << skip_info;
+  }
+
+  return Status::OK();
+}
+
+rocksdb::Status RDB::exist(const std::string &key) {
```
torwig: I would rename this function to `keyExist` or as @git-hulk suggested `isKeyExist`.

src/storage/rdb.cc
```diff
@@ -449,3 +510,188 @@ Status RDB::Restore(const std::string &key, uint64_t ttl_ms) {
...
+  int cnt = 0;
+  std::vector<rocksdb::Slice> keys;
+  keys.emplace_back(key);
+  redis::Database redis(storage_, ns_);
+  auto s = redis.Exists(keys, &cnt);
```
torwig: I was thinking of introducing the function `Database::KeyExist` to make the check, in this case, simpler without creating `std::vector` and renaming `Database::Exists` to `Database::MultipleKeysExist`. However, I'm not sure about it. @git-hulk @PragmaTwice What do you think?

torwig (commented):
@xq2010 Good job, thank you for your effort. Just a little nitpicking from me :)

src/common/rdb_stream.h
```diff
@@ -0,0 +1,80 @@
...
+class RdbStream {
+ public:
+  RdbStream() = default;
+  virtual ~RdbStream() = default;
+
+  virtual StatusOr<size_t> Read(char *buf, size_t len) = 0;
+  virtual StatusOr<uint64_t> GetCheckSum() const = 0;
+  StatusOr<uint8_t> ReadByte() {
+    uint8_t value = 0;
+    auto s = Read(reinterpret_cast<char *>(&value), 1);
```
PragmaTwice, line 39: not sure if we need to consider endian here.

src/storage/rdb.cc
git-hulk: Yes,this solution will be better.

xq2010:
Thanks for your review, I will review it later

src/storage/rdb.h
xq2010: The last commit has deleted the config parameter

src/storage/rdb.cc
xq2010: The lint report falsely indicates a memory leak, but I can't find any potential for a memory leak

src/storage/rdb.cc
xq2010: yes, accepted

src/storage/rdb.cc
xq2010: okay

src/storage/rdb.cc
xq2010: Okay

src/commands/cmd_server.cc
xq2010: Yes, it's better to use unique_ptr.

src/commands/cmd_server.cc
xq2010: Yes, I copied it and forgot to delete.

src/common/rdb_stream.cc
xq2010: Okay

src/common/rdb_stream.h
xq2010: I don't think so. It only reads one byte. If you plan to use it to read multiple bytes, you should determine how to handle endianness

src/storage/rdb.cc
git-hulk: ok, it should be if the ASAN in CI is good.

src/common/rdb_stream.cc
```diff
@@ -0,0 +1,70 @@
...
+  pos_ += n;
+  return n;
+}
+
+StatusOr<uint64_t> RdbStringStream::GetCheckSum() const {
+  if (input_.size() < 8) {
+    return {Status::NotOK, "invalid payload length"};
+  }
+  uint64_t crc = crc64(0, reinterpret_cast<const unsigned char *>(input_.data()), input_.size() - 8);
+  memrev64ifbe(&crc);
```
mapleFU, line 40: It's a bit tricky here. Should we `memrev64ifbe` for the crc64 of the Stream?

src/common/rdb_stream.cc
```diff
@@ -0,0 +1,70 @@
...
+
+StatusOr<size_t> RdbFileStream::Read(char *buf, size_t len) {
+  size_t n = 0;
+  while (len) {
+    size_t read_bytes = max_read_chunk_size_ < len ? max_read_chunk_size_ : len;
+    ifs_.read(buf, static_cast<std::streamsize>(read_bytes));
+    if (!ifs_.good()) {
+      return Status(Status::NotOK, fmt::format("read failed: {}:", strerror(errno)));
+    }
+    check_sum_ = crc64(check_sum_, (const unsigned char *)buf, read_bytes);
```
mapleFU: using `static_cast` for  `buf`?

src/common/rdb_stream.cc
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  return Status::OK();
+}
+
+StatusOr<size_t> RdbFileStream::Read(char *buf, size_t len) {
+  size_t n = 0;
+  while (len) {
+    size_t read_bytes = max_read_chunk_size_ < len ? max_read_chunk_size_ : len;
+    ifs_.read(buf, static_cast<std::streamsize>(read_bytes));
```
mapleFU, line 58: https://en.cppreference.com/w/cpp/io/basic_istream/read
Do we need to handle the exception here?

src/common/rdb_stream.h
```diff
@@ -0,0 +1,80 @@
...
+ public:
+  explicit RdbFileStream(std::string file_name, size_t chunk_size = 1024 * 1024)
+      : file_name_(std::move(file_name)), check_sum_(0), total_read_bytes_(0), max_read_chunk_size_(chunk_size){};
+  RdbFileStream(const RdbFileStream &) = delete;
+  RdbFileStream &operator=(const RdbFileStream &) = delete;
+  ~RdbFileStream() override = default;
+
+  Status Open();
+  StatusOr<size_t> Read(char *buf, size_t len) override;
+  StatusOr<uint64_t> GetCheckSum() const override { return check_sum_; }
```
mapleFU: Actually I think `GetCheckSum()` here is a bit tricky. Can we use `std::optional<uint64_t> check_sum_` as member, and return `Error` when `check_sum_` is unset?

src/common/rdb_stream.h
PragmaTwice: Oh, didn't notice it. Looks fine then.

src/storage/rdb.cc
xq2010: I believe the original version is better :)

src/storage/rdb.cc
xq2010: Yes, if the checksum does not match, we will not perform a rollback. In Redis, if a checksum error occurs, the server will exit during start-up. However, as a command to load the RDB, it is not suitable to exit. Instead, I will add an error message to be sent to the client.

src/common/rdb_stream.cc
xq2010: yes, we should, , that's what Redis does.

src/common/rdb_stream.h
xq2010: i don't think so, if not set, it should not return an error

git-hulk:
Hi @xq2010 How's this PR going? Should we take another round review?

xq2010:
@git-hulk  Thank you for your review. I've made some corrections. Could you please review it once more?

git-hulk:
> @git-hulk Thank you for your review. I've made some corrections. Could you please review it once more?

Sure, thanks for your efforts!

git-hulk (approved):

git-hulk:
I would like to merge this PR today if no more comments.

mapleFU:
I'm a bit busy these few days. I'll go through it before Thursday. You can merge this first if other doesn't have any comments.

torwig (approved):
Generally, LGTM.

src/common/rdb_stream.cc
```diff
@@ -0,0 +1,69 @@
...
+ * under the License.
+ *
+ */
+
+#include "rdb_stream.h"
+
+#include "fmt/format.h"
+#include "vendor/crc64.h"
+
+StatusOr<size_t> RdbStringStream::Read(char *buf, size_t n) {
```
mapleFU, line 26: The syntax of `RdbStringStream` is different from `RdbFileStream`. The previous one will return a smaller size when n is greater than remaining buffer, but `RdbFileStream` will always return `Status::NotOK`. Is this expected?

src/storage/rdb.cc
```diff
@@ -156,10 +173,9 @@ StatusOr<std::string> RDB::loadEncodedString() {
   }
 
   // Normal string
-  GET_OR_RET(peekOk(static_cast<size_t>(len)));
-  auto value = std::string(input_.data() + pos_, len);
-  pos_ += len;
-  return value;
+  std::vector<char> vec(len);
+  GET_OR_RET(stream_->Read(vec.data(), len));
```
mapleFU, line 177: Could it be `len == 0`?

src/common/rdb_stream.cc
mapleFU: Oh I make it wrong. So they're same.

src/storage/rdb.cc
xq2010: Yes, it is possible, could you  help me fix it?

mapleFU: https://github.com/apache/kvrocks/pull/1839
I've create a pull request for it.

Thanks again for your contribution!

PR #1795, opened 2023-10-07
[WIP] Implement Xreadgroup

LiuYuHui:
This PR is still in a very early stage, it is not ready for review.

PR #1799, opened 2023-10-11
Add support of the XGROUP CREATECONSUMER command

Yangsx-1:
This PR finish [XGROUP CREATECONSUMER](https://redis.io/commands/xgroup-createconsumer/).
The implementation of XGROUP DELCONSUMER may rely on the design of pending list, so it is better to implement it after XREADGROUP finished.

### DIFF
--- a/src/types/redis_stream.cc
+++ b/src/types/redis_stream.cc
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "db_util.h"
+#include "time_util.h"
 
 namespace redis {
 
@@ -211,6 +212,26 @@ StreamConsumerGroupMetadata Stream::decodeStreamConsumerGroupMetadataValue(const
   return consumer_group_metadata;
 }
 
+std::string Stream::internalKeyFromConsumerName(const std::string &ns_key, const StreamMetadata &metadata,
+                                                const std::string &group_name, const std::string &consumer_name) const {
+  std::string sub_key;
+  PutFixed64(&sub_key, group_name.size());
+  sub_key += group_name;
+  PutFixed64(&sub_key, consumer_name.size());
+  sub_key += consumer_name;
+  sub_key += consumerGroupMetadataDelimiter;
+  std::string entry_key = InternalKey(ns_key, sub_key, metadata.version, storage_->IsSlotIdEncoded()).Encode();
+  return entry_key;
+}
+
+std::string Stream::encodeStreamConsumerMetadataValue(const StreamConsumerMetadata &consumer_metadata) {
+  std::string dst;
+  PutFixed64(&dst, consumer_metadata.pending_number);
+  PutFixed64(&dst, consumer_metadata.last_idle);
+  PutFixed64(&dst, consumer_metadata.last_active);
+  return dst;
+}
+
 rocksdb::Status Stream::CreateGroup(const Slice &stream_name, const StreamXGroupCreateOptions &options,
                                     const std::string &group_name) {
   if (std::isdigit(group_name[0])) {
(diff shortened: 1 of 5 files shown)


LiuYuHui (approved):
LGTM, maybe we can merge this PR first, I will update the pending entries list part in https://github.com/apache/kvrocks/pull/1795 ASAP!

PragmaTwice:
Thanks for your contribution!

Seems there is a failure in the golang test cases, could you try to locate and fix it?

Yangsx-1:
> 
> Seems there is a failure in the golang test cases, could you try to locate and fix it?

Fixed.

src/types/redis_stream.cc
```diff
@@ -313,6 +334,62 @@ rocksdb::Status Stream::DestroyGroup(const Slice &stream_name, const std::string
...
+  consumer_metadata.last_active = now;
+  std::string consumer_key = internalKeyFromConsumerName(ns_key, metadata, group_name, consumer_name);
+  std::string consumer_value = encodeStreamConsumerMetadataValue(consumer_metadata);
+  std::string get_consumer_value;
+  s = storage_->Get(rocksdb::ReadOptions(), stream_cf_handle_, consumer_key, &get_consumer_value);
+  if (!s.IsNotFound()) {
+    if (!s.ok()) {
+      return s;
+    }
+    return rocksdb::Status::OK();
```
mapleFU: Hmmm can we simplify this branch?

mapleFU (commented):
General looks ok to me, just some nit

src/types/redis_stream.cc
Yangsx-1: I thought about this, we can move `if (!s.ok())` out of the outer `if`, but this may cause the if condition longer.
The implementation now may be simpler.

mapleFU:
```
  if (!s.IsNotFound()) {
    // if s.ok(), it means it already ...
    return s;
  }
```

Would code like this ok?

Yangsx-1: Oh, that would be nice!

torwig (dismissed):
Generally, LGTM.

src/types/redis_stream.cc
```diff
@@ -313,6 +334,59 @@ rocksdb::Status Stream::DestroyGroup(const Slice &stream_name, const std::string
...
+  }
+
+  std::string entry_key = internalKeyFromGroupName(ns_key, metadata, group_name);
+  std::string get_entry_value;
+  s = storage_->Get(rocksdb::ReadOptions(), stream_cf_handle_, entry_key, &get_entry_value);
+  if (!s.ok() && !s.IsNotFound()) {
+    return s;
+  }
+  if (s.IsNotFound()) {
+    return rocksdb::Status::InvalidArgument("NOGROUP No such consumer group" + group_name + "for key name" +
```
git-hulk:
```suggestion
    return rocksdb::Status::InvalidArgument("NOGROUP No such consumer group " + group_name + " for key name " +
```

git-hulk:
left one comment, others are good to me. Thank you!

src/types/redis_stream.cc
```diff
@@ -313,6 +334,59 @@ rocksdb::Status Stream::DestroyGroup(const Slice &stream_name, const std::string
...
+  auto now = util::GetTimeStampMS();
+  consumer_metadata.last_idle = now;
+  consumer_metadata.last_active = now;
+  std::string consumer_key = internalKeyFromConsumerName(ns_key, metadata, group_name, consumer_name);
+  std::string consumer_value = encodeStreamConsumerMetadataValue(consumer_metadata);
+  std::string get_consumer_value;
+  s = storage_->Get(rocksdb::ReadOptions(), stream_cf_handle_, consumer_key, &get_consumer_value);
+  if (!s.IsNotFound()) {
+    return s;
+  }
```
mapleFU, line 374: Should we also check `!s.ok()` after `IsNotFound`?

mapleFU (approved):

git-hulk (approved):

git-hulk:
Thanks all, merging..

PR #1803, opened 2023-10-15
Add support of new command JSON.SET and JSON.GET

PragmaTwice:
We add support of two new commands `JSON.SET` and `JSON.GET` (despite some options missing).

Besides, we add two new methods `IsSingleKVType` and `IsEmptyableType` to class `Metadata` for more intuitive coding.

JSON Metadata encoding:
```
key => flag | expire | format | payload
```

The field `format` is to make JSON data type more extensible. Currently it can only be `0` to indicate that the data stored in `payload` is in the JSON format. 

TODO:
- missing options (maybe in next PR)
- compatible issue between jsoncons jsonpath and redis jsonpath (maybe leave as FIXME)
- tests

### DIFF
--- a/src/types/json.h
+++ b/src/types/json.h
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#pragma once
+
+#include <jsoncons/json.hpp>
+#include <jsoncons/json_error.hpp>
+#include <jsoncons_ext/jsonpath/json_query.hpp>
+
+#include "jsoncons_ext/jsonpath/jsonpath_error.hpp"
+#include "status.h"
+
+struct JsonValue {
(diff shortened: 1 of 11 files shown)


git-hulk (approved):
LGTM, thanks for your great work!

git-hulk:
We can create an issue to track other new commands. And others can help to implement the new commands if they like.

src/types/json.h
```diff
@@ -0,0 +1,77 @@
...
+
+struct JsonValue {
+  JsonValue() = default;
+  explicit JsonValue(jsoncons::basic_json<char> value) : value(std::move(value)) {}
+
+  static StatusOr<JsonValue> FromString(std::string_view str) {
+    jsoncons::json val;
+
+    try {
+      val = jsoncons::json::parse(str);
```
mapleFU, line 38: I think we'd better use `parse` with `erroc` version to avoid costing from the exception.

src/types/json.h
```diff
@@ -0,0 +1,77 @@
...
+    return JsonValue(std::move(val));
+  }
+
+  std::string Dump() const { return value.to_string(); }
+
+  Status Set(std::string_view path, JsonValue &&new_value) {
+    try {
+      jsoncons::jsonpath::json_replace(value, path, [&new_value](const std::string & /*path*/, jsoncons::json &origin) {
+        origin = std::move(new_value.value);
+      });
```
mapleFU, line 52: Assume the path matches mutlple sub json object, this will be wrong

mapleFU (commented):
General it would be a good point to start, so just some nits.

src/types/json.h
PragmaTwice: I think since it follows *preorder traversal* so it will goes well when there are mutlple sub json objects matched in the original object.

But it will be wrong if there are  json objects matched in the new_value, which will be fixed later.

src/types/json.h
PragmaTwice: It will be fixed in later PRs, or you can pick up it if you have free time : )

PragmaTwice:
Tests will be added in further PRs to make an example/reference for other issues in the tracking issue : )

src/types/json.h
mapleFU: I don't think so, since `new_value.value` has been moved.

PragmaTwice: Oh, good point. I will fix it.

PR #1837, opened 2023-10-19
Add the support of the json.arrappend command

### LINKED ISSUE #1806: Add support for the JSON.ARRAPPEND command
Please refer to #1805.

lieck:
This PR adds support for JSON.ARRAPEND command.
Closes #1806

### DIFF
--- a/src/commands/cmd_json.cc
+++ b/src/commands/cmd_json.cc
@@ -52,7 +52,37 @@ class CommandJsonGet : public Commander {
   }
 };
 
+class CommandJsonArrAppend : public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::Json json(svr->storage, conn->GetNamespace());
+
+    std::vector<uint64_t> result_count;
+
+    auto s = json.ArrAppend(args_[1], args_[2], {args_.begin() + 3, args_.end()}, result_count);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    if(result_count.empty()) {
+      result_count.emplace_back(0);
+    }
+
+    std::vector<std::string> result_values;
+    result_values.reserve(result_count.size());
+    for(auto c : result_count) {
+      if(c != 0) {
+        result_values.emplace_back("(integer) " + std::to_string(c));
+      } else {
+        result_values.emplace_back();
+      }
+    }
+
+    *output = redis::MultiBulkString(result_values, true);
+    return Status::OK();
+  }
+};
+
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandJsonSet>("json.set", -3, "write", 1, 1, 1),
-                        MakeCmdAttr<CommandJsonGet>("json.get", -2, "read-only", 1, 1, 1), );
+                        MakeCmdAttr<CommandJsonGet>("json.get", -2, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandJsonArrAppend>("json.arrappend", -4, "write", 1, 1, 1), );
 
 }  // namespace redis
--- a/src/commands/cmd_json.cc
+++ b/src/commands/cmd_json.cc
@@ -59,24 +59,22 @@ class CommandJsonArrAppend : public Commander {
 
     std::vector<uint64_t> result_count;
 
-    auto s = json.ArrAppend(args_[1], args_[2], {args_.begin() + 3, args_.end()}, result_count);
+    auto s = json.ArrAppend(args_[1], args_[2], {args_.begin() + 3, args_.end()}, &result_count);
     if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
 
-    if(result_count.empty()) {
+    if (result_count.empty()) {
       result_count.emplace_back(0);
     }
 
-    std::vector<std::string> result_values;
-    result_values.reserve(result_count.size());
-    for(auto c : result_count) {
-      if(c != 0) {
-        result_values.emplace_back("(integer) " + std::to_string(c));
+    *output = redis::MultiLen(result_count.size());
+    for (uint64_t c : result_count) {
+      if (c != 0) {
+        *output += redis::Integer(c);
       } else {
-        result_values.emplace_back();
+        *output += redis::NilString();
       }
     }
 
-    *output = redis::MultiBulkString(result_values, true);
     return Status::OK();
   }
 };
--- a/src/commands/cmd_json.cc
+++ b/src/commands/cmd_json.cc
@@ -62,10 +62,6 @@ class CommandJsonArrAppend : public Commander {
     auto s = json.ArrAppend(args_[1], args_[2], {args_.begin() + 3, args_.end()}, &result_count);
     if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
 
-    if (result_count.empty()) {
-      result_count.emplace_back(0);
-    }
-
     *output = redis::MultiLen(result_count.size());
     for (uint64_t c : result_count) {
       if (c != 0) {
--- a/src/types/json.h
+++ b/src/types/json.h
@@ -74,18 +74,17 @@ struct JsonValue {
   }
 
   Status ArrAppend(std::string_view path, const std::vector<jsoncons::json> &append_values,
-                   std::vector<uint64_t> &result_count) {
+                   std::vector<uint64_t> *result_count) {
     try {
-      std::vector<std::string> paths;
-      jsoncons::jsonpath::json_replace(value, path, [&append_values, &result_count]
-                                       (const std::string &path, jsoncons::json &val) {
-        if(val.is_array()) {
-          val.insert(val.array_range().end(), append_values.begin(), append_values.end());
-          result_count.emplace_back(val.size());
-        } else {
-          result_count.emplace_back(0);
-        }
-      });
+      jsoncons::jsonpath::json_replace(
+          value, path, [&append_values, result_count](const std::string &path, jsoncons::json &val) {
+            if (val.is_array()) {
+              val.insert(val.array_range().end(), append_values.begin(), append_values.end());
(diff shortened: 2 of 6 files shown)


src/types/json.h
```diff
@@ -73,6 +73,25 @@ struct JsonValue {
     }
   }
 
+  Status ArrAppend(std::string_view path, const std::vector<jsoncons::json> &append_values,
+                   std::vector<uint64_t> &result_count) {
```
mapleFU: Prefer `std::vector<uint64_t>* result_count` to specify it's an output type

src/types/json.h
```diff
@@ -73,6 +73,25 @@ struct JsonValue {
...
+    try {
+      std::vector<std::string> paths;
```
mapleFU: `paths` Not used?

src/types/redis_json.cc
```diff
@@ -117,4 +117,38 @@ rocksdb::Status Json::Get(const std::string &user_key, const std::vector<std::st
...
+    return rocksdb::Status::NotSupported("JSON storage format not supported");
+
+  auto value_res = JsonValue::FromString(rest.ToStringView());
+  if(!value_res) return rocksdb::Status::Corruption(value_res.Msg());
+  auto value = *std::move(value_res);
+
+  auto append_res = value.ArrAppend(path, append_values, result_count);
+  if(!append_res) return rocksdb::Status::InvalidArgument(append_res.Msg());
+
+  return write(ns_key, &metadata, value);
```
mapleFU, line 154: Can we optimize the case when nothing in `value` has been changed?

src/types/redis_json.cc
```diff
@@ -117,4 +117,38 @@ rocksdb::Status Json::Get(const std::string &user_key, const std::vector<std::st
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status Json::ArrAppend(const std::string &user_key, const std::string &path,
+                                const std::vector<std::string> &values,
+                                std::vector<uint64_t> &result_count) {
+  auto ns_key = AppendNamespacePrefix(user_key);
+
+  std::vector<jsoncons::json> append_values;
```
mapleFU, line 124: better `reserve` here?

src/commands/cmd_json.cc
```diff
@@ -52,7 +52,37 @@ class CommandJsonGet : public Commander {
...
+    auto s = json.ArrAppend(args_[1], args_[2], {args_.begin() + 3, args_.end()}, result_count);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    if(result_count.empty()) {
+      result_count.emplace_back(0);
+    }
+
+    std::vector<std::string> result_values;
+    result_values.reserve(result_count.size());
+    for(auto c : result_count) {
```
mapleFU:
```suggestion
    for(uint64_t c : result_count) {
```

Since `uint64_t` is not a too complex type, maybe we'd better using literal rather than auto

mapleFU (commented):
General looks good to me, just some nits.

src/commands/cmd_json.cc
```diff
@@ -52,7 +52,37 @@ class CommandJsonGet : public Commander {
...
+      if(c != 0) {
+        result_values.emplace_back("(integer) " + std::to_string(c));
```
mapleFU: I think we can just use `redis::Integer` and `redis::Nil`

mapleFU:
> The current of CommandJsonArrAppend::Execute uses MultiBulkString. Is there a need to provide an MultiBulkInteger?

https://github.com/redis/redis-specifications/blob/master/protocol/RESP2.md#resp-integers Redis has a resp protocol, and we don't need to using "(Integer)" literal. Just using `redis::Integer` is ok here.

By the way, you should also add some integration test in `tests/gocase`. It will test the command from the client side. You can refer to `unit/type/json/json_test.go`

tests/cppunit/types/json_test.cc
```diff
@@ -119,3 +119,42 @@ TEST_F(RedisJsonTest, Get) {
...
+  ASSERT_EQ(json_val_.Dump(), R"({"x":[1,2,{"z":3},1],"y":[]})");
+  res.clear();
+
+  ASSERT_TRUE(json_->ArrAppend(key_, "$..y", {"1", "2", "3"}, &res).ok());
+  ASSERT_EQ(res.size(), 1);
+  ASSERT_EQ(res[0], 3);
+  ASSERT_TRUE(json_->Get(key_, {}, &json_val_).ok());
+  ASSERT_EQ(json_val_.Dump(), R"({"x":[1,2,{"z":3},1],"y":[1,2,3]})");
+  res.clear();
+
```
mapleFU, line 150: Do we need add a test for key is not exist?

src/commands/cmd_json.cc
```diff
@@ -52,7 +52,35 @@ class CommandJsonGet : public Commander {
...
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::Json json(svr->storage, conn->GetNamespace());
+
+    std::vector<uint64_t> result_count;
+
+    auto s = json.ArrAppend(args_[1], args_[2], {args_.begin() + 3, args_.end()}, &result_count);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    if (result_count.empty()) {
```
mapleFU: So this is to make a "nothing updated" `ArrAppend` to return a `Nil`?

src/types/redis_json.cc
```diff
@@ -117,4 +117,41 @@ rocksdb::Status Json::Get(const std::string &user_key, const std::vector<std::st
...
+                                const std::vector<std::string> &values, std::vector<uint64_t> *result_count) {
+  auto ns_key = AppendNamespacePrefix(user_key);
+
+  std::vector<jsoncons::json> append_values;
+  append_values.reserve(values.size());
+  for (auto &v : values) {
+    auto value_res = JsonValue::FromString(v);
+    if (!value_res) return rocksdb::Status::InvalidArgument(value_res.Msg());
+    auto value = *std::move(value_res);
+    append_values.emplace_back(value.value);
```
mapleFU: You can just `emplace_back( ... the moved value )`, I guess current copy might copy the value?

mapleFU (commented):
Thanks! Just few questions

src/commands/cmd_json.cc
lieck: Yes, that's right. :)

mapleFU:
```
    r.assertOk(r.execute_command('JSON.SET', 'doc1', '$', json.dumps(types_data)))
    res = r.execute_command('JSON.GET', 'doc1', '$')
    r.assertEqual([types_data], json.loads(res))

    r.assertEqual(r.execute_command('HSET', 'hash_key', 'a', '1', 'b', '2'), 2)

    # Notice: redis client is parsing error responses and trimming prefixes such as 'ERR'

    # ARRAPPEND
    r.assertEqual(r.execute_command('JSON.ARRAPPEND', 'doc1', '$.string', '"abc"'), [None])
    r.assertEqual(r.execute_command('JSON.ARRAPPEND', 'doc1', '$.nowhere', '"abc"'), [])
    r.expect('JSON.ARRAPPEND', 'doc_none', '$.string', '"abc"').raiseError().contains("doesn't exist")
    r.expect('JSON.ARRAPPEND', 'hash_key', '$.string', '"abc"').raiseError().contains("wrong Redis type")
```

https://github.com/RedisJSON/RedisJSON/blob/master/tests/pytest/test_multi.py#L1029-L1039

Though we don't need to compatible with RedisJson( I don't think we need to return same error code, lol), but seems that it return the empty when nothing has matched.

mapleFU: I guess you can:

```
std::vector<std::string> results;
results.reserve(result_count.size());
    for (uint64_t c : result_count) {
      if (c != 0) {
        results.push_back(redis::Integer(c));
      } else {
        results.push_back(redis::NilString());
      }
    }
   *output = MultiBulkString(...);
```

(The code above might failed to compile or doesn't have right style, just to show the way here.)

So that we can simplify our code, and use `MultiBulkString` to handle the return value with zero counts.

lieck: Thank you, using `MultiBulkString` will invoke `BulkString` for each element in `results`, making the return result not an integer array.
```cpp
std::string MultiBulkString(const std::vector<std::string> &values, const std::vector<rocksdb::Status> &statuses) {
  std::string result = "*" + std::to_string(values.size()) + CRLF;
  for (size_t i = 0; i < values.size(); i++) {
    if (i < statuses.size() && !statuses[i].ok()) {
      result += NilString();
    } else {
      result += BulkString(values[i]);
    }
  }
  return result;
}
```

mapleFU (dismissed):

src/commands/cmd_json.cc
mapleFU: Yeah the way you're using is right, my bad

mapleFU:
Thanks @lieck . The patch lgtm!

But you still need to fix the golang-lint:

```
unit/type/json/json_test.go:60: File is not `gofmt`-ed with `-s` (gofmt)
	    require.NoError(t, rdb.Do(ctx, "SET", "a", `1`).Err())
        require.Error(t, rdb.Do(ctx, "JSON.ARRAPPEND", "a", "$", `1`).Err())
        require.NoError(t, rdb.Do(ctx, "DEL", "a").Err())
$ /home/runner/go/bin/golangci-lint run -v ./...
Traceback (most recent call last):
```

You can just run `go fmt` to format the file.

mapleFU (approved):
Thanks @lieck . I'll wait for other committers comment, and will merge it if no other doesn't have comments on this and ci passes.

mapleFU:
cc @git-hulk @PragmaTwice

git-hulk (approved):

mapleFU:
Thanks @lieck . Merged now

PR #1840, opened 2023-10-19
Support format options for `JSON.GET`

PragmaTwice:
- support format options `INDENT` `SPACE` and `NEWLINE` for command`JSON.GET`
- add a configuration option `json_max_nesting_depth`

### DIFF
--- a/src/commands/cmd_json.cc
+++ b/src/commands/cmd_json.cc
@@ -18,7 +18,10 @@
  *
  */
 
+#include <algorithm>
+
 #include "commander.h"
+#include "commands/command_parser.h"
 #include "server/redis_reply.h"
 #include "server/server.h"
 #include "types/redis_json.h"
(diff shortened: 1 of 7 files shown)


src/commands/cmd_json.cc
```diff
@@ -40,16 +43,57 @@ class CommandJsonSet : public Commander {
...
+        }
+
+        indent_size_ = indent.size();
+      } else if (parser.EatEqICase("NEWLINE")) {
+        new_line_chars_ = GET_OR_RET(parser.TakeStr());
+      } else if (parser.EatEqICase("SPACE")) {
+        auto space = GET_OR_RET(parser.TakeStr());
+
+        if (space != "" && space != " ") {
+          return {Status::RedisParseErr, "Currently only SPACE ' ' is supported"};
```
mapleFU, line 64: what about empty?

PragmaTwice: I think we do not need to emphasize it in error message. User should just not pass this option.

git-hulk (dismissed):

git-hulk (dismissed):

src/commands/cmd_json.cc
```diff
@@ -40,16 +43,57 @@ class CommandJsonSet : public Commander {
...
 class CommandJsonGet : public Commander {
  public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 2);
+
+    while (parser.Good()) {
+      if (parser.EatEqICase("INDENT")) {
+        auto indent = GET_OR_RET(parser.TakeStr());
+
+        if (std::any_of(indent.begin(), indent.end(), [](char v) { return v != ' '; })) {
```
mapleFU, line 53: Won't `std::all_of` be more clear here?

src/types/json.h
```diff
@@ -22,35 +22,76 @@
...
 
-#include "jsoncons_ext/jsonpath/jsonpath_error.hpp"
 #include "status.h"
 
 struct JsonValue {
   JsonValue() = default;
   explicit JsonValue(jsoncons::basic_json<char> value) : value(std::move(value)) {}
 
-  static StatusOr<JsonValue> FromString(std::string_view str) {
+  static StatusOr<JsonValue> FromString(std::string_view str, int max_nesting_depth = std::numeric_limits<int>::max()) {
```
mapleFU, line 36: Should we add a constexpr rather than direct using `std::numeric_limits<int>::max()`? Since it's difference from default 1024?

mapleFU (approved):

src/commands/cmd_json.cc
PragmaTwice: little difference to me

src/types/json.h
PragmaTwice: I think it is like INT_MAX, which do not require a constant variable.

Issue #1824, opened 2023-10-15
Add support for the JSON.STRLEN command


PragmaTwice:
Please refer to #1805.

PR #1841, opened 2023-10-20
Add support for JSON.STRAPPEND and JSON.STRLEN command

### LINKED ISSUE #1823: Add support for the JSON.STRAPPEND command

guojidan:
This closes #1823 #1824


src/commands/cmd_json.cc
```diff
@@ -52,7 +52,33 @@ class CommandJsonGet : public Commander {
...
+
+      *output = redis::Integer(cnt);
+      return Status::OK();
+    }
+};
+
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandJsonSet>("json.set", -3, "write", 1, 1, 1),
-                        MakeCmdAttr<CommandJsonGet>("json.get", -2, "read-only", 1, 1, 1), );
+                        MakeCmdAttr<CommandJsonGet>("json.get", -2, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandJsonStrAppend>("json.strappend", -2, "read-only", 1, 1, 1), );
```
mapleFU: I don't know why arity is -2. The syntax for it is:

```
JSON.STRAPPEND key [path] value
```

Seems that the arity should be `-3` since it should have at lease `JSON.STRAPPEND, key, value` ?

src/commands/cmd_json.cc
```diff
@@ -52,7 +52,33 @@ class CommandJsonGet : public Commander {
...
 };
 
+class CommandJsonStrAppend : public Commander {
+public:
+    Status Execute(Server *svr, Connection *conn, std::string *output) override {
+      redis::Json json(svr->storage, conn->GetNamespace());
+
+      JsonValue result;
+      std::vector<std::string> paths;
+      std::vector<std::string> values;
```
mapleFU: Seems that `path` and `value` should be `std::string`?

You can also take a look at `CommandParser`.

```
std::string path_or_root = "$";
std::string value;
if (args_.size() == 3) {
  value = ...
} else {
  path = 
  value = 
}
```

src/commands/cmd_json.cc
```diff
@@ -52,7 +52,33 @@ class CommandJsonGet : public Commander {
...
+      for (int i=2;i<(int)args_.size();++i) {
+        auto val = args_[i];
+        if (val.substr(0, 1) == "$") {
+          paths.push_back(val);
+        } else{
+          values.push_back(val);
+        }
+      }
+      uint64_t cnt = 0;
```
mapleFU: The return value should be an array of int like https://github.com/apache/kvrocks/pull/1837 rather than a int

tests/gocase/unit/type/json/json_test.go
```diff
@@ -34,7 +35,7 @@ func TestJson(t *testing.T) {
 	rdb := srv.NewClient()
 	defer func() { require.NoError(t, rdb.Close()) }()
 
-	t.Run("JSON.SET and JSON.GET basics", func(t *testing.T) {
+	t.Run("JSON.SET and JSON.GET and JSON.STRAPPEND basics", func(t *testing.T) {
```
mapleFU: You'd better start a new test here?

```
t.Run("JSON.STRAPPEND tests", func(t *testing.T) {})
```

src/types/redis_json.cc
```diff
@@ -117,4 +117,30 @@ rocksdb::Status Json::Get(const std::string &user_key, const std::vector<std::st
...
+
+  auto json_res = JsonValue::FromString(rest.ToStringView());
+  if (!json_res) return rocksdb::Status::Corruption(json_res.Msg());
+  auto json_val = *std::move(json_res);
+
+  for (const auto &path : paths) {
+    auto append_res = json_val.Append(path, value, deleted_cnt);
+    if (!append_res) return rocksdb::Status::InvalidArgument(append_res.Msg());
+  }
+  return write(ns_key, &metadata, json_val);
```
mapleFU, line 489: Can we avoid a `write` when nothing changed in underlying json (.i.e `deleted_count == 0` in your case )

src/commands/cmd_json.cc
guojidan: should be -4, at lease `JSON.STRAPPEND, key, path, value`, should have one `path`

mapleFU: From the RedisJson code impl[1] and document[3], I think it's -3

1. https://redis.io/commands/json.strappend/
2. https://github.com/RedisJSON/RedisJSON/blob/d4ba815fb13412c879b2cd3cd29ab60aee3c211b/redis_json/src/commands.rs#L939

But I think you can also use `4`. I think it's not a good idea that RedisJson make `[path]` as optional...

guojidan: ok, thanks

tests/gocase/unit/type/json/json_test.go
```diff
@@ -54,5 +55,13 @@ func TestJson(t *testing.T) {
 		require.Equal(t, rdb.Do(ctx, "JSON.GET", "a", "$").Val(), `[{"x":1,"y":{"x":{"y":2},"y":3}}]`)
 		require.Equal(t, rdb.Do(ctx, "JSON.GET", "a", "$..x").Val(), `[1,{"y":2}]`)
 		require.Equal(t, rdb.Do(ctx, "JSON.GET", "a", "$..x", "$..y").Val(), `{"$..x":[1,{"y":2}],"$..y":[{"x":{"y":2},"y":3},3,2]}`)
+
+		require.NoError(t, rdb.Do(ctx, "JSON.SET", "a", "$", `{"a":"foo", "nested": {"a": "hello"}, "nested2": {"a": 31}}`).Err())
+		require.Equal(t, rdb.Do(ctx, "JSON.STRAPPEND", "a", "$.a", "ba").Val(), int64(1))
+        require.NoError(t, rdb.Do(ctx, "JSON.SET", "a", "$", `{"a":"foo", "nested": {"a": "hello"}, "nested2": {"a": 31}}`).Err())
```
PragmaTwice: Please fix the format : )

guojidan:
I fixed the problem you pointed out : )

src/types/json.h
```diff
@@ -65,6 +66,29 @@ struct JsonValue {
...
+        if (origin.is_string()) {
+          auto origin_str = origin.to_string();
+          // need trim "
+          auto origin_trimmed = util::Trim(origin_str, "\"");
+          len = origin_trimmed.length() + append_trimmed.length();
+          std::string new_value;
+          new_value.append("\"").append(origin_trimmed).append(append_trimmed).append("\"");
+          origin = jsoncons::json::parse(new_value);
+        }
+        append_cnt.push_back(len);
```
```suggestion
        } else {
          append_cnt.push_back(0);
        }
```

src/types/redis_json.cc
```diff
@@ -117,4 +117,39 @@ rocksdb::Status Json::Get(const std::string &user_key, const std::vector<std::st
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status Json::StrAppend(const std::string &user_key, const std::vector<std::string> &paths, const std::string &value, std::vector<uint64_t> &append_cnt) {
```
git-hulk: Path should be std::string?

src/commands/cmd_json.cc
```diff
@@ -52,7 +52,24 @@ class CommandJsonGet : public Commander {
   }
 };
 
+class CommandJsonStrAppend : public Commander {
+public:
+    Status Execute(Server *svr, Connection *conn, std::string *output) override {
+      redis::Json json(svr->storage, conn->GetNamespace());
+
+      std::vector<uint64_t> cnt;
```
git-hulk: can rename cnt to others like results because cnt looks like a single value.

src/types/json.h
guojidan: `len` initialization 0，so there logic is right

src/types/redis_json.cc
guojidan: Get && Set function`s path is std::string，Can you explain your doubt?

git-hulk: Yes, my point is should use std::string instead of const std::vector<std::string>.

src/types/json.h
git-hulk: Yes, but using `0` should be more intuitive, we don't care how the length was initialized.

src/types/redis_json.cc
guojidan: but, strAppend is support multipath

src/types/json.h
guojidan: ok, I understand

src/types/redis_json.cc
git-hulk: It should be only one or none: https://redis.io/commands/json.strappend/

guojidan: `[path]` means option, I misunderstood... ,I thought means a array

src/types/json.h
```diff
@@ -65,6 +66,47 @@ struct JsonValue {
...
+
+  Status StrLen(std::string_view path, std::vector<uint64_t> &str_lens) const {
+    try {
+      jsoncons::jsonpath::json_query(value, path,
+                                     [&str_lens](const std::string & /*path*/, const jsoncons::json &origin) {
+                                       if (origin.is_string()) {
+                                         // need subtraction "
+                                         str_lens.push_back(origin.to_string().length() - 2);
+                                       } else {
+                                         str_lens.push_back(0);
```
PragmaTwice: How can you distinguish between zero-sized array and non-array here?

src/types/json.h
```diff
@@ -65,6 +66,47 @@ struct JsonValue {
...
+    return Status::OK();
+  }
+
+  Status StrLen(std::string_view path, std::vector<uint64_t> &str_lens) const {
+    try {
+      jsoncons::jsonpath::json_query(value, path,
+                                     [&str_lens](const std::string & /*path*/, const jsoncons::json &origin) {
+                                       if (origin.is_string()) {
+                                         // need subtraction "
+                                         str_lens.push_back(origin.to_string().length() - 2);
```
PragmaTwice: IMHO even you trim these two quotation marks here, I think the result still can be incorrect. e.g. `"\n"` -> `2`

src/types/json.h
```diff
@@ -65,6 +66,47 @@ struct JsonValue {
...
+              auto origin_str = origin.to_string();
+              // need trim "
+              auto origin_trimmed = util::Trim(origin_str, "\"");
+              append_cnt.push_back(origin_trimmed.length() + append_trimmed.length());
+              std::string new_value;
+              new_value.append("\"").append(origin_trimmed).append(append_trimmed).append("\"");
+              origin = jsoncons::json::parse(new_value);
+            } else {
+              append_cnt.push_back(0);
+            }
```
PragmaTwice: IMHO it is not the correct way to process the string object in the JSON document.

You are expected to use `as<string>`, rather than `to_string`.

`to_string` is to re-serialize the object, not to retrieve the value.

PragmaTwice (dismissed):

src/types/json.h
guojidan: ohhh,  should be use `as<std::string>`  or `as_string`, it will automatically process `"`

src/types/json.h
guojidan: no need distinguish? I don't quite understand what you mean.
...
127.0.0.1:6666> json.set doc $ '{"a":"foo\n", "nested": {"a": "hello"}, "nested2": {"a": 31}, "nested3": [{"a": "world"}], "nested4": {"a": []}}'
OK
127.0.0.1:6666> json.strlen doc $..a
1) (integer) 4
2) (integer) 5
3) (nil)
4) (integer) 5
5) (nil)

...

PragmaTwice:
```
> json.set x $ '["", 1]'
OK
> json.strlen x $[0]
1) (integer) 0
> json.strlen x $[1]
1) (nil)
```

guojidan: I understand, need distinguish between null and zero, zero is not null

src/commands/cmd_json.cc
```diff
@@ -171,10 +171,49 @@ class CommandJsonClear : public Commander {
...
+    std::vector<int64_t> results;
+    auto s = json.StrLen(args_[1], path, results);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = redis::IntegerArray(results);
+    return Status::OK();
+  }
+};
+
+REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandJsonSet>("json.set", -3, "write", 1, 1, 1),
```
PragmaTwice: Seems this line is changed unintentionally.

src/server/redis_reply.cc
```diff
@@ -34,6 +34,18 @@ std::string BulkString(const std::string &data) { return "$" + std::to_string(da
 
 std::string NilString() { return "$-1" CRLF; }
 
+std::string IntegerArray(const std::vector<int64_t> &values, bool output_nil_for_negative) {
```
PragmaTwice: IMHO the `output_nil_for_negative` is too specific to be used in other places.

I do not suggest you put this function in `redis_reply.h`.

src/commands/cmd_json.cc
guojidan: yes，I am overwrite someone else's code when rebase

src/types/json.h
```diff
@@ -106,6 +108,44 @@ struct JsonValue {
...
+            if (origin.is_string()) {
+              auto origin_trimmed = origin.as<std::string>();
+              append_cnt.push_back((int64_t)(origin_trimmed.length() + append_trimmed.length()));
+              std::string new_value;
+              new_value.append("\"").append(origin_trimmed).append(append_trimmed).append("\"");
+              origin = jsoncons::json::parse(new_value);
+            } else {
+              append_cnt.push_back(-1);
+            }
+          });
```
PragmaTwice, line 125: Same as https://github.com/apache/kvrocks/pull/1841#discussion_r1367656620.

And I think it will be a **blocking issue** of this PR.

PragmaTwice: It confuses me actually. Don't know why you need to manually concat these strings with `"`.

It will be a **blocking issue** of this PR.

PragmaTwice (commented):
Even if you want to use `-1` as a flag number, you DO NOT need to change unsigned integer type to signed integer. It is useless and make the situation more complicted.

If you don't get this point, you can refer to https://en.cppreference.com/w/cpp/string/basic_string/npos.

It also gonna be a **blocking issue** of this PR.

src/types/json.h
guojidan: `jsoncons::json::parse()` need `"string"`, use `string` will throw err

guojidan:
> Even if you want to use `-1` as a flag number, you DO NOT need to change unsigned integer type to signed integer. It is useless and make the situation more complicted.
> 
> If you don't get this point, you can refer to https://en.cppreference.com/w/cpp/string/basic_string/npos.
> 
> It also gonna be a **blocking issue** of this PR.

I understand

src/types/json.h
PragmaTwice: Here it does not need a `parse()`. You can check the document of jsoncons for more details.

src/server/redis_reply.cc
guojidan: I use `std::string::npos` instead negative, I think this func be used for struct RESP,  put in `redis_reply.h` is ok

PragmaTwice: You do not need to use `npos`. It is just a reference of the concept.

guojidan: signed integer `-1` wil conversion to largest positive value, ohhh, thanks for your help

PragmaTwice:
Hi @guojidan, before we review this PR, you can first try to solve these conflicts with unstable.

guojidan:
> Hi @guojidan, before we review this PR, you can first try to solve these conflicts with unstable.

hi @PragmaTwice Thank you very much for your help, now I already rebase this pr

src/commands/cmd_json.cc
```diff
@@ -204,11 +204,62 @@ class CommandJsonArrLen : public Commander {
...
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = IntegerArray(results);
+    return Status::OK();
+  }
+
+  static std::string IntegerArray(const std::vector<uint64_t> &values, bool output_nil_for_negative = true) {
+    std::string result = "*" + std::to_string(values.size()) + CRLF;
+    for (const auto &value : values) {
+      if (value == (uint64_t)-1 and output_nil_for_negative) {
```
PragmaTwice: `(uint64_t)-1` is NOT a negative number. You can just remove `output_nil_for_negative`.

src/commands/cmd_json.cc
```diff
@@ -204,11 +204,62 @@ class CommandJsonArrLen : public Commander {
...
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::Json json(svr->storage, conn->GetNamespace());
+
+    // path default, if not provided
+    std::string path = "$";
+    if (args_.size() == 4) {
+      path = args_[2];
+    }
+    std::vector<uint64_t> results;
+    auto s = json.StrAppend(args_[1], path, args_[3], results);
```
PragmaTwice: The argument processing is wrong if `size() != 4`.

src/types/json.h
```diff
@@ -106,6 +108,43 @@ struct JsonValue {
     return Status::OK();
   }
 
+  Status StrAppend(std::string_view path, const std::string &append_value, std::vector<uint64_t> &append_cnt) {
+    try {
+      auto append_trimmed = util::Trim(append_value, "\"");
```
PragmaTwice: It is wrong to just trim `"`. You need to parse it.

PragmaTwice (commented):
Sorry but it still has several issues which need to be addressed before merging.

src/types/redis_json.cc
```diff
@@ -201,4 +201,63 @@ rocksdb::Status Json::ArrLen(const std::string &user_key, const std::string &pat
...
+  Slice rest;
+  auto s = GetMetadata(kRedisJson, ns_key, &bytes, &metadata, &rest);
+  if (!s.ok()) return s;
+
+  if (metadata.format != JsonStorageFormat::JSON)
+    return rocksdb::Status::NotSupported("JSON storage format not supported");
+
+  auto json_res = JsonValue::FromString(rest.ToStringView());
+  if (!json_res) return rocksdb::Status::Corruption(json_res.Msg());
+  auto json_val = *std::move(json_res);
```
PragmaTwice: Could you use the `read` method instead these code? You can refer to other method for how to use `read`.

src/types/redis_json.cc
```diff
@@ -201,4 +201,63 @@ rocksdb::Status Json::ArrLen(const std::string &user_key, const std::string &pat
...
+  JsonMetadata metadata;
+  Slice rest;
+  auto s = GetMetadata(kRedisJson, ns_key, &bytes, &metadata, &rest);
+  if (!s.ok()) return s;
+
+  if (metadata.format != JsonStorageFormat::JSON)
+    return rocksdb::Status::NotSupported("JSON storage format not supported");
+
+  auto json_res = JsonValue::FromString(rest.ToStringView());
+  if (!json_res) return rocksdb::Status::Corruption(json_res.Msg());
```
PragmaTwice: ditto

src/types/redis_json.cc
guojidan: ok

src/types/redis_json.cc
PragmaTwice: Seems it is not changed yet : )

guojidan: sorry, it too careless

tests/cppunit/types/json_test.cc
```diff
@@ -272,3 +273,36 @@ TEST_F(RedisJsonTest, ArrLen) {
...
+
+  append_cnt_.clear();
+  ASSERT_TRUE(json_->Set(key_, "$", R"({"a":"foo", "nested": {"a": "hello"}, "nested2": {"a": 31}})").ok());
+  ASSERT_TRUE(json_->StrLen(key_, "$..a", append_cnt_).ok());
+  ASSERT_EQ(append_cnt_.size(), 3);
+  int64_t result1[] = {3, 5, -1};
+  for (int i = 0; i < 3; ++i) {
+    ASSERT_EQ(append_cnt_[i], result1[i]);
+  }
+}
```
PragmaTwice:
```suggestion
}

```

src/types/json.h
```diff
@@ -144,6 +146,50 @@ struct JsonValue {
     return Status::OK();
   }
 
+  Status StrAppend(std::string_view path, const std::string &append_value, std::vector<uint64_t> &append_cnt) {
+    try {
+      std::string append_trimmed;
```
PragmaTwice: I think `append_str` may be a better name.

src/types/json.h
```diff
@@ -144,6 +146,50 @@ struct JsonValue {
...
+
+    return Status::OK();
+  }
+
+  Status StrLen(std::string_view path, std::vector<uint64_t> &str_lens) const {
+    try {
+      jsoncons::jsonpath::json_query(value, path,
+                                     [&str_lens](const std::string & /*path*/, const jsoncons::json &origin) {
+                                       if (origin.is_string()) {
+                                         str_lens.push_back((int64_t)origin.as<std::string>().length());
```
PragmaTwice: Why does it need to be casted to `int64_t` here?

src/types/json.h
```diff
@@ -144,6 +146,50 @@ struct JsonValue {
...
+      if (append_json.is_string()) {
+        append_trimmed = append_json.as_string();
+      } else {
+        return {Status::NotOK, "STRAPPEND need input a string to append"};
+      }
+
+      jsoncons::jsonpath::json_replace(
+          value, path, [&append_trimmed, &append_cnt](const std::string & /*path*/, jsoncons::json &origin) {
+            if (origin.is_string()) {
+              auto origin_trimmed = origin.as<std::string>();
```
PragmaTwice: ditto

src/types/redis_json.cc
```diff
@@ -230,4 +230,43 @@ rocksdb::Status Json::ArrLen(const std::string &user_key, const std::string &pat
...
+  if (!s.ok()) return s;
+
+  auto append_res = json_val.StrAppend(path, value, append_cnt);
+  if (!append_res) return rocksdb::Status::InvalidArgument(append_res.Msg());
+  // need Reverse order
+  std::reverse(append_cnt.begin(), append_cnt.end());
+
+  bool need_overwrite = false;
+  for (auto append : append_cnt) {
+    if (append > 0) {
```
PragmaTwice: The `max` situation is missed here.

src/types/redis_json.cc
```diff
@@ -230,4 +230,43 @@ rocksdb::Status Json::ArrLen(const std::string &user_key, const std::string &pat
...
+                                std::vector<uint64_t> &append_cnt) {
+  auto ns_key = AppendNamespacePrefix(user_key);
+  JsonMetadata metadata;
+  JsonValue json_val;
+  auto s = read(ns_key, &metadata, &json_val);
+  if (!s.ok()) return s;
+
+  auto append_res = json_val.StrAppend(path, value, append_cnt);
+  if (!append_res) return rocksdb::Status::InvalidArgument(append_res.Msg());
+  // need Reverse order
```
PragmaTwice: Please avoid UPPERCASE without a reason.

src/types/json.h
```diff
@@ -144,6 +146,50 @@ struct JsonValue {
...
+              append_cnt.push_back(origin_trimmed.length() + append_trimmed.length());
+              jsoncons::json new_value(origin_trimmed + append_trimmed);
+              origin = new_value;
```
```suggestion
              origin = origin_str + append_str;
```

src/types/redis_json.cc
```diff
@@ -230,4 +230,43 @@ rocksdb::Status Json::ArrLen(const std::string &user_key, const std::string &pat
...
+  std::reverse(append_cnt.begin(), append_cnt.end());
```
PragmaTwice, line 477: I am curious why we need reverse here but not in `StrLen`?

src/types/json.h
guojidan: but, other functions in `json.h` use Camel-Case， like `ArrAppend`

PragmaTwice: I mean the variable name.

src/types/redis_json.cc
guojidan: `jsoncons::jsonpath::json_replace()` look like is reverse order, like `ArrAppend`: 

> 127.0.0.1:6666> json.set doc $ '{"x":[1],"y":[1, 2], "z": {"y": [1]}}'
OK
127.0.0.1:6666> json.arrappend doc $..y 1
> 1) (integer) 2
> 2) (integer) 3

git-hulk:
@guojidan Are you still working on this PR?

guojidan:
> @guojidan Are you still working on this PR?



> @guojidan Are you still working on this PR?

yes, this pr wait review

git-hulk (dismissed):

git-hulk:
Some places need to be improved, but I will do it in another PR:

- Keep consistent of the integer receiver, should be `std::vector<std::optional<uint64_t>>`.
- Add a function to covert `std::vector<std::optional<uint64_t>>` to the Redis output to avoid duplicate code blocks.

To avoid PR being pending too long and those improvements are code style-related. So it's good for me to approve this PR first and improve it later.

git-hulk:
@guojidan Cloud you please resolve conflicts again

git-hulk (approved):

Issue #1802, opened 2023-10-13
Add the support of changing worker threads in-flight


git-hulk:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

It's hard to know the application workload as the time goes,
so that's impossible to estimate the right number of worker threads
to fulfill the workload requirement. So it'd be great if we can change
the worker thread number in-flight, and we don't need to worry much about
it's suitable at the beginning.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

guojidan:
provide a server command to change worker thread num in-flight?

guojidan:
assign to me : )

PragmaTwice:
Hi @guojidan , assigned!

BTW here we can just use `CONFIG SET ..` instead of a new command.

Refer to [config.cc](https://github.com/apache/kvrocks/blob/unstable/src/config/config.cc) for more details.

guojidan:
thanks

xiaobiaozhao:
@guojidan Do you have any ideas? I'm interested in this. I think we can talk about this.

guojidan:
I have implemented a version，but I don't think it's good，Do you have any better ideas？ https://github.com/apache/kvrocks/commit/4b627a99cbbb071f9e00d57d9e1ccfe52ead88bd

xiaobiaozhao:
If I look at this commit, there is nothing wrong with the big logic, what needs to be considered is how to do thread exit.

guojidan:
Is it limited worker_num to greater than 1, I think 1 workers is Meaningless ,  and I think long connections can be transferred to other threads on exit

guojidan:
take look @xiaobiaozhao  https://github.com/guojidan/kvrocks/commit/3b0c4b7d9190c5ffc6d1f3a4658cdf48e75bfdba

guojidan:
I can't handle long connections properly, so let other people interested in it implement this issue, I will close pr and unassign to me.

git-hulk:
@guojidan Thanks for your efforts. Let me take it and welcome to help review if you like.

PR #1842, opened 2023-10-20
Change workers in-flight

guojidan:
support of changing worker threads in-flight

Related #1802

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -105,7 +105,7 @@ Config::Config() {
       {"tls-session-cache-timeout", false, new IntField(&tls_session_cache_timeout, 300, 0, INT_MAX)},
       {"tls-replication", true, new YesNoField(&tls_replication, false)},
 #endif
-      {"workers", true, new IntField(&workers, 8, 1, 256)},
+      {"workers", false, new IntField(&workers, 8, 1, 256)},
       {"timeout", false, new IntField(&timeout, 0, 0, INT_MAX)},
       {"tcp-backlog", true, new IntField(&backlog, 511, 0, INT_MAX)},
       {"maxclients", false, new IntField(&maxclients, 10240, 0, INT_MAX)},
@@ -301,6 +301,19 @@ void Config::initFieldValidator() {
          }
          return Status::OK();
        }},
+      {"workers",
+       [](const std::string &k, const std::string &v) -> Status{
+         auto result_parse = ParseInt<int>(v, 10);
+         if (!result_parse) {
+           return {Status::NotOK, "Invalid arg type, need Integer"};
+         }
+         int new_workers = *result_parse;
+
+         if (new_workers < 1) {
+           return {Status::NotOK, "workers number must more than 1"};
+         }
+         return Status::OK();
+       }},
   };
   for (const auto &iter : validators) {
     auto field_iter = fields_.find(iter.first);
@@ -399,6 +412,21 @@ void Config::initFieldCallback() {
        [this](Server *srv, const std::string &k, const std::string &v) -> Status {
          std::vector<std::string> args = util::Split(v, " \t");
          binds = std::move(args);
+         return Status::OK();
+       }},
+      {"workers",
+       [this](Server *srv, const std::string &k, const std::string &v) -> Status{
+         // for Config::load()
+         if (!srv) return Status::OK();
+
+         auto result_parse = ParseInt<int>(v, 10);
+         if (!result_parse) {
+           return {Status::NotOK, "Invalid arg type, need Integer"};
+         }
+         int new_workers = *result_parse;
+         Status s = srv->AdjustWorkerThreadNum(new_workers);
+         if (!s.IsOK()) return s;
+
          return Status::OK();
        }},
       {"maxclients",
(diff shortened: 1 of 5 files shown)


src/server/server.cc
```diff
@@ -232,8 +232,12 @@ void Server::Stop() {
 }
 
 void Server::Join() {
-  for (const auto &worker : worker_threads_) {
-    worker->Join();
+//  for (const auto &worker : worker_threads_) {
+//    worker->Join();
+//    LOG(INFO) << "worker_threads_ len: " <<worker_threads_.size();
+//  }
+  for (int i = 0; i < (int )worker_threads_.size(); ++i) {
```
torwig, line 239: Could you please use `auto` / `size_t` instead of `int` here?

src/server/server.cc
```diff
@@ -232,8 +232,12 @@ void Server::Stop() {
 }
 
 void Server::Join() {
-  for (const auto &worker : worker_threads_) {
-    worker->Join();
+//  for (const auto &worker : worker_threads_) {
```
torwig, line 235: This commented code should be removed.

src/server/server.cc
```diff
@@ -1863,3 +1867,77 @@ std::string Server::GetKeyNameFromCursor(const std::string &cursor, CursorType c
 
   return {};
 }
+
+Status Server::AdjustWorkerThreadNum(int new_num) {
+  int now_num = (int)worker_threads_.size();
+  LOG(INFO) << __func__ << ", now_num: " << now_num << ", new_num: " << new_num;
```
torwig, line 1873: I guess it's here just for usage only during the development process?

src/server/server.cc
```diff
@@ -1863,3 +1867,77 @@ std::string Server::GetKeyNameFromCursor(const std::string &cursor, CursorType c
...
+
+  if (now_num == new_num) {
+    return Status::OK();
+  } else if (now_num < new_num) {
```
torwig, line 1877: You delete `else` and write the rest of the logic without indentation due to an early return before.

torwig:
@guojidan You can run `./x.py format` to format your code before pushing (to pass the CI checks).

src/config/config.cc
```diff
@@ -301,6 +301,19 @@ void Config::initFieldValidator() {
          }
          return Status::OK();
        }},
+      {"workers",
+       [](const std::string &k, const std::string &v) -> Status{
+         auto result_parse = ParseInt<int>(v, 10);
+         if (!result_parse) {
+           return {Status::NotOK, "Invalid arg type, need Integer"};
+         }
+         int new_workers = *result_parse;
```
PragmaTwice, line 310:
```suggestion
         int new_workers = GET_OR_RET(ParseInt<int>(v, 10));
```

src/config/config.cc
```diff
@@ -301,6 +301,19 @@ void Config::initFieldValidator() {
...
+
+         if (new_workers < 1) {
+           return {Status::NotOK, "workers number must more than 1"};
+         }
```
PragmaTwice, line 314: if you mean ***more than 1*** here, the condition should be `new_workers <= 1`

src/config/config.cc
```diff
@@ -301,6 +301,19 @@ void Config::initFieldValidator() {
          }
          return Status::OK();
        }},
+      {"workers",
+       [](const std::string &k, const std::string &v) -> Status{
```
PragmaTwice, line 305: Actually I think maybe there is no necessary to give a validator, since these checks are trivial and already done in `IntField`.

src/config/config.cc
```diff
@@ -399,6 +412,21 @@ void Config::initFieldCallback() {
...
+      {"workers",
+       [this](Server *srv, const std::string &k, const std::string &v) -> Status{
+         // for Config::load()
+         if (!srv) return Status::OK();
+
+         auto result_parse = ParseInt<int>(v, 10);
+         if (!result_parse) {
+           return {Status::NotOK, "Invalid arg type, need Integer"};
+         }
+         int new_workers = *result_parse;
```
PragmaTwice, line 426:
```suggestion
         int new_workers = GET_OR_RET(ParseInt<int>(v, 10));
```

src/config/config.cc
```diff
@@ -399,6 +412,21 @@ void Config::initFieldCallback() {
...
+         Status s = srv->AdjustWorkerThreadNum(new_workers);
+         if (!s.IsOK()) return s;
+
          return Status::OK();
```
PragmaTwice, line 430:
```suggestion
         return srv->AdjustWorkerThreadNum(new_workers);
```

PragmaTwice:
Hi @guojidan , thanks for your contribution.

Could you add some golang test cases for this patch?

src/server/worker.cc
```diff
@@ -322,6 +327,10 @@ Status Worker::AddConnection(redis::Connection *c) {
   return Status::OK();
 }
 
+std::map<int, redis::Connection *> Worker::GetConnection() {
+  return std::move(conns_);
```
PragmaTwice, line 331: Is it intentional to clear `conns_` after this function? IMHO it is weird to move class field in a getter function.

git-hulk:
It's unnecessary to allow changing worker's number for the Unix socket listening mode because it cannot be shared like the TCP socket. Maybe we can take the reduce and increase worker's number into two functions which would make the function clear.

PR #1850, opened 2023-10-22
Support for the JSON.CLEAR command

### LINKED ISSUE #1812: Add support for the JSON.CLEAR command

2rueSid:
support JSON.CLEAR, this closes #1812 
please review @mapleFU

[JSON.CLEAR command](https://redis.io/commands/json.clear)

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -135,7 +135,7 @@ log-retention-days -1
 # When running in daemonize mode, kvrocks writes a PID file in ${CONFIG_DIR}/kvrocks.pid by
 # default. You can specify a custom pid file location here.
 # pidfile /var/run/kvrocks.pid
-pidfile ""
+pidfile /tmp/kvrocks/kvrocks.pid
 
 # You can configure a slave instance to accept writes or not. Writing against
 # a slave instance may be useful to store some ephemeral data (because data
@@ -859,3 +859,14 @@ rocksdb.rate_limiter_auto_tuned yes
 
 ################################ NAMESPACE #####################################
 # namespace.test change.me
+auto-resize-block-and-sst yes
+backup-dir /tmp/kvrocks/backup
+force-compact-file-age 172800
+force-compact-file-min-deleted-percentage 10
+fullsync-recv-file-delay 0
+persist-cluster-nodes-enabled yes
+replica-announce-port 0
+rocksdb.metadata_block_cache_size 2048
+rocksdb.share_metadata_and_subkey_block_cache yes
+rocksdb.subkey_block_cache_size 2048
+unixsocketperm 777
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -135,7 +135,7 @@ log-retention-days -1
 # When running in daemonize mode, kvrocks writes a PID file in ${CONFIG_DIR}/kvrocks.pid by
 # default. You can specify a custom pid file location here.
 # pidfile /var/run/kvrocks.pid
-pidfile /tmp/kvrocks/kvrocks.pid
+pidfile ""
 
 # You can configure a slave instance to accept writes or not. Writing against
 # a slave instance may be useful to store some ephemeral data (because data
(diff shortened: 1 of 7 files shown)


torwig:
@2rueSid You can format your code with the `./x.py` command. It allows your code to pass the lint stage in the CI pipeline.

kvrocks.conf
```diff
@@ -858,4 +858,4 @@ rocksdb.write_options.memtable_insert_hint_per_batch no
 rocksdb.rate_limiter_auto_tuned yes
 
 ################################ NAMESPACE #####################################
-# namespace.test change.me
+# namespace.test change.me
```
git-hulk: Can revert this change

src/common/status.h
```diff
@@ -168,7 +168,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> :
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
-      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
+      : BaseType((typename StringInStatusOr<U>::BaseType&&)(std::move(v))) {}
```
git-hulk: Should revert this file. It seems the different behaviors between different versions of clang-format. For clang-format >= 14, it will add a space for type and parameter name, but not before clang-format 14.

git-hulk:
Rest are good to me, great work! @2rueSid

src/common/status.h
mapleFU: We can add a clang-format rule for this? Seems check by clang-tidy version is not a good way.

mapleFU (commented):
Overall LGTM except some style problems!
Thanks!

src/types/json.h
```diff
@@ -172,6 +172,34 @@ struct JsonValue {
...
+                                         bool is_object = val.is_object() && !val.empty();
+                                         bool is_number = val.is_number() && val.as<double>() != 0;
+
+                                         if (!is_number && !is_array && !is_object) {
+                                           return;
+                                         }
+
+                                         if (is_array) val = jsoncons::json::array();
+                                         if (is_object) val = jsoncons::json::object();
+                                         if (is_number) val = 0;
```
```suggestion
                                         if (is_array) val = jsoncons::json::array();
                                         else if (is_object) val = jsoncons::json::object();
                                         else if (is_number) val = 0;
                                         else return;
```

src/common/status.h
2rueSid: Ok, I will check the versions and will revert it.

2rueSid: I reverted the changes. It was because I use clang-format version 17.*.

Should I investigate this a bit more and try to add a rule?


Also, I'm trying to run linter with the command (actually each time I try to run linter it fails):
```bash
$ ./x.py check tidy --clang-tidy-path /.../llvm@14/bin/clang-tidy --run-clang-tidy-path /.../llvm@14/bin/run-clang-tidy
```

and receive an output:

```
4975 warnings and 1 error generated.
Error while processing kvrocks/src/types/bloom_filter.cc.
Suppressed 4974 warnings (4966 in non-user code, 8 NOLINT).
9 warnings treated as errors
Traceback (most recent call last):
  File "./x.py", line 422, in <module>
    args.func(**arg_dict)
  File "./x.py", line 195, in clang_tidy
    run(run_command, *options, *regexes, verbose=True, cwd=basedir)
  File "./x.py", line 68, in run
    raise RuntimeError(err)
RuntimeError:
failed to run: ('/opt/homebrew/opt/llvm@14/bin/run-clang-tidy', '-p', 'build', '-clang-tidy-binary', '/opt/homebrew/opt/llvm@14/bin/clang-tidy', '-j8', '-header-filter=kvrocks/src/|utils/kvrocks2redis/|tests/cppunit/', 'kvrocks/src/', 'utils/kvrocks2redis/', 'tests/cppunit/')
exit with code: 1
```

And the problem is, I guess that it's not a normal behavior, and I can't see an actual error, or if it actually exists. Are there some suggestions which I can try to resolve this?

mapleFU: Let's do it in a separate patch, since it's not a good idea making this patch too complex

2rueSid: Sure, you can assign me to an issue and I will do it ASAP.

torwig (approved):
Generally, LGTM.

mapleFU (approved):
Will wait for twice's review

git-hulk (approved):

PragmaTwice (approved):

PragmaTwice:
Thanks for your contribution!

PR #1852, opened 2023-10-24
Add support for the JSON.MERGE command 

### LINKED ISSUE #1815: Add support for the JSON.MERGE command

2rueSid:
support JSON.MERGE, this closes https://github.com/apache/kvrocks/issues/1815

[JSON.MERGE](https://redis.io/commands/json.merge/)

It's a bit messy, so I added comments. (After review I may remove them)

handling of the null value is a bit tricky so I utilize these commands to do this:
- [make expression](https://github.com/danielaparker/jsoncons/blob/master/doc/ref/jsonpath/make_expression.md)
- [flatten](https://github.com/danielaparker/jsoncons/blob/master/doc/ref/jsonpath/flatten.md)

Also here is info about [mergepatch](https://github.com/danielaparker/jsoncons/blob/master/doc/ref/mergepatch/mergepatch.md)

Mostly for references I used the original command: [JSON.MERGE](https://github.com/RedisJSON/RedisJSON/blob/90ba30273ce0dcd3e7a91704f7f4900b1e5a8c4a/redis_json/src/commands.rs#L237)

please review @mapleFU @PragmaTwice @torwig


src/types/json.h
```diff
@@ -224,4 +301,4 @@ struct JsonValue {
   ~JsonValue() = default;
 
   jsoncons::json value;
-};
+};
```
```suggestion
};

```

src/types/json.h
```diff
@@ -215,6 +218,80 @@ struct JsonValue {
...
   }
 
+  StatusOr<bool> Merge(const std::string_view path, const std::string &merge_value) {
+    bool is_updated = false;
+    const std::string json_root_path = "$";
+    try {
+      jsoncons::json patch_value = jsoncons::json::parse(merge_value);
+      bool not_exists = jsoncons::jsonpath::json_query(value, path).empty();
+
+      if (path == json_root_path || not_exists) {
```
git-hulk: I guess the meaning of `non-existing keys the path must be $` is when the key is NOT found instead of path. That said:

```
// Bad
$ JSON.MERGE no_exists_key $.a '{}'
// Good, will create key `no_exists_key`
$ JSON.MERGE no_exists_key $ '{}'
// Good if the object path is missing
$ JSON.MERGE exists_key $.a '1'

```

src/types/json.h
```diff
@@ -215,6 +218,80 @@ struct JsonValue {
...
+    }
+
+    jsoncons::json result = value;
+    for (auto it = parts.rbegin(); it != parts.rend(); ++it) {
+      jsoncons::json obj;
+      obj[*it] = result;
+      result = obj;
+    }
+
+    return result;
```
PragmaTwice: Seems it is not general enough, e.g. it cannot process path like `$[0]`, `$["hello"]` or `$..a`. 
I think here we can just use `Get`.

2rueSid: Sorry, I was busy the last couple of days. Now, I can finally finish it.

What exactly do you mean by `I think here we can just use Get.`?
I was thinking about extracting keys from a string and using jsonpointer to create an object. However, it still won't handle all possible grammar.

PragmaTwice: I mean maybe we can just use `jsonpath::json_query` here (as well as `JsonValue::Get`).

2rueSid: A brief update: I have researched a bit, and it appears that we might require the path parser.

The function `jsonpath::json_query` can return a normalized path instead of a value, but only if a value exists at that path. The same applies to `jsonpath::json_replace`—it operates only when a value exists.

Moreover, the parser could be beneficial for the Set function, as it currently cannot process the following expression:
```bash
127.0.0.1:6666> JSON.SET doc $ '{"a":2}'
OK
127.0.0.1:6666> JSON.SET doc $.b '8'
OK
127.0.0.1:6666> JSON.GET doc $
"[{\"a\":2}]"
```

Ideally, in the `Merge()` function, we should invoke the `Set()` function for this case, I presume.

I will also reach out to the maintainer of `jsoncons` to check whether there is a more elegant implementation of this.

PragmaTwice: Oh, I check the document of JSON.MERGE and now get your point.

I think currently you can just implement the one that is supported by jsoncons (specifically, the case when a value exists in the jsonpath provided by the user, which can be handled by `json_query`) and return some error message in other cases. (you can add a `TODO` comment in the code.)

And after this PR being merged, you can try to cover more cases if you are interested. 

And of course it  is a great idea to open a discussion or issue in the jsoncons community to address this problem : )

2rueSid: I've updated the method slightly and added a TODO comment. I've also initiated a discussion https://github.com/danielaparker/jsoncons/discussions/468

I will attempt to implement this in a separate pull request, as it could be quite complex.

src/types/redis_json.cc
```diff
@@ -91,6 +91,14 @@ rocksdb::Status Json::Info(const std::string &user_key, JsonStorageFormat *stora
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status Json::create(const std::string &ns_key, JsonMetadata metadata, const std::string &value) {
```
PragmaTwice:
```suggestion
rocksdb::Status Json::create(const std::string &ns_key, JsonMetadata &metadata, const std::string &value) {
```

I prefer lvalue reference or pointer since it will be changed in `write`.

2rueSid: yes, it's a good point.

PragmaTwice:
Seems you also need to change the method declaration to make it pass the compiler.

src/types/json.h
```diff
@@ -253,6 +255,54 @@ struct JsonValue {
...
+
+      if (path == json_root_path) {
+        // Merge with the root. Patch function complies with RFC7396 Json Merge Patch
+        jsoncons::mergepatch::apply_merge_patch(value, patch_value);
+        is_updated = true;
+      } else if (!patch_value.is_null()) {
+        // Replace value by path
+        jsoncons::jsonpath::json_replace(
+            value, path, [&patch_value, &is_updated](const std::string & /*path*/, jsoncons::json &target) {
+              target = patch_value;
```
```suggestion
              jsoncons::mergepatch::apply_merge_patch(target, patch_value);
```

src/types/json.h
```diff
@@ -253,6 +255,54 @@ struct JsonValue {
...
+        expr.evaluate(
+            value, [&](const jsoncons::string_view &p, const jsoncons::json &val) { converted_path = p; },
+            jsoncons::jsonpath::result_options::path);
+        // Unify object state
+        jsoncons::json flattened = jsoncons::jsonpath::flatten(value);
+        if (flattened.contains(converted_path)) {
+          flattened.erase(converted_path);
+          value = jsoncons::jsonpath::unflatten(flattened);
+          is_updated = true;
+        }
```
PragmaTwice, line 358: Seems we can just use `json_replace` here?

2rueSid: We need to delete the key if the merge value is `null`. Since `json_replace` does not provide access to the parent keys, we cannot use it in this case.

Consider the following example:
```bash
redis> JSON.SET doc $ '{"a":2}'
OK
redis> JSON.MERGE doc $.a 'null'
OK
redis> JSON.GET doc $
```
In the `json_replace` function with this path, we would retrieve the value of `a`. However, to call the `erase()` function, we need to pass a key name, and the key `a` cannot point to itself. Therefore, to erase it, we need to have access to the parent key.

We need something like this:

```cpp
json_replace("{\"a\": 2}", path="$", pseudo_cb={
  // This callback function attempts to erase the key 'a' from the JSON object.
  value.erase("$.a");
})
```

2rueSid:
Is a SonarCloud analysis required for a PR to pass? It seems to display a duplication error for C++ test cases, although there are no duplicates.

@PragmaTwice

PragmaTwice:
> Is a SonarCloud analysis required for a PR to pass? It seems to display a duplication error for C++ test cases, although there are no duplicates.
> 
> @PragmaTwice

Nope. It is fine to leave it alone.

@git-hulk Do you have free time to help review and merge this PR?

git-hulk:
> > Is a SonarCloud analysis required for a PR to pass? It seems to display a duplication error for C++ test cases, although there are no duplicates.
> > @PragmaTwice
> 
> Nope. It is fine to leave it alone.
> 
> @git-hulk Do you have free time to help review and merge this PR?

Sure, I would take a look soon.

git-hulk (approved):

git-hulk:
Thank you! @2rueSid

PragmaTwice:
Hi @2rueSid , in the new version of jsoncons, some new features have been added for removing nodes.

Refer to https://github.com/danielaparker/jsoncons/issues/467#issuecomment-1826868711.

Are you willing to improve this MERGE command using that?

2rueSid:
> Hi @2rueSid , in the new version of jsoncons, some new features have been added for removing nodes.
> 
> Refer to [danielaparker/jsoncons#467 (comment)](https://github.com/danielaparker/jsoncons/issues/467#issuecomment-1826868711).
> 
> Are you willing to improve this MERGE command using that?

Hello. Yes for sure! I will take a look tomorrow.

PR #1853, opened 2023-10-24
Fix status mis-return in JSON.CLEAR and polish code style

PragmaTwice:
https://github.com/apache/kvrocks/blob/35e041f628847d67a75e03573ac29265ba046ede/src/types/redis_json.cc#L177-L178

It returns a wrong status (which will lead to an always-OK status returned).

And for `JSON.TYPE`, it does not return the status which is required to handle.

Besides, here are also some style changes related to `size_t` and `StatusOr`.

cc @2rueSid @git-hulk


torwig (dismissed):
@PragmaTwice Good catch!

src/types/redis_json.cc
```diff
@@ -159,11 +159,14 @@ rocksdb::Status Json::Type(const std::string &user_key, const std::string &path,
   auto s = read(ns_key, &metadata, &json_val);
   if (!s.ok()) return s;
 
-  auto res = json_val.Type(path, results);
+  auto res = json_val.Type(path);
+  if (!res) return rocksdb::Status::InvalidArgument(res.Msg());
```
mapleFU, line 163: Would it be an invalid value here? Rather than just InvalidArgument

mapleFU (dismissed):

src/types/redis_json.cc
PragmaTwice: It will return error status iff the `path` is invalid.

torwig (dismissed):

src/types/redis_json.cc
mapleFU: Oh I see, when type is unknown it will just return a unknown string

mapleFU (approved):

torwig (approved):

git-hulk (approved):

PR #1855, opened 2023-10-24
Allow to change the worker thread number in-flight

### LINKED ISSUE #1802: Add the support of changing worker threads in-flight

git-hulk:
This closes #1802

Scaling up is as simple as starting new threads and adding them to the worker pool.
But scaling down is a bit complex since we need to migrate existing connections from
the to-be-removed workers to survived workers. This process is NOT so easy for the sake
of the blocking commands that will store its context(includes fd and owner pointer) in
different global variables and we need to find out and remove them. As far as I know,
it contains the below commands:

- All LIST blocking command
- Subscribe/PSubscribe
- XREAD

To make it simple, we would like to close those connections if the last command is
a blocking command. And we can improve this if any users complain about this.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -364,7 +364,8 @@ void Config::initFieldCallback() {
           {"workers",
            [](Server *srv, const std::string &k, const std::string &v) -> Status {
              if (!srv) return Status::OK();
-             return srv->AdjustWorkerThreads();
+             srv->AdjustWorkerThreads();
+             return Status::OK();
            }},
           {"dir",
            [this](Server *srv, const std::string &k, const std::string &v) -> Status {
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -105,7 +105,7 @@ Config::Config() {
       {"tls-session-cache-timeout", false, new IntField(&tls_session_cache_timeout, 300, 0, INT_MAX)},
       {"tls-replication", true, new YesNoField(&tls_replication, false)},
 #endif
-      {"workers", true, new IntField(&workers, 8, 1, 256)},
+      {"workers", false, new IntField(&workers, 8, 1, 256)},
       {"timeout", false, new IntField(&timeout, 0, 0, INT_MAX)},
       {"tcp-backlog", true, new IntField(&backlog, 511, 0, INT_MAX)},
       {"maxclients", false, new IntField(&maxclients, 10240, 0, INT_MAX)},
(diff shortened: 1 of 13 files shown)


guojidan:
nice, long connect will not disconnect when change workers, I use redis-benchmark continuous write

git-hulk:
> nice, long connect will not disconnect when change workers, I use redis-benchmark continuous write

Thank you! I'm working on test cases.

mapleFU:
Should we prevent from setting the thread-cnt to a dangerous value( .e.g `0`)?

git-hulk:
> Should we prevent from setting the thread-cnt to a dangerous value( .e.g `0`)?

@mapleFU Yes, IntField will check the value range before setting.

`{"workers", true, new IntField(&workers, 8, 1, 256)},`

src/server/redis_connection.cc
```diff
@@ -74,6 +75,17 @@ void Connection::Close() {
...
 
+bool Connection::IsBlockingMode() const {
+  if (current_cmd == nullptr) return false;
+  if (dynamic_cast<BlockingCommander *>(current_cmd.get()) != nullptr) return true;
+
+  auto args = current_cmd->GetArgs();
+  if (args.empty()) return false;
+
+  std::string cmd_name = util::ToLower(args.front());
+  return cmd_name == "xread";
```
PragmaTwice: Seems it can cause data race since `args` are used in multiple threads. And the check `cmd_name == "xread"` is also a disaster of maintainability (we can easily forget while facing new blocking commands).

So I think a better solution is add an new attribute like `blocking` to these commands and check it here. Then we actually do not need `args` and even `dynamic_cast`

git-hulk: Gotcha, thank you!

src/config/config.cc
```diff
@@ -359,296 +359,304 @@ void Config::initFieldCallback() {
...
-           if (cmd == "*") {
-             profiling_sample_all_commands = true;
+  std::map<std::string, CallbackFn> callbacks =
+      {
+          {"workers",
+           [](Server *srv, const std::string &k, const std::string &v) -> Status {
+             if (!srv) return Status::OK();
+             srv->AdjustWorkerThreads();
+             return Status::OK();
+           }},
```
git-hulk, line 369: Only those lines are newly added, rest is formatted by clang-format

src/server/server.cc
```diff
@@ -417,8 +419,6 @@ void Server::SubscribeChannel(const std::string &channel, redis::Connection *con
   std::lock_guard<std::mutex> guard(pubsub_channels_mu_);
 
   auto conn_ctx = ConnContext(conn->Owner(), conn->GetFD());
-  conn_ctxs_[conn_ctx] = true;
```
git-hulk, line 420: No where is using conn_ctxs_, so delete it.

tests/gocase/unit/config/config_test.go
```diff
@@ -141,3 +144,92 @@ func TestStartWithoutConfigurationFile(t *testing.T) {
...
+			}
+			wg.Wait()
+		}
+		// Reduce worker threads to 4
+		runCommands(4)
+
+		// Reduce worker threads to 1
+		runCommands(1)
+
+		// Reduce worker threads to 12
```
torwig: Just nitpicking: in this case, we **increasing** worker threads to 12 :)

src/server/worker.cc
```diff
@@ -344,6 +344,31 @@ redis::Connection *Worker::removeConnection(int fd) {
...
+//
+// To make it simple, we would close the connection if it's
+// blocked on a key or stream.
+void Worker::MigrateConnection(Worker *target, redis::Connection *conn) {
+  if (!target || !conn) return;
+  if (conn->current_cmd != nullptr && conn->current_cmd->IsBlocking()) {
+    // don't need to close the connection since destroy worker thread will close it
+    return;
+  }
+
```
torwig, line 358: Could any blocking command be received after the previous check and before the actual connection migrating?

torwig (commented):
A few comments, but the rest LGTM. @git-hulk Good job!

tests/gocase/unit/config/config_test.go
git-hulk: Good catch, thank you!

src/server/worker.cc
git-hulk: It's not since Config::Set is an exclusive command, so it won't receive any other commands before this process was finished.

torwig (dismissed):
LGTM.

src/server/server.cc
```diff
@@ -245,6 +244,9 @@ void Server::Join() {
   if (auto s = task_runner_.Join(); !s) {
     LOG(WARNING) << s.Msg();
   }
+  for (const auto &worker : worker_threads_) {
```
mapleFU, line 246: Just curious, why worker would shutdown after the other threads shutdown?

src/server/server.cc
```diff
@@ -1680,6 +1677,64 @@ void Server::AdjustOpenFilesLimit() {
   }
 }
 
+void Server::AdjustWorkerThreads() {
```
mapleFU, line 1679: Do we need some kind of locking ( Or CAS ) here to avoid concurrent `AdjustWorkerThreads`?

src/server/server.cc
```diff
@@ -1680,6 +1677,64 @@ void Server::AdjustOpenFilesLimit() {
...
+    auto worker = std::make_unique<Worker>(this, config_);
+    auto worker_thread = std::make_unique<WorkerThread>(std::move(worker));
+    worker_thread->Start();
+    worker_threads_.emplace_back(std::move(worker_thread));
+  }
+}
+
+void Server::decreaseWorkerThreads(size_t delta) {
+  auto current_worker_threads = worker_threads_.size();
+  auto remain_worker_threads = current_worker_threads - delta;
```
mapleFU, line 1710: Should we better adding a `DCHECK(current_worker_threads > delta);`

src/server/server.cc
```diff
@@ -1680,6 +1677,64 @@ void Server::AdjustOpenFilesLimit() {
...
+      worker_thread->GetWorker()->MigrateConnection(target_worker, iter.second);
+    }
+    worker_thread->Stop();
+    // Don't join the worker thread here, because it may join itself.
+    recycle_worker_threads_.push(std::move(worker_thread));
+  }
+}
+
+void Server::cleanupExitedWorkerThreads() {
+  std::unique_ptr<WorkerThread> worker_thread = nullptr;
```
mapleFU, line 1731: Should we dcheck thread-id not eq?

src/server/server.cc
git-hulk: It doesn't wait for other threads but can't be the first one, because worker_threads_ may be modified after starting, so I move it to the last.

src/server/server.cc
git-hulk: It's unnecessary since Config::Set is an exclusive command, so it won't have the data race condition.

src/server/server.cc
git-hulk, line 1710: @mapleFU Do you mean this?
```suggestion
  DCHECK(current_worker_threads > delta);
  auto remain_worker_threads = current_worker_threads - delta;
```

src/server/server.cc
git-hulk: I guess it's fine since those threads are from decreaseWorkerThreads

src/server/server.cc
mapleFU: Yeah

src/server/server.cc
mapleFU: (That's why we only need DCHECK?)

git-hulk: Yes

mapleFU: Also, should `Server::Join` also try to check `recycle_worker_threads_`?

mapleFU:
I'm a bit confused about the ctor ordering here:

1. `Server::~Server` calls `cleanupExitedWorkerThreads()`
2. cron calls `cleanupExitedWorkerThreads()`

However, when `Join`, would the exited worker thread be cleaned?

src/server/server.cc
git-hulk: It has been done before destroying the server.

git-hulk:
> I'm a bit confused about the ctor ordering here:
> 
> 1. `Server::~Server` calls `cleanupExitedWorkerThreads()`
> 2. cron calls `cleanupExitedWorkerThreads()`
> 
> However, when `Join`, would the exited worker thread be cleaned?

The `cron` thread periodic call `cleanupExitedWorkerThreads()` but it can't promise it will call `cleanupExitedWorkerThreads()` before exiting the cron thread. So for the server destructor, it MUST check again to prevent exceptions.

src/server/server.cc
```diff
@@ -117,6 +117,9 @@ Server::~Server() {
   }
 
   lua::DestroyState(lua_);
```
mapleFU: Since `conn_ctxs_` is removed, should we remove or change the comment in L114?

mapleFU: Also. should we adjust the ordering of `lua::DestroyState` and cleanup?( not so familiar with lua here, so don't know would we meet the problem)

mapleFU:
Got it, so finally resource is handled by dtor.

src/server/server.cc
git-hulk: It should be fine to leave `lua::DestroyState` as it is because workers are stopped.

git-hulk: > Since conn_ctxs_ is removed, should we remove or change the comment in L114?

Yes, deleted.

mapleFU (approved):
LGTM. Thanks

torwig (approved):

git-hulk:
Thanks all, merging..

PR #1861, opened 2023-10-30
Unify server's shorten form, change svr to srv

git-hulk:
Currently, we have two shorten forms for the server variable name,
and it will be better to keep a consistent name.


mapleFU (approved):
Have a glance and seems that all modified variables is Server

git-hulk:
> Have a glance and seems that all modified variables is Server

Yes, only changed the variable name of server from `svr` to `srv`.

mapleFU:
(I mean nothing weird is changed, aha )

tisonkun (approved):
Thank you!

PR #1863, opened 2023-10-31
Graceful shutdown the workers when reducing worker threads

git-hulk:
#1855 introduced the data race even only migrating the non-blocking connections. For example:

T1: C0(connection) is running the command Config::Set on W0(worker) and C1 is running another command on W1. C0 got the exclusive lock for executing commands and C1 will wait for the lock inside ExecuteCommands.

T2: C0 migrates C1 to W0 after reducing the number of worker threads, but the ExecuteCommands function will continue executing after migrating. So both W0 and W1 will access the C1 at the same time.

To avoid this, I simply don't migrate the connection if it has any running command and delay to shut down the worker, so that old connections have a chance to complete the running command.

### DIFF
--- a/src/commands/blocking_commander.h
+++ b/src/commands/blocking_commander.h
@@ -44,7 +44,6 @@ class BlockingCommander : public Commander,
   // in other words, returning true indicates ending the blocking
   virtual bool OnBlockingWrite() = 0;
 
-  bool IsBlocking() const override { return true; }
   // to start the blocking process
   // usually put to the end of the Execute method
   Status StartBlocking(int64_t timeout, std::string *output) {
--- a/src/server/redis_connection.h
+++ b/src/server/redis_connection.h
@@ -111,7 +111,7 @@ class Connection : public EvbufCallbackBase<Connection> {
   Worker *Owner() { return owner_; }
   void SetOwner(Worker *new_owner) { owner_ = new_owner; };
   int GetFD() { return bufferevent_getfd(bev_); }
-  bool HasRunningCommand() const { return current_cmd != nullptr; }
+  bool HasRunningCommand() const { return current_cmd_ != nullptr; }
   evbuffer *Input() { return bufferevent_get_input(bev_); }
   evbuffer *Output() { return bufferevent_get_output(bev_); }
   bufferevent *GetBufferEvent() { return bev_; }
@@ -129,7 +129,6 @@ class Connection : public EvbufCallbackBase<Connection> {
   std::deque<redis::CommandTokens> *GetMultiExecCommands() { return &multi_cmds_; }
 
   std::function<void(int)> close_cb = nullptr;
-  std::unique_ptr<Commander> current_cmd;
 
   std::set<std::string> watched_keys;
   std::atomic<bool> watched_keys_modified = false;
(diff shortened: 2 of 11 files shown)


src/server/worker.cc
```diff
@@ -252,7 +252,8 @@ Status Worker::listenTCP(const std::string &host, uint32_t port, int backlog) {
     }
 
     evutil_make_socket_nonblocking(fd);
-    auto lev = NewEvconnlistener<&Worker::newTCPConnection>(base_, LEV_OPT_CLOSE_ON_FREE, backlog, fd);
+    auto lev =
+        NewEvconnlistener<&Worker::newTCPConnection>(base_, LEV_OPT_THREADSAFE | LEV_OPT_CLOSE_ON_FREE, backlog, fd);
```
PragmaTwice, line 256: I wonder if `THREADSAFE` flag will affect performance in a explicit way?

src/commands/blocking_commander.h
```diff
@@ -44,7 +44,6 @@ class BlockingCommander : public Commander,
   // in other words, returning true indicates ending the blocking
   virtual bool OnBlockingWrite() = 0;
 
-  bool IsBlocking() const override { return true; }
```
PragmaTwice, line 47: Now we do not need to check if it is blocking or not?

src/server/worker.cc
git-hulk: Yes, I guess it will affect the performance since it needs an extra lock and unlock operation. But it should be fine since the bottlenecks of accept connections won't be here. Let's me have a performance compare later.

src/commands/blocking_commander.h
git-hulk: Yes, we don't migrate the connection once it has the running command to avoid data race between the new and old worker. The data race condition was mentioned in PR summary.

src/server/worker.cc
git-hulk: I tested on Ubuntu 22.04, it's no different. The performance of connecting is around 1.4W QPS.

git-hulk:
This PR opened before changing the sonar reconfigured, so just ignore it. cc @PragmaTwice

src/server/redis_connection.h
```diff
@@ -111,6 +111,7 @@ class Connection : public EvbufCallbackBase<Connection> {
   Worker *Owner() { return owner_; }
   void SetOwner(Worker *new_owner) { owner_ = new_owner; };
   int GetFD() { return bufferevent_getfd(bev_); }
+  bool HasRunningCommand() const { return saved_current_command_ != nullptr; }
```
PragmaTwice: I'm wondering if there are non-blocking commands which are running in some situation?

git-hulk: Yes, that's possible. 

For example,  connection0 is running on worker0 with the command `Get`, but it will wait for the lock if the migrating connection process is ongoing, then connection0 will be migrated to a new worker. Since then the new worker will start schedule the connection0 and the old worker will be still using the connection while resuming the ExecuteCommands.

PragmaTwice: Oh, I mean, only blocking commands will be stored in `saved_current_command_ `.

git-hulk: Yes, I also misunderstood what you mean. You're right, it should only have the data race in current_command, and the other variables should have the same issue.

git-hulk: Fixed

torwig (dismissed):
LGTM

torwig (approved):

Issue #1808, opened 2023-10-15
Add support for the JSON.ARRINSERT command


PragmaTwice:
Please refer to https://github.com/apache/kvrocks/issues/1805.

PR #1867, opened 2023-11-02
Support for the JSON.ARRINSERT command 

jihuayu:
[#1808](https://github.com/apache/kvrocks/issues/1808)


git-hulk:
Looks generally good to me, thank you!

jihuayu:
The logic for when `index > array size` is missing here. I need to first check the behavior of the Redis stack in this situation.

jihuayu:
😿 😿 😿 
https://github.com/RedisJSON/RedisJSON/issues/1129
RedisJSON's `json.arrinsert` has some bugs. We need to wait for the upstream to determine the correct logic.

mapleFU:
We don't need to follow it ( but I guess mention it in comment is better)

src/types/json.h
```diff
@@ -179,6 +179,33 @@ struct JsonValue {
...
+  StatusOr<std::vector<std::optional<uint64_t>>> ArrInsert(std::string_view path, const int64_t &index,
+                                                           const std::vector<jsoncons::json> &insert_values) {
+    std::vector<std::optional<uint64_t>> result_count;
+
+    try {
+      jsoncons::jsonpath::json_replace(
+          value, path, [&insert_values, &result_count, index](const std::string & /*path*/, jsoncons::json &val) {
+            if (val.is_array()) {
+              auto len = static_cast<int64_t>(val.size());
+              if (index >= len /*index > 0*/ || len + index < 0 /*index < 0*/) {
```
```suggestion
              if (index >= len /*index is out of upper bound*/ || len + index < 0 /*index < 0*/) {
```

git-hulk (dismissed):

git-hulk:
Only one typo in the comment inline, rest are good to me.

src/types/json.h
jihuayu: It is not typo, It is mean when index > 0.
I think I can change this line to ` if (index >= len /*when index > 0*/ || len + index < 0 /*when index < 0*/) {`

git-hulk: `/*when index > 0*/` it should be >= length instead > 0? since index > 0 is good if it doesn't exceed the upper bound of the array.

jihuayu: I means that when index > 0, we should check if index >= len. And when index < 0, we should check if index + len < 0.
It seems that this comment could be easily misunderstood. Should I delete it?

git-hulk: oh, I did misunderstand your point. 

> It seems that this comment could be easily misunderstood. Should I delete it?

Sounds good, or add a longer comment line outside to make it clearer.

jihuayu: Ok, I have added a longer comment above.

git-hulk (approved):

PR #1872, opened 2023-11-03
Support for the JSON.OBJKEYS command

### LINKED ISSUE #1820: Add support for the JSON.OBJKEYS command

jihuayu:
close https://github.com/apache/kvrocks/issues/1820

Redis doc  [json.objkeys](https://redis.io/commands/json.objkeys/)

### DIFF
--- a/src/commands/cmd_json.cc
+++ b/src/commands/cmd_json.cc
@@ -148,6 +148,29 @@ class CommandJsonType : public Commander {
   }
 };
 
+class CommandJsonObjkeys : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    redis::Json json(srv->storage, conn->GetNamespace());
+
+    std::vector<std::optional<std::vector<std::string>>> results;
+
+    auto s = json.ObjKeys(args_[1], args_[2], results);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = redis::MultiLen(results.size());
+    for (const auto &item : results) {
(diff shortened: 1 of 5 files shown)


git-hulk:
@jihuayu I think only Go tests is ok if most of them can be covered.

jihuayu:
@git-hulk  I'm not quite understanding the difference in using between cpp test and go test in kvrocks. When should I add a cpp test, and when go test is OK?

git-hulk:
> rstanding the difference in using between cpp test and go test in kvrocks. When should I add a cpp test, and when go test is

No, only Go is ok if the key path can be tested. And cpp test is used to cover the path which is hard to test through the redis protocol way.

src/commands/cmd_json.cc
```diff
@@ -167,6 +167,29 @@ class CommandJsonType : public Commander {
...
 
+class CommandJsonObjkeys : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    redis::Json json(srv->storage, conn->GetNamespace());
+
+    std::vector<std::optional<std::vector<std::string>>> results;
+
+    auto s = json.ObjKeys(args_[1], args_[2], results);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
```
git-hulk: Should return null if key is non-exists, rest are good to me.

jihuayu: fixed

git-hulk (dismissed):

git-hulk (approved):

Issue #1825, opened 2023-10-15
Add support for the JSON.TOGGLE command


PragmaTwice:
Please refer to #1805.

chakkk309:
Hi, I want to try this task~

PR #1875, opened 2023-11-03
 support for the JSON.TOGGLE command

MaheshMadushan:
[Add support for the JSON.TOGGLE command #1825]

### DIFF
--- a/src/commands/cmd_json.cc
+++ b/src/commands/cmd_json.cc
@@ -171,6 +171,24 @@ class CommandJsonClear : public Commander {
   }
 };
 
+class CommandJsonToggle : public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::Json json(svr->storage, conn->GetNamespace());
+
+    std::string path = (args_.size() > 2) ? args_[2] : "$";
+    auto s = json.Toggle(args_[1], path);
+
+    if (s.IsNotFound()) {
+      return Status::OK();
+    }
+
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    return Status::OK();
+  }
+};
+
 class CommandJsonArrLen : public Commander {
  public:
   Status Execute(Server *svr, Connection *conn, std::string *output) override {
@@ -209,6 +227,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandJsonSet>("json.set", 4, "write", 1, 1
                         MakeCmdAttr<CommandJsonType>("json.type", -2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandJsonArrAppend>("json.arrappend", -4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandJsonClear>("json.clear", -2, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandJsonClear>("json.toggle", -2, "write", 1, 1, 1),
                         MakeCmdAttr<CommandJsonArrLen>("json.arrlen", -2, "read-only", 1, 1, 1), );
 
 }  // namespace redis
--- a/src/commands/cmd_json.cc
+++ b/src/commands/cmd_json.cc
@@ -195,10 +195,20 @@ class CommandJsonToggle : public Commander {
     redis::Json json(svr->storage, conn->GetNamespace());
 
     std::string path = (args_.size() > 2) ? args_[2] : "$";
-    auto s = json.Toggle(args_[1], path);
+    std::vector<std::optional<int>> results;
+    auto s = json.Toggle(args_[1], path, results);
 
     if (s.IsNotFound()) {
+      *output = redis::NilString();
       return Status::OK();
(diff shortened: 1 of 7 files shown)


src/types/json.h
```diff
@@ -173,6 +173,22 @@ struct JsonValue {
     return types;
   }
 
+  Status Toggle(std::string_view path) {
+    try {
+      jsoncons::jsonpath::json_replace(value, path, [&](const std::string & /*path*/, jsoncons::json &val) {
+        bool is_boolean = val.is_bool() && !val.empty();
```
PragmaTwice: May I ask why we check `empty()` here? Seems `is_bool()` is enough to me.

MaheshMadushan: Thanks for pointing out. fixed

src/types/json.h
```diff
@@ -173,6 +173,22 @@ struct JsonValue {
...
 
+  Status Toggle(std::string_view path) {
+    try {
+      jsoncons::jsonpath::json_replace(value, path, [&](const std::string & /*path*/, jsoncons::json &val) {
+        bool is_boolean = val.is_bool();
+
+        if (is_boolean)
+          val = !val.as_bool();
+        else
+          return;
```
```suggestion
        if (val.is_bool()) val = !val.as_bool();
```

MaheshMadushan: Thanks for the suggestion. fixed it.

src/commands/cmd_json.cc
```diff
@@ -189,6 +189,24 @@ class CommandJsonClear : public Commander {
...
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::Json json(svr->storage, conn->GetNamespace());
+
+    std::string path = (args_.size() > 2) ? args_[2] : "$";
+    auto s = json.Toggle(args_[1], path);
+
+    if (s.IsNotFound()) {
+      return Status::OK();
+    }
```
PragmaTwice: Have you tested this case? I think here the connection will be blocked.

MaheshMadushan: Not yet. Writing more test cases for happy path. Will update asap with error scenarios.

PragmaTwice: You can add `*output = redis::NilString()` before returning to solve it, just like in other commands.

MaheshMadushan: Sure will do

PragmaTwice (dismissed):

PragmaTwice:
You can try to build it to make sure it will pass the compiler before pushing.

MaheshMadushan:
> You can try to build it to make sure it will pass the compiler before pushing.

issue fixed.

src/types/json.h
```diff
@@ -211,6 +211,17 @@ struct JsonValue {
     return types;
   }
 
+  Status Toggle(std::string_view path) {
+    try {
+      jsoncons::jsonpath::json_replace(value, path, [&](const std::string & /*path*/, jsoncons::json &val) {
+        if (val.is_bool()) val = !val.as_bool();
```
git-hulk: > JSON.TOGGLE returns an array of integer replies for each path, the new value (0 if false or 1 if true), or nil for JSON values matching the path that are not Boolean. 

The return value should be an array of integers. That said if there are two paths matched and one of them is NOT a boolean value, then it should return an array with:

1) 0(or 1)
2) nil

Can refer to other command implementations.

MaheshMadushan: Thanks for poiniting out. will fix this

git-hulk: Thank you!

src/types/json.h
```diff
@@ -211,6 +211,23 @@ struct JsonValue {
     return types;
   }
 
+  StatusOr<std::vector<std::optional<int>>> Toggle(std::string_view path) {
+    std::vector<std::optional<int>> result;
```
PragmaTwice: Seems here we can just use a `std::vector<std::optional<bool>>`?

MaheshMadushan: Thanks for pointing out. That is more suitable. Fixed.

src/commands/cmd_json.cc
```diff
@@ -219,6 +219,35 @@ class CommandJsonClear : public Commander {
...
+
+    if (s.IsNotFound()) {
+      *output = redis::NilString();
+      return Status::OK();
+    }
+
+    *output = redis::MultiLen(results.size());
+    for (auto it = results.rbegin(); it != results.rend(); ++it) {
+      if ((*it).has_value()) {
+        *output += redis::Integer((*it).value());
```
PragmaTwice:
```suggestion
      if (it->has_value()) {
        *output += redis::Integer(it->value());
```

MaheshMadushan: Thanks for poiniting out. fixed

git-hulk (approved):
LGTM, thanks

MaheshMadushan:
> LGTM, thanks

Thanks for the feedback

git-hulk:
@MaheshMadushan Thanks for your contribution.

PR #1877, opened 2023-11-05
Add more sanity checks for docker image

### LINKED ISSUE #1793: Add the sanity check for docker image
We'd better add some sanity checks in:

https://github.com/apache/kvrocks/blob/7f0c5156493dc3601bbc6f2157f9fa1d8462ea51/.github/workflows/kvrocks.yaml#L292-L309

_Originally posted by @tisonkun in https://github.com/apache/kvrocks/issues/1784#issuecomment-1735890155_

As mentioned in the comments, we at least check if the docker listening port can be accessed from outside.

jyf111:
This PR should fix #1793.

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -310,7 +310,18 @@ jobs:
           tags: kvrocks:ci
           outputs: type=docker
       - name: Test built image
-        run: docker run --rm kvrocks:ci -v
+        run: |
+          ID="$(docker run --rm -d -p 6666:6666 kvrocks:ci)"
+          sleep 1m
+          if [ "$(docker inspect --format='{{.State.Health.Status}}' $ID)" != "healthy" ]; then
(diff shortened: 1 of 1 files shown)


git-hulk (dismissed):

.github/workflows/kvrocks.yaml
```diff
@@ -310,7 +310,18 @@ jobs:
           tags: kvrocks:ci
           outputs: type=docker
       - name: Test built image
-        run: docker run --rm kvrocks:ci -v
```
PragmaTwice, line 313: I think we still need this one since it is not covered in the new checks.

jyf111: When starting kvrocks, the version is output but cannot be seen due to detach. 
I have restored this version check to output the version to github action log.

PragmaTwice (approved):

git-hulk (approved):

Issue #1864, opened 2023-11-01
Compact on all cf with same range?


raffertyyu:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

1142a9d87ac6c404dcf4e0d06778af9ae738539b

### Minimal reproduce step

none

### What did you expect to see?

call compact on specific cf

### What did you see instead?

call compact on all cf

### Anything Else?

In method `CompactionChecker::PickCompactionFiles`, kvrocks will search all sst files in cf specified by `cf_name`. If a sst file should be compacted, kvrocks will call 
```
storage_->Compact(&start_key, &stop_key)
```
But, `Compact` method will CompactRange on all cfs.

Expected:
`Compact` method needs a `cf_name` parameter to do compact on right cf not all cfs.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PragmaTwice:
Good catch! Feel free to submit a PR if you are willing to fix it.

raffertyyu:
OK, I will fix it.

PR #1878, opened 2023-11-05
Add column family parameter to Compact methon

raffertyyu:
fix issue #1864 
Added a parampeter to specify the column family that should be compacted.
If `cf` is `nullptr`, it means all `CF`s should be compacted.

This patch also fixed a calculation bug I found in `compaction_checker.cc`
```c++
auto force_compact_min_ratio =
    static_cast<double>(storage_->GetConfig()->force_compact_file_min_deleted_percentage / 100);
```
The `force_compact_file_min_deleted_percentage` is an interger and must be between 1 and 100. So the `force_compact_min_ratio` result is always either 0.0 or 1.0.


git-hulk (approved):
Good catch, LGTM.

PR #1882, opened 2023-11-06
Don't rewrite unchanged items into the configuration file

git-hulk:
Currently, config rewrite will also add some default configurations like backup-dir, pidfile into the configuration file. Some of them are for compatible purposes and don't expect to be seen by users.

So we resolve this by ignoring the configuration items if they do not exist in the configuration file and the value was not modified.

Thanks for @PragmaTwice finding this issue and proposing the solution.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -853,7 +853,7 @@ Status Config::Set(Server *srv, std::string key, const std::string &value) {
   return Status::OK();
 }
 
-bool Config::checkOptionIsDefault(const std::string &key, const std::string &value) const {
+bool Config::checkFieldValueIsDefault(const std::string &key, const std::string &value) const {
   auto iter = fields_.find(key);
   return iter != fields_.end() && iter->second->Default() == value;
 }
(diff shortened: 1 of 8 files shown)


src/config/config.cc
```diff
@@ -859,6 +853,11 @@ Status Config::Set(Server *srv, std::string key, const std::string &value) {
   return Status::OK();
 }
 
+bool Config::ignoreIfNotModified(const std::string &key, const std::string &value) const {
+  auto iter = fields_.find(key);
+  return iter != fields_.end() && iter->second->Default() == value;
+}
```
PragmaTwice: From the semantics of this function, I think it's better to be named as `checkOptionIsDefault` or something like that since it has no operation about "ignore".

torwig: Longer option from me: `isFieldValueSetToDefault` :)

git-hulk: ok, makes sense.

git-hulk: I guess we can use `checkFieldValueIsDefault`

torwig (approved):
LGTM

PR #1886, opened 2023-11-08
fix integer paring error in listpack and ziplist when load RDB

w4096:
When parsing listpack and ziplist when loading RDB, they don't correctly handle negetive integer. in This PR i try to fix this.

When I fix the problem, I think we can make the code more simpler, and I can't stop myself do those changes:

- fix int parse bug in listpack and ziplist
- reimplement the next method using a input string stream
- remove unused public method of ZipList and ListPack (Length and Next). User can't call this method because it consume the input. Next and Length is used by Entries, we should make them private.

In `InputStringStream` I use execption rather than `StatusOr`. If using `StatusOr`, I must wrap every `stream.Read` call in `GET_OR_RET`, this  make code hard to read.

By using `InputStringStream` you can easily handle offset, you can easily read certain bytes integer and string.

### DIFF
--- a/src/common/string_stream.h
+++ b/src/common/string_stream.h
@@ -21,25 +21,49 @@
 #pragma once
 
 #include <string>
-#include <string_view>
-#include <vector>
 
-#include "common/status.h"
-#include "common/string_stream.h"
 #include "endianconv.h"
+#include "status.h"
 
-class ListPack {
+class InputStringStream {
  public:
-  explicit ListPack(std::string_view input) : stream_(input) {}
-  ~ListPack() = default;
+  InputStringStream() = default;
 
-  StatusOr<std::vector<std::string>> Entries();
+  explicit InputStringStream(std::string_view input);
+
+  size_t ReadableSize() const;
+  const char *Data() const;
+  void Consume(size_t size);
+
+  template <typename T>
+  T Read();
+
+  std::string Read(size_t n);
 
  private:
-  InputStringStream stream_;
+  void peak(size_t n) const;
 
-  static uint32_t encodeBackLen(uint32_t len);
-
-  uint32_t length();
-  std::string next();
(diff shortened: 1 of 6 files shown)


mapleFU:
Thanks for the bugfix

However, changing the style from Status to exception only in this submodule is so weird...

src/storage/rdb_listpack.cc
```diff
@@ -137,88 +120,66 @@ uint32_t ListPack::encodeBackLen(uint32_t len) {
...
+    value = stream_.Read(len);
+    entry_bytes = 2 + len;
+  } else if ((encoding & ListPack32BitStringMask) == ListPack32BitString) {
+    // 1111|0000 <4 bytes len> <large string>
+    len = stream_.Read<uint32_t>();
+    value = stream_.Read(len);
+    entry_bytes += 1 + sizeof(uint32_t) + len;
   } else {
-    return {Status::NotOK, "invalid listpack entry"};
+    throw std::runtime_error("invalid listpack entry");
```
PragmaTwice, line 178: I don't think it is reasonable to change from `StatusOr` back to exceptions here and only here, since it breaks the consistency of error handling.

PragmaTwice (changes requested):

src/common/string_stream.h
```diff
@@ -0,0 +1,70 @@
...
+  template <typename T>
+  T Read();
+
+  std::string Read(size_t n);
+
+ private:
+  void check(size_t n) const;
+
+  std::string_view input_;
+  size_t pos_{0};
```
PragmaTwice, line 47: IMHO this is weird, since you can just slice the `string_view` instead of adding a `pos_`.

git-hulk:
Close PR due to lack of active, free feel to reopen if the issue is still living.

PR #1888, opened 2023-11-09
support pinnable slice get

### LINKED ISSUE #1722: Support `Get` with `rocksdb::PinnableSlice` to decrease copy for BLOB
### Search before asking

### Motivation

RocksDB has support Get with `PinnableSlice` [1][2], and it's already used in `MultiGet` in our kvrocks[3]. For most of our objects, they're small object. So this might not get any benefits. However, as for type like Bitmap, BloomFilter. This could avoid copying memory.

[1] https://rocksdb.org/blog/2017/08/24/pinnableslice.html
[2] https://github.com/facebook/rocksdb/blob/main/include/rocksdb/db.h
[3] https://github.com/apache/kvrocks/blob/e2baeb1b2cdeb480d02b74d4eff23f2c418bc3e7/src/storage/storage.h#L109

### Solution

_No response_

chrisxu333:
This closes https://github.com/apache/kvrocks/issues/1722
Support Get with rocksdb::PinnableSlice, so that relatively large objects like Bitmap and BloomFilter could avoid in-memory copy.

### DIFF
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -531,6 +531,19 @@ rocksdb::Status Storage::Get(const rocksdb::ReadOptions &options, rocksdb::Colum
   return db_->Get(options, column_family, key, value);
 }
 
+rocksdb::Status Storage::Get(const rocksdb::ReadOptions &options, const rocksdb::Slice &key,
+                             rocksdb::PinnableSlice *value) {
+  return Get(options, db_->DefaultColumnFamily(), key, value);
+}
+
+rocksdb::Status Storage::Get(const rocksdb::ReadOptions &options, rocksdb::ColumnFamilyHandle *column_family,
+                             const rocksdb::Slice &key, rocksdb::PinnableSlice *value) {
+  if (is_txn_mode_ && txn_write_batch_->GetWriteBatch()->Count() > 0) {
+    return txn_write_batch_->GetFromBatchAndDB(db_.get(), options, column_family, key, value);
+  }
+  return db_->Get(options, column_family, key, value);
+}
+
 rocksdb::Iterator *Storage::NewIterator(const rocksdb::ReadOptions &options) {
   return NewIterator(options, db_->DefaultColumnFamily());
 }
--- a/src/types/redis_bitmap.cc
+++ b/src/types/redis_bitmap.cc
@@ -251,7 +251,7 @@ rocksdb::Status Bitmap::BitCount(const Slice &user_key, int64_t start, int64_t s
     std::string sub_key =
         InternalKey(ns_key, std::to_string(i * kBitmapSegmentBytes), metadata.version, storage_->IsSlotIdEncoded())
             .Encode();
-    s = storage_->Get(read_options, storage_->GetDB()->DefaultColumnFamily(), sub_key, &pin_value);
+    s = storage_->Get(read_options, sub_key, &pin_value);
     if (!s.ok() && !s.IsNotFound()) return s;
     if (s.IsNotFound()) continue;
     size_t j = 0;
@@ -309,7 +309,7 @@ rocksdb::Status Bitmap::BitPos(const Slice &user_key, bool bit, int64_t start, i
     std::string sub_key =
         InternalKey(ns_key, std::to_string(i * kBitmapSegmentBytes), metadata.version, storage_->IsSlotIdEncoded())
             .Encode();
-    s = storage_->Get(read_options, storage_->GetDB()->DefaultColumnFamily(), sub_key, &pin_value);
+    s = storage_->Get(read_options, sub_key, &pin_value);
     if (!s.ok() && !s.IsNotFound()) return s;
     if (s.IsNotFound()) {
       if (!bit) {
--- a/src/types/redis_bitmap.cc
+++ b/src/types/redis_bitmap.cc
@@ -246,19 +246,19 @@ rocksdb::Status Bitmap::BitCount(const Slice &user_key, int64_t start, int64_t s
   uint32_t start_index = u_start / kBitmapSegmentBytes;
   uint32_t stop_index = u_stop / kBitmapSegmentBytes;
   // Don't use multi get to prevent large range query, and take too much memory
-  std::string value;
   for (uint32_t i = start_index; i <= stop_index; i++) {
+    rocksdb::PinnableSlice pin_value;
     std::string sub_key =
         InternalKey(ns_key, std::to_string(i * kBitmapSegmentBytes), metadata.version, storage_->IsSlotIdEncoded())
             .Encode();
-    s = storage_->Get(read_options, sub_key, &value);
+    s = storage_->Get(read_options, storage_->GetDB()->DefaultColumnFamily(), sub_key, &pin_value);
     if (!s.ok() && !s.IsNotFound()) return s;
     if (s.IsNotFound()) continue;
     size_t j = 0;
     if (i == start_index) j = u_start % kBitmapSegmentBytes;
-    auto k = static_cast<int64_t>(value.size());
+    auto k = static_cast<int64_t>(pin_value.size());
     if (i == stop_index) k = u_stop % kBitmapSegmentBytes + 1;
-    *cnt += BitmapString::RawPopcount(reinterpret_cast<const uint8_t *>(value.data()) + j, k);
+    *cnt += BitmapString::RawPopcount(reinterpret_cast<const uint8_t *>(pin_value.data()) + j, k);
   }
   return rocksdb::Status::OK();
 }
@@ -304,12 +304,12 @@ rocksdb::Status Bitmap::BitPos(const Slice &user_key, bool bit, int64_t start, i
   uint32_t start_index = u_start / kBitmapSegmentBytes;
   uint32_t stop_index = u_stop / kBitmapSegmentBytes;
   // Don't use multi get to prevent large range query, and take too much memory
-  std::string value;
+  rocksdb::PinnableSlice pin_value;
   for (uint32_t i = start_index; i <= stop_index; i++) {
     std::string sub_key =
         InternalKey(ns_key, std::to_string(i * kBitmapSegmentBytes), metadata.version, storage_->IsSlotIdEncoded())
             .Encode();
-    s = storage_->Get(read_options, sub_key, &value);
+    s = storage_->Get(read_options, storage_->GetDB()->DefaultColumnFamily(), sub_key, &pin_value);
     if (!s.ok() && !s.IsNotFound()) return s;
     if (s.IsNotFound()) {
       if (!bit) {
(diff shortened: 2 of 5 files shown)


src/storage/storage.cc
```diff
@@ -528,6 +528,14 @@ rocksdb::Status Storage::Get(const rocksdb::ReadOptions &options, rocksdb::Colum
   return db_->Get(options, column_family, key, value);
 }
 
+rocksdb::Status Storage::Get(const rocksdb::ReadOptions &options, rocksdb::ColumnFamilyHandle *column_family,
```
mapleFU: we can also support a 

```
rocksdb::Status Storage::Get(const rocksdb::ReadOptions &options, const rocksdb::Slice &key, rocksdb::PinnableSlice *value) {
  return Get(options, db_->DefaultColumnFamily(), key, value);
}
```

Wrapper, and make caller more convient

src/types/redis_bloom_chain.cc
```diff
@@ -51,11 +51,13 @@ rocksdb::Status BloomChain::getBFDataList(const std::vector<std::string> &bf_key
...
-    std::string bf_data;
-    rocksdb::Status s = storage_->Get(read_options, bf_key, &bf_data);
+  rocksdb::PinnableSlice pin_value;
+  for (int i = 0; i < bf_key_list.size(); ++i) {
+    auto &bf_key = bf_key_list[i];
+    rocksdb::Status s = storage_->Get(read_options, storage_->GetDB()->DefaultColumnFamily(), bf_key, &pin_value);
     if (!s.ok()) return s;
-    bf_data_list->push_back(std::move(bf_data));
+    bf_data_list->emplace_back(pin_value.ToString());
+    pin_value.Reset();
```
mapleFU: I think `getBFDataList` can also get a slice to avoid copying here 🤔

mapleFU (commented):
General LGTM! Looking forward testing here!

src/types/redis_bloom_chain.cc
chrisxu333: Do you mean bf_data_list would hold a list of slice instead of raw string? Namely change `std::vector<std::string>` to `std::vector<Slice>`?

mapleFU: Yes and no, changing `std::vector<std::string>` to `std::vector<Slice>` might suffer from lifetime problem, we need to change it to `std::vector<PinnableSlice>`

chrisxu333: Got it. Yea that's what I meant :)

src/types/redis_bitmap.cc
```diff
@@ -246,19 +246,20 @@ rocksdb::Status Bitmap::BitCount(const Slice &user_key, int64_t start, int64_t s
   uint32_t start_index = u_start / kBitmapSegmentBytes;
   uint32_t stop_index = u_stop / kBitmapSegmentBytes;
   // Don't use multi get to prevent large range query, and take too much memory
-  std::string value;
+  rocksdb::PinnableSlice pin_value;
```
git-hulk: We should move this line into for-loop? so that we don't need to manually call the pin_value.Reset.

chrisxu333: Yes you're right. Reset() cost more time than creating for-loop local vars.

src/types/redis_bloom_chain.cc
```diff
@@ -45,17 +45,18 @@ void BloomChain::getBFKeyList(const Slice &ns_key, const BloomChainMetadata &met
...
+                                          std::vector<rocksdb::Slice> *bf_data_list) {
   LatestSnapShot ss(storage_);
   rocksdb::ReadOptions read_options;
   read_options.snapshot = ss.GetSnapShot();
 
   bf_data_list->reserve(bf_key_list.size());
-  for (const auto &bf_key : bf_key_list) {
-    std::string bf_data;
-    rocksdb::Status s = storage_->Get(read_options, bf_key, &bf_data);
+  for (size_t i = 0; i < bf_key_list.size(); ++i) {
```
PragmaTwice: I wonder if there is a reason to change the range-based for loop to the classic one?

chrisxu333: I think I accidentally did that for testing purposes. I'll change it back.

PragmaTwice:
CI failed on macOS and ASAN build (use-after-free behaviors are found).

src/types/redis_bloom_chain.cc
```diff
@@ -45,17 +45,17 @@ void BloomChain::getBFKeyList(const Slice &ns_key, const BloomChainMetadata &met
 }
 
 rocksdb::Status BloomChain::getBFDataList(const std::vector<std::string> &bf_key_list,
-                                          std::vector<std::string> *bf_data_list) {
+                                          std::vector<rocksdb::Slice> *bf_data_list) {
```
```suggestion
                                          std::vector<rocksdb::PinnableSlice> *bf_data_list) {
```

chrisxu333:
> CI failed on macOS and ASAN build (use-after-free behaviors are found).

Okay I went through the code and I thought it might be related to pinnable slice copy into bf_data_list container.

src/types/redis_bitmap.cc
```diff
@@ -246,19 +246,19 @@ rocksdb::Status Bitmap::BitCount(const Slice &user_key, int64_t start, int64_t s
...
   uint32_t stop_index = u_stop / kBitmapSegmentBytes;
   // Don't use multi get to prevent large range query, and take too much memory
-  std::string value;
   for (uint32_t i = start_index; i <= stop_index; i++) {
+    rocksdb::PinnableSlice pin_value;
     std::string sub_key =
         InternalKey(ns_key, std::to_string(i * kBitmapSegmentBytes), metadata.version, storage_->IsSlotIdEncoded())
             .Encode();
-    s = storage_->Get(read_options, sub_key, &value);
+    s = storage_->Get(read_options, storage_->GetDB()->DefaultColumnFamily(), sub_key, &pin_value);
```
mapleFU: This can use the default version

src/types/redis_bitmap.cc
```diff
@@ -304,12 +304,12 @@ rocksdb::Status Bitmap::BitPos(const Slice &user_key, bool bit, int64_t start, i
...
   uint32_t stop_index = u_stop / kBitmapSegmentBytes;
   // Don't use multi get to prevent large range query, and take too much memory
-  std::string value;
+  rocksdb::PinnableSlice pin_value;
   for (uint32_t i = start_index; i <= stop_index; i++) {
     std::string sub_key =
         InternalKey(ns_key, std::to_string(i * kBitmapSegmentBytes), metadata.version, storage_->IsSlotIdEncoded())
             .Encode();
-    s = storage_->Get(read_options, sub_key, &value);
+    s = storage_->Get(read_options, storage_->GetDB()->DefaultColumnFamily(), sub_key, &pin_value);
```
mapleFU: ditto

src/types/redis_bloom_chain.cc
```diff
@@ -180,7 +180,8 @@ rocksdb::Status BloomChain::InsertCommon(const Slice &user_key, const std::vecto
     bool exist = false;
     // TODO: to test which direction for searching is better
     for (int ii = static_cast<int>(bf_data_list.size()) - 1; ii >= 0; --ii) {
-      exist = bloomCheck(item_hash_list[i], bf_data_list[ii]);
+      std::string data = bf_data_list[ii].ToString();
+      exist = bloomCheck(item_hash_list[i], data);
```
mapleFU: Would you mind change it to `std::string_view` to avoid a round of clone?

src/types/redis_bloom_chain.cc
```diff
@@ -190,17 +191,22 @@ rocksdb::Status BloomChain::InsertCommon(const Slice &user_key, const std::vecto
     } else {
       if (metadata.size + 1 > metadata.GetCapacity()) {
         if (metadata.IsScaling()) {
-          batch->Put(bf_key_list.back(), bf_data_list.back());
+          batch->Put(bf_key_list.back(), bf_data_list.back().ToString());
```
mapleFU: Would you mind change it to `std::string_view` to avoid a round of clone?

chrisxu333:
Hi @mapleFU , I saw all tests are passed for this pr. So do you mind helping merge this pr after I rebase it with the `unstable` branch

mapleFU:
Generall LGTM. I'll take a carefully pass before merging tomorrow

src/types/redis_bloom_chain.cc
```diff
@@ -190,17 +192,22 @@ rocksdb::Status BloomChain::InsertCommon(const Slice &user_key, const std::vecto
     } else {
       if (metadata.size + 1 > metadata.GetCapacity()) {
         if (metadata.IsScaling()) {
-          batch->Put(bf_key_list.back(), bf_data_list.back());
+          batch->Put(bf_key_list.back(), bf_data_list.back().ToStringView());
           std::string bf_data;
           createBloomFilterInBatch(ns_key, &metadata, batch, &bf_data);
-          bf_data_list.push_back(std::move(bf_data));
+          rocksdb::PinnableSlice pin_slice;
+          pin_slice.PinSelf(rocksdb::Slice(bf_data));
```
mapleFU: Hmmm this part the logic is:

```
get bf_data ( a string ) in createBloomFilterInBatch
pin_slice.Pin(bf_data), which will clone the bf_data
```

Would you mind try:

```
*pin_slice.GetSelf() = std::move(bf_data);
pin_slice.PinSelf();
```

I think it might be better than `PinSelf` to avoid a round of copying?

src/types/redis_bloom_chain.cc
```diff
@@ -190,17 +192,22 @@ rocksdb::Status BloomChain::InsertCommon(const Slice &user_key, const std::vecto
...
         } else {
           (*rets)[i] = BloomFilterAddResult::kFull;
           continue;
         }
       }
-      bloomAdd(item_hash_list[i], bf_data_list.back());
+      std::string data = bf_data_list.back().ToString();
+      bloomAdd(item_hash_list[i], data);
+      bf_data_list.back().Reset();
+      bf_data_list.back().PinSelf(rocksdb::Slice(data));
```
mapleFU: ditto

mapleFU (dismissed):
LGTM, just a few nits

src/types/redis_bloom_chain.cc
mapleFU: (This trick is used in `PinnableWideColumns::MoveValue` )

```
inline void PinnableWideColumns::MoveValue(std::string&& value) {
  std::string* const buf = value_.GetSelf();
  assert(buf);

  *buf = std::move(value);
  value_.PinSelf();
}
```

chrisxu333: Thanks for bring this up! I actually didn't realize this way to implementing it :)

mapleFU (approved):
Will merge in one day if CI passed and other reviewer think it's ok

git-hulk (approved):

PR #1891, opened 2023-11-11
Fix potentially overflow msg_namelen value in SupervisedSystemd

PragmaTwice:
https://github.com/apache/kvrocks/blob/78f87c409348bbde6839ccf4d7bb2be207efbf5e/src/cli/daemon_util.h#L75

The value of `msg_namelen` may cause potentially overflow since `notify_socket` is not the real truncated one. So we need to use `strlen(notify_socket)` instead of `strlen(su.sun_path)`.

Besides, some other code has been cleaned up.

### DIFF
--- a/src/cli/daemon_util.h
+++ b/src/cli/daemon_util.h
@@ -56,23 +56,20 @@ inline bool SupervisedSystemd() {
     return false;
   }
 
-  sockaddr_un su;
-  memset(&su, 0, sizeof(su));
+  sockaddr_un su = {};
   su.sun_family = AF_UNIX;
   strncpy(su.sun_path, notify_socket, sizeof(su.sun_path) - 1);
   su.sun_path[sizeof(su.sun_path) - 1] = '\0';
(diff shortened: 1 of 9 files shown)


git-hulk (approved):

git-hulk:
@PragmaTwice Sonar Lint reports hotspot security due to using the stren function in daemon_util.h, but I guess it should be fine since the variable is always set the `\0`.

torwig (approved):

PragmaTwice:
> @PragmaTwice Sonar Lint reports hotspot security due to using the stren function in daemon_util.h, but I guess it should be fine since the variable is always set the `\0`.

This `strlen` is not introduced by me lol.

git-hulk:
> > @PragmaTwice Sonar Lint reports hotspot security due to using the stren function in daemon_util.h, but I guess it should be fine since the variable is always set the `\0`.
> 
> This `strlen` is not introduced by me lol.

aha, yes i know. What I mean is to ignore this hotspot security report.

PR #1892, opened 2023-11-12
fix go test `JSON.TOGGLE basics` parameters order

jihuayu:
The PR has been fixed in the `JSON.TOGGLE basics` test case, where the parameters order of the `require.Equal` function was incorrect.


torwig (approved):
@jihuayu Good catch.
LGTM if there are no lint issues.

git-hulk (approved):

Issue #1896, opened 2023-11-18
Add support for the BITFIELD command


julic20s:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

This issue considers to support the [bitfield](https://redis.io/commands/bitfield/) command.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PR #1901, opened 2023-11-20
Support for the BITFIELD command

julic20s:
#1896

### DIFF
--- a/src/commands/cmd_bit.cc
+++ b/src/commands/cmd_bit.cc
@@ -19,6 +19,7 @@
  */
 
 #include "commander.h"
+#include "commands/command_parser.h"
 #include "error_constants.h"
 #include "server/server.h"
 #include "types/redis_bitmap.h"
@@ -132,6 +133,8 @@ class CommandBitCount : public Commander {
 
 class CommandBitPos : public Commander {
  public:
+  using Commander::Parse;
+
   Status Parse(const std::vector<std::string> &args) override {
     if (args.size() >= 4) {
       auto parse_start = ParseInt<int64_t>(args[3], 10);
@@ -225,10 +228,158 @@ class CommandBitOp : public Commander {
   BitOpFlags op_flag_;
 };
 
+class CommandBitfield : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    BitfieldOperation cmd;
+
+    read_only_ = true;
+    // BITFIELD <key> [commands...]
+    for (CommandParser group(args, 2); group.Good();) {
+      auto remains = group.Remains();
+
+      std::string opcode = util::ToLower(group[0]);
+      if (opcode == "get") {
+        cmd.type = BitfieldOperation::Type::kGet;
+      } else if (opcode == "set") {
+        cmd.type = BitfieldOperation::Type::kSet;
+        read_only_ = false;
+      } else if (opcode == "incrby") {
+        cmd.type = BitfieldOperation::Type::kIncrBy;
+        read_only_ = false;
+      } else if (opcode == "overflow") {
+        constexpr auto kOverflowCmdSize = 2;
+        if (remains < kOverflowCmdSize) {
+          return {Status::RedisParseErr, errWrongNumOfArguments};
+        }
+        auto s = parseOverflowSubCommand(group[1], &cmd);
+        if (!s.IsOK()) {
+          return s;
+        }
+
+        group.Skip(kOverflowCmdSize);
+        continue;
+      } else {
+        return {Status::RedisParseErr, errUnknownSubcommandOrWrongArguments};
+      }
+
+      if (remains < 3) {
+        return {Status::RedisParseErr, errWrongNumOfArguments};
+      }
+
+      // parse encoding
+      auto encoding = parseBitfieldEncoding(group[1]);
+      if (!encoding.IsOK()) {
+        return encoding.ToStatus();
+      }
+      cmd.encoding = encoding.GetValue();
+
+      // parse offset
+      if (!GetBitOffsetFromArgument(group[2], &cmd.offset).IsOK()) {
+        return {Status::RedisParseErr, "bit offset is not an integer or out of range"};
+      }
+
+      if (cmd.type != BitfieldOperation::Type::kGet) {
+        if (remains < 4) {
+          return {Status::RedisParseErr, errWrongNumOfArguments};
+        }
+
+        auto value = ParseInt<int64_t>(group[3], 10);
+        if (!value.IsOK()) {
+          return value.ToStatus();
+        }
+        cmd.value = value.GetValue();
+
+        // SET|INCRBY <encoding> <offset> <value>
+        group.Skip(4);
+      } else {
+        // GET <encoding> <offset>
+        group.Skip(3);
+      }
+
+      cmds_.push_back(cmd);
+    }
+
+    return Commander::Parse(args);
+  }
+
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    redis::Bitmap bitmap_db(srv->storage, conn->GetNamespace());
+    std::vector<std::optional<BitfieldValue>> rets;
+    rocksdb::Status s;
+    if (read_only_) {
+      s = bitmap_db.BitfieldReadOnly(args_[1], cmds_, &rets);
+    } else {
+      s = bitmap_db.Bitfield(args_[1], cmds_, &rets);
+    }
+    std::vector<std::string> str_rets(rets.size());
+    for (size_t i = 0; i != rets.size(); ++i) {
+      if (rets[i].has_value()) {
+        if (rets[i]->Encoding().IsSigned()) {
+          str_rets[i] = redis::Integer(CastToSignedWithoutBitChanges(rets[i]->Value()));
+        } else {
+          str_rets[i] = redis::Integer(rets[i]->Value());
+        }
+      } else {
+        str_rets[i] = redis::NilString();
+      }
+    }
+    *output = redis::Array(str_rets);
+    return Status::OK();
+  }
+
+ private:
+  static Status parseOverflowSubCommand(const std::string &overflow, BitfieldOperation *cmd) {
+    std::string lower = util::ToLower(overflow);
+    if (lower == "wrap") {
+      cmd->overflow = BitfieldOverflowBehavior::kWrap;
+    } else if (lower == "sat") {
+      cmd->overflow = BitfieldOverflowBehavior::kSat;
+    } else if (lower == "fail") {
+      cmd->overflow = BitfieldOverflowBehavior::kFail;
+    } else {
+      return {Status::RedisParseErr, errUnknownSubcommandOrWrongArguments};
+    }
+    return Status::OK();
+  }
+
+  static StatusOr<BitfieldEncoding> parseBitfieldEncoding(const std::string &token) {
+    if (token.empty()) {
+      return {Status::RedisParseErr, errUnknownSubcommandOrWrongArguments};
+    }
+
+    auto sign = std::tolower(token[0]);
+    if (sign != 'u' && sign != 'i') {
+      return {Status::RedisParseErr, errUnknownSubcommandOrWrongArguments};
+    }
+
+    auto type = BitfieldEncoding::Type::kUnsigned;
+    if (sign == 'i') {
+      type = BitfieldEncoding::Type::kSigned;
+    }
+
+    auto bits_parse = ParseInt<uint8_t>(token.substr(1), 10);
+    if (!bits_parse.IsOK()) {
+      return bits_parse.ToStatus();
+    }
+    uint8_t bits = bits_parse.GetValue();
+
+    auto encoding = BitfieldEncoding::Create(type, bits);
+    if (!encoding.IsOK()) {
+      return {Status::RedisParseErr, errUnknownSubcommandOrWrongArguments};
+    }
+    return encoding.GetValue();
+  }
+
+  std::vector<BitfieldOperation> cmds_;
+  bool read_only_;
+};
+
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandGetBit>("getbit", 3, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandSetBit>("setbit", 4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandBitCount>("bitcount", -2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandBitPos>("bitpos", -3, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandBitOp>("bitop", -4, "write", 2, -1, 1), )
+                        MakeCmdAttr<CommandBitOp>("bitop", -4, "write", 2, -1, 1),
+                        MakeCmdAttr<CommandBitfield>("bitfield", -2, "write", 1, 1, 1), )
 
 }  // namespace redis
--- a/src/commands/command_parser.h
+++ b/src/commands/command_parser.h
@@ -41,6 +41,9 @@ struct CommandParser {
  public:
   using ValueType = typename Iter::value_type;
 
+  static constexpr bool IsRandomAccessIter =
+      std::is_base_of_v<std::random_access_iterator_tag, typename std::iterator_traits<Iter>::iterator_category>;
+
   CommandParser(Iter begin, Iter end) : begin_(std::move(begin)), end_(std::move(end)) {}
 
   template <typename Container>
@@ -56,11 +59,40 @@ struct CommandParser {
 
   decltype(auto) RawPeek() const { return *begin_; }
 
+  decltype(auto) operator[](size_t index) const {
+    Iter iter = begin_;
+    std::advance(iter, index);
+    return *iter;
+  }
+
   decltype(auto) RawTake() { return *begin_++; }
 
   decltype(auto) RawNext() { ++begin_; }
 
   bool Good() const { return begin_ != end_; }
+
+  std::enable_if_t<IsRandomAccessIter, size_t> Remains() const {
+    // O(1) iff Iter is random access iterator.
+    auto d = std::distance(begin_, end_);
+    DCHECK(d >= 0);
+    return d;
+  }
+
+  size_t Skip(size_t count) {
+    if constexpr (IsRandomAccessIter) {
+      size_t steps = std::min(Remains(), count);
+      begin_ += steps;
+      return steps;
+    } else {
+      size_t steps = 0;
+      while (count != 0 && Good()) {
+        ++begin_;
+        ++steps;
+        --count;
+      }
+      return steps;
+    }
+  }
 
   template <typename Pred>
   bool EatPred(Pred&& pred) {
--- a/src/common/bitfield_util.cc
+++ b/src/common/bitfield_util.cc
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#include "bitfield_util.h"
+
+namespace detail {
+
+StatusOr<bool> SignedBitfieldPlus(uint64_t value, int64_t incr, uint8_t bits, BitfieldOverflowBehavior overflow,
+                                  uint64_t *dst) {
+  if (!BitfieldEncoding::IsSupportedBitLengths(BitfieldEncoding::Type::kSigned, bits)) {
+    return Status::NotOK;
+  }
+
+  auto max = std::numeric_limits<int64_t>::max();
+  if (bits != 64) {
+    max = (static_cast<int64_t>(1) << (bits - 1)) - 1;
+  }
+  int64_t min = -max - 1;
+
+  int64_t signed_value = CastToSignedWithoutBitChanges(value);
+  int64_t max_incr = CastToSignedWithoutBitChanges(static_cast<uint64_t>(max) - value);
+  int64_t min_incr = min - signed_value;
+
+  auto wrap = [=]() {
+    uint64_t res = value + static_cast<uint64_t>(incr);
+    if (bits < 64) {
+      auto mask = std::numeric_limits<uint64_t>::max() << bits;
+      if ((res & (1 << (bits - 1))) != 0) {
+        res |= mask;
+      } else {
+        res &= ~mask;
+      }
+    }
+    return res;
+  };
+
+  if (signed_value > max || (bits != 64 && incr > max_incr) || (signed_value >= 0 && incr >= 0 && incr > max_incr)) {
+    if (overflow == BitfieldOverflowBehavior::kWrap) {
+      *dst = wrap();
+    } else {
+      *dst = max;
+    }
+    return true;
+  } else if (signed_value < min || (bits != 64 && incr < min_incr) ||
+             (signed_value < 0 && incr < 0 && incr < min_incr)) {
+    if (overflow == BitfieldOverflowBehavior::kWrap) {
+      *dst = wrap();
+    } else {
+      *dst = min;
+    }
+    return true;
+  }
+
+  *dst = signed_value + incr;
+  return false;
+}
+
+// return true if overflow.
+StatusOr<bool> UnsignedBitfieldPlus(uint64_t value, int64_t incr, uint8_t bits, BitfieldOverflowBehavior overflow,
+                                    uint64_t *dst) {
+  if (!BitfieldEncoding::IsSupportedBitLengths(BitfieldEncoding::Type::kUnsigned, bits)) {
+    return Status::NotOK;
+  }
+  auto max = (static_cast<uint64_t>(1) << bits) - 1;
+  int64_t max_incr = CastToSignedWithoutBitChanges(max - value);
+  int64_t min_incr = CastToSignedWithoutBitChanges((~value) + 1);
+
+  auto wrap = [=]() {
+    uint64_t mask = std::numeric_limits<uint64_t>::max() << bits;
+    uint64_t res = value + incr;
+    res &= ~mask;
+    return res;
+  };
+
+  if (value > max || (incr > 0 && incr > max_incr)) {
+    if (overflow == BitfieldOverflowBehavior::kWrap) {
+      *dst = wrap();
+    } else if (overflow == BitfieldOverflowBehavior::kSat) {
+      *dst = max;
+    }
+    return true;
+  } else if (incr < 0 && incr < min_incr) {
+    if (overflow == BitfieldOverflowBehavior::kWrap) {
+      *dst = wrap();
+    } else if (overflow == BitfieldOverflowBehavior::kSat) {
+      *dst = 0;
+    }
+    return true;
+  }
+
+  *dst = value + incr;
+  return false;
+}
+
+}  // namespace detail
--- a/src/common/bitfield_util.cc
+++ b/src/common/bitfield_util.cc
@@ -22,10 +22,24 @@
 
 namespace detail {
 
+static uint64_t WrappedSignedBitfieldPlus(uint64_t value, int64_t incr, uint8_t bits) {
+  uint64_t res = value + static_cast<uint64_t>(incr);
+  if (bits < 64) {
+    auto mask = std::numeric_limits<uint64_t>::max() << bits;
+    if ((res & (1 << (bits - 1))) != 0) {
+      res |= mask;
+    } else {
+      res &= ~mask;
+    }
+  }
+  return res;
+}
+
 StatusOr<bool> SignedBitfieldPlus(uint64_t value, int64_t incr, uint8_t bits, BitfieldOverflowBehavior overflow,
                                   uint64_t *dst) {
-  if (!BitfieldEncoding::IsSupportedBitLengths(BitfieldEncoding::Type::kSigned, bits)) {
-    return Status::NotOK;
+  Status bits_status(BitfieldEncoding::CheckSupportedBitLengths(BitfieldEncoding::Type::kSigned, bits));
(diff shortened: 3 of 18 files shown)


src/common/bitfield_util.h
```diff
@@ -0,0 +1,135 @@
...
+#include <stdexcept>
+#include <cstdint>
+#include <limits>
+#include <type_traits>
+
+enum class BitfieldOverflowBehavior : uint8_t { kWrap, kSat, kFail };
+
+struct BitfieldEncoding {
+  bool is_signed;
+  // 1 ~ 63 if unsigned, 1 ~ 64 if signed.
```
mapleFU: Should we `static_assert(sizeof(...))` for this?

src/common/bitfield_util.h
```diff
@@ -0,0 +1,135 @@
...
+  // INCRBY amount or SET value
+  int64_t value;
+};
+
+namespace detail {
+
+// safe cast from unsigned to signed.
+// see also "Integral conversions" on https://en.cppreference.com/w/cpp/language/implicit_conversion
+// If the destination type is signed, the result when overflow is implementation-defined until C++20
+union SafeInt64 {
```
mapleFU: Would we better just using `uint64_t` and maintaining a group of `SafeOp` rather than `SafeInt64`?

src/types/redis_bitmap.cc
```diff
@@ -538,6 +545,302 @@ rocksdb::Status Bitmap::BitOp(BitOpFlags op_flag, const std::string &op_name, co
   return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
 }
 
+namespace {
+union IntResult {
```
mapleFU: ditto. Seems this has duplicate IntResult...

mapleFU (commented):
Ooops seems this patch is extremly huge...I need take time for review, just some tiny comments first.

git-hulk:
@julic20s Can use `./x.py format` to format your code before pushing.

julic20s:
> @julic20s Can use `./x.py format` to format your code before pushing.

Sorry, I didn't see it at first, I just used shortcut formatting in the ide

git-hulk:
> > @julic20s Can use `./x.py format` to format your code before pushing.
> 
> Sorry, I didn't see it at first, I just used shortcut formatting in the ide

ooh, that's fine. You can also configure clang-format in IDE if supported.

src/common/bitfield_util.h
julic20s: > Should we `static_assert(sizeof(...))` for this?

I don't get it. Should I add likes static_assert(sizeof(BitfieldEncoding) == 2)?
The comment in code means redis only support i1\~i64 and u1\~u63 in bitfield.

mapleFU: Oh I misunderstanding this

julic20s: The code comment incomprehensible. I will update it after.

src/common/bitfield_util.h
```diff
@@ -0,0 +1,180 @@
...
+
+#include <cstdint>
+#include <cstring>
+#include <limits>
+#include <stdexcept>
+#include <string>
+
+enum class BitfieldOverflowBehavior : uint8_t { kWrap, kSat, kFail };
+
+class BitfieldBitsLengthNotFitExecption : public std::runtime_error {
```
mapleFU: Hmmm for Style kvrocks uses `Status` rather than Exception, would you mind using status to wrap this?

src/common/bitfield_util.h
```diff
@@ -0,0 +1,180 @@
...
+  // see also https://redis.io/commands/bitfield/
+  static bool IsSupportedBitLengths(Type type, uint8_t bits) {
+    uint8_t max_bits = 64;
+    if (type == Type::kUnsigned) {
+      max_bits = 63;
+    }
+    return 1 <= bits && bits <= max_bits;
+  }
+
+  BitfieldEncoding(Type type, uint8_t bits) : type_(type), bits_(bits) {
```
mapleFU: This can be wrapped as a `Result<BitfieldEncoding>` builder, and `BitfieldEncoding` as a private ctor.

src/common/bitfield_util.h
```diff
@@ -0,0 +1,180 @@
...
+// see also "Integral conversions" on https://en.cppreference.com/w/cpp/language/implicit_conversion
+// If the destination type is signed, the result when overflow is implementation-defined until C++20
+inline int64_t CastToSignedWithoutBitChanges(uint64_t x) {
+  int64_t res = 0;
+  memcpy(&res, &x, sizeof(res));
+  return res;
+}
+
+// return true if overflow.
+inline auto BitfieldPlus(uint64_t value, int64_t incr, BitfieldEncoding enc, BitfieldOverflowBehavior overflow,
```
mapleFU: Better using `bool` as response explicit?

src/common/bitfield_util.h
```diff
@@ -0,0 +1,180 @@
...
+  if (op.type == BitfieldOperation::Type::kSet) {
+    overflow = BitfieldPlus(op.value, 0, op.encoding, op.overflow, new_value);
+  } else {
+    overflow = BitfieldPlus(old_value, op.value, op.encoding, op.overflow, new_value);
+  }
+
+  return op.overflow != BitfieldOverflowBehavior::kFail || !overflow;
+}
+
+template <class BitmapView>
```
mapleFU: We can leave a todo here for optimizations. I didn't try it in `-O3` but it looks slow if bits isn't small

src/common/bitfield_util.h
```diff
@@ -0,0 +1,180 @@
...
+    uint32_t byteval = view[byte];
+    byteval &= ~(1 << bit);
+    byteval |= int(v) << bit;
+    view[byte] = char(byteval);
+    offset++;
+  }
+}
+
+template <class BitmapView>
+inline uint64_t GetUnsignedBitfield(const BitmapView &view, uint64_t offset, uint64_t bits) {
```
mapleFU: ditto, we can leave a todo here?

src/common/bitfield_util.h
```diff
@@ -0,0 +1,180 @@
...
+template <class BitmapView>
+inline int64_t GetSignedBitfield(const BitmapView &view, uint64_t offset, uint64_t bits) {
+  if (!BitfieldEncoding::IsSupportedBitLengths(BitfieldEncoding::Type::kSigned, bits)) {
+    throw BitfieldBitsLengthNotFitExecption();
+  }
+  int64_t value = CastToSignedWithoutBitChanges(GetUnsignedBitfield(view, offset, bits));
+
+  if ((value & (static_cast<uint64_t>(1) << (bits - 1))) != 0) {
+    value |= CastToSignedWithoutBitChanges(std::numeric_limits<uint64_t>::max() << bits);
+  }
```
mapleFU: Hmmm this is a bit tricky here I think...Can we add some comment here?

src/storage/batch_extractor.cc
```diff
@@ -232,6 +232,10 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
               first_seen_ = false;
             }
             break;
+          case kRedisCmdBitfield:
```
mapleFU, line 235: Would you mind testing the slot-migration case here?

src/types/redis_bitmap.cc
```diff
@@ -38,6 +39,22 @@ const char kErrBitmapStringOutOfRange[] =
...
 
+// min_bytes can not more than kBitmapSegmentBytes
+void ExpandBitmapSegment(std::string *segment, size_t min_bytes) {
+  auto old_size = segment->size();
+  if (min_bytes > old_size) {
+    size_t expand_size = 0;
+    if (min_bytes > old_size * 2) {
+      expand_size = min_bytes - old_size;
+    } else if (old_size * 2 > kBitmapSegmentBytes) {
+      expand_size = kBitmapSegmentBytes - old_size;
```
mapleFU: I think we need to check or dcheck `kBitmapSegmentBytes` and size here (as an separate helper function, more checks seems better)

src/commands/cmd_bit.cc
```diff
@@ -225,10 +227,145 @@ class CommandBitOp : public Commander {
...
 };
 
+class CommandBitfield : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    BitfieldOperation cmd;
+
+    read_only_ = true;
+    // BITFIELD <key> [commands...]
+    for (auto group = args.begin() + 2; group != args.end();) {
```
mapleFU: Though this code LGTM, we need to check the argument of the number of `group` doesn't visit the `args.end()`. Would you mind just using the `Parser` here?

mapleFU:
Thanks @julic20s . This patch is huge, and I need some time to review the whole patch. I just submit some small comments.

src/types/redis_bitmap.cc
```diff
@@ -538,6 +545,281 @@ rocksdb::Status Bitmap::BitOp(BitOpFlags op_flag, const std::string &op_name, co
...
+      }
+
+      if (op.encoding.IsSigned()) {
+        ret = redis::Integer(CastToSignedWithoutBitChanges(unsigned_new_value));
+      } else {
+        ret = redis::Integer(unsigned_new_value);
+      }
+    } else {
+      ret = redis::NilString();
+    }
```
PragmaTwice: I think it would be better to return `vector<optional<integer>>` here rather than vector of RESP strings here, in case that we can reuse the result.

src/types/redis_bitmap_string.cc
```diff
@@ -204,4 +205,113 @@ int64_t BitmapString::RawBitpos(const uint8_t *c, int64_t count, bool bit) {
...
+      }
+
+      if (op.encoding.IsSigned()) {
+        ret = redis::Integer(CastToSignedWithoutBitChanges(unsigned_new_value));
+      } else {
+        ret = redis::Integer(unsigned_new_value);
+      }
+    } else {
+      ret = redis::NilString();
+    }
```
PragmaTwice: ditto

src/types/redis_bitmap_string.cc
```diff
@@ -204,4 +205,113 @@ int64_t BitmapString::RawBitpos(const uint8_t *c, int64_t count, bool bit) {
...
+  std::string_view str_;
+};
+
+class ReadWriteStringBitmapView {
+ public:
+  explicit ReadWriteStringBitmapView(std::string *str) : str_(str) {}
+
+  char &operator[](uint32_t index) {
+    if (index >= str_->size()) {
+      str_->append((index + 1) - str_->size(), 0);
```
PragmaTwice:
```suggestion
      str_->resize(index + 1);
```

PragmaTwice (commented):
Thanks for your contribution!

Just some random thoughts without going into the details.

Besides, I believe it would be beneficial to create golang test cases for the recently added command.

src/types/redis_bitmap.cc
julic20s: Should I use std::variant<nullptr_t, uint64_t, int64_t> or define a new type to handle this? There are 3 of types could store in result list (null, unsigned and signed). These two way above can store each result in 16 bytes. But it doesn't seem intuitive enough. (variant is unreadable. new type is hard to reuse, which fits for the 
`Bitmap::Bitfield` function only)

src/commands/cmd_bit.cc
mapleFU: Oh seems `size` is already checked, you can decide whether using parser or current code.

julic20s:
The mutex pool in lock manager may be reallocated by MultiGet(). Then the old mutex pointers in LockGuard all be invalid...I guess that's why I failed the test. Should I use a std::deque to replace the vector?

git-hulk:
> The mutex pool in lock manager may be reallocated by MultiGet(). Then the old mutex pointers in LockGuard all be invalid...I guess that's why I failed the test. Should I use a std::deque to replace the vector?

Guess you should use optional<LockGuard> instead of LockGuard directly in https://github.com/apache/kvrocks/pull/1901/files#diff-642915be939e08850c3ce460033f3cdf0969ab9bfa67b63639e82ab9daf454f6R606-R609.

Or it will release the mutex after exiting the if block, and exiting this function will cause to free the mutex again.

git-hulk:
@julic20s Guess not, you should use `std::optional<LockGuard>` instead of LockGuard directly since the LockGuard constructor will lock the mutex, and if you use LockGuard = default, it won't lock the mutex at all.

refer: https://github.com/apache/kvrocks/blob/a18800976176da06f21f6dd33a7522ef905f5c2f/src/types/redis_set.cc#L205

julic20s:
Oh... I define a default move assignment. Which cause a dangling pointer

julic20s:
Thanks, I get it. The actual cause is that I define a default move assignment function, which cause a dangling pointer.
________________________________
发件人: hulk ***@***.***>
发送时间: 2023年11月22日 16:30
收件人: apache/kvrocks ***@***.***>
抄送: julic20s ***@***.***>; Mention ***@***.***>
主题: Re: [apache/kvrocks] Support for the BITFIELD command (PR #1901)


@julic20s<https://github.com/julic20s> Guess not, you should use std::optional<LockGuard> instead of LockGuard directly since the LockGuard constructor will lock the mutex, and if you use LockGuard = default, it won't lock the mutex at all.

―

julic20s:
```cpp
 LockGuard(LockGuard &&guard) : lock_(guard.lock_) { guard.lock_ = nullptr; } // correct constructor
 LockGuard &operator=(LockGuard &&other) = default; // uncorrect assignment.
```
😭

git-hulk:
@julic20s BTW, would you like to take anonymous functions(get/set/get_segment) from Bitmap::bitfield into named functions. This function is a bit hard to read for now.

src/types/redis_bitmap.cc
PragmaTwice: Maybe it can be `std::optional<struct X>`, where `struct X { uint64_t v; bool signed; std::string to_string(){ ... } }`, if you are willing to change it.

src/common/lock_manager.h
```diff
@@ -81,6 +81,8 @@ class LockManager {
 
 class LockGuard {
  public:
+  LockGuard() = default;
```
PragmaTwice: This ctor will make the state of a LockGuard object more complicated, which is not what we want in my view.

You can refer to https://en.cppreference.com/w/cpp/thread/lock_guard/lock_guard that also doesn't provide such a ctor.

And if you do not want to initialize the object when you define it, please use `std::optional`

src/common/lock_manager.h
```diff
@@ -94,8 +96,10 @@ class LockGuard {
 
   LockGuard(LockGuard &&guard) : lock_(guard.lock_) { guard.lock_ = nullptr; }
 
+  LockGuard &operator=(LockGuard &&) = default;
```
PragmaTwice: IMHO it's not a well-formed move assignment operator, which can lead to double unlock.

src/common/parse_util.h
```diff
@@ -53,13 +58,18 @@ struct ParseIntFunc<long long> {  // NOLINT
   constexpr static const auto value = std::strtoll;
 };
 
+template <>
+struct ParseIntFunc<unsigned char> {  // NOLINT
+  constexpr static const auto value = std::strtol;
```
PragmaTwice:
```suggestion
  constexpr static const auto value = std::strtoul;
```

src/commands/command_parser.h
```diff
@@ -41,6 +41,10 @@ struct CommandParser {
  public:
   using ValueType = typename Iter::value_type;
 
+  static constexpr bool IsRandomAccessIter() noexcept {
+    return std::is_base_of_v<std::random_access_iterator_tag, typename std::iterator_traits<Iter>::iterator_category>;
+  }
```
PragmaTwice:
```suggestion
  static constexpr bool IsRandomAccessIter =
    std::is_base_of_v<std::random_access_iterator_tag, typename std::iterator_traits<Iter>::iterator_category>;
```

src/commands/command_parser.h
```diff
@@ -56,12 +60,39 @@ struct CommandParser {
 
   decltype(auto) RawPeek() const { return *begin_; }
 
+  decltype(auto) operator[](size_t index) const { return begin_[index]; }
```
PragmaTwice: I think you can use something like std::advance here to avoid the requirement of random access iterator.

src/commands/command_parser.h
```diff
@@ -56,12 +60,39 @@ struct CommandParser {
...
   decltype(auto) RawNext() { ++begin_; }
 
   bool Good() const { return begin_ != end_; }
 
+  std::enable_if_t<IsRandomAccessIter(), size_t> Remains() const {
+    // O(1) iff Iter is random access iterator.
+    auto d = std::distance(begin_, end_);
+    if (d < 0) {
+      d = 0;
+    }
```
PragmaTwice: I think maybe we can just DCHECK or assert here, then we can find problem when we encounter an invalid state.

src/commands/command_parser.h
```diff
@@ -56,12 +60,39 @@ struct CommandParser {
...
+    return d;
+  }
+
+  size_t Eat(size_t count) {
```
PragmaTwice:
```suggestion
  size_t Skip(size_t count) {
```

src/types/redis_bitmap.cc
```diff
@@ -538,6 +546,320 @@ rocksdb::Status Bitmap::BitOp(BitOpFlags op_flag, const std::string &op_name, co
...
+
+    if constexpr (ReadOnly) {
+      rets->emplace_back() = {op.encoding, unsigned_old_value};
+      continue;
+    }
+
+    auto &ret = rets->emplace_back();
+    uint64_t unsigned_new_value = 0;
+    // BitfieldOp failed only when the length or bits not illegal.
```
julic20s: There is a typo. Should be `BitfieldOp failed only when the length or bits illegal.`

src/storage/batch_extractor.cc
julic20s: Could you tell me how to run only one test instead of all go test? I wrote the tests and I tried `./x.py test go build redis-cli -run <test function name>`, but no effect. Thanks.

mapleFU:
I'm a bit busy on workdays, I'll go through a pass this weekend. If other thinks it's ok we can review an merge it first

julic20s:
@git-hulk Could you help me approval the workflow? I fix some problems that I found. But I can't reproduction the unittest failed on my local machine. I tried both gcc and clang.

mapleFU:
done

src/common/bitfield_util.h
```diff
@@ -0,0 +1,306 @@
...
+    uint8_t max_bits = 64;
+    if (type == Type::kUnsigned) {
+      max_bits = 63;
+    }
+    return 1 <= bits && bits <= max_bits;
+  }
+
+  static StatusOr<BitfieldEncoding> Create(Type type, uint8_t bits) {
+    if (!IsSupportedBitLengths(type, bits)) {
+      return Status::NotOK;
```
mapleFU: Nit: with a reason to help debugging?

src/common/bitfield_util.h
```diff
@@ -0,0 +1,306 @@
...
+  enum class Type : uint8_t { kSigned, kUnsigned };
+
+  // check whether the bit length is fit to given number sign.
+  // Redis has bits length limitation to bitfield.
+  // Quote:
+  // The supported encodings are up to 64 bits for signed integers, and up to 63 bits for unsigned integers. This
+  // limitation with unsigned integers is due to the fact that currently the Redis protocol is unable to return 64 bit
+  // unsigned integers as replies.
+  // see also https://redis.io/commands/bitfield/
+  static bool IsSupportedBitLengths(Type type, uint8_t bits) {
```
mapleFU: though noexcept is not widely used in kvrocks, since you mark `GetType()` and other as `noexcept`, should this function be `noexcept` as well?

src/common/bitfield_util.h
```diff
@@ -0,0 +1,306 @@
...
+    *new_value = old_value;
+    return true;
+  }
+
+  bool overflow = false;
+  if (op.type == BitfieldOperation::Type::kSet) {
+    auto plus_res = BitfieldPlus(op.value, 0, op.encoding, op.overflow, new_value);
+    if (!plus_res) {
+      return plus_res;
+    }
```
mapleFU: `GET_OR_RET` can used to simplify this code

src/common/bitfield_util.h
```diff
@@ -0,0 +1,306 @@
...
+
+  bool isOutOfBound(uint32_t offset, uint32_t bytes) const noexcept {
+    return offset < offset_ || offset_ + kSize < offset + bytes;
+  }
+
+  StatusOr<uint64_t> getBitfield(uint64_t offset, uint8_t bits) const {
+    uint32_t first_byte = offset / 8;
+    uint32_t last_byte = (offset + bits - 1) / 8 + 1;
+    if (isOutOfBound(first_byte, last_byte - first_byte)) {
+      return Status::NotOK;
```
mapleFU: Maybe a error msg is better here.

src/common/bitfield_util.cc
```diff
@@ -0,0 +1,112 @@
...
+  if (bits != 64) {
+    max = (static_cast<int64_t>(1) << (bits - 1)) - 1;
+  }
+  int64_t min = -max - 1;
+
+  int64_t signed_value = CastToSignedWithoutBitChanges(value);
+  int64_t max_incr = CastToSignedWithoutBitChanges(static_cast<uint64_t>(max) - value);
+  int64_t min_incr = min - signed_value;
+
+  auto wrap = [=]() {
```
mapleFU: Expect extract `wrap` as a function here. Since it's logic can be test easily.

src/common/bitfield_util.h
```diff
@@ -0,0 +1,306 @@
...
+  uint64_t Value() const noexcept { return value_; }
+
+ private:
+  BitfieldEncoding encoding_;
+  uint64_t value_;
+};
+
+// return true if overflow. Status is not ok iff calling BitfieldEncoding::IsSupportedBitLengths()
+// for given op return false.
+inline StatusOr<bool> BitfieldPlus(uint64_t value, int64_t incr, BitfieldEncoding enc,
```
mapleFU, line 166: This function needs more detailed document. It means like check "(value, enc) + incr"

src/common/bitfield_util.cc
```diff
@@ -0,0 +1,112 @@
...
+      auto mask = std::numeric_limits<uint64_t>::max() << bits;
+      if ((res & (1 << (bits - 1))) != 0) {
+        res |= mask;
+      } else {
+        res &= ~mask;
+      }
+    }
+    return res;
+  };
+
```
mapleFU: I tried hard to understand the logic here. The logic is same as redis logic, it tries to:
1. check whether `signed_value` itself overflows
2. if bits < 64, it would be ok to check `max_incr` without caring about underflow
3. do something like `checked_add/checked_minus` and then compare

Do you think just `checked_add` or `checked_sub` using int64_t by `incr` first, and then if that not overflow/underflow, comparing with min-max. Would that be more understandable?

src/common/bitfield_util.cc
```diff
@@ -0,0 +1,112 @@
...
+  auto wrap = [=]() {
+    uint64_t mask = std::numeric_limits<uint64_t>::max() << bits;
+    uint64_t res = value + incr;
+    res &= ~mask;
+    return res;
+  };
+
+  if (value > max || (incr > 0 && incr > max_incr)) {
+    if (overflow == BitfieldOverflowBehavior::kWrap) {
```
mapleFU, line 101: Should we `DCHECK` the behavior is not `kFail`?

src/common/bitfield_util.h
```diff
@@ -0,0 +1,306 @@
...
+  Status Get(uint32_t offset, uint32_t bytes, uint8_t *dst) const {
+    if (isOutOfBound(offset, bytes)) {
+      return Status::NotOK;
+    }
+    offset -= offset_;
+    memcpy(dst, buf_ + offset, bytes);
+    return Status::OK();
+  }
+
+  StatusOr<uint64_t> GetUnsignedBitfield(uint64_t offset, uint64_t bits) const {
```
mapleFU: Can we extract `bit_offset` and byte_offset in variable name?

src/common/bitfield_util.h
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  StatusOr<int64_t> GetSignedBitfield(uint64_t offset, uint64_t bits) const {
+    if (!BitfieldEncoding::IsSupportedBitLengths(BitfieldEncoding::Type::kSigned, bits)) {
+      return Status::NotOK;
+    }
+    auto status = getBitfield(offset, bits);
+    if (!status) {
+      return status;
+    }
```
mapleFU: Using `GET_OR_RET` here

src/types/redis_bitmap.cc
```diff
@@ -38,6 +38,24 @@ const char kErrBitmapStringOutOfRange[] =
...
+  if (min_bytes > old_size) {
+    size_t new_size = 0;
+    if (min_bytes > old_size * 2) {
+      new_size = min_bytes;
+    } else if (old_size * 2 > kBitmapSegmentBytes) {
+      new_size = kBitmapSegmentBytes;
+    } else {
+      new_size = old_size * 2;
+    }
+    segment->resize(new_size);
```
mapleFU: explicit write `resize(new_size, 0)` might better here

src/types/redis_bitmap.cc
```diff
@@ -187,17 +205,7 @@ rocksdb::Status Bitmap::SetBit(const Slice &user_key, uint32_t offset, bool new_
...
-    if (byte_index >= value.size() * 2) {
-      expand_size = byte_index - value.size() + 1;
-    } else if (value.size() * 2 > kBitmapSegmentBytes) {
-      expand_size = kBitmapSegmentBytes - value.size();
-    } else {
-      expand_size = value.size();
-    }
-    value.append(expand_size, 0);
-  }
+  ExpandBitmapSegment(&value, byte_index + 1);
```
mapleFU, line 209: @git-hulk Hmmm I'm a bit confused why original `expand_size = byte_index - value.size() + 1;` has a `+1` here?

src/types/redis_bitmap.cc
```diff
@@ -539,6 +547,321 @@ rocksdb::Status Bitmap::BitOp(BitOpFlags op_flag, const std::string &op_name, co
...
+                    std::string namespace_key, const Metadata &bitmap_metadata)
+      : storage_(storage),
+        metadata_cf_handle_(metadata_cf_handle),
+        ns_key_(std::move(namespace_key)),
+        metadata_(bitmap_metadata) {}
+
+  // Get a read-only segment by given index
+  rocksdb::Status Get(uint32_t index, const std::string **cache) {
+    std::string *res = nullptr;
+    auto s = get(index, false, &res);
```
mapleFU:
```suggestion
    auto s = get(index, /*set_dirty=*/false, &res);
```

src/types/redis_bitmap.cc
```diff
@@ -539,6 +547,321 @@ rocksdb::Status Bitmap::BitOp(BitOpFlags op_flag, const std::string &op_name, co
...
+    if (s.ok()) {
+      *cache = res;
+    }
+    return s;
+  }
+
+  // Get a segment by given index, and mark it dirty.
+  rocksdb::Status GetMut(uint32_t index, std::string **cache) { return get(index, true, cache); }
```
mapleFU:
```suggestion
  rocksdb::Status GetMut(uint32_t index, std::string **cache) { return get(index, /*set_dirty=*/true, cache); }
```

src/types/redis_bitmap.cc
julic20s: @mapleFU byte_index is the last index will be used. It can be explained as `expand_size = (byte_index + 1) /* new size */ - value.size() /* old size */`;

mapleFU: Got it, my bad

src/common/bitfield_util.cc
mapleFU: We can keep the logic here, and we may need to paste the code link to redis

I think later we may use `checked_add` to simplify this logic

julic20s:
I am trying to use rocksdb::PinnableSlice to improve Bitmap::SegmentCacheStore. But I failed because of some strange error. When I use `std::unordered_map<uint32_t, rocksdb::PinnableSlice> cache_` and get its iterator, the iterator compare function is failed to instantiate (such as: `it != cache_.end()`), which makes me can not travel `cache_` in `BatchForFlush()`. `rocksdb::Cleanable` has a bug in current rocksdb version(It's copy constructor deleted, but the param of it is not marked const).

PragmaTwice:
> I am trying to use rocksdb::PinnableSlice to improve Bitmap::SegmentCacheStore. But I failed because of some strange error. When I use `std::unordered_map<uint32_t, rocksdb::PinnableSlice> cache_` and get its iterator, the iterator compare function is failed to instantiate (such as: `it != cache_.end()`), which makes me can not travel `cache_` in `BatchForFlush()`. `rocksdb::Cleanable` has a bug in current rocksdb version(It's copy constructor deleted, but the param of it is not marked const).

I don't think using PinnableSlice is necessary for this PR to be merged.

julic20s:
@mapleFU Hi, I finished all changes you pointed, could you review them again? Thanks.

mapleFU:
Sure, I'm now becoming very busy on workdays, I'll finish review these few days.

src/types/redis_bitmap_string.cc
```diff
@@ -204,4 +205,88 @@ int64_t BitmapString::RawBitpos(const uint8_t *c, int64_t count, bool bit) {
...
+    uint32_t first_byte = op.offset / 8;
+    uint32_t last_byte = (op.offset + op.encoding.Bits() - 1) / 8 + 1;
+
+    ArrayBitfieldBitmap bitfield(first_byte);
+    auto s = bitfield.Set(first_byte, last_byte - first_byte,
+                          reinterpret_cast<const uint8_t *>(string_value.data() + first_byte));
+
+    if (op.encoding.IsSigned()) {
+      int64_t value = bitfield.GetSignedBitfield(op.offset, op.encoding.Bits()).GetValue();
+      rets->emplace_back() = {op.encoding, static_cast<uint64_t>(value)};
```
PragmaTwice:
```suggestion
      rets->emplace_back(std::in_place, op.encoding, static_cast<uint64_t>(value));
```

src/types/redis_bitmap_string.cc
```diff
@@ -204,4 +205,88 @@ int64_t BitmapString::RawBitpos(const uint8_t *c, int64_t count, bool bit) {
...
+    } else {
+      uint64_t value = bitfield.GetUnsignedBitfield(op.offset, op.encoding.Bits()).GetValue();
+      rets->emplace_back() = {op.encoding, value};
```
PragmaTwice: ditto

git-hulk:
@PragmaTwice @mapleFU I think we can go ahead to merge this PR if overall looks good.

PragmaTwice (approved):

PragmaTwice:

Issue #1905, opened 2023-11-23
Quickstart "Error: Server closed the connection" for any command


theronic:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

kvrocks version 2.6.0 (commit c0451df0)

macOS Ventura 13.4 on Apple M2 Pro Macbook Pro 14-inch 2023
Kvrocks: 

### Minimal reproduce step

As per [the docs](https://kvrocks.apache.org/blog/go-redis-kvrocks-opentelemetry/#getting-started-with-kvrocks):

    docker run -it -p 6666:6666 apache/kvrocks

In a separate terminal:

    redis-cli -p 6666
    > GET foo

### What did you expect to see?

Return nil or key value.

### What did you see instead?

On docker terminal:

```
I20231123 11:59:40.362723     1 main.cc:334] kvrocks version 2.6.0 (commit c0451df0)
I20231123 11:59:40.410612     1 storage.cc:332] [storage] Success to load the data from disk: 16 ms
I20231123 11:59:40.413663     1 worker.cc:72] [worker] Listening on: 127.0.0.1:6666
I20231123 11:59:40.414007     1 worker.cc:72] [worker] Listening on: 127.0.0.1:6666
I20231123 11:59:40.414085     1 worker.cc:72] [worker] Listening on: 127.0.0.1:6666
I20231123 11:59:40.414157     1 worker.cc:72] [worker] Listening on: 127.0.0.1:6666
I20231123 11:59:40.414258     1 worker.cc:72] [worker] Listening on: 127.0.0.1:6666
I20231123 11:59:40.414347     1 worker.cc:72] [worker] Listening on: 127.0.0.1:6666
I20231123 11:59:40.414443     1 worker.cc:72] [worker] Listening on: 127.0.0.1:6666
I20231123 11:59:40.414533     1 worker.cc:72] [worker] Listening on: 127.0.0.1:6666
I20231123 11:59:40.414712     1 worker.cc:515] [worker] Thread #281473666143544 started
I20231123 11:59:40.414757     1 worker.cc:515] [worker] Thread #281473654420792 started
I20231123 11:59:40.414800     1 worker.cc:515] [worker] Thread #281473654244664 started
I20231123 11:59:40.414839     1 worker.cc:515] [worker] Thread #281473654068536 started
I20231123 11:59:40.414875     1 worker.cc:515] [worker] Thread #281473653892408 started
[...]

Issue #1851, opened 2023-10-23
Wrong bind 127.0.0.1 for docker image


rnz:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

Bind to 127.0.0.1 in the configuration is not suitable for the docker image

### Minimal reproduce step

```
$ docker run -d -p 6666:6666 apache/kvrocks:latest
67a4d65e172a67ce9d3c32a159fbdd1c9fd137294f9ca91a97a5728965558631

$ docker ps
CONTAINER ID  IMAGE                            COMMAND               CREATED         STATUS         PORTS                     NAMES
67a4d65e172a  docker.io/apache/kvrocks:latest                        4 minutes ago   Up 4 minutes   0.0.0.0:6666->6666/tcp  focused_grothendieck

$ docker exec -it 67 sh 
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
/ # netstat -tlnp
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 127.0.0.1:6666          0.0.0.0:*               LISTEN      1/kvrocks
tcp        0      0 127.0.0.1:6666          0.0.0.0:*               LISTEN      1/kvrocks
tcp        0      0 127.0.0.1:6666          0.0.0.0:*               LISTEN      1/kvrocks
tcp        0      0 127.0.0.1:6666          0.0.0.0:*               LISTEN      1/kvrocks
tcp        0      0 127.0.0.1:6666          0.0.0.0:*               LISTEN      1/kvrocks
tcp        0      0 127.0.0.1:6666          0.0.0.0:*               LISTEN      1/kvrocks
tcp        0      0 127.0.0.1:6666          0.0.0.0:*               LISTEN      1/kvrocks
tcp        0      0 127.0.0.1:6666          0.0.0.0:*               LISTEN      1/kvrocks

/ # exit
```
```
$ redis-cli -p 6666 info | wc -l
Error: Connection reset by peer
```

### What did you expect to see?

Connecting by redis-cli to kvrocks instance which was run in docker, such for example as it make with redis images:

```
$ docker ps
CONTAINER ID  IMAGE                            COMMAND               CREATED         STATUS         PORTS                     NAMES
[...]

PR #1906, opened 2023-11-23
Add the bind argument to the docker quickstart section

git-hulk:
The Docker image forgot to change the bind address to 0.0.0.0 at release 2.6.0,
and users will be frustrated if they cannot connect to the server after starting the server.

So it'd be better to fix the quickstart guide document.

refer: #1905 #1851

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ $ ./build/kvrocks -c kvrocks.conf
 ### Running Kvrocks using Docker
 
 ```shell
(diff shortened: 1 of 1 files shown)


PragmaTwice (approved):
Maybe we can also add this workaround in the website.

git-hulk:
> Maybe we can also add this workaround in the website.

Good point, will do it soon.

PR #1916, opened 2023-12-03
Fix name of rocksdb's write delay and stop stats in INFO command

### LINKED ISSUE #1913: output for 'level0_file_limit_stop' from ino is empty
### Search before asking

### Version

from 2.4 to 2.6

### Minimal reproduce step

info command

level0_file_limit_slowdown[default]:0
level0_file_limit_stop[default]:0
pending_compaction_bytes_slowdown[default]:0
pending_compaction_bytes_stop[default]:0
memtable_count_limit_slowdown[default]:0
memtable_count_limit_stop[default]:0

### What did you see instead?

level0_file_limit_slowdown[default]:
level0_file_limit_stop[default]:
pending_compaction_bytes_slowdown[default]:
pending_compaction_bytes_stop[default]:
memtable_count_limit_slowdown[default]:
memtable_count_limit_stop[default]:

### Anything Else?

_No response_

git-hulk:
This closes #1913 

rocksdb rename the write stall stats of the DB and column family in rocksdb #11300,
which organizes them more structurally. So it will return an empty string in the INFO command
since we're still using the old stat key to fetch them.

The INFO command will return "0" instead of the empty string after this patch:

```
level0_file_limit_slowdown[metadata]:0
level0_file_limit_stop[metadata]:0
pending_compaction_bytes_slowdown[metadata]:0
pending_compaction_bytes_stop[metadata]:0
level0_file_limit_stop_with_ongoing_compaction[metadata]:0
level0_file_limit_slowdown_with_ongoing_compaction[metadata]:0
memtable_count_limit_slowdown[metadata]:0
memtable_count_limit_stop[metadata]:0
```

and the current output is:

```
level0_file_limit_slowdown[default]:
level0_file_limit_stop[default]:
pending_compaction_bytes_slowdown[default]:
pending_compaction_bytes_stop[default]:
memtable_count_limit_slowdown[default]:
memtable_count_limit_stop[default]:
```

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -784,17 +784,21 @@ void Server::GetRocksDBInfo(std::string *info) {
                   << "\r\n";
     db->GetMapProperty(cf_handle, rocksdb::DB::Properties::kCFStats, &cf_stats_map);
     string_stream << "level0_file_limit_slowdown[" << cf_handle->GetName()
-                  << "]:" << cf_stats_map["io_stalls.level0_slowdown"] << "\r\n";
+                  << "]:" << cf_stats_map["l0-file-count-limit-delays"] << "\r\n";
     string_stream << "level0_file_limit_stop[" << cf_handle->GetName()
-                  << "]:" << cf_stats_map["io_stalls.level0_numfiles"] << "\r\n";
+                  << "]:" << cf_stats_map["l0-file-count-limit-stops"] << "\r\n";
     string_stream << "pending_compaction_bytes_slowdown[" << cf_handle->GetName()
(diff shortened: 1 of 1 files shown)


PragmaTwice (approved):

torwig (approved):

enjoy-binbin:
How did we lose this change when upgrading rocksdb? It seems like a good opportunity to think about how to handle the upgrade of dependency more safely.

git-hulk:
> How did we lose this change when upgrading rocksdb? It seems like a good opportunity to think about how to handle the upgrade of dependency more safely.

Yes, we now don't have enough cases to confirm whether it's good to upgrade.

PR #1918, opened 2023-12-04
Fix JSON.DEL should return 0 if the key was not found

git-hulk:


src/types/redis_json.cc
```diff
@@ -401,18 +401,23 @@ rocksdb::Status Json::ArrTrim(const std::string &user_key, const std::string &pa
...
-  if (!s.ok()) return s;
+  if (!s.ok() && !s.IsNotFound()) return s;
+  if (s.IsNotFound()) {
+    return rocksdb::Status::OK();
+  }
+
+  if (path == "$") {
+    *result = 1;
+    return del(ns_key);
+  }
```
PragmaTwice, line 420: I think it will lead to more i/o operations to move the code here rather than before `read`.

PragmaTwice: Ahhh I got why you move it here.

PragmaTwice (approved):

PR #1920, opened 2023-12-05
Minor refactor the return value of JSON methods

git-hulk:
Currently, many places return an optional array but are implemented in different ways, and the optional array to Redis response was duplicated in many times.


torwig (approved):
LGTM

PR #1928, opened 2023-12-08
Fix redundant ERR prefix in cluster redirect error message

### LINKED ISSUE #1922: Got (error) ERR MOVED when use redis-cli -c
### Version

Redis-cli version: 6.2.5
Kvrocks version: 2.6.0

### Minimal reproduce step

Run follow command:

```
redis-cli -h 172.16.49.83 -p 7383 -c

set test abc
```

ok

### What did you see instead?

(error) ERR MOVED 6918 172.16.49.96:7384

git-hulk:
Currently, Kvrocks cluster mode returns a MOVED/TRYAGAIN error message with the 'ERR' prefix, which will cause the Redis client can't recognize the redirection message. To make it compatible with the Redis client, we should remove this prefix.

Before this patch, Kvrocks will return a MOVED error message:

```
> redis-cli -p 30001 -c set a 1
(error) ERR MOVED 15495 127.0.0.1:30005
```

After applying this patch, it works well with redis-cli redirection:

```
> redis-cli -p 30001 -c set a 1
OK
```

This close #1922

### DIFF
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -393,7 +393,7 @@ void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
       s = srv_->cluster->CanExecByMySelf(attributes, cmd_tokens, this);
       if (!s.IsOK()) {
         if (is_multi_exec) multi_error_ = true;
-        Reply(redis::Error("ERR " + s.Msg()));
+        Reply(redis::Error(s.Msg()));
         continue;
       }
     }
--- a/tests/gocase/integration/cluster/cluster_test.go
+++ b/tests/gocase/integration/cluster/cluster_test.go
@@ -237,7 +237,7 @@ func TestClusterSlotSet(t *testing.T) {
 
 	slotKey := util.SlotTable[0]
 	require.NoError(t, rdb1.Set(ctx, slotKey, 0, 0).Err())
-	util.ErrorRegexp(t, rdb2.Set(ctx, slotKey, 0, 0).Err(), fmt.Sprintf(".*MOVED 0.*%d.*", srv1.Port()))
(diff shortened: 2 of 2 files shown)


enjoy-binbin (dismissed):
LGTM

enjoy-binbin (approved):

src/server/redis_connection.cc
```diff
@@ -393,7 +393,7 @@ void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
       s = srv_->cluster->CanExecByMySelf(attributes, cmd_tokens, this);
       if (!s.IsOK()) {
         if (is_multi_exec) multi_error_ = true;
-        Reply(redis::Error("ERR " + s.Msg()));
+        Reply(redis::Error(s.Msg()));
```
PragmaTwice, line 396: IMHO it is a huge change. How about just change the redirect command.

PragmaTwice: It can affect many many error messages. And some of these may be not expected to be changed.

git-hulk: @PragmaTwice It only affects the cluster redirect error message which returns by `CanExecByMySelf`.

enjoy-binbin: I did checked the error returns in CanExecByMySelf, they both no need the ERR code

PragmaTwice: Hmmm. Then no problem to me.

But for the general situation (we cannot control it for nested returned status right? it will cause huge maintanance effort), I recommend to add some custom error status and check it here.

I mean, we cannot check every commit every time whether every status need the ERR code, which is not a neat method to me.

PragmaTwice (approved):

src/server/redis_connection.cc
git-hulk: Yes, @enjoy-binbin also mentioned this before. It's unclear when the "ERR" prefix is needed.

Issue #1929, opened 2023-12-09
Add version information for each command to the Supported Commands page in website


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

https://kvrocks.apache.org/docs/supported-commands

We'd better to add which version each command started to be supported, e.g.

|Command|Since Version|
|-|-|
|FCALL|2.7.0|

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

tisonkun:
This may not be a good first issue but actually the contributor should know detail about the command impl history :D

jihuayu:
Contributors can write a string matching program to check if each command exists in a specific version of the source code. 
So I think the contributor don't need to know detail about the command impl history

kay011:
Hello, can you assign this issue to me?

git-hulk:
@kay011 Thank you! As @jihuayu mentioned, maybe a quick way is to write a program to detect the support commands in each version.

kay011:
@git-hulk Yeah！I plan to check the commands supported by each version through Python regular expressions. I will submit this script file and the final markdown file. This should be quick. Just paste the final results on the final docs.

PR #1935, opened 2023-12-16
Add SET command KEEPTLE and GET options supported

### LINKED ISSUE #1925: Set command supports NX and GET options used together
### Search before asking

### Motivation

New features in Redis 7.0.0.

## History
- Starting with Redis version 2.6.12: Added the EX, PX, NX and XX options.
- Starting with Redis version 6.0.0: Added the KEEPTTL option.
- Starting with Redis version 6.2.0: Added the GET, EXAT and PXAT option.
- Starting with Redis version 7.0.0: Allowed the NX and GET options to be used together.

redis docs: https://redis.io/commands/set/

### Solution

_No response_

jihuayu:
closed #1925 

In this PR, I have added support for `KEEPTLE` and `GET` options in the `SET` command.  
I have also updated the `GETSET`, `SETEX`, `SETNX`, and `SETXX` commands to use the new `set` function.

By the way,  I have a question about when to use `std::optional<std::string>&` and when to use `std::string*`? 
I think that both options are equivalent, but using `std::optional<std::string>&` can help avoid null in the code.
So I have used `std::optional<std::string>&` in my code.

### DIFF
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -244,7 +244,7 @@ rocksdb::Status String::Set(const std::string &user_key, const std::string &valu
     return rocksdb::Status::OK();
   } else {
     // if GET option not given, make ret not nil
-    if (!get) ret = std::make_optional("");
+    if (!get) ret = "";
   }
 
   // Handle expire time
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -23,7 +23,6 @@
 #include <cmath>
 #include <cstddef>
 #include <cstdint>
-#include <limits>
 #include <optional>
 #include <string>
 
@@ -188,19 +187,10 @@ rocksdb::Status String::GetEx(const std::string &user_key, std::string *value, u
 }
 
 rocksdb::Status String::GetSet(const std::string &user_key, const std::string &new_value, std::string *old_value) {
-  std::string ns_key = AppendNamespacePrefix(user_key);
-
-  LockGuard guard(storage_->GetLockManager(), ns_key);
-  rocksdb::Status s = getValue(ns_key, old_value);
-  if (!s.ok() && !s.IsNotFound()) return s;
-
-  std::string raw_value;
-  Metadata metadata(kRedisString, false);
-  metadata.Encode(&raw_value);
-  raw_value.append(new_value);
-  auto write_status = updateRawValue(ns_key, raw_value);
-  // prev status was used to tell whether old value was empty or not
-  return !write_status.ok() ? write_status : s;
+  std::optional<std::string> ret;
+  auto s = Set(user_key, new_value, 0, StringSetType::NX, /*get*/ true, /*keep_ttl*/ false, ret);
+  *old_value = *ret;
+  return s;
 }
 rocksdb::Status String::GetDel(const std::string &user_key, std::string *value) {
   std::string ns_key = AppendNamespacePrefix(user_key);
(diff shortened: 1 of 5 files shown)


src/types/redis_string.cc
```diff
@@ -217,38 +206,87 @@ rocksdb::Status String::Set(const std::string &user_key, const std::string &valu
   return MSet(pairs, /*ttl=*/0, /*lock=*/true);
 }
 
-rocksdb::Status String::SetEX(const std::string &user_key, const std::string &value, uint64_t ttl) {
-  std::vector<StringPair> pairs{StringPair{user_key, value}};
-  return MSet(pairs, /*ttl=*/ttl, /*lock=*/true);
-}
+rocksdb::Status String::Set(const std::string &user_key, const std::string &value, uint64_t ttl, StringSetType type,
```
git-hulk: Is it good to group Set Flags / keepttl/get into one parameter? To avoid too many input parameters.

git-hulk (commented):
Overall is good, thanks for your contribution.

src/types/redis_string.cc
```diff
@@ -217,38 +206,87 @@ rocksdb::Status String::Set(const std::string &user_key, const std::string &valu
...
+    // if GET option not given, operation aborted: return nil
+    if (!get) ret = std::nullopt;
+    return rocksdb::Status::OK();
+  } else if (!old_key_found && type == StringSetType::XX) {
+    // if GET option not given, operation aborted: return nil
+    if (!get) ret = std::nullopt;
+    return rocksdb::Status::OK();
+  } else {
+    // if GET option not given, make ret not nil
+    if (!get) ret = std::make_optional("");
```
```suggestion
    if (!get) ret = "";
```

src/types/redis_string.cc
jihuayu: OK, I will add a struct likes go-redis `SetArgs`.
```go
// SetArgs provides arguments for the SetArgs function.
type SetArgs struct {
	// Mode can be `NX` or `XX` or empty.
	Mode string

	// Zero `TTL` or `Expiration` means that the key has no expiration time.
	TTL      time.Duration
	ExpireAt time.Time

	// When Get is true, the command returns the old value stored at key, or nil when key did not exist.
	Get bool

	// KeepTTL is a Redis KEEPTTL option to keep existing TTL, it requires your redis-server version >= 6.0,
	// otherwise you will receive an error: (error) ERR syntax error.
	KeepTTL bool
}
```

git-hulk (dismissed):

src/types/redis_string.cc
```diff
@@ -187,20 +186,10 @@ rocksdb::Status String::GetEx(const std::string &user_key, std::string *value, u
...
-  std::string raw_value;
-  Metadata metadata(kRedisString, false);
-  metadata.Encode(&raw_value);
-  raw_value.append(new_value);
-  auto write_status = updateRawValue(ns_key, raw_value);
-  // prev status was used to tell whether old value was empty or not
-  return !write_status.ok() ? write_status : s;
+rocksdb::Status String::GetSet(const std::string &user_key, const std::string &new_value,
+                               std::optional<std::string> &old_value) {
+  auto s = Set(user_key, new_value, {/*ttl=*/0, StringSetType::NONE, /*get*/ true, /*keep_ttl*/ false}, old_value);
```
enjoy-binbin:
```suggestion
  auto s = Set(user_key, new_value, {/*ttl=*/0, StringSetType::NONE, /*get=*/true, /*keep_ttl=*/false}, old_value);
```

enjoy-binbin (dismissed):

git-hulk (approved):

enjoy-binbin (approved):

jihuayu:
@git-hulk  Maybe we need to add a note to the documentation, telling users that the next version（2.8）supports the SET option with KEEPTLE and GET options?

git-hulk:
@jihuayu Sure, thank you!

enjoy-binbin:
It feels like it could also be included in #1929 and done together

PR #1940, opened 2023-12-18
Add filesystem related lib

### LINKED ISSUE #1939: Can't build unittest with gcc8.5 and clang 12
### Search before asking

### Version

centos8/gcc8.5/clang12
kvrocks: unstable/2ecf764d

### Minimal reproduce step

./x.py build -j 20 --unittest
```
Consolidate compiler generated dependencies of target unittest
[ 94%] Linking CXX executable unittest
CMakeFiles/unittest.dir/tests/cppunit/compact_test.cc.o: In function `Compact_Filter_Test::TestBody()':
/usr/include/c++/8/bits/fs_path.h:185: undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
CMakeFiles/unittest.dir/tests/cppunit/compact_test.cc.o: In function `Compact_Filter_Test::TestBody()':
/home/yanzhao.tang/Code/kvrocks/tests/cppunit/compact_test.cc:128: undefined reference to `std::filesystem::remove_all(std::filesystem::__cxx11::path const&, std::error_code&)'
CMakeFiles/unittest.dir/tests/cppunit/disk_test.cc.o: In function `TestBase::~TestBase()':
/usr/include/c++/8/bits/fs_path.h:185: undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
CMakeFiles/unittest.dir/tests/cppunit/disk_test.cc.o: In function `TestBase::~TestBase()':
```
./x.py build -j 20 --unittest --compiler clang
```
[100%] Linking CXX executable unittest
ld.lld: error: undefined symbol: std::filesystem::__cxx11::path::_M_split_cmpts()
>>> referenced by fs_path.h:185 (/usr/lib/gcc/x86_64-redhat-linux/8/../../../../include/c++/8/bits/fs_path.h:185)
>>>               CMakeFiles/unittest.dir/tests/cppunit/compact_test.cc.o:(Compact_Filter_Test::TestBody())
>>> referenced by fs_path.h:185 (/usr/lib/gcc/x86_64-redhat-linux/8/../../../../include/c++/8/bits/fs_path.h:185)
>>>               CMakeFiles/unittest.dir/tests/cppunit/disk_test.cc.o:(TestBase::~TestBase())
>>> referenced by fs_path.h:185 (/usr/lib/gcc/x86_64-redhat-linux/8/../../../../include/c++/8/bits/fs_path.h:185)
>>>               CMakeFiles/unittest.dir/tests/cppunit/rdb_test.cc.o:(std::filesystem::__cxx11::path::path<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::filesystem::__cxx11::path>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::filesystem::__cxx11::path::format))
```

code compiled

### What did you see instead?

code not compiled

### Anything Else?

_No response_

Smityz:
close: https://github.com/apache/kvrocks/issues/1939


CMakeLists.txt
```diff
@@ -169,6 +169,7 @@ list(APPEND EXTERNAL_LIBS Threads::Threads)
 list(APPEND EXTERNAL_LIBS ${Backtrace_LIBRARY})
 list(APPEND EXTERNAL_LIBS xxhash)
 list(APPEND EXTERNAL_LIBS span-lite)
+list(APPEND EXTERNAL_LIBS stdc++fs)
```
mapleFU: Is this test only, or we may using it outside testing?

Smityz: done

PragmaTwice (commented):
I am fine to link `stdc++fs` into `kvrocks_objs` rather than just tests, but here are some points need to be considered:
- it seems in higher-version (>= 9) gcc (or, libstdc++) we don't need to explicitly link this library any more, so I wonder if it (`stdc++fs`) will be duplicated. I think we'd better to add a gcc version check here. BTW I encourage everyone to use new version of compiler as much as possible, e.g. gcc 12 or 13
- what will happen if we use libc++ rather than libstdc++? Does libc++ also provide `stdc++fs`?

PragmaTwice:
CI failed on macOS, which is related to the case about libc++ rather than libstdc++ that I mentioned before. We need to add a check about the std lib impl that are currently using.

Smityz:
~~It is somewhat difficult to determine the versions of gcc/clang/libstdc++/libc++. We can first fix the case for gcc8, and then set a minimum version limit of 13 for Clang @PragmaTwice~~

Passed CI tests, PTAL @PragmaTwice @mapleFU

CMakeLists.txt
```diff
@@ -148,7 +148,6 @@ endif()
 find_package(Threads REQUIRED)
 
 list(APPEND EXTERNAL_LIBS glog)
-list(APPEND EXTERNAL_LIBS snappy)
```
PragmaTwice, line 153: Hmmmm. Why is it deleted?

Smityz:
```
ld: warning: ignoring duplicate libraries: '_deps/snappy-build/libsnappy.a', 'clang_17.0_cxx17_64_relwithdebinfo/libtbb.a'
```
I found a warning when compiling with clang 17. It appears that deleting this line does not affect the compilation process.

tisonkun: I'd prefer to explicit list this dep instead of depend on transitive dep.

Smityz: fixed

PragmaTwice:
Close as stale. Feel free to reopen for any new activities.

PR #1943, opened 2023-12-18
Add the support of ANALYZE command to inspect the performance of RocksDB

### LINKED ISSUE #783: Expose some stats in RocksDB Perf Context and IO Stats Context
### Search before asking

### Motivation

Stats in Perf Context or IO Stat Context are useful to pinpoint some performance issues of redis commands.
For example, Counters `internal_delete_skipped_count` could illustrate why `zrangebyscore`  is so slow if there is too many tombstones.

### Solution

_No response_

chrisxu333:
This pr tries to close #783 .

To expose essential perf_context and iostats_context from rocksdb, a new command ANALYZE is introduced.

A typical usage would be as follows:
perf

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -18,6 +18,9 @@
  *
  */
 
+#include <rocksdb/iostats_context.h>
+#include <rocksdb/perf_context.h>
+
 #include "command_parser.h"
 #include "commander.h"
 #include "commands/scan_base.h"
@@ -1123,6 +1126,71 @@ class CommandRdb : public Commander {
   uint32_t db_index_ = 0;
 };
 
+class CommandAnalyze : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() <= 1) return {Status::RedisExecErr, errInvalidSyntax};
+    for (int i = 1; i < args.size(); ++i) {
+      command_args_.push_back(args[i]);
+    }
+    return Status::OK();
+  }
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    auto commands = redis::CommandTable::Get();
+    auto cmd_iter = commands->find(util::ToLower(command_args_[0]));
+    if (cmd_iter == commands->end()) {
+      // unsupported redis command
+      return {Status::RedisExecErr, errInvalidSyntax};
+    }
+    auto redis_cmd = cmd_iter->second;
+    auto cmd = redis_cmd->factory();
+    cmd->SetAttributes(redis_cmd);
+    cmd->SetArgs(command_args_);
+
+    int arity = cmd->GetAttributes()->arity;
+    if ((arity > 0 && command_args_.size() != arity) || (arity < 0 && command_args_.size() < -arity)) {
+      *output = redis::Error("ERR wrong number of arguments");
+      return {Status::RedisExecErr, errWrongNumOfArguments};
+    }
+
+    auto s = cmd->Parse(command_args_);
+    if (!s.IsOK()) {
+      return s;
+    }
+
+    auto prev_perf_level = rocksdb::GetPerfLevel();
+    rocksdb::SetPerfLevel(rocksdb::PerfLevel::kEnableTimeExceptForMutex);
+    rocksdb::get_perf_context()->Reset();
+    rocksdb::get_iostats_context()->Reset();
+
+    std::string command_output;
+    s = cmd->Execute(srv, conn, &command_output);
+    if (!s.IsOK()) {
+      return s;
+    }
+
+    if (command_output[0] == '-') {
+      *output = command_output;
+      return s;
+    }
+
+    std::string perf_context = rocksdb::get_perf_context()->ToString(true);
+    std::string iostats_context = rocksdb::get_iostats_context()->ToString(true);
+    rocksdb::get_perf_context()->Reset();
+    rocksdb::get_iostats_context()->Reset();
+    rocksdb::SetPerfLevel(prev_perf_level);
+
+    *output = redis::MultiLen(3);  // command output + perf context + iostats context
+    *output += command_output;
+    *output += redis::BulkString(perf_context);
+    *output += redis::BulkString(iostats_context);
+    return Status::OK();
+  }
+
+ private:
+  std::vector<std::string> command_args_;
+};
+
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandAuth>("auth", 2, "read-only ok-loading", 0, 0, 0),
                         MakeCmdAttr<CommandPing>("ping", -1, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandSelect>("select", 2, "read-only", 0, 0, 0),
@@ -1157,6 +1225,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandAuth>("auth", 2, "read-only ok-loadin
                         MakeCmdAttr<CommandFlushBackup>("flushbackup", 1, "read-only no-script", 0, 0, 0),
                         MakeCmdAttr<CommandSlaveOf>("slaveof", 3, "read-only exclusive no-script", 0, 0, 0),
                         MakeCmdAttr<CommandStats>("stats", 1, "read-only", 0, 0, 0),
-                        MakeCmdAttr<CommandRdb>("rdb", -3, "write exclusive", 0, 0, 0), )
+                        MakeCmdAttr<CommandRdb>("rdb", -3, "write exclusive", 0, 0, 0),
+                        MakeCmdAttr<CommandAnalyze>("analyze", -1, "", 0, 0, 0), )
 
 }  // namespace redis


src/commands/cmd_server.cc
```diff
@@ -1123,6 +1126,60 @@ class CommandRdb : public Commander {
...
+
+    tmp_output = redis::ExtractData(tmp_output);
+
+    *output = redis::MultiBulkString(
+        {tmp_output, "(perf_context): " + perf_context, "(iostats_context): " + iostats_context});
+    return Status::OK();
+  }
+
+ private:
+  std::vector<std::string> true_args;
```
mapleFU:
```suggestion
  std::vector<std::string> true_args_;
```

src/commands/cmd_server.cc
```diff
@@ -1123,6 +1126,60 @@ class CommandRdb : public Commander {
...
+    if (cmd_iter == commands->end()) {
+      // unsupported redis command
+      return {Status::RedisExecErr, errInvalidSyntax};
+    }
+    auto redis_cmd = cmd_iter->second;
+    auto cmd = redis_cmd->factory();
+    cmd->SetAttributes(redis_cmd);
+    cmd->SetArgs(true_args);
+
+    auto s = cmd->Parse(true_args);
```
torwig: Should we check the status here?

src/commands/cmd_server.cc
```diff
@@ -1123,6 +1126,60 @@ class CommandRdb : public Commander {
...
+
+    rocksdb::SetPerfLevel(rocksdb::PerfLevel::kEnableTimeExceptForMutex);
+    rocksdb::get_perf_context()->Reset();
+    rocksdb::get_iostats_context()->Reset();
+
+    std::string tmp_output;
+    s = cmd->Execute(srv, conn, &tmp_output);
```
torwig: If we have hundreds/thousands of operations per second, how do we know that the perf context and its data are related for this particular command?

src/commands/cmd_server.cc
```diff
@@ -1157,6 +1214,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandAuth>("auth", 2, "read-only ok-loadin
                         MakeCmdAttr<CommandFlushBackup>("flushbackup", 1, "read-only no-script", 0, 0, 0),
                         MakeCmdAttr<CommandSlaveOf>("slaveof", 3, "read-only exclusive no-script", 0, 0, 0),
                         MakeCmdAttr<CommandStats>("stats", 1, "read-only", 0, 0, 0),
-                        MakeCmdAttr<CommandRdb>("rdb", -3, "write exclusive", 0, 0, 0), )
+                        MakeCmdAttr<CommandRdb>("rdb", -3, "write exclusive", 0, 0, 0),
+                        MakeCmdAttr<CommandAnalyze>("analyze", -1, "", 0, 0, 0), )
```
mapleFU, line 1229: at least, it should be an exclusive call?

src/commands/cmd_server.cc
```diff
@@ -1123,6 +1126,60 @@ class CommandRdb : public Commander {
...
+    }
+
+    if (tmp_output[0] == '-') {
+      *output = tmp_output;
+      return s;
+    }
+
+    std::string perf_context = rocksdb::get_perf_context()->ToString(true);
+    std::string iostats_context = rocksdb::get_iostats_context()->ToString(true);
+    rocksdb::SetPerfLevel(rocksdb::PerfLevel::kDisable);
```
mapleFU: at least we should record pervious perf level, and resume it after this function( even if the status is error after execution )

mapleFU (commented):
This is a bit confusing because it does far less than `PROFILE` in SQL. Also it might break the previous config about perflevel, so I wonder whether we have better way for this...

src/commands/cmd_server.cc
chrisxu333: That's a good point I'll fix it accordingly :)

mapleFU:
Maybe it's better to first having a RAII guard like TiKV: https://github.com/tikv/tikv/blob/master/src/coprocessor/tracker.rs ( We can be easier than them) ?

src/commands/cmd_server.cc
mapleFU: Sorry maybe I made it wrong, since PerfContext is a tls in implemention, maybe current implemention doesn't need to considering exclusive

chrisxu333:
> Maybe it's better to first having a RAII guard like TiKV: https://github.com/tikv/tikv/blob/master/src/coprocessor/tracker.rs ( We can be easier than them) ?

Can you elaborate more on this? I didn't quite get the point.

src/commands/cmd_server.cc
```diff
@@ -1123,6 +1126,60 @@ class CommandRdb : public Commander {
...
+    if (perf_context.empty()) return Status::OK();  // request without db operation
+
+    tmp_output = redis::ExtractData(tmp_output);
+
+    *output = redis::MultiBulkString(
+        {tmp_output, "(perf_context): " + perf_context, "(iostats_context): " + iostats_context});
```
mapleFU:
```suggestion
        {std::move(tmp_output), "(perf_context): " + perf_context, "(iostats_context): " + iostats_context});
```

src/server/redis_reply.cc
```diff
@@ -34,6 +34,29 @@ std::string BulkString(const std::string &data) { return "$" + std::to_string(da
 
 std::string NilString() { return "$-1" CRLF; }
 
+std::string ExtractData(const std::string &reply) {
```
mapleFU: Emmm, what about the multi bulk string like?

mapleFU:
> > Maybe it's better to first having a RAII guard like TiKV: https://github.com/tikv/tikv/blob/master/src/coprocessor/tracker.rs ( We can be easier than them) ?
> 
> Can you elaborate more on this? I didn't quite get the point.

Hmm we can just leave it here in this patch. I believe something like a :

```
class StoragePerfContext {
  
  void close();
  ~StoragePerfContext() {
      if (!reset_) {
         rocksdb::Perfcontext // resume
      }
   }
private:
  Storage* // or rocksdb?
  bool reseted_ = false;
};
```

Might be better? If other command want to debugging, they can also uses this tool? Also we can considering RAII here.

chrisxu333:
> > > Maybe it's better to first having a RAII guard like TiKV: https://github.com/tikv/tikv/blob/master/src/coprocessor/tracker.rs ( We can be easier than them) ?
> > 
> > 
> > Can you elaborate more on this? I didn't quite get the point.
> 
> Hmm we can just leave it here in this patch. I believe something like a :
> 
> ```
> class StoragePerfContext {
>   
>   void close();
>   ~StoragePerfContext() {
>       if (!reset_) {
>          rocksdb::Perfcontext // resume
>       }
>    }
> private:
>   Storage* // or rocksdb?
>   bool reseted_ = false;
> };
> ```
> 
> Might be better? If other command want to debugging, they can also uses this tool? Also we can considering RAII here.

I see what you meant. So just to be clear, do we still need to do this in this patch? Or we wait until when it's actually needed to do so?

src/commands/cmd_server.cc
mapleFU: This might related to context, it uses ThreadLocal storage, and metric in the thread local context

mapleFU:
> do we still need to do this in this patch

I agree it's not neccessary in this patch. Personally I think metrics can be widely used, but lets first get it in used...

cc @PragmaTwice @git-hulk This command looks like SQL "PROFILE" with rocksdb io/perf stats enabled and output. It can be used in some cases, but for some blocking command I don't know whether it's ok. Would you mind take a look?

chrisxu333:
> > do we still need to do this in this patch
> 
> I agree it's not neccessary in this patch. Personally I think metrics can be widely used, but lets first get it in used...
> 
> cc @PragmaTwice @git-hulk This command looks like SQL "PROFILE" with rocksdb io/perf stats enabled and output. It can be used in some cases, but for some blocking command I don't know whether it's ok. Would you mind take a look?

@git-hulk @PragmaTwice Hello guys, if you have time could you help review this pr? Thank you :)

git-hulk:
Sure, sorry for missing this message. I will take a look recently.

src/commands/cmd_server.cc
```diff
@@ -1123,6 +1126,75 @@ class CommandRdb : public Commander {
...
+    if (tmp_output.at(0) == '*') {
+      // if multi bulkstring, append directly
+      tmp_output.append(redis::BulkString(perf_context));
+      tmp_output.append(redis::BulkString(iostats_context));
+      *output = tmp_output;
+    } else {
+      tmp_output = redis::ExtractData(tmp_output);
+      *output = redis::MultiBulkString(
+          {std::move(tmp_output), "(perf_context): " + perf_context, "(iostats_context): " + iostats_context});
+    }
```
```suggestion
    *output = redis::MultiLen(3); // command output + perf context + iostats context
    *output += std::move(tmp_output);
    *output += redis::BulkString(perf_context);
    *output += redis::BulkString(iostats_context);
```

This implementation isn't correct. For the multi-bulk strings, the first element is the number of array elements, so you can't simply append the element.

git-hulk: Can also remove function `ExtractData` and rename `tmp_output` to `command_output`

chrisxu333: What do you mean by saying remove `ExtractData` function?

chrisxu333: Sounds good I've changed them accordingly :)

git-hulk:
@chrisxu333 Can rename `true_args_` to `command_args_`, rest are good to me. Sorry for missing this point.

git-hulk (dismissed):

mapleFU:
Would you mind fix the lint first?

torwig:
@chrisxu333  `./x.py format` should do the trick.

chrisxu333:
> Would you mind fix the lint first?

Sounds good I just did.

src/commands/cmd_server.cc
```diff
@@ -1123,6 +1126,65 @@ class CommandRdb : public Commander {
...
+    if (cmd_iter == commands->end()) {
+      // unsupported redis command
+      return {Status::RedisExecErr, errInvalidSyntax};
+    }
+    auto redis_cmd = cmd_iter->second;
+    auto cmd = redis_cmd->factory();
+    cmd->SetAttributes(redis_cmd);
+    cmd->SetArgs(command_args_);
+
+    auto s = cmd->Parse(command_args_);
```
mapleFU, line 1156: Should we check the arity before `Parse`? Since wrong `command_args_` might make our program crash

PragmaTwice: We need to reuse the code in ExecuteCommand rather than write them again here.

chrisxu333: I just pushed a newer version which added the arity checking logic. As for reusing / refactoring the code in `ExecuteCommand`, maybe it's better to do it in a separate pr? As the `ExecuteCommand` code needs to be dissected into several chunks for reusing purpose. @mapleFU @PragmaTwice

src/commands/cmd_server.cc
```diff
@@ -1123,6 +1126,71 @@ class CommandRdb : public Commander {
   uint32_t db_index_ = 0;
 };
 
+class CommandAnalyze : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() <= 1) return {Status::RedisExecErr, errInvalidSyntax};
+    for (int i = 1; i < args.size(); ++i) {
+      command_args_.push_back(args[i]);
+    }
```
mapleFU, line 1135: nit: we can directly `insert` rather than push back

mapleFU (approved):

chrisxu333:
> @chrisxu333 `./x.py format` should do the trick.

Hi @torwig, could you help review this pr whenever you're able to? Thx!

chrisxu333:
Hi @git-hulk @PragmaTwice,

Could you help review and approve this pr if possible? Thx!

git-hulk:
@chrisxu333 Thanks for your follow-up. This PR looks good to me, to see if @PragmaTwice and @torwig has any comments, if not, we can move forward to merge.

git-hulk (approved):

PragmaTwice (approved):

PR #1963, opened 2023-12-22
Refactor kvrocks2redis via rocksdb secondary instance

### LINKED ISSUE #1832: Refactor kvrocks2redis via rocksdb secondary instance
### Search before asking

### Motivation

Currently kvrocks2redis using network connections to catch up with incremental data. We can use [Read only and Secondary instances](https://github.com/facebook/rocksdb/wiki/Read-only-and-Secondary-instances) in RocksDB to remove such network connections.

cc @git-hulk 

### Solution

_No response_

maochongxin:
Using RSI (Secondary Instance) for synchronization, replacing the previous method of using `OpenForReadOnly` followed by PSync.
First, open the data directory using `OpenAsSecondary`, and reuse the previous method of parsing the checkpoint to complete a "full sync". Then, use `TryCatchUpWithPrimary` to synchronize the new wal.

Try to fix #1832


utils/kvrocks2redis/kvrocks2redis.conf
```diff
@@ -17,6 +17,11 @@ daemonize no
 # Default: ./data
 data-dir ./data
 
+# The Secondary Instance directory.
+#
+#  Default: ./secondary
+secondary-dir ./secondary
+
```
PragmaTwice: I'm wondering if we could just use `data-dir` here instead of introducing a `secondary-dir`?

maochongxin:
I'm still trying to add some integration tests

utils/kvrocks2redis/kvrocks2redis.conf
PragmaTwice: I mean, maybe we can make secondary instance the default way, so that we can remove all data-fetching network connections.

maochongxin: I may have not understood your meaning correctly, but are you referring to configuring the secondary dir similar to `db_dir = data_dir + "/db"`, without introducing a `secondary-dir`?

PragmaTwice: Currently we use `OpenForReadOnly` in kvrocks2redis, we can change it to `OpenAsSecondary` and then we will be able to use `TryCatchUpWithPrimary` to capture incremental data rather than network connections.

That's the intention of #1832.

When we change it to `OpenAsSecondary`, the original `OpenForReadOnly` mode is useless and can be removed.

tisonkun:
Thanks for your contribution @maochongxin!

The PR description is empty. Can you add some background so that reviewers can understand the motivation and how you plan to implement it?

utils/kvrocks2redis/kvrocks2redis.conf
maochongxin: oh, this pr was created for that issue. Then I should be able to remove the support for the readonly mode in the `storage::Open`.

maochongxin:
> Thanks for your contribution @maochongxin!
> 
> The PR description is empty. Can you add some background so that reviewers can understand the motivation and how you plan to implement it?

Thank you for the reminder. I have now provided the background information for this matter.

utils/kvrocks2redis/kvrocks2redis.conf
PragmaTwice: I think it is optional to clean up Storage::Open, but I suggest you to clean the code for original logic and useless network connections in kvrocks2redis.

maochongxin: Ok, I understand what you mean now

src/config/config.h
```diff
@@ -119,6 +119,7 @@ struct Config {
   std::vector<std::string> binds;
   std::string dir;
   std::string db_dir;
+  std::string secondary_dir;
```
PragmaTwice:
```suggestion
```

src/storage/storage.cc
```diff
@@ -243,7 +243,7 @@ Status Storage::CreateColumnFamilies(const rocksdb::Options &options) {
   return Status::OK();
 }
 
-Status Storage::Open(bool read_only) {
+Status Storage::Open(bool read_only, bool as_secondary) {
```
PragmaTwice: You can add an `enum OpenMode`, then we can have 3 state here rather than 2*2=4.

utils/kvrocks2redis/config.h
```diff
@@ -43,14 +43,12 @@ struct Config {
   std::string data_dir = "./data";
   std::string output_dir = "./";
   std::string db_dir = data_dir + "/db";
+  std::string secondary_dir = output_dir + "/secondary";
```
PragmaTwice:
```suggestion
```

utils/kvrocks2redis/config.cc
```diff
@@ -78,6 +78,7 @@ Status Config::parseConfigFromString(const std::string &input) {
     }
     pidfile = output_dir + "kvrocks2redis.pid";
     next_seq_file_path = output_dir + "last_next_seq.txt";
+    secondary_dir = output_dir + "secondary";
```
PragmaTwice:
```suggestion
```

src/storage/storage.cc
maochongxin: Sure, however, I also noticed an issue [here](https://github.com/apache/kvrocks/blob/unstable/src/storage/storage.cc#L329). The check for `s` doesn't make sense; it doesn't actually validate the return value of the DBOpen operation. I can fix this issue together.

src/storage/storage.cc
```diff
@@ -322,17 +322,32 @@ Status Storage::Open(bool read_only) {
...
+      db_ = GET_OR_RET(util::DBOpenForReadOnly(options, config_->db_dir, column_families, &cf_handles_));
+      break;
+    }
+    case DBOpenMode::kDBOpenModeAsSecondaryInstance: {
+      db_ = GET_OR_RET(
+          util::DBOpenAsSecondaryInstance(options, config_->db_dir, config_->dir, column_families, &cf_handles_));
+      break;
+    }
+    default:
+      break;
```
```suggestion
      __builtin_unreachable();
```

PragmaTwice (commented):
The rest looks good to me! It would be great if you can add some test for it.

utils/kvrocks2redis/sync.cc
```diff
@@ -104,98 +78,68 @@ void Sync::Stop() {
...
-    std::string line = GET_OR_RET(util::SockReadLine(sock_fd_).Prefixed("read auth response err"));
-    if (line.compare(0, 3, "+OK") != 0) {
-      return {Status::NotOK, "auth got invalid response"};
-    }
-  }
-  LOG(INFO) << "[kvrocks2redis] Auth succ, continue...";
-  return Status::OK();
+Status Sync::tryCatchUpWithPrimary() {
+  auto s = storage_->GetDB()->TryCatchUpWithPrimary();
+  return s.ok() ? Status() : Status::NotOK;
```
PragmaTwice, line 83:
```suggestion
  return storage_->GetDB()->TryCatchUpWithPrimary();
```

Why not just return it?

PragmaTwice: Ahh I see. It's rocksdb::Status.

maochongxin:
> The rest looks good to me! It would be great if you can add some test for it.

Added test for data consistency between the source and destination nodes.

PragmaTwice:
> > The rest looks good to me! It would be great if you can add some test for it.
> 
> Added test for data consistency between the source and destination nodes.

Great! Could you also add tests for the incremental data? e.g. kvrocks2redis can catch new data after starting if it is written to the source kvrocks node.

PragmaTwice (dismissed):

PragmaTwice:
We need to add apache license header to newly added files.

```
ERROR the following files don't have a valid license header: 
utils/kvrocks2redis/tests/check_consistency.py 
```

Refer to x.py for the same header.

maochongxin:
> We need to add apache license header to newly added files.
> 
> ```
> ERROR the following files don't have a valid license header: 
> utils/kvrocks2redis/tests/check_consistency.py 
> ```
> 
> Refer to x.py for the same header.

Thank you, updated.

PragmaTwice (approved):

git-hulk (approved):

PragmaTwice:

Issue #1937, opened 2023-12-17
Add `keyspace_hits` and `keyspace_misses` statistics that be consistent with Redis


git-hulk:

git-hulk:
It needs to do some minor refactoring before adding this metric since now many places will read metadata directly. I will do it while I'm free before the next release.

PR #1964, opened 2023-12-23
Unify the way of parsing metadata

git-hulk:
Currently, three places will parse the metadata and
it may introduce duplicate codes if we would like to add
some statistics when parsing the metadata.

This PR split into three commits:

1. Minor refactor the readMulti in redis_json.cc to avoid duplicate codes
2. Allow to pass multiple data types when parsing metadata, so that bitmap can remove duplicate parsing metadata code
3. Remove an unused function

relate issue: https://github.com/apache/kvrocks/issues/1937


tisonkun (approved):
Thank you

Issue #1948, opened 2023-12-18
Add support of the command PEXPIRETIME


PragmaTwice:
Refer to https://redis.io/commands/pexpiretime/

kay011:
I think this command can be supported together with #1947 , and assigned to me.

PragmaTwice:
Looks good to me! Assigned.

Issue #1947, opened 2023-12-18
Add support of the command EXPIRETIME


PragmaTwice:
Refer to https://redis.io/commands/expiretime/

kay011:
Can you assign it to me?

git-hulk:
@kay011 Assigned, thank you!

PR #1965, opened 2023-12-24
Add support of EXPIRETIME and PEXPIRETIME

kay011:
This commit finish the  #1948 and  #1947. 
It should be noted that since the timeout precision currently supported by kvrocks is seconds, the results obtained by executing PEXPIRETIME will be different from those obtained by redis. This is my first time trying to implement a kvrocks command, please give me your code review comments. Thanks!

### DIFF
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -67,18 +67,18 @@ rocksdb::Status Database::ParseMetadata(RedisTypes types, Slice *bytes, Metadata
   return s;
 }
 
-rocksdb::Status Database::GetMetadata(RedisType type, const Slice &ns_key, Metadata *metadata) {
+rocksdb::Status Database::GetMetadata(RedisTypes types, const Slice &ns_key, Metadata *metadata) {
   std::string raw_value;
   Slice rest;
-  return GetMetadata(type, ns_key, &raw_value, metadata, &rest);
+  return GetMetadata(types, ns_key, &raw_value, metadata, &rest);
 }
 
-rocksdb::Status Database::GetMetadata(RedisType type, const Slice &ns_key, std::string *raw_value, Metadata *metadata,
+rocksdb::Status Database::GetMetadata(RedisTypes types, const Slice &ns_key, std::string *raw_value, Metadata *metadata,
                                       Slice *rest) {
   auto s = GetRawMetadata(ns_key, raw_value);
   *rest = *raw_value;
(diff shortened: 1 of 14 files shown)


PragmaTwice:
Thanks for your effort!

> It should be noted that since the timeout precision currently supported by kvrocks is seconds, the results obtained by executing PEXPIRETIME will be different from those obtained by redis.

If you build kvrocks with `-DENABLE_NEW_ENCODING=ON`, you can see that the precision of these expire time is in milliseconds.

src/storage/redis_db.cc
```diff
@@ -226,6 +226,30 @@ rocksdb::Status Database::TTL(const Slice &user_key, int64_t *ttl) {
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status Database::GetExpireTime(const Slice &user_key, uint64_t *timestamp) {
```
git-hulk, line 228: We can use `GetMetadata` directly instead of decoding here.

PragmaTwice: IMHO it is hard to use since we need to pass RedisType.

git-hulk: Yes, maybe we need to pass an empty redis type array to ignore the type check later.

PragmaTwice: We can use `RedisTypes::All()` now.

git-hulk: @kay011 Would you mind changing this implementation to `GetMetadata` since we can use `RedisTypes::All()` now.

kay011: @git-hulk Of course, I'm thinking about how to use it more elegantly

git-hulk: How about changing the parameter `type` of `Database::GetMetadata` to `RedisTypes` like ParseMetadata?

kay011: @git-hulk I think that is the best solution! But the existed code will change the call parameter type. So there is a trade off.

git-hulk: I think that's fine since only a few places are using this function.

kay011: @git-hulk Nice, I will implement it as stated above.

git-hulk:
@kay011 Thanks for your contribution and quick response.

git-hulk (approved):

PR #1967, opened 2023-12-24
Optimize ParseMetadata by introducing RedisTypes

PragmaTwice:
Both construction of `std::vector<RedisType>` and `vector<T>::find` is not efficient enough.

Here we use bitmap to make it faster and avoid memory allocation.


git-hulk (dismissed):
LGTM, cool!

torwig (approved):

git-hulk (approved):

PR #1970, opened 2023-12-25
Fix `GET` return no error when key type is not string

### LINKED ISSUE #1969: Using `GET` to get json type key will not return error.
### Search before asking

### Version

unstable 

### Minimal reproduce step

Using `GET` to get json type key will not return error.
I writed below test case.

```go
t.Run("GET JSON type key", func(t *testing.T) {
	require.NoError(t, rdb.Do(ctx, "JSON.SET", "a", "$", ` {"x":1,"y":2} `).Err())
	require.EqualValues(t, "\x00{\"x\":1,\"y\":2}", rdb.Get(ctx, "a").Val())
})
```

test failed

### What did you see instead?

test Passed

### Anything Else?

Maybe is this line problem.https://github.com/apache/kvrocks/blob/unstable/src/types/redis_string.cc#L86
```C++
if (metadata.Type() != kRedisString && metadata.size > 0) {
    return rocksdb::Status::InvalidArgument(kErrMsgWrongType);
  }
```
Why do we need to check `metadata.size > 0`?

jihuayu:
fixed #1969

### DIFF
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -58,7 +58,7 @@ std::vector<rocksdb::Status> String::getRawValues(const std::vector<Slice> &keys
       statuses[i] = rocksdb::Status::NotFound(kErrMsgKeyExpired);
       continue;
     }
-    if (metadata.Type() != kRedisString && metadata.size > 0) {
+    if (metadata.Type() != kRedisString && (metadata.IsEmptyableType() || metadata.size > 0)) {
       (*raw_values)[i].clear();
       statuses[i] = rocksdb::Status::InvalidArgument(kErrMsgWrongType);
       continue;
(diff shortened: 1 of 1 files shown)


git-hulk:
@jihuayu Could you add a test case to make sure this issue was resolved?

jihuayu:
@git-hulk No problem, I will add test cases for each type.

git-hulk:
> @git-hulk No problem, I will add test cases for each type.

Thank you!

git-hulk:
@jihuayu This bug should be resolved after https://github.com/apache/kvrocks/pull/1971, but can add test cases to make sure it works well if you like.

jihuayu:
@git-hulk All right, I'm a bit busy these days, I will come back later to solve it.

jihuayu:
Many commands have this kind of constraint (cannot operate on keys of other types), and we may need a universal method to test them.
I will find a better way that covers this issue. 
Close this PR.

PR #1971, opened 2023-12-26
Record and export the keyspace hit/miss count to INFO command

### LINKED ISSUE #1937: Add `keyspace_hits` and `keyspace_misses` statistics that be consistent with Redis

git-hulk:
Currently, Redis supports inspecting the keyspace hit/miss count via
INFO command to see if the hit ratio is expected. So we also export
the same metric to align with Redis metrics.

This closes #1937

### DIFF
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -200,12 +200,6 @@ Status Storage::SetOptionForAllColumnFamilies(const std::string &key, const std:
   return Status::OK();
 }
 
-Status Storage::SetOption(const std::string &key, const std::string &value) {
-  auto s = db_->SetOptions({{key, value}});
-  if (!s.ok()) return {Status::NotOK, s.ToString()};
-  return Status::OK();
-}
-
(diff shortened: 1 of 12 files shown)


torwig (approved):

src/storage/storage.cc
```diff
@@ -641,6 +659,23 @@ Status Storage::ReplicaApplyWriteBatch(std::string &&raw_batch) {
...
+void Storage::RecordStat(StatType type, uint64_t v) {
+  switch (type) {
+    case StatType::FlushCount:
+      db_stats_.flush_count += v;
+      break;
+    case StatType::CompactionCount:
+      db_stats_.compaction_count += v;
+      break;
+    case StatType::KeyspaceHits:
+      db_stats_.keyspace_hits += v;
```
mapleFU, line 671: Would `fetch_add(relaxed)` better, which might not need to hold the execution pipeline?

git-hulk: @mapleFU Yes, it don't require strict sequence increase.

Issue #1966, opened 2023-12-24
The storage engine type can be awareness in Kvrocks


git-hulk:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

We're now supporting the speedb and rocksdb as its storage engine
and maybe even more in the future. I believe it'd would be better
to allow each engine to have its specific configurations.

### Solution

We can export the engine type information to Kvrocks while building
the source stage. Can also take a look at how the new encoding format
is recognized: https://github.com/apache/kvrocks/blob/unstable/CMakeLists.txt#L222-L226 

 

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

MaheshMadushan:
Hi, is this issue available? if so, shall I try on this?

git-hulk:
@MaheshMadushan assigned, thank you!

PR #1973, opened 2023-12-31
The storage engine type awareness in Kvrocks

MaheshMadushan:
https://github.com/apache/kvrocks/issues/1966

Storage Engine Type Enum
`enum class StorageEngineType : uint16_t {
  RocksDB,
  Speedb,
};`

Storage type stored in variable STORAGE_ENGINE_TYPE in `src/storage.h`. it can be `StorageEngineType::RocksDB` or `StorageEngineType::Speedb`


git-hulk (dismissed):
LGTM, thank you!

CMakeLists.txt
```diff
@@ -216,6 +216,11 @@ elseif((CMAKE_CXX_COMPILER_ID STREQUAL "Clang") OR (CMAKE_CXX_COMPILER_ID STREQU
 endif()
 target_link_libraries(kvrocks_objs PUBLIC -fno-omit-frame-pointer)
 target_link_libraries(kvrocks_objs PUBLIC ${EXTERNAL_LIBS})
+if (ENABLE_SPEEDB)
+    target_compile_definitions(kvrocks_objs PUBLIC SPEEDB_STORAGE_ENGINE)
```
PragmaTwice: Could we just use one macro definition with different value?

PragmaTwice (changes requested):
BTW, I think we can have an enum type in our code.

CMakeLists.txt
MaheshMadushan: Thank you for the feedback. I'm thinking of having numbered values for each storage engine type. Would that be ideal?
or we could have string names like `"ROCKSDB"` and `"SPEEDB"`.

- First case:

`STORAGE_ENGINE=0 # for rocksdb`
`STORAGE_ENGINE=1 # for speedb`

- Second case:

`STORAGE_ENGINE="ROCKSDB"`
`STORAGE_ENGINE="SPEEDB"`

mapleFU: I vote for the second one

MaheshMadushan:
> BTW, I think we can have an enum type in our code.

Under the namespace `engine` in `src/storage.h` following enum can be created.
`enum class Type {
  RocksDB,
  Speedb,
};
`

Then we can have macros for storage engine as following,
`STORAGE_ENGINE=0 # for rocksdb`
`STORAGE_ENGINE=1 # for speedb`

Then we can determine the engine type as following,
`if (STORAGE_ENGINE == static_cast<int>(engine::Type::RocksDB)) {} ...`

Would that be ideal?

PragmaTwice:
> > BTW, I think we can have an enum type in our code.
> 
> Under the namespace `engine` in `src/storage.h` following enum can be created.
> `enum class Type {
>   RocksDB,
>   Speedb,
> };
> `
> 
> Then we can have macros for storage engine as following,
> `STORAGE_ENGINE=0 # for rocksdb`
> `STORAGE_ENGINE=1 # for speedb`
> 
> Then we can determine the engine type as following,
> `if (STORAGE_ENGINE == static_cast<int>(engine::Type::RocksDB)) {} ...`
> 
> Would that be ideal?

Firstly, the enum type can definitely have a better name than `Type`.

Besides, we can have a static-storage-duration constant to wrap the macro, so that we don't need to do any type cast.

CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ option(ENABLE_IPO "enable interprocedural optimization" ON)
 option(ENABLE_UNWIND "enable libunwind in glog" ON)
 option(ENABLE_SPEEDB "enable speedb instead of rocksdb" OFF)
 set(PORTABLE 0 CACHE STRING "build a portable binary (disable arch-specific optimizations)")
+set(ROCKSDB_STORAGE_ENGINE 0 STRING "ROCKSDB")
+set(SPEEDB_STORAGE_ENGINE 1 STRING "SPEEDB")
```
PragmaTwice, line 35:
```suggestion
```

```suggestion
```

CMakeLists.txt
```diff
@@ -216,6 +218,11 @@ elseif((CMAKE_CXX_COMPILER_ID STREQUAL "Clang") OR (CMAKE_CXX_COMPILER_ID STREQU
 endif()
 target_link_libraries(kvrocks_objs PUBLIC -fno-omit-frame-pointer)
 target_link_libraries(kvrocks_objs PUBLIC ${EXTERNAL_LIBS})
+if (ENABLE_SPEEDB)
+    target_compile_definitions(kvrocks_objs PUBLIC STORAGE_ENGINE=${SPEEDB_STORAGE_ENGINE})
+else()
+    target_compile_definitions(kvrocks_objs PUBLIC STORAGE_ENGINE=${ROCKSDB_STORAGE_ENGINE})
+endif()
```
```suggestion
if (ENABLE_SPEEDB)
    target_compile_definitions(kvrocks_objs PUBLIC KVROCKS_STORAGE_ENGINE=Speedb)
else()
    target_compile_definitions(kvrocks_objs PUBLIC KVROCKS_STORAGE_ENGINE=RocksDB)
endif()
```

src/storage/storage.h
```diff
@@ -40,6 +40,13 @@
 #include "observer_or_unique.h"
 #include "status.h"
 
+enum class StorageEngineType : uint16_t {
+  RocksDB,
+  Speedb,
+};
+
+static constexpr StorageEngineType STORAGE_ENGINE_TYPE = StorageEngineType { STORAGE_ENGINE };
```
```suggestion
inline constexpr StorageEngineType STORAGE_ENGINE_TYPE = StorageEngineType::KVROCKS_STORAGE_ENGINE;
```

PragmaTwice (approved):

PragmaTwice:

PR #1979, opened 2024-01-03
Fix parse BITCOUNT

### LINKED ISSUE #1978: `BITCOUNT` return wrong when args > 4
### Search before asking

### Version

unstable

### Minimal reproduce step

```shell
127.0.0.1:6666> SET mykey "foobar"
OK
127.0.0.1:6666> BITCOUNT mykey 5 30
(integer) 4
127.0.0.1:6666> BITCOUNT mykey 5 30 111
(integer) 26
```

`BITCOUNT mykey 5 30 111` return 4

### What did you see instead?

`BITCOUNT mykey 5 30 111` return ERROR

### Anything Else?

https://github.com/apache/kvrocks/blob/574d438b9acde7b45b0666564af4450c057aceea/src/commands/cmd_bit.cc#L94-L117

We need return ERROR when `args.size() > 4`

tisonkun:
This closes #1978 .


src/commands/cmd_bit.cc
```diff
@@ -113,6 +118,15 @@ class CommandBitCount : public Commander {
       stop_ = *parse_stop;
     }
 
+    if (args.size() == 5) {
+      if (util::EqualICase(args[4], "BYTE")) {
+      } else if (util::EqualICase(args[4], "BIT")) {
```
tisonkun, line 123: BIT modifier is supported in Redis 7.0.

But we assume counting in bytes everywhere so it's non-trivial to implement. It can be a follow-up issue.

tisonkun: https://redis.io/commands/bitcount/

tisonkun:
The patch is trivial and it's not worthy to add a test case ..

git-hulk (approved):

PragmaTwice (approved):

mapleFU (approved):

PR #1987, opened 2024-01-04
Allow to get the latest sequence number when creating the backup

git-hulk:
As mentioned in #1986, some users want to know the sequence number of the backup
for the increment backup scenario. Fortunately, rocksdb allows us to get the sequence
number when creating the checkpoint. So what we need to do is to pass the receiver.


tests/cppunit/storage_test.cc
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#include <__filesystem/operations.h>
```
mapleFU: Hmmm is this expected?

tests/cppunit/storage_test.cc
```diff
@@ -0,0 +1,57 @@
...
+
+  constexpr int cnt = 10;
+  for (int i = 0; i < cnt; i++) {
+    rocksdb::WriteBatch batch;
+    batch.Put("k", "v");
+    assert(storage->Write(rocksdb::WriteOptions(), &batch).ok());
+  }
+  uint64_t sequence_number = 0;
+  s = storage->CreateBackup(&sequence_number);
+  assert(s.IsOK());
```
mapleFU: assert -> ASSERT_TRUE?

src/storage/storage.h
```diff
@@ -123,7 +123,7 @@ class Storage {
   Status SetOptionForAllColumnFamilies(const std::string &key, const std::string &value);
   Status SetDBOption(const std::string &key, const std::string &value);
   Status CreateColumnFamilies(const rocksdb::Options &options);
-  Status CreateBackup();
+  Status CreateBackup(uint64_t *sequence_number = nullptr);
```
mapleFU, line 129: Adding some comments?

tests/cppunit/storage_test.cc
git-hulk: No, Clion auto-imported the wrong header. Good catch！

src/storage/storage.cc
```diff
@@ -363,7 +363,7 @@ Status Storage::CreateBackup() {
   }
 
   std::unique_ptr<rocksdb::Checkpoint> checkpoint_guard(checkpoint);
-  s = checkpoint->CreateCheckpoint(tmpdir, config_->rocks_db.write_buffer_size * MiB);
+  s = checkpoint->CreateCheckpoint(tmpdir, config_->rocks_db.write_buffer_size * MiB, sequence_number);
```
mapleFU, line 366: Just a question:

```
  // sequence_number_ptr: if it is not nullptr, the value it points to will be
  // set to a sequence number guaranteed to be part of the DB, not necessarily
  // the latest. The default value of this parameter is nullptr.
```

From rocksdb, seems this might a bit less than backup? Or I misunderstanding this?

git-hulk: Maybe, I didn't investigate when it would be less than the DB.

mapleFU (dismissed):

src/storage/storage.cc
mapleFU: It was introduced here: https://github.com/facebook/rocksdb/commit/3ffb3baa0b5ce782ea36c8c53adb211127a92dcc

Maybe we should notice user to pay-attention to that, replaying kv might not harm

git-hulk: @mapleFU I have updated the comment to address this, please take a look again.

mapleFU (approved):

src/storage/storage.h
```diff
@@ -123,7 +123,10 @@ class Storage {
   Status SetOptionForAllColumnFamilies(const std::string &key, const std::string &value);
   Status SetDBOption(const std::string &key, const std::string &value);
   Status CreateColumnFamilies(const rocksdb::Options &options);
-  Status CreateBackup();
+  // The sequence_number will be pointed to the value of the sequence number in range of DB,
+  // but can't promise it's the latest sequence number. So you must check it by yourself before
+  // using it.
+  Status CreateBackup(uint64_t *sequence_number = nullptr);
```
PragmaTwice, line 129: Where will we use it? since kvrocks is currently not a library but a execuable.

git-hulk: As mentioned in https://github.com/apache/kvrocks/discussions/1986, they would like to implement the increment backup based on the codebase, so need to export the backup sequence number for this purpose.

PragmaTwice: I think if they have such a requirement, maybe they can implement it in their codebase rather than modifying our codebase.

git-hulk: Yes, I consider adding this because it won't add the extra maintenance burden for the codebase. And it's a common requirement to let them know where's the sequence number of the current backup.

PragmaTwice: I see.

PragmaTwice (approved):

src/storage/storage.cc
git-hulk: I have confirmed the implementation, it is the latest sequence number while starting to create the checkpoint, so it should be less the current backup DB if has any writes during this process.

mapleFU: Yeah I guess maybe it's not atomic between:

```c++
  *sequence_number = db_->GetLatestSequenceNumber();

  LiveFilesStorageInfoOptions opts;
  opts.include_checksum_info = get_live_table_checksum;
  opts.wal_size_for_flush = log_size_for_flush;

  std::vector<LiveFileStorageInfo> infos;
  {
    Status s = db_->GetLiveFilesStorageInfo(opts, &infos);
```

PR #1990, opened 2024-01-06
Add support of BITFIELD_RO

### LINKED ISSUE #1954: Add support of the command BITFIELD_RO
Refer to https://redis.io/commands/bitfield_ro/

jyf111:
This PR adds support of the BITFIELD_RO command.

closes #1954.

Note: The [doc](https://redis.io/commands/bitfield_ro/) shows that the syntax of `bitfield_ro` only allows the get subcommand, but the overflow subcommand can also appear without error, although it has no effect (tested on redis-server).

### DIFF
--- a/src/commands/cmd_bit.cc
+++ b/src/commands/cmd_bit.cc
@@ -242,6 +242,7 @@ class CommandBitOp : public Commander {
   BitOpFlags op_flag_;
 };
 
+template <bool ReadOnly>
 class CommandBitfield : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
@@ -312,6 +313,12 @@ class CommandBitfield : public Commander {
       }
 
       cmds_.push_back(cmd);
+    }
+
+    if constexpr (ReadOnly) {
+      if (!read_only_) {
+        return {Status::RedisParseErr, "BITFIELD_RO only supports the GET subcommand"};
+      }
     }
 
     return Commander::Parse(args);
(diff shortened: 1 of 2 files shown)


src/commands/cmd_bit.cc
```diff
@@ -258,9 +259,15 @@ class CommandBitfield : public Commander {
       } else if (opcode == "set") {
         cmd.type = BitfieldOperation::Type::kSet;
         read_only_ = false;
+        if constexpr (ReadOnly) {
+          return {Status::RedisParseErr, "BITFIELD_RO only supports the GET subcommand"};
+        }
```
PragmaTwice:
```suggestion
        if constexpr (ReadOnly) {
          return {Status::RedisParseErr, "BITFIELD_RO only supports the GET subcommand"};
        }
        cmd.type = BitfieldOperation::Type::kSet;
        read_only_ = false;
```

src/commands/cmd_bit.cc
```diff
@@ -258,9 +259,15 @@ class CommandBitfield : public Commander {
...
       } else if (opcode == "incrby") {
         cmd.type = BitfieldOperation::Type::kIncrBy;
         read_only_ = false;
+        if constexpr (ReadOnly) {
```
PragmaTwice: ditto

src/commands/cmd_bit.cc
```diff
@@ -258,9 +259,15 @@ class CommandBitfield : public Commander {
       } else if (opcode == "set") {
         cmd.type = BitfieldOperation::Type::kSet;
         read_only_ = false;
+        if constexpr (ReadOnly) {
```
mapleFU: Could we unifying these flow above `return Commander::Parse(args);`?

mapleFU (commented):
General looks ok to me

src/commands/cmd_bit.cc
```diff
@@ -314,6 +322,12 @@ class CommandBitfield : public Commander {
       cmds_.push_back(cmd);
     }
 
+    if constexpr (ReadOnly) {
+      if (group.Good()) {
```
mapleFU: Hmmm:

```
... // Parse as origin

    if constexpr (ReadOnly) {
      if (!readonly_) {
         return {Status::RedisParseErr, "BITFIELD_RO only supports the GET subcommand"};
      }
```

Would it be more straitforward?

jyf111: Yes, I was just struggling about this, but it would require an additional assignment to `read_only_`.

```cpp
read_only_ = false;
if constexpr (ReadOnly) {
  break;
}
```

I'm not sure which one is better

mapleFU: Hmm maybe we don't need to break? Just parsing it is ok for me, seems it should not be a bottleneck?

jyf111: ok, get your point

mapleFU (approved):

PragmaTwice (approved):

git-hulk (approved):

PR #1999, opened 2024-01-10
Add support of new command: RESET

### LINKED ISSUE #1944: Add support of the command RESET
Refer to https://redis.io/commands/reset/

chrisxu333:
This closes #1944


src/commands/cmd_server.cc
```diff
@@ -1221,6 +1221,39 @@ class CommandAnalyze : public Commander {
...
+    if (conn->IsFlagEnabled(Connection::kMonitor)) {
+      conn->Owner()->QuitMonitorConn(conn);
+    }
+    // 10. Aborts Pub/Sub's subscription state (SUBSCRIBE and PSUBSCRIBE), when appropriate.
+    if (conn->SubscriptionsCount() != 0) {
+      conn->UnsubscribeAll();
+    }
+    if (conn->PSubscriptionsCount() != 0) {
+      conn->PUnsubscribeAll();
+    }
```
git-hulk, line 1248: The connection will not receive any commands after entering the monitor or subscribe mode, so I guess it should be unable to cancel them here.

tests/gocase/unit/reset/reset_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+		require.NoError(t, rdb.Set(ctx, "x", 40, 0).Err())
+		require.NoError(t, rdb.Do(ctx, "multi").Err())
+		require.NoError(t, rdb.Do(ctx, "ping").Err())
+		require.Equal(t, rdb.Do(ctx, "exec").Val(), []interface{}{"PONG"})
+	})
+
+	t.Run("unsub and punsub", func(t *testing.T) {
+		require.NoError(t, rdb.Do(ctx, "subscribe", "chan1").Err())
+		require.NoError(t, rdb.Do(ctx, "reset").Err())
```
git-hulk, line 64: This test case works well because go-redis will use another connection while sending the reset command. If you re-subscribe the `chan1`, I guess the subscriber count will be 2 instead of 1.

src/commands/cmd_server.cc
chrisxu333: Oh I didn't know that. I thought only redis-cli would behave this way. Do you have any idea how do bypass this?

tests/gocase/unit/reset/reset_test.go
chrisxu333: I changed the code to the following snippet, and it works well too.

```	
t.Run("unsub and punsub", func(t *testing.T) {
		require.NoError(t, rdb.Do(ctx, "subscribe", "chan1").Err())
		require.NoError(t, rdb.Do(ctx, "reset").Err())
        require.Equal(t, rdb.Do(ctx, "subscribe", "chan1").Val(), []interface{}{"subscribe", "chan1", (int64)(1)})
})
```

git-hulk: @chrisxu333 could you print some logs inside the RESET command, to see if the subscriber was cancelled

src/commands/cmd_server.cc
git-hulk: @chrisxu333 Maybe I was wrong. We didn't remove the READABLE event from fd after entering the monitor/subscriber mode on the server side. So it may be good for the current implementation. You can print some logs or use debug mode to confirm this.

tests/gocase/unit/reset/reset_test.go
chrisxu333: image

I think this shows that the subscriber is actually cancelled. And below is the corresponding code:

image

src/commands/cmd_server.cc
chrisxu333: Here is the log and corresponding code:
image
image

Is it enough to verify?

git-hulk: Sure, thank you!

git-hulk (approved):

PragmaTwice (approved):

PragmaTwice:

PR #2004, opened 2024-01-11
Implement an unify key-value iterator for Kvrocks

### LINKED ISSUE #1989: Implement the migrate iterator for iterating key values
Currently, we need to iterate all keys in the database in different places like the cluster migration and kvrocks2redis, but don't have an iterator for this purpose. It's very error-prone to implement this in different places since Kvrocks may add a new column family in the future, and we must be careful to iterate all keys in all column families. This would be a burden for maintenance, So I think we could add an iterator to iterate all keys in all column families.

## Proposal

For the API, we could generally comply with the rocksdb's iterator API, but we allow to add some extra functions if needed.

```C++

class Iterator {
public:
    Iterator(Storage *storage, const rocksdb::ReadOptions &options, const int slot = -1);
    ~KeyIterator();

    bool Valid() const;
    void Next();

    rocksdb::WriteBatch *Batch() const;
    Slice Key() const;
    Slice Value() const;
    RedisType Type() const;
};

```

And when implementing this iterator, it will iterate the metadata column family first and check its type, if it's not a string, then it will iterate the corresponding column family to get subkeys. That said, if we have a key `foo` with type `hash`, then the iterator will iterate `foo` and `foo:field1`, `foo:field2`, and so on.

This solution can bring those benefits:

- The codes look more intutive
- Can reuse this iterator if we want to iterate keys only

git-hulk:
Currently, we need to iterate all keys in the database in different places like the cluster migration and kvrocks2redis, but don't have an iterator for this purpose. It's very error-prone to implement this in different places since Kvrocks may add a new column family in the future, and we must be careful to iterate all keys in all column families. This would be a burden for maintenance, So we want to implement an iterator for iterating keys.

```C++

DBIter iter(storage, read_option);
for (iter.Seek(); iter.Valid(); iter.Next()) {
    if (iter.Type() == kRedisString || iter.Type() == kRedisJSON) {
        // the string/json type didn't have subkeys
        continue;
    }

    auto subkey_iter = iter.GetSubKeyIterator();
    for (subkey_iter.Seek(); subkey_iter.Valid(); subkey_iter.Next()) {
        // handle its subkey and value here
    }
}

```

When using this iterator, it will iterate the metadata column family first and check its type, if it's not a string or JSON, then it will iterate the corresponding column family to get subkeys. That said, if we have a key foo with type hash, then the iterator will iterate foo and foo:field1, foo:field2, and so on.

This solution can bring those benefits:

- The codes look more intuitive
- Can reuse this iterator if we want to iterate keys only

This closes #1989


caipengbo (dismissed):
Excellent code, LGTM!

src/storage/iterator.h
```diff
@@ -0,0 +1,82 @@
...
+#pragma once
+
+#include <rocksdb/iterator.h>
+#include <rocksdb/options.h>
+
+#include "storage.h"
+
+namespace engine {
+
+class SubKeyIterator {
```
caipengbo, line 29: Should we provide an interface? Will `Iterator` for incremental data be implemented later?

git-hulk: This PR doesn't involve this scope, we can add it in another PR for parsing the increment data. What do you think? @caipengbo

caipengbo: Got it

caipengbo (dismissed):

jihuayu:
I have 2 question:
1. I found the test doesn't have Bloom, JSON and SortedInt type. Is that OK?
2. When I iter the zset subkey, it only returns the subkeys `key|version|member=>score`, not return subkeys `key|version|score|member=>NULL`?

git-hulk:
@jihuayu Thanks for your review.

> 1. I found the test doesn't have Bloom, JSON and SortedInt type. Is that OK?

I'm not quite sure about how to add test cases for bloom. Welcome to send a PR after this if anyone knows that. I will also add a test case for JSON and SortedInt.

> When I iter the zset subkey, it only returns the subkeys key|version|member=>score, not return subkeys key|version|score|member=>NULL?

Yes, it's not necessary to scan the score column family since it has enough information from the subkey column family.

src/storage/iterator.cc
```diff
@@ -0,0 +1,162 @@
...
+}
+
+SubKeyIterator::SubKeyIterator(Storage* storage, rocksdb::ReadOptions read_options, RedisType type, std::string prefix)
+    : storage_(storage), read_options_(std::move(read_options)), type_(type), prefix_(std::move(prefix)) {
+  if (type_ == kRedisStream) {
+    cf_handle_ = storage_->GetCFHandle(kStreamColumnFamilyName);
+  } else {
+    cf_handle_ = storage_->GetCFHandle(kSubkeyColumnFamilyName);
+  }
+  iter_ = std::move(util::UniqueIterator(storage_->NewIterator(read_options_, cf_handle_)));
```
mapleFU: don't need `std::move`?

src/storage/iterator.cc
```diff
@@ -0,0 +1,162 @@
...
+}
+
+void SubKeyIterator::Next() {
+  iter_->Next();
```
mapleFU, line 129: Should we DCHECK valid before calling `iter_->Next()`?

src/storage/iterator.cc
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  // The string/json type doesn't have sub keys
+  RedisType type = metadata_.Type();
+  if (type == kRedisNone || type == kRedisString || type == kRedisJson) {
+    return nullptr;
+  }
+
+  auto prefix = InternalKey(Key(), "", metadata_.version, storage_->IsSlotIdEncoded()).Encode();
+  return std::make_unique<SubKeyIterator>(storage_, read_options_, type, prefix);
```
mapleFU:
```suggestion
  return std::make_unique<SubKeyIterator>(storage_, read_options_, type, std::move(prefix));
```

src/storage/iterator.cc
```diff
@@ -0,0 +1,162 @@
...
+
+#include <cluster/redis_slot.h>
+
+#include "db_util.h"
+
+namespace engine {
+DBIterator::DBIterator(Storage* storage, rocksdb::ReadOptions read_options, int slot)
+    : storage_(storage), read_options_(std::move(read_options)), slot_(slot) {
+  metadata_cf_handle_ = storage_->GetCFHandle(kMetadataColumnFamilyName);
+  metadata_iter_ = std::move(util::UniqueIterator(storage_->NewIterator(read_options_, metadata_cf_handle_)));
```
mapleFU: remove `std::move`?

src/storage/iterator.cc
```diff
@@ -0,0 +1,162 @@
...
+}
+
+void DBIterator::Next() {
+  metadata_iter_->Next();
+  nextUntilValid();
+}
+
+void DBIterator::nextUntilValid() {
```
mapleFU, line 41: So actually the `valid` means "unexpired"?

src/storage/iterator.cc
```diff
@@ -0,0 +1,162 @@
...
+      return;
+    }
+  }
+
+  while (metadata_iter_->Valid()) {
+    Metadata metadata(kRedisNone, false);
+    // Skip the metadata if it's expired
+    if (metadata.Decode(metadata_iter_->value()).ok() && !metadata.Expired()) {
+      metadata_ = metadata;
+      break;
```
mapleFU, line 58: Would after that, the `metadata_iter_` point to another slot?

git-hulk: That's impossible because we have checked it in line 45.

src/storage/iterator.cc
git-hulk: Yes, should be valid and the key is not expired yet.

caipengbo (approved):

PR #2015, opened 2024-01-14
Add type check before metadata decode 

### LINKED ISSUE #2012: XADD to other type key returns wrong error message
### Search before asking

### Version

unstable

### Minimal reproduce step

```shell
127.0.0.1:6666> set a test
OK
127.0.0.1:6666> lpush a xx
(error) ERR Invalid argument: WRONGTYPE Operation against a key holding the wrong kind of value
127.0.0.1:6666> XADD a * a hello
(error) ERR Invalid argument: metadata is too short
```

`XADD a * a hello` command return "ERR Invalid argument: WRONGTYPE Operation against a key holding the wrong kind of value"

### What did you see instead?

`XADD a * a hello` command return "ERR Invalid argument: metadata is too short"

### Anything Else?

Other stream commands may also have this problem.

jihuayu:
In this PR I add type check before metadata decode.
I also add some test cases about operating with the wrong type
It fixes #2012


src/storage/redis_metadata.cc
```diff
@@ -372,29 +372,31 @@ rocksdb::Status StreamMetadata::Decode(Slice *input) {
     return s;
   }
 
-  if (input->size() < 8 * 11) {
-    return rocksdb::Status::InvalidArgument(kErrMetadataTooShort);
-  }
+  if (Type() == kRedisStream) {
```
PragmaTwice: It is already assumed that in `StreamMetadata::Decode` the type must be kRedisStream, which should be checked in other places. So I don't think it is meaningful.

tests/gocase/unit/type/types_test.go
```diff
@@ -0,0 +1,54 @@
...
+		message := "ERR Invalid argument: WRONGTYPE Operation against a key holding the wrong kind of value"
+		require.NoError(t, rdb.Set(ctx, "a", "hello", 0).Err())
+		require.EqualError(t, rdb.Do(ctx, "XADD", "a", "*", "a", "test").Err(), message)
+		require.EqualError(t, rdb.Do(ctx, "LPUSH", "a", 1).Err(), message)
+		require.EqualError(t, rdb.Do(ctx, "HSET", "a", "1", "2").Err(), message)
+		require.EqualError(t, rdb.Do(ctx, "SADD", "a", "1", "2").Err(), message)
+		require.EqualError(t, rdb.Do(ctx, "ZADD", "a", "1", "2").Err(), message)
+		require.EqualError(t, rdb.Do(ctx, "JSON.SET", "a", "$", "{}").Err(), message)
+		require.EqualError(t, rdb.Do(ctx, "BF.ADD", "a", "test").Err(), message)
+		require.EqualError(t, rdb.Do(ctx, "SADD", "a", 100).Err(), message)
```
PragmaTwice, line 47: I think the test case can pass without these changes, regardless of error messages.

PragmaTwice (changes requested):
See comments : )

PragmaTwice:
Instead, if you want to fix #2012, I think a better way is to change `Metadata::ParseMetadata`.

src/storage/redis_metadata.cc
jihuayu: I am according to old code, maybe this code is unused now.  https://github.com/apache/kvrocks/blob/8d30e1eb2476f7c42d6c90f5cf112924e1d48ba9/src/storage/redis_metadata.cc#L338-L344


> I think a better way is to change `Metadata::ParseMetadata`.

Great, I will move all to `Metadata::ParseMetadata`.

tests/gocase/unit/type/types_test.go
jihuayu: This test case is designed to ensure that the error message is consistent with Redis. 
Based on my experience, it's difficult for me to associate the problem with inconsistent key types when I see other error messages.
So I wrote this test case to prevent future users from being confused by meaningless error messages.

jihuayu:
@PragmaTwice 
I delay InvalidArgument error check after type match check in `Metadata::ParseMetadata`.

For the tests, I think it is necessary.

git-hulk (approved):

torwig (approved):

PragmaTwice (approved):
LGTM

PR #2016, opened 2024-01-14
Add support of new command: ZRANDMEMBER 

JxLi0921:
Add support of the ZRANDMEMBER command. This is related to #1985

PR #2017, opened 2024-01-14
Implement the RESP3 null for the nil string and array

### LINKED ISSUE #1993: Add RESP3 null type

git-hulk:
For the null type in RESP2, the nil string is represented by `$-1\r\n`
and nil array is `*-1\r\n`. But after RESP3, they will be reduced into
a unified string `_\r\n`. So in this PR, most of them are renamed redis::MultiBulkString/Redis::NilString to connection::MultiBulkString and connection::NilString.

This closes https://github.com/apache/kvrocks/issues/1993

### DIFF
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -130,6 +130,39 @@ void Connection::Reply(const std::string &msg) {
   redis::Reply(bufferevent_get_output(bev_), msg);
 }
 
+std::string Connection::Bool(bool b) const {
+  if (protocol_version_ == RESP::v3) {
+    return b ? "#t" CRLF : "#f" CRLF;
+  }
+  return Integer(b ? 1 : 0);
+}
+
+std::string Connection::MultiBulkString(const std::vector<std::string> &values,
+                                        bool output_nil_for_empty_string) const {
+  std::string result = "*" + std::to_string(values.size()) + CRLF;
+  for (const auto &value : values) {
+    if (value.empty() && output_nil_for_empty_string) {
+      result += NilString();
+    } else {
+      result += BulkString(value);
+    }
+  }
+  return result;
+}
+
+std::string Connection::MultiBulkString(const std::vector<std::string> &values,
+                                        const std::vector<rocksdb::Status> &statuses) const {
+  std::string result = "*" + std::to_string(values.size()) + CRLF;
+  for (size_t i = 0; i < values.size(); i++) {
+    if (i < statuses.size() && !statuses[i].ok()) {
+      result += NilString();
+    } else {
+      result += BulkString(values[i]);
+    }
+  }
+  return result;
+}
+
 void Connection::SendFile(int fd) {
   // NOTE: we don't need to close the fd, the libevent will do that
   auto output = bufferevent_get_output(bev_);
--- a/src/server/redis_reply.cc
+++ b/src/server/redis_reply.cc
@@ -41,7 +41,7 @@ std::string Array(const std::vector<std::string> &list) {
   return result;
 }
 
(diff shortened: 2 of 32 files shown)


git-hulk:
This PR involves many files, but most of them are renaming the function.

torwig (dismissed):
LGTM. Good job, @git-hulk !

git-hulk:
@torwig I changed NilString in the last commit for the new command ZRANDMEMBER which was introduced in #2016, please retake a look.

torwig (dismissed):

src/server/redis_connection.cc
```diff
@@ -130,6 +130,39 @@ void Connection::Reply(const std::string &msg) {
   redis::Reply(bufferevent_get_output(bev_), msg);
 }
 
+std::string Connection::Bool(bool b) const {
+  if (protocol_version_ == RESP::v3) {
+    return b ? "#t" CRLF : "#f" CRLF;
+  }
+  return Integer(b ? 1 : 0);
```
PragmaTwice, line 137: In RESPv2, will the `false` be mapped to `0` or `nil`?

src/server/redis_connection.cc
```diff
@@ -130,6 +130,39 @@ void Connection::Reply(const std::string &msg) {
...
+}
+
+std::string Connection::MultiBulkString(const std::vector<std::string> &values,
+                                        bool output_nil_for_empty_string) const {
```
PragmaTwice, line 141: Is there any scenario that `output_nil_for_empty_string` is false?

src/server/redis_connection.cc
git-hulk: Only one place will map false to nil, see: https://github.com/apache/kvrocks/blob/e239e83d55dd2ece43f8b1cd8cbd2a5e44653b42/src/storage/scripting.cc#L1089

src/server/redis_connection.cc
PragmaTwice: Ahh I see it is used in Array2RESP. But if it is just used here maybe we can remove it.

src/server/redis_reply.cc
```diff
@@ -67,6 +41,12 @@ std::string Array(const std::vector<std::string> &list) {
   return result;
 }
 
-std::string Command2RESP(const std::vector<std::string> &cmd_args) { return MultiBulkString(cmd_args, false); }
+std::string Array2RESP(const std::vector<std::string> &elems) {
```
PragmaTwice:
```suggestion
std::string ArrayOfBulkStrings(const std::vector<std::string> &elems) {
```

src/server/redis_connection.cc
PragmaTwice: Ahh I see. I wonder what the behavior is in Redis?

git-hulk: Yes, I have confirmed this while changing this line.

PragmaTwice: Got it.

src/server/redis_connection.cc
git-hulk: Yes, it will set to false if we don't want to convert the empty string to nil. For example, in SMEMBERS command we don't want to convert the empty member string to nil.

PragmaTwice: I see. Thank you.

src/server/redis_reply.cc
git-hulk: Done in the last commit

PragmaTwice (approved):

git-hulk:
Thanks for @torwig @PragmaTwice review.

PR #2020, opened 2024-01-15
persist cluster-enabled status in RocksDB

### LINKED ISSUE #1410: Persist the cluster-enabled status in RocksDB

chrisxu333:
To close #1410 .

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -43,8 +43,11 @@ daemonize no
 # If you enable cluster, kvrocks will encode key with its slot id calculated by
 # CRC16 and modulo 16384, encoding key with its slot id makes it efficient to
 # migrate keys based on the slot. So if you enabled at first time, cluster mode must
-# not be disabled after restarting, and vice versa. That is to say, data is not
-# compatible between standalone mode with cluster mode, you must migrate data
+# not be disabled after restarting, and vice versa. Currently, kvrocks will keep 
+# using the cluster-enabled status that is persisted at first time, regardless of 
+# what cluster-enabled status is provided afterwards.
+# Note that even if kvrocks has such protection, you should also be aware that data
+# is not compatible between standalone mode with cluster mode, you must migrate data
 # if you want to change mode, otherwise, kvrocks will make data corrupt.
 #
 # Default: no
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -152,6 +152,11 @@ Status Server::Start() {
     }
   }
 
+  s = checkClusterMode();
+  if (!s.IsOK()) {
+    return s;
+  }
+
   if (config_->cluster_enabled) {
     if (config_->persist_cluster_nodes_enabled) {
       auto s = cluster->LoadClusterNodes(config_->NodesFilePath());
@@ -1836,6 +1841,27 @@ void Server::cleanupExitedWorkerThreads(bool force) {
       // Push the worker thread back to the queue if it's still running.
       recycle_worker_threads_.push(std::move(worker_thread));
     }
+  }
+}
+
+Status Server::checkClusterMode() {
+  std::string value;
+  auto cf = storage->GetCFHandle(engine::kPropagateColumnFamilyName);
+  rocksdb::Status check_cluster_enabled =
+      storage->Get(rocksdb::ReadOptions(), cf, rocksdb::Slice(engine::kClusterEnabledKey), &value);
+
+  if (check_cluster_enabled.IsNotFound()) {
+    return storage->WriteToPropagateCF(engine::kClusterEnabledKey, std::to_string(config_->cluster_enabled));
+  }
+  if (check_cluster_enabled.ok()) {
+    auto is_enabled = GET_OR_RET(ParseInt<int64_t>(value, 10));
+    if (config_->cluster_enabled != is_enabled) {
+      LOG(ERROR) << "cluster_enabled status from config file is inconsistent with the persisted one";
+      return {Status::NotOK, "inconsistent cluster_enabled status"};
+    }
+    return Status::OK();
+  } else {
+    return {Status::NotOK, "failed to load cluster_enabled from storage: " + check_cluster_enabled.ToString()};
   }
 }
 
--- a/src/storage/storage.h
+++ b/src/storage/storage.h
@@ -80,6 +80,8 @@ constexpr const char *kLuaFuncSHAPrefix = "lua_f_";
 constexpr const char *kLuaFuncLibPrefix = "lua_func_lib_";
 constexpr const char *kLuaLibCodePrefix = "lua_lib_code_";
 
+const std::string kClusterEnabledKey = "config_cluster_enabled";
+
 struct CompressionOption {
   rocksdb::CompressionType type;
   const std::string name;
--- a/src/server/server.h
+++ b/src/server/server.h
@@ -311,6 +311,7 @@ class Server {
   void increaseWorkerThreads(size_t delta);
   void decreaseWorkerThreads(size_t delta);
   void cleanupExitedWorkerThreads(bool force);
+  Status checkClusterMode();
 
   std::atomic<bool> stop_ = false;
   std::atomic<bool> is_loading_ = false;
--- a/tests/gocase/integration/cluster/cluster_test.go
+++ b/tests/gocase/integration/cluster/cluster_test.go
@@ -93,6 +93,17 @@ func TestClusterNodes(t *testing.T) {
 		require.EqualValues(t, []redis.ClusterNode{{ID: nodeID, Addr: srv.HostPort()}}, slots[0].Nodes)
 	})
 
+	t.Run("enable/disable cluster-enabled option", func(t *testing.T) {
+		// force change cluster-enabled status in kvrocks.conf file
+		srv.ForceChangeClusterMode(false)
+		defer func() {
+			srv.ForceChangeClusterMode(true)
+			srv.Restart(util.RestartOpt{Nowait: false, Noclose: true})
+		}()
+		srv.Restart(util.RestartOpt{Nowait: true, Noclose: false})
+		require.ErrorContains(t, rdb.Do(ctx, "clusterx", "version").Err(), "connection refused")
+	})
+
 	t.Run("enable/disable the persist cluster nodes", func(t *testing.T) {
 		require.NoError(t, rdb.ConfigSet(ctx, "persist-cluster-nodes-enabled", "yes").Err())
 		srv.Restart()
--- a/tests/gocase/util/server.go
+++ b/tests/gocase/util/server.go
@@ -28,6 +28,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"sync"
 	"syscall"
 	"testing"
@@ -38,6 +39,11 @@ import (
 	"github.com/stretchr/testify/require"
 	"golang.org/x/exp/slices"
 )
+
+type RestartOpt struct {
+	Nowait  bool
+	Noclose bool
+}
 
 type KvrocksServer struct {
 	t   testing.TB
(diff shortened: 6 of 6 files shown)


chrisxu333:
@git-hulk Please help review :)

src/server/server.cc
```diff
@@ -152,6 +152,19 @@ Status Server::Start() {
     }
   }
 
+  std::string value;
```
caipengbo: We'd better clarify this behavior in the configuration file:

https://github.com/apache/kvrocks/blob/9d656f1476583651e2fae20ed1c37cf35d0679ca/kvrocks.conf#L46-L48

src/server/server.cc
```diff
@@ -152,6 +152,19 @@ Status Server::Start() {
...
+  auto cf = storage->GetCFHandle(engine::kPropagateColumnFamilyName);
+  rocksdb::Status check_cluster_enabled = storage->Get(rocksdb::ReadOptions(), cf, rocksdb::Slice(engine::kClusterEnabledKey), &value);
+  if(check_cluster_enabled.IsNotFound()) {
+    s = storage->WriteToPropagateCF(engine::kClusterEnabledKey, std::to_string(config_->cluster_enabled));
+    if(!s.IsOK()) return s;
+  } else if(check_cluster_enabled.ok()) {
+    // use the one that's previously persisted.
+    config_->cluster_enabled = std::stoi(value);
```
caipengbo: In this line, I suggest printing a warning log if the persistence is inconsistent with the configuration file.

src/server/server.cc
```diff
@@ -152,6 +152,19 @@ Status Server::Start() {
...
+  } else {
+    return {Status::NotOK, "failed reading from storage"};
```
caipengbo: "failed to load cluster_enable from storage:  "+ check_cluster_enabled.ToString()

src/storage/storage.h
```diff
@@ -80,6 +80,8 @@ constexpr const char *kLuaFuncSHAPrefix = "lua_f_";
 constexpr const char *kLuaFuncLibPrefix = "lua_func_lib_";
 constexpr const char *kLuaLibCodePrefix = "lua_lib_code_";
 
+const std::string kClusterEnabledKey = "cluster_enabled";
```
```suggestion
const std::string kClusterEnabledKey = "config_cluster_enabled";
```

src/server/server.cc
```diff
@@ -152,6 +152,19 @@ Status Server::Start() {
...
+  }
```
PragmaTwice, line 158: Could you move these lines to a separate function?

src/server/server.cc
PragmaTwice: I prefer a fatal error rather than a warning.

torwig:
@chrisxu333 Please, run `./x.py format` to format the codebase and CI will be happy :)

src/server/server.cc
```diff
@@ -1839,6 +1844,26 @@ void Server::cleanupExitedWorkerThreads(bool force) {
   }
 }
 
+Status Server::forceClusterMode() {
```
git-hulk: Is overrideClusterMode better than forceClusterMode?

src/server/server.cc
```diff
@@ -1839,6 +1844,26 @@ void Server::cleanupExitedWorkerThreads(bool force) {
...
+  auto cf = storage->GetCFHandle(engine::kPropagateColumnFamilyName);
+  rocksdb::Status check_cluster_enabled =
+      storage->Get(rocksdb::ReadOptions(), cf, rocksdb::Slice(engine::kClusterEnabledKey), &value);
+
+  if (check_cluster_enabled.IsNotFound()) {
+    Status s = storage->WriteToPropagateCF(engine::kClusterEnabledKey, std::to_string(config_->cluster_enabled));
+    if (!s.IsOK()) return s;
+  } else {
+    if (check_cluster_enabled.ok()) {
+      LOG(WARNING) << "cluster_enable status is inconsistent. Using the previously persisted one.";
```
git-hulk: Need to compare the configuration between the config file and rocksdb? or it might confuse users.

src/server/server.cc
```diff
@@ -1839,6 +1844,26 @@ void Server::cleanupExitedWorkerThreads(bool force) {
...
 
+Status Server::forceClusterMode() {
+  std::string value;
+  auto cf = storage->GetCFHandle(engine::kPropagateColumnFamilyName);
+  rocksdb::Status check_cluster_enabled =
+      storage->Get(rocksdb::ReadOptions(), cf, rocksdb::Slice(engine::kClusterEnabledKey), &value);
+
+  if (check_cluster_enabled.IsNotFound()) {
+    Status s = storage->WriteToPropagateCF(engine::kClusterEnabledKey, std::to_string(config_->cluster_enabled));
+    if (!s.IsOK()) return s;
```
git-hulk: Can return the status directly if the key is not found
```suggestion
    return storage->WriteToPropagateCF(engine::kClusterEnabledKey, std::to_string(config_->cluster_enabled));
```

src/server/server.cc
```diff
@@ -1839,6 +1844,26 @@ void Server::cleanupExitedWorkerThreads(bool force) {
...
 }
 
+Status Server::forceClusterMode() {
+  std::string value;
+  auto cf = storage->GetCFHandle(engine::kPropagateColumnFamilyName);
+  rocksdb::Status check_cluster_enabled =
+      storage->Get(rocksdb::ReadOptions(), cf, rocksdb::Slice(engine::kClusterEnabledKey), &value);
+
+  if (check_cluster_enabled.IsNotFound()) {
+    Status s = storage->WriteToPropagateCF(engine::kClusterEnabledKey, std::to_string(config_->cluster_enabled));
```
git-hulk: I prefer using `yes` or `no` instead of the bool value.

kvrocks.conf
```diff
@@ -43,8 +43,11 @@ daemonize no
 # If you enable cluster, kvrocks will encode key with its slot id calculated by
 # CRC16 and modulo 16384, encoding key with its slot id makes it efficient to
 # migrate keys based on the slot. So if you enabled at first time, cluster mode must
-# not be disabled after restarting, and vice versa. That is to say, data is not
-# compatible between standalone mode with cluster mode, you must migrate data
+# not be disabled after restarting, and vice versa. Currently, kvrocks will keep 
+# using the cluster-enabled status that is persisted at first time, regardless of 
+# what cluster-enabled status is provided afterwards.
```
caipengbo, line 48: force rewrite or can not start the server? HDYT @git-hulk @PragmaTwice

src/server/server.cc
chrisxu333: Could you elaborate more on this?

kvrocks.conf
git-hulk: Both of them are good for me. I prefer throwing an error if we have to choose one.

chrisxu333: Okay I'll change accordingly. By throwing an error you mean also to prevent the user from starting the server on mismatch right?

caipengbo: Me too, I also tend to just throw an error that prevents the user from starting the server. This is equivalent to adding a quick pre check.

git-hulk: > By throwing an error you mean also to prevent the user from starting the server on mismatch right?

That's right.

src/server/server.cc
git-hulk: Use `yes` or `no` as the value in rocksdb instead of `0` or `1` with to_string.

src/server/server.cc
```diff
@@ -1839,6 +1844,26 @@ void Server::cleanupExitedWorkerThreads(bool force) {
...
+  std::string value;
+  auto cf = storage->GetCFHandle(engine::kPropagateColumnFamilyName);
+  rocksdb::Status check_cluster_enabled =
+      storage->Get(rocksdb::ReadOptions(), cf, rocksdb::Slice(engine::kClusterEnabledKey), &value);
+
+  if (check_cluster_enabled.IsNotFound()) {
+    return storage->WriteToPropagateCF(engine::kClusterEnabledKey, std::to_string(config_->cluster_enabled));
+  }
+  if (check_cluster_enabled.ok()) {
+    if (config_->cluster_enabled != stoi(value)) {
```
PragmaTwice: Please use `ParseInt` in `parse_utl.h` instead of `stoi` here to avoid exception throwing.

src/server/server.cc
```diff
@@ -1839,6 +1844,26 @@ void Server::cleanupExitedWorkerThreads(bool force) {
...
+      LOG(ERROR) << "cluster_enable status from config file is inconsistent with the persisted one";
+      return {Status::NotOK, "mismatching cluster_enable status"};
```
```suggestion
      return {Status::NotOK, "mismatched cluster_enabled status"};
```

src/server/server.cc
```diff
@@ -1839,6 +1844,26 @@ void Server::cleanupExitedWorkerThreads(bool force) {
...
+  auto cf = storage->GetCFHandle(engine::kPropagateColumnFamilyName);
+  rocksdb::Status check_cluster_enabled =
+      storage->Get(rocksdb::ReadOptions(), cf, rocksdb::Slice(engine::kClusterEnabledKey), &value);
+
+  if (check_cluster_enabled.IsNotFound()) {
+    return storage->WriteToPropagateCF(engine::kClusterEnabledKey, std::to_string(config_->cluster_enabled));
+  }
+  if (check_cluster_enabled.ok()) {
+    if (config_->cluster_enabled != stoi(value)) {
+      LOG(ERROR) << "cluster_enable status from config file is inconsistent with the persisted one";
```
```suggestion
      LOG(ERROR) << "cluster_enabled status from config file is inconsistent with the persisted one";
```

src/server/server.cc
```diff
@@ -1839,6 +1844,26 @@ void Server::cleanupExitedWorkerThreads(bool force) {
   }
 }
 
+Status Server::overrideClusterMode() {
```
PragmaTwice: Could you give it a better name? IMHO "override" is not appropriate here.

git-hulk: How about checkClusterMode or validateClusterMode if it throws an error once the cluster mode is mismatched between the db and the configuration file?

chrisxu333: I think `checkClusterMode` would be better. Changed accordingly.

src/server/server.cc
```diff
@@ -1839,6 +1844,27 @@ void Server::cleanupExitedWorkerThreads(bool force) {
...
+      storage->Get(rocksdb::ReadOptions(), cf, rocksdb::Slice(engine::kClusterEnabledKey), &value);
+
+  if (check_cluster_enabled.IsNotFound()) {
+    return storage->WriteToPropagateCF(engine::kClusterEnabledKey, std::to_string(config_->cluster_enabled));
+  }
+  if (check_cluster_enabled.ok()) {
+    auto is_enabled = GET_OR_RET(ParseInt<int64_t>(value, 10));
+    if (config_->cluster_enabled != is_enabled) {
+      LOG(ERROR) << "cluster_enabled status from config file is inconsistent with the persisted one";
+      return {Status::NotOK, "mismatching cluster_enabled status"};
```
```suggestion
      return {Status::NotOK, "inconsistent cluster_enabled status"};
```

chrisxu333:
Hi @git-hulk could you also help diagnose why the newly introduced testcase is failed? It was successful when I did locally.

git-hulk:
> Hi @git-hulk could you also help diagnose why the newly introduced testcase is failed? It was successful when I did locally.

perhaps you can try to debug the test case yourself, and I will also take a look while I'm free.

src/server/server.cc
```diff
@@ -152,6 +152,11 @@ Status Server::Start() {
     }
   }
 
+  s = checkClusterMode();
```
PragmaTwice, line 155: Seems this should be before the `namespace_.LoadAndRewrite()`. To avoid some weird error.

Issue #1984, opened 2024-01-04
Add support of the command ZDIFF and ZDIFFSTORE


PragmaTwice:
Refer to https://redis.io/commands/zdiff/

MaheshMadushan:
Hi, is this issue available? if so, shall I try on this?

PragmaTwice:
Sure! Assigned.

jihuayu:
Close via #2021

PR #2021, opened 2024-01-15
Add support of the command ZDIFF and ZDIFFSTORE

MaheshMadushan:
Add support of the command ZDIFF and ZDIFFSTORE #1984
- [x] ZDIFF
- [x] ZDIFFSTORE


torwig:
@MaheshMadushan Thank you for the contribution. 
You can format the codebase by running `./x.py format`.

MaheshMadushan:
> @MaheshMadushan Thank you for the contribution. You can format the codebase by running `./x.py format`.

Hi @torwig , formatting issue fixed. Thank you.

torwig:
@MaheshMadushan LGTM, nicely done!
Could you please add some edge-cases tests,? For example, if the first set is empty, doesn't exist, one of the other sets doesn't exist, etc.

MaheshMadushan:
Hi @torwig . Thanks for the feedback. will add edge test cases for the feature.

src/types/redis_zset.cc
```diff
@@ -930,4 +930,46 @@ rocksdb::Status ZSet::RandMember(const Slice &user_key, int64_t command_count,
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status ZSet::Diff(const std::vector<Slice> &keys, MemberScores *members) {
+  std::vector<std::string> lock_keys;
```
git-hulk: It should be fine to read the snapshot instead of locking keys.

git-hulk:
Overall are good, one comment inline.

src/types/redis_zset.cc
git-hulk: Other places lock keys, so let's leave it as it is and fix them together.

git-hulk (dismissed):

src/types/redis_zset.cc
MaheshMadushan: Hi @git-hulk, I really appreciate your feedback. I'd be happy to address this in the current PR if that works for you. Additionally, I was wondering if you could provide some more insights into the snapshotting process in kvrocks, specifically whether reading from snapshots is mutually exclusive? Your clarification would be incredibly helpful. Thanks a lot!

git-hulk: @MaheshMadushan Thanks for your contribution. We don't need to use a lock while reading keys. For the read snapshot, please refer: https://github.com/apache/kvrocks/blob/unstable/src/types/redis_zset.cc#L839

>  specifically whether reading from snapshots is mutually exclusive? Your clarification would be incredibly helpful. 

It's not mutually exclusive to readers. You can regard it as a mechanism like MVCC, the snapshot won't be changed it's reading.

MaheshMadushan: Thank you. Got it.

torwig:
@git-hulk Are we waiting for additional test in this PR or it will be a separate one?

git-hulk:
> @git-hulk Are we waiting for additional test in this PR or it will be a separate one?

@torwig ohh, sorry for missing your previous comment. @MaheshMadushan Could you please add the edge cases as @torwig mentioned in this PR?

MaheshMadushan:
> > @git-hulk Are we waiting for additional test in this PR or it will be a separate one?
> 
> @torwig ohh, sorry for missing your previous comment. @MaheshMadushan Could you please add the edge cases as @torwig mentioned in this PR?

will add them asap.

src/types/redis_zset.cc
```diff
@@ -931,14 +931,6 @@ rocksdb::Status ZSet::RandMember(const Slice &user_key, int64_t command_count,
 }
 
 rocksdb::Status ZSet::Diff(const std::vector<Slice> &keys, MemberScores *members) {
-  std::vector<std::string> lock_keys;
```
MaheshMadushan: Hi @git-hulk , I have omitted the locking in the ZSet::Diff. is this what you had in mind?

git-hulk: Yes, I'm good with this change. We can improve this in the future by using the same snapshot between keys in RangeByScore.

tests/gocase/unit/type/zset/zset_test.go
```diff
@@ -1463,6 +1463,168 @@ func basicTests(t *testing.T, rdb *redis.Client, ctx context.Context, encoding s
...
+			{Score: 4, Member: "e"},
+		})
+		createZset(rdb, ctx, "zsetb", []redis.Z{
+			{Score: 1, Member: "b"},
+			{Score: 2, Member: "c"},
+			{Score: 4, Member: "f"},
+			{Score: 4, Member: "e"},
+		})
+		rdb.Del(ctx, "zsetc")
+		rdb.Del(ctx, "zsetc")
```
torwig: Why do you repeat the deletion?

torwig:
@MaheshMadushan @git-hulk  I'm also thinking about one optimization: when the source zset is empty - I guess in this case we can make early return and there is no sense to traverse all the others zsets. Is it correct?

git-hulk:
> @MaheshMadushan @git-hulk I'm also thinking about one optimization: when the source zset is empty - I guess in this case we can make early return and there is no sense to traverse all the others zsets. Is it correct?

Good point.

tests/gocase/unit/type/zset/zset_test.go
MaheshMadushan: Thanks for pointing out. This is a mistake. will fix this.

MaheshMadushan:
> @MaheshMadushan @git-hulk I'm also thinking about one optimization: when the source zset is empty - I guess in this case we can make early return and there is no sense to traverse all the others zsets. Is it correct?

Will fix this.

torwig (approved):

git-hulk (approved):

git-hulk:
@MaheshMadushan Thanks for your contribution.

MaheshMadushan:
@git-hulk @torwig Thank you for feedback.

PR #2024, opened 2024-01-16
Implement RESP3 set type for the set data structure

### LINKED ISSUE #1997: Add RESP3 set type

git-hulk:
This closes #1997


caipengbo:
I have a question about  last PR #2017, why put `MultiBulkString` to `Connection`? @git-hulk

git-hulk:
> I have a question about last PR #2017, why put `MultiBulkString` to `Connection`? @git-hulk

@caipengbo Because `MultiBulkString` should use the RESP3's null to represent the nil string if RESP3 was enabled. That said, all types related to RESP3 will move into `Connnection`.

torwig (dismissed):
LGTM.

BTW, the line:
` *output = conn->SetLen(3);`
made me stop and think for a while :) ("How do we set a length to a connection?")

git-hulk:
> LGTM.
> 
> BTW, the line: ` *output = conn->SetLen(3);` made me stop and think for a while :) ("How do we set a length to a connection?")
 
Yes, it's a bit confusing. Do you have any idea about this name? I'm not sure if it's better to append `Build` before those functions like `BuildSetLen()`, `BuildMultiBulkStrings()`

@torwig What do you think about this solution?

caipengbo:
> MultiString should use the RESP3's null to represent the nil string if RESP3 was enabled.

Why not pass a version flag to this function?

git-hulk:
> > MultiString should use the RESP3's null to represent the nil string if RESP3 was enabled.
> 
> Why not pass a version flag to this function?

It's okay to add the version flag, but moving them into connection can avoid misusing those functions. That said, use RESP2 or RESP3 should be related to connection instead of developers.

caipengbo:
> That said, use RESP2 or RESP3 should be related to connection instead of developers.

Got it !

torwig:
@git-hulk The `Build` prefix doesn't fit because the function doesn't build anything. I'm thinking of something like `SizeOfSet` similar to the `ArrayOfSet` function you introduced as well.

git-hulk:
> @git-hulk The `Build` prefix doesn't fit because the function doesn't build anything. I'm thinking of something like `SizeOfSet` similar to the `ArrayOfSet` function you introduced as well.

cool. SizeOfSet is better, I have updated it.

caipengbo (approved):
LGTM

torwig (approved):

src/server/redis_connection.cc
```diff
@@ -163,6 +163,15 @@ std::string Connection::MultiBulkString(const std::vector<std::string> &values,
   return result;
 }
 
+std::string Connection::ArrayOfSet(const std::vector<std::string> &elems) const {
```
PragmaTwice, line 166: IMHO I cannot understand the name.

It is just a **set**, not an **array of sets**. Instead, it is a **set of bulk strings**.

An **array of sets** is typed `Array<Set<T>>`.

Naming is important, and these confusing names can lead to bugs.

git-hulk: @PragmaTwice So do you suggest changing this name to SetOfBulkStrings? If yes, I will bring this fix in the next RESP3 PR.

PragmaTwice: Yeah that's better to me. ArrayOfSet seems like an array composed by sets.

git-hulk: Thanks for your good suggestion.

PR #2026, opened 2024-01-17
Add support of the command RENAME&RENAMENX 

### LINKED ISSUE #1951: Add support of the command RENAMENX
Refer to https://redis.io/commands/renamenx/

jihuayu:
close https://github.com/apache/kvrocks/issues/1951

### DIFF
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -32,6 +32,7 @@
 #include "server/server.h"
 #include "storage/iterator.h"
 #include "storage/redis_metadata.h"
+#include "storage/storage.h"
 #include "time_util.h"
 
 namespace redis {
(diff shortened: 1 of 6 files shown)


src/storage/redis_db.cc
```diff
@@ -690,4 +693,61 @@ Status WriteBatchLogData::Decode(const rocksdb::Slice &blob) {
...
+      // copy sub key
+      batch->Put(subkey_iter->ColumnFamilyHandle(), to_ikey, subkey_iter->Value());
+
+      if (type == kRedisZSet) {
+        std::string score_bytes = subkey_iter->Value().ToString();
+        score_bytes.append(from_ikey.GetSubKey().ToString());
+        // copy score key
+        std::string score_key =
+            InternalKey(new_ns_key, score_bytes, from_ikey.GetVersion(), storage_->IsSlotIdEncoded()).Encode();
+        batch->Put(storage_->GetCFHandle(engine::kZSetScoreColumnFamilyName), score_key, Slice());
```
jihuayu: Is there a cleaner way to avoid repetitive calls to the GetCFHandle function?
Or, this way is ok, as CPU branch prediction will help us deal with it.

src/storage/redis_db.cc
```diff
@@ -690,4 +693,61 @@ Status WriteBatchLogData::Decode(const rocksdb::Slice &blob) {
...
+      *ret = false;
+      return rocksdb::Status::OK();
+    }
+  }
+
+  if (key == new_key) return rocksdb::Status::OK();
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(type);
+  batch->PutLogData(log_data.Encode());
```
jihuayu, line 728: I don't know how WriteBatchLogData works, It is written ok?

src/storage/redis_db.cc
torwig: Sure. The one way is to just make it before the loop.

src/storage/redis_db.cc
git-hulk: Yes, that's good for this command.

src/storage/redis_db.cc
```diff
@@ -695,4 +698,65 @@ Status WriteBatchLogData::Decode(const rocksdb::Slice &blob) {
...
+    auto zset_score_cf = type == kRedisZSet ? storage_->GetCFHandle(engine::kZSetScoreColumnFamilyName) : nullptr;
+
+    for (subkey_iter->Seek(); subkey_iter->Valid(); subkey_iter->Next()) {
+      InternalKey from_ikey(subkey_iter->Key(), storage_->IsSlotIdEncoded());
+      std::string to_ikey =
+          InternalKey(new_ns_key, from_ikey.GetSubKey(), from_ikey.GetVersion(), storage_->IsSlotIdEncoded()).Encode();
+      // copy sub key
+      batch->Put(subkey_iter->ColumnFamilyHandle(), to_ikey, subkey_iter->Value());
+
+      if (type == kRedisZSet) {
```
git-hulk, line 751:
```suggestion
      // The ZSET type stores an extra score and member field inside `zset_score` column family
      // while compared to other composed data structures. The purpose is to allow to seek by score.
      if (type == kRedisZSet) {
```

git-hulk (commented):
Looks good to me. 

My only suggestion is to add a comment line to explain why we need to do another write operation for ZSET.

Thanks for your nice test suites for the RENAME command.

git-hulk (approved):

torwig (approved):

PR #2028, opened 2024-01-18
Add the support of RESP3 map type

### LINKED ISSUE #1996: Add RESP3 map type

git-hulk:
This PR mainly supports the RESP3 map type. But we found the FUNCTION LIST command
output now is an array of bulk strings like `["function_name","add","library":"lib1"]`.
To align with Redis's format which is an array of arrays that should be:
`[["function_name","add"], ["library":"lib1"]]`.

**We must mention the breaking change in the FUNCTION LIST command output in our next release notes.**

Most codes are changed in test cases to ensure RESP3 is working well.

This closes #1996

### DIFF
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -165,7 +165,7 @@ std::string Connection::MultiBulkString(const std::vector<std::string> &values,
 
 std::string Connection::SetOfBulkStrings(const std::vector<std::string> &elems) const {
   std::string result;
-  result += SizeOfSet(elems.size());
+  result += HeaderOfSet(elems.size());
   for (const auto &elem : elems) {
     result += BulkString(elem);
   }
@@ -176,7 +176,7 @@ std::string Connection::MapOfBulkStrings(const std::vector<std::string> &elems) 
   CHECK(elems.size() % 2 == 0);
 
   std::string result;
-  result += SizeOfMap(elems.size() / 2);
+  result += HeaderOfMap(elems.size() / 2);
   for (const auto &elem : elems) {
     result += BulkString(elem);
   }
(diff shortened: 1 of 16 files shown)


src/server/redis_connection.cc
```diff
@@ -172,6 +172,15 @@ std::string Connection::ArrayOfSet(const std::vector<std::string> &elems) const
   return result;
 }
 
+std::string Connection::MapOfBulkStrings(const std::vector<std::string> &elems) const {
+  std::string result;
```
PragmaTwice, line 178: Could you add an assertion here? e.g. `CHECK(elems.size() % 2 == 0)`

git-hulk: @PragmaTwice Sure, done. Thank you!

src/server/redis_connection.h
```diff
@@ -74,7 +74,12 @@ class Connection : public EvbufCallbackBase<Connection> {
   std::string SizeOfSet(T len) const {
     return protocol_version_ == RESP::v3 ? "~" + std::to_string(len) + CRLF : MultiLen(len);
   }
-  std::string ArrayOfSet(const std::vector<std::string> &elems) const;
+  std::string SetOfBulkStrings(const std::vector<std::string> &elems) const;
+  template <typename T, std::enable_if_t<std::is_integral_v<T>, int> = 0>
+  std::string SizeOfMap(T len) const {
```
PragmaTwice: From my perspective, the function name can be something like `StartMap` or `MapHeader`.

git-hulk: How about changing them to `StartOutputOfMap` or `StartMapOutput` due to  `conn->StartMap` feels like we will create a map inside the connection at first look.

And then change `[Set|Map]OfBulkStrings` to `OutputOf[Set|Map]`?

PragmaTwice: I don't think `Output` is a better idea. It seems like an IO operation, rather than a no-side-effect string operation.

If `StartMap` looks weird to you, maybe we can try `MapHeader`?

PragmaTwice (approved):

PR #2030, opened 2024-01-19
Run SonarQube scan in Github Actions to support coverage analysis

PragmaTwice:
Refer to https://docs.sonarsource.com/sonarcloud/advanced-setup/automatic-analysis/, there are generally two methods to use SonarCloud: automatic analysis and running analysis via CI.

We need to switch to the second mode to support to generate coverage data and report.

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -196,6 +196,8 @@ jobs:
             with_speedb: -DENABLE_SPEEDB=ON
 
     runs-on: ${{ matrix.os }}
+    env:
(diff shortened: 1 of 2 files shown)


PragmaTwice:
Due to GitHub Actions limits, the newly-added workflow will take effect after this PR is merged.

git-hulk (approved):
LGTM, cool!!

Just out of curious. Is SONARCLOUD_TOKEN pre-defined secret inside organization scope.

PragmaTwice:
Actually I don't know if it works, so let's try.

Issue #2022, opened 2024-01-15
Allow to use hyper clock cache to replace the LRU cache


git-hulk:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

The hyperclock cache was introduced in rocksdb in 7.7.3 and it may bring performance benefits for the block/row cache while compared to the LRU cache. This idea is mentioned by @PragmaTwice.

### Solution

Allow to enable the hyper cache via the configuration.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

JxLi0921:
hello! I have a question for this. Noticed that this issue has a `need benchmark` tag, is there a standard way to benchmark kvrocks? For example, does simlply running redis-benchmark on my computer is enough , or we need to do some other experiments on some platforms?

git-hulk:
@JxLi0921 Your local environment is enough. What we want to know is if the hyper clock cache can bring performance benefits while compared to LRU. And the benchmark result is also not a blocking item for merging PR.

mapleFU:
> Noticed that this issue has a need benchmark tag, is there a standard way to benchmark kvrocks?

Perhaps you can try ycsb in your local environment. This is a way we can follow: https://github.com/apache/kvrocks/pull/1424#issuecomment-1546521178

JxLi0921:
@git-hulk @mapleFU Thanks for your answers, got it! Could you please assign this to me? I would like to give it a try.

git-hulk:
@JxLi0921 assigned, thank you!

LiuYuHui:
It seems like that this issue had been solved?

git-hulk:
@LiuYuHui Sorry for forgetting to close this issue, it's done in #2031.

PR #2031, opened 2024-01-20
Add support of using hyper clock cache as the block cache

JxLi0921:
This is related to this issue #2022 . I will soon post my simple benchmark result for this PR below.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -571,6 +571,12 @@ rocksdb.block_cache_size 4096
 
 # Specify the type of cache used in the block cache.
 # Accept value: "lru", "hcc"
+# "lru" stands for the cache with the LRU(Least Recently Used) replacement policy.
+#
+# "hcc" stands for the Hyper Clock Cache, a lock-free cache alternative
+# that offers much improved CPU efficiency vs. LRU cache under high parallel
+# load or high contention.
+#
 # default lru
 rocksdb.block_cache_type lru
 
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -569,12 +569,22 @@ migrate-sequence-gap 10000
 # Default: 4096MB
 rocksdb.block_cache_size 4096
 
+# Specify the type of cache used in the block cache.
+# Accept value: "lru", "hcc"
+# default lru
+rocksdb.block_cache_type lru
+
 # A global cache for table-level rows in RocksDB. If almost always point
 # lookups, enlarging row cache may improve read performance. Otherwise,
 # if we enlarge this value, we can lessen metadata/subkey block cache size.
 #
 # Default: 0 (disabled)
 rocksdb.row_cache_size 0
+
+# Specify the type of cache used in the row cache.
+# Accept value: "lru", "hcc"
+# default lru
+rocksdb.row_cache_type lru
 
 # Number of open files that can be used by the DB.  You may need to
 # increase this if your database has a large working set. Value -1 means
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -581,11 +581,6 @@ rocksdb.block_cache_type lru
 # Default: 0 (disabled)
 rocksdb.row_cache_size 0
 
-# Specify the type of cache used in the row cache.
-# Accept value: "lru", "hcc"
-# default lru
-rocksdb.row_cache_type lru
-
 # Number of open files that can be used by the DB.  You may need to
 # increase this if your database has a large working set. Value -1 means
 # files opened are always kept open. You can estimate number of files based
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -155,10 +155,7 @@ rocksdb::Options Storage::InitRocksDBOptions() {
   }
 
   if (config_->rocks_db.row_cache_size) {
-    // Parameters for the rocksdb::LRUCacheOptions here are set in line with the default parameters for the
-    // rocksdb::NewLRUCache function, which is now deprecated
-    rocksdb::LRUCacheOptions lru_cache_options(config_->rocks_db.row_cache_size * MiB, -1, false, 0.5);
-    options.row_cache = lru_cache_options.MakeSharedCache();
+    options.row_cache = rocksdb::NewLRUCache(config_->rocks_db.row_cache_size * MiB);
   }
 
   options.enable_pipelined_write = config_->rocks_db.enable_pipelined_write;
@@ -264,9 +261,8 @@ Status Storage::Open(DBOpenMode mode) {
 
   std::shared_ptr<rocksdb::Cache> shared_block_cache;
 
-  if (config_->rocks_db.block_cache_type == rocksdb::PrimaryCacheType::kCacheTypeLRU) {
-    rocksdb::LRUCacheOptions lru_cache_options(block_cache_size, -1, false, 0.75);
-    shared_block_cache = lru_cache_options.MakeSharedCache();
+  if (config_->rocks_db.block_cache_type == BlockCacheType::kCacheTypeLRU) {
+    shared_block_cache = rocksdb::NewLRUCache(block_cache_size, -1, false, 0.75);
   } else {
     rocksdb::HyperClockCacheOptions hcc_cache_options(block_cache_size, 0);
     shared_block_cache = hcc_cache_options.MakeSharedCache();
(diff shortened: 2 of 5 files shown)


JxLi0921:
The script for pressing data into kvrocks: 
```
memtier_benchmark -t 4 -c 20 -h 127.0.0.1 -p 6666 --distinct-client-seed --key-minimum=1000000000 --key-maximum=10000000000 --random-data --key-prefix="" --data-size=50 -n 2000000 --pipeline=100 --command "mset key data key data key data key data key data key data key data key data key data key data"
``` 

The script for benchmarking: where the READ_RATIO is 0, 50, 90 or 100, respectively.
```
memtier_benchmark -t 4 -c 20 -s 127.0.0.1 -p 6666 --distinct-client-seed --key-minimum=1000000000 --key-maximum=10000000000 --key-prefix="" -n 10000 --command "get key" --command-ratio=READ_RATIO --command "set key value" --command-ratio=100-READ_RATIO
```

Here is the P99, P999 and P9999 latency with difference read ratios for the SET and GET command in the benchmark above.

SET: 
latency_SET

GET: 
latency_GET

wanghenshui:
> --data-size=50 -n 2 000 000

50 * 2,000,000 * 20(c) = 2,000,000,000

how much rocksdb sst data your make, 2G? please make at least 10G data

mapleFU:
I think the clock might has better concurrency and perhaps a bit worse cache hit rate. We can also report the cache hit rate etc for analyzing the performance here

src/storage/storage.cc
```diff
@@ -256,7 +259,14 @@ Status Storage::Open(DBOpenMode mode) {
     }
   }
 
-  std::shared_ptr<rocksdb::Cache> shared_block_cache = rocksdb::NewLRUCache(block_cache_size, -1, false, 0.75);
+  std::shared_ptr<rocksdb::Cache> shared_block_cache;
+
+  if (config_->rocks_db.block_cache_type == rocksdb::PrimaryCacheType::kCacheTypeLRU) {
+    shared_block_cache = rocksdb::NewLRUCache(block_cache_size, -1, false, 0.75);
```
mapleFU: nit: should we also use `rocksdb::LruCacheOptions` because `NewLRUCache` is deprecated?

mapleFU (commented):
+1 for this pr.
And maybe we can dive into benchmark result and find some reason of performance🤔

JxLi0921:
> > --data-size=50 -n 2 000 000
> 50 * 2,000,000 * 20(c) = 2,000,000,000
> how much rocksdb sst data your make, 2G? please make at least 10G data

@wanghenshui  Thanks for your review! The command I used to press the data into kvrocks is `mset key data * 10`, so maybe it's 20G instead of 2G? I am not convinced with my understanding, please correct me if I am wrong.

src/storage/storage.cc
JxLi0921: > nit: should we also use `rocksdb::LruCacheOptions` because `NewLRUCache` is deprecated?

thanks, fixed

JxLi0921:
> I think the clock might has better concurrency and perhaps a bit worse cache hit rate. We can also report the cache hit rate etc for analyzing the performance here

Thanks for your review! I would add these statistics later today.

mapleFU:
As for the statistics, you can refer `Server::GetRocksDBInfo` for help

kvrocks.conf
```diff
@@ -569,6 +569,11 @@ migrate-sequence-gap 10000
 # Default: 4096MB
 rocksdb.block_cache_size 4096
 
+# Specify the type of cache used in the block cache.
+# Accept value: "lru", "hcc"
+# default lru
+rocksdb.block_cache_type lru
```
PragmaTwice: Could you explain the meaning of lru and hcc for users in the comment>

src/storage/storage.cc
```diff
@@ -152,9 +153,14 @@ rocksdb::Options Storage::InitRocksDBOptions() {
       options.compression_per_level[i] = config_->rocks_db.compression;
     }
   }
+
   if (config_->rocks_db.row_cache_size) {
-    options.row_cache = rocksdb::NewLRUCache(config_->rocks_db.row_cache_size * MiB);
+    // Parameters for the rocksdb::LRUCacheOptions here are set in line with the default parameters for the
+    // rocksdb::NewLRUCache function, which is now deprecated
+    rocksdb::LRUCacheOptions lru_cache_options(config_->rocks_db.row_cache_size * MiB, -1, false, 0.5);
```
PragmaTwice: Could we add some named constants for these default arguments (e.g. -1, false, 0.5)?

src/storage/storage.cc
```diff
@@ -256,7 +262,15 @@ Status Storage::Open(DBOpenMode mode) {
...
   }
 
-  std::shared_ptr<rocksdb::Cache> shared_block_cache = rocksdb::NewLRUCache(block_cache_size, -1, false, 0.75);
+  std::shared_ptr<rocksdb::Cache> shared_block_cache;
+
+  if (config_->rocks_db.block_cache_type == rocksdb::PrimaryCacheType::kCacheTypeLRU) {
+    rocksdb::LRUCacheOptions lru_cache_options(block_cache_size, -1, false, 0.75);
+    shared_block_cache = lru_cache_options.MakeSharedCache();
+  } else {
+    rocksdb::HyperClockCacheOptions hcc_cache_options(block_cache_size, 0);
```
PragmaTwice: ditto for `0` here.

src/storage/storage.cc
JxLi0921: > because `NewLRUCache` is deprecated?

Oh maybe for now we have to still use `NewLRUCache` for being compatible with Speedb. Speedb currently does not offer the `LRUCacheOptions::MakeSharedCache` API like Rocksdb.

src/storage/storage.cc
JxLi0921: I have did this, but I am not sure whether the names of my constants are proper.

git-hulk:
@JxLi0921 Great, thanks for your PR. I'm not sure if the QPS throughput is close between Hyper Clock Cache and LRU since I didn't see this in previous report.

JxLi0921:
Sorry for late update. I found that the way I use `memtier_benchmark` is totally wrong before. 

Here is my current script for putting data and running benchmark: 

Putting data into kvrocks:
```
memtier_benchmark -t 8 -c 20 -h 127.0.0.1 -p 6666 --distinct-client-seed --key-minimum=1000000000 --key-maximum=10000000000 --random-data --key-prefix="" --data-size
=50 -n 2000000 --pipeline=100 --command "mset __key__ __data__ __key__ __data__ __key__ __data__  __key__ __data__ __key__ __data__ __key__ __data__ __key__ __data__ __key__ 
__data__ __key__ __data__ __key__ __data__"
```

I failed to run the script above to complete because I am almost run out of disk space. After running the script above for about 20% of progress, the keyspace statistics of kvrocks is as follows. 
```
#  Keyspace
# Last DBSIZE SCAN time: Wed Jan 24 12:04:59 2024
db0:keys=341377306,expires=0,avg_ttl=0,expired=0
sequence:615748420
used_db_size:18388474802
max_db_size:0
used_percent: 0%
 ```

Running benchmark:

```
memtier_benchmark -t 4 -c 20 -s 127.0.0.1 -p 6666 --distinct-client-seed --key-minimum=1000000000  --key-maximum=10000000000 --key-prefix="" -n 50000 --ratio={100-read_ratio}:{read_ratio}
```

where the `read_ratio` above is one of 0, 50, 90 and 100.

Result:

GET Latency: 

latency_GET

SET Latency:

latency_SET

GET Ops/sec

ops_GET


SET Ops/sec

ops_SET


Overall(Get+Set) Ops/sec

ops_Overall

According to the results listed above, it seems that HCC does outperform LRU for latency and throughput in this simple benchmark.

Block Cache Hit Ratio:

cacheHit_SET

I'm not sure whether it's normal to have the block cache hit ratios to be almost 100. Should I use a smaller block cache size? I used 4096MB block cache size in this benchmark, which is kvrocks' default configuration.

mapleFU (approved):
Thanks for the investigate. I may go for this this weekend. At lease we can first checkin this patch?

Also @JxLi0921 would you mind tell the CPU and disk you're using?

git-hulk (approved):
Cool, LGTM

JxLi0921:
> would you mind tell the CPU and disk you're using?

@mapleFU 

CPU: AMD Ryzen 6800H
Disk: SAMSUNG MZVL2512HCJQ-00B00

PR #2032, opened 2024-01-20
Make the result of SRANDMEMBER random

### LINKED ISSUE #2014: Improve the implementation of SRANDMEMBER
### Search before asking

### Motivation

Currently it just returns the first N members, which is not actually random.

We can improve it just like the implementation of HRANDMEMBER.
 (Or if you have better ideas for random, feel free to implement them)

### Solution

_No response_

wsehjk:
This pr is to resolve #2014.
I use **Reservoir Sampling**. When the number we want to select is larger than that of elements in the set, all elements are selected

### DIFF
--- a/src/types/redis_set.cc
+++ b/src/types/redis_set.cc
@@ -247,8 +247,8 @@ rocksdb::Status Set::Take(const Slice &user_key, std::vector<std::string> *membe
       batch->Delete(key);
     }
   }
-  if (pop && n > 0) {
-    metadata.size -= n;
+  if (pop && !iter_keys.empty()) {
+    metadata.size -= iter_keys.size();
     std::string bytes;
     metadata.Encode(&bytes);
     batch->Put(metadata_cf_handle_, ns_key, bytes);
--- a/src/types/redis_set.cc
+++ b/src/types/redis_set.cc
@@ -23,6 +23,7 @@
 #include <map>
 #include <memory>
 #include <optional>
+#include <random>
 
 #include "db_util.h"
 
@@ -222,12 +223,29 @@ rocksdb::Status Set::Take(const Slice &user_key, std::vector<std::string> *membe
   rocksdb::Slice upper_bound(next_version_prefix);
   read_options.iterate_upper_bound = &upper_bound;
 
+  std::vector<std::string> iter_keys;
+  iter_keys.reserve(count);
+  std::random_device rd;
+  std::mt19937 gen(rd());
   auto iter = util::UniqueIterator(storage_, read_options);
   for (iter->Seek(prefix); iter->Valid() && iter->key().starts_with(prefix); iter->Next()) {
-    InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+    ++n;
+    if (n <= count) {
+      iter_keys.push_back(iter->key().ToString());
+    } else {  // n > count
+      std::uniform_int_distribution<> distrib(0, n - 1);
+      int random = distrib(gen);  // [0,n-1]
+      if (random < count) {
+        iter_keys[random] = iter->key().ToString();
+      }
+    }
+  }
+  for (Slice key : iter_keys) {
+    InternalKey ikey(key, storage_->IsSlotIdEncoded());
     members->emplace_back(ikey.GetSubKey().ToString());
-    if (pop) batch->Delete(iter->key());
-    if (++n >= count) break;
+    if (pop) {
+      batch->Delete(key);
+    }
   }
   if (pop && n > 0) {
     metadata.size -= n;


src/types/redis_set.cc
```diff
@@ -222,12 +223,26 @@ rocksdb::Status Set::Take(const Slice &user_key, std::vector<std::string> *membe
...
 
+  std::vector<std::string> vec;
+  vec.reserve(count);
   auto iter = util::UniqueIterator(storage_, read_options);
   for (iter->Seek(prefix); iter->Valid() && iter->key().starts_with(prefix); iter->Next()) {
-    InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+    if (n < count) {
+      vec.push_back(iter->key().ToString());
+    } else { // n >= count
+      int random = rand()%n;
```
mapleFU: Better use rand in https://en.cppreference.com/w/cpp/numeric/random

mapleFU: Also, should this be `rand() % count` ?

jihuayu:
> This pr is to solve https://github.com/apache/kvrocks/issues/2014.

Hi @wsehjk, you can use resolve #2014 instead of solve #2014. It will make GitHub auto close issue when PR has been merged.

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PragmaTwice:
Please use `./x.py format` to format your code, otherwise the CI cannot pass.

src/types/redis_set.cc
```diff
@@ -222,12 +223,26 @@ rocksdb::Status Set::Take(const Slice &user_key, std::vector<std::string> *membe
   rocksdb::Slice upper_bound(next_version_prefix);
   read_options.iterate_upper_bound = &upper_bound;
 
+  std::vector<std::string> vec;
+  vec.reserve(count);
```
PragmaTwice: The name `vec` here seems meaningless (any vector can be named `vec`). Could you rename it?

src/types/redis_set.cc
```diff
@@ -222,12 +223,29 @@ rocksdb::Status Set::Take(const Slice &user_key, std::vector<std::string> *membe
...
+  std::mt19937 gen(rd());
   auto iter = util::UniqueIterator(storage_, read_options);
   for (iter->Seek(prefix); iter->Valid() && iter->key().starts_with(prefix); iter->Next()) {
-    InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+    ++n;
+    if (n <= count) {
+      iter_keys.push_back(iter->key().ToString());
+    } else {  // n > count
+      std::uniform_int_distribution<> distrib(0, n - 1);
+      int random = distrib(gen);  // [0,n-1)
```
mapleFU:
```suggestion
      int random = distrib(gen);  // [0,n-1]
```

mapleFU:
Though a different sampling algorithm is used, general LGTM

src/common/status.h
```diff
@@ -168,7 +168,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> :
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
-      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
+      : BaseType((typename StringInStatusOr<U>::BaseType&&)(std::move(v))) {}
```
PragmaTwice: Could you revert this change?

jihuayu:
You can enable the GitHub action in your repo (It is free). So that you can find the CI failing issues earlier and more easily.

https://github.com/wsehjk/kvrocks/actions

src/common/status.h
wsehjk: It's the result of `./x.py format`

git-hulk: @wsehjk The format is introduced by clang-format-16 and Kvrocks is now using clang-format-14, so you can change the clang-format to 14 to make the CI happy.

wsehjk: Ok, i will reinstall clang-format-14

jihuayu:
It seems like the test run fails. Fix it, victory is within sight.

By the way, you don't need to use force push to make PR only one commit; simply use the Update branch Button is ok
image

wsehjk:
> It seems like the test run fails. Fix it, victory is within sight.
> 
> By the way, you don't need to use force push to make PR only one commit; simply use the Update branch Button is ok image

This is because the maintainer of last repo i have contributed required pr to be one commit, i think it still makes sense here.
Everytime `dev` branch lags behind the `unstable` branch， I run`git pull` and ` git rebase unstable dev` to catch up, then I have to use `force-push`....

git-hulk (approved):

mapleFU (approved):
LGTM

Issue #2007, opened 2024-01-12
Implement WAL iterator for migrating by WAL logs


git-hulk:

caipengbo:
In https://github.com/apache/kvrocks/pull/2004, we implement the unified key-value iterators, I want to change data to provide a unified iterators. 

For incremental data, we need to call `rocksdb::WALIter` and then implement the `rocksdb::WriteBatch::Handler` interface. 

I want to provide an interface that behaves similarly to `engine::DBIterator`: 

https://github.com/apache/kvrocks/blob/c772109516a6698f8e74d50c1a43bb1c02a697c5/src/storage/iterator.h#L52-L81. 

It will wrap `rocksdb::WALIter` and can return different value types. It is possible to implement their processing logic depending on the concrete type.

HDYT @git-hulk

git-hulk:
@caipengbo Sounds good to me.

PR #2040, opened 2024-01-22
Implement the unify WAL iterator

caipengbo:
Related to https://github.com/apache/kvrocks/issues/2007

Implement  the unify WAL iterator for kvrocks, similar to `DBIterator`. It will wrap rocksdb WAL Iterator and can return different item types. It is possible to implement their processing logic depending on the concrete type.

### DIFF
--- a/src/storage/iterator.cc
+++ b/src/storage/iterator.cc
@@ -25,7 +25,7 @@
 #include "db_util.h"
 
 namespace engine {
-DBIterator::DBIterator(Storage* storage, rocksdb::ReadOptions read_options, int slot)
+DBIterator::DBIterator(Storage *storage, rocksdb::ReadOptions read_options, int slot)
     : storage_(storage), read_options_(std::move(read_options)), slot_(slot) {
   metadata_cf_handle_ = storage_->GetCFHandle(kMetadataColumnFamilyName);
   metadata_iter_ = util::UniqueIterator(storage_->NewIterator(read_options_, metadata_cf_handle_));
@@ -80,7 +80,7 @@ void DBIterator::Reset() {
   if (metadata_iter_) metadata_iter_.reset();
 }
 
-void DBIterator::Seek(const std::string& target) {
+void DBIterator::Seek(const std::string &target) {
   if (!metadata_iter_) return;
 
   // Iterate with the slot id but storage didn't enable slot id encoding
@@ -112,7 +112,7 @@ std::unique_ptr<SubKeyIterator> DBIterator::GetSubKeyIterator() const {
   return std::make_unique<SubKeyIterator>(storage_, read_options_, type, std::move(prefix));
 }
 
-SubKeyIterator::SubKeyIterator(Storage* storage, rocksdb::ReadOptions read_options, RedisType type, std::string prefix)
+SubKeyIterator::SubKeyIterator(Storage *storage, rocksdb::ReadOptions read_options, RedisType type, std::string prefix)
     : storage_(storage), read_options_(std::move(read_options)), type_(type), prefix_(std::move(prefix)) {
   if (type_ == kRedisStream) {
     cf_handle_ = storage_->GetCFHandle(kStreamColumnFamilyName);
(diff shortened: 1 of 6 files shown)


tests/cppunit/iterator_test.cc
```diff
@@ -363,4 +365,149 @@ TEST_F(SlotIteratorTest, LiveKeys) {
...
+      }
+      default:
+        FAIL() << "Unexpected wal item type" << uint8_t(item.type);
+    }
+  }
+  ASSERT_TRUE(std::equal(expected_put_keys.begin(), expected_put_keys.end(), put_keys.begin()));
+  ASSERT_TRUE(std::equal(expected_delete_keys.begin(), expected_delete_keys.end(), delete_keys.begin()));
+}
+
+TEST_F(WALIteratorTest, ComplexType) {
```
caipengbo: I wanted to implement all types of single tests, but I found that many of them were repetitive and did not help for improving test coverage, so I chose two categories of data types (`string` and `hash`).

git-hulk: @caipengbo Perhaps adding all types except BloomFilter is to ensure it works well if we change some of them.

PragmaTwice:
Could you run `./x.py format` to make the CI pass?

If you've run it but the CI still fail in format check, please use clang-format with exact version 14.

caipengbo:
Ok, my machine clang-format is version 17, let me install version 14.

src/storage/iterator.cc
```diff
@@ -164,4 +164,124 @@ void SubKeyIterator::Reset() {
   if (iter_) iter_.reset();
 }
 
+rocksdb::Status WALBatchExtractor::PutCF(uint32_t column_family_id, const Slice &key, const Slice &value) {
+  if (slot_ != -1) {
+    if (slot_ != ExtractSlotId(key)) {
+      return rocksdb::Status::OK();
+    }
+  }
```
git-hulk:
```suggestion
  if (slot_ != -1 && slot_ != ExtractSlotId(key)) {
    return rocksdb::Status::OK();
  }
```

src/storage/iterator.cc
```diff
@@ -164,4 +164,124 @@ void SubKeyIterator::Reset() {
...
+  items_.emplace_back(WALItem::Type::kTypePut, column_family_id, key.ToString(), value.ToString());
+  return rocksdb::Status::OK();
+}
+
+rocksdb::Status WALBatchExtractor::DeleteCF(uint32_t column_family_id, const rocksdb::Slice &key) {
+  if (slot_ != -1) {
+    if (slot_ != ExtractSlotId(key)) {
+      return rocksdb::Status::OK();
+    }
+  }
```
git-hulk:
```suggestion
  if (slot_ != -1 && slot_ != ExtractSlotId(key)) {
    return rocksdb::Status::OK();
  }
```

src/storage/iterator.cc
```diff
@@ -164,4 +164,124 @@ void SubKeyIterator::Reset() {
...
+  items_.emplace_back(WALItem::Type::kTypeDelete, column_family_id, key.ToString(), std::string{});
+  return rocksdb::Status::OK();
+}
+
+rocksdb::Status WALBatchExtractor::DeleteRangeCF(uint32_t column_family_id, const rocksdb::Slice &begin_key,
+                                                 const rocksdb::Slice &end_key) {
+  items_.emplace_back(WALItem::Type::kTypeDeleteRange, column_family_id, begin_key.ToString(), end_key.ToString());
```
git-hulk, line 185: Should we check if the begin_key and end_key are in the same slot when the slot != -1

src/storage/iterator.cc
```diff
@@ -164,4 +164,124 @@ void SubKeyIterator::Reset() {
...
+
+void WALBatchExtractor::Clear() { items_.clear(); }
+
+WALBatchExtractor::Iter WALBatchExtractor::GetIter() { return Iter(&items_); }
+
+bool WALBatchExtractor::Iter::Valid() { return items_ && cur_ < items_->size(); }
+
+void WALBatchExtractor::Iter::Next() { cur_++; }
+
+WALItem WALBatchExtractor::Iter::Value() { return (*items_)[cur_]; }
```
git-hulk: Should we check if the `cur_` is out of the index here?

src/storage/iterator.cc
caipengbo: This range can contain more than one slot, and checking the slot might lose some information, so I want to keep the whole range and let the caller deal with it.

git-hulk: got it, makes sense.

git-hulk (approved):
LGTM

PR #2043, opened 2024-01-22
Use condition argument of cv.wait instead of loop

PragmaTwice:
See changes.


tisonkun (approved):

torwig (approved):

git-hulk (approved):

mapleFU (approved):
So they have same meaning, just a style enhancement?

PragmaTwice:
> So they have same meaning, just a style enhancement?

It can avoid spurious awakenings.

PR #2047, opened 2024-01-24
Ignore max-db-size limit when deleting data or writing aux informations

caipengbo:
When we delete data in kvrocks, we will write delete makers to rocksdb. When our data size reaches `max-db-size`, we can't delete data. 

For operations that delete data, we should ignore the `max-db-size` limit and allow the deletion to proceed, so we can clean up the data.

In addition, for some aux information(written into `propagate` column family), we should also ignore `max-db-size`.


jihuayu:
Will this logic confuse users?
For example, when the data size reaches the max-db-size, if the user deletes a large amount of data, it will cause the db size to significantly exceed the db-size they have configured.

caipengbo:
> Will this logic confuse users?

@jihuayu There is also a risk of confusion, which we can explain in the configuration file. In addition, these delete makers will be eliminated during compaction, and the data size will decrease. 

If `max-db-size` is not reached, users deleting large amounts of data will also cause disk explosion, this can also bother users.

git-hulk:
@caipengbo The better way is to check if it reached the max db size if it's a write command before running. And add an extra flag to identify if the write command is allowed to run even though reached the limit of db size. What you do think about this?

caipengbo:
> The better way is to check if it reached the max db size if it's a write command. And add an extra flag to identify if the write command is allowed to run even though reached the limit of db size.

That sounds better, I'll try to modify it.

jihuayu:
> @caipengbo The better way is to check if it reached the max db size if it's a write command before running. And add an extra flag to identify if the write command is allowed to run even though reached the limit of db size. What you do think about this?

I think we can make some reserved db space, such as:
- When free db space is below 100MB, only delete commands write and system write can work.
- When free db space is below 10MB, only system write can work.

What do you think? @git-hulk @caipengbo

caipengbo:
Your idea is similar to what I did on our personalized version kvrocks. @jihuayu 

 It's just that we don't want to save disk space(we reserve some disk space), we want to **avoid OOM**, because if we delete a lot of data, we can create a very large WAL, which caches a lot of data in memory and very easy to cause OOM(we have encountered it), and even worse, we can't restart our program, because we need to replay the WAL at startup. At this point, all unflushed data will be loaded into memory, again leading to an OOM!

So we find all the places in the program where kvrocks can generate a large WAL, check it, and if it's larger than `max-write-batch-size`, we return an error so the user can't do it.
> This is a feature I plan to implement in some subsequent PRs (about memory control).

src/storage/storage.h
```diff
@@ -171,6 +171,8 @@ class Storage {
   void PurgeOldBackups(uint32_t num_backups_to_keep, uint32_t backup_max_keep_hours);
   uint64_t GetTotalSize(const std::string &ns = kDefaultNamespace);
   void CheckDBSizeLimit();
+  bool ReachedDBSizeLimit() { return db_size_limit_reached_; }
+  void SetDBSizeLimit(bool limit) { db_size_limit_reached_ = limit; }
```
caipengbo, line 186: This flag will be modified by flush or compact, but for now it's only for testing (debug), so I think it's good enough.

src/storage/storage.h
```diff
@@ -243,7 +245,7 @@ class Storage {
   Config *config_ = nullptr;
   std::vector<rocksdb::ColumnFamilyHandle *> cf_handles_;
   LockManager lock_mgr_;
-  bool db_size_limit_reached_ = false;
+  std::atomic<bool> db_size_limit_reached_{false};
```
jihuayu, line 259: Why change it to atomic? I don't quite understand.

caipengbo: This variable will be modified by multiple threads during compact and flush.

src/commands/cmd_list.cc
```diff
@@ -861,9 +861,9 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandBLPop>("blpop", -3, "write no-script"
                         MakeCmdAttr<CommandLPush>("lpush", -3, "write", 1, 1, 1),
                         MakeCmdAttr<CommandLPushX>("lpushx", -3, "write", 1, 1, 1),
                         MakeCmdAttr<CommandLRange>("lrange", 4, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandLRem>("lrem", 4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandLRem>("lrem", 4, "write no-dbsize-check", 1, 1, 1),
                         MakeCmdAttr<CommandLSet>("lset", 4, "write", 1, 1, 1),
-                        MakeCmdAttr<CommandLTrim>("ltrim", 4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandLTrim>("ltrim", 4, "write no-dbsize-check", 1, 1, 1),
                         MakeCmdAttr<CommandLMPop>("lmpop", -4, "write", CommandLMPop::keyRangeGen),
                         MakeCmdAttr<CommandRPop>("rpop", -2, "write", 1, 1, 1),
```
PragmaTwice, line 868: Are these commands also suitable for `no-dbsize-check`?

Maybe we need to carefully cover all such commands as much as possible.

caipengbo: I didn't add this flag for any of these commands. The way I think about it is that there might be a lot of fields for this type(list, zset...),  and the db space release will be slow if the user removes them one by one. I want to guide the user to remove many elements(remove a range) at once. So I only added trim, etc.

PragmaTwice (approved):

jihuayu (approved):

git-hulk (approved):

PR #2048, opened 2024-01-24
Fix sonar action can't get PR number

jihuayu:
We only save PR info in the CI action pull_request event.

It was tested under https://github.com/jihuayu/kvrocks-test

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -481,3 +481,4 @@ jobs:
           fi
       - name: Sentinel
         run: true
+ 
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -334,20 +334,7 @@ jobs:
         with:
           name: sonarcloud-data
           path: ${{ env.SONARCLOUD_OUTPUT_DIR }}
-          
-      - name: Save PR info
-        if: ${{ (github.event_name == 'pull_request') && matrix.sonarcloud }}
-        run: |
-          mkdir -p ./pr
-          echo "number=${{ github.event.number }}" >> ./pr/info
-          
-      - name: Upload PR info
-        uses: actions/upload-artifact@v3
-        if: ${{ matrix.sonarcloud }}
-        with:
-          name: pr
-          path: pr/
-          
+
   check-docker:
     name: Check Docker image
     needs: [precondition, check-and-lint, check-typos]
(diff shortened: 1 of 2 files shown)


.github/workflows/sonar.yaml
```diff
@@ -39,30 +39,45 @@ jobs:
...
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                artifact_id: matchArtifact.id,
+                archive_format: 'zip',
+              });
+              let fs = require('fs');
+              fs.writeFileSync(`${process.env.GITHUB_WORKSPACE}/${name}.zip`, Buffer.from(download.data));
+            }
+            await getData("sonarcloud-data");
+            await getData("pr");
```
git-hulk: May it get the wrong PR number while two PRs are running at the same time?

jihuayu: No, this is a Github action doc example.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow

.github/workflows/kvrocks.yaml
```diff
@@ -327,7 +327,20 @@ jobs:
...
+        run: |
+          mkdir -p ./pr
+          echo "number=${{ github.event.number }}" >> ./pr/info
+          
+      - name: Upload PR info
+        uses: actions/upload-artifact@v3
+        if: ${{ matrix.sonarcloud }}
+        with:
+          name: pr
+          path: pr/
```
PragmaTwice: You can just add a new file to `${{ env.SONARCLOUD_OUTPUT_DIR }}`.

.github/workflows/sonar.yaml
git-hulk: OK, thanks for your reference. That said, this artifact is only visible in the current run.

.github/workflows/kvrocks.yaml
jihuayu: You are right, Thank you. I will do it after #2049 has been merged.

.github/workflows/kvrocks.yaml
```diff
@@ -480,4 +480,4 @@ jobs:
             exit 1
           fi
       - name: Sentinel
-        run: true
+        run: true
```
git-hulk: Keep the empty line at the end of the file.
```suggestion
        run: true
 
```

git-hulk (commented):
Looks good to me. Thanks for your great efforts.

git-hulk (approved):

PR #2049, opened 2024-01-24
Dump github event information in CI

PragmaTwice:
cc @jihuayu


jihuayu (approved):
LGTM.
Do you mean we can parser this JSON in sonar action and get more information?

git-hulk (approved):

PragmaTwice:
> LGTM. Do you mean we can parser this JSON in sonar action and get more information?

Yeah. You can use some cli tools like `jq`, which is already installed.

PR #2051, opened 2024-01-25
Fix crash in zset store getkeys, fix zdiff/bzmpop range, add tests

enjoy-binbin:
These following cases will crash the server, the reason is that
the index of numkeys is wrong:
```
command getkeys zdiffstore dst 2 src1 src2
command getkeys zinterstore dst 2 src1 src2
command getkeys zunionstore dst 2 src1 src2
```

These following getkeys output is wrong:
```
> command getkeys zdiff 2 key1 key2
1) "key1"
2) "key2"
3) (nil)

> command getkeys bzmpop 0 2 key1 key2
1) "key1"
```

These are ok:
```
command getkeys zinter 2 key1 key2
command getkeys zunion 2 key1 key2
command getkeys sintercard 2 key1 key2
command getkeys zintercard 2 key1 key2
command getkeys zmpop 2 key1 key2
command getkeys lmpop 2 key1 key2
command getkeys blmpop 0 2 key1 key2
```

However, at present, there is still a problem with our zset store.
We do not support returning dst key, but let's do it later...
```
127.0.0.1:6379> command getkeys zinterstore dst 2 src1 src2
1) "dst"
2) "src1"
3) "src2"

127.0.0.1:6666> command getkeys zinterstore dst 2 src1 src2
1) "src1"
2) "src2"
```


git-hulk (approved):
LGTM, thanks

jihuayu (approved):
LGTM too, thanks.

jihuayu:
I think we need to inform developers that in versions of kvrocks before 2.8, some commands may cause server crashes.
Can we add this to the release log and [supported-commands desc](https://kvrocks.apache.org/docs/supported-commands)?

git-hulk:
@jihuayu Thanks for your warm reminder. Yes we should mention this in our next release note.

PR #2055, opened 2024-01-26
Use CommandKeyRangeVecGen to support retrieving store keys

enjoy-binbin:
For store key related commands, we cannot extract the store
key, see #2051. This PR use CommandKeyRangeVecGen to support
retrieving store keys.

This PR handle these store key commands:
- zunionstore
- zinterstore
- zdiffstore
- georadius
- georadiusbymember
- geosearchstore

before:
```
> command getkeys zinterstore dst 2 src1 src2
1) "src1"
2) "src2"
> command getkeys GEORADIUS src 1 1 1 km store dst
1) "src"
> command getkeys GEORADIUSBYMEMBER src member radius m store dst
1) "src"
> command getkeys GEOSEARCHSTORE dst src frommember member byradius 10 m
1) "dst"
```

after:
```
> command getkeys zinterstore dst 2 src1 src2
1) "dst"
2) "src1"
3) "src2"
> command getkeys GEORADIUS src 1 1 1 km store dst
1) "src"
2) "dst"
> command getkeys GEORADIUSBYMEMBER src member radius m store dst
1) "src"
2) "dst"
> command getkeys GEOSEARCHSTORE dst src frommember member byradius 10 m
1) "dst"
2) "src"
```

### DIFF
--- a/src/commands/cmd_geo.cc
+++ b/src/commands/cmd_geo.cc
@@ -344,7 +344,7 @@ class CommandGeoRadius : public CommandGeoBase {
     return redis::Array(list);
   }
 
-  static CommandKeyRange Range(const std::vector<std::string> &args) {
+  static std::vector<CommandKeyRange> Range(const std::vector<std::string> &args) {
     int store_key = 0;
 
     // Check for the presence of the stored key in the command args.
@@ -358,7 +358,10 @@ class CommandGeoRadius : public CommandGeoBase {
       }
     }
 
-    return {1, 1, 1, store_key};
+    if (store_key > 0) {
+      return {{1, 1, 1}, {store_key, store_key, 1}};
+    }
+    return {{1, 1, 1}};
   }
 
  protected:
@@ -622,6 +625,10 @@ class CommandGeoSearchStore : public CommandGeoSearch {
     return Status::OK();
   }
 
+  static std::vector<CommandKeyRange> Range(const std::vector<std::string> &args) {
+    return {{1, 1, 1}, {2, 2, 1}};
+  }
+
  private:
   bool store_distance_ = false;
   std::string store_key_;
@@ -665,7 +672,7 @@ class CommandGeoRadiusByMember : public CommandGeoRadius {
     return Status::OK();
   }
 
-  static CommandKeyRange Range(const std::vector<std::string> &args) {
+  static std::vector<CommandKeyRange> Range(const std::vector<std::string> &args) {
     int store_key = 0;
 
     // Check for the presence of the stored key in the command args.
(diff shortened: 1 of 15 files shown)


git-hulk (dismissed):
LGTM, good catch.

src/commands/commander.h
```diff
@@ -93,18 +93,22 @@ class CommanderWithParseMove : Commander {
...
   int last_key;
 
-  // step length of key position
+  // step length of key (non-store key) position
   // e.g. key step 2 means "key other key other ..." sequence
   int key_step;
+
+  // index of the store key in command tokens
+  // 0 stands for no store key, since the first index of command arguments is command name
+  int store_key;
```
jihuayu: If I'm a new developer, I'm not quite sure what this variable is used for. When do I need to set this variable? 
Could you please add some comments to clarify?

jihuayu (dismissed):
LGTM!

PragmaTwice (changes requested):
I don't think we need to distinguish between store and non-store keys. Could you just use `CommandKeyRangeVecGen` which is already well-designed for such a use case?

enjoy-binbin:
Can you give an example of how to use key_range_vec_gen? I don't seem to see it being used anywhere in the code base

enjoy-binbin:
for example, zunionstore, two range:
- store key: {1, 1, 1}
- non-store key: {3, 2 + num_key, 1}

something like this?

PragmaTwice:
> for example, zunionstore, two range:
> 
> * store key: {1, 1, 1}
> * non-store key: {3, 2 + num_key, 1}
> 
> something like this?

Yeah that's right. It is used to generate multiple key ranges at runtime.

PragmaTwice (approved):

src/commands/cmd_geo.cc
```diff
@@ -604,6 +624,8 @@ class CommandGeoSearchStore : public CommandGeoSearch {
     return Status::OK();
   }
 
+  static std::vector<CommandKeyRange> Range(const std::vector<std::string> &args) { return {{1, 1, 1}, {2, 2, 1}}; }
```
PragmaTwice, line 627: Maybe we can just return `{1,2,1}` here?

enjoy-binbin: yes, this will work too, personally i want to separate them (more explicit?)
but i am ok with both of them, i can change it if you ask

jihuayu (approved):

src/commands/cmd_geo.cc
PragmaTwice: fine to me.

enjoy-binbin:
Next time can we try to use the top comment (if it is useful) as a squash commit message and bring it into the commit?

This is useful for viewing the context of changes based on commit messages without having to open a web PR page.

PragmaTwice:
> Next time can we try to use the top comment (if it is useful) as a squash commit message and bring it into the commit?
> 
> This is useful for viewing the context of changes based on commit messages without having to open a web PR page.

Sure! Sorry for not caring about it.

PR #2066, opened 2024-01-28
Initialize metadata and tag indexing encoding for RediSearch

### LINKED ISSUE #2065: Add search metadata and tag field metadata encoding for RediSearch
Refer to https://redis.io/docs/interact/search-and-query/advanced-concepts/tags/.

PragmaTwice:
It close #2065.

The current encoding design:
```
search data type metadata:

key (index name) -> flag | expire | version | size | on_data_type (HASH or JSON)

prefixes encoding:

key (index name) | PREFIXES -> prefix1 prefix2 ...

tag field metadata encoding:

key (index name) | TAG_FIELD_META | field name -> separator | case sensitive

tag field index encoding:

key (index name) | TAG_FIELD | field name | tag | key -> (nil)

```

Refs:
https://redis.io/docs/interact/search-and-query/
https://redis.io/docs/interact/search-and-query/advanced-concepts/tags/

### DIFF
--- a/src/types/redis_search_encoding.h
+++ b/src/types/redis_search_encoding.h
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#pragma once
+
+#include <encoding.h>
+#include <storage/redis_metadata.h>
+
+namespace redis {
+
+inline constexpr auto kErrorInsufficientLength = "insufficient length while decoding metadata";
+
+enum class SearchSubkeyType : uint8_t {
+  // search global metadata
+  PREFIXES = 1,
+
+  // field metadata for different types
+  TAG_FIELD_META = 64 + 1,
+
+  // field indexing for different types
+  TAG_FIELD = 128 + 1,
+};
+
+inline std::string ConstructSearchPrefixesSubkey() { return {(char)SearchSubkeyType::PREFIXES}; }
+
+struct SearchPrefixesMetadata {
+  std::vector<std::string> prefixes;
+
+  void Encode(std::string *dst) const {
+    for (const auto &prefix : prefixes) {
+      PutFixed32(dst, prefix.size());
+      dst->append(prefix);
+    }
+  }
+
+  rocksdb::Status Decode(Slice *input) {
+    uint32_t size = 0;
+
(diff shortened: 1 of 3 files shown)


src/types/redis_search_encoding.h
```diff
@@ -0,0 +1,103 @@
...
+
+  rocksdb::Status Decode(Slice *input) {
+    uint32_t size = 0;
+
+    while (GetFixed32(input, &size)) {
+      if (input->size() < size) return rocksdb::Status::Corruption(kErrorInsufficientLength);
+      prefixes.emplace_back(input->data(), size);
+      input->remove_prefix(size);
+    }
+
```
mapleFU, line 61: Should we check `input->size() == 0` here?

src/types/redis_search_encoding.h
```diff
@@ -0,0 +1,103 @@
...
+#pragma once
+
+#include <encoding.h>
+#include <storage/redis_metadata.h>
+
+namespace redis {
+
+inline constexpr auto kErrorInsufficientLength = "insufficient length while decoding metadata";
+
+enum class SearchSubkeyType : uint8_t {
```
mapleFU, line 30: What do `TAG_FIELD_META` and `TAG_FIELD` means?

src/types/redis_search_encoding.h
PragmaTwice: I think it can be guarenteed by users.

src/types/redis_search_encoding.h
PragmaTwice: `TAG_FIELD_META` means metadata of this index (field, typed `tag`), and `TAG_FIELD` simply stores the index.

e.g. for a tag field `x` in key `k`, if we insert json `a = {"x": "hello;hi"}`, then the low-level ops are:
```
write:  k | TAG_FIELD | x | hello | a -> (nil)
write:  k | TAG_FIELD | x | hi | a -> (nil)
```

And the metadata of `x` may be:
```
k | TAG_FIELD_META | x -> ; | false
```

mapleFU: 🤔So this is different type of subkey?

mapleFU:
How can we list all indexes efficiently? 🤔
RediSearch seems would tracking the records for insert and maintaining index on that, should we support some global tracking structure?

PragmaTwice:
> How can we list all indexes efficiently? 🤔 RediSearch seems would tracking the records for insert and maintaining index on that, should we support some global tracking structure?

Like RediSearch, we will track all HASH and JSON commands and adjust our indexes accordingly.

mapleFU:
> Like RediSearch, we will track all HASH and JSON commands and adjust our indexes accordingly.

I mean, when start the process, we should "LIST" all indexes

PragmaTwice:
> > Like RediSearch, we will track all HASH and JSON commands and adjust our indexes accordingly.
> 
> I mean, when start the process, we should "LIST" all indexes

Ahh I got your point.

I think we can add a common prefix for all index key, or put them to a new CF. Currently I prefer the latter.

git-hulk:
> > > Like RediSearch, we will track all HASH and JSON commands and adjust our indexes accordingly.
> > 
> > 
> > I mean, when start the process, we should "LIST" all indexes
> 
> Ahh I got your point.
> 
> I think we can add a common prefix for all index key, or put them to a new CF. Currently I prefer the latter.

This skeleton code looks good to me. I also prefer the latter one if we want to keep the common prefix since the current column families are not for this purpose. But we need to do a little backward compatibility work while adding a new column family.

git-hulk (approved):

mapleFU:
So only "tag" is support currently?

PragmaTwice:
> So only "tag" is support currently?

Yeah, I also plan to add the support of numeric indexes.

But for text and vector fields, it is currently not in the plan.

PR #2067, opened 2024-01-28
Allow to migrate cluster slot by raw write batch

### LINKED ISSUE #2008: Implement the new migration type by applying raw batch

caipengbo:
This PR implemented the cluster slot migration by the rocksdb write batch, which will bring performance benefits. Also, we don't need to depend on parsing commands from the write batch log data if enabled. This makes the cluster migration feature more robust since we don't have to take care of the log data in each write batch.

This closes #2008 and the original PR is #1534


src/storage/storage.h
```diff
@@ -177,6 +177,7 @@ class Storage {
   bool IsClosing() const { return db_closing_; }
   std::string GetName() const { return config_->db_name; }
   rocksdb::ColumnFamilyHandle *GetCFHandle(const std::string &name);
+  rocksdb::ColumnFamilyHandle *GetCFHandle(ColumnFamilyID name);
```
git-hulk: Can rename it to cfd_id

src/cluster/slot_migrate.cc
```diff
@@ -1130,3 +1148,154 @@ void SlotMigrator::resumeSyncCtx(const Status &migrate_result) {
...
+        break;
+    }
+    if (batch_sender->IsFull()) {
+      GET_OR_RET(sendMigrationBatch(batch_sender));
+    }
+  }
+
+  // send the remaining data
+  return sendMigrationBatch(batch_sender);
+}
```
git-hulk:
```suggestion
}

```

src/cluster/slot_migrate.cc
```diff
@@ -1130,3 +1148,164 @@ void SlotMigrator::resumeSyncCtx(const Status &migrate_result) {
     blocking_context_ = nullptr;
   }
 }
+
+Status SlotMigrator::sendMigrationBatch(BatchSender *batch) {
+  // user may dynamically change some configs, apply it when send data
+  batch->SetMaxBytes(migrate_batch_size_bytes_);
+  batch->SetBytesPerSecond(migrate_batch_bytes_per_sec_);
```
git-hulk, line 1153:
```suggestion
  // Those configurations may be changed dynamically, so apply them here before sending.
  batch->SetMaxBytes(migrate_batch_size_bytes_);
  batch->SetBytesPerSecond(migrate_batch_bytes_per_sec_);
```

@caipengbo I cannot see those configurations will be changed except for constructing the migrator. Did I miss anything?

src/cluster/slot_migrate.cc
```diff
@@ -1130,3 +1148,164 @@ void SlotMigrator::resumeSyncCtx(const Status &migrate_result) {
...
+        GET_OR_RET(batch_sender->Put(storage_->GetCFHandle(static_cast<ColumnFamilyID>(item.column_family_id)),
+                                     item.key, item.value));
+        break;
+      }
+      case engine::WALItem::Type::kTypeDelete: {
+        GET_OR_RET(
+            batch_sender->Delete(storage_->GetCFHandle(static_cast<ColumnFamilyID>(item.column_family_id)), item.key));
+        break;
+      }
+      default:
```
git-hulk, line 1303: Should we care the kTypeDeleteRange write batch?

git-hulk:
@caipengbo Great! Overall is quite clear and concise. Just a few questions inline.

caipengbo:
> Just a few questions inline.

I have modified the question you mentioned before. Are there any other questions? @git-hulk

git-hulk:
@caipengbo Sorry for didn't send out the comments. 😭

src/cluster/slot_migrate.cc
caipengbo: `SetMigrateBatchSize` will be called in config set.

src/cluster/slot_migrate.cc
caipengbo: I don't have a good way to handle delete range because it may overwrite multiple slots. Do you have any good ideas? 
I don't see that the redis-command migration handles delete range either.

src/cluster/slot_migrate.cc
git-hulk: oh, I missed this place. Thank you!

src/cluster/slot_migrate.cc
git-hulk: The delete range API is only used when flushing the db/all, so it should be a challenge to handle this while migrating. Can you add this switch branch and comment a bit the reason like:

```
// Do nothing in DeleteRange due to it might cross multiple slots. It's only used in
// FLUSHDB/FLUSHALL commands for now and maybe we can disable them while migrating.
```

caipengbo: Sound good :)

git-hulk (dismissed):
LGTM

git-hulk (approved):

Issue #1212, opened 2022-12-28
BITPOS doesn't support BIT option


lt1946:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Version

latest

### Minimal reproduce step

from redis command : [https://redis.io/commands/bitpos/](url)
redis> SET mykey "\x00\xff\xf0"
"OK"
 

### What did you expect to see?

redis> BITPOS mykey 1 7 15 BIT
(integer) 9

### What did you see instead?

redis> BITPOS mykey 1 7 15 BIT
(integer) 57

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

torwig:
Hi @lt1946, thank you for your report. 
On my machine, with Redis I have the following:
```
[yaroslav@localhost ~]$ redis-cli
127.0.0.1:6379> SET mykey1 "\x00\xff\xf0"
OK
127.0.0.1:6379> BITPOS mykey1 1 7 15 BIT
(integer) 8
```
However, you said that you expect to see `9`.

Kvrocks behaves in the following way:
```
127.0.0.1:6666> SET mykey "\x00\xff\xf0"
OK
127.0.0.1:6666> BITPOS mykey 1 7 15 BIT
(integer) -1
```
And it needs an investigation.

torwig:
@lt1946 I got the reason why the result is `-1`. Currently, the `BITPOS` command doesn't support `BIT` flag, so it considers `start` and `end` values as **bytes**.
```
127.0.0.1:6379> BITPOS mykey1 1 1 2
(integer) 8
127.0.0.1:6379> BITPOS mykey1 1 2 3
(integer) 16
```

tisonkun:
I'd suggest adding a comment/description in the document https://kvrocks.apache.org/docs/supported-commands#bit-commands and closing this issue.

sheharyaar:
@jihuayu , I would like to take up this issue, please assign me this.

jihuayu:
@sheharyaar Assign! You can refer to https://github.com/apache/kvrocks/pull/2087

sheharyaar:
@jihuayu , I think of two approaches :

1. We can convert the bits to bytes and when the bit position is retrieved we can check if it falls with [start,end) bits.
[...]

PR #2087, opened 2024-02-01
Support BIT and BYTE options in the BITCOUNT command

### LINKED ISSUE #2006: Add BIT option support for BITCOUNT command
### Search before asking

### Motivation

I found that the `BITCOUNT` command doesn't have the `BIT` option supported. [Source](https://github.com/jihuayu/incubator-kvrocks/blob/588650f51c434b161c8a8de65ea80871c7d2573e/src/commands/cmd_bit.cc#L124)

### Solution

Add BIT option support for `BITCOUNT` command.

relate issues: https://github.com/apache/kvrocks/issues/1212

kay011:
For #2006, This PR enables the BITCOUNT command to support **bit** and **byte** options, but the go-redis library does not seem to support this option yet, so only the cpp unit tests are added, and the correctness is verified locally through redis-cli.


PragmaTwice:
> but the go-redis library does not seem to support this option yet

You can use `rdb.Do` in go-redis to send any command and options to kvrocks. We don't need go-redis to support anything.

jihuayu:
It looks like clang-tidy run failed. You need to fix it before merging.
https://github.com/apache/kvrocks/actions/runs/7744539810/job/21124606184?pr=2087#step:8:1406

src/commands/cmd_bit.cc
```diff
@@ -143,6 +143,7 @@ class CommandBitCount : public Commander {
  private:
   int64_t start_ = 0;
   int64_t stop_ = -1;
+  bool is_bit_ = false;
```
jihuayu: Other commands have different meanings of `bit_`, which may cause misunderstanding. 
Can you change it to another name? Such as `type` or `flag`? If you can change it to an enum, it is better! 

https://github.com/kay011/kvrocks/blob/64ba5f9e72e85429fe625878da8d6442ba0ff475/src/commands/cmd_bit.cc#L91
https://github.com/kay011/kvrocks/blob/64ba5f9e72e85429fe625878da8d6442ba0ff475/src/commands/cmd_bit.cc#L201

src/commands/cmd_bit.cc
```diff
@@ -121,7 +121,7 @@ class CommandBitCount : public Commander {
     if (args.size() == 5) {
       if (util::EqualICase(args[4], "BYTE")) {
```
jihuayu, line 122: Can you explicitly set the value of the variable? This will make the code more readable.

src/types/redis_bitmap.cc
```diff
@@ -226,14 +226,20 @@ rocksdb::Status Bitmap::BitCount(const Slice &user_key, int64_t start, int64_t s
...
+  auto totlen = static_cast<int64_t>(metadata.size);
+  if (is_bit) totlen <<= 3;
   // Counting bits in byte [start, stop].
-  std::tie(start, stop) = BitmapString::NormalizeRange(start, stop, static_cast<int64_t>(metadata.size));
+  std::tie(start, stop) = BitmapString::NormalizeRange(start, stop, totlen);
   // Always return 0 if start is greater than stop after normalization.
   if (start > stop) return rocksdb::Status::OK();
 
+  uint8_t first_byte_neg_mask = 0, last_byte_neg_mask = 0;
+  std::tie(start, stop) =
```
jihuayu: It seems like from this line, `start` and `end` means change from start BIT/BYTE to start BYTE. If it's true, can you change the params name? 
Variables with different meanings require different names.

src/types/redis_bitmap_string.h
```diff
@@ -56,6 +57,9 @@ class BitmapString : public Database {
   // Return:
   //  The normalized [start, end] range.
   static std::pair<int64_t, int64_t> NormalizeRange(int64_t origin_start, int64_t origin_end, int64_t length);
+
+  static std::pair<int64_t, int64_t> AdjustMaskWithRange(bool is_bit, int64_t origin_start, int64_t origin_end,
```
jihuayu: Can you add a comment to tell others what this function does?

src/types/redis_bitmap_string.cc
```diff
@@ -68,17 +68,31 @@ rocksdb::Status BitmapString::SetBit(const Slice &ns_key, std::string *raw_value
...
+  std::tie(start, stop) = NormalizeRange(start, stop, totlen);
+  // Always return 0 if start is greater than stop after normalization.
+  if (start > stop) return rocksdb::Status::OK();
+
+  uint8_t first_byte_neg_mask = 0, last_byte_neg_mask = 0;
+  std::tie(start, stop) = AdjustMaskWithRange(is_bit, start, stop, &first_byte_neg_mask, &last_byte_neg_mask);
 
   /* Precondition: end >= 0 && end < strlen, so the only condition where
    * zero can be returned is: start > stop. */
   if (start <= stop) {
```
jihuayu: It seems like this will always be true because we checked it before.

src/types/redis_bitmap_string.cc
```diff
@@ -68,17 +68,31 @@ rocksdb::Status BitmapString::SetBit(const Slice &ns_key, std::string *raw_value
...
   auto strlen = static_cast<int64_t>(string_value.size());
-  std::tie(start, stop) = NormalizeRange(start, stop, strlen);
+  int64_t totlen = strlen;
+  if (is_bit) totlen <<= 3;
+  std::tie(start, stop) = NormalizeRange(start, stop, totlen);
+  // Always return 0 if start is greater than stop after normalization.
+  if (start > stop) return rocksdb::Status::OK();
+
+  uint8_t first_byte_neg_mask = 0, last_byte_neg_mask = 0;
+  std::tie(start, stop) = AdjustMaskWithRange(is_bit, start, stop, &first_byte_neg_mask, &last_byte_neg_mask);
```
jihuayu: This line also has variable meanings changed.

jihuayu (commented):
We need to clarify whether the "start" variable in the function is calculated in bits or bytes.
Others LGTM.

src/commands/cmd_bit.cc
mapleFU: I think something like `position_is_bit_` would be better

kay011:
@jihuayu Thank you for review carefully, I will fix it when I am free ～

src/types/redis_bitmap_string.h
mapleFU: Something like "normalized bit to byte range" would help

src/types/redis_bitmap_string.cc
```diff
@@ -202,6 +216,18 @@ std::pair<int64_t, int64_t> BitmapString::NormalizeRange(int64_t origin_start, i
   return {origin_start, origin_end};
 }
 
+std::pair<int64_t, int64_t> BitmapString::AdjustMaskWithRange(bool is_bit, int64_t origin_start, int64_t origin_end,
+                                                              uint8_t *first_byte_neg_mask,
+                                                              uint8_t *last_byte_neg_mask) {
+  if (is_bit) {
```
mapleFU, line 230: Nit: should we add `DCHECK(origin_start <= origin_end);` here?

tests/cppunit/types/bitmap_test.cc
```diff
@@ -72,39 +72,56 @@ TEST_P(RedisBitmapTest, BitCount) {
...
   EXPECT_EQ(cnt, 6);
-  bitmap_->BitCount(key_, 0, -1, &cnt);
+  bitmap_->BitCount(key_, 0, -1, false, &cnt);
   EXPECT_EQ(cnt, 6);
   auto s = bitmap_->Del(key_);
 }
 
+TEST_P(RedisBitmapTest, BitCountBitmapString) {
+  std::string str = "hello";
+  string_->Set(key_, str);
```
mapleFU: we can avoid hackings for `str = "hello"`, just set specific bits. Setting a string would be hacking

https://github.com/apache/kvrocks/pull/2075 This patch unify their behaviors

tests/cppunit/types/bitmap_test.cc
```diff
@@ -115,15 +132,43 @@ TEST_P(RedisBitmapTest, BitCountNegative) {
...
 
   auto s = bitmap_->Del(key_);
 }
 
+TEST_P(RedisBitmapTest, BitCountBITOption) {
+  uint32_t offsets[] = {0, 100, 1024 * 8, 1024 * 8 + 1, 3 * 1024 * 8, 3 * 1024 * 8 + 1};
+  for (const auto &offset : offsets) {
+    bool bit = false;
+    bitmap_->SetBit(key_, offset, true, &bit);
+  }
```
mapleFU, line 132: Nit: we can add more test case to just get self. ( `BitCount(key_, offset, offset, ..)` )

mapleFU (commented):
LGTM except some nits and testings

tests/cppunit/types/bitmap_test.cc
mapleFU: You can add something like:

```
for (uint32_t bit_offset = 0; bit_offset <= 3 * 1024 * 8 + 1; ++bit_offset) {
  bitmap_->BitCount(key_, offset, offset, /*...*/true, &positive_cnt);
  bitmap_->BitCount(key_, offset, offset,  /*...*/false, &negative_cnt);
  // check the result.
}
```

mapleFU:
Others LGTM except https://github.com/apache/kvrocks/pull/2087#discussion_r1477628027

tests/cppunit/types/bitmap_test.cc
mapleFU:
```c++
  for (uint32_t offset = 0; offset <= 3 * 1024 * 8 + 1; ++offset) {
    uint32_t cnt = 0;
    EXPECT_TRUE(bitmap_->BitCount(key_, offset, offset, /*bit=*/true, &cnt).ok());
    if (offset == 0 || offset == 100 || offset == 1024 * 8 || offset == 1024 * 8 + 1 || offset == 3 * 1024 * 8 ||
        offset == 3 * 1024 * 8 + 1) {
      EXPECT_EQ(1, cnt) << "offset: " << offset;
    } else {
      EXPECT_EQ(0, cnt) << "offset: " << offset;
    }
  }
```

E.g. Add the code above

src/types/redis_bitmap.cc
```diff
@@ -252,16 +261,31 @@ rocksdb::Status Bitmap::BitCount(const Slice &user_key, int64_t start, int64_t s
...
+    }
+    if (i == stop_index) {
+      stop_in_segment = u_stop % kBitmapSegmentBytes;
+      if (is_bit_index && start_in_segment <= stop_in_segment && last_byte_neg_mask != 0) {
+        uint8_t last_mask_byte = kBitSwapTable[static_cast<uint8_t>(pin_value[stop_in_segment])] & last_byte_neg_mask;
+        mask_cnt += BitmapString::RawPopcount(&last_mask_byte, 1);
+      }
+    }
+    int64_t bytes = 0;
+    if (stop_in_segment >= start_in_segment) {
```
mapleFU: This might visit undefined memory, which could introduce bug

mapleFU:
@kay011 I found adding the test would find Bitmap has a bug here(BitmapString is ok). I've a fixing for that. Would you mind fix it yourself or let me fix that?

kay011:
> @kay011 I found adding the test would find Bitmap has a bug here(BitmapString is ok). I've a fixing for that. Would you mind fix it yourself or let me fix that?

you can fix it and I will merge it. I think maybe the origin BITCOUNT which used BYTE index also has this bug

mapleFU:
>  I think maybe the origin BITCOUNT which used BYTE index also has this bug

Yeah

tests/cppunit/types/bitmap_test.cc
```diff
@@ -115,15 +115,46 @@ TEST_P(RedisBitmapTest, BitCountNegative) {
...
+TEST_P(RedisBitmapTest, BitCountBITOption) {
+  uint32_t offsets[] = {0, 100, 1024 * 8, 1024 * 8 + 1, 3 * 1024 * 8, 3 * 1024 * 8 + 1};
+  for (const auto &offset : offsets) {
+    bool bit = false;
+    uint32_t cnt = 0;
+    bitmap_->SetBit(key_, offset, true, &bit);
+    bitmap_->BitCount(key_, offset, offset, true, &cnt);
+    EXPECT_EQ(cnt, 1);
+  }
+  uint32_t cnt = 0;
```
mapleFU, line 144:
```suggestion
  for (uint32_t offset = 0; offset <= 3 * 1024 * 8 + 10; ++offset) {
    uint32_t cnt = 0;
    bitmap_->BitCount(key_, offset, offset, /*is_bit_index=*/true, &cnt);
    if (offset == 0 || offset == 100 || offset == 1024 * 8 || offset == 1024 * 8 + 1 || offset == 3 * 1024 * 8 ||
        offset == 3 * 1024 * 8 + 1) {
      ASSERT_EQ(cnt, 1) << "offset: " << offset;
    } else {
      ASSERT_EQ(cnt, 0) << "offset: " << offset;
    }
  }
  uint32_t cnt = 0;
```

src/types/redis_bitmap.cc
```diff
@@ -252,16 +261,31 @@ rocksdb::Status Bitmap::BitCount(const Slice &user_key, int64_t start, int64_t s
...
+      bytes = stop_in_segment - start_in_segment + 1;
+    }
+    *cnt += BitmapString::RawPopcount(reinterpret_cast<const uint8_t *>(pin_value.data()) + start_in_segment, bytes);
```
mapleFU:
```suggestion
    // Counting bits in [start_in_segment, stop_in_segment]
    int64_t start_in_segment = 0;                                                // start_index int 1024 bytes segment
    auto readable_stop_in_segment = static_cast<int64_t>(pin_value.size() - 1);  // stop_index int 1024 bytes segment
    auto stop_in_segment = readable_stop_in_segment;
    if (i == start_index) {
      start_in_segment = u_start % kBitmapSegmentBytes;
      if (is_bit_index && start_in_segment <= readable_stop_in_segment && first_byte_neg_mask != 0) {
        uint8_t first_mask_byte =
            kBitSwapTable[static_cast<uint8_t>(pin_value[start_in_segment])] & first_byte_neg_mask;
        mask_cnt += BitmapString::RawPopcount(&first_mask_byte, 1);
      }
    }
    if (i == stop_index) {
      stop_in_segment = u_stop % kBitmapSegmentBytes;
      if (is_bit_index && stop_in_segment <= readable_stop_in_segment && last_byte_neg_mask != 0) {
        uint8_t last_mask_byte = kBitSwapTable[static_cast<uint8_t>(pin_value[stop_in_segment])] & last_byte_neg_mask;
        mask_cnt += BitmapString::RawPopcount(&last_mask_byte, 1);
      }
    }
    if (stop_in_segment >= start_in_segment && readable_stop_in_segment >= start_in_segment) {
      int64_t bytes = 0;
      bytes = std::min(stop_in_segment, readable_stop_in_segment) - start_in_segment + 1;
      *cnt += BitmapString::RawPopcount(reinterpret_cast<const uint8_t *>(pin_value.data()) + start_in_segment, bytes);
    }
```

mapleFU:
@kay011 My edits are listed above

kay011:
> @kay011 My edits are listed above
Got it. Let me merge it on this commit.

mapleFU (approved):

mapleFU:
Will wait before Friday to waiting others review it, since I'm one of the editor.

PragmaTwice (approved):

jihuayu:
Thanks @kay011. BITOPS also doesn't support BIT option https://github.com/apache/kvrocks/issues/1212

Do you like to have a try?

kay011:
> Thanks @kay011. BITOPS also doesn't support BIT option #1212
> 
> Do you like to have a try?

@jihuayu Yeah, I am willing to do it.

PR #2100, opened 2024-02-13
Use require.JSONEq instead of require.Equal for the JSON string

jihuayu:
I have used `require.JSONEq` instead of `require.Equal` for the JSON string.
We don't need to test the order of the elements in the JSON.


git-hulk (dismissed):
LGTM, thanks

tisonkun (dismissed):

PragmaTwice (changes requested):
At least in some test cases, we need to validate the indent and spaces of the output JSON. And `JSONEq` will just ignore all of them.

For exmaple, all cases in `JSON.GET with options`.

jihuayu:
> At least in some test cases, we need to validate the indent and spaces of the output JSON. And `JSONEq` will just ignore all of them.
> 
> For exmaple, all cases in `JSON.GET with options`.

Thank you, I fixed it.

PragmaTwice (approved):

PR #2103, opened 2024-02-15
Set the redis-cursor-compatible field to yes in the initial configuration

jihuayu:
After a long time, no bugs have been reported with this Redis cursor format compatibility feature.

I think we can set the `redis-cursor-compatible field` in the initial configuration to `yes` so that new users can get out-of-the-box support for the Redis cursor format compatibility feature.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -309,7 +309,7 @@ max-bitmap-to-string-mb 16
 # If enabled, the cursor will be unsigned 64-bit integers.
 # If disabled, the cursor will be a string.
 # Default: no
-redis-cursor-compatible no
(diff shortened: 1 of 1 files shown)


git-hulk (approved):

git-hulk:
Changing the default value in the configuration file won't affect the legacy users, but need to highlight this modification in our next release note. cc @PragmaTwice @caipengbo

caipengbo:
> need to highlight this modification in our next release note.

Yes, this is an important enhancement and we should have highlighted it.

PragmaTwice (approved):

PR #2111, opened 2024-02-21
Initialize the unified index updater for HASH and JSON data type

PragmaTwice:
In this PR, we add the basic structure of the index updater: used to maintain the index structure based on rocksdb key-values.

In this implementation, we choose the before-write-after-write-hooking method, instead of 
- logical logging (since we don't have a decent logical logging system, but glad to have if someone want to implement)
- per-command-hooking (hard to maintain)

About concurrency safety:
- In order to ensure concurrency safety, write commands that need to build the index will be weakened into serial execution (via global locking).
- We can rewrite this part if kvrocks gets a good transaction model later.

The general workflow is:
- before writing (if data type and prefix matched)
  - indexed field recording
- user's HASH/JSON writing ops
- after writing (if writing ops succeed and the before-writing record succeed)
  - indexed field re-recording
  - recorded field diff
  - index writing

Besides, we introduce a HAT-trie (https://github.com/Tessil/hat-trie), to support the RediSearch key prefix matching.

TODO:
- index key-value building
- before-write/after-write difference comparing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -138,6 +138,7 @@ include(cmake/fmt.cmake)
 include(cmake/jsoncons.cmake)
 include(cmake/xxhash.cmake)
 include(cmake/span.cmake)
+include(cmake/trie.cmake)
 
 if (ENABLE_LUAJIT)
     include(cmake/luajit.cmake)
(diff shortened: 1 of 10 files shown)


git-hulk (dismissed):

PragmaTwice:
Done. Other parts will be submitted as further PRs. cc @git-hulk

git-hulk (approved):

PR #2112, opened 2024-02-23
Check if the type is an entry  while iterating the stream subkeys

### LINKED ISSUE #2109: Create stream consumer group will make the xread command error
### Search before asking

### Version

unstable

### Minimal reproduce step

```shell
XADD test * a a
XGROUP create test test 0
xread STREAMS test 0
```

```shell
1) 1) "test"
   2) 1) 1) "1708506496723-0"
         2) 1) "a"
            2) "a"
```

### What did you see instead?

(error) ERR Invalid argument: failed to decode stream entry value

### Anything Else?

I discovered this when developing the xreadgroup command.

The subkey of the consumer group overlapped with the subkey of the stream entry.
This causes the [range function](https://github.com/apache/kvrocks/blob/30f6bee42d4352715ce254a52c4403fe6b0ab5e7/src/types/redis_stream.cc#L686) to get the stream Entry with the group's subkey.

I think there is something wrong with the subkey design of the consumer & consumer group metadata.
https://github.com/apache/kvrocks/blob/30f6bee42d4352715ce254a52c4403fe6b0ab5e7/src/types/redis_stream.cc#L168-L176

I don't really understand why we puts the consumerGroupMetadataDelimiter at the end of the subkey. Shouldn't we put it at the start?

jihuayu:
fixed #2109

### DIFF
--- a/src/types/redis_stream.cc
+++ b/src/types/redis_stream.cc
@@ -250,7 +250,7 @@ StreamConsumerMetadata Stream::decodeStreamConsumerMetadataValue(const std::stri
   return consumer_metadata;
 }
 
-StreamSubkeyType Stream::identifySubkeyType(const rocksdb::Slice &key) {
+StreamSubkeyType Stream::identifySubkeyType(const rocksdb::Slice &key) const {
   InternalKey ikey(key, storage_->IsSlotIdEncoded());
   Slice subkey = ikey.GetSubKey();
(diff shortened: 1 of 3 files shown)


Yangsx-1 (approved):
LGTM.

git-hulk (approved):

PR #2116, opened 2024-02-25
Add the support of the LCS command

### LINKED ISSUE #1955: Add support of the command LCS
Refer to https://redis.io/commands/lcs/

JoverZhang:
Close #1955


src/commands/cmd_string.cc
```diff
@@ -637,6 +826,5 @@ REDIS_REGISTER_COMMANDS(
     MakeCmdAttr<CommandIncrByFloat>("incrbyfloat", 3, "write", 1, 1, 1),
     MakeCmdAttr<CommandIncr>("incr", 2, "write", 1, 1, 1), MakeCmdAttr<CommandDecrBy>("decrby", 3, "write", 1, 1, 1),
     MakeCmdAttr<CommandDecr>("decr", 2, "write", 1, 1, 1), MakeCmdAttr<CommandCAS>("cas", -4, "write", 1, 1, 1),
-    MakeCmdAttr<CommandCAD>("cad", 3, "write", 1, 1, 1), )
-
+    MakeCmdAttr<CommandCAD>("cad", 3, "write", 1, 1, 1), MakeCmdAttr<CommandLCS>("lcs", -3, "read-only", 1, 1, 1), )
```
jihuayu:
```suggestion
    MakeCmdAttr<CommandCAD>("cad", 3, "write", 1, 1, 1), MakeCmdAttr<CommandLCS>("lcs", -3, "read-only", 1, 2, 1), )
```
> LCS key1 key2 [LEN] [IDX] [MINMATCHLEN min-match-len] [WITHMATCHLEN]

It means that param 1 and param 2 are key.

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -895,4 +895,95 @@ func TestString(t *testing.T) {
...
+
+	rna1 := "CACCTTCCCAGGTAACAAACCAACCAACTTTCGATCTCTTGTAGATCTGTTCTCTAAACGAACTTTAAAATCTGTGTGGCTGTCACTCGGCTGCATGCTTAGTGCACTCACGCAGTATAATTAATAACTAATTACTGTCGTTGACAGGACACGAGTAACTCGTCTATCTTCTGCAGGCTGCTTACGGTTTCGTCCGTGTTGCAGCCGATCATCAGCACATCTAGGTTTCGTCCGGGTGTG"
+	rna2 := "ATTAAAGGTTTATACCTTCCCAGGTAACAAACCAACCAACTTTCGATCTCTTGTAGATCTGTTCTCTAAACGAACTTTAAAATCTGTGTGGCTGTCACTCGGCTGCATGCTTAGTGCACTCACGCAGTATAATTAATAACTAATTACTGTCGTTGACAGGACACGAGTAACTCGTCTATCTTCTGCAGGCTGCTTACGGTTTCGTCCGTGTTGCAGCCGATCATCAGCACATCTAGGTTT"
+	rnalcs := "ACCTTCCCAGGTAACAAACCAACCAACTTTCGATCTCTTGTAGATCTGTTCTCTAAACGAACTTTAAAATCTGTGTGGCTGTCACTCGGCTGCATGCTTAGTGCACTCACGCAGTATAATTAATAACTAATTACTGTCGTTGACAGGACACGAGTAACTCGTCTATCTTCTGCAGGCTGCTTACGGTTTCGTCCGTGTTGCAGCCGATCATCAGCACATCTAGGTTT"
+
+	t.Run("LCS basic", func(t *testing.T) {
+		require.NoError(t, rdb.Set(ctx, "virus1", rna1, 0).Err())
+		require.NoError(t, rdb.Set(ctx, "virus2", rna2, 0).Err())
+		require.Equal(t, rnalcs, rdb.LCS(ctx, &redis.LCSQuery{Key1: "virus1", Key2: "virus2"}).Val().MatchString)
+		// require.Equal(t, rnalcs, rdb.Do(ctx, "LCS", "virus1", "virus2").Val())
```
jihuayu: If this line is not useful, please delete it.

src/server/redis_request.cc
```diff
@@ -37,7 +37,6 @@
 namespace redis {
 
 const size_t PROTO_INLINE_MAX_SIZE = 16 * 1024L;
-const size_t PROTO_BULK_MAX_SIZE = 512 * 1024L * 1024L;
 const size_t PROTO_MULTI_MAX_SIZE = 1024 * 1024L;
```
jihuayu: I think that moving only a constant definition to a header file will break consistency.
If it is really necessary to move. Moving all three definitions to the header file would be better.

src/commands/cmd_string.cc
```diff
@@ -620,6 +621,201 @@ class CommandCAD : public Commander {
...
+    auto alen = static_cast<uint32_t>(a.length());
+    auto blen = static_cast<uint32_t>(b.length());
+
+    // Allocate the LCS table.
+    uint64_t dp_size = (alen + 1) * (blen + 1);
+    uint64_t bulk_size = dp_size * sizeof(uint32_t);
+    if (bulk_size >= SIZE_MAX || bulk_size / dp_size != sizeof(uint32_t)) {
+      return {Status::RedisExecErr, "Insufficient memory, failed allocating transient memory for LCS"};
+    }
+    if (bulk_size > PROTO_BULK_MAX_SIZE) {
```
jihuayu: I don't quite understand why we need this judgment.

jihuayu (commented):
Thank you @JoverZhang, wonderful algorithm.
I think it's not good to put so much logic in the commander. 
Would it be better to move the specific logic to redis::String? @git-hulk

src/commands/cmd_string.cc
JoverZhang: I think Kvrocks‘s `PROTO_BULK_MAX_SIZE` should be equivalent to Redis's `proto_max_bulk_len`.

https://github.com/redis/redis/blob/14176484697dc96df62c76630eef2b97a89bb8d0/src/t_string.c#L804-L807

git-hulk: This prevents the command occupies too much memory while calculating LCS table.

jihuayu: Thank you!

src/commands/cmd_string.cc
```diff
@@ -620,6 +621,201 @@ class CommandCAD : public Commander {
...
+    }
+
+    // Compute the LCS using the vanilla dynamic programming technique of
+    // building a table of LCS(x, y) substrings.
+    auto alen = static_cast<uint32_t>(a.length());
+    auto blen = static_cast<uint32_t>(b.length());
+
+    // Allocate the LCS table.
+    uint64_t dp_size = (alen + 1) * (blen + 1);
+    uint64_t bulk_size = dp_size * sizeof(uint32_t);
```
mapleFU: Should we better check `dp_size` not overflow uint32_t? Since the memory usage will be 16GB if dp_size reaches `numeric_limits<uint32_t>::max()` (Though redis uses same algorithm)

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -895,4 +895,94 @@ func TestString(t *testing.T) {
...
+		require.NoError(t, rdb.Set(ctx, "virus2", rna2, 0).Err())
+		matches := rdb.LCS(ctx, &redis.LCSQuery{Key1: "virus1", Key2: "virus2", Idx: true, WithMatchLen: true, MinMatchLen: 5}).Val().Matches
+		require.Equal(t, []redis.LCSMatchedPosition{
+			{
+				Key1:     redis.LCSPosition{Start: 1, End: 222},
+				Key2:     redis.LCSPosition{Start: 13, End: 234},
+				MatchLen: 222,
+			},
+		}, matches)
+	})
```
mapleFU, line 987: Would you mind add tests case for empty string?

src/commands/cmd_string.cc
```diff
@@ -620,6 +621,201 @@ class CommandCAD : public Commander {
...
+          lcs(i, j) = std::max(lcs(i - 1, j), lcs(i, j - 1));
+        }
+      }
+    }
+
+    // Store the actual LCS string if needed.
+    std::string result;
+    uint32_t idx = lcs(alen, blen);
+
+    // Do we need to compute the actual LCS string? Allocate it in that case.
```
mapleFU:
```suggestion
    // Allocate when we need to compute the actual LCS string
```

src/commands/cmd_string.cc
```diff
@@ -620,6 +621,201 @@ class CommandCAD : public Commander {
...
+    uint32_t arange_start = alen;  // alen signals that values are not set.
+    uint32_t arange_end = 0;
+    uint32_t brange_start = 0;
+    uint32_t brange_end = 0;
+    while (compute_lcs && i > 0 && j > 0) {
+      bool emit_range = false;
+      if (a[i - 1] == b[j - 1]) {
+        // If there is a match, store the character and reduce
+        // the indexes to look for a new match.
+        result[idx - 1] = a[i - 1];
```
mapleFU: Better using `result.at()` to make it safer, since `result` would not be resize in some cases

mapleFU (commented):
I didn't review the logic for response, just some few comments

jihuayu:
Can someone confirm whether this logic is correct? I am also unable to determine it.
Alternatively, can we ignore whether the algorithm is correct and fix it when a bug arises?

src/server/redis_request.h
```diff
@@ -32,6 +32,10 @@ class Server;
 
 namespace redis {
 
+const size_t PROTO_INLINE_MAX_SIZE = 16 * 1024L;
```
mapleFU: Nit: change to constexpr?

src/commands/cmd_string.cc
```diff
@@ -620,6 +621,201 @@ class CommandCAD : public Commander {
...
+    std::string matches;
+    uint32_t matches_len = 0;
+
+    uint32_t i = alen;
+    uint32_t j = blen;
+    uint32_t arange_start = alen;  // alen signals that values are not set.
+    uint32_t arange_end = 0;
+    uint32_t brange_start = 0;
+    uint32_t brange_end = 0;
+    while (compute_lcs && i > 0 && j > 0) {
```
mapleFU: Check `compute_lcs` in loop is so weird, isn't it?

src/commands/cmd_string.cc
```diff
@@ -620,6 +621,201 @@ class CommandCAD : public Commander {
...
+          i--;
+        else
+          j--;
+        if (arange_start != alen) emit_range = true;
+      }
+
+      // Emit the current range if needed.
+      uint32_t match_len = arange_end - arange_start + 1;
+      if (emit_range) {
+        if (get_idx_ && (min_match_len_ == 0 || match_len >= min_match_len_)) {
```
mapleFU: sigh, don't understand why a `minmatchlen == 0` is checked here...Is this a fast path?

mapleFU (commented):
General LGTM. Just some few nits

@jihuayu Actually I think this impl works(Since it's code it's similiar with Redis' impl). However, I think the most important thing is prevent malicious user from using lcs to attack the system, or lcs making the server crash...

PragmaTwice (commented):
As mentioned by @jihuayu, the implementation of the algorithm should be put into redis_string.cc instead of cmd_string.cc.

src/commands/cmd_string.cc
JoverZhang: Ah. When `minmatchlen == 0`, the `minmatchlen` is ignored because its default value is `0`.

src/commands/cmd_string.cc
JoverZhang: Ok, I have modified it.

JoverZhang:
> As mentioned by @jihuayu, the implementation of the algorithm should be put into redis_string.cc instead of cmd_string.cc.

Understood, I have moved the algorithm to `redis_string.cc` as suggested. Please review the implementation and let me know if any further modifications are required.

src/types/redis_string.cc
```diff
@@ -530,4 +531,156 @@ rocksdb::Status String::CAD(const std::string &user_key, const std::string &valu
...
+    *rst = StringLCSIdxResult{{}, 0};
+  } else {
+    *rst = std::string{};
+  }
+
+  std::string a;
+  std::string b;
+  std::string ns_key1 = AppendNamespacePrefix(user_key1);
+  std::string ns_key2 = AppendNamespacePrefix(user_key2);
+  MultiLockGuard guard(storage_->GetLockManager(), std::vector{ns_key1, ns_key2});
```
PragmaTwice: I don't think we need a lock since we don't change anything here.

JoverZhang: Oh, sorry. I have removed it.

src/types/redis_string.cc
```diff
@@ -530,4 +531,155 @@ rocksdb::Status String::CAD(const std::string &user_key, const std::string &valu
...
+  auto blen = static_cast<uint32_t>(b.length());
+
+  // Allocate the LCS table.
+  uint64_t dp_size = (alen + 1) * (blen + 1);
+  uint64_t bulk_size = dp_size * sizeof(uint32_t);
+  if (bulk_size > PROTO_BULK_MAX_SIZE || bulk_size / dp_size != sizeof(uint32_t)) {
+    return rocksdb::Status::Aborted("Insufficient memory, transient memory for LCS exceeds proto-max-bulk-len");
+  }
+  std::vector<uint32_t> dp(dp_size, 0);
+  auto lcs = [&](const uint32_t i, const uint32_t j) -> uint32_t & { return dp[i * (blen + 1) + j]; };
```
```suggestion
  auto lcs = [&dp, blen](const uint32_t i, const uint32_t j) -> uint32_t & { return dp[i * (blen + 1) + j]; };
```

src/types/redis_string.cc
```diff
@@ -530,4 +531,155 @@ rocksdb::Status String::CAD(const std::string &user_key, const std::string &valu
...
+      uint32_t lcs2 = lcs(i, j - 1);
+      if (lcs1 > lcs2)
+        i--;
+      else
+        j--;
+      if (arange_start != alen) emit_range = true;
+    }
+
+    // Emit the current range if needed.
+    uint32_t match_len = arange_end - arange_start + 1;
```
mapleFU: put this in `emit_range`?

src/types/redis_string.h
```diff
@@ -42,8 +44,34 @@ struct StringSetArgs {
...
+
+struct StringLCSMatchedRange {
+  StringLCSRange a;
+  StringLCSRange b;
+  uint32_t match_len;
+};
+
+struct StringLCSIdxResult {
+  // Matched ranges.
+  std::list<StringLCSMatchedRange> matches;
```
mapleFU: Though I don't think performance of LCS it's important, I think `deque` or `vector` ( using vector without pop_front) would be better here.

mapleFU (commented):
General LGTM

src/types/redis_string.cc
```diff
@@ -530,4 +531,155 @@ rocksdb::Status String::CAD(const std::string &user_key, const std::string &valu
...
+    if (emit_range) {
+      if (auto result = std::get_if<StringLCSIdxResult>(rst)) {
+        // Always emit the range when the `min_match_len` is not set.
+        if (args.min_match_len == 0 || match_len >= args.min_match_len) {
+          result->matches.emplace_back(StringLCSMatchedRange{StringLCSRange{arange_start, arange_end},
+                                                             StringLCSRange{brange_start, brange_end}, match_len});
```
```suggestion
          result->matches.emplace_back(StringLCSRange{arange_start, arange_end},
                                                             StringLCSRange{brange_start, brange_end}, match_len);
```

src/types/redis_string.h
PragmaTwice: Yeah, deque or vector is better here.

src/types/redis_string.cc
JoverZhang: I might not fully grasp your point. Are you suggesting that we should make `emit_range = arange_end - arange_start + 1`?

src/types/redis_string.h
JoverZhang: Thank. I have modified this to use the `vector`.

Initially, I hadn't used the `vector` because I was concerned that a very large `vector` would significantly increase memory usage when building the `output` at the end.
However, I forgot that `dp` actually frees up more memory space at the end of `string_db.LCS()`.

src/types/redis_string.cc
mapleFU:
```
// Emit the current range if needed.
- uint32_t match_len = arange_end - arange_start + 1;
if (emit_range) {
+   uint32_t match_len = arange_end - arange_start + 1;
```

mapleFU (dismissed):
LGTM

mapleFU (approved):

PragmaTwice (approved):

jihuayu (approved):

PragmaTwice:

jihuayu:
Hi @JoverZhang, can you add this command to our docs? 
You can set `Since Version` with unstable.
https://kvrocks.apache.org/docs/supported-commands#string-commands

JoverZhang:
> Hi @JoverZhang, can you add this command to our docs? You can set `Since Version` with unstable. https://kvrocks.apache.org/docs/supported-commands#string-commands

Okay, I added it.
https://github.com/apache/kvrocks-website/pull/208

PR #2117, opened 2024-02-26
Fallback to use the redis command migration type if the target don't support the ApplyBatch command

### LINKED ISSUE #2078: Use redis-command data migration type when the destination does not support ApplyBatch command
### Search before asking

### Motivation

In https://github.com/apache/kvrocks/pull/2010, we support the new `APPLYBATCH` command for data migration scenarios. In https://github.com/apache/kvrocks/pull/2067, we use this command, to realize the new `raw-key-value` migration type. 

But the `raw-key-value` migration won't work if the target doesn't support the `APPLYBATCH` command, so we can add a fallback mechanism. We can use the old `redis-command` migration type when the destination does not support `ApplyBatch` command.

### Solution

_No response_

caipengbo:
This close #2078

### DIFF
--- a/src/cluster/slot_migrate.cc
+++ b/src/cluster/slot_migrate.cc
@@ -290,24 +290,36 @@ Status SlotMigrator::startMigration() {
     return s.Prefixed(errFailedToSetImportStatus);
   }
 
+  migration_type_ = srv_->GetConfig()->migrate_type;
+
+  // If the APPLYBATCH command is not supported on the destination,
+  // we will fall back to the redis-command migration type.
+  if (migration_type_ == MigrationType::kRawKeyValue) {
+    bool supported = GET_OR_RET(supporttedApplyBatchCommandOnDstNode(*dst_fd_));
+    if (!supported) {
+      LOG(INFO) << "APPLYBATCH command is not supported, use redis command for migration" << s.Msg();
+      migration_type_ = MigrationType::kRedisCommand;
+    }
+  }
+
   LOG(INFO) << "[migrate] Start migrating slot " << migrating_slot_ << ", connect destination fd " << *dst_fd_;
 
   return Status::OK();
 }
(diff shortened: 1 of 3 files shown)


src/cluster/slot_migrate.cc
```diff
@@ -485,6 +497,33 @@ Status SlotMigrator::setImportStatusOnDstNode(int sock_fd, int status) {
   return Status::OK();
 }
 
+StatusOr<bool> SlotMigrator::supporttedApplyBatchCommandOnDstNode(int sock_fd) {
+  std::string cmd = redis::ArrayOfBulkStrings({"command", "info", "applybatch——"});
+  auto s = util::SockSend(sock_fd, cmd);
+  if (!s.IsOK()) {
+    return s.Prefixed("failed to send command info to the destination node");
+  }
+
```
caipengbo, line 506: It's not too strict here, maybe we can support redis client lib in the future.

src/cluster/slot_migrate.cc
```diff
@@ -485,6 +497,33 @@ Status SlotMigrator::setImportStatusOnDstNode(int sock_fd, int status) {
   return Status::OK();
 }
 
+StatusOr<bool> SlotMigrator::supportedApplyBatchCommandOnDstNode(int sock_fd) {
+  std::string cmd = redis::ArrayOfBulkStrings({"command", "info", "applybatch——"});
```
PragmaTwice: `applybatch——`?

caipengbo: sorry, for local test :(

caipengbo:
When #2123 is merged (fix a bug about `COMMAND` command), the unit tests will be passed :)

PragmaTwice (approved):

git-hulk (approved):

PR #2119, opened 2024-02-26
Add support of RESP3 in Lua

### LINKED ISSUE #2035: Support RESP3 in Lua scripts

git-hulk:
This closes #2035


src/storage/scripting.cc
```diff
@@ -1009,12 +1015,30 @@ const char *RedisProtocolToLuaTypeAggregate(lua_State *lua, const char *reply, i
...
+    lua_newtable(lua);
+    for (j = 0; j < mbulklen; j++) {
+      lua_pushnumber(lua, j + 1);
+      p = RedisProtocolToLuaType(lua, p);
+      lua_settable(lua, -3);
+    }
+    return p;
+  }
+
+  // atype == '%' or atype == '~'
```
PragmaTwice: I think we need to do some check to make it more stable here, e.g. in the future we have a new type with a new starting character.

git-hulk: Yes, that makes a lot of sense. How about doing a check before pushing into lua stack like:

```
CHECK(atype == '%' || atype == '~')
if (atype == '%' || atype == '~') {
   // do something
}

```

PragmaTwice (approved):

PR #2120, opened 2024-02-27
Add the support of the XREADGROUP command

### LINKED ISSUE #1588: Add the support of the XREADGROUP command
### Search before asking

### Motivation

This issue intend to implement [XREADGROUP](https://redis.io/commands/xreadgroup/) command like redis.

### Solution

_No response_

Yangsx-1:
Close #1588 

Sorry for the late PR, this command is important for consumer group.

Please refer to [xreadgroup](https://redis.io/commands/xreadgroup/) and design discussion #1654 .


src/types/redis_stream.cc
```diff
@@ -927,6 +971,143 @@ rocksdb::Status Stream::Range(const Slice &stream_name, const StreamRangeOptions
...
+  }
+  if (s.IsNotFound()) {
+    int created_number = 0;
+    s = CreateConsumer(stream_name, group_name, consumer_name, &created_number);
+    if (!s.ok()) {
+      return s;
+    }
+  }
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
```
jihuayu: Look like we don't modify the stream metadata.
Can we only lock the consumer group?

src/types/redis_stream.cc
```diff
@@ -927,6 +971,143 @@ rocksdb::Status Stream::Range(const Slice &stream_name, const StreamRangeOptions
...
+        consumer_metadata.pending_number += 1;
+      }
+    }
+    if (maxid > consumergroup_metadata.last_delivered_id) {
+      consumergroup_metadata.last_delivered_id = maxid;
+    }
+  } else {
+    std::string next_version_prefix_key =
+        InternalKey(ns_key, "", metadata.version + 1, storage_->IsSlotIdEncoded()).Encode();
+    std::string prefix_key = InternalKey(ns_key, "", metadata.version, storage_->IsSlotIdEncoded()).Encode();
```
jihuayu: Can we only scan this consumer group PEL?
Items in the same consumer group should be consecutive.
```c++
std::string prefix_key = internalPelKeyFromGroupAndId(ns_key,metadata,group_name,StreamEntryID::Minimum());
std::string end_key = internalPelKeyFromGroupAndId(ns_key,metadata,group_name,StreamEntryID::Maximum());
```

jihuayu (commented):
I think the block-read logic has some problems.
When a new entry comes, all consumers in the consumer group will be awakened. Only one will return data, others return empty lists.
We should randomly wake up consumers of a consumer group.

**Of course, if it doesn't cause terrible problems we can also improve this issue in later PRs.**

src/types/redis_stream.cc
Yangsx-1: We can but i don't suggest this way, because there maybe other threads do something about the stream, stream's group and group's consumer, and other operations all handle the lock of the stream key, if we use a lock of group key here, some potential problems may exist.

torwig: For example, a stream can be deleted during the XREADGROUP if we obtain a lock only on a consumer group.

src/types/redis_stream.h
```diff
@@ -82,6 +85,11 @@ class Stream : public SubKeyScanner {
   std::string consumerNameFromInternalKey(rocksdb::Slice key) const;
   static std::string encodeStreamConsumerMetadataValue(const StreamConsumerMetadata &consumer_metadata);
   static StreamConsumerMetadata decodeStreamConsumerMetadataValue(const std::string &value);
+  std::string internalPelKeyFromGroupAndId(const std::string &ns_key, const StreamMetadata &metadata,
+                                           const std::string &group_name, const StreamEntryID &id);
+  StreamEntryID groupandidFromPelInternalkey(rocksdb::Slice key, std::string &group_name);
```
torwig: `groupandidFromPelInternalkey` => `groupAndEntryIdFromPelInternalKey`

src/types/redis_stream.h
```diff
@@ -82,6 +85,11 @@ class Stream : public SubKeyScanner {
   std::string consumerNameFromInternalKey(rocksdb::Slice key) const;
   static std::string encodeStreamConsumerMetadataValue(const StreamConsumerMetadata &consumer_metadata);
   static StreamConsumerMetadata decodeStreamConsumerMetadataValue(const std::string &value);
+  std::string internalPelKeyFromGroupAndId(const std::string &ns_key, const StreamMetadata &metadata,
```
torwig: `internalPelKeyFromGroupAndId` => `internalPelKeyFromGroupAndEntryId`

src/types/redis_stream.cc
torwig: BTW, @Yangsx-1 If you want to lock the stream, I believe it's better to obtain a lock **before** getting the metadata of that stream.

jihuayu: Ok. This is indeed a problem. We can think about how to improve the throughput of the stream later.

jihuayu (dismissed):

src/types/redis_stream.cc
```diff
@@ -927,6 +971,147 @@ rocksdb::Status Stream::Range(const Slice &stream_name, const StreamRangeOptions
...
+  std::string consumer_key = internalKeyFromConsumerName(ns_key, metadata, group_name, consumer_name);
+  std::string get_consumer_value;
+  s = storage_->Get(rocksdb::ReadOptions(), stream_cf_handle_, consumer_key, &get_consumer_value);
+  if (!s.ok() && !s.IsNotFound()) {
+    return s;
+  }
+  if (s.IsNotFound()) {
+    int created_number = 0;
+    lock.unlock();
+    s = CreateConsumer(stream_name, group_name, consumer_name, &created_number);
```
Yangsx-1: There will be a deadlock here, because CreateConsumer also requires a lock of stream key, i use unique_lock rather than Lockguard.

torwig: @Yangsx-1 Perhaps you need to move the code related to creating a consumer (but not acquiring a lock) to a helper function and use it here. Because unlock before creating a consumer is a hack and technically any operations can sneak between unlock->lock calls from other threads. What do you think?

Yangsx-1: That would be better. I thought many status may help check the result of some operations, but it's not that safe. I'll revise it.

jihuayu (dismissed):

src/types/redis_stream.cc
```diff
@@ -927,6 +977,145 @@ rocksdb::Status Stream::Range(const Slice &stream_name, const StreamRangeOptions
...
+        StreamEntryID entry_id = groupAndEntryIdFromPelInternalKey(iter->key(), tmp_group_name);
+        if (tmp_group_name != group_name) continue;
+        StreamPelEntry pel_entry = decodeStreamPelEntryValue(iter->value().ToString());
+        if (pel_entry.consumer_name != consumer_name) continue;
+        std::string raw_value;
+        rocksdb::Status st = getEntryRawValue(ns_key, metadata, entry_id, &raw_value);
+        if (!st.ok() && !st.IsNotFound()) {
+          return st;
+        }
+        std::vector<std::string> values;
```
torwig, line 1100: What will be if `st.IsNotFound()` is `true`?

torwig: Is it the case described in the section `What happens when a pending message is deleted?` here https://redis.io/commands/xreadgroup/?

Yangsx-1: Yes, it handles this situation, and i add a new test for it.

jihuayu (approved):

torwig (approved):

matchyc:
Impressive PR!

PR #2130, opened 2024-02-29
Add support of lua function 'redis.setresp()'

git-hulk:
Except for the `redis.setresp` function, it fixes the different behavior with Redis in Lua script.

Before applying this PR, the Lua script will return the result in RESP3 format
if the connection is connected with `HELLO 3` command. But for Redis,
it will always use RESP2 unless users explicitly set it with `redis.setresp(3)`.

```
// Kvrocks

❯ redis-cli -3 -p 6666

127.0.0.1:6666> EVAL 'return redis.call("hgetall","hash")' 0
1# "f1" => "v1"

// Redis

❯ redis-cli -3 
127.0.0.1:6379> EVAL 'return redis.call("hgetall","hash")' 0
1) "f1"
2) "v1"
```

After applying this PR, it will behaves consistently with Redis:

```
❯ redis-cli -3 -p 6666
127.0.0.1:6666> EVAL 'redis.setresp(3);return redis.call("hgetall","hash")' 0
1# "f1" => "v1"
127.0.0.1:6666> EVAL 'return redis.call("hgetall","hash")' 0
1) "f1"
2) "v1"
```

### DIFF
--- a/src/storage/scripting.cc
+++ b/src/storage/scripting.cc
@@ -87,6 +87,11 @@ void LoadFuncs(lua_State *lua, bool read_only) {
   lua_pushcfunction(lua, RedisPCallCommand);
   lua_settable(lua, -3);
 
+  /* redis.setresp */
+  lua_pushstring(lua, "setresp");
+  lua_pushcfunction(lua, RedisSetResp);
+  lua_settable(lua, -3);
+
   /* redis.log and log levels. */
   lua_pushstring(lua, "log");
   lua_pushcfunction(lua, RedisLogCommand);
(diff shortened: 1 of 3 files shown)


src/storage/scripting.cc
```diff
@@ -848,6 +853,29 @@ int RedisReturnSingleFieldTable(lua_State *lua, const char *field) {
...
+  }
+
+  int resp = static_cast<int>(lua_tonumber(lua, -1));
+  if (resp != 2 && resp != 3) {
+    PushError(lua, "RESP version must be 2 or 3.");
+    return 1;
+  }
+  conn->SetProtocolVersion(resp == 2 ? redis::RESP::v2 : redis::RESP::v3);
+  if (resp == 3 && !srv->GetConfig()->resp3_enabled) {
+    PushError(lua, "you need set resp_enabled to yes to enable RESP3.");
```
enjoy-binbin: It looks like there is no way we can use resp2 and resp3 at the same time?
For example, use resp2 for one lua and resp3 for another lua.

```suggestion
    PushError(lua, "You need set resp3-enabled to yes to enable RESP3.");
```

git-hulk: @enjoy-binbin Yes, it will support the RESP2 and RESP3 at them same time while the `resp3-enabled` was enabled. We now disable the RESP3 by default for stable consideration, so we need to check if the RESP3 was enabled here.

enjoy-binbin (dismissed):

torwig (dismissed):

enjoy-binbin (approved):

torwig (approved):

git-hulk:
Thanks for your review. @enjoy-binbin @torwig

PR #2142, opened 2024-03-07
feat(hyperloglog): add support of the Hyperloglog data structure

tutububug:
storage format description: https://github.com/apache/kvrocks-website/pull/207/files.

Only dense is supported now, and Merge is moved in https://github.com/apache/kvrocks/pull/2142/commits/d3b29781b1a0ab3a4b0ce94f7785df363a27d339 / https://github.com/apache/kvrocks/pull/2142/commits/22923f9ca3468dd9898fa16a38f04f8d75822fc7 and would be back in coming patches


src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,359 @@
...
+    const int r = 47;
+    uint64_t h = seed ^ (len * m);
+    const uint8_t *data = (const uint8_t *)key;
+    const uint8_t *end = data + (len-(len&7));
+
+    while(data != end) {
+        uint64_t k;
+
+#if (BYTE_ORDER == LITTLE_ENDIAN)
+    #ifdef USE_ALIGNED_ACCESS
```
mapleFU: Would you mind extract `USE_ALIGNED_ACCESS` from `redis_bitmap.cc`?

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,359 @@
...
+    unsigned long _fb = regnum*HLL_BITS&7; \
+    unsigned long _fb8 = 8 - _fb; \
+    unsigned long b0 = _p[_byte]; \
+    unsigned long b1 = _p[_byte+1]; \
+    target = ((b0 >> _fb) | (b1 << _fb8)) & HLL_REGISTER_MAX; \
+} while(0)
+
+/* ========================= HyperLogLog algorithm  ========================= */
+
+/* Our hash function is MurmurHash2, 64 bit version.
```
mapleFU: Should we just vendor murmurhash or use it under rocksdb? @PragmaTwice

mapleFU (commented):
I'll take a careful review this week

src/types/redis_hyperloglog.cc
tutububug: sorry, I do not understand ''extract" clearly, do you mean to add a cmake compile option for USE_ALIGNED_ACCESS?

mapleFU: Currently, `USE_ALIGNED_ACCESS` is defined here: https://github.com/apache/kvrocks/blob/83b3adfdeabec572ec00bb15e93eb6f4000dfb59/src/types/redis_bitmap.h#L32

It would be only used in `redis_bitmap.cc`. You can extract it to a common header or define it in cmake

tutububug: ok, got it.

torwig:
@tutububug Thank you for your contribution. Running `./x.py format` should eliminate the linting issues.

Yangsx-1:
Thank you for your contribution! Maybe you need to add a command parser to use the data structure with actual command like redis.

PragmaTwice:

Could you include your design in the PR description? For example, explain how to encode the metadata and HLL data (subkeys), similar to what is shown on https://kvrocks.apache.org/community/data-structure-on-rocksdb.

src/storage/redis_db.cc
```diff
@@ -99,10 +99,18 @@ rocksdb::Status Database::GetMetadata(RedisTypes types, const Slice &ns_key, std
...
 
 rocksdb::Status Database::GetRawMetadata(const Slice &ns_key, std::string *bytes) {
+  return getRaw(ns_key, bytes);
+}
+
+rocksdb::Status Database::GetRawData(const Slice &key, std::string *bytes) {
+  return getRaw(key, bytes);
+}
+
+rocksdb::Status Database::getRaw(const Slice &key, std::string *bytes) {
```
PragmaTwice: Could you explain a little about these newly-added methods?

src/storage/redis_metadata.h
```diff
@@ -329,3 +330,8 @@ class SearchMetadata : public Metadata {
   void Encode(std::string *dst) const override;
   rocksdb::Status Decode(Slice *input) override;
 };
+
+class HyperloglogMetadata: public Metadata {
+ public:
+  explicit HyperloglogMetadata(bool generate_version = true) : Metadata(kRedisHyperloglog, generate_version) {}
+};
```
PragmaTwice: Please put a newline to the end of source files : )

src/storage/redis_db.cc
tutububug: 'GetRawData' is added just because need to read 'sub_key' value as the method literal meaning, corresponds to 'GetRawMetadata'.

tutububug:
> Thank you for your contribution! Maybe you need to add a command parser to use the data structure with actual command like redis.

Yes, I will give the commit later.

> Thank you for your contribution!
> 
> Could you include your design in the PR description? For example, explain how to encode the metadata and HLL data (subkeys), similar to what is shown on https://kvrocks.apache.org/community/data-structure-on-rocksdb.

OK.

src/storage/redis_db.cc
tutububug: oh, but the column family should be specified, I'll fix it.

tutububug: The code changed here should be reset, it's not in use now.

src/types/redis_hyperloglog.cc
tutububug: Had moved to 'storage.h'.

tutububug:
@PragmaTwice I create a PR(https://github.com/apache/kvrocks-website/pull/207) for describe hyperloglog storage format.

PragmaTwice:
> @PragmaTwice I create a PR([apache/kvrocks-website#207](https://github.com/apache/kvrocks-website/pull/207)) for describe hyperloglog storage format.

Thank you!
Currently, it is only necessary to include your design in the PR description, not on the website as the design is not finalized yet.

Regarding your design, I have some questions:
1) Based on the current design, typically one Redis key will introduce a maximum of 16384 RocksDB keys (registers). Each value corresponding to a RocksDB key contains only one integer. This may be inefficient; merging multiple registers onto one key could reduce the number of keys introduced. WDYT?
2) I noticed that integers are stored as string representations (`std::to_string`) rather than their binary form (e.g., subkey and register value). What is the reason for this approach?
3) Having a constant `size` seems illogical since the number of subkeys linked to this Redis key varies. (However, if every write operation modifies the metadata, it may lead to a decrease in performance. I don't have a clear idea about this aspect.)

Concerning the code, although I haven't reviewed it thoroughly yet, there are some points worth mentioning:
1) The complete source code of MurmurHash can be placed in the `vendor` directory.
2) It appears that using `PFADD` in the code leads to an increasing number of RocksDB keys (registers). There seems to be no operation that reduces these keys until deleting this Redis key. How can we prevent an increase in RocksDB keys without a mechanism to decrease them?

cc @git-hulk @mapleFU

git-hulk:
@tutububug As @PragmaTwice mentioned in https://github.com/apache/kvrocks/pull/2142#issuecomment-1986756740, it's unnecessary to use a static number of 16384 since it may heavily affect the read performance while using PFMERGE, I guess a smaller one like 16 is enough and every subkey has 1000 integers.

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+  *ret = 0;
+  std::string ns_key = AppendNamespacePrefix(user_key);
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+  HyperloglogMetadata metadata;
+  rocksdb::Status s = GetMetadata(ns_key, &metadata);
+  if (!s.ok() && !s.IsNotFound()) return s;
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisHyperloglog);
```
git-hulk: `kRedisHyperloglog` should be `kRedisHyperLogLog` or `kRedisHLL`

src/common/status.h
```diff
@@ -168,7 +168,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> :
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
-      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
+      : BaseType((typename StringInStatusOr<U>::BaseType&&)(std::move(v))) {}
```
git-hulk: Need to use clang-format 14 to align with the GitHub Actions.

PragmaTwice:
I suggest that we can store the number of registers in one rocksdb key in the metadata (for example, referred to as `register_number_in_one_key`), so that even if adjustments are made later for performance reasons, the compatibility of kvrocks data can be maintained.

 Currently, a fixed value for `register_number_in_one_key` can be hardcoded in the code.

cc @tutububug

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+  std::string next_version_prefix = InternalKey(ns_key, "", metadata.version + 1, storage_->IsSlotIdEncoded()).Encode();
+
+  rocksdb::ReadOptions read_options;
+  LatestSnapShot ss(storage_);
+  read_options.snapshot = ss.GetSnapShot();
+  rocksdb::Slice upper_bound(next_version_prefix);
+  read_options.iterate_upper_bound = &upper_bound;
+
+  auto iter = util::UniqueIterator(storage_, read_options);
+  for (iter->Seek(prefix); iter->Valid() && iter->key().starts_with(prefix); iter->Next()) {
```
git-hulk: Can consider using `MultiGet` while the key's indexes are known, guess the performance will be better than the seek way.

mapleFU:
> Based on the current design, typically one Redis key will introduce a maximum of 16384 RocksDB keys (registers). Each value corresponding to a RocksDB key contains only one integer. This may be inefficient; merging multiple registers onto one key could reduce the number of keys introduced. WDYT?

There're two parts of things.

1. Bitmap type can be used to do the dense logic, which would reducing the keycount
2. sparse hll type can be introduced to save some overhead here.

PragmaTwice:
> > Based on the current design, typically one Redis key will introduce a maximum of 16384 RocksDB keys (registers). Each value corresponding to a RocksDB key contains only one integer. This may be inefficient; merging multiple registers onto one key could reduce the number of keys introduced. WDYT?
> 
> There're two parts of things.
> 
> 1. Bitmap type can be used to do the dense logic, which would reducing the keycount
> 2. sparse hll type can be introduced to save some overhead here.

So I think the question is, should we also introduce two mode of hll encoding (sparse and dense layout) and an auto switching policy between these two layout?

mapleFU:
>  So I think the question is, should we also introduce two mode of hll encoding (sparse and dense layout) and an auto switching policy between these two layout?

I prefer to do this :-) But we can regard it as a further optimization

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+
+#define HLL_P 14 /* The greater is P, the smaller the error. */
+#define HLL_Q                                                                                            \
+  (64 - HLL_P) /* The number of bits of the hash value used for determining the number of leading zeros. \
+                */
+#define HLL_REGISTERS (1 << HLL_P)     /* With P=14, 16384 registers. */
+#define HLL_P_MASK (HLL_REGISTERS - 1) /* Mask to index register. */
+#define HLL_BITS 8
+#define HLL_REGISTER_MAX ((1 << HLL_BITS) - 1)
+#define HLL_ALPHA_INF 0.721347520444481703680 /* constant for 0.5/ln(2) */
```
mapleFU: Would you mind:
1. define these variables as `constexpr` and functions, then avoid hack `P` `Q` naming? e.g:

```
constexpr size_t HyperLogLogBucketCount = 14;
constexpr size_t HyperLogLogBucketPayload = 64 - HyperLogLogBucketCount;
...
```

src/types/redis_hyperloglog.h
```diff
@@ -0,0 +1,44 @@
...
+  rocksdb::Status Count(const Slice &user_key, int *ret);
+  rocksdb::Status Merge(const std::vector<Slice> &user_keys);
+
+ private:
+  uint64_t hllCount(const std::vector<uint8_t> &counts);
+  void hllMerge(uint8_t *max, const std::vector<uint8_t> &counts);
+  rocksdb::Status getRegisters(const Slice &user_key, std::vector<uint8_t> *registers);
+
+  rocksdb::Status GetMetadata(const Slice &ns_key, HyperloglogMetadata *metadata);
+  int hllPatLen(unsigned char *ele, size_t elesize, long *regp);
```
mapleFU: could this be a static function?

src/types/redis_hyperloglog.h
```diff
@@ -0,0 +1,44 @@
...
+class Hyperloglog : public Database {
+ public:
+  explicit Hyperloglog(engine::Storage *storage, const std::string &ns) : Database(storage, ns) {}
+  rocksdb::Status Add(const Slice &user_key, const std::vector<Slice> &elements, int *ret);
+  rocksdb::Status Count(const Slice &user_key, int *ret);
+  rocksdb::Status Merge(const std::vector<Slice> &user_keys);
+
+ private:
+  uint64_t hllCount(const std::vector<uint8_t> &counts);
+  void hllMerge(uint8_t *max, const std::vector<uint8_t> &counts);
```
mapleFU: could this be a static function?

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+  HyperloglogMetadata metadata;
+  rocksdb::Status s = GetMetadata(ns_key, &metadata);
+  if (!s.ok() && !s.IsNotFound()) return s;
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisHyperloglog);
+  batch->PutLogData(log_data.Encode());
+  if (s.IsNotFound()) {
+    std::string bytes;
+    metadata.size = HLL_REGISTERS;  // 'size' must non-zone, or 'GetMetadata' will failed as 'expired'.
```
mapleFU: When could `metadata.size != HLL_REGISTERS`?

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+    metadata.Encode(&bytes);
+    batch->Put(metadata_cf_handle_, ns_key, bytes);
+  }
+  for (const auto &element : elements) {
+    long index = 0;
+    uint8_t count = hllPatLen((unsigned char *)element.data(), element.size(), &index);
```
mapleFU: why using `uint8_t` to accept a `int` as return value?

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+ *
+ * The hll object must be already validated via isHLLObjectOrReply()
+ * or in some other way.
+ *
+ * If the HyperLogLog is sparse and is found to be invalid, C_ERR
+ * is returned, otherwise the function always succeeds. */
+void Hyperloglog::hllMerge(uint8_t *max, const std::vector<uint8_t> &counts) {
+  int i;
+  uint8_t val;
+  auto registers = (uint8_t *)(&counts[0]);
```
mapleFU: nit

```suggestion
  uint8_t* registers = counts.data();
```

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+        InternalKey(ns_key, std::to_string(i), metadata.version, storage_->IsSlotIdEncoded()).Encode();
+
+    batch->Put(sub_key, std::to_string(max[i]));
+  }
+  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+/* Store the value of the register at position 'regnum' into variable 'target'.
+ * 'p' is an array of unsigned bytes. */
+#define HLL_DENSE_GET_REGISTER(target, p, regnum)             \
```
mapleFU: can you change it to a function?

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+  index = hash & HLL_P_MASK;      /* Register index. */
+  hash >>= HLL_P;                 /* Remove bits used to address the register. */
+  hash |= ((uint64_t)1 << HLL_Q); /* Make sure the loop terminates
+                                     and count will be <= Q+1. */
+  bit = 1;
+  count = 1; /* Initialized to 1 since we count the "00000...1" pattern. */
+  while ((hash & bit) == 0) {
+    count++;
+    bit <<= 1;
+  }
```
mapleFU: would you mind use counting leading zero instrinct ( clz )?

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+  *regp = (int)index;
+  return count;
+}
+
+/* Compute the register histogram in the dense representation. */
+void hllDenseRegHisto(uint8_t *registers, int *reghisto) {
```
mapleFU: prefer an `std::vector` or even `std::array` as second argument

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,401 @@
...
+    x = sqrt(x);
+    zPrime = z;
+    y *= 0.5;
+    z -= pow(1 - x, 2) * y;
+  } while (zPrime != z);
+  return z / 3;
+}
+
+/* Return the approximated cardinality of the set based on the harmonic
+ * mean of the registers values. 'hdr' points to the start of the SDS
```
mapleFU: we don't have redis sds, do we?

mapleFU:
Other Looks ok to me

tutububug:
> Regarding your design, I have some questions:
> 
>     1. Based on the current design, typically one Redis key will introduce a maximum of 16384 RocksDB keys (registers). Each value corresponding to a RocksDB key contains only one integer. This may be inefficient; merging multiple registers onto one key could reduce the number of keys introduced. WDYT?
> 
>     2. I noticed that integers are stored as string representations (`std::to_string`) rather than their binary form (e.g., subkey and register value). What is the reason for this approach?
> 
>     3. Having a constant `size` seems illogical since the number of subkeys linked to this Redis key varies. (However, if every write operation modifies the metadata, it may lead to a decrease in performance. I don't have a clear idea about this aspect.)
> 
> 
> Concerning the code, although I haven't reviewed it thoroughly yet, there are some points worth mentioning:
> 
>     1. The complete source code of MurmurHash can be placed in the `vendor` directory.
> 
>     2. It appears that using `PFADD` in the code leads to an increasing number of RocksDB keys (registers). There seems to be no operation that reduces these keys until deleting this Redis key. How can we prevent an increase in RocksDB keys without a mechanism to decrease them?
> 
> 
> cc @git-hulk @mapleFU

tutububug:
> Regarding your design, I have some questions:
> 
>     1. Based on the current design, typically one Redis key will introduce a maximum of 16384 RocksDB keys (registers). Each value corresponding to a RocksDB key contains only one integer. This may be inefficient; merging multiple registers onto one key could reduce the number of keys introduced. WDYT?
Not really. The register(subkey) only be stored which its count is not zero. This point is different from the memory implementation with static array as dense encode. On disk, I think its sparse encode naturally.
> 
>     2. I noticed that integers are stored as string representations (`std::to_string`) rather than their binary form (e.g., subkey and register value). What is the reason for this approach?
The number of consecutive 0s is calculated from the last 50 digits of the hash value, so the maximum value is 50, and the maximum value stored in a string is 2 bytes. It should not waste a lot of space, and at the same time save the overhead of integer encoding and decoding. For keys, it may be more efficient, but the largest index is only 5 bytes (16383).
> 
>     3. Having a constant `size` seems illogical since the number of subkeys linked to this Redis key varies. (However, if every write operation modifies the metadata, it may lead to a decrease in performance. I don't have a clear idea about this aspect.)
Currently, the ‘size’ variable has no practical purpose; the only requirement is that it be non-zero. Due to the implementation of kvrocks getmetadata, non-string type data structures with a size of 0 are judged to be expired, and the HLL add parameter that redis has implemented allows no parameters but the key will be stored. For compatibility, size is used as a constant to prevent expiration.
> 
> 
> Concerning the code, although I haven't reviewed it thoroughly yet, there are some points worth mentioning:
> 
>     1. The complete source code of MurmurHash can be placed in the `vendor` directory.
OK.
> 
>     2. It appears that using `PFADD` in the code leads to an increasing number of RocksDB keys (registers). There seems to be no operation that reduces these keys until deleting this Redis key. How can we prevent an increase in RocksDB keys without a mechanism to decrease them?
For an HLL user key, the maximum register value is 16384, and it cannot be larger. In fact, I think this should be considered controllable compared to data structures such as hash, set, list, etc. whose size is determined by user input.
> 
> 
@PragmaTwice  cc @git-hulk @mapleFU

mapleFU:
> The number of consecutive 0s is calculated from the last 50 digits of the hash value, so the maximum value is 50, and the maximum value stored in a string is 2 bytes. It should not waste a lot of space, and at the same time save the overhead of integer encoding and decoding. For keys, it may be more efficient, but the largest index is only 5 bytes (16383).

For rocksdb value, it's should be 1-2 bytes payload, the value size is also included. So, it introduce an extremly huge overhead. So I prefer the impl of bitmap/bitfield.

Get 2^14 times in rocksdb would also be heavy, and might break some statistics in rocksdb, which making it tent to compaction more or caching the unneccessary blocks.

PragmaTwice:
> The number of consecutive 0s is calculated from the last 50 digits of the hash value, so the maximum value is 50, and the maximum value stored in a string is 2 bytes. It should not waste a lot of space, and at the same time save the overhead of integer encoding and decoding. For keys, it may be more efficient, but the largest index is only 5 bytes (16383).

Sorry but I cannot get your point. In every sense, using a string representation cannot be more efficient. In binary representation: if the maximum value is less than 2^8, you should use one byte; and if the maximum value is less than 2^16, you should use two bytes.

> Not really. The register(subkey) only be stored which its count is not zero. This point is different from the memory implementation with static array as dense encode. On disk, I think its sparse encode naturally.

I don't think you answered my question. The maximum number of subkeys is still 2^16. I think multiple registers should be put into one subkey to reduce the number of keys and improve the efficiency of count.

tutububug:
@mapleFU @PragmaTwice Thank you for carefully code review. Since the code for the algorithm part comes from redis, some function arguments and comments need to be adjusted. Regarding optimizing key encoding and the number of subkeys, I will refer to your suggestions.
cc @git-hulk

git-hulk:
@tutububug Thanks for your great efforts.

mapleFU:
@tutububug FYI:
1. bitfield: https://github.com/apache/kvrocks/blob/unstable/src/common/bitfield_util.h#L33 . You can refer to these logic or rewrite the HLL logic using these yourself. A value in HLL using redis' impl would be about 13KB, we can storing it in one value. You can also refer to my bit utils: https://github.com/apache/kvrocks/blob/unstable/src/common/bit_util.h . It's not hard but maybe need some time to get familiar with it. Redis uses the similiar logic, but maybe not encapsulate them as tools.
2. Coding style: the redis using macro and some raw-pointers, for maintaining, we'd better using C++ style code
3. Protocol and metadata: currently, this patch only implements "dense" HLL. I'm ok with this. But maybe we can leave a field for "format" in metadata. For example

```
| ... | format-version (1Byte)|
```

And set "dense == 0" here.

Thank you for your patient and your code

PragmaTwice:
> A value in HLL using redis' impl would be about 13KB, we can storing it in one value. 

Do you mean that we can store the whole HLL data in one rocksdb key?

IMHO I think it is worth exploring how many rocksdb key value pairs is suitable to be stored here. 

Kvrocks does not have MVCC. Considering that write ops are very frequent for HLL, I think putting them into multiple keys may improve efficiency.


cc @mapleFU

mapleFU:
@PragmaTwice We can first implement that, and optimize writing to HLL step-by-step. Since 2.9 release would be a long time from now.

tutububug:
@mapleFU @PragmaTwice The code refactoring is ready for review, and the storage format description is sorted out at https://github.com/apache/kvrocks-website/pull/207/files. Please take a look if have time. 
cc @git-hulk @torwig

git-hulk:
Sure, @tutububug can help to fix the lint error while you're free: https://github.com/apache/kvrocks/actions/runs/8538011240/job/23657605010?pr=2142#step:7:1368

tutububug:
> Sure, @tutububug can help to fix the lint error while you're free: https://github.com/apache/kvrocks/actions/runs/8538011240/job/23657605010?pr=2142#step:7:1368

Strange, there is no such warning in my branch(https://github.com/tutububug/incubator-kvrocks/pull/6). I triggered the action again to check.

git-hulk:
@tutububug you can pull the unstable into this branch and have a try.

tutububug:
> @tutububug you can pull the unstable into this branch and have a try.
fixed

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,347 @@
...
+    if (!s.ok()) return s;
+    if (segment->size() == 0) {
+      *segment = registers;
+    }
+  }
+  cache.BatchForFlush(batch);
+  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+/* ========================= HyperLogLog algorithm  ========================= */
```
git-hulk: We can add a comment line to notice if those codes are generally modified from other repository like: https://github.com/apache/kvrocks/blob/unstable/src/common/rocksdb_crc32c.h#L10

git-hulk:
LGTM, one comment to check if we need to add a notice for the source code.

PragmaTwice:
@mapleFU Could you review it again?

git-hulk (dismissed):

git-hulk:
@mapleFU Can help to take a look while you get time. We can merge this PR if it looks good since it's been pending too long.

tests/cppunit/types/hyperloglog_test.cc
tutububug, line 1: Adding 'srand' here to make HLL elements completely random should be necessary, but it may cause larger error and the error range cannot be evaluated (now it is 5.1%), this will affect the stability of ci's unit test execution. So it can only be removed first.

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,349 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+// NOTE: this file is copy from redis's source: `src/hyperloglog.c`
```
PragmaTwice: Could we move these copied functions to a new file? e.g. `types/hyperloglog.h`, to distinguish from our own code.

tutububug: done

src/storage/redis_metadata.cc
```diff
@@ -489,3 +489,20 @@ rocksdb::Status SearchMetadata::Decode(Slice *input) {
...
+  Metadata::Encode(dst);
+  PutFixed8(dst, static_cast<uint8_t>(encode_type_));
+}
+
+rocksdb::Status HyperloglogMetadata::Decode(Slice *input) {
+  if (auto s = Metadata::Decode(input); !s.ok()) {
+    return s;
+  }
+
+  if (!GetFixed8(input, reinterpret_cast<uint8_t *>(&encode_type_))) {
```
mapleFU: @PragmaTwice Is this casting a valid cast? Since `encode_type_` supports less than `uint8_t`'s domain?

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,154 @@
...
+  rocksdb::ReadOptions read_options = storage_->DefaultScanOptions();
+  read_options.snapshot = ss.GetSnapShot();
+  rocksdb::Slice upper_bound(next_version_prefix);
+  read_options.iterate_upper_bound = &upper_bound;
+
+  auto iter = util::UniqueIterator(storage_, read_options);
+  for (iter->Seek(prefix); iter->Valid() && iter->key().starts_with(prefix); iter->Next()) {
+    InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+
+    int register_index = std::stoi(ikey.GetSubKey().ToString());
```
mapleFU: nit: kvrocks already has parseInt, we can unify using that rather than `std::stoi`

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,154 @@
...
+    if (register_index / kHyperLogLogRegisterCountPerSegment < 0 ||
+        register_index / kHyperLogLogRegisterCountPerSegment >= kHyperLogLogSegmentCount ||
+        register_index % kHyperLogLogRegisterCountPerSegment != 0) {
+      return rocksdb::Status::Corruption("invalid subkey index: idx=" + ikey.GetSubKey().ToString());
+    }
+    auto val = iter->value().ToString();
```
mapleFU: nit:

```
std::string_view val = iter->value();
```

This can avoid a around of materialize to `std::string`

src/types/redis_hyperloglog.h
```diff
@@ -0,0 +1,41 @@
...
+class HyperLogLog : public Database {
+ public:
+  explicit HyperLogLog(engine::Storage *storage, const std::string &ns) : Database(storage, ns) {}
+  rocksdb::Status Add(const Slice &user_key, const std::vector<Slice> &elements, uint64_t *ret);
+  rocksdb::Status Count(const Slice &user_key, uint64_t *ret);
+  rocksdb::Status Merge(const std::vector<Slice> &user_keys);
+
+ private:
+  rocksdb::Status GetMetadata(Database::GetOptions get_options, const Slice &ns_key, HyperloglogMetadata *metadata);
+  rocksdb::Status getRegisters(const Slice &user_key, std::vector<uint8_t> *registers);
```
mapleFU: Nit: you can add a todo, and optimizing the return value for `getRegisters`

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,154 @@
...
+    auto ele_str = element.ToString();
+    std::vector<uint8_t> ele(ele_str.begin(), ele_str.end());
+    uint8_t count = HllPatLen(ele, &register_index);
+    uint32_t segment_index = register_index / kHyperLogLogRegisterCountPerSegment;
+    uint32_t register_index_in_segment = register_index % kHyperLogLogRegisterCountPerSegment;
+
+    std::string *segment = nullptr;
+    auto s = cache.GetMut(segment_index, &segment);
+    if (!s.ok()) return s;
+    if (segment->size() == 0) {
```
mapleFU: if `segment->size() != 0`, should we check it's size is equal to `kHyperLogLogRegisterBytesPerSegment` before `HllDenseGetResgister`?

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,154 @@
...
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisHyperLogLog);
+  batch->PutLogData(log_data.Encode());
+
+  Bitmap::SegmentCacheStore cache(storage_, metadata_cf_handle_, ns_key, &metadata);
+  for (const auto &element : elements) {
+    uint32_t register_index = 0;
+    auto ele_str = element.ToString();
+    std::vector<uint8_t> ele(ele_str.begin(), ele_str.end());
```
mapleFU: nit: how about using `std::string_view` here?

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,256 @@
...
+   * as index). The max run can be 64-kHyperLogLogRegisterCountPow+1 = kHyperLogLogHashBitCount+1 bits.
+   *
+   * Note that the final "1" ending the sequence of zeroes must be
+   * included in the count, so if we find "001" the count is 3, and
+   * the smallest count possible is no zeroes at all, just a 1 bit
+   * at the first position, that is a count of 1.
+   *
+   * This may sound like inefficient, but actually in the average case
+   * there are high probabilities to find a 1 after a few iterations. */
+  hash = MurmurHash64A(element.data(), elesize, 0xadc83b19ULL);
```
mapleFU: Nit: we can extract this hash to a constant rather than hard code a magic number?

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,256 @@
...
+
+// NOTE: this file is copy from redis's source: `src/hyperloglog.c`
+
+#include "hyperloglog.h"
+
+#include "murmurhash2.h"
+
+/* Store the value of the register at position 'index' into variable 'val'.
+ * 'registers' is an array of unsigned bytes. */
+void HllDenseGetRegister(uint8_t *val, uint8_t *registers, uint32_t index) {
```
mapleFU:
```suggestion
 uint8_t HllDenseGetRegister(const uint8_t *registers, uint32_t index) {
```

What about changing to this?

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,256 @@
...
+  while ((hash & bit) == 0) {
+    count++;
+    bit <<= 1;
+  }
+  *register_index = (int)index;
+  return count;
+}
+
+/* Compute the register histogram in the dense representation. */
+void HllDenseRegHisto(uint8_t *registers, int *reghisto) {
```
mapleFU: registers -> const uint8_t*

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,256 @@
...
+  /* Redis default is to use 16384 registers 6 bits each. The code works
+   * with other values by modifying the defines, but for our target value
+   * we take a faster path with unrolled loops. */
+  if (kHyperLogLogRegisterCount == 16384 && kHyperLogLogBits == 6) {
+    uint8_t *r = registers;
+    unsigned long r0 = 0, r1 = 0, r2 = 0, r3 = 0, r4 = 0, r5 = 0, r6 = 0, r7 = 0, r8 = 0, r9 = 0, r10 = 0, r11 = 0,
+                  r12 = 0, r13 = 0, r14 = 0, r15 = 0;
+    for (auto j = 0; j < 1024; j++) {
```
mapleFU: This code is a bit hacking, can we change `63` to `kHyperLogLogRegisterMax` ?

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,256 @@
...
+  *register_index = (int)index;
+  return count;
+}
+
+/* Compute the register histogram in the dense representation. */
+void HllDenseRegHisto(uint8_t *registers, int *reghisto) {
+  /* Redis default is to use 16384 registers 6 bits each. The code works
+   * with other values by modifying the defines, but for our target value
+   * we take a faster path with unrolled loops. */
+  if (kHyperLogLogRegisterCount == 16384 && kHyperLogLogBits == 6) {
```
mapleFU: Can we remove `if` since else would never happens?

mapleFU (commented):
This looks much better than previous version. The remaining are mostly code-style problem

tutububug:
> This looks much better than previous version. The remaining are mostly code-style problem

fixed.

tests/cppunit/types/hyperloglog_test.cc
```diff
@@ -0,0 +1,121 @@
...
+  std::vector<uint64_t> cards(3);
+  ASSERT_TRUE(hll_->Count("hll0", &cards[0]).ok());
+  ASSERT_TRUE(hll_->Count("hll1", &cards[1]).ok());
+  ASSERT_TRUE(hll_->Count("hll2", &cards[2]).ok());
+
+  auto card = static_cast<double>(cards[0] + cards[1] + cards[2]);
+  auto realcard = static_cast<double>(realcard_vec.size());
+  double left = std::abs(card - realcard);
+  double right = card / 100 * 5;
+  ASSERT_TRUE(left < right) << "left : " << left << ", right: " << right;
```
PragmaTwice: IIRC there's an `ASSERT_LT`.

src/storage/redis_metadata.h
```diff
@@ -50,6 +50,7 @@ enum RedisType : uint8_t {
   kRedisBloomFilter = 9,
   kRedisJson = 10,
   kRedisSearch = 11,
+  kRedisHyperLogLog = 12,
```
PragmaTwice: You should also change `RedisTypeNames`.

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,253 @@
...
+/* Merge by computing MAX(registers_max[i],registers[i]) the HyperLogLog 'registers'
+ * with an array of uint8_t kHyperLogLogRegisterCount registers pointed by 'registers_max'. */
+void HllMerge(std::vector<uint8_t> *registers_max, const std::vector<uint8_t> &registers) {
+  uint8_t val = 0, max_val = 0;
+
+  for (uint32_t i = 0; i < kHyperLogLogRegisterCount; i++) {
+    val = HllDenseGetRegister(const_cast<uint8_t *>(registers.data()), i);
+    max_val = HllDenseGetRegister(reinterpret_cast<uint8_t *>(registers_max->data()), i);
+    if (val > max_val) {
+      HllDenseSetRegister(reinterpret_cast<uint8_t *>(registers_max->data()), i, val);
```
PragmaTwice: Why do we need the `reinterpret_cast` here?

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,253 @@
...
+}
+
+/* Merge by computing MAX(registers_max[i],registers[i]) the HyperLogLog 'registers'
+ * with an array of uint8_t kHyperLogLogRegisterCount registers pointed by 'registers_max'. */
+void HllMerge(std::vector<uint8_t> *registers_max, const std::vector<uint8_t> &registers) {
+  uint8_t val = 0, max_val = 0;
+
+  for (uint32_t i = 0; i < kHyperLogLogRegisterCount; i++) {
+    val = HllDenseGetRegister(const_cast<uint8_t *>(registers.data()), i);
+    max_val = HllDenseGetRegister(reinterpret_cast<uint8_t *>(registers_max->data()), i);
```
PragmaTwice: ditto

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,253 @@
...
+  int j = 0;
+  /* Note that reghisto size could be just kHyperLogLogHashBitCount+2, because kHyperLogLogHashBitCount+1 is
+   * the maximum frequency of the "000...1" sequence the hash function is
+   * able to return. However it is slow to check for sanity of the
+   * input: instead we history array at a safe size: overflows will
+   * just write data to wrong, but correctly allocated, places. */
+  int reghisto[64] = {0};
+
+  /* Compute register histogram */
+  HllDenseRegHisto(const_cast<uint8_t *>(registers.data()), reghisto);
```
PragmaTwice: ditto

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,253 @@
...
+  /* Estimate cardinality from register histogram. See:
+   * "New cardinality estimation algorithms for HyperLogLog sketches"
+   * Otmar Ertl, arXiv:1702.01284 */
+  double z = m * HllTau((m - reghisto[kHyperLogLogHashBitCount + 1]) / (double)m);
+  for (j = kHyperLogLogHashBitCount; j >= 1; --j) {
+    z += reghisto[j];
+    z *= 0.5;
+  }
+  z += m * HllSigma(reghisto[0] / (double)m);
+  e = static_cast<double>(llroundl(kHyperLogLogAlphaInf * m * m / z));
```
PragmaTwice: ditto

PragmaTwice (changes requested):
There is a lot of strange type casting in the code.

It is advisable to minimize the use of type casting, particularly `const_cast` and `reinterpret_cast`. It is recommended to consider modifying the function signature instead of relying solely on casts.

src/storage/redis_metadata.cc
PragmaTwice: I think it's fine here.

tutububug:
> There is a lot of strange type casting in the code.
> 
> It is advisable to minimize the use of type casting, particularly `const_cast` and `reinterpret_cast`. It is recommended to consider modifying the function signature instead of relying solely on casts.

fixed.

src/types/redis_hyperloglog.h
tutububug: Do you mean the output argument 'registers' needs to be modified?

tutububug (commented):
Had fixed.

mapleFU:
I just back from holidays and catch a cold. I'll take a carefully around this week. Would you mind me edit directly on this patch @tutububug

tutububug:
> I just back from holidays and catch a cold. I'll take a carefully around this week. Would you mind me edit directly on this patch @tutububug

OK, feel free to modify if needed. And also the description at https://github.com/apache/kvrocks-website/pull/207/files.

src/vendor/murmurhash2.h
```diff
@@ -0,0 +1,96 @@
...
+  const uint64_t m = 0xc6a4a7935bd1e995;
+  const int r = 47;
+  uint64_t h = seed ^ (len * m);
+  const auto *data = (const uint8_t *)key;
+  const uint8_t *end = data + (len - (len & 7));
+
+  while (data != end) {
+    uint64_t k = 0;
+
+#if (BYTE_ORDER == LITTLE_ENDIAN)
```
mapleFU: Maybe this header cannot require BYTE_ORDER macro

mapleFU (commented):
Looks much better than previous, I'll change and merge after update the proposal in https://github.com/apache/kvrocks-website/pull/207

src/vendor/murmurhash2.h
PragmaTwice: You can use `__BYTE_ORDER__` .

src/storage/redis_metadata.h
PragmaTwice: Seems we can delete kRedisSearch since it's not used anymore.

src/storage/redis_metadata.h
```diff
@@ -49,6 +49,8 @@ enum RedisType : uint8_t {
   kRedisStream = 8,
   kRedisBloomFilter = 9,
   kRedisJson = 10,
+  kRedisSearch = 11,
```
PragmaTwice: need to delete

src/storage/redis_metadata.h
```diff
@@ -90,8 +92,9 @@ enum RedisCommand {
   kRedisCmdLMove,
 };
 
-const std::vector<std::string> RedisTypeNames = {"none",   "string",    "hash",   "list",      "set",      "zset",
-                                                 "bitmap", "sortedint", "stream", "MBbloom--", "ReJSON-RL"};
+const std::vector<std::string> RedisTypeNames = {"none",      "string", "hash",       "list",   "set",
+                                                 "zset",      "bitmap", "sortedint",  "stream", "MBbloom--",
+                                                 "ReJSON-RL", "search", "hyperloglog"};
```
PragmaTwice: ditto (for `"search"`)

src/storage/redis_metadata.h
```diff
@@ -313,3 +316,39 @@ class JsonMetadata : public Metadata {
...
+class SearchMetadata : public Metadata {
+ public:
+  SearchOnDataType on_data_type;
+
+  explicit SearchMetadata(bool generate_version = true) : Metadata(kRedisSearch, generate_version) {}
+
+  void Encode(std::string *dst) const override;
+  rocksdb::Status Decode(Slice *input) override;
+};
+
```
PragmaTwice: ditto

src/commands/cmd_hll.cc
```diff
@@ -0,0 +1,81 @@
...
+    if (s.IsNotFound()) {
+      ret = 0;
+    }
+    *output = redis::Integer(ret);
+    return Status::OK();
+  }
+};
+
+REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandPfAdd>("pfadd", -2, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandPfCount>("pfcount", 2, "read-only", 1, 1, 1), );
```
PragmaTwice, line 79: How about `PFMERGE`?

src/storage/redis_metadata.cc
```diff
@@ -472,3 +472,44 @@ rocksdb::Status JsonMetadata::Decode(Slice *input) {
...
+  if (auto s = Metadata::Decode(input); !s.ok()) {
+    return s;
+  }
+
+  if (!GetFixed8(input, reinterpret_cast<uint8_t *>(&on_data_type))) {
+    return rocksdb::Status::InvalidArgument(kErrMetadataTooShort);
+  }
+
+  return rocksdb::Status::OK();
+}
```
PragmaTwice: need to delete

src/types/hyperloglog.cc
```diff
@@ -0,0 +1,246 @@
...
+   * the smallest count possible is no zeroes at all, just a 1 bit
+   * at the first position, that is a count of 1.
+   *
+   * This may sound like inefficient, but actually in the average case
+   * there are high probabilities to find a 1 after a few iterations. */
+  uint32_t index = hash & kHyperLogLogRegisterCountMask; /* Register index. */
+  DCHECK_LT(index, kHyperLogLogRegisterCount);
+  hash >>= kHyperLogLogRegisterCountPow; /* Remove bits used to address the register. */
+  hash |= (static_cast<uint64_t>(1U) << kHyperLogLogHashBitCount);
+  uint8_t ctz = __builtin_ctzll(hash) + 1;
```
mapleFU, line 106: This is a little difference from valkey code, I've send a pr to them

src/types/redis_hyperloglog.cc
```diff
@@ -0,0 +1,224 @@
...
+    std::string data;
+    bool dirty;
+  };
+  std::map<uint32_t, SegmentEntry> segments;
+
+  /// Get the segment from cache or storage.
+  ///
+  /// If the segment in not in the cache and storage, it will be initialized with
+  /// string(kHyperLogLogSegmentBytes, 0) and return OK.
+  rocksdb::Status Get(uint32_t segment_index,
```
PragmaTwice: here we can avoid using std::function by template parameters.

src/types/redis_hyperloglog.h
```diff
@@ -0,0 +1,44 @@
...
+#include "storage/redis_metadata.h"
+
+namespace redis {
+
+class HyperLogLog : public Database {
+ public:
+  explicit HyperLogLog(engine::Storage *storage, const std::string &ns) : Database(storage, ns) {}
+  rocksdb::Status Add(const Slice &user_key, const std::vector<uint64_t> &element_hashes, uint64_t *ret);
+  rocksdb::Status Count(const Slice &user_key, uint64_t *ret);
+  rocksdb::Status Merge(const std::vector<Slice> &user_keys);
```
PragmaTwice: hmm I didn't find its definition 🤔

mapleFU: Aha, I've update the pr description, I remove merge related code to make it simple, I'll add it back in next pr

src/commands/cmd_hll.cc
mapleFU: Would be add in next patch

src/storage/redis_metadata.h
```diff
@@ -49,6 +49,7 @@ enum RedisType : uint8_t {
   kRedisStream = 8,
   kRedisBloomFilter = 9,
   kRedisJson = 10,
+  kRedisHyperLogLog = 12,
```
PragmaTwice: It can be 11 now?

PragmaTwice (dismissed):

git-hulk:
I think it's good to merge after fixing the lint issues.

src/commands/cmd_hll.cc
PragmaTwice, line 79: the implementation is not consistent with `PFCOUNT key [key ...]` in the comment.

mapleFU: You can re-check the todo in the pfcount command, without merge we cannot support multiple keys

git-hulk (approved):

PragmaTwice (approved):

mapleFU:
Merged, thanks all!

PR #2145, opened 2024-03-09
Extract common/port.h and optimize dbstats

mapleFU:
This patch does two things:
1. Extract a common/port.h, which handles difference between platforms
2. DBStats is used heavily in kvrocks. So, we need to enhance it's performance. This patch uses fetch_add and aligned it with cacheline

### DIFF
--- a/src/common/port.h
+++ b/src/common/port.h
@@ -25,24 +25,15 @@
 #endif
 
 #ifndef CACHE_LINE_SIZE
-// To test behavior with non-native cache line size, e.g. for
-// Bloom filters, set TEST_CACHE_LINE_SIZE to the desired test size.
-// This disables ALIGN_AS to keep it from failing compilation.
-#ifdef TEST_CACHE_LINE_SIZE
-#define CACHE_LINE_SIZE TEST_CACHE_LINE_SIZE
-#define ALIGN_AS(n) /*empty*/
-#else
 #if defined(__s390__)
 #if defined(__GNUC__) && __GNUC__ < 7
-#define CACHE_LINE_SIZE 64U
+constexpr size_t CACHE_LINE_SIZE = 64U;
 #else
-#define CACHE_LINE_SIZE 256U
+constexpr size_t CACHE_LINE_SIZE = 256U;
 #endif
 #elif defined(__powerpc__) || defined(__aarch64__)
-#define CACHE_LINE_SIZE 128U
+constexpr size_t CACHE_LINE_SIZE = 128U;
 #else
-#define CACHE_LINE_SIZE 64U
-#endif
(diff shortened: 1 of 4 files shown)


src/storage/storage.h
```diff
@@ -107,18 +108,18 @@ inline const std::vector<CacheOption> CacheOptions = {
...
   KeyspaceHits,
   KeyspaceMisses,
 };
 
-struct DBStats {
-  std::atomic<uint64_t> compaction_count = 0;
-  std::atomic<uint64_t> flush_count = 0;
-  std::atomic<uint64_t> keyspace_hits = 0;
-  std::atomic<uint64_t> keyspace_misses = 0;
+struct alignas(CACHE_LINE_SIZE) DBStats {
```
mapleFU: Like rocksdb, an Per core array optimization can also be introduced here. Maybe I'll try it in another patch.

mapleFU: Another problem is that, should we change `DBStats` to `unique_ptr<DBStats>` in Storage? @git-hulk @PragmaTwice

git-hulk: I feel good about both of them since the lifetime of DBStats is the same as the storage.

git-hulk (dismissed):
LGTM, good patch.

src/common/port.h
```diff
@@ -0,0 +1,39 @@
...
+#define USE_ALIGNED_ACCESS
+#endif
+
+#ifndef CACHE_LINE_SIZE
+#if defined(__s390__)
+#if defined(__GNUC__) && __GNUC__ < 7
+constexpr size_t CACHE_LINE_SIZE = 64U;
+#else
+constexpr size_t CACHE_LINE_SIZE = 256U;
+#endif
```
PragmaTwice, line 32: I wonder why it will be changed for different GCC versions?

mapleFU: The code is borrow from rocksdb, seems it's a bug in gcc, we can ingoring it?
https://github.com/facebook/rocksdb/commit/65996dd757b5d4c073c557f2841fc9d3da264b80

src/storage/storage.h
mapleFU: Emm I think previously it's size is 8*8B=64B, now `sizeof(DBStats)` becomes 8 * 64B. Which would be a large object

mapleFU:
Performance for sadd(On my WSL2 Ubuntu22, with gcc11.4, CPU AMD 3800X) with default release build:

```
before:
SADD: 27173.91 requests per second

after:
SADD: 30211.48 requests per second
```

cc @git-hulk @PragmaTwice

PragmaTwice (approved):

PR #2148, opened 2024-03-09
Fix build warnings: local variable 's' will be copied despite being returned by name

aleksraiden:
Fix warning at build like:

```
[ 85%] Building CXX object CMakeFiles/kvrocks_objs.dir/src/cluster/slot_import.cc.o
/Users/runner/work/kvrocks/kvrocks/src/cluster/replication.cc:78:10: warning: local variable 's' will be copied despite being returned by name [-Wreturn-std-move]
  return s;
         ^
/Users/runner/work/kvrocks/kvrocks/src/cluster/replication.cc:78:10: note: call 'std::move' explicitly to avoid copying
  return s;
         ^
         std::move(s)
[ 85%] Building CXX object CMakeFiles/kvrocks_objs.dir/src/cluster/slot_migrate.cc.o
[ 85%] Building CXX object CMakeFiles/kvrocks_objs.dir/src/cluster/sync_migrate_context.cc.o
```

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -75,7 +75,7 @@ Status FeedSlaveThread::Start() {
     conn_ = nullptr;  // prevent connection was freed when failed to start the thread
   }
 
-  return s;
+  return std::move(s);
 }
 
 void FeedSlaveThread::Stop() {
--- a/src/cluster/replication.h
+++ b/src/cluster/replication.h
@@ -175,7 +175,7 @@ class ReplicationThread : private EventCallbackBase<ReplicationThread> {
   enum IncrementBatchLoopState {
     Incr_batch_size,
     Incr_batch_data,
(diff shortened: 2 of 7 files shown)


PragmaTwice (approved):

git-hulk (approved):

src/cluster/replication.cc
```diff
@@ -75,7 +75,7 @@ Status FeedSlaveThread::Start() {
     conn_ = nullptr;  // prevent connection was freed when failed to start the thread
   }
 
-  return s;
+  return std::move(s);
```
mapleFU, line 78: Hmm C++17 would applying nrvo for this ? @PragmaTwice

PragmaTwice: The type of returned value is not the same as the return type. So no.

mapleFU (approved):

PR #2150, opened 2024-03-10
Fix missing fields in HELLO command response

git-hulk:
Currently, we're missing the 'version', 'role', 'modules' in the HELLO command which may cause the compatibility issue in some clients like Java lettuce:

```
Caused by: java.lang.IllegalArgumentException: Version must not be null
	at io.lettuce.core.internal.LettuceAssert.notNull(LettuceAssert.java:71)
	at io.lettuce.core.RedisHandshake$RedisVersion.<init>(RedisHandshake.java:333)
	at io.lettuce.core.RedisHandshake$RedisVersion.of(RedisHandshake.java:362)
	at io.lettuce.core.RedisHandshake.applyPostHandshake(RedisHandshake.java:249)
	at io.lettuce.core.RedisHandshake.lambda$initialize$0(RedisHandshake.java:99)
```

For the HELLO command response refer: https://redis.io/commands/hello/

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -800,6 +800,9 @@ class CommandHello final : public Commander {
     std::vector<std::string> output_list;
     output_list.push_back(redis::BulkString("server"));
     output_list.push_back(redis::BulkString("redis"));
+    output_list.push_back(redis::BulkString("version"));
+    // What the client want is the Redis compatible version instead of the Kvrocks version.
+    output_list.push_back(redis::BulkString(REDIS_VERSION));
     output_list.push_back(redis::BulkString("proto"));
     if (srv->GetConfig()->resp3_enabled) {
       output_list.push_back(redis::Integer(protocol));
@@ -815,6 +818,11 @@ class CommandHello final : public Commander {
     } else {
       output_list.push_back(redis::BulkString("standalone"));
     }
+    output_list.push_back(redis::BulkString("role"));
+    output_list.push_back(redis::BulkString(srv->IsSlave() ? "slave" : "master"));
+    // For Kvrocks, the modules is not supported.
+    output_list.push_back(redis::BulkString("modules"));
+    output_list.push_back(conn->NilArray());
     *output = conn->HeaderOfMap(output_list.size() / 2);
     for (const auto &item : output_list) {
       *output += item;
--- a/src/server/server.h
+++ b/src/server/server.h
@@ -53,6 +53,8 @@
 #include "tls_util.h"
 #include "worker.h"
 
+constexpr const char *REDIS_VERSION = "4.0.0";
+
 struct DBScanInfo {
   time_t last_scan_time = 0;
   KeyNumStats key_num_stats;
--- a/tests/gocase/unit/hello/hello_test.go
+++ b/tests/gocase/unit/hello/hello_test.go
@@ -72,8 +72,8 @@ func TestHello(t *testing.T) {
 	t.Run("hello with non protocol", func(t *testing.T) {
 		r := rdb.Do(ctx, "HELLO", "2", "SETNAME", "kvrocks")
 		rList := r.Val().([]interface{})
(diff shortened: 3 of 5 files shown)


git-hulk:
This issue was reported by the user `Asad Awadia` in Slack.

torwig (dismissed):

asad-awadia:
Thank you for the prompt fix

What is the nightly docker img - i can try it out

git-hulk:
> Thank you for the prompt fix
> 
> What is the nightly docker img - i can try it out

@asad-awadia Need to wait for this PR to be merged, and then you can have a try at `apache/kvrocks:nightly`.

src/server/server.h
```diff
@@ -53,6 +53,8 @@
 #include "tls_util.h"
 #include "worker.h"
 
+constexpr const char *REDIS_VERSION = "4.0.0";
```
asad-awadia, line 56: why so far back? is kvrocks compatible with up to redis 4 ? not more than that?

git-hulk: @asad-awadia It's the least promised compatibility version, we also would like to retrospect if need to update this version since most 6.0/7.0 commands are supported now. But it'd be better to discuss this in another place.

torwig (approved):

tests/gocase/unit/hello/hello_test.go
```diff
@@ -44,15 +44,19 @@ func TestHello(t *testing.T) {
...
+		require.EqualValues(t, 2, rList[5])
 	})
 
 	t.Run("hello with protocol 3", func(t *testing.T) {
 		r := rdb.Do(ctx, "HELLO", "3")
 		rList := r.Val().([]interface{})
-		require.EqualValues(t, rList[2], "proto")
-		require.EqualValues(t, rList[3], 2)
+		require.EqualValues(t, "version", rList[2])
+		require.EqualValues(t, "4.0.0", rList[3])
```
jihuayu, line 57: Would it be better to use regular expressions here instead of specific versions?
such as `^\d+\.\d+\.\d+$`

jihuayu (approved):

tests/gocase/unit/hello/hello_test.go
git-hulk: @jihuayu Yes, but maybe we can update this when we decide to change the Redis compatibility version.

mapleFU:
Thanks for catching this!

src/commands/cmd_server.cc
```diff
@@ -815,6 +818,11 @@ class CommandHello final : public Commander {
     } else {
       output_list.push_back(redis::BulkString("standalone"));
     }
+    output_list.push_back(redis::BulkString("role"));
+    output_list.push_back(redis::BulkString(srv->IsSlave() ? "slave" : "master"));
+    // For Kvrocks, the modules is not supported.
+    output_list.push_back(redis::BulkString("modules"));
+    output_list.push_back(conn->NilArray());
```
PragmaTwice, line 825: I don't think it should be an empty array. We DO support some redis modules.

cc @git-hulk

git-hulk: @PragmaTwice Yes, we did support some modules like JSON, but the mechanism is different from the Redis modules. For the Redis HELLO command, the module field needs to return the module library path and integer version which NOT exist in Kvrocks, so I kept this as an empty array.

```
   1) 1# "name" => "ReJSON"
      2# "ver" => (integer) 20606
      3# "path" => "/opt/redis-stack/lib/rejson.so"
      4# "args" => (empty array)
```

PragmaTwice: Ahh I see.

asad-awadia:
```
Status: Downloaded newer image for apache/kvrocks:nightly
I20240311 13:09:12.585850     1 main.cc:142] kvrocks unstable (commit 86d34ac)
```

Tried the nightly image and works as expected with resp3 default in lettuce :)

PR #2152, opened 2024-03-11
Add arm64 macOS support for kvrocks CI

jihuayu:
Now GitHub action has M1 macOS runner now.
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/


PragmaTwice:
The changes have been introduced in https://github.com/apache/kvrocks/tree/actions-m1.

Also the bug has already been fixed.

PR #2169, opened 2024-03-14
Add support of the XACK command

### LINKED ISSUE #1732: Add support of the XACK command
### Search before asking

### Motivation

This issue wants to implement [XACK](https://redis.io/commands/xack/) command like redis.

### Solution

_No response_

Yangsx-1:
Close #1732 

Refer to [XACK](https://redis.io/commands/xack/).

### DIFF
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -31,6 +31,41 @@
 
 namespace redis {
 
+class CommandXAck : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    stream_name_ = args[1];
+    group_name_ = args[2];
+    StreamEntryID tmp_id;
+    for (int i = 3; i < args.size(); ++i) {
+      auto s = ParseStreamEntryID(args[i], &tmp_id);
+      if (!s.IsOK()) {
+        return {Status::RedisParseErr, s.Msg()};
+      }
+      entry_set_.emplace_back(tmp_id);
+    }
+
+    return Status::OK();
+  }
+
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    redis::Stream stream_db(srv->storage, conn->GetNamespace());
+    uint64_t acknowledged = 0;
+    auto s = stream_db.DeletePelEntries(stream_name_, group_name_, entry_set_, acknowledged);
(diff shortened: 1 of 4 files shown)


jihuayu (dismissed):

src/commands/cmd_stream.cc
```diff
@@ -31,6 +31,41 @@
...
+      entry_set_.emplace_back(tmp_id);
+    }
+
+    return Status::OK();
+  }
+
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    redis::Stream stream_db(srv->storage, conn->GetNamespace());
+    uint64_t acknowledged = 0;
+    auto s = stream_db.DeletePelEntries(stream_name_, group_name_, entry_set_, acknowledged);
```
torwig: Maybe here we can pass the `acknowledged` value as a pointer instead of a reference, just to be consistent across the codebase - we usually pass output parameters by the pointer as far as I can recall.

src/commands/cmd_stream.cc
```diff
@@ -31,6 +31,41 @@
...
+    }
+    *output = redis::Integer(acknowledged);
+
+    return Status::OK();
+  }
+
+ private:
+  std::string stream_name_;
+  std::string group_name_;
+  std::vector<StreamEntryID> entry_set_;
```
torwig: You can name this variable just `entries_` (or even better - `entry_ids_`) without suffix `set/list/map/etc` - because currently the type of the variable is a std::vector but the name has the `set` as a suffix.

src/types/redis_stream.cc
```diff
@@ -314,6 +314,48 @@ StreamSubkeyType Stream::identifySubkeyType(const rocksdb::Slice &key) const {
   return StreamSubkeyType::StreamConsumerMetadata;
 }
 
+rocksdb::Status Stream::DeletePelEntries(const Slice &stream_name, const std::string &group_name,
+                                         const std::vector<StreamEntryID> &entry_set, uint64_t &acknowledged) {
```
torwig: And here we can change `entry_set` to `entry_ids`.

torwig (approved):

jihuayu (approved):

PR #2171, opened 2024-03-14
Add compression level option

### LINKED ISSUE #2143: add options for set compress level
### Search before asking

### Motivation

https://github.com/facebook/rocksdb/wiki/Space-Tuning#compression-level
Default Zstandard compression level is 3 and it is also a default in RocksDB. Zstandard allows to set compression level from 1 to 19 and it trades compression speed and ratio (https://github.com/facebook/zstd). In one typical DB, we use level 6 and in general we use “compression_opts=-14:6:0”.

### Solution

_No response_

Beihao-Zhou:
close: #2143

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -932,5 +932,3 @@ rocksdb.rate_limiter_auto_tuned yes
 
 ################################ NAMESPACE #####################################
 # namespace.test change.me
-backup-dir /tmp/kvrocks/backup
-pidfile /tmp/kvrocks/kvrocks.pid
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -727,7 +727,9 @@ rocksdb.cache_index_and_filter_blocks yes
 # default snappy
 rocksdb.compression snappy
 
-# Specify the compression level to use.
+# Specify the compression level to use. It trades compression speed 
+#   and ratio, might be useful when tuning for disk space.
+#   See details: https://github.com/facebook/rocksdb/wiki/Space-Tuning
 # For zstd: valid range is from 1 to 19, 
 #   where 1 is fastest compression and 19 is best compression.
 # For zlib: valid range is from 1 to 9, 
@@ -743,7 +745,7 @@ rocksdb.compression snappy
 #   - kLZ4: -1 (i.e., `acceleration=1`; see `CompressionOptions::level` doc)
 # For all others, RocksDB does not specify a compression level. 
 # If the compression type doesn't support the setting, it will be a no-op.
-# Default: 6 DB space tuning
+# Default: 6 (DB space tuning)
 rocksdb.compression_level 6
 
 # If non-zero, we perform bigger reads when doing compaction. If you're
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -727,6 +727,25 @@ rocksdb.cache_index_and_filter_blocks yes
 # default snappy
 rocksdb.compression snappy
 
+# Specify the compression level to use.
+# For zstd: valid range is from 1 to 19, 
+#   where 1 is fastest compression and 19 is best compression.
+# For zlib: valid range is from 1 to 9, 
+#   where 1 is fastest compression and 9 is best compression.
+# For lz4: adjusting the level influences the 'acceleration'. 
+#   RocksDB sets a negative level to indicate acceleration directly, 
+#   with more negative values indicating higher speed and less compression.
+# Note: This setting is ignored for compression algorithms like Snappy that 
+#   do not support variable compression levels.
+# RocksDB Default: 
+#   - zstd: 3
+#   - zlib: Z_DEFAULT_COMPRESSION (currently -1)
+#   - kLZ4: -1 (i.e., `acceleration=1`; see `CompressionOptions::level` doc)
+# For all others, RocksDB does not specify a compression level. 
+# If the compression type doesn't support the setting, it will be a no-op.
+# Default: 6 DB space tuning
+rocksdb.compression_level 6
+
 # If non-zero, we perform bigger reads when doing compaction. If you're
 # running RocksDB on spinning disks, you should set this to at least 2MB.
 # That way RocksDB's compaction is doing sequential instead of random reads.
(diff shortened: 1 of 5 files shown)


kvrocks.conf
```diff
@@ -911,3 +930,5 @@ rocksdb.rate_limiter_auto_tuned yes
 
 ################################ NAMESPACE #####################################
 # namespace.test change.me
+backup-dir /tmp/kvrocks/backup
+pidfile /tmp/kvrocks/kvrocks.pid
```
jihuayu: These two lines should not be committed.

jihuayu (commented):
@Beihao-Zhou Great!
Just one small issue, the rest is LGTM.

kvrocks.conf
git-hulk: Yes, I will take a look at why those two lines are generated after running.

kvrocks.conf
```diff
@@ -727,6 +727,25 @@ rocksdb.cache_index_and_filter_blocks yes
...
+#   with more negative values indicating higher speed and less compression.
+# Note: This setting is ignored for compression algorithms like Snappy that 
+#   do not support variable compression levels.
+# RocksDB Default: 
+#   - zstd: 3
+#   - zlib: Z_DEFAULT_COMPRESSION (currently -1)
+#   - kLZ4: -1 (i.e., `acceleration=1`; see `CompressionOptions::level` doc)
+# For all others, RocksDB does not specify a compression level. 
+# If the compression type doesn't support the setting, it will be a no-op.
+# Default: 6 DB space tuning
```
caipengbo: `DB space tuning`  What does that mean?

kvrocks.conf
```diff
@@ -727,6 +727,25 @@ rocksdb.cache_index_and_filter_blocks yes
 # default snappy
 rocksdb.compression snappy
 
+# Specify the compression level to use.
```
caipengbo: I thought it would be better to add: `It trades compression speed and ratio, might be useful when tuning for disk space`.

Or we attach https://github.com/facebook/rocksdb/wiki/Space-Tuning link to guide users.

kvrocks.conf
Beihao-Zhou: Removed!

kvrocks.conf
Beihao-Zhou: Sure, add both comment and link in the updated PR :)

kvrocks.conf
```diff
@@ -727,6 +727,27 @@ rocksdb.cache_index_and_filter_blocks yes
...
+# Note: This setting is ignored for compression algorithms like Snappy that 
+#   do not support variable compression levels.
+# RocksDB Default: 
+#   - zstd: 3
+#   - zlib: Z_DEFAULT_COMPRESSION (currently -1)
+#   - kLZ4: -1 (i.e., `acceleration=1`; see `CompressionOptions::level` doc)
+# For all others, RocksDB does not specify a compression level. 
+# If the compression type doesn't support the setting, it will be a no-op.
+# Default: 6 (DB space tuning)
+rocksdb.compression_level 6
```
mapleFU: https://github.com/facebook/rocksdb/blob/096fb9b67d19a9a180e7c906b4a0cdb2b2d0c1f6/include/rocksdb/compression_type.h#L43-L68

Why setting it to 6 by default？Though the doc says 6 is space tuning ( https://github.com/facebook/rocksdb/wiki/Space-Tuning#compression-level ), I think we should keep it "default in rocksdb" by default

jihuayu (dismissed):

kvrocks.conf
caipengbo: I agree. I also think we should use the rocksdb default value.

Beihao-Zhou: Sure thing! Set `kDefaultCompressionLevel = 32767` as default and add comment

caipengbo (approved):

src/config/config.cc
```diff
@@ -190,6 +190,7 @@ Config::Config() {
       {"rocksdb.compression", false,
        new EnumField<rocksdb::CompressionType>(&rocks_db.compression, compression_types,
                                                rocksdb::CompressionType::kNoCompression)},
+      {"rocksdb.compression_level", true, new IntField(&rocks_db.compression_level, 32767, INT_MIN, INT_MAX)},
```
mapleFU, line 193: 🤔why compression_level can be dynamically changed?

Beihao-Zhou: Seems like RocksDB just hardcode this value 32767 into their compression level logic. 
exposed level option const: https://github.com/facebook/rocksdb/blob/096fb9b67d19a9a180e7c906b4a0cdb2b2d0c1f6/include/rocksdb/compression_type.h#L53
Zlib: https://github.com/facebook/rocksdb/blob/096fb9b67d19a9a180e7c906b4a0cdb2b2d0c1f6/util/compression.h#L805
Zstd: https://github.com/facebook/rocksdb/blob/096fb9b67d19a9a180e7c906b4a0cdb2b2d0c1f6/util/compression.h#L183

mapleFU: aha, my bad. I mean why it's different from compression, which could be changed dynamically.

This looks ok to me

mapleFU (approved):

src/config/config.cc
Beihao-Zhou: Oh gotcha, nwnw, I did have the same question while I was working on it. 

So when we change the compression type, it doesn't require re-encoding of the existing data because each block or SST file includes metadata specifying the compression algorithm used for that particular block or file. This allows RocksDB to understand how to decompress the data regardless of the global compression settings at any given time. 

However, compression level, unlike compression types, which are recorded and recognized on a per-block or per-file basis, the specific level detail isn't stored with each block or file. Therefore, if we want to change it, it will require re-encoding everything, which is a huge overhead. I think that's the reason why RocksDB doesn't allow us to dynamically SetOption() for it. 

Ref: https://github.com/facebook/rocksdb/wiki/Compression (The first line explains a bit about the granularity of encoding block)

mapleFU: Rocksdb don't have this interface is fair enough, however, I don't think changing the level need re-encoding. And decompress block also doesn't need the level

mapleFU: Besides, maybe refer to https://github.com/facebook/rocksdb/pull/6615 helps.

Beihao-Zhou: I see I see, thank you very much for correcting me and finding the reference!!

PR #2173, opened 2024-03-15
Feat: add support of READONLY and READWRITE

### LINKED ISSUE #2166: Add support of the command READONLY & READWRITE
### Search before asking

### Motivation

1. redis cluster utilize [readonly](https://redis.io/commands/readonly/) to support reading from slave node
2. go-redis will send a 'readonly' command to server when 'ReadOnly' option is true during initialisation which would incur an error response from kvrocks 'ERR unknown command readonly' because kvrocks doesn't support command 'readonly'
3. this is important because without reading from slave a shard  is not readable during failover which is not acceptable for some businesses

### Solution

_No response_

LiuYuHui:
closes #2166 
local test pass
image
image

### DIFF
--- a/src/cluster/cluster.h
+++ b/src/cluster/cluster.h
@@ -68,8 +68,6 @@ class SyncMigrateContext;
 
 class Cluster {
  public:
-  enum class ClusterMode { READONLY, READWRITE };
-
   explicit Cluster(Server *srv, std::vector<std::string> binds, int port);
   Status SetClusterNodes(const std::string &nodes_str, int64_t version, bool force);
   Status GetClusterNodes(std::string *nodes_str);
--- a/src/cluster/cluster.h
+++ b/src/cluster/cluster.h
@@ -68,6 +68,8 @@ class SyncMigrateContext;
 
 class Cluster {
  public:
+  enum class ClusterMode { READONLY, READWRITE };
+
   explicit Cluster(Server *srv, std::vector<std::string> binds, int port);
   Status SetClusterNodes(const std::string &nodes_str, int64_t version, bool force);
   Status GetClusterNodes(std::string *nodes_str);
@@ -92,6 +94,11 @@ class Cluster {
 
   static bool SubCommandIsExecExclusive(const std::string &subcommand);
 
+  Status SetClusterMode(ClusterMode mode) {
+    cluster_mode_ = mode;
+    return Status::OK();
+  }
+
  private:
   std::string genNodesDescription();
   std::string genNodesInfo();
@@ -111,4 +118,6 @@ class Cluster {
 
   std::map<int, std::string> migrated_slots_;
   std::set<int> imported_slots_;
+
+  ClusterMode cluster_mode_{ClusterMode::READWRITE};
 };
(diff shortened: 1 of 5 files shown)


src/cluster/cluster.h
```diff
@@ -111,4 +118,6 @@ class Cluster {
 
   std::map<int, std::string> migrated_slots_;
   std::set<int> imported_slots_;
+
+  ClusterMode cluster_mode_{ClusterMode::READONLY};
```
caipengbo: These two commands should be in the **connection** dimension.

src/commands/cmd_cluster.cc
```diff
@@ -292,8 +292,25 @@ static uint64_t GenerateClusterFlag(const std::vector<std::string> &args) {
   return 0;
 }
 
+class CommandReadOnly : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    *output = redis::SimpleString("READONLY");
```
caipengbo: The reply should be `OK`.

src/commands/cmd_cluster.cc
```diff
@@ -292,8 +292,25 @@ static uint64_t GenerateClusterFlag(const std::vector<std::string> &args) {
...
+    return srv->cluster->SetClusterMode(Cluster::ClusterMode::READONLY);
+  }
+};
+
+class CommandReadWrite : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    *output = redis::SimpleString("READWRITE");
```
caipengbo: The reply should be `OK`.

src/cluster/cluster.h
LiuYuHui: thanks, fixed it.

src/commands/cmd_cluster.cc
LiuYuHui: thanks, fixed it.

src/server/redis_connection.h
```diff
@@ -45,6 +45,7 @@ class Connection : public EvbufCallbackBase<Connection> {
     kCloseAfterReply = 1 << 6,
     kCloseAsync = 1 << 7,
     kMultiExec = 1 << 8,
+    KReadWrite = 1 << 9,
```
jihuayu: The default state of the flag be disabled? Should it be `KReadOnly` here?

jihuayu (commented):
Thanks! Good job!

src/server/redis_connection.h
LiuYuHui: It seems that the original logic is `KReadOnly` ==1, and the default flag is 0, so I choose `KReadWrite` here, which means the default option is `KReadWrite` ==0 (`KReadOnly` ==1).

jihuayu: What is the logic for Redis? The original logic may be wrong.
We need to align with Redis, not with the original.

LiuYuHui: https://redis.io/commands/readonly/
> Normally replica nodes will redirect clients to the authoritative master for the hash slot involved in a given command, however clients can use replicas in order to scale reads using the READONLY command.

I see, Redis default option is `READWRTIE`, will change it .

jihuayu (approved):
Thank you!

caipengbo (approved):

git-hulk (approved):

git-hulk:
@LiuYuHui Thanks for your contribution.

PR #2182, opened 2024-03-19
Bump rocksdb to v9.0.0

aleksraiden:
Bump rocksdb to v9.0.0. Full changelog - https://github.com/facebook/rocksdb/releases/tag/v9.0.0

**Important!** format_version=6 is the new default setting in BlockBasedTableOptions, for more robust data integrity checking. DBs and SST files written with this setting cannot be read by RocksDB versions before 8.6.0

**Features and changes**:
- Mark wal_compression feature as production-ready. Currently only compatible with ZSTD compression
- Removed a lot of deprecated option from early release
- format_version=6 is the new default setting in BlockBasedTableOptions, for more robust data integrity checking. DBs and SST files written with this setting cannot be read by RocksDB versions before 8.6.0
- Compactions can be scheduled in parallel in an additional scenario: multiple files are marked for compaction within a single column family
- For leveled compaction, RocksDB will try to do intra-L0 compaction if the total L0 size is small compared to Lbase
- Fix some perf context statistics error in write steps
- Fixed a bug that can, under rare circumstances, cause MultiGet to return an incorrect result for a duplicate key in a MultiGet batch
- Fix a bug where older data of an ingested key can be returned for read when universal compaction is used

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -902,11 +902,6 @@ rocksdb.write_options.sync no
 # Default: no
 rocksdb.write_options.disable_wal no
 
-# Algorithm to use for WAL compression. none to disable.
-# Supports from rocksdb 9.0. At now only none and zstd are support.
-# Default: none
-rocksdb.wal_compression = none
-
 # If enabled and we need to wait or sleep for the write request, fails
 # immediately.
 #
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -895,6 +895,11 @@ rocksdb.write_options.sync no
 # Default: no
 rocksdb.write_options.disable_wal no
 
+# Algorithm to use for WAL compression. none to disable.
+# Supports from rocksdb 9.0. At now only none and zstd are support.
+# Default: none
+rocksdb.wal_compression = none
+
 # If enabled and we need to wait or sleep for the write request, fails
 # immediately.
 #
--- a/cmake/rocksdb.cmake
+++ b/cmake/rocksdb.cmake
@@ -26,8 +26,8 @@ endif()
 include(cmake/utils.cmake)
 
 FetchContent_DeclareGitHubWithMirror(rocksdb
-  facebook/rocksdb v8.11.3
-  MD5=6b1faf94b3880913725ddb09ab60d7a6
+  facebook/rocksdb v9.0.0
+  MD5=ce5630234491f81fc8d180b327d39ef5
 )
 
 FetchContent_GetProperties(jemalloc)


PragmaTwice:
> Mark wal_compression feature as production-ready. Currently only compatible with ZSTD compression

Maybe we can expose it in the config file.

mapleFU:
> Important! format_version=6 is the new default setting in BlockBasedTableOptions, for more robust data integrity checking. DBs and SST files written with this setting cannot be read by RocksDB versions before 8.6.0

Would this be dangerous? We need check which version does 2.0 kvrocks cluster uses?

git-hulk:
> > Important! format_version=6 is the new default setting in BlockBasedTableOptions, for more robust data integrity checking. DBs and SST files written with this setting cannot be read by RocksDB versions before 8.6.0
> 
> Would this be dangerous? We need check which version does 2.0 kvrocks cluster uses?

We have pinned the format version to 5 since 2.0.x: https://github.com/apache/kvrocks/blob/unstable/src/storage/storage.cc#L128

This change won't affect us, but we should be more careful since it's a major version.

caipengbo:
I wonder if we're being a little too radical in upgrading rocksdb.

PragmaTwice:
> I wonder if we're being a little too radical in upgrading rocksdb.

Just feeling "radical" is not a valid objection.

Please attach detailed reason.

git-hulk:
> > I wonder if we're being a little too radical in upgrading rocksdb.
> 
> Just feeling "radical" is not a valid objection.
> 
> Please attach detailed reason.

The underlying storage may affect the performance and even break users' data,
so we must take care while upgrading, especially for the major version.
But the main issue is that except for the rocksdb release logs, we don't have other
metrics to tell if we're good to upgrade. I suggest holding this change for
a while and keeping an eye on the rocksdb community, and bringing this change after
the next release if all goes well in the rocksdb community.

It's good to introduce it in the unstable branch but exclude this change
in the next release to make sure we have enough time to collect feedback.

tisonkun:
Instead, what test coverage we lack that prevents us to confidently upgrade ..

PragmaTwice:
> we don't have other metrics to tell if we're good to upgrade

IMHO it's not the problem of rocksdb, instead it's a problem of our project.

The important thing here is to think how we can improve the test suite and regression testing to make us more confident for any dependency upgrade.

> and bringing this change after the next release if all goes well in the rocksdb community

We are talking about kvrocks on the unstable branch, so I don't think we should wait for a more stable version for dependencies.

Instead we can become testers for the rocksdb community as a method of contributing back : )

git-hulk:
> IMHO it's not the problem of rocksdb, instead it's a problem of our project.

Yes, we don't have enough tests to let us be confident to upgrade. My point is the number of the rocksdb users is far larger than Kvrocks, it could wait for more feedback from the rocksdb community.

> We are talking about kvrocks on the unstable branch, so I don't think we should wait for a more stable version for dependencies.
>
> Instead we can become testers for the rocksdb community as a method of contributing back : )

Yes, agree we should become tests for the rocksdb community, but we need to avoid letting users test in production. So I don't recommend bringing this change into the next release.

mapleFU:
We can be radical in upgrade the bug fix versions, but maybe we should be more carefully when upgrading the big-version changing

aleksraiden:
> We can be radical in upgrade the bug fix versions, but maybe we should be more carefully when upgrading the big-version changing

As I think, an rocksdb team will not  continuous support older version, so if we not upgrade, we will use unsupported version - no patch and bugfix in 8.х

git-hulk:
> As I think, an rocksdb team will not continuous support older version, so if we not upgrade, we will use unsupported version - no patch and bugfix in 8.х

@aleksraiden Yes, we exactly want to keep the rocksdb version as fresh as possible, the main concern is the timing to introduce this major upgrade.

mapleFU:
Sigh, actually we're not that team which too familiar with RocksDB(and we don't has strictly performance requirement on it), this has a bad part and a good part. The bad part is mostly we don't know the risk of RocksDB's new version, Meta might rush tools for zippydb, myrocks or other internal user. The good part is, we just use some basic api in RocksDB, which would hardly trigger the "new bug" in RocksDB, its tests can cover our basic usages. So personally I think we can upgrade it, but all  release notes should be review carefully, and maybe we can have a performance or other testing here

aleksraiden:
Also, maybe we can support a build flag, e.g. -DUSEROCKSDB=8.11.3|9.0.0|upstream(latest commit) ?

mapleFU:
> Also, maybe we can support a build flag, e.g. -DUSEROCKSDB=8.11.3|9.0.0|upstream(latest commit) ?

I think it could be a good choice for library, however, most users use default version or using our builded mirror. 🤔 It could be useful for only few "advanced" users

aleksraiden:
And so, what is our plan to do with this changes? As I see in rocksdb repository, yesterday they are starting an 9.1 branch. 

At my side, in our project, we are switch to our unstable version of kvrocks which are included all latest unstable commits for core components (rocksdb, zstd, libevent, zlib and so on).

PragmaTwice:
From my own side, I believe we don't block dep update **WITHOUT A REASON**, especially for unstable branch.

I'll merge it if no objection.

git-hulk:
I'm fine to merge into the unstable branch, but excluding this commit in the next release would be better.

aleksraiden:
> I'm fine to merge into the unstable branch, but excluding this commit in the next release would be better.

I think, OK. As I understand - exclude in next release by default and in next+1 release - use as default, right?

git-hulk:
> > I'm fine to merge into the unstable branch, but excluding this commit in the next release would be better.
> 
> I think, OK. As I understand - exclude in next release by default and in next+1 release - use as default, right?

Yes, exactly.

mapleFU (dismissed):
+1 from myside

Besides. Maybe we should mark some config, like clock, as experiment. I think rocksdb will keep basic usage great. But anyway new feature might containing bugs.

git-hulk (dismissed):

kvrocks.conf
```diff
@@ -902,6 +902,11 @@ rocksdb.write_options.sync no
 # Default: no
 rocksdb.write_options.disable_wal no
 
+# Algorithm to use for WAL compression. none to disable.
+# Supports from rocksdb 9.0. At now only none and zstd are support.
+# Default: none
+rocksdb.wal_compression = none
```
PragmaTwice: We need to add corresponding config item in the code.

aleksraiden: Of course, @torwig could you help me with this, please?

torwig: @aleksraiden Sure.

git-hulk: @aleksraiden We can revert kvrocks.conf and then @torwig submit the PR while he gets time.

aleksraiden: Thanks, done

git-hulk (approved):

torwig (approved):

caipengbo (approved):

PragmaTwice (approved):

Issue #2071, opened 2024-01-29
Proposal: A unified query intermediate representation for RediSearch and SQL


PragmaTwice:
It is part of #2064.

RediSearch provides [a special syntax](https://redis.io/docs/interact/search-and-query/advanced-concepts/query_syntax/) for data query, while its semantics has no difference than a subset of SQL DQL.

To be comatible with RediSearch, we should support such query syntax. However, if we can also offer a SQL-like syntax simultaneously, it will greatly benefit users and minimize the learning curve. (Especially since the Redis query syntax is weird and has huge semantic gaps between different versions: v1 and v2.)

To support both syntaxes simultaneously, we can design a unified intermediate representation (called Kvrocks Query Intermediate Representation, KQIR) and transform both syntaxes into this representation when processing user input.

Hence, the frontend of KQIR will be:
- Redis query syntax (v2) and FT.SEARCH fields (for maintainance consideration, v1 will NOT be supported),
- SQL select statement (try to be compatible with MySQL)

Detailed design will be provided soon.

jihuayu:
It seems like we don't need support JOIN because `FT.SEARCH` doesn't support JOIN.
It is a good news for KQIR design.

PR #2183, opened 2024-03-20
Initialize the data structure of KQIR

PragmaTwice:
Part of #2071.

In this first version we initialized a relatively simple intemediate representation, which doesn't support numeric calculation but supports full logical calculation.


git-hulk (approved):
LGTM,cool

mapleFU (approved):

PR #2184, opened 2024-03-20
Add SQL SELECT parser via PEGTL for search module

### LINKED ISSUE #2126: Add SQL select query parser for RediSearch
We generally follow a subset of the MySQL SELECT statement syntax:
https://dev.mysql.com/doc/refman/8.0/en/select.html

PragmaTwice:
Part of #2071.
Close #2126 .

In this PR, a simple parser and its IR transformer for the SELECT statement are added. As in the KQIR, we currently don't support numeric calculation.


PragmaTwice:
This PR is ready for review now : ) cc @git-hulk @mapleFU

git-hulk:
@PragmaTwice Seems the newer cmake version failed to install in CentOS 7: https://github.com/apache/kvrocks/actions/runs/8393130640/job/22987944193?pr=2184#step:8:967

git-hulk:
Great job, looks good to me.

cmake/pegtl.cmake
```diff
@@ -0,0 +1,27 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+include_guard()
+
+include(cmake/utils.cmake)
+
+FetchContent_DeclareGitHubWithMirror(pegtl
```
git-hulk: We need to avoid wrongly introducing further versions from its main branch or fork it if we want to fix bugs. I think we can add a comment to notice here.

PragmaTwice: Why not just use new versions?

PragmaTwice: BSL-1.0 is Boost Software License.

git-hulk: oh, sorry. I thought it's the Business Software License.

git-hulk (approved):

src/search/ir.h
```diff
@@ -20,87 +20,243 @@
...
-struct SortBy {
-  enum { ASC, DESC } order;
-  FieldRef field;
+struct SortBy : Node {
+  enum Order { ASC, DESC } order = ASC;
+  std::unique_ptr<FieldRef> field;
+
+  SortBy(Order order, std::unique_ptr<FieldRef> &&field) : order(order), field(std::move(field)) {}
+
+  static constexpr const char *OrderToString(Order order) { return order == ASC ? "asc" : "desc"; }
```
mapleFU, line 215: would string_view better here?

mapleFU (approved):

src/search/ir.h
PragmaTwice: Hmm I see. I'll change it later.

PR #2190, opened 2024-03-21
Feat: add support of XPENDING

### LINKED ISSUE #1734: Add support of the XPENDING command
### Search before asking

### Motivation

This issue wants to implement [XPENDING](https://redis.io/commands/xpending/) command like redis.

### Solution

_No response_

LiuYuHui:
closes #1734

### DIFF
--- a/src/types/redis_stream.cc
+++ b/src/types/redis_stream.cc
@@ -23,11 +23,13 @@
 #include <rocksdb/status.h>
 
 #include <memory>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
 #include "db_util.h"
 #include "time_util.h"
+#include "types/redis_stream_base.h"
 
 namespace redis {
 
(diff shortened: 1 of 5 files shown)


src/types/redis_stream.cc
```diff
@@ -343,8 +346,26 @@ rocksdb::Status Stream::DeletePelEntries(const Slice &stream_name, const std::st
...
+      s = storage_->Get(rocksdb::ReadOptions(), stream_cf_handle_, consumer_key, &get_consumer_value);
+      if (!s.ok() && !s.IsNotFound()) {
+        return s;
+      }
+      if (s.IsNotFound()) {
+        return rocksdb::Status::OK();
+      }
+
+      StreamConsumerMetadata consumer_metadata = decodeStreamConsumerMetadataValue(get_consumer_value);
+      consumer_metadata.pending_number -= 1;
```
LiuYuHui, line 366: @Yangsx-1 , It seems that we should also update `pending_number` in `StreamConsumerMetadata`, when `xack` id, currently, we only update `pending_number` in `StreamConsumerGroupMetadata`.

Yangsx-1: Good catch! You're right.

src/types/redis_stream.cc
```diff
@@ -343,8 +346,26 @@ rocksdb::Status Stream::DeletePelEntries(const Slice &stream_name, const std::st
...
+
+      batch->Put(stream_cf_handle_, consumer_key, encodeStreamConsumerMetadataValue(consumer_metadata));
       batch->Delete(stream_cf_handle_, entry_key);
```
Yangsx-1, line 369: I think this line should be put after `*acknowledged += 1` , because the entry_key may not be delete due to the early return.

src/types/redis_stream.cc
```diff
@@ -343,8 +346,26 @@ rocksdb::Status Stream::DeletePelEntries(const Slice &stream_name, const std::st
...
+
+      const std::string &consumer_name = pel_entry.consumer_name;
+      std::string consumer_key = internalKeyFromConsumerName(ns_key, metadata, group_name, consumer_name);
+      std::string get_consumer_value;
+      s = storage_->Get(rocksdb::ReadOptions(), stream_cf_handle_, consumer_key, &get_consumer_value);
+      if (!s.ok() && !s.IsNotFound()) {
+        return s;
+      }
+      if (s.IsNotFound()) {
+        return rocksdb::Status::OK();
```
Yangsx-1, line 362: I don't think it's a return here. The above line is the same. It should not affect other consumers and the final result.

src/types/redis_stream.cc
```diff
@@ -1393,4 +1414,91 @@ rocksdb::Status Stream::SetId(const Slice &stream_name, const StreamEntryID &las
   return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
 }
 
+rocksdb::Status Stream::GetPendingEntries(StreamPendingOptions &options, StreamGetPendingEntryResult &pending_infos,
+                                          std::vector<StreamGetExtPendingEntryResult> &ext_results) {
+  const std::string &stream_name = options.stream_name;
+  const std::string &group_name = options.group_name;
+  std::string ns_key = AppendNamespacePrefix(stream_name);
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
```
jihuayu, line 1423: Can we use the snapshot instead of the lock? @mapleFU 
https://github.com/apache/kvrocks/pull/2174

src/types/redis_stream.h
```diff
@@ -66,6 +66,8 @@ class Stream : public SubKeyScanner {
   rocksdb::Status GetLastGeneratedID(const Slice &stream_name, StreamEntryID *id);
   rocksdb::Status SetId(const Slice &stream_name, const StreamEntryID &last_generated_id,
                         std::optional<uint64_t> entries_added, std::optional<StreamEntryID> max_deleted_id);
+  rocksdb::Status GetPendingEntries(StreamPendingOptions &options, StreamGetPendingEntryResult &pending_infos,
+                                    std::vector<StreamGetExtPendingEntryResult> &ext_results);
```
jihuayu, line 70: `options` here can be const.


The signature of this function is different from what we are accustomed to.
1. We are accustomed to using pointers instead of references to represent return values. 
2. In general, we usually have only one return value.
3. We usually place the key name(here is stream_name), group_name, consumer_name outside instead of inside the options.

src/types/redis_stream.cc
```diff
@@ -23,11 +23,14 @@
 #include <rocksdb/status.h>
 
 #include <memory>
+#include <unordered_set>
```
jihuayu, line 26: Do we need this included?

src/types/redis_stream.cc
jihuayu: Has this patch about `add support of XPENDING`?
If not can you move the change to a new PR? @LiuYuHui

jihuayu (commented):
Thank you! Pretty good.

Yangsx-1:
@LiuYuHui Hi, how's it going now?

LiuYuHui:
> @LiuYuHui Hi, how's it going now?

@Yangsx-1 Sorry for the long delay, will pick it up this weekend.

PR #2194, opened 2024-03-23
Allow to retrieve the cluster node id via the CLUSTERX command

git-hulk:
Currently, we have no way to retrieve the cluster node id
which makes it hard to debug while we want to know if the node
id is our expected.


caipengbo (approved):

jihuayu (approved):

git-hulk:
Thanks for your review.

PR #2196, opened 2024-03-24
Fix missing migrating/importing information in the CLUSTER NODES command

git-hulk:
In Redis, it will add the migrating/importing slot section
for the source and target node. For the migrating source node,
it will contain the below section:

```
[{slot_id}->-{target_node_id}]
```

And for the importing node, it will add:

```
[{slot_id}-<-{source_node_id}]
```


git-hulk:
In this PR, I also removed the unused import_fd_ field in the importor.

torwig (approved):

PR #2205, opened 2024-03-27
Add support for dumping DOT graphs from KQIR

PragmaTwice:
In this PR, we support dump the IR to a DOT graph for debugging, especially for the case that the IR is more complicated and transformed by some IR pass that's different from the original AST.
image


PragmaTwice:
Ready for review now.

src/search/ir.h
```diff
@@ -22,33 +22,49 @@
...
 #include "string_util.h"
+#include "type_util.h"
 
 // kqir stands for Kvorcks Query Intermediate Representation
 namespace kqir {
 
 struct Node {
   virtual std::string Dump() const = 0;
+  virtual std::string_view Name() const = 0;
+  virtual std::string Content() const { return {}; }
```
git-hulk, line 46: Not sure if my understanding is correct:

`Content` should be always overridden by the leaf node and `ChildBegin` and `ChildEnd` for the non-leaf node like SELECT and HASTAG query?

PragmaTwice: No. `Content` is just to show some additional information about the node, e.g. in `NumericCompareExpr` will show the operator (==/</...).

git-hulk: oh? I found some of non-leaf nodes(e.g. TagContainExpr/OrExpr) didn't implement the `Content`, so I thought the `Content` is only meaningful to the leaf node.

PragmaTwice: Because there's no additional information in such node. Also, even for leaf node, if there is no additional information, its content can also be empty.

git-hulk: i see, thanks

git-hulk (approved):

PR #2210, opened 2024-03-30
Add support of the CLUSTER RESET command

### LINKED ISSUE #2185: Add support of the command CLUSTER RESET
Like the Redis [CLUSTER SET](https://redis.io/commands/cluster-reset/) command, we would like to reset the Kvrocks cluster via redis-cli.

To be noticed, Redis required the master node not to contain any keys before executing the CLUSTER SET command. For Kvrocks, we don't have the living counter for the key number, but we can use the rocksdb seek API to check if any keys exist.

git-hulk:
This closes #2185

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -53,7 +53,7 @@ Cluster::Cluster(Server *srv, std::vector<std::string> binds, int port)
 // cluster data, so these commands should be executed exclusively, and ReadWriteLock
 // also can guarantee accessing data is safe.
 bool Cluster::SubCommandIsExecExclusive(const std::string &subcommand) {
-  for (auto v : {"setnodes", "setnodeid", "setslot", "import"}) {
+  for (auto v : {"setnodes", "setnodeid", "setslot", "import", "reset"}) {
     if (util::EqualICase(v, subcommand)) return true;
   }
   return false;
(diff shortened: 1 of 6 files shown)


src/commands/cmd_cluster.cc
```diff
@@ -105,6 +106,14 @@ class CommandCluster : public Commander {
...
         return {Status::RedisExecErr, s.Msg()};
       }
+    } else if (subcommand_ == "reset") {
+      Status s = srv->cluster->Reset();
+      if (s.IsOK()) {
+        *output = redis::SimpleString("OK");
+      } else {
+        return {Status::RedisExecErr, s.Msg()};
+      }
+      return Status::OK();
```
caipengbo: This line can be deleted.

caipengbo (approved):

torwig (approved):

src/cluster/cluster.cc
```diff
@@ -828,3 +828,31 @@ Status Cluster::CanExecByMySelf(const redis::CommandAttributes *attributes, cons
...
+
+  version_ = -1;
+  size_ = 0;
+  myid_.clear();
+  myself_.reset();
+
+  nodes_.clear();
+  for (auto &n : slots_nodes_) {
+    n = nullptr;
+  }
```
PragmaTwice, line 851: why we need to set all elements to `nullptr` instead of `clear()`?

git-hulk: slot_nodes_ is an array and the number of slot is a fixed number, so I think it should be fine to clear them with the null pointer.

jihuayu (approved):

PR #2218, opened 2024-04-03
Add NotExpr pushing pass for KQIR

PragmaTwice:
In this pass, `not` expr will be pushed into a more inner place.

### DIFF
--- a/src/search/passes/push_down_not_expr.h
+++ b/src/search/passes/push_down_not_expr.h
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


git-hulk (approved):

src/search/passes/push_down_not_expr.h
```diff
@@ -0,0 +1,59 @@
...
+    } else if (auto v = Node::As<TagContainExpr>(std::move(node->inner))) {
+      return std::make_unique<NotExpr>(std::move(v));
+    } else if (auto v = Node::As<AndExpr>(std::move(node->inner))) {
+      std::vector<std::unique_ptr<QueryExpr>> nodes;
+      for (auto& n : v->inners) {
+        nodes.push_back(std::make_unique<NotExpr>(std::move(n)));
+      }
+      res = std::make_unique<OrExpr>(std::move(nodes));
+    } else if (auto v = Node::As<OrExpr>(std::move(node->inner))) {
+      std::vector<std::unique_ptr<QueryExpr>> nodes;
```
mapleFU, line 46: May I ask how `NULL or ` or `NULL and` is handled in this case?

PragmaTwice: No. Currently we don't support NULL. We support TRUE and FALSE and it's handled in SimplifyBoolean pass.

mapleFU: Oh that's ok

mapleFU (approved):

PR #2220, opened 2024-04-03
Disable the ANALYZE command to reconsider the design

PragmaTwice:
### Why to disable:

Currently, in kvrocks codebase there are three places for the logic of executing command:
- Connection::ExecuteCommands
- redis command execution in lua scripting
- the ANALYSE command

It's very hard to maintain in these three places and even implement the correct logic.

And there are so many wrong/missing point in the command execution part of ANALYZE:
- command attribute flags are totally be ignored (it can cause issue in some cases like replica executing write commnds)
- it can cause issue in MULTI-EXEC and also scripting since the flags are not dynamically added properly
- it cannot be traced by WATCH/UNWATCH
- it cannot handle blocking commands
- cluster related staff

### How to design:

Actually we should reconsider how to design such a command and its workflow, to trade off between maintainibility, completeness and user exprience. e.g. 
- How to maintain the command executing logic properly, to avoid so many missing?
- How to respond to user properly? Could we use attributes reply in RESP3?
- Could we have a better analyzing control rather than just forward command arguments in every ANALYZE?
- ...

Since it a relatively big decison, I want to have more discussion and options. Feel free to comment : )

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -1207,72 +1207,6 @@ class CommandRdb : public Commander {
   uint32_t db_index_ = 0;
 };
 
-class CommandAnalyze : public Commander {
- public:
-  Status Parse(const std::vector<std::string> &args) override {
-    if (args.size() <= 1) return {Status::RedisExecErr, errInvalidSyntax};
-    for (unsigned int i = 1; i < args.size(); ++i) {
-      command_args_.push_back(args[i]);
-    }
-    return Status::OK();
-  }
-  Status Execute(Server *srv, Connection *conn, std::string *output) override {
-    auto commands = redis::CommandTable::Get();
-    auto cmd_iter = commands->find(util::ToLower(command_args_[0]));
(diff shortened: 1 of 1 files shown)


git-hulk:
From my perspective, I agree what @PragmaTwice mentioned. Expect for the design, the main obstacle for Kvrocks performance tracing is how to export the underlying storage metrics to users with an easy way. Maybe we can discuss this in another thread and +1 for this PR.

git-hulk (dismissed):

git-hulk (approved):

mapleFU:
Seems that performance analysis should be for developers...

mapleFU (approved):
I'm ok with this but after this only kvrocks programmer know how to debug the performance

PragmaTwice:
> I'm ok with this but after this only kvrocks programmer know how to debug the performance

We need to reconsider how to design the ANALYZE command and propose a new spec.

And then we can ship it to users.

PR #2221, opened 2024-04-04
Fix wrongly try to rewrite the namespace in the cluster mode

### LINKED ISSUE #2214: Slave does not sync key updates during downtime after restart
### Search before asking

### Version

Kvrocks 2.7.0

### Minimal reproduce step

1. shutdown slave node
2. update key `test1` and create a new key `test2` in master node
3. bring up the slave node and check the keys in slave, `test1` in slave is not updated and `test2` is created

after restart, slave should have the updated value of `test1`

### What did you see instead?

`test1` key in slave is not updated 

### Anything Else?

I also tried updating key `test2` in master node to see if this could trigger the sync of key `test1` between slave and master, but only to find `test2` is updated in slave

git-hulk:
This closes #2214

The namespace mechanism is NOT allowed in cluster mode, so it's
unnecessary to rewrite while the cluster mode is enabled. This
config rewrite behavior will cause the replication issue which was
mentioned in #2214 when starting the cluster node.

The root cause is that the server will try to rewrite the namespace
into the rocksdb if the option `repl-namespace-enabled` is enabled.
So it will increase the server's rocksdb sequence and replication will
start with the wrong offset. We have checked if the role is a slave 
before rewriting, but the cluster replication is NOT set at that time(master-replica is good).

The good news is it only affects the cluster users who enabled
the option `repl-namespace-enabled`, so I guess almost no user
will do this since the namespace replication is meaningless to the cluster mode.

Please refer to the first commit in this PR for the reproduced code.
The replica will lose one update before applying this patch:

```
=== RUN   TestClusterReplication/Cluster_replication_should_work_normally_after_restart
    replication_test.go:88: 
        	Error Trace:	/Users/hulk/code/cxx/kvrocks/tests/gocase/integration/replication/replication_test.go:88
        	Error:      	Not equal: 
        	            	expected: "v1"
        	            	actual  : "v0"
```

And it works well after this patch.

### DIFF
--- a/src/server/namespace.cc
+++ b/src/server/namespace.cc
@@ -53,6 +53,9 @@ bool Namespace::IsAllowModify() const {
 
 Status Namespace::LoadAndRewrite() {
   auto config = storage_->GetConfig();
+  // Namespace is NOT allowed in the cluster mode, so we don't need to rewrite here.
+  if (config->cluster_enabled) return Status::OK();
+
   // Load from the configuration file first
   tokens_ = config->load_tokens;
 
--- a/src/storage/redis_pubsub.cc
+++ b/src/storage/redis_pubsub.cc
@@ -23,6 +23,9 @@
 namespace redis {
 
 rocksdb::Status PubSub::Publish(const Slice &channel, const Slice &value) {
+  if (storage_->GetConfig()->IsSlave()) {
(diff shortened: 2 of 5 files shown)


torwig (approved):
LGTM

jihuayu (approved):

PR #2217, opened 2024-04-02
Add cron to recalculate db size

kinoute:
Following https://github.com/RocksLabs/kvrocks_exporter/pull/25, I was wondering if there is any way to automatically recalculate the db size instead of doing it manually through `dbsize scan` with the redis-cli. Because without it, the panel on the Grafana dashboard doesn't make any sense without manual intervention, same for the `info` command through the CLI or redis client.

I thought it would be nice to add a cronjob that could run periodically to run this command on all namespaces to have up-to-date estimated keys number (default: disabled).

This is my first PR here and in C++, I hope I didn't make any mess.

PR #2222, opened 2024-04-04
Allow to scan keys with the suffix

baobaomaomeng:
After modifying the code, I'm encountering some strange issues. I don't understand why this test is failing.

first

second

third
 At the same time, there's a bit of a performance issue. I'm not sure about the optimal number of scans for substring matching and suffix matching, I haven't set a limit yet. In the next couple of days, I'll try to add support for HSCAN, SSCAN, and ZSCAN.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,5 @@ testdb
 
 build
 cmake-build-*
+debug-build
+test
--- a/.gitignore
+++ b/.gitignore
@@ -42,9 +42,8 @@ version.h
 .cache
 
 compactdb
+test
 testdb
 
 build
 cmake-build-*
-debug-build
-test
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -840,7 +840,8 @@ class CommandScan : public CommandScanBase {
 
     std::vector<std::string> keys;
     std::string end_key;
-    auto s = redis_db.Scan(key_name, limit_, prefix_, &keys, &end_key, type_, pm_);
+    ScanConfig scan_config(limit_, srv->GetConfig()->max_scan_num, prefix_);
+    auto s = redis_db.Scan(key_name, &keys, scan_config, *match_mode_,&end_key, type_);
     if (!s.ok()) {
       return {Status::RedisExecErr, s.ToString()};
     }
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -840,7 +840,7 @@ class CommandScan : public CommandScanBase {
 
     std::vector<std::string> keys;
     std::string end_key;
-    auto s = redis_db.Scan(key_name, limit_, prefix_, &keys, &end_key, type_);
+    auto s = redis_db.Scan(key_name, limit_, prefix_, &keys, &end_key, type_, pm_);
     if (!s.ok()) {
       return {Status::RedisExecErr, s.ToString()};
     }
--- a/src/commands/scan_base.h
+++ b/src/commands/scan_base.h
@@ -20,6 +20,8 @@
 
 #pragma once
 
+#include <vector>
+
 #include "commander.h"
 #include "commands/command_parser.h"
 #include "error_constants.h"
@@ -44,17 +46,16 @@ class CommandScanBase : public Commander {
     while (parser.Good()) {
       if (parser.EatEqICase("match")) {
         prefix_ = GET_OR_RET(parser.TakeStr());
-        if(prefix_.size()>=2&&prefix_.front() == '*'&&prefix_.back() =='*'){
-          prefix_ = prefix_.substr(1,prefix_.size()-2);
-          pm_ = 2;
-        }else if (!prefix_.empty() && prefix_.back() == '*') {
+        if (prefix_.size() >= 2 && prefix_.front() == '*' && prefix_.back() == '*') {
+          prefix_ = prefix_.substr(1, prefix_.size() - 2);
+          match_mode_ = std::make_unique<SubStringMatchType>();
+        } else if (!prefix_.empty() && prefix_.back() == '*') {
           prefix_ = prefix_.substr(0, prefix_.size() - 1);
-          pm_ = 0;
-        } else if(!prefix_.empty()&& prefix_.front()=='*'){
+          match_mode_ = std::make_unique<PerfixMatchType>();
+        } else if (!prefix_.empty() && prefix_.front() == '*') {
           prefix_ = prefix_.substr(1, prefix_.size() - 1);
-          pm_ = 1;
-        }
-        else {
+          match_mode_ = std::make_unique<SubStringMatchType>();
+        } else {
           return {Status::RedisParseErr, "currently only key prefix matching is supported"};
         }
       } else if (parser.EatEqICase("count")) {
@@ -87,16 +88,10 @@ class CommandScanBase : public Commander {
     }
   }
 
-  std::string GenerateOutput(Server *srv, const Connection *conn, const std::vector<std::string> &keys,
-                             CursorType cursor_type) const {
+  static std::string GenerateOutput(Server *srv, const Connection *conn, const std::vector<std::string> &keys,
+                                    const std::string &end_cursor) {
     std::vector<std::string> list;
-    if (keys.size() == static_cast<size_t>(limit_)) {
-      auto end_cursor = srv->GenerateCursorFromKeyName(keys.back(), cursor_type);
-      list.emplace_back(redis::BulkString(end_cursor));
-    } else {
-      list.emplace_back(redis::BulkString("0"));
-    }
-
+    list.emplace_back(end_cursor);
     list.emplace_back(ArrayOfBulkStrings(keys));
 
     return redis::Array(list);
@@ -107,7 +102,7 @@ class CommandScanBase : public Commander {
   std::string prefix_;
   int limit_ = 20;
   RedisType type_ = kRedisNone;
-  int pm_ = 0;
+  std::unique_ptr<BaseMatchType> match_mode_;
 };
 
 class CommandSubkeyScanBase : public CommandScanBase {
--- a/src/commands/scan_base.h
+++ b/src/commands/scan_base.h
@@ -39,15 +39,22 @@ class CommandScanBase : public Commander {
 
     return ParseAdditionalFlags<true>(parser);
   }
-
   template <bool IsScan, typename Parser>
   Status ParseAdditionalFlags(Parser &parser) {
     while (parser.Good()) {
       if (parser.EatEqICase("match")) {
         prefix_ = GET_OR_RET(parser.TakeStr());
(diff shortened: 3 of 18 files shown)


jihuayu:
@baobaomaomeng You can add a config to set the optimal number. We can talk about the default value later.
If this test case doesn't your need, you can hide it in your local.
You can use `./x.py check format` to format the code.

baobaomaomeng:
> @baobaomaomeng You can add a config to set the optimal number. We can talk about the default value later. If this test case doesn't your need, you can hide it in your local. You can use `./x.py check format` to format the code.

I've now added some tests,add support hsacn zscan and SSCAN, and the final step is to add configuration options. I'm not sure how to add configuration options, nor do I know what a good default number of search times would be. I feel like I need some help in this area.

jihuayu:
@baobaomaomeng That's great, thank you for your efforts. 

Regarding how to add configuration items, you can refer to [this PR](https://github.com/apache/kvrocks/pull/2217).

The default number may be any number(such as 100,000). We can have the Reviewers confirm the final value, this modification is quite simple.

baobaomaomeng:
> @baobaomaomeng That's great, thank you for your efforts.
> 
> Regarding how to add configuration items, you can refer to [this PR](https://github.com/apache/kvrocks/pull/2217).
> 
> The default number may be any number(such as 100,000). We can have the Reviewers confirm the final value, this modification is quite simple.

One point that confuses me is why the limit in the scan operator is not a configuration item, and now I need to add the maximum number to scan at once into the configuration items.

jihuayu:
@baobaomaomeng Hi, can you resolve the conflicts?

src/commands/scan_base.h
```diff
@@ -83,6 +89,7 @@ class CommandScanBase : public Commander {
   std::string cursor_;
   std::string prefix_;
   int limit_ = 20;
+  int pm_ = 0;
```
jihuayu: Can you use enum instead of it? 
Can you change and rename it to `prefix_mode_`?

src/commands/cmd_server.cc
```diff
@@ -878,7 +878,7 @@ class CommandScan : public CommandScanBase {
 
     std::vector<std::string> keys;
     std::string end_key;
-    auto s = redis_db.Scan(key_name, limit_, prefix_, &keys, &end_key);
+    auto s = redis_db.Scan(key_name, limit_, srv->GetConfig()->max_scan_num, prefix_, &keys, &end_key, pm_);
```
jihuayu: This code looks ugly. Can we have some good way to deal with config?

src/storage/redis_db.cc
```diff
@@ -359,28 +363,57 @@ rocksdb::Status Database::Scan(const std::string &cursor, uint64_t limit, const
...
+      // reaches its limit
       Metadata metadata(kRedisNone, false);
       auto s = metadata.Decode(iter->value());
       if (!s.ok()) continue;
 
       if (metadata.Expired()) continue;
       std::tie(std::ignore, user_key) = ExtractNamespaceKey<std::string>(iter->key(), storage_->IsSlotIdEncoded());
+      if (cnt_false >= scan_false_limit) {
+        break;
+      }
```
jihuayu: Why not put those codes to begin?
```C++
if (!scan_match_check()) {
    cnt_false++;
    if (cnt_false < scan_false_limit) {
        continue;
    } else {
        break;
    }
}
```

src/storage/redis_db.cc
```diff
@@ -359,28 +363,57 @@ rocksdb::Status Database::Scan(const std::string &cursor, uint64_t limit, const
...
     }
-    if (!storage_->IsSlotIdEncoded() || prefix.empty()) {
-      if (!keys->empty() && cnt >= limit) {
+    if (!storage_->IsSlotIdEncoded() || fix.empty()) {
+      if (!keys->empty() && (cnt_success >= scan_success_limit)) {
+        end_cursor->append(user_key);
+      }
+      if (keys->empty() && cnt_false >= scan_false_limit) {
         end_cursor->append(user_key);
       }
```
jihuayu: There may be a situation where some keys have been found, but `cnt_false >= scan_false_limit`.

src/storage/redis_db.cc
```diff
@@ -322,10 +327,12 @@ rocksdb::Status Database::Keys(const std::string &prefix, std::vector<std::strin
   return rocksdb::Status::OK();
 }
 
-rocksdb::Status Database::Scan(const std::string &cursor, uint64_t limit, const std::string &prefix,
-                               std::vector<std::string> *keys, std::string *end_cursor) {
+rocksdb::Status Database::Scan(const std::string &cursor, uint64_t scan_success_limit, uint64_t scan_false_limit,
+                               const std::string &fix, std::vector<std::string> *keys, std::string *end_cursor,
+                               const int pm) {
```
jihuayu: Can you change the param `fix` to a more meaningful name? Can you keep the name the same as the header file?

jihuayu (commented):
Good work! Thank you for your contribution. 
Can you please resolve the code conflicts first? I will continue to review the code after resolving the conflicts.

src/commands/cmd_server.cc
baobaomaomeng: I have no clue about this, do you have any suggestions?

.gitignore
```diff
@@ -40,9 +40,13 @@ version.h
...
+.gitignore
 
 compactdb
+test
 testdb
 
 build
 cmake-build-*
+debug-build
+test
```
git-hulk: duplicate line, it has been added in previous section.

src/config/config.cc
```diff
@@ -152,6 +153,7 @@ Config::Config() {
       {"pidfile", true, new StringField(&pidfile, kDefaultPidfile)},
       {"max-io-mb", false, new IntField(&max_io_mb, 0, 0, INT_MAX)},
       {"max-bitmap-to-string-mb", false, new IntField(&max_bitmap_to_string_mb, 16, 0, INT_MAX)},
+      {"max-scan-num", false, new IntField(&max_scan_num, 2, 0, INT_MAX)},
```
git-hulk, line 156: Seems `2` is too smaller to be as the `max_scan_num`?

src/storage/redis_db.cc
```diff
@@ -309,10 +314,12 @@ rocksdb::Status Database::Keys(const std::string &prefix, std::vector<std::strin
   return rocksdb::Status::OK();
 }
 
-rocksdb::Status Database::Scan(const std::string &cursor, uint64_t limit, const std::string &prefix,
-                               std::vector<std::string> *keys, std::string *end_cursor, RedisType type) {
+rocksdb::Status Database::Scan(const std::string &cursor, uint64_t scan_success_limit, uint64_t scan_false_limit,
```
git-hulk: `scan_matched_limt` and `scan_mismatched_limit` should be a better name in this scenario.

.gitignore
```diff
@@ -40,9 +40,13 @@ version.h
 .idea
 .vscode
 .cache
+.gitignore
```
PragmaTwice: why it's ignored?

.gitignore
PragmaTwice: duplicated `test`

tests/gocase/unit/scan/scan_test.go
```diff
@@ -187,7 +187,27 @@ func ScanTest(t *testing.T, rdb *redis.Client, ctx context.Context) {
 		require.Equal(t, []string{"1", "10", "foo", "foobar"}, keys)
 	})
 
-	t.Run("ZSCAN with PATTERN", func(t *testing.T) {
+	t.Run("HSCAN with subkey PATTERN", func(t *testing.T) {
```
PragmaTwice, line 190: seems we need more test cases

src/commands/scan_base.h
```diff
@@ -39,14 +41,20 @@ class CommandScanBase : public Commander {
 
     return ParseAdditionalFlags<true>(parser);
   }
-
```
PragmaTwice, line 42: please don't introduce random changes.

src/storage/redis_db.cc
git-hulk: The parameter of the SCAN command is too long now, it's time to group the scan parameters into a struct.

src/storage/redis_db.cc
```diff
@@ -309,10 +314,12 @@ rocksdb::Status Database::Keys(const std::string &prefix, std::vector<std::strin
...
+                               const std::string &match_str, std::vector<std::string> *keys, std::string *end_cursor,
+                               RedisType type, const MatchType match_mode) {
```
PragmaTwice: I have some questions about your design.

But before that, could you give a little introduction about your general design and supported patterns in the PR description?

Then we can have a detailed review on your code.

Refer to Redis glob-style patterns:
>Supported glob-style patterns:
>
>`h?llo` matches hello, hallo and hxllo
`h*llo` matches hllo and heeeello
`h[ae]llo` matches hello and hallo, but not hillo
`h[^e]llo` matches hallo, hbllo, ... but not hello
`h[a-b]llo` matches hallo and hbllo
>

PragmaTwice:
In my opinion, we need a well-designed solution for implementing Redis glob-style patterns that is **scalable**, rather than just handling specific cases. If the design is not scalable, we may have to remove all modifications when aiming to support the complete pattern cases.

.gitignore
baobaomaomeng: I'm sorry, that was my oversight.

src/config/config.cc
baobaomaomeng: it's just for test easy，jihuayu say they will determine the final value.

src/storage/redis_db.cc
baobaomaomeng: I just added two matching patterns.
such as
*llo matches ello,hello,eeeello
*l* matches l,ll,llo,ello,hello

And，i add scan_false_time limit to config,beacasue if we use suffix,substring match, we need scan one by one, and the potentially long time may be unacceptable. so I add this argument in scan,zsan,sscan,hscan command。I found that the three scan methods to get the last scanned key are all different, which is not convenient for me to modify，so i add a string value  in  CommandSubkeyScanBase to represented the last key that was scanned。

baobaomaomeng:
> In my opinion, we need a well-designed solution for implementing Redis glob-style patterns that is **scalable**, rather than just handling specific cases. If the design is not scalable, we may have to remove all modifications when aiming to support the complete pattern cases.

Thank you for your review, but this is not consistent with the goal I communicated at the beginning. I now need to rewrite my code, and I think I need to consider how to write it.Could you give me some reference suggestions？

src/storage/redis_db.cc
baobaomaomeng: I've been thinking about how to design a scalable scan, but I don't have any ideas. This might be a bit difficult for me. I think I need some guidance on the thought process right now.

PragmaTwice:
closed by #2608.

PR #2225, opened 2024-04-05
Add support of the command MOVEX

### LINKED ISSUE #1950: Add support of the command MOVE
Refer to https://redis.io/commands/move/

Chiro11:
Close #1950 
Some desc:
- `MOVE key db` is treated as dummy command, to be compatible with Redis.
- `MOVEX key ns_token` moves a key to new namespace.
  1. When namespace does not correspond to token, report an error.
  2. Do not support when `requirepass` is empty, report an error.
  3. When the source key does not exist or the target key already exists, return 0, otherwise, return 1.

### DIFF
--- a/src/commands/cmd_key.cc
+++ b/src/commands/cmd_key.cc
@@ -49,59 +49,54 @@ class CommandType : public Commander {
 class CommandMove : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    if (args.size() == 3) {
-      // When the args' size is 3, we treat it as a dummy command to keep compatibility with Redis
-      move_type_ = MoveType::kDummy;
-    } else if (args.size() == 4) {
-      move_type_ = MoveType::kNormal;
-    } else {
-      return {Status::RedisParseErr, errWrongNumOfArguments};
-    }
+    GET_OR_RET(ParseInt<int64_t>(args[2], 10));
     return Status::OK();
   }
 
   Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    int count = 0;
     redis::Database redis(srv->storage, conn->GetNamespace());
-    if (move_type_ == MoveType::kDummy) {
-      int count = 0;
-      rocksdb::Status s = redis.Exists({args_[1]}, &count);
-      if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
-      *output = count ? redis::Integer(1) : redis::Integer(0);
+    rocksdb::Status s = redis.Exists({args_[1]}, &count);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = count ? redis::Integer(1) : redis::Integer(0);
+    return Status::OK();
+  }
+};
+
+class CommandMoveX : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    std::string &key = args_[1], &ns = args_[2], &token = args_[3];
+
+    redis::Database redis(srv->storage, conn->GetNamespace());
+    // auth
+    const auto &requirepass = srv->GetConfig()->requirepass;
+    if (requirepass.empty()) {
+      return {Status::NotOK, "Forbidden to move key when requirepass is empty"};
+    }
+    auto get_ns = srv->GetNamespace()->GetByToken(token);
+    if (get_ns.IsOK()) {
+      if (get_ns.GetValue() != ns) {
+        return {Status::NotOK, "Incorrect namespace or token"};
+      }
     } else {
-      std::string &key = args_[1], &ns = args_[2], &token = args_[3];
-
-      // auth
-      const auto &requirepass = srv->GetConfig()->requirepass;
-      if (requirepass.empty()) {
-        return {Status::NotOK, "Forbidden to move key when requirepass is empty"};
+      if (!(ns == kDefaultNamespace && token == requirepass)) {
+        return {Status::NotOK, "Incorrect namespace or token"};
       }
-      auto get_ns = srv->GetNamespace()->GetByToken(token);
-      if (get_ns.IsOK()) {
-        if (get_ns.GetValue() != ns) {
-          return {Status::NotOK, "Incorrect namespace or token"};
-        }
-      } else {
-        if (!(ns == kDefaultNamespace && token == requirepass)) {
-          return {Status::NotOK, "Incorrect namespace or token"};
-        }
-      }
+    }
 
-      bool ret = true;
-      rocksdb::Status s = redis.Move(key, ns, &ret);
-      if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+    bool ret = true;
+    rocksdb::Status s = redis.Move(key, ns, &ret);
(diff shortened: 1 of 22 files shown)


jihuayu:
Hi @Chiro11, thanks for your PR.
In the issues, we think we need to add a new command. Can you move the code to a new command?
https://github.com/apache/kvrocks/issues/1950#issuecomment-1963244776

Chiro11:
> Hi @Chiro11, thanks for your PR. In the issues, we think we need to add a new command. Can you move the code to a new command? [#1950 (comment)](https://github.com/apache/kvrocks/issues/1950#issuecomment-1963244776)

You mean to use a new name? Do you have any idea?

jihuayu:
> > Hi @Chiro11, thanks for your PR. In the issues, we think we need to add a new command. Can you move the code to a new command? [#1950 (comment)](https://github.com/apache/kvrocks/issues/1950#issuecomment-1963244776)
> 
> You mean to use a new name? Do you have any idea?

Yes. @git-hulk @PragmaTwice @mapleFU @enjoy-binbin Do you have some idea about new command name?

git-hulk:
> > > Hi @Chiro11, thanks for your PR. In the issues, we think we need to add a new command. Can you move the code to a new command? [#1950 (comment)](https://github.com/apache/kvrocks/issues/1950#issuecomment-1963244776)
> > 
> > 
> > You mean to use a new name? Do you have any idea?
> 
> Yes. @git-hulk @PragmaTwice @mapleFU @enjoy-binbin Do you have some idea about new command name?

Perhaps MOVEX is a good name for aligning with the CLUSTERX command.

Chiro11:
> > > > Hi @Chiro11, thanks for your PR. In the issues, we think we need to add a new command. Can you move the code to a new command? [#1950 (comment)](https://github.com/apache/kvrocks/issues/1950#issuecomment-1963244776)
> > > 
> > > 
> > > You mean to use a new name? Do you have any idea?
> > 
> > 
> > Yes. @git-hulk @PragmaTwice @mapleFU @enjoy-binbin Do you have some idea about new command name?
> 
> Perhaps MOVEX is a good name for aligning with the CLUSTERX command.

@jihuayu @git-hulk Modified as suggested.

src/storage/redis_db.cc
```diff
@@ -721,6 +721,51 @@ rocksdb::Status Database::Rename(const std::string &key, const std::string &new_
 
   if (key == new_key) return rocksdb::Status::OK();
 
+  return moveInternal(ns_key, new_ns_key, type);
+}
+
+rocksdb::Status Database::Move(const std::string &key, const std::string &new_ns, bool *ret) {
```
git-hulk: Seems the only difference between the RENAME and MOVE is how to construct its new key. The rename command requires the same namespace and a different key, and the move command requires a different namespace but the same key. So I think we need to move the construction of the new key out of Database::Rename, and then both of rename/move can share the same implementation.

src/commands/cmd_key.cc
```diff
@@ -64,6 +64,41 @@ class CommandMove : public Commander {
   }
 };
 
+class CommandMoveX : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    std::string &key = args_[1], &ns = args_[2], &token = args_[3];
+
+    redis::Database redis(srv->storage, conn->GetNamespace());
```
git-hulk, line 72: Is it better to move the old AuthenticateUser as a member function and set the connection namespace outside? so that we can reuse the authenticate code block:

```
AuthResult Server::AuthenticateUser(const std::string &user_password, std::string *ns);
```

git-hulk:
@Chiro11 Thanks for your efforts and quick response.

jihuayu (commented):
Thank you for your code. @Chiro11 
This PR contains a lot of refactoring unrelated to the `movex` command. Could you please move it to a new PR?

src/storage/redis_db.h
```diff
@@ -82,10 +82,10 @@ class Database {
   [[nodiscard]] rocksdb::Status Expire(const Slice &user_key, uint64_t timestamp);
   [[nodiscard]] rocksdb::Status Del(const Slice &user_key);
   [[nodiscard]] rocksdb::Status MDel(const std::vector<Slice> &keys, uint64_t *deleted_cnt);
-  [[nodiscard]] rocksdb::Status Exists(const std::vector<Slice> &keys, int *ret);
+  [[nodiscard]] rocksdb::Status Exists(const std::vector<Slice> &keys, int *ret, bool internal_key = false);
```
git-hulk: How about keeping the old method signature and adding a new private method like:

```
 rocksdb::Status Exists(const std::vector<Slice> &keys, int *ret) {
   ns_keys.reserve(keys.size());
   for (const auto &user_key : keys) {
     ns_keys.emplace_back(AppendNamespacePrefix(key));
   }
   return exits(ns_keys, ret);
 }
 
 // private method
 rocksdb::Status exists(const std::vector<Slice> &ns_keys, int *ret) {
 }
```
So it should be more intuitive since we don't need to check if it's an internal inside `exitst` method. What do you think?

git-hulk: Also change the Type method if sounds good.

git-hulk:
Rest are good to me, one comment inline.

PragmaTwice:
`MOVEX key ns_name ns_token`

I think here the `ns_name` is useless.

Chiro11:
> `MOVEX key ns_name ns_token`
> 
> I think here the `ns_name` is useless.

Theoretically it is. But I think this is more clear. @git-hulk @jihuayu What are the opinions of you two?

src/storage/redis_db.h
Chiro11: Done

Chiro11:
> Thank you for your code. @Chiro11 This PR contains a lot of refactoring unrelated to the `movex` command. Could you please move it to a new PR?

@jihuayu I think it's basically ok. Just some typos and best practices.

git-hulk:
> > `MOVEX key ns_name ns_token`
> > I think here the `ns_name` is useless.
> 
> Theoretically it is. But I think this is more clear. @git-hulk @jihuayu What are the opinions of you two?

The point of @PragmaTwice is we can find the `ns` by the token, so it's unnecessary to pass the namespace in the command arguments.

Chiro11:
> > > `MOVEX key ns_name ns_token`
> > > I think here the `ns_name` is useless.
> > 
> > 
> > Theoretically it is. But I think this is more clear. @git-hulk @jihuayu What are the opinions of you two?
> 
> The point of @PragmaTwice is we can find the `ns` by the token, so it's unnecessary to pass the namespace in the command arguments.

I know. But I think `MOVEX key token` is a bit strange. I think it's reasonable to have a clear namespace. But I'm actually okay with both. Depends on you.

git-hulk:
@Chiro11 So let's remove it if you feel good about this.

git-hulk (approved):

PragmaTwice (approved):

jihuayu:
Thanks @Chiro11. Good job!  Can you add this command to our website?
https://kvrocks.apache.org/docs/supported-commands#key-commands

If you're willing, can you work for https://github.com/apache/kvrocks/issues/2229?

Issue #1001, opened 2022-10-16
Make kvrocks dump rdb file which redis can load.


ColinChamber:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

  Making Kvrocks write RDB is more straightforward than making it parse RDB.
  Because Redis is backward compatible, which can load the original RDB file. 
  We can easily utilize the primitive RDB object types, which follow the simple 
  encoding pattern:  <type> <len> <key> <len> <value> 
  or <type> <len> <key> <size> <data> (<data> consists of <len><key(value)>)
  RDB_TYPE_STRING, RDB_TYPE_LIST, RDB_TYPE_SET, RDB_TYPE_ZSET, RDB_TYPE_HASH,
  these five types follow this pattern. We ignore the RDB_TYPE_HASH_ZIPMAP, 
  RDB_TYPE_LIST_ZIPLIST, RDB_TYPE_SET_INTSET, and other complicated types 
  only implement the five simple type.
 
  There is a mapping relationship between Redis object type and RDB object type.
  But you don't have to worry about Redis using ineffective object types after loading RDB. 
  Redis will convert it to an efficient type automatically.
  E.g
  	kvrocks> hmset foo foo1 bar1 foo2 bar2
	kvrocks> save (use RDB_TYPE_HASH rather than RDB_TYPE_HASH_ZIPLIST)
   --------  Redis loads the RDB that kvrocks dumped  --------
   redis> OBJECT ENCODING foo
   "ziplist"
   
  There are more detailed references to RDB.
   https://github.com/sripathikrishnan/redis-rdb-tools/wiki/Redis-RDB-Dump-File-Format
   https://github.com/sripathikrishnan/redis-rdb-tools/blob/master/docs/RDB_Version_History.textile
   https://rdb.fnordig.de/file_format.html

### Solution

#958 

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

git-hulk:
cool, thanks for @ColinChamber great work.

ShooterIT:
Hi @ColinChamber @git-hulk i don't think supporting dumping RDB in Kvrocks is a good idea.
[...]

PR #2227, opened 2024-04-06
Add support of  the command DUMP

### LINKED ISSUE #1946: Add support of the command DUMP
Refer to https://redis.io/commands/dump/

AntiTopQuark:
Close #1946 

Command Desc：
1. Adds support for the Dump command, which can be used as dump key.
2. Currently, the supported data types include: String, Hash, List, Set, ZSet. Redis also supports Stream and Module types; however, due to significant differences between kvrocks and Redis, Stream and Module types are not supported at this time.
3. Since kvrocks does not support encoding methods such as listpack and lzf, most types calculate the dump payload string using the RAW method.

Redis Documentation: [https://redis.io/commands/dump/](https://redis.io/commands/dump/)

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -1273,6 +1273,48 @@ class CommandApplyBatch : public Commander {
   bool low_pri_ = false;
 };
 
+class CommandDump : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() != 2) {
+      return {Status::RedisExecErr, errWrongNumOfArguments};
+    }
+    return Status::OK();
+  }
+
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    rocksdb::Status db_status;
+    std::string &key = args_[1];
+    redis::Database redis(srv->storage, conn->GetNamespace());
+    int count = 0;
+    db_status = redis.Exists({key}, &count);
+    if (!db_status.ok()) {
+      if (db_status.IsNotFound()) {
+        *output = conn->NilString();
+        return Status::OK();
+      }
+      return {Status::RedisExecErr, db_status.ToString()};
+    }
+    if (count == 0) {
+      *output = conn->NilString();
+      return Status::OK();
+    }
+
+    RedisType type = kRedisNone;
+    db_status = redis.Type(key, &type);
+    if (!db_status.ok()) return {Status::RedisExecErr, db_status.ToString()};
+
+    std::string result;
+    auto stream_ptr = std::make_unique<RdbStringStream>(result);
+    RDB rdb(srv->storage, conn->GetNamespace(), std::move(stream_ptr));
+    auto s = rdb.Dump(key, type);
+    if (!s.IsOK()) return {Status::RedisExecErr, s.Msg()};
+    CHECK(dynamic_cast<RdbStringStream *>(rdb.GetStream().get()) != nullptr);
+    *output = redis::BulkString(static_cast<RdbStringStream *>(rdb.GetStream().get())->GetInput());
+    return Status::OK();
+  }
+};
+
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandAuth>("auth", 2, "read-only ok-loading", 0, 0, 0),
                         MakeCmdAttr<CommandPing>("ping", -1, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandSelect>("select", 2, "read-only", 0, 0, 0),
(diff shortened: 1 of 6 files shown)


torwig:
@AntiTopQuark Running the `./x.py format` should fix the linting issues.

git-hulk:
> Since kvrocks does not support encoding methods such as listpack and lzf, most types calculate the dump payload string using the RAW method.

I feel good about this, and we can make it work first and then improve.

src/commands/cmd_server.cc
```diff
@@ -34,6 +34,8 @@
 #include "storage/rdb.h"
 #include "string_util.h"
 #include "time_util.h"
+#include "types/redis_string.h"
+#include "vendor/crc64.h"
```
PragmaTwice: seems not necessary here?

AntiTopQuark: done

src/storage/rdb.cc
```diff
@@ -680,3 +679,323 @@ Status RDB::LoadRdb(uint32_t db_index, bool overwrite_exist_key) {
...
+
+  // When the length is less than 11, value may be an integer,
+  // so special encoding is performed.
+  if (len <= 11) {
+    unsigned char buf[5];
+    // convert string to long long
+    std::istringstream iss(value);
+    long long integer_value = 0;
+    if (iss.eof() && !iss.fail()) {
+      iss >> integer_value;
```
PragmaTwice: Could you please use `util::ParseInt` instead?

src/storage/rdb.cc
```diff
@@ -680,3 +679,323 @@ Status RDB::LoadRdb(uint32_t db_index, bool overwrite_exist_key) {
...
+  /* RDB version */
+  buf[0] = MaxRDBVersion & 0xff;
+  buf[1] = (MaxRDBVersion >> 8) & 0xff;
+  s = stream_->Write((const char *)buf, 2);
+  if (!s.IsOK()) {
+    return {Status::RedisExecErr, s.Msg()};
+  }
+
+  /* CRC64 */
+  std::string &output = static_cast<RdbStringStream *>(stream_.get())->GetInput();
```
PragmaTwice, line 709: If you want to assume it must be an `RdbStringStream` here, you can add a `CHECK` statement.

PragmaTwice: But TBH, to support output files rather than strings, we need a better solution here.

src/storage/rdb.cc
```diff
@@ -680,3 +679,323 @@ Status RDB::LoadRdb(uint32_t db_index, bool overwrite_exist_key) {
...
+    return 5;
+  } else {
+    return 0;
+  }
+}
+
+Status RDB::rdbSaveBinaryDoubleValue(double val) {
+  memrev64ifbe(&val);
+  return stream_->Write((const char *)(&val), sizeof(val));
+}
```
git-hulk:
```suggestion
}

```

src/storage/rdb.cc
git-hulk: For the Redis dump command, it needs to echo the result to the client, so it should be good to keep in memory. Not sure if I misunderstood @PragmaTwice point.

AntiTopQuark: > If you want to assume it must be an `RdbStringStream` here, you can add a `CHECK` statement.

I added a member attribute named type_ to RdbStream, and I ensure its safety by checking it before performing a static_cast

src/storage/rdb.cc
```diff
@@ -680,3 +679,323 @@ Status RDB::LoadRdb(uint32_t db_index, bool overwrite_exist_key) {
...
+  /* RDB version */
+  buf[0] = MaxRDBVersion & 0xff;
+  buf[1] = (MaxRDBVersion >> 8) & 0xff;
+  s = stream_->Write((const char *)buf, 2);
+  if (!s.IsOK()) {
+    return {Status::RedisExecErr, s.Msg()};
+  }
+
+  /* CRC64 */
+  CHECK(stream_.get()->GetType() == String);
```
PragmaTwice: Maybe just `dynamic_cast<RdbStringStream *>(stream_.get()) != nullptr` ?

src/storage/rdb.cc
AntiTopQuark: > But TBH, to support output files rather than strings, we need a better solution here.

Is it absolutely necessary to dump a value into a file? 
If the goal is simply to dump the contents of specific keys, this can be readily achieved by redirecting the output of redis-cli to a designated file. 
For dumping an entire RDB, might this feature be unnecessary? Directly utilizing kvrocks2redis would suffice. This requirement has been previously discussed: [https://github.com/apache/kvrocks/issues/1001#issuecomment-1304982649](https://github.com/apache/kvrocks/issues/1001#issuecomment-1304982649)

src/storage/rdb.cc
AntiTopQuark: `dynamic_cast` takes advantage of run-time type identification (RTTI), which may have an impact on performance. Although this scene is relatively simple, it may not have much impact. I'll modify it according to the comments.

src/storage/rdb.cc
AntiTopQuark: done

src/storage/rdb.cc
AntiTopQuark: done

src/storage/rdb.cc
AntiTopQuark: done

jihuayu (approved):
The beautiful code that ignites a delightful start to my day.
Thank you!

git-hulk (approved):

Issue #1952, opened 2023-12-18
Add support of the command SORT and SORT_RO


PragmaTwice:
Refer to https://redis.io/commands/sort/ and https://redis.io/commands/sort_ro/

HolyLow:
Hello, I am interested in this issue.

PragmaTwice:
Great! Assigned.

PokIsemaine:
Hello @HolyLow @PragmaTwice 
Is there any progress on this issue? If no one is working on it at the moment, I'm willing to try this question.

git-hulk:
@PokIsemaine Thanks for your interest. I think it's fine to reassign since it's inactive for a long time.

HolyLow:
@PokIsemaine My progress is limited (just annotations), so feel free to take this over.
@git-hulk Sorry for the huge delay, I could spare some time on another issue https://github.com/apache/kvrocks/issues/1372 recently, so please reserve that one for me if it's ok.

git-hulk:
@HolyLow Thanks for your feedback.

> @git-hulk Sorry for the huge delay, I could spare some time on another issue https://github.com/apache/kvrocks/issues/1372 recently, so please reserve that one for me if it's ok.

Absolutely yes. No sorry and very appreciative of your efforts and contributions.

PR #2230, opened 2024-04-08
feat: support for the sort command[draft]

PokIsemaine:
* issues: https://github.com/apache/kvrocks/issues/1952

I'm trying to support the `sort` command and later the `sort_ro` command for kvrocks, this is my first time submitting a pull request for this project and I need some advice, so I set it to draft temporarily.

The help I need is as follows：
* There are three TODOs in the code and I don’t know how to deal with them. It seems that they are related to `ACL`, `Cluster`, `CLIENT_SCRIPT`, but I have not found relevant support for kvrocks at present. What should I do? The following is the relevant code in the redis implementation:
    * TODO1: https://github.com/redis/redis/blob/f4481e657f905074fa515701af3f695757817d88/src/sort.c#L237
    * TODO2: https://github.com/redis/redis/blob/f4481e657f905074fa515701af3f695757817d88/src/sort.c#L218
    * TODO3: https://github.com/redis/redis/blob/f4481e657f905074fa515701af3f695757817d88/src/sort.c#L296

* How to support `denyoom` and `movablekeys` flags, or whether kvrocks needs to support it?
```
127.0.0.1:6379> command info sort
1)  1) "sort"
    2) (integer) -2
    3) 1) write
       2) denyoom
       3) movablekeys
```
* Whether there is any need to modify or add to the existing code (especially exceptions), comments, and tests (especially boundary cases) in the draft?

If you can provide me with any advice I would be very grateful！
    

The information and code I referenced are as follows:
* [1] https://redis.io/commands/sort/
* [2] https://redis.io/commands/sort_ro/
* [3] https://github.com/redis/redis/blob/unstable/src/sort.c imitate its implementation
* [4] Execute `command docs sort` and `command info sort` to obtain the parameters, flags and other information of the sort command

### DIFF
--- a/src/search/ir.h
+++ b/src/search/ir.h
@@ -85,6 +85,7 @@ struct FieldRef : Ref {
   const FieldInfo *info = nullptr;
 
   explicit FieldRef(std::string name) : name(std::move(name)) {}
+  FieldRef(std::string name, const FieldInfo *info) : name(std::move(name)), info(info) {}
 
   std::string_view Name() const override { return "FieldRef"; }
   std::string Dump() const override { return name; }
--- a/src/search/ir_pass.h
+++ b/src/search/ir_pass.h
@@ -28,6 +28,8 @@ namespace kqir {
 struct Pass {
   virtual std::unique_ptr<Node> Transform(std::unique_ptr<Node> node) = 0;
 
+  virtual void Reset() {}
+
   virtual ~Pass() = default;
 };
 
--- a/src/search/passes/interval_analysis.h
+++ b/src/search/passes/interval_analysis.h
@@ -21,6 +21,7 @@
 #pragma once
 
 #include <algorithm>
+#include <cmath>
 #include <memory>
 #include <set>
 #include <type_traits>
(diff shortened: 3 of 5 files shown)


mapleFU:
> It seems that they are related to ACL, Cluster, CLIENT_SCRIPT, but I have not found relevant support for kvrocks at present. What should I do? 

I think we can leave them unimplemented first

> How to support denyoom and movablekeys flags

About oom currently it's a bit hard to manage the memory by custom allocator. Maybe we can separate a machine/process level global memory tracking module and check that. This can separate to an other patch or discuss it firstly

PokIsemaine:
> > It seems that they are related to ACL, Cluster, CLIENT_SCRIPT, but I have not found relevant support for kvrocks at present. What should I do?
> 
> I think we can leave them unimplemented first
> 
> > How to support denyoom and movablekeys flags
> 
> About oom currently it's a bit hard to manage the memory by custom allocator. Maybe we can separate a machine/process level global memory tracking module and check that. This can separate to an other patch or discuss it firstly

Thanks!
So what do we need to do to modify these TODO comments? Delete them or keep them (perhaps with more formal comments)

jihuayu (commented):
@PokIsemaine Thanks for your PR. Your code is amazingly clean. The test cases are also very detailed.
For these TODO comments, I think we can keep them.

The logic LGTM.  
But according to our conventions, this command should be placed in the `cmd_key.cc` file, and the core logic of the Execute function should be extracted and placed in the `redis::Database` class.

Below is a example:
https://github.com/apache/kvrocks/blob/26cbdb978b53c08216c6dfb5d45ed950ccde5936/src/commands/cmd_key.cc#L138-L149

src/commands/commander.h
```diff
@@ -185,6 +187,10 @@ inline uint64_t ParseCommandFlags(const std::string &description, const std::str
       flags |= kCmdCluster;
     else if (flag == "no-dbsize-check")
       flags |= kCmdNoDBSizeCheck;
+    else if (flag == "deny-oom")
+      flags |= kCmdDenyOom;
+    else if (flag == "movable-keys")
+      flags |= kCmdMovableKeys;
```
PragmaTwice: This flag is meaningless in Kvrocks. Please remove it and refer to KeyRangeVecGen.

src/commands/commander.h
```diff
@@ -185,6 +187,10 @@ inline uint64_t ParseCommandFlags(const std::string &description, const std::str
       flags |= kCmdCluster;
     else if (flag == "no-dbsize-check")
       flags |= kCmdNoDBSizeCheck;
+    else if (flag == "deny-oom")
+      flags |= kCmdDenyOom;
```
PragmaTwice: I think we don't plan to implement it in this PR, so please remove it.

src/commands/commander.h
```diff
@@ -65,6 +65,8 @@ enum CommandFlags : uint64_t {
   kCmdROScript = 1ULL << 10,       // "ro-script" flag for read-only script commands
   kCmdCluster = 1ULL << 11,        // "cluster" flag
   kCmdNoDBSizeCheck = 1ULL << 12,  // "no-dbsize-check" flag
+  kCmdDenyOom = 1ULL << 13,        // "deny-oom" flag
+  kCmdMovableKeys = 1ULL << 14,    // "movable-keys" flag
```
PragmaTwice: ditto

src/commands/cmd_sort.cc
```diff
@@ -0,0 +1,355 @@
...
+
+  std::string sortby_;                    // BY
+  bool dontsort_ = false;                 // DONT SORT
+  long offset_ = 0;                       // LIMIT OFFSET
+  long count_ = -1;                       // LIMIT COUNT
+  std::vector<std::string> getpatterns_;  // GET
+  bool desc_ = false;                     // ASC/DESC
+  bool alpha_ = false;                    // ALPHA
+  std::string storekey_;                  // STORE
+};
```
PragmaTwice: The implementation of SORT should be in `redis_db.cc`. The parsing part can be in `cmd_key.cc`.

PokIsemaine:
> @PokIsemaine Thanks for your PR. Your code is amazingly clean. The test cases are also very detailed. For these TODO comments, I think we can keep them.
> 
> The logic LGTM. But according to our conventions, this command should be placed in the `cmd_key.cc` file, and the core logic of the Execute function should be extracted and placed in the `redis::Database` class.
> 
> Below is a example:
> 
> https://github.com/apache/kvrocks/blob/26cbdb978b53c08216c6dfb5d45ed950ccde5936/src/commands/cmd_key.cc#L138-L149

Thanks for the advice, I'll probably do it this weekend

src/commands/commander.h
PokIsemaine: Okay, I need to learn about KeyRangeVecGen.

src/commands/commander.h
PokIsemaine: I agree with this, thanks for the heads up, I'll remove the flags later on.

PR #2232, opened 2024-04-08
Initialize the query plan part of KQIR

PragmaTwice:
In this PR, we added new nodes to KQIR in order to construct the query plan.

### DIFF
--- a/src/search/ir_plan.h
+++ b/src/search/ir_plan.h
@@ -1,0 +1,117 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#pragma once
+
+#include <limits>
+#include <memory>
+
+#include "ir.h"
+#include "search/ir_sema_checker.h"
+#include "string_util.h"
+
+namespace kqir {
+
+namespace plan {
+
+struct PlanOperator : Node {};
+
+struct FullIndexScan : PlanOperator {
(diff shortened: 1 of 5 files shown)


src/search/ir_plan.h
```diff
@@ -0,0 +1,117 @@
...
+  std::string Content() const override { return fmt::format("{}, {}", field->name, range.ToString()); };
+  std::string Dump() const override { return fmt::format("numeric-scan {}", Content()); }
+};
+
+struct TagFieldScan : FieldScan {
+  std::string tag;
+
+  std::string_view Name() const override { return "TagFieldScan"; };
+  std::string Content() const override { return fmt::format("{}, {}", field->name, tag); };
+  std::string Dump() const override { return fmt::format("tag-scan {}", Content()); }
```
mapleFU, line 65: is nested struct better than just white space?

PragmaTwice: I cannot get you. Please be more specific : )

PragmaTwice: Ahh I think I've got you now. Yeah I prefer S-expr and omit the parentheses as much as possible, rather than `f(..)`.
And `Dump` is only used for equality checking in unit test. For debugging, we can use the dot graph dumper.

git-hulk (approved):

src/search/ir_plan.h
```diff
@@ -0,0 +1,113 @@
...
+  std::unique_ptr<QueryExpr> filter_expr;
+
+  std::string_view Name() const override { return "Filter"; };
+  std::string Dump() const override { return fmt::format("(filter {}, {})", source->Dump(), Content()); }
+
+  NodeIterator ChildBegin() override { return {source.get(), filter_expr.get()}; }
+  NodeIterator ChildEnd() override { return {}; }
+};
+
+struct Merge : PlanOperator {
```
mapleFU, line 79: You can add description on Merge, since it's far from "SQL MERGE", which means doing complex if-else on selected data

src/search/ir_plan.h
```diff
@@ -0,0 +1,113 @@
...
+  std::string_view Name() const override { return "Limit"; };
+  std::string Dump() const override { return fmt::format("(limit {}, {}: {})", offset, count, op->Dump()); }
+
+  NodeIterator ChildBegin() override { return NodeIterator{op.get()}; }
+  NodeIterator ChildEnd() override { return {}; }
+};
+
+struct Projection : PlanOperator {
+  std::unique_ptr<PlanOperator> source;
+  std::unique_ptr<SelectExpr> select;
```
mapleFU, line 104: I just noticed that SelectExpr is array of field, what would `select 1;` like?

PragmaTwice: It currently just doesn't support such form.

src/search/ir_plan.h
PragmaTwice: I see. Good suggestion.

PR #2234, opened 2024-04-09
Add clone method to KQIR Node

PragmaTwice:
In this PR we added a virtial function `Node::Clone` and impl it in every subclasses.

### DIFF
--- a/src/search/ir.h
+++ b/src/search/ir.h
@@ -48,6 +48,8 @@ struct Node {
   virtual NodeIterator ChildBegin() { return {}; };
   virtual NodeIterator ChildEnd() { return {}; };
 
+  virtual std::unique_ptr<Node> Clone() const = 0;
+
   virtual ~Node() = default;
 
   template <typename T, typename U = Node, typename... Args>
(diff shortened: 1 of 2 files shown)


src/search/ir.h
```diff
@@ -106,16 +116,23 @@ struct TagContainExpr : BoolAtomExpr {
...
   NodeIterator ChildEnd() override { return {}; };
+
+  std::unique_ptr<Node> Clone() const override {
+    return std::make_unique<TagContainExpr>(Node::MustAs<FieldRef>(field->Clone()),
+                                            Node::MustAs<StringLiteral>(tag->Clone()));
+  }
 };
 
-struct NumericLiteral : Node {
+struct NumericLiteral : Literal {
```
mapleFU, line 126: Btw, do we have a Literal class for all literals(which is not a node), which could be used in filters?

mapleFU (commented):
General LGTM

git-hulk (approved):

src/search/ir.h
PragmaTwice: Currently no. Is it necessary?

PR #2236, opened 2024-04-11
Fix additional parameter parsing in SCAN

jackyyyyyssss:
When kvrocks executes a scan command, such as SCAN 0 key match * he 1000, it will change the originally recognized match to a key without reporting any errors. Finally, the value of prefix_ will match all keys, and this command will report an error under normal Redis. Add mandatory matching verification code and use the current test case for  testing

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -828,7 +828,7 @@ class CommandScan : public CommandScanBase {
         if (!prefix_.empty() && prefix_[prefix_.size() - 1] == '*') {
           prefix_ = prefix_.substr(0, prefix_.size() - 1);
         } else {
-          return {Status::RedisParseErr, "only keys prefix match was supported"};
+          return {Status::RedisParseErr, "currently only key prefix matching is supported"};
         }
       } else if (parser.EatEqICase("count")) {
         limit_ = GET_OR_RET(parser.TakeInt());
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -823,9 +823,7 @@ class CommandScan : public CommandScanBase {
     CommandParser parser(args, 1);
     ParseCursor(GET_OR_RET(parser.TakeStr()));
     while (parser.Good()) {
-      if (parser.EatEqICase("scan")) {
-        ParseCursor(GET_OR_RET(parser.TakeStr()));
-      } else if (parser.EatEqICase("match")) {
+      if (parser.EatEqICase("match")) {
         prefix_ = GET_OR_RET(parser.TakeStr());
         if (!prefix_.empty() && prefix_[prefix_.size() - 1] == '*') {
           prefix_ = prefix_.substr(0, prefix_.size() - 1);
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -826,18 +826,9 @@ class CommandScan : public CommandScanBase {
 
     ParseCursor(args[1]);
     if (args.size() >= 4) {
-      std::string match_flag = util::ToLower(args[2]);
-      if (match_flag != "match") {
-        return {Status::RedisParseErr, errWrongNumOfArguments};
-      }
-      Status s = ParseMatchAndCountParam(match_flag, args_[3]);
-      if (!s.IsOK()) {
-        return s;
-      }
-    }
-
-    if (args.size() >= 6) {
-      Status s = ParseMatchAndCountParam(util::ToLower(args[4]), args_[5]);
+      ScanParameters parameters;
+      Status s = ParseScanParameters(args, &parameters);
+      FillScanAttributes(&parameters);
       if (!s.IsOK()) {
         return s;
       }
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -836,7 +836,7 @@ class CommandScan : public CommandScanBase {
           return {Status::RedisParseErr, errInvalidSyntax};
         }
       } else if (parser.EatEqICase("type")) {
-        return {Status::RedisParseErr, "type was not supported"};
+        return {Status::RedisParseErr, "TYPE flag is currently not supported"};
       } else {
         return {Status::RedisParseErr, errWrongNumOfArguments};
       }
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -819,32 +819,6 @@ class CommandScan : public CommandScanBase {
  public:
   CommandScan() : CommandScanBase() {}
 
-  Status Parse(const std::vector<std::string> &args) override {
-    CommandParser parser(args, 1);
-    ParseCursor(GET_OR_RET(parser.TakeStr()));
-    while (parser.Good()) {
-      if (parser.EatEqICase("match")) {
-        prefix_ = GET_OR_RET(parser.TakeStr());
-        if (!prefix_.empty() && prefix_.back() == '*') {
-          prefix_ = prefix_.substr(0, prefix_.size() - 1);
-        } else {
-          return {Status::RedisParseErr, "currently only key prefix matching is supported"};
-        }
-      } else if (parser.EatEqICase("count")) {
-        limit_ = GET_OR_RET(parser.TakeInt());
-        if (limit_ <= 0) {
-          return {Status::RedisParseErr, errInvalidSyntax};
-        }
-      } else if (parser.EatEqICase("type")) {
-        return {Status::RedisParseErr, "TYPE flag is currently not supported"};
-      } else {
-        return {Status::RedisParseErr, errWrongNumOfArguments};
-      }
-    }
-
-    return Commander::Parse(args);
-  }
-
   static std::string GenerateOutput(Server *srv, const Connection *conn, const std::vector<std::string> &keys,
                                     const std::string &end_cursor) {
     std::vector<std::string> list;
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -842,7 +842,11 @@ class CommandScan : public CommandScanBase {
 
     ParseCursor(args[1]);
     if (args.size() >= 4) {
-      Status s = ParseMatchAndCountParam(util::ToLower(args[2]), args_[3]);
+      std::string match_flag= util::ToLower(args[2]);
+      if (match_flag!="match"){
+        return {Status::RedisParseErr, errWrongNumOfArguments};
+      }
+      Status s = ParseMatchAndCountParam(match_flag, args_[3]);
       if (!s.IsOK()) {
         return s;
       }
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -820,10 +820,8 @@ class CommandScan : public CommandScanBase {
   CommandScan() : CommandScanBase() {}
 
   Status Parse(const std::vector<std::string> &args) override {
-    if (args.size() % 2 != 0) {
-      return {Status::RedisParseErr, errWrongNumOfArguments};
-    }
-    CommandParser parser(args, 0);
+    CommandParser parser(args, 1);
+    ParseCursor(GET_OR_RET(parser.TakeStr()));
     while (parser.Good()) {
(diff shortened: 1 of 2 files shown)


src/commands/cmd_server.cc
```diff
@@ -826,7 +826,11 @@ class CommandScan : public CommandScanBase {
 
     ParseCursor(args[1]);
     if (args.size() >= 4) {
-      Status s = ParseMatchAndCountParam(util::ToLower(args[2]), args_[3]);
+      std::string match_flag = util::ToLower(args[2]);
```
git-hulk: https://redis.io/docs/latest/commands/scan/

The third parameter from the scan command documentation can also be TYPE/COUNT. We can improve the `ParseMatchAndCountParam` instead of a temporary fix here.

My initial idea is like below:
```
struct ScanParameters {
  std::string Match;
  int64_t count;
  RedisType type; // only allowed in SCAN command
}

Status ParseScanParameters(const std::vector<std::string> &args, ScanParameters *params) {
   // ...
}
```

jackyyyyyssss: @git-hulk Thank you for your CR. I have already made the necessary modifications PTAL

git-hulk: @jackyyyyyssss sure, thank you!

src/commands/scan_base.h
```diff
@@ -79,6 +85,38 @@ class CommandScanBase : public Commander {
     return redis::Array(list);
   }
 
+  static Status ParseScanParameters(const std::vector<std::string> &args, ScanParameters *params) {
+    for (size_t i = 2; i < args.size(); i = i + 2) {
```
PragmaTwice: Could you implement this function via `CommandParser`?

src/commands/scan_base.h
```diff
@@ -79,6 +85,38 @@ class CommandScanBase : public Commander {
...
+      } else if (util::ToLower(args[i]) == "count") {
+        auto parse_result = ParseInt<int>(args[i + 1], 10);
+        if (!parse_result) {
+          return {Status::RedisParseErr, "count param should be type int"};
+        }
+        params->count = *parse_result;
+        if (params->count <= 0) {
+          return {Status::RedisParseErr, errInvalidSyntax};
+        }
+      } else if (util::ToLower(args[i]) == "type") {
```
PragmaTwice: Please try to imagine what will happen if the input is `scan xxx type`.

src/commands/scan_base.h
```diff
@@ -79,6 +85,38 @@ class CommandScanBase : public Commander {
...
+        return {Status::RedisParseErr, errWrongNumOfArguments};
+      }
+    }
+
+    return Status::OK();
+  }
+  void FillScanAttributes(const ScanParameters *params) {
+    prefix_ = params->match;
+    limit_ = params->count;
+  }
```
PragmaTwice: No need. Please just assign these values in `ParseScanParameters`.

src/commands/cmd_server.cc
```diff
@@ -823,16 +823,33 @@ class CommandScan : public CommandScanBase {
...
+        ParseCursor(GET_OR_RET(parser.TakeStr()));
+      } else if (parser.EatEqICase("match")) {
+        prefix_ = GET_OR_RET(parser.TakeStr());
+        if (!prefix_.empty() && prefix_[prefix_.size() - 1] == '*') {
+          prefix_ = prefix_.substr(0, prefix_.size() - 1);
+        } else {
+          return {Status::RedisParseErr, "only keys prefix match was supported"};
+        }
+      } else if (parser.EatEqICase("count")) {
+        auto parse_result = ParseInt<int>(GET_OR_RET(parser.TakeStr()), 10);
```
PragmaTwice: Please use `parser.TakeInt()`

src/commands/cmd_server.cc
```diff
@@ -823,16 +823,33 @@ class CommandScan : public CommandScanBase {
...
-    if (args.size() >= 4) {
-      ScanParameters parameters;
-      Status s = ParseScanParameters(args, &parameters);
-      FillScanAttributes(&parameters);
-      if (!s.IsOK()) {
-        return s;
+    CommandParser parser(args, 0);
+    while (parser.Good()) {
+      if (parser.EatEqICase("scan")) {
+        ParseCursor(GET_OR_RET(parser.TakeStr()));
```
PragmaTwice: It's not a correct implementation. e.g. please consider if the input is `match ..`. `SCAN` itself cannot be parsed as an additional flag.
You should implement it like:
```
ParseCursor(GET_OR_RET(parser.TakeStr()));
while (..) ...
```

src/commands/cmd_server.cc
```diff
@@ -823,16 +823,33 @@ class CommandScan : public CommandScanBase {
     if (args.size() % 2 != 0) {
       return {Status::RedisParseErr, errWrongNumOfArguments};
     }
```
PragmaTwice: Please remove this.

jackyyyyyssss: I am a C++beginner. Thank you very much for your guidance on my code. I have learned a lot

src/commands/cmd_server.cc
```diff
@@ -820,24 +820,28 @@ class CommandScan : public CommandScanBase {
...
+        } else {
+          return {Status::RedisParseErr, "only keys prefix match was supported"};
+        }
+      } else if (parser.EatEqICase("count")) {
+        limit_ = GET_OR_RET(parser.TakeInt());
+        if (limit_ <= 0) {
+          return {Status::RedisParseErr, errInvalidSyntax};
+        }
+      } else if (parser.EatEqICase("type")) {
+        return {Status::RedisParseErr, "type was not supported"};
```
```suggestion
        return {Status::RedisParseErr, "TYPE flag is currently not supported"};
```

src/commands/cmd_server.cc
```diff
@@ -820,24 +820,28 @@ class CommandScan : public CommandScanBase {
...
-        return s;
+    CommandParser parser(args, 1);
+    ParseCursor(GET_OR_RET(parser.TakeStr()));
+    while (parser.Good()) {
+      if (parser.EatEqICase("match")) {
+        prefix_ = GET_OR_RET(parser.TakeStr());
+        if (!prefix_.empty() && prefix_[prefix_.size() - 1] == '*') {
+          prefix_ = prefix_.substr(0, prefix_.size() - 1);
+        } else {
+          return {Status::RedisParseErr, "only keys prefix match was supported"};
```
```suggestion
          return {Status::RedisParseErr, "currently only key prefix matching is supported"};
```

src/commands/cmd_server.cc
```diff
@@ -820,24 +820,28 @@ class CommandScan : public CommandScanBase {
...
-    if (args.size() >= 4) {
-      Status s = ParseMatchAndCountParam(util::ToLower(args[2]), args_[3]);
-      if (!s.IsOK()) {
-        return s;
+    CommandParser parser(args, 1);
+    ParseCursor(GET_OR_RET(parser.TakeStr()));
+    while (parser.Good()) {
+      if (parser.EatEqICase("match")) {
+        prefix_ = GET_OR_RET(parser.TakeStr());
+        if (!prefix_.empty() && prefix_[prefix_.size() - 1] == '*') {
```
```suggestion
        if (!prefix_.empty() && prefix_.back() == '*') {
```

git-hulk:
@jackyyyyyssss CommandSubkeyScanBase should have the same issue which is used for HSCAN/SSCAN/ZSCAN: https://redis.io/docs/latest/commands/sscan/.

jackyyyyyssss:
> @jackyyyyyssss CommandSubkeyScanBase should have the same issue which is used for HSCAN/SSCAN/ZSCAN: https://redis.io/docs/latest/commands/sscan/.


Okay, can I use different PR for these modifications？

git-hulk:
@jackyyyyyssss Perhaps you can take a look, it's in the same file and they share the same parse logic except the CommandSubkeyScanBase starts parsing the match/count from the 2nd parameter, and type is not allowed.

git-hulk:
@jackyyyyyssss Can remove ParseMatchAndCountParam if no place is using it.

src/commands/scan_base.h
```diff
@@ -90,23 +91,24 @@ class CommandSubkeyScanBase : public CommandScanBase {
...
+        } else {
+          return {Status::RedisParseErr, "currently only key prefix matching is supported"};
+        }
+      } else if (parser.EatEqICase("count")) {
+        limit_ = GET_OR_RET(parser.TakeInt());
+        if (limit_ <= 0) {
+          return {Status::RedisParseErr, errInvalidSyntax};
+        }
+      } else {
+        return {Status::RedisParseErr, errWrongNumOfArguments};
```
PragmaTwice: Please avoid [copy and paste programming](https://en.wikipedia.org/wiki/Duplicate_code).

jackyyyyyssss: I have a question about whether this parsing is written in CommandScanBase, or if both have differences written in their respective places, with the public part written in base. However, if so, what should be the template value for CommandParser as a method parameter? like this  
   template < typename Iter >
  Status ProcessCommonParser(CommandParser<Iter>  &parser) {..}

jackyyyyyssss:
> @jackyyyyyssss Can remove ParseMatchAndCountParam if no place is using it.

ok

src/commands/scan_base.h
```diff
@@ -31,28 +32,33 @@ inline constexpr const char *kCursorPrefix = "_";
...
-      if (!parse_result) {
-        return {Status::RedisParseErr, "count param should be type int"};
-      }
-
-      limit_ = *parse_result;
-      if (limit_ <= 0) {
-        return {Status::RedisParseErr, errInvalidSyntax};
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 1);
+    if (util::ToLower(args[0]) != "scan") {
```
PragmaTwice: Please use parameters instead of such check.

src/commands/scan_base.h
```diff
@@ -31,28 +32,33 @@ inline constexpr const char *kCursorPrefix = "_";
...
+        } else {
+          return {Status::RedisParseErr, "currently only key prefix matching is supported"};
+        }
+      } else if (parser.EatEqICase("count")) {
+        limit_ = GET_OR_RET(parser.TakeInt());
+        if (limit_ <= 0) {
+          return {Status::RedisParseErr, errInvalidSyntax};
+        }
+      } else if (parser.EatEqICase("type")) {
+        // HSCAN SSCAN ZSCAN Will not enter
```
PragmaTwice: Why do you think that "HSCAN SSCAN ZSCAN will not enter" this block?

jackyyyyyssss: Maybe my description is a bit inaccurate. After other commands are added, error prompts will be return

src/commands/scan_base.h
```diff
@@ -31,28 +32,33 @@ inline constexpr const char *kCursorPrefix = "_";
...
-        return {Status::RedisParseErr, "count param should be type int"};
-      }
-
-      limit_ = *parse_result;
-      if (limit_ <= 0) {
-        return {Status::RedisParseErr, errInvalidSyntax};
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 1);
+    if (util::ToLower(args[0]) != "scan") {
+      key_ = GET_OR_RET(parser.TakeStr());
```
PragmaTwice: This should be only in `CommandSubkeyScanBase`.

PragmaTwice:
I will prepare a patch to succeed this PR.

jackyyyyyssss:
> I will prepare a patch to succeed this PR.

Can you help me take a look at the optimized version thks

PragmaTwice:
It's more close. But I think we can save some communication effort by my patch..

I'll make you the co-author of the patch, and hope you can learn some C++ knowledege from my code.

Also, in the next time, please don't forget to write test cases for your patch, as it's mandatory.

jackyyyyyssss:
> It's more close. But I think we can save some communication effort by my patch..
> 
> I'll make you the co-author of the patch, and hope you can learn some C++ knowledege from my code.
> 
> Also, in the next time, please don't forget to write test cases for your patch, as it's mandatory.


Thank you for your patient answer

PR #2239, opened 2024-04-11
Rename SelectExpr to SelectClause to prevent misunderstanding

PragmaTwice:
`SelectExpr` is not an appropriate name since it can represent the whole `select .. from ..` expression.

### DIFF
--- a/src/search/ir.h
+++ b/src/search/ir.h
@@ -322,10 +322,10 @@ struct SortByClause : Node {
   }
 };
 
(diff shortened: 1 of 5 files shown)


mapleFU (dismissed):
+1 for this, expr should be simplier, select should be clause/node/operator...

torwig (dismissed):

torwig (approved):

git-hulk (approved):

PR #2247, opened 2024-04-14
Add plan lowering pass for KQIR

PragmaTwice:
In this PR, we added a pass to lower original IR to planner IR.

There's no optimization and index selection in this pass, and these pass will be seperated to another following pass.

### DIFF
--- a/src/search/passes/lower_to_plan.h
+++ b/src/search/passes/lower_to_plan.h
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


src/search/passes/lower_to_plan.h
```diff
@@ -0,0 +1,51 @@
...
+struct LowerToPlan : Visitor {
+  std::unique_ptr<Node> Visit(std::unique_ptr<SearchExpr> node) override {
+    auto scan = std::make_unique<FullIndexScan>(node->index->CloneAs<IndexRef>());
+    auto filter = std::make_unique<Filter>(std::move(scan), std::move(node->query_expr));
+
+    std::unique_ptr<PlanOperator> op = std::move(filter);
+
+    // order is important here, since limit(sort(op)) is different from sort(limit(op))
+    if (node->sort_by) {
+      op = std::make_unique<Sort>(std::move(op), std::move(node->sort_by));
```
mapleFU, line 40: why does this called lower? Since it extract a Sort node, don't know what's the naming rule here

PragmaTwice: lowering is a term in compiler. especially in IR optimization.

https://news.ycombinator.com/item?id=26928724

PragmaTwice: If you want more detail:
here we lower the IR from a syntactically-intuitive structure to the planning subset.

git-hulk (approved):

mapleFU (approved):

PR #2249, opened 2024-04-14
Fix change type of `ttl_` to `int64_t` in `commands` to avoid disturbing

### LINKED ISSUE #2248: bug: inappropriate type of `ttl_` of `CommandSet`
### Search before asking

### Version

Since db2860d until now.

### Minimal reproduce step

https://github.com/apache/kvrocks/blob/a2380535d5602692f3746783c48ccc7bb05cc58e/src/commands/cmd_string.cc#L307

Here determine whether `ttl_` less than `0`, if so, doing deletion. However, the type of `ttl_` if `uint64_t` which is impossible to less than 0.

https://github.com/apache/kvrocks/blob/a2380535d5602692f3746783c48ccc7bb05cc58e/src/commands/cmd_string.cc#L340

If `ttl_` less than `0`, doing deletion.

### What did you see instead?

The condition will never be true, thus, deletion will never be done.

### Anything Else?

_No response_

J0HN50N133:
fix #2248

### DIFF
--- a/src/commands/cmd_key.cc
+++ b/src/commands/cmd_key.cc
@@ -186,7 +186,7 @@ class CommandExpire : public Commander {
   }
 
  private:
-  uint64_t ttl_ = 0;
+  int64_t ttl_ = 0;
 };
 
 class CommandPExpire : public Commander {
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -1117,7 +1117,7 @@ class CommandRestore : public Commander {
  private:
   bool replace_ = false;
   bool absttl_ = false;
-  uint64_t ttl_ms_ = 0;
+  int64_t ttl_ms_ = 0;
 };
 
 // command format: rdb load <path> [NX]  [DB index]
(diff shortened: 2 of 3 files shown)


J0HN50N133:
@git-hulk I've modified all code with similar problem in #2248

mapleFU (approved):
@PragmaTwice should we enable some checks for signed-unsigned number casting?

git-hulk (approved):

PragmaTwice:
@J0HN50N133 Hi, could you revert your changes in `cmd_string.cc`? 

I'll handle it since it's a bit complicated than just changing a type.

Also it makes the CI fail to pass.

PragmaTwice:
> I'll handle it since it's a bit complicated than just changing a type.

See #2250.

J0HN50N133:
> @J0HN50N133 Hi, could you revert your changes in `cmd_string.cc`?
> 
> I'll handle it since it's a bit complicated than just changing a type.
> 
> Also it makes the CI fail to pass.

Ok.

PragmaTwice:
Actually you can keep the PR with your changes (except `cmd_string.cc`), but it's also fine if you don't want to continue : )

Thank you for your contribution.

PR #2250, opened 2024-04-14
Refactor TTL processing in string type

### LINKED ISSUE #2248: bug: inappropriate type of `ttl_` of `CommandSet`

PragmaTwice:
Previously, we use TTL as a parameter in string operations, and it's not good since we should use an absolute value in internal implementations to avoid double casting (`EXAT timestamp` -> cast to TTL -> cast back to expire timestamp, these introduce two `GetTimestampMS()` with probably different result).

Here we change TTL to timestamp as the parameter of string operations.

Close #2248.

### DIFF
--- a/src/commands/cmd_string.cc
+++ b/src/commands/cmd_string.cc
@@ -66,10 +66,10 @@ class CommandGetEx : public Commander {
     CommandParser parser(args, 2);
     std::string_view ttl_flag;
     while (parser.Good()) {
-      if (auto v = GET_OR_RET(ParseTTL(parser, ttl_flag))) {
-        ttl_ = *v;
+      if (auto v = GET_OR_RET(ParseExpireFlags(parser, ttl_flag))) {
+        expire_ = *v;
       } else if (parser.EatEqICaseFlag("PERSIST", ttl_flag)) {
-        persist_ = true;
+        expire_ = 0;
       } else {
         return parser.InvalidSyntax();
       }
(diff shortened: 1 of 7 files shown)


git-hulk (approved):

src/commands/cmd_string.cc
```diff
@@ -107,8 +103,7 @@ class CommandGetEx : public Commander {
   }
 
  private:
-  uint64_t ttl_ = 0;
-  bool persist_ = false;
+  std::optional<uint64_t> expire_;
```
mapleFU, line 106: Would we have a class wrapper to distinguish the syntax?

PragmaTwice: Is it necessary?

mapleFU: I think a interface with something like `has_ttl()` would look better, but feel free to just merge this

PR #2255, opened 2024-04-19
Support TYPE flag in the SCAN command

jackyyyyyssss:
support  scan  additional type

### DIFF
--- a/src/commands/scan_base.h
+++ b/src/commands/scan_base.h
@@ -56,9 +56,10 @@ class CommandScanBase : public Commander {
           return {Status::RedisParseErr, "limit should be a positive integer"};
         }
       } else if (IsScan && parser.EatEqICase("type")) {
-        std::string key = GET_OR_RET(parser.TakeStr());
-        if (RedisNamesMapType.count(key)) {
-          type_ = RedisNamesMapType.at(key);
+        std::string type_str = GET_OR_RET(parser.TakeStr());
+        if (auto iter = std::find(RedisTypeNames.begin(), RedisTypeNames.end(), type_str);
+            iter != RedisTypeNames.end()) {
+          type_ = static_cast<RedisType>(iter - RedisTypeNames.begin());
         } else {
           return {Status::RedisExecErr, "Invalid type"};
         }
--- a/src/commands/scan_base.h
+++ b/src/commands/scan_base.h
@@ -56,7 +56,12 @@ class CommandScanBase : public Commander {
           return {Status::RedisParseErr, "limit should be a positive integer"};
         }
       } else if (IsScan && parser.EatEqICase("type")) {
-        return {Status::RedisParseErr, "TYPE flag is currently not supported"};
+        std::string key = util::ToLower(GET_OR_RET(parser.TakeStr()));
+        if (RedisNamesMapType.count(key)) {
+          type_ = RedisNamesMapType.at(key);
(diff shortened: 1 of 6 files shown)


src/commands/scan_base.h
```diff
@@ -56,7 +56,12 @@ class CommandScanBase : public Commander {
           return {Status::RedisParseErr, "limit should be a positive integer"};
         }
       } else if (IsScan && parser.EatEqICase("type")) {
-        return {Status::RedisParseErr, "TYPE flag is currently not supported"};
+        std::string key = GET_OR_RET(parser.TakeStr());
+        if (RedisNamesMapType.count(key)) {
+          type_ = RedisNamesMapType.at(key);
+        } else {
+          return {Status::RedisExecErr, "Invalid type"};
+        }
```
PragmaTwice:
```suggestion
        if (auto iter = std::find(RedisTypeNames.begin(), RedisTypeNames.end(), type_str); iter != RedisTypeNames.end()) {
          type_ = static_cast<RedisType>(iter - RedisTypeNames.begin());
        } else {
          return {Status::RedisExecErr, "Invalid type"};
        }
```

src/commands/scan_base.h
```diff
@@ -56,7 +56,12 @@ class CommandScanBase : public Commander {
           return {Status::RedisParseErr, "limit should be a positive integer"};
         }
       } else if (IsScan && parser.EatEqICase("type")) {
-        return {Status::RedisParseErr, "TYPE flag is currently not supported"};
+        std::string key = GET_OR_RET(parser.TakeStr());
```
PragmaTwice:
```suggestion
        std::string type_str = GET_OR_RET(parser.TakeStr());
```

what's `key`?

src/storage/redis_metadata.h
```diff
@@ -94,6 +95,12 @@ enum RedisCommand {
 const std::vector<std::string> RedisTypeNames = {"none",   "string",    "hash",   "list",      "set",      "zset",
                                                  "bitmap", "sortedint", "stream", "MBbloom--", "ReJSON-RL"};
 
+const std::map<std::string, RedisType> RedisNamesMapType = {
+    {"none", kRedisNone},     {"string", kRedisString}, {"hash", kRedisHash},
+    {"list", kRedisList},     {"set", kRedisSet},       {"zset", kRedisZSet},
+    {"bitmap", kRedisBitmap}, {"stream", kRedisStream}, {"MBbloom--", kRedisBloomFilter},
+    {"ReJSON-RL", kRedisJson}};
+
```
PragmaTwice: Please remove this.

jackyyyyyssss: Thank you for your suggestion. I originally intended this design to reduce time complexity, but do you think it's not necessary to have too few array elements

PragmaTwice: Yeah, the maintenance cost will be far more than the performance benefit we get.

jackyyyyyssss: I understand. Thank you for your answer. I also missed an unverified type of test case. I will run the test case on the local branch first, and once it passes, please take a look on your end

PragmaTwice (approved):

git-hulk (approved):

jihuayu (approved):
Good work, thank you!

Issue #2193, opened 2024-03-23
Reduce the number of write buffers for infrequently used column families


git-hulk:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Currently, all column families are shared the same number of write buffers
which is unnecessary for infrequently used column families including `pubsub`,
`propagte`, and `zset_score`. We can keep them to a const value like 1 to reduce
the memory usage in some scenarios.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

LiuYuHui:
Please assign it to me, thanks a lot :D

git-hulk:
@LiuYuHui Thank you!

git-hulk:
Hi @LiuYuHui Are you still working on this issue?

LiuYuHui:
Hi @git-hulk, I have had no extra time bandwidth recently, please feel free to assign it to other people.

git-hulk:
@LiuYuHui Thanks for your quick response. Welcome back while you get time.

jjz921024:
Hi @git-hulk , May i question? How do we implement this feature? 

1. For each column family add a config, specifying its number of write buffers. eg: `pubsub.max_write_buffer_number`、`propagte.max_write_buffer_number`、`zset_score.max_write_buffer_number` ...
2. Add a `lite` mode config, When this config is set to `true`, create a new `lite_table_opts` for infrequently used column families. like:

```java
rocksdb::BlockBasedTableOptions lite_table_opts = InitTableOptions();
rocksdb::ColumnFamilyOptions lite_opts(subkey_opts);
lite_opts.write_buffer_size = 32 * MiB;
lite_opts.max_write_buffer_number = 0;

// replace subkey_opts with lite_opts 
column_families.emplace_back(kZSetScoreColumnFamilyName, lite_opts);
column_families.emplace_back(kPubSubColumnFamilyName, lite_opts);
column_families.emplace_back(kPropagateColumnFamilyName, lite_opts);
column_families.emplace_back(kStreamColumnFamilyName, lite_opts);```

git-hulk:
@jjz921024 Yes, I'm good with this implementation. Would you like to submit a PR?

jjz921024:
Yes, I want to try

git-hulk:
@jjz921024 Assigned, thank you!

PR #2258, opened 2024-04-20
Add a option to specify write_buffer_size for minor column families

jjz921024:
For #2193 

Add a `lite_mode` option to special kvrock running at lite_mode,
At this mode, the `lite_opts` will replace for `subkey_opts` on some infrequently used column families 
for reduce memory usage in some scenarios.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -333,8 +333,8 @@ json-storage-format json
 # including pubsub, propagte, zset_score, stream and search. 
 # It will reduce the memory usage in some scenarios.
 #
-# Default: 65536 KB (same with rocksdb.write_buffer_size)
-minor-columns-write-buffer-size 65536
+# Default: 64 MB (same with rocksdb.write_buffer_size)
+minor-columns-write-buffer-size 64
 
 ################################## TLS ###################################
 
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -329,6 +329,13 @@ json-max-nesting-depth 1024
 # Default: json
 json-storage-format json
 
+# The write buffer size of minor column families those infrequently be used
+# including pubsub, propagte, zset_score, stream and search. 
+# It will reduce the memory usage in some scenarios.
+#
+# Default: 65536 KB (same with rocksdb.write_buffer_size)
+minor-columns-write-buffer-size 65536
+
 ################################## TLS ###################################
 
 # By default, TLS/SSL is disabled, i.e. `tls-port` is set to 0.
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -190,7 +190,7 @@ Config::Config() {
       {"json-max-nesting-depth", false, new IntField(&json_max_nesting_depth, 1024, 0, INT_MAX)},
       {"json-storage-format", false,
        new EnumField<JsonStorageFormat>(&json_storage_format, json_storage_formats, JsonStorageFormat::JSON)},
-      {"minor-columns-write-buffer-size", false, new IntField(&minor_columns_write_buffer_size, 16, 0, 4194304)},
+      {"minor-columns-write-buffer-size", false, new IntField(&minor_columns_write_buffer_size, 65536, 16, 4194304)},
 
       /* rocksdb options */
       {"rocksdb.compression", false,
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -41,6 +41,7 @@
 #include "server/server.h"
 #include "status.h"
 #include "storage/redis_metadata.h"
+#include "storage/storage.h"
 
 constexpr const char *kDefaultDir = "/tmp/kvrocks";
 constexpr const char *kDefaultBackupDir = "/tmp/kvrocks/backup";
@@ -189,6 +190,7 @@ Config::Config() {
       {"json-max-nesting-depth", false, new IntField(&json_max_nesting_depth, 1024, 0, INT_MAX)},
       {"json-storage-format", false,
        new EnumField<JsonStorageFormat>(&json_storage_format, json_storage_formats, JsonStorageFormat::JSON)},
+      {"minor-columns-write-buffer-size", false, new IntField(&minor_columns_write_buffer_size, 16, 0, 4194304)},
 
       /* rocksdb options */
       {"rocksdb.compression", false,
@@ -574,6 +576,19 @@ void Config::initFieldCallback() {
            [](Server *srv, const std::string &k, const std::string &v) -> Status {
              if (!srv) return Status::OK();
              return srv->GetNamespace()->LoadAndRewrite();
+           }},
+          {"minor-columns-write-buffer-size",
+           [this](Server *srv, const std::string &k, const std::string &v) -> Status {
+             if (!srv) return Status::OK();
+             const std::vector<ColumnFamilyID> column_families = {kColumnFamilyIDZSetScore, kColumnFamilyIDPubSub,
+                                                                  kColumnFamilyIDPropagate, kColumnFamilyIDStream,
+                                                                  kColumnFamilyIDSearch};
+             for (const auto &cf : column_families) {
+               auto s = srv->storage->SetOptionForColumnFamily(cf, "write_buffer_size",
+                                                               std::to_string(minor_columns_write_buffer_size * KiB));
+               if (!s.IsOK()) return s;
+             }
+             return Status::OK();
            }},
 
           {"rocksdb.target_file_size_base",
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -190,7 +190,7 @@ Config::Config() {
       {"json-max-nesting-depth", false, new IntField(&json_max_nesting_depth, 1024, 0, INT_MAX)},
       {"json-storage-format", false,
        new EnumField<JsonStorageFormat>(&json_storage_format, json_storage_formats, JsonStorageFormat::JSON)},
-      {"minor-columns-write-buffer-size", false, new IntField(&minor_columns_write_buffer_size, 65536, 16, 4194304)},
+      {"minor-columns-write-buffer-size", false, new IntField(&minor_columns_write_buffer_size, 64, 0, 4096)},
 
       /* rocksdb options */
       {"rocksdb.compression", false,
@@ -585,7 +585,7 @@ void Config::initFieldCallback() {
                                                                   kColumnFamilyIDSearch};
              for (const auto &cf : column_families) {
                auto s = srv->storage->SetOptionForColumnFamily(cf, "write_buffer_size",
-                                                               std::to_string(minor_columns_write_buffer_size * KiB));
+                                                               std::to_string(minor_columns_write_buffer_size * MiB));
                if (!s.IsOK()) return s;
              }
              return Status::OK();
--- a/src/config/config.h
+++ b/src/config/config.h
@@ -119,6 +119,7 @@ struct Config {
   bool auto_resize_block_and_sst = true;
   int fullsync_recv_file_delay = 0;
   bool use_rsid_psync = false;
+  int minor_columns_write_buffer_size;
   std::vector<std::string> binds;
   std::string dir;
   std::string db_dir;
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -37,6 +37,7 @@
 #include <random>
 
 #include "compact_filter.h"
+#include "config/config.h"
 #include "db_util.h"
 #include "event_listener.h"
 #include "event_util.h"
@@ -226,6 +227,12 @@ Status Storage::SetOptionForAllColumnFamilies(const std::string &key, const std:
   return Status::OK();
 }
 
+Status Storage::SetOptionForColumnFamily(ColumnFamilyID id, const std::string &key, const std::string &value) {
+  auto s = db_->SetOptions(GetCFHandle(id), {{key, value}});
+  if (!s.ok()) return {Status::NotOK, s.ToString()};
+  return Status::OK();
+}
+
 Status Storage::SetDBOption(const std::string &key, const std::string &value) {
   auto s = db_->SetDBOptions({{key, value}});
   if (!s.ok()) return {Status::NotOK, s.ToString()};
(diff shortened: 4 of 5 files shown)


src/storage/storage.cc
```diff
@@ -337,15 +337,26 @@ Status Storage::Open(DBOpenMode mode) {
   propagate_opts.disable_auto_compactions = config_->rocks_db.disable_auto_compactions;
   SetBlobDB(&propagate_opts);
 
+  rocksdb::ColumnFamilyOptions lite_opts(subkey_opts);
+  if (config_->lite_mode) {
```
mapleFU: Seems the logic is write twice? (generate a lite_opts and update all previous opts)

src/storage/storage.cc
```diff
@@ -337,15 +337,26 @@ Status Storage::Open(DBOpenMode mode) {
...
+    lite_opts.write_buffer_size = 1L;
```
git-hulk: We can keep the write_buffer_size as a relatively smaller one like 16MiB/32MiB instead of 1 byte.

src/config/config.cc
```diff
@@ -189,6 +189,7 @@ Config::Config() {
       {"json-max-nesting-depth", false, new IntField(&json_max_nesting_depth, 1024, 0, INT_MAX)},
       {"json-storage-format", false,
        new EnumField<JsonStorageFormat>(&json_storage_format, json_storage_formats, JsonStorageFormat::JSON)},
+      {"lite-mode", false, new YesNoField(&lite_mode, false)},
```
git-hulk: It should be necessary to add an option to enable this since it's safe to shrink the infrequently used column families. cc @mapleFU @torwig @PragmaTwice

git-hulk:
@jjz921024 Thanks for your contributions.

src/storage/storage.cc
jjz921024: Yes, It seems that the min value is 16KiB ? Even though it is set to 1, the rocksdb log still shows `Options.write_buffer_size: 65536`.

PragmaTwice:
I believe "lite mode" is not a suitable name as it does not clearly convey its purpose.

We should find a more specific name instead of labeling any random behavior as "lite mode".

PragmaTwice (changes requested):
See above.

src/storage/storage.cc
jjz921024: Because the lite_opts need to inherit all of subkey_opts properties, so create a `lite_opts` using `lite_opts(subkey_opts)` ?

src/storage/storage.cc
git-hulk: 16KiB is good for me.

jjz921024: done

jjz921024:
> I believe "lite mode" is not a suitable name as it does not clearly convey its purpose.
> 
> We should find a more specific name instead of labeling any random behavior as "lite mode".

Maybe we can rename this option to `reduce_infrequently_column_memory_usage` so that explicitly tell the user that it is used to reduce the memory size of an infrequently used column family?

src/config/config.cc
jjz921024: @git-hulk Hi，I submit a commit in this pr to support dynamic configuration of this option，I'm not sure that's what you mean，please review

src/config/config.cc
```diff
@@ -575,6 +577,40 @@ void Config::initFieldCallback() {
              if (!srv) return Status::OK();
              return srv->GetNamespace()->LoadAndRewrite();
            }},
+          {"reduce-infrequent-column-memory-usage",
+           [](Server *srv, const std::string &k, const std::string &v) -> Status {
+             if (!srv) return Status::OK();
+
+             std::string buff_size;
+             if (v == "yes") {
+               buff_size = std::to_string(16 * KiB);
```
PragmaTwice: It should be a named constant.

src/storage/storage.cc
```diff
@@ -337,15 +344,22 @@ Status Storage::Open(DBOpenMode mode) {
   propagate_opts.disable_auto_compactions = config_->rocks_db.disable_auto_compactions;
   SetBlobDB(&propagate_opts);
 
+  rocksdb::ColumnFamilyOptions lite_opts(subkey_opts);
+  if (config_->reduce_infrequent_column_memory_usage) {
+    lite_opts.write_buffer_size = 16 * KiB;
+    pubsub_opts.write_buffer_size = 16 * KiB;
+    propagate_opts.write_buffer_size = 16 * KiB;
+  }
```
PragmaTwice: ditto

src/storage/storage.cc
```diff
@@ -337,15 +344,22 @@ Status Storage::Open(DBOpenMode mode) {
   propagate_opts.disable_auto_compactions = config_->rocks_db.disable_auto_compactions;
   SetBlobDB(&propagate_opts);
 
+  rocksdb::ColumnFamilyOptions lite_opts(subkey_opts);
+  if (config_->reduce_infrequent_column_memory_usage) {
+    lite_opts.write_buffer_size = 16 * KiB;
+    pubsub_opts.write_buffer_size = 16 * KiB;
+    propagate_opts.write_buffer_size = 16 * KiB;
```
PragmaTwice: These code are useless.

src/config/config.cc
```diff
@@ -575,6 +577,40 @@ void Config::initFieldCallback() {
...
+               return s;
+             }
+             s = srv->storage->SetOptionForColumnFamily(kColumnFamilyIDStream, "write_buffer_size", buff_size);
+             if (!s.IsOK()) {
+               return s;
+             }
+             s = srv->storage->SetOptionForColumnFamily(kColumnFamilyIDSearch, "write_buffer_size", buff_size);
+             if (!s.IsOK()) {
+               return s;
+             }
```
PragmaTwice: Please use a loop to rewrite this part.

kvrocks.conf
```diff
@@ -329,6 +329,14 @@ json-max-nesting-depth 1024
 # Default: json
 json-storage-format json
 
+# Whether to reduce infrequent column families memory usage.
+# If enable, the write buffers will keep 1 for infrequently used column families
```
PragmaTwice: Seems it's 16k, not 1.

src/storage/storage.cc
```diff
@@ -337,15 +344,22 @@ Status Storage::Open(DBOpenMode mode) {
   propagate_opts.disable_auto_compactions = config_->rocks_db.disable_auto_compactions;
   SetBlobDB(&propagate_opts);
 
+  rocksdb::ColumnFamilyOptions lite_opts(subkey_opts);
+  if (config_->reduce_infrequent_column_memory_usage) {
```
PragmaTwice: Here the `if` is useless if you delete the assignment for pubsub_opts and propagate_opts.

PragmaTwice:
> > I believe "lite mode" is not a suitable name as it does not clearly convey its purpose.
> > We should find a more specific name instead of labeling any random behavior as "lite mode".
> 
> Maybe we can rename this option to `reduce_infrequently_column_memory_usage` so that explicitly tell the user that it is used to reduce the memory size of an infrequently used column family?

It can be `minor_columns_write_buffer_size num`.

src/config/config.cc
jjz921024: update

kvrocks.conf
jjz921024: update

src/storage/storage.cc
jjz921024: update, please review

git-hulk:
@jjz921024 I'm good with this implementation.

src/config/config.cc
```diff
@@ -189,6 +190,7 @@ Config::Config() {
       {"json-max-nesting-depth", false, new IntField(&json_max_nesting_depth, 1024, 0, INT_MAX)},
       {"json-storage-format", false,
        new EnumField<JsonStorageFormat>(&json_storage_format, json_storage_formats, JsonStorageFormat::JSON)},
+      {"minor-columns-write-buffer-size", false, new IntField(&minor_columns_write_buffer_size, 16, 0, 4194304)},
```
mapleFU: inconsistent buffer-size?

src/config/config.h
```diff
@@ -119,6 +119,7 @@ struct Config {
   bool auto_resize_block_and_sst = true;
   int fullsync_recv_file_delay = 0;
   bool use_rsid_psync = false;
+  int minor_columns_write_buffer_size;
```
mapleFU, line 122: Better a default value?

mapleFU (commented):
Looks nice now!

src/config/config.cc
jjz921024: Do you meant units？because the minimum of write_buffer_size is 16KiB，so the unit of this option is KiB.

src/config/config.cc
```diff
@@ -189,6 +190,7 @@ Config::Config() {
       {"json-max-nesting-depth", false, new IntField(&json_max_nesting_depth, 1024, 0, INT_MAX)},
       {"json-storage-format", false,
        new EnumField<JsonStorageFormat>(&json_storage_format, json_storage_formats, JsonStorageFormat::JSON)},
+      {"minor-columns-write-buffer-size", false, new IntField(&minor_columns_write_buffer_size, 65536, 16, 4194304)},
```
mapleFU: Besides, would this conflict with `rocksdb.write_buffer_size`? Should we denote that?

src/config/config.cc
```diff
@@ -575,6 +577,19 @@ void Config::initFieldCallback() {
              if (!srv) return Status::OK();
              return srv->GetNamespace()->LoadAndRewrite();
            }},
+          {"minor-columns-write-buffer-size",
+           [this](Server *srv, const std::string &k, const std::string &v) -> Status {
+             if (!srv) return Status::OK();
+             const std::vector<ColumnFamilyID> column_families = {kColumnFamilyIDZSetScore, kColumnFamilyIDPubSub,
```
mapleFU, line 583: Should we limit the CF set by `"rocksdb.write_buffer_size"`?

src/config/config.cc
```diff
@@ -575,6 +577,19 @@ void Config::initFieldCallback() {
...
+                                                                  kColumnFamilyIDPropagate, kColumnFamilyIDStream,
+                                                                  kColumnFamilyIDSearch};
+             for (const auto &cf : column_families) {
+               auto s = srv->storage->SetOptionForColumnFamily(cf, "write_buffer_size",
+                                                               std::to_string(minor_columns_write_buffer_size * KiB));
```
mapleFU: Should this be an argument of `rocks_db`?

kvrocks.conf
```diff
@@ -329,6 +329,13 @@ json-max-nesting-depth 1024
 # Default: json
 json-storage-format json
 
+# The write buffer size of minor column families those infrequently be used
+# including pubsub, propagte, zset_score, stream and search. 
+# It will reduce the memory usage in some scenarios.
```
mapleFU, line 334: I didn't check the communication previously, why this uses KB but major uses MB?

mapleFU (commented):
I'm so sorry that I was so busy these days. I've checked this patch. Should we add a tag for kvrocks owned Column Familys, and mark some CF as "minor"?

This patch looks ok, however, when adjust the `rocksdb.write_buffer_size`, the buffer size for minor would be overwrite. Besides, our code enforce "write_buffer_size" in some scenerio

I'm also OK with this as a first patch here.

kvrocks.conf
git-hulk: > We can keep the write_buffer_size as a relatively smaller one like 16MiB/32MiB instead of 1 byte.

Yes, it will be better to use MiB to keep consistent with the option `write_buffer_size`. Rest are good to me.

src/storage/storage.cc
```diff
@@ -337,15 +344,20 @@ Status Storage::Open(DBOpenMode mode) {
   propagate_opts.disable_auto_compactions = config_->rocks_db.disable_auto_compactions;
   SetBlobDB(&propagate_opts);
 
+  rocksdb::ColumnFamilyOptions minor_opts(subkey_opts);
+  minor_opts.write_buffer_size = config_->minor_columns_write_buffer_size * KiB;
+  pubsub_opts.write_buffer_size = config_->minor_columns_write_buffer_size * KiB;
+  propagate_opts.write_buffer_size = config_->minor_columns_write_buffer_size * KiB;
```
PragmaTwice: Could you explain for these two lines of code?

src/config/config.cc
PragmaTwice: Good suggestion.

src/config/config.cc
jjz921024: Yes, If the user just want to adjust write_buffer_size for `subkey` and keep other minor cf.  The user had to execute 2 `config` command. 
1. first exec `config set rocksdb.write_buffer_size num`  to adjust all cf 
2. and exec `config set minor-columns-write-buffer-size num`  for recover minor cf 
we should denote it in comment

@git-hulk @PragmaTwice what do your think?

src/config/config.cc
jjz921024: Do you mean prefix this option with rocksdb，like `rocksdb.minor-columns-write-buffer-size` ?

I'm struggling with this. Because all options prefix with `rocksdb` correspond to the options of rocksdb, but this is a behavior of kvrocks.

src/storage/storage.cc
jjz921024: Don't we want to specify `write_buffer_size` of all minor column families except for metadata and subkey?

src/config/config.cc
mapleFU: > Because all options prefix with rocksdb correspond to the options of rocksdb, but this is a behavior of kvrocks.

`rocksdb.write_buffer_size` previously is also a part of rocksdb and written inside `config.rocks_db`, maybe we should regard them as same thing?

src/storage/storage.cc
PragmaTwice: What will happen if `minor_columns_write_buffer_size` is disabled?

kvrocks.conf
```diff
@@ -329,6 +329,13 @@ json-max-nesting-depth 1024
...
+#
+# Default: 64 MB (same with rocksdb.write_buffer_size)
+minor-columns-write-buffer-size 64
```
PragmaTwice, line 337: I think by default it can be 0, which means its value as same as `rocksdb.write_buffer_size`.

mapleFU:
@jjz921024 Maybe a simpler way is:
1. Config an extra "rocksdb.minor_write_buffer_size" as extra arguments, it can be a bit smaller
2. "default" and "metadata" cf is "major", other column families are "minor"
3. Adjust the size for these two parts. There are two kinds of ways:  (1) using same default "rocksdb.minor_write_buffer_size", or even a smaller "rocksdb.minor_write_buffer_size", and separate the two args. This would nice for new user and most users, since it could reducing the arg here and just improve some performance (2) Using current way, this would be good for some extra user who adjusted the write_buffer_size previously

Personally (3) is a bit hard but I think these CF should be tuned themselves in the future, maintaining two configs both for "minor"( like if "rocksdb.write_buffer_size" is set and "rocksdb.minor_write_buffer_size"  doesn't set, using "rocksdb.write_buffer_size", and otherwise separate them) is better for users, but might be a bit tricky. Maybe we can separate it into other patch?

Thanks for the efforts!

PR #2259, opened 2024-04-20
Fix the replica should remove the master link when receiving the CLUSTER RESET command

git-hulk:
This PR also allows using SOFT|HARD in the CLUSTER RESET command,
but it has the same meaning in the Kvrocks cluster since keeping
the cluster's current epoch(version) is meaningless to it.

### DIFF
--- a/tests/gocase/integration/cluster/cluster_test.go
+++ b/tests/gocase/integration/cluster/cluster_test.go
@@ -473,19 +473,37 @@ func TestClusterReset(t *testing.T) {
 	id1 := "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx01"
 	require.NoError(t, rdb1.Do(ctx, "clusterx", "SETNODEID", id1).Err())
 
+	srv2 := util.StartServer(t, map[string]string{"cluster-enabled": "yes"})
+	defer func() { srv2.Close() }()
+	rdb2 := srv2.NewClientWithOption(&redis.Options{PoolSize: 1})
+	defer func() { require.NoError(t, rdb2.Close()) }()
+	id2 := "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx02"
+	require.NoError(t, rdb1.Do(ctx, "clusterx", "SETNODEID", id1).Err())
+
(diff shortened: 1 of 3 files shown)


tests/gocase/integration/cluster/cluster_test.go
```diff
@@ -473,33 +473,51 @@ func TestClusterReset(t *testing.T) {
...
+	require.NoError(t, rdb2.Do(ctx, "clusterx", "SETNODES", clusterNodes, "1").Err())
 
 	t.Run("cannot reset cluster if the db is not empty", func(t *testing.T) {
 		key := util.SlotTable[0]
 		require.NoError(t, rdb0.Set(ctx, key, "value", 0).Err())
-		require.Contains(t, rdb0.Do(ctx, "cluster", "reset").Err(), "Can't reset cluster while database is not empty")
+		require.Contains(t, rdb0.ClusterResetHard(ctx).Err(), "Can't reset cluster while database is not empty")
 		require.NoError(t, rdb0.Del(ctx, key).Err())
-		require.NoError(t, rdb0.Do(ctx, "cluster", "reset").Err())
+		require.NoError(t, rdb0.ClusterResetSoft(ctx).Err())
```
git-hulk, line 495: Use to confirm the `CLUSTER RESET SOFT` also works well.

src/storage/rdb.cc
```diff
@@ -125,7 +125,7 @@ StatusOr<uint64_t> RDB::loadObjectLen(bool *is_encoded) {
         GET_OR_RET(stream_->Read(reinterpret_cast<char *>(&len), sizeof(uint64_t)));
         return ntohu64(len);
       } else {
-        return {Status::NotOK, fmt::format("Unknown RDB string encoding type {} byte {}", type, c)};
+        return {Status::NotOK, fmt::format("Unknown RDB co -string encoding type {} byte {}", type, c)};
```
PragmaTwice: Is there any special meaning of `co -`?

git-hulk: No, I wrongly typed the keyboard and submit the change.

git-hulk: Fixed now.

PragmaTwice (approved):

PR #2262, opened 2024-04-21
Add support of the SORT command

PokIsemaine:
issues: https://github.com/apache/kvrocks/issues/1952


jihuayu (dismissed):
Good job. Thank you for providing such comprehensive test cases.

src/storage/redis_db.h
```diff
@@ -23,13 +23,33 @@
...
+  bool alpha = false;                    // ALPHA
+  std::string storekey;                  // STORE
+};
+
+struct RedisSortObject {
+  std::string obj;
+  std::variant<double, std::string> v;
+};
+
+bool SortCompare(const RedisSortObject &a, const RedisSortObject &b, const SortArgument &args);
```
mapleFU: What does the bool return value in "SortCompare" means? Would you mind add a doc string or change the function name?

PokIsemaine: > What does the bool return value in "SortCompare" means? Would you mind add a doc string or change the function name?


This function is used to assist in implementing the third parameter `Compare comp` of `std::sort`, so that `RedisSortObject` can be sorted based on `SortArgument`

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,111 @@ class CommandCopy : public Commander {
...
 };
 
+template <bool ReadOnly>
+class CommandSort : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 2);
+    while (parser.Good()) {
+      if (parser.EatEqICase("BY")) {
+        if (parser.Remains() < 1) {
```
mapleFU: Isn't `TakeStr` already checks remaining?

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,111 @@ class CommandCopy : public Commander {
...
+        if (parser.Remains() < 2) {
+          return parser.InvalidSyntax();
+        }
+        sort_argument_.offset = GET_OR_RET(parser.template TakeInt<int>());
+        sort_argument_.count = GET_OR_RET(parser.template TakeInt<int>());
+      } else if (parser.EatEqICase("GET") && parser.Remains() >= 1) {
+        if (parser.Remains() < 1) {
+          return parser.InvalidSyntax();
+        }
+        /* TODO:
```
mapleFU, line 451: Why checks `Remains() < 1` after `Remains >= 1` ?

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,111 @@ class CommandCopy : public Commander {
...
+        sort_argument_.getpatterns.push_back(GET_OR_RET(parser.TakeStr()));
+      } else if (parser.EatEqICase("ASC")) {
+        sort_argument_.desc = false;
+      } else if (parser.EatEqICase("DESC")) {
+        sort_argument_.desc = true;
+      } else if (parser.EatEqICase("ALPHA")) {
+        sort_argument_.alpha = true;
+      } else if (parser.EatEqICase("STORE")) {
+        if constexpr (ReadOnly) {
+          return parser.InvalidSyntax();
```
mapleFU: Can the error code be more specifically?

src/storage/redis_db.cc
```diff
@@ -777,4 +782,208 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+   * even if no sort order is requested, so they remain stable across
+   * scripting and replication.
+   *
+   * TODO: support CLIENT_SCRIPT flag, (!storekey_.empty() || c->flags & CLIENT_SCRIPT)) */
+  if (args.dontsort && type == RedisType::kRedisSet && (!args.storekey.empty())) {
+    /* Force ALPHA sorting */
+    args.dontsort = false;
+    args.alpha = true;
+    args.sortby = "";
+  }
```
mapleFU: Can we normalize this when build the args? Now the `SortArgument` is build outside and normalized in `Database::Sort`. Which is a bit weird

src/storage/redis_db.cc
```diff
@@ -777,4 +782,208 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+
+  // Obtain the length of the object to sort.
+  const std::string ns_key = AppendNamespacePrefix(key);
+  Metadata metadata(type, false);
+  auto s = GetMetadata(GetOptions{}, {type}, ns_key, &metadata);
+  if (!s.ok()) {
+    return s;
+  }
+
+  int vectorlen = (int)metadata.size;
```
mapleFU: Better using `static_cast`

mapleFU (commented):
I'll take a carefully pass tonight, just some minor comments now

src/commands/cmd_key.cc
PragmaTwice: Yeah, we can remove this check.

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,111 @@ class CommandCopy : public Commander {
...
+           * If BY is specified with a real pattern, we can't accept it in cluster mode,
+           * unless we can make sure the keys formed by the pattern are in the same slot
+           * as the key to sort.
+           * If BY is specified with a real pattern, we can't accept
+           * it if no full ACL key access is applied for this command. */
+        }
+      } else if (parser.EatEqICase("LIMIT")) {
+        if (parser.Remains() < 2) {
+          return parser.InvalidSyntax();
+        }
```
PragmaTwice: ditto

PragmaTwice: ditto

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,111 @@ class CommandCopy : public Commander {
...
+        sort_argument_.offset = GET_OR_RET(parser.template TakeInt<int>());
+        sort_argument_.count = GET_OR_RET(parser.template TakeInt<int>());
+      } else if (parser.EatEqICase("GET") && parser.Remains() >= 1) {
+        if (parser.Remains() < 1) {
+          return parser.InvalidSyntax();
+        }
```
PragmaTwice: ditto

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,111 @@ class CommandCopy : public Commander {
...
+        sort_argument_.desc = true;
+      } else if (parser.EatEqICase("ALPHA")) {
+        sort_argument_.alpha = true;
+      } else if (parser.EatEqICase("STORE")) {
+        if constexpr (ReadOnly) {
+          return parser.InvalidSyntax();
+        }
+        if (parser.Remains() < 1) {
+          return parser.InvalidSyntax();
+        }
```
PragmaTwice: ditto

src/storage/redis_db.cc
```diff
@@ -777,4 +782,208 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+    }
+    for (const std::string &pattern : args.getpatterns) {
+      std::string val = lookupKeyByPattern(pattern, elem.obj);
+      if (val.empty()) {
+        output_vec->emplace_back(redis::NilString(version));
+      } else {
+        output_vec->emplace_back(val);
+      }
+    }
+  }
```
PragmaTwice: I don't think it's a good idea to generate output in `Database::Sort`, instead it should be in `Execute`.

`Database::Sort` can just focus on the algorithm and rocksdb operations.

src/storage/redis_db.h
```diff
@@ -23,13 +23,44 @@
...
+/// It can assist in implementing the third parameter `Compare comp` required by `std::sort`
+///
+/// \param args The basis used to compare two RedisSortObjects.
+/// If `args.alpha` is false, `RedisSortObject.v` will be taken as double for comparison
+/// If `args.alpha` is true and `args.sortby` is not empty, `RedisSortObject.v` will be taken as string for comparison
+/// If `args.alpha` is true and `args.sortby` is empty, the comparison is by `RedisSortObject.obj`.
+///
+/// \return If `desc` is false, returns true when `a < b`, otherwise returns true when `a > b`
+
+bool SortCompare(const RedisSortObject &a, const RedisSortObject &b, const SortArgument &args);
```
PragmaTwice: I think this function can be a static (or friend) member function of `RedisSortObject`.

mapleFU: ( Better being a member or std::function to accelerate )

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,111 @@ class CommandCopy : public Commander {
...
+
+        if (sort_argument_.sortby.find('*') == std::string::npos) {
+          sort_argument_.dontsort = true;
+        } else {
+          /* TODO:
+           * If BY is specified with a real pattern, we can't accept it in cluster mode,
+           * unless we can make sure the keys formed by the pattern are in the same slot
+           * as the key to sort.
+           * If BY is specified with a real pattern, we can't accept
+           * it if no full ACL key access is applied for this command. */
```
mapleFU, line 445: Should we raise un-support first?

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,111 @@ class CommandCopy : public Commander {
...
+  }
+
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    redis::Database redis(srv->storage, conn->GetNamespace());
+    RedisType type = kRedisNone;
+    auto s = redis.Type(args_[1], &type);
+    if (s.ok()) {
+      if (type >= RedisTypeNames.size()) {
+        return {Status::RedisExecErr, "Invalid type"};
+      } else if (type != RedisType::kRedisList && type != RedisType::kRedisSet && type != RedisType::kRedisZSet) {
```
mapleFU: Would it enough for just check the type equal to List/Set/Zset?

src/storage/redis_db.cc
```diff
@@ -777,4 +782,208 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+        }
+      } else {
+        list_db.Range(key, 0, -1, &str_vec);
+      }
+    } else if (type == RedisType::kRedisSet) {
+      auto set_db = redis::Set(storage_, namespace_);
+      set_db.Members(key, &str_vec);
+
+      if (args.dontsort) {
+        str_vec = std::vector(str_vec.begin() + offset, str_vec.begin() + offset + count);
```
mapleFU: why desc isn't checked here?

src/storage/redis_db.cc
```diff
@@ -777,4 +782,208 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+        zset_db.RangeByRank(key, spec, &member_scores, nullptr);
+
+        for (auto &member_score : member_scores) {
+          str_vec.emplace_back(member_score.member);
+        }
+      } else {
+        zset_db.GetAllMemberScores(key, &member_scores);
+
+        for (auto &member_score : member_scores) {
+          str_vec.emplace_back(member_score.member);
```
mapleFU: nit: can using move to handling this

src/storage/redis_db.cc
```diff
@@ -777,4 +782,208 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+
+      if (args.dontsort) {
+        RangeRankSpec spec;
+        spec.start = offset;
+        spec.stop = offset + count - 1;
+        spec.reversed = args.desc;
+        zset_db.RangeByRank(key, spec, &member_scores, nullptr);
+
+        for (auto &member_score : member_scores) {
+          str_vec.emplace_back(member_score.member);
```
mapleFU: nit: can use std::move to handling this

src/storage/redis_db.cc
mapleFU: Besides, should we limit the element count here? @PragmaTwice . A huge vectorlen can easily causing OOM in our current impl?

src/storage/redis_db.cc
```diff
@@ -777,4 +782,208 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+        byval = str_vec[i];
+      }
+
+      if (args.alpha) {
+        if (!args.sortby.empty()) {
+          sort_vec[i].v = byval;
+        }
+      } else {
+        try {
+          sort_vec[i].v = std::stod(byval);
```
mapleFU: Seems that we already has parsing in util?

src/storage/redis_db.cc
```diff
@@ -777,4 +782,208 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
   return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
 }
 
+std::string Database::lookupKeyByPattern(const std::string &pattern, const std::string &subst) {
```
mapleFU: Should better has a doc string here?

src/commands/cmd_key.cc
PokIsemaine: SORT only supports sorting Keys of List/Set/ZSet type

ref:
* https://redis.io/docs/latest/commands/sort/
* https://github.com/redis/redis/blob/804110a487f048669aa9d9412e5789ec43f4fe39/src/sort.c#L271

src/commands/cmd_key.cc
PokIsemaine: Sorry, this was written by mistake

src/storage/redis_db.cc
PokIsemaine: Agree

src/commands/cmd_key.cc
PokIsemaine: How to propose it specifically? This part actually has some checks.
https://github.com/redis/redis/blob/804110a487f048669aa9d9412e5789ec43f4fe39/src/sort.c#L218


Should we remind users in the code or in the kvrocks documentation?

src/storage/redis_db.cc
PokIsemaine: Can you please indicate the specific location? Thanks

src/storage/redis_db.cc
PokIsemaine: How about moving it after the type check in the Execute function?
```
Status Execute(Server *srv, Connection *conn, std::string *output) override {
  redis::Database redis(srv->storage, conn->GetNamespace());
  RedisType type = kRedisNone;
  auto s = redis.Type(args_[1], &type);
  if (s.ok()) {
    if (type >= RedisTypeNames.size()) {
      return {Status::RedisExecErr, "Invalid type"};
    } else if (type != RedisType::kRedisList && type != RedisType::kRedisSet && type != RedisType::kRedisZSet) {
      *output = Error("WRONGTYPE Operation against a key holding the wrong kind of value");
      return Status::OK();
    }
  } else {
  }
  // MOVE TO HERE
  std::vector<std::string> output_vec;
```

src/storage/redis_db.cc
PokIsemaine: ASC/DESC is not valid for `dontsort` Set key.
ref: https://github.com/redis/redis/blob/804110a487f048669aa9d9412e5789ec43f4fe39/src/sort.c#L383

In Redis:
```
127.0.0.1:6379> sadd skey 2 3 4 1
(integer) 4
127.0.0.1:6379> SMEMBERS skey
1) "1"
2) "2"
3) "3"
4) "4"
127.0.0.1:6379> sort skey by not-exist-key ASC
1) "1"
2) "2"
3) "3"
4) "4"
127.0.0.1:6379> sort skey by not-exist-key DESC
1) "1"
2) "2"
3) "3"
4) "4"
```

src/storage/redis_db.cc
PokIsemaine: Like this?
```
for (auto &member_score : member_scores) {
    str_vec.emplace_back(std::move(member_score.member));
}
```

src/storage/redis_db.cc
PokIsemaine: Do subsequent store operations also need to be placed in `Execute`? It relies on `output_vec` while operating on rocksdb

```
  Perform storage
  if (!args.storekey.empty()) {
    redis::List list_db(storage_, namespace_);
    std::vector<Slice> elems(output_vec->begin(), output_vec->end());
    list_db.Trim(args.storekey, 0, -1);
    uint64_t new_size = 0;
    list_db.Push(args.storekey, elems, false, &new_size);
  }
```
Maybe we can replace a variable name instead of using `output_vec`. What do you think is better?

src/storage/redis_db.cc
mapleFU: I prefer this one

src/commands/cmd_key.cc
mapleFU: Yes, would we just check the "invalid type" here?

`type != RedisType::kRedisList && type != RedisType::kRedisSet && type != RedisType::kRedisZSet`

src/storage/redis_db.cc
mapleFU: I mean, seems "set" not check the asc/desc?

src/storage/redis_db.cc
mapleFU: Yes

src/storage/redis_db.cc
mapleFU: Oh, all values are loaded here..

src/storage/redis_db.cc
PragmaTwice: > Do subsequent store operations also need to be placed in Execute?

No. As I said, rocksdb read/write should be in `redis_db.cc`.

src/storage/redis_db.cc
PragmaTwice: here: https://github.com/apache/kvrocks/blob/unstable/src/common/parse_util.h

PragmaTwice:
> Besides, should we limit the element count here? @PragmaTwice . A huge vectorlen can easily causing OOM in our current impl?

Actually I'm thinking to specify a special command tag for such commands that loads all of the data structure, like this one, and HRANDFIELD..

src/commands/cmd_key.cc
PokIsemaine: Yes, do you think there is anything that needs to be improved?

src/storage/redis_db.cc
PokIsemaine: This seems unfeasible. The code in the output part **is also necessary** code in the store part and cannot be moved out separately.
At present, I have renamed `output_vec` -> `elems`, and removed the comment of output, so that semantically it can be considered that it is just a store operation.
https://github.com/apache/kvrocks/pull/2262/commits/6b8880db9f350f5a351eec49061ba2083d6c6730#diff-df4860055937f1bb78946c49a63a6c4120812a1ec6229a5ded3aa3536220c253

src/storage/redis_db.cc
```diff
@@ -777,4 +782,193 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+
+  // Perform storage
+  for (auto &elem : sort_vec) {
+    if (args.getpatterns.empty()) {
+      elems->emplace_back(elem.obj);
+    }
+    for (const std::string &pattern : args.getpatterns) {
+      std::string val = lookupKeyByPattern(pattern, elem.obj);
+      if (val.empty()) {
+        elems->emplace_back(redis::NilString(version));
```
PragmaTwice: Seems here the implementation is wrong. We cannot store `redis::NilString(version)` into database. Instead we need to store an empty string.

PragmaTwice: And then, this function will be unrelated to RESP version.

src/storage/redis_db.cc
```diff
@@ -937,7 +937,7 @@ rocksdb::Status Database::Sort(const RedisType &type, const std::string &key, co
     for (const std::string &pattern : args.getpatterns) {
       std::string val = lookupKeyByPattern(pattern, elem.obj);
       if (val.empty()) {
-        elems->emplace_back(redis::NilString(version));
+        elems->emplace_back("");
```
PragmaTwice: Hmmm seems we need to distinguish between nil and empty string here?

If so we need to use something like `vector<optional<string>>`.

PokIsemaine: I agree with you, this also requires modifying `lookupKeyByPattern` and adding relevant tests

src/commands/cmd_key.cc
PokIsemaine: I initially referred to the implementation of CmdType, but there were unnecessary checks. This code would be changed to the following:
```
    if (s.ok()) {
      if (type != RedisType::kRedisList && type != RedisType::kRedisSet && type != RedisType::kRedisZSet) {
        *output = Error("WRONGTYPE Operation against a key holding the wrong kind of value");
        return Status::OK();
      }
    } else {
    }
```

src/commands/cmd_key.cc
mapleFU: Maybe return a InvalidArg in Parse phase is ok. Report error is better than un-supported behavior?

src/commands/cmd_key.cc
mapleFU:
```
if (!s.ok()) {
  return {...};
}
// check type
```

I think return error eariler is ok, and we'd better not making indent too deep

src/commands/cmd_key.cc
PokIsemaine: Is there any way to programmatically warn the user without directly aborting the execution of the command? If an error is reported directly, it basically means that this parameter cannot be used normally, but what we lack is just a check under specific circumstances.

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,115 @@ class CommandCopy : public Commander {
...
+        sort_argument_.getpatterns.push_back(GET_OR_RET(parser.TakeStr()));
+      } else if (parser.EatEqICase("ASC")) {
+        sort_argument_.desc = false;
+      } else if (parser.EatEqICase("DESC")) {
+        sort_argument_.desc = true;
+      } else if (parser.EatEqICase("ALPHA")) {
+        sort_argument_.alpha = true;
+      } else if (parser.EatEqICase("STORE")) {
+        if constexpr (ReadOnly) {
+          return {Status::RedisParseErr, "SORT_RO is read-only and does not support the STORE parameter"};
```
mapleFU, line 464: InvalidArg?

src/storage/redis_db.cc
```diff
@@ -777,4 +782,195 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+        }
+      } else {
+        list_db.Range(key, 0, -1, &str_vec);
+      }
+    } else if (type == RedisType::kRedisSet) {
+      auto set_db = redis::Set(storage_, namespace_);
+      set_db.Members(key, &str_vec);
+
+      if (args.dontsort) {
+        str_vec = std::vector(str_vec.begin() + offset, str_vec.begin() + offset + count);
```
mapleFU: Would move iterator better here?

mapleFU (commented):
Sort is too complex here, I'll dive into it later, thanks for the contribution

src/commands/cmd_key.cc
PragmaTwice: It's fine here.

PragmaTwice (commented):
The implementation looks generally good to me.

See if @mapleFU or others have any concern.

src/storage/redis_db.cc
PokIsemaine: Sorry, I don't quite understand. Do you mean using `std::move` or something like `std::advance`? Can you paste the desired code directly? Or you can make changes directly. Thank you so much!

PragmaTwice: @mapleFU means you can use [move_iterator](https://en.cppreference.com/w/cpp/iterator/move_iterator) here to avoid string copy.

src/storage/redis_db.cc
```diff
@@ -951,7 +953,10 @@ rocksdb::Status Database::Sort(const RedisType &type, const std::string &key, co
     redis::List list_db(storage_, namespace_);
     list_db.Trim(args.storekey, -1, 0);
     uint64_t new_size = 0;
-    list_db.Push(args.storekey, std::vector<Slice>(store_elems.cbegin(), store_elems.cend()), false, &new_size);
+    list_db.Push(
+        args.storekey,
+        std::vector<Slice>(std::make_move_iterator(store_elems.cbegin()), std::make_move_iterator(store_elems.cend())),
```
PragmaTwice: The value type is Slice here, so we don't need to use move iterator.

src/storage/redis_db.h
```diff
@@ -21,15 +21,46 @@
...
+  std::variant<double, std::string> v;
+
+  /// SortCompare is a helper function that enables `RedisSortObject` to be sorted based on `SortArgument`.
+  ///
+  /// It can assist in implementing the third parameter `Compare comp` required by `std::sort`
+  ///
+  /// \param args The basis used to compare two RedisSortObjects.
+  /// If `args.alpha` is false, `RedisSortObject.v` will be taken as double for comparison
+  /// If `args.alpha` is true and `args.sortby` is not empty, `RedisSortObject.v` will be taken as string for comparison
+  /// If `args.alpha` is true and `args.sortby` is empty, the comparison is by `RedisSortObject.obj`.
```
mapleFU, line 64: So why do here has `std::variant` and a `std::string`? We should use only `variant` or `double + string` composite?

src/storage/redis_db.h
```diff
@@ -119,6 +153,21 @@ class Database {
   // Already internal keys
   [[nodiscard]] rocksdb::Status existsInternal(const std::vector<std::string> &keys, int *ret);
   [[nodiscard]] rocksdb::Status typeInternal(const Slice &key, RedisType *type);
+
+  /// lookupKeyByPattern is a helper function of `Sort` to support `GET` and `BY` fields.
+  ///
+  /// \param pattern can be the value of a `BY` or `GET` field
+  /// \param subst is used to replace the "*" or "#" matched in the pattern string.
+  /// \return  Return the value associated to the key with a name obtained using the following rules:
```
mapleFU:
```suggestion
  /// \return  Returns the value associated to the key with a name obtained using the following rules:
```

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,115 @@ class CommandCopy : public Commander {
...
 };
 
+template <bool ReadOnly>
+class CommandSort : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 2);
+    while (parser.Good()) {
+      if (parser.EatEqICase("BY")) {
+        sort_argument_.sortby = GET_OR_RET(parser.TakeStr());
```
mapleFU, line 435: Should we check `sortby.empty` here?

src/commands/cmd_key.cc
```diff
@@ -424,6 +424,115 @@ class CommandCopy : public Commander {
...
+    }
+
+    /* When sorting a set with no sort specified, we must sort the output
+     * so the result is consistent across scripting and replication.
+     *
+     * The other types (list, sorted set) will retain their native order
+     * even if no sort order is requested, so they remain stable across
+     * scripting and replication.
+     *
+     * TODO: support CLIENT_SCRIPT flag, (!storekey_.empty() || c->flags & CLIENT_SCRIPT)) */
```
mapleFU, line 494: @git-hulk do we already have flag like this?

src/storage/redis_db.cc
```diff
@@ -768,4 +773,197 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+    RedisType type = RedisType::kRedisNone;
+    if (auto s = string_db.Type(key, &type); !s.ok() || type != RedisType::kRedisString) {
+      return std::nullopt;
+    }
+    string_db.Get(key, &value);
+  }
+  return value;
+}
+
+rocksdb::Status Database::Sort(const RedisType &type, const std::string &key, const SortArgument &args,
```
mapleFU: `RedisType` better passing by value?

src/storage/redis_db.cc
```diff
@@ -768,4 +773,197 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+  if (!s.ok()) {
+    return s;
+  }
+
+  int vectorlen = static_cast<int>(metadata.size);
+
+  // Adjust the offset and count of the limit
+  int offset = args.offset >= vectorlen ? 0 : std::clamp(args.offset, 0, vectorlen - 1);
+  int count = args.offset >= vectorlen ? 0 : std::clamp(args.count, -1, vectorlen - offset);
+  if (count == -1) count = vectorlen - offset;
```
mapleFU, line 836: `count == -1` is a bit weird, can we make it more strict to `count < 0`?

src/storage/redis_db.cc
```diff
@@ -768,4 +773,197 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+        spec.start = offset;
+        spec.stop = offset + count - 1;
+        spec.reversed = args.desc;
+        zset_db.RangeByRank(key, spec, &member_scores, nullptr);
+
+        for (auto &member_score : member_scores) {
+          str_vec.emplace_back(std::move(member_score.member));
+        }
+      } else {
+        zset_db.GetAllMemberScores(key, &member_scores);
```
mapleFU: Should we check the return value of these helper function? If get failed, we should return here rather than go to the other logic

src/storage/redis_db.cc
```diff
@@ -768,4 +773,197 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+                               std::vector<std::optional<std::string>> *elems, SortResult *res) {
+  // Obtain the length of the object to sort.
+  const std::string ns_key = AppendNamespacePrefix(key);
+  Metadata metadata(type, false);
+  auto s = GetMetadata(GetOptions{}, {type}, ns_key, &metadata);
+  if (!s.ok()) {
+    return s;
+  }
+
+  int vectorlen = static_cast<int>(metadata.size);
```
mapleFU: Nit: we can limit the vectorlen. Currently kvrocks don't have memory limit logic, so a large length can easily crack the system? We can first limit the length to 512, and enlarge the size when we have better machanism

src/storage/redis_db.cc
```diff
@@ -768,4 +773,197 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+        byval = str_vec[i];
+      }
+
+      if (args.alpha && !args.sortby.empty()) {
+        sort_vec[i].v = byval.value();
+      } else if (!args.alpha && !byval.value().empty()) {
+        auto double_byval = ParseFloat<double>(byval.value());
+        if (!double_byval) {
+          *res = SortResult::DOUBLE_CONVERT_ERROR;
+          return rocksdb::Status::OK();
```
mapleFU, line 918: I think it can be ok, but we can denote the rocksdb return value here

src/storage/redis_db.cc
```diff
@@ -768,4 +773,197 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+
+  std::vector<RedisSortObject> sort_vec(str_vec.size());
+  for (size_t i = 0; i < str_vec.size(); ++i) {
+    sort_vec[i].obj = str_vec[i];
+  }
+
+  // Sort by BY, ALPHA, ASC/DESC
+  if (!args.dontsort) {
+    for (size_t i = 0; i < sort_vec.size(); ++i) {
+      std::optional<std::string> byval;
```
mapleFU: Why it's a `std::optional` here?

```c++
std::string byval;
      if (!args.sortby.empty()) {
        auto lookup = lookupKeyByPattern(args.sortby, str_vec[i]);
        if (!lookup.has_value()) continue;
        byval = std::move(lookup.value());
      } else {
        byval = str_vec[i];
      }
```

Could this be better?

src/commands/cmd_key.cc
git-hulk: @mapleFU Not for now.

src/storage/redis_db.cc
PokIsemaine: Because `std::clamp(args.count, -1, vectorlen - offset);`,  `count` will not be any other negative number

src/storage/redis_db.h
PokIsemaine: I think it is feasible to only use `std::variant<std::string, double>`, but this may require distinguishing between two types of `std::string`: BY(RedisSortObject.v) and the sort key itself(RedisSortObject.obj), resulting in unclear sorting logic.

mapleFU: Sigh, I think maybe struct or other is ok, since the passed in argument ( `RedisSortObject` ) is defined by `args`. It cannot have another state ( like holding a string when `!args.alpha` ), and this introduce a double-checking here.

`std::variant` seems has same space usage as `struct {double; string;}`. But both is ok for me

src/storage/redis_db.cc
PokIsemaine: Please make it clearer or use the code modification suggestions directly, thank you.

src/commands/cmd_key.cc
PokIsemaine: I don't think it's necessary, redis allows receiving `""` as the value of `BY`, but this reminds me that I can add some test cases, thanks

mapleFU: I mean, would "BY a BY b*" sequence making `sortby` has conflict stats (both dontsort and have properties)

```
if (!sort_argument_.sortby.empty()) return InvalidArg();
sort_argument_.sortby = GET_OR_RET(parser.TakeStr());
```

Would this check better here?

src/storage/redis_db.cc
mapleFU:
```
  enum class SortResult { UNKNOWN_TYPE, DOUBLE_CONVERT_ERROR, DONE };
  [[nodiscard]] rocksdb::Status Sort(const RedisType &type, const std::string &key, const SortArgument &args,
                                     std::vector<std::optional<std::string>> *elems, SortResult *res);
```

Generally, kvrocks have 2-styles of errors:
1. Return `rocksdb::Status` for storage
2. Return `Status` for itself

`SortResult` defines a self-defined error handling logic here and return it in `Sort`. So we can add docstring for that like:

```
/// When status is not ok, `res` should not been checked.
/// Otherwise, res represents...
```

src/commands/cmd_key.cc
PokIsemaine: I try to test behavior like this in redis
```
127.0.0.1:6379> lpush uid 1 3 2
(integer) 3
127.0.0.1:6379> mset uid_rank_1 3 uid_rank_2 2 uid_rank_3 1
OK
127.0.0.1:6379> mset uid_level_1 1 uid_level_2 2 uid_level_3 3
OK
127.0.0.1:6379> sort uid by uid_rank_*
1) "3"
2) "2"
3) "1"
127.0.0.1:6379> sort uid by uid_level_*
1) "1"
2) "2"
3) "3"
127.0.0.1:6379> sort uid by uid_level_* by uid_rank_*
1) "3"
2) "2"
3) "1"
127.0.0.1:6379> sort uid by uid_rank_* by uid_level_*
1) "1"
2) "2"
3) "3"
127.0.0.1:6379> sort uid by dontsort by uid_level_*
1) "2"
2) "3"
3) "1"
127.0.0.1:6379> sort uid by dontsort by uid_rank_*
1) "2"
2) "3"
3) "1"
127.0.0.1:6379> sort uid by uid_level_* by dontsort
1) "2"
2) "3"
3) "1"
127.0.0.1:6379> sort uid by uid_rank_* by dontsort
1) "2"
2) "3"
3) "1"
```
It seems that when multiple BY appears, the next value will prevail. If dontsort is included, the output will be according to dontsort. 
Using multiple BYs does not seem like normal behavior, what do you think would be better? Align with redis or check?

mapleFU: I think we can add a check and return invalid, following the redis would be a bit weird

src/storage/redis_db.cc
```diff
@@ -768,4 +773,211 @@ rocksdb::Status Database::Copy(const std::string &key, const std::string &new_ke
...
+  }
+
+  std::string key = pattern.substr(0, match_pos + 1);
+  key.replace(match_pos, 1, subst);
+
+  std::string value;
+  RedisType type = RedisType::kRedisNone;
+  if (!field.empty()) {
+    auto hash_db = redis::Hash(storage_, namespace_);
+    if (auto s = hash_db.Type(key, &type); !s.ok() || type != RedisType::kRedisHash) {
```
mapleFU, line 800:
```suggestion
    if (auto s = hash_db.Type(key, &type); !s.ok() || type != RedisType::kRedisHash) {
```

Maybe this can return `StatusOr<std::optional<std::string>>`, but current impl is also ok for me

mapleFU (approved):

mapleFU:
Will merge in 1-days if no negative comments

git-hulk (approved):

PR #2273, opened 2024-04-26
Support ASKING command

### LINKED ISSUE #2168: Add support of the command ASKING

jjz921024:
Close #2168 

Support ASKING command，but didn't do anythins. Because the kvrocks seems will serving the imported slot despite the lack of asking flag

https://github.com/apache/kvrocks/blob/ae5784dbe83bce07f0e5534f84404e7e075daaa4/src/cluster/cluster.cc#L857-L862

### DIFF
--- a/src/commands/cmd_cluster.cc
+++ b/src/commands/cmd_cluster.cc
@@ -338,9 +338,19 @@ class CommandReadWrite : public Commander {
   }
 };
 
+class CommandAsking : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    conn->EnableFlag(redis::Connection::kAsking);
+    *output = redis::SimpleString("OK");
+    return Status::OK();
+  }
+};
+
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandCluster>("cluster", -2, "cluster no-script", 0, 0, 0, GenerateClusterFlag),
                         MakeCmdAttr<CommandClusterX>("clusterx", -2, "cluster no-script", 0, 0, 0, GenerateClusterFlag),
                         MakeCmdAttr<CommandReadOnly>("readonly", 1, "cluster no-multi", 0, 0, 0),
-                        MakeCmdAttr<CommandReadWrite>("readwrite", 1, "cluster no-multi", 0, 0, 0), )
+                        MakeCmdAttr<CommandReadWrite>("readwrite", 1, "cluster no-multi", 0, 0, 0),
+                        MakeCmdAttr<CommandAsking>("asking", 1, "cluster", 0, 0, 0), )
 
 }  // namespace redis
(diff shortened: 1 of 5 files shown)


git-hulk:
> Support ASKING command，but didn't do anythins. Because the kvrocks seems will serving the imported slot despite the lack of asking flag

Yes, exactly. But we can fix this behavior after supporting the ASKING command. By the way, guess we should clear the ASKING flag after executing the next query?

jjz921024:
> > Support ASKING command，but didn't do anythins. Because the kvrocks seems will serving the imported slot despite the lack of asking flag
> 
> Yes, exactly. But we can fix this behavior after supporting the ASKING command. By the way, guess we should clear the ASKING flag after executing the next query?

Yes, Fixing it will make kvorcks more like with redis in behavior. Let me do it.

jjz921024:
@git-hulk Hi, done it, please review

jjz921024:
I find that the kvrocks doesn't seem to return `ASK` error when query a key in migrating slot ?

tests/gocase/integration/slotimport/slotimport_test.go
```diff
@@ -172,3 +172,45 @@ func TestImportedServer(t *testing.T) {
...
+		require.NoError(t, cli.Type(ctx, slotKey).Err())
+	})
+
+	t.Run("IMPORT - asking flag will be reset after executing", func(t *testing.T) {
+		require.Equal(t, "OK", cli.Do(ctx, "asking").Val())
+		require.NoError(t, cli.Type(ctx, slotKey).Err())
+		util.ErrorRegexp(t, cli.Type(ctx, slotKey).Err(), fmt.Sprintf("MOVED %d.*%d.*", slotNum, mockSrv0Port))
+	})
+
+}
```
git-hulk:
```suggestion
}

```

git-hulk:
@jjz921024 Thank you! Generally looks good to me, you can fix the Go format to pass the CI.

jjz921024:
> @jjz921024 Thank you! Generally looks good to me, you can fix the Go format to pass the CI.

Yes, I had fix it

git-hulk (approved):

src/commands/cmd_cluster.cc
```diff
@@ -338,9 +338,19 @@ class CommandReadWrite : public Commander {
   }
 };
 
+class CommandAsking : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    conn->EnableFlag(redis::Connection::kAsking);
+    *output = redis::SimpleString("OK");
```
caipengbo, line 345: LGTM, it would be nice to write a simple comment that says we're doing nothing but compatibility.

caipengbo (approved):

src/commands/cmd_cluster.cc
git-hulk: It should align with the Redis behavior for now, since it only allows the ASKING flag connection to query keys of the importing slot.

caipengbo: Roger that.

PR #2277, opened 2024-04-27
Use the QUICKLIST encoding to dump the list payload to compatible with Redis >= 4

### LINKED ISSUE #2253: DUMP command error with list
### Search before asking

### Version

OS: Debian 11 bulleye
KV Rocks: unstable, git_sha1:4ecb35b9
Redis: 6.2.10

### Minimal reproduce step

Redis service: localhost:6381
KV Rocks: localhost:7381

1. Start Redis 6.2 with attached dump.rdb
[dump.zip](https://github.com/apache/kvrocks/files/14998662/dump.zip)

2. Dump Key "ABCD" , which is a list, from Redis to a local file
```
redis-cli -p 6381 --raw dump ABCD | head -c-1 > ABCD-REDIS.dump
```
3. Restore ABCD to KV rocks
```
cat ABCD-REDIS.dump | redis-cli -p 7381 -x restore ABCD 0
OK
```
4. Dump Key "ABCD" from KV Rocks to local file
```
redis-cli -p 7381 --raw dump ABCD | head -c-1 > ABCD-KVROCKS.dump
```
5. Restore ABCD back to Redis with a new name
```
cat ABCD-KVROCKS.dump | redis-cli -p 6381 -x restore NEW_ABCD 0
(error) ERR Bad data format
```

Restore to Redis with KV Rocks dump should be successful

### What did you see instead?

Restore to Redis with KV Rocks dump is failed. Error message:
(error) ERR Bad data format

### Anything Else?

I attached 2 dump files. KV Rocks dump size is way bigger than the Redis one
[ABCD_dump.zip](https://github.com/apache/kvrocks/files/14998854/ABCD_dump.zip)

AntiTopQuark:
Close #2253

In order to support the compatibility of old versions of redis, 

when using the dump command, `QUICKLIST`encoding is used instead of `QUICKLIST2`for LIST encoding.


src/storage/rdb.cc
```diff
@@ -1005,3 +1000,37 @@ Status RDB::rdbSaveBinaryDoubleValue(double val) {
...
   return stream_->Write((const char *)(&val), sizeof(val));
 }
+
+Status RDB::rdbSaveZipListObject(const std::string &elem) {
+  // calc total ziplist size
+  uint prevlen = 0;
+  const size_t ziplist_size = zlHeaderSize + zlEndSize + elem.length() +
+                              ZipList::zipStorePrevEntryLength(nullptr, prevlen) +
+                              ZipList::zipStoreEntryEncoding(nullptr, elem.length());
+  auto zl = new unsigned char[ziplist_size];
```
mapleFU: why new explicitly here?

AntiTopQuark: I have changed the code from using `new` directly to using `std::make_unique` for memory management. 
Thank you for your review and suggestions!

PragmaTwice:
Could you write some test cases for your patch? Thank you!

AntiTopQuark:
> Could you write some test cases for your patch? Thank you!

Here I just changed a coding method. The previous tests can still be used. I have now added some corner cases.

PragmaTwice:
> > Could you write some test cases for your patch? Thank you!
> 
> Here I just changed a coding method. The previous tests can still be used. I have now added some corner cases.

Generally, if we fixed a bug in a patch, we need to add a test case, that should fail without the patch, and succeed after the patch is applied.

src/storage/rdb_ziplist.h
```diff
@@ -41,4 +53,4 @@ class ZipList {
   Status peekOK(size_t n);
   void setPreEntryLen(uint32_t len) { pre_entry_len_ = len; }
   static uint32_t getEncodedLengthSize(uint32_t len);
-};
+};
```
git-hulk:
```suggestion
};

```

Keep the newline at the end of file.

src/storage/rdb_ziplist.cc
```diff
@@ -152,3 +148,65 @@ Status ZipList::peekOK(size_t n) {
...
+    return (len < ZipListBigLen) ? 1 : sizeof(uint32_t) + 1;
+  } else {
+    if (len < ZipListBigLen) {
+      p[0] = len;
+      return 1;
+    } else {
+      return ZipStorePrevEntryLengthLarge(p, len);
+    }
+  }
+}
```
git-hulk:
```suggestion
uint32_t ZipList::ZipStorePrevEntryLength(unsigned char *p, unsigned int len) {
  if (p == nullptr) {
    return (len < ZipListBigLen) ? 1 : sizeof(uint32_t) + 1;
  }
  if (len < ZipListBigLen) {
    p[0] = len;
    return 1;
  }
  return ZipStorePrevEntryLengthLarge(p, len);
}
```

git-hulk:
@AntiTopQuark Thanks for your efforts.

src/storage/rdb.cc
```diff
@@ -1005,3 +1000,35 @@ Status RDB::rdbSaveBinaryDoubleValue(double val) {
...
   return stream_->Write((const char *)(&val), sizeof(val));
 }
+
+Status RDB::rdbSaveZipListObject(const std::string &elem) {
+  // calc total ziplist size
+  uint prevlen = 0;
+  const size_t ziplist_size = zlHeaderSize + zlEndSize + elem.length() +
+                              ZipList::ZipStorePrevEntryLength(nullptr, prevlen) +
+                              ZipList::ZipStoreEntryEncoding(nullptr, elem.length());
+  auto zl = std::make_unique<unsigned char[]>(ziplist_size);
```
mapleFU: would an std::string better here? which also avoid the copying in L:1027

AntiTopQuark: thank you very much, done

src/storage/rdb_ziplist.cc
AntiTopQuark: done

src/storage/rdb_ziplist.h
AntiTopQuark: done

git-hulk (dismissed):

git-hulk:
I can confirm it also works in Redis 4:

```
127.0.0.1:6379> RESTORE list 0 "\x0e\x04\x0e\x0e\x00\x00\x00\n\x00\x00\x00\x01\x00\x00\x01d\xff\x0e\x0e\x00\x00\x00\n\x00\x00\x00\x01\x00\x00\x01c\xff\x0e\x0e\x00\x00\x00\n\x00\x00\x00\x01\x00\x00\x01b\xff\x0e\x0e\x00\x00\x00\n\x00\x00\x00\x01\x00\x00\x01a\xff\x06\x003\x06\xd4\x86\x93\xfbvB"
OK
127.0.0.1:6379> lrange list 0 -1
1) "d"
2) "c"
3) "b"
4) "a"
```

src/storage/rdb.cc
```diff
@@ -1005,3 +1000,35 @@ Status RDB::rdbSaveBinaryDoubleValue(double val) {
...
   return stream_->Write((const char *)(&val), sizeof(val));
 }
+
+Status RDB::rdbSaveZipListObject(const std::string &elem) {
+  // calc total ziplist size
+  uint prevlen = 0;
+  const size_t ziplist_size = zlHeaderSize + zlEndSize + elem.length() +
+                              ZipList::ZipStorePrevEntryLength(nullptr, prevlen) +
+                              ZipList::ZipStoreEntryEncoding(nullptr, elem.length());
+  auto zl_string = std::string(ziplist_size, 0);
```
mapleFU, line 1010: Why the size would be 0? Isn't it same as `ziplist_size`?

src/storage/rdb_ziplist.cc
```diff
@@ -152,3 +148,63 @@ Status ZipList::peekOK(size_t n) {
...
+  memcpy(p, buf, len);
+  return len;
+}
+
+void ZipList::SetZipListBytes(unsigned char *zl, uint32_t value) { (*((uint32_t *)(zl))) = value; }
+void ZipList::SetZipListTailOffset(unsigned char *zl, uint32_t value) {
+  (*((uint32_t *)((zl) + sizeof(uint32_t)))) = value;
+}
+void ZipList::SetZipListLength(unsigned char *zl, uint16_t value) {
+  (*((uint16_t *)((zl) + sizeof(uint32_t) * 2))) = value;
```
mapleFU: Would you mind use memcpy? Since this might be a type puning here

src/storage/rdb_ziplist.cc
```diff
@@ -152,3 +148,63 @@ Status ZipList::peekOK(size_t n) {
...
+  }
+
+  /* Store this length at p. */
+  memcpy(p, buf, len);
+  return len;
+}
+
+void ZipList::SetZipListBytes(unsigned char *zl, uint32_t value) { (*((uint32_t *)(zl))) = value; }
+void ZipList::SetZipListTailOffset(unsigned char *zl, uint32_t value) {
+  (*((uint32_t *)((zl) + sizeof(uint32_t)))) = value;
```
mapleFU: ditto, would you mind use memcpy here?

src/storage/rdb_ziplist.cc
```diff
@@ -152,3 +148,63 @@ Status ZipList::peekOK(size_t n) {
...
+    buf[3] = (rawlen >> 8) & 0xff;
+    buf[4] = rawlen & 0xff;
+  }
+
+  /* Store this length at p. */
+  memcpy(p, buf, len);
+  return len;
+}
+
+void ZipList::SetZipListBytes(unsigned char *zl, uint32_t value) { (*((uint32_t *)(zl))) = value; }
```
mapleFU: Would you mind use memcpy here?

src/storage/rdb.cc
```diff
@@ -1005,3 +1000,35 @@ Status RDB::rdbSaveBinaryDoubleValue(double val) {
...
+  auto zl_ptr = reinterpret_cast<unsigned char *>(&zl_string[0]);
+
+  // set ziplist header
+  ZipList::SetZipListBytes(zl_ptr, intrev32ifbe(ziplist_size));
```
mapleFU: `ziplist_size` is not a uint32, right? Would we check and cast it to u32?

src/storage/rdb_ziplist.cc
```diff
@@ -152,3 +148,63 @@ Status ZipList::peekOK(size_t n) {
...
+    buf[2] = (rawlen >> 16) & 0xff;
+    buf[3] = (rawlen >> 8) & 0xff;
+    buf[4] = rawlen & 0xff;
+  }
+
+  /* Store this length at p. */
+  memcpy(p, buf, len);
+  return len;
+}
+
```
mapleFU, line 204: Besides, would we also applying length check for these set below?

src/storage/rdb_ziplist.cc
```diff
@@ -152,3 +148,63 @@ Status ZipList::peekOK(size_t n) {
...
 
 uint32_t ZipList::getEncodedLengthSize(uint32_t len) { return len < ZipListBigLen ? 1 : 5; }
+
+uint32_t ZipList::ZipStorePrevEntryLengthLarge(unsigned char *p, unsigned int len) {
+  uint32_t u32 = 0;
+  if (p != nullptr) {
+    p[0] = ZipListBigLen;
+    u32 = len;
+    memcpy(p + 1, &u32, sizeof(u32));
+    memrev32ifbe(p + 1);
```
mapleFU, line 161: Just curious why memrev after memcpy? Wouldn't that be faster to rev the register before writing into memory?

mapleFU (commented):
General LGTM. Remaining problems are about code styles

src/storage/rdb_ziplist.cc
AntiTopQuark: done

src/storage/rdb_ziplist.cc
AntiTopQuark: done

src/storage/rdb_ziplist.cc
AntiTopQuark: done

src/storage/rdb.cc
AntiTopQuark: done

src/storage/rdb_ziplist.cc
AntiTopQuark: done

src/storage/rdb_ziplist.cc
AntiTopQuark: Currently, it seems that Redis is implemented in the same way. I suspect it is to avoid the risks associated with operating directly on the original data.

src/storage/rdb.cc
```diff
@@ -1005,3 +1000,36 @@ Status RDB::rdbSaveBinaryDoubleValue(double val) {
...
+  pos += ZipList::ZipStorePrevEntryLength(pos, ziplist_size, prevlen);
+  pos += ZipList::ZipStoreEntryEncoding(pos, ziplist_size, elem.length());
+  assert(pos + elem.length() <= zl_ptr + ziplist_size);
+  memcpy(pos, elem.c_str(), elem.length());
+
+  // set ziplist end
+  ZipList::SetZipListLength(zl_ptr, ziplist_size, 1);
+  zl_ptr[ziplist_size - 1] = zlEnd;
+
+  if (ziplist_size > 0) {
```
mapleFU: I'm reading the logic here, may I ask that when would `ziplist_size == 0`? Since if `ziplist_size == 0`, the code above would hit memory issue? If it cannot be 0, we should remove this branch?

mapleFU:
I didn't dive into logics of RDB carefully, generally this LGTM

src/storage/rdb.cc
AntiTopQuark: oh, you are right. done～

git-hulk:
@AntiTopQuark Need to reformat codes.

git-hulk (approved):

PR #2278, opened 2024-04-27
Add index selection pass for KQIR planning

PragmaTwice:
We add a simple index selection algorithm upon KQIR planning part based on a simple cost model.

Also, we introduce range-v3 since it's essential for modern C++ and is a part of C++20.


PragmaTwice:
Ready for review now : )

git-hulk (approved):

PR #2303, opened 2024-05-08
Fix compiler warnings in KQIR and RDB

raffertyyu:
According to gcc output, fix two compiler warnings.

### DIFF
--- a/src/search/executors/filter_executor.h
+++ b/src/search/executors/filter_executor.h
@@ -100,6 +100,9 @@ struct QueryExprEvaluator {
         return l > r;
       case NumericCompareExpr::GET:
         return l >= r;
+      default:
+        CHECK(false) << "unreachable";
+        __builtin_unreachable();
     }
   }
 };
--- a/src/storage/rdb_ziplist.cc
+++ b/src/storage/rdb_ziplist.cc
@@ -196,7 +196,7 @@ uint32_t ZipList::ZipStoreEntryEncoding(unsigned char *p, size_t zl_size, unsign
     buf[3] = (rawlen >> 8) & 0xff;
     buf[4] = rawlen & 0xff;
   }
-  assert(zl_size >= zlHeaderSize + len);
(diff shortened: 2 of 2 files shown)


src/storage/rdb_ziplist.cc
```diff
@@ -196,7 +196,7 @@ uint32_t ZipList::ZipStoreEntryEncoding(unsigned char *p, size_t zl_size, unsign
     buf[3] = (rawlen >> 8) & 0xff;
     buf[4] = rawlen & 0xff;
   }
-  assert(zl_size >= zlHeaderSize + len);
+  assert(zl_size >= static_cast<size_t>(zlHeaderSize) + len);
```
mapleFU, line 199: Would `DCHECK` better here?

src/search/executors/filter_executor.h
```diff
@@ -100,6 +100,9 @@ struct QueryExprEvaluator {
         return l > r;
       case NumericCompareExpr::GET:
         return l >= r;
+      default:
+        assert(false);
+        return false;
```
mapleFU: Should we elaborate all branches rather than a unreachable "default"

raffertyyu: Actually all branches are covered here, but gcc thought it should have a default branch to cover any unforeseen case.

src/storage/rdb_ziplist.cc
raffertyyu: static_cast is enough here, imo

src/search/executors/filter_executor.h
PragmaTwice:
```suggestion
      default:
        CHECK(false) << "unreachable";
        __builtin_unreachable();
```

PragmaTwice (approved):

git-hulk (approved):

git-hulk:
The CI failure is not related to this PR, will take a look later.

PragmaTwice (approved):

PR #2323, opened 2024-05-20
feat: add support of the command JSON.DEBUG MEMORY

jackyyyyyssss:
Add support of the command JSON.DEBUG MEMORY
https://redis.io/docs/latest/commands/json.debug-memory/


src/commands/cmd_json.cc
```diff
@@ -624,6 +624,37 @@ class CommandJsonMSet : public Commander {
...
 
+class CommandJsonDebug : public Commander {
+ public:
+  Status Execute(Server *svr, Connection *conn, std::string *output) override {
+    redis::Json json(svr->storage, conn->GetNamespace());
+
+    std::string path = "$";
+
+    if (!util::EqualICase(args_[1], "memory")) {
+      return {Status::RedisExecErr, "The number of arguments is more than expected"};
```
mapleFU: why this is number more than expected?

src/types/redis_json.cc
```diff
@@ -626,5 +626,17 @@ std::vector<rocksdb::Status> Json::readMulti(const std::vector<Slice> &ns_keys,
   }
   return statuses;
 }
+rocksdb::Status Json::DebugMemory(const std::string &user_key, const std::string &path, Optionals<uint64_t> *results) {
```
mapleFU: format or adding a space between them?

src/types/json.h
```diff
@@ -217,6 +217,25 @@ struct JsonValue {
...
 
+  StatusOr<Optionals<uint64_t>> StrBytes(std::string_view path) const {
+    Optionals<uint64_t> results;
+    try {
+      jsoncons::jsonpath::json_query(
+          value, path, [&results](const std::string & /*path*/, const jsoncons::json &origin) {
+            if (!origin.empty()) {
+              // This is only a rough calculation of the size of the string size char, not the entire byte occupied by
+              // the object
+              results.emplace_back(origin.as_string().size() * sizeof(std::string::value_type));
```
mapleFU: Could this json_query matches more than 1 element?

mapleFU (commented):
Another problem is that, should we emit the memory size ( And memory size might larger than str size ) or just export the size on disk?

PragmaTwice (changes requested):
There are several issues in this implemetation:
- the result of `as_string().size()` is not equal to the real size, we should at least use `to_string()`
- but actually `to_string` is also wrong, since it maybe not equal to the result of `compact_json_string_encoder`. also it would be totally wrong if CBOR format is enabled.
- there may be multiple location found from one jsonpath, which is not handled in this impl
- if jsonpath is `$`, no json decoding is required

src/types/json.h
jackyyyyyssss: If multiple elements are hit, an array is returned. This is the return value of Redis 7.0 
251ef557533affef376c009aaf4d218

jackyyyyyssss:
> There are several issues in this implemetation:
> 
> * the result of `as_string().size()` is not equal to the real size, we should at least use `to_string()`
> * but actually `to_string` is also wrong, since it maybe not equal to the result of `compact_json_string_encoder`. also it would be totally wrong if CBOR format is enabled.
> * there may be multiple location found from one jsonpath, which is not handled in this impl
> * if jsonpath is `$`, no json decoding is required

Thank you for your help. The modifications have been completed, please help me take a look again

src/types/json.h
```diff
@@ -217,6 +217,42 @@ struct JsonValue {
...
 
+  StatusOr<Optionals<uint64_t>> StrBytes(std::string_view path, JsonStorageFormat format,
+                                         int max_nesting_depth = std::numeric_limits<int>::max()) const {
+    Optionals<uint64_t> results;
+    try {
+      jsoncons::jsonpath::json_query(
+          value, path,
+          [&results, max_nesting_depth, format](const std::string & /*path*/, const jsoncons::json &origin) {
+            if (!origin.empty()) {
+              Status s;
```
PragmaTwice: Is it useful here?

src/types/json.h
```diff
@@ -217,6 +217,42 @@ struct JsonValue {
...
+              if (format == JsonStorageFormat::JSON) {
+                jsoncons::json_options options;
+                options.max_nesting_depth(max_nesting_depth);
+                jsoncons::compact_json_string_encoder encoder{buffer, options};
+                origin.dump(encoder, ec);
+              } else if (format == JsonStorageFormat::CBOR) {
+                jsoncons::cbor::cbor_options options;
+                options.max_nesting_depth(max_nesting_depth);
+                jsoncons::cbor::basic_cbor_encoder<jsoncons::string_sink<std::string>> encoder{buffer, options};
+                origin.dump(encoder, ec);
```
PragmaTwice: Please use the exception version of `dump` and catch it later.

src/types/json.h
```diff
@@ -217,6 +217,42 @@ struct JsonValue {
     return results;
   }
 
+  StatusOr<Optionals<uint64_t>> StrBytes(std::string_view path, JsonStorageFormat format,
```
PragmaTwice: Is `optional` useful here? Could we just use `vector`?

```suggestion
  StatusOr<Optionals<uint64_t>> GetBytes(std::string_view path, JsonStorageFormat format,
```

src/types/redis_json.cc
```diff
@@ -627,4 +627,24 @@ std::vector<rocksdb::Status> Json::readMulti(const std::vector<Slice> &ns_keys,
...
 
+rocksdb::Status Json::DebugMemory(const std::string &user_key, const std::string &path, Optionals<uint64_t> *results) {
+  auto ns_key = AppendNamespacePrefix(user_key);
+  JsonMetadata metadata;
+  if (path == "$") {
+    std::string bytes;
+    Slice rest;
+    auto s = GetMetadata(GetOptions{}, {kRedisJson}, ns_key, &bytes, &metadata, &rest);
+    if (!s.ok()) return s;
+    results->emplace_back(bytes.size());
```
PragmaTwice:
```suggestion
    results->emplace_back(rest.size());
```

src/types/json.h
```diff
@@ -217,6 +217,42 @@ struct JsonValue {
     return results;
   }
 
+  StatusOr<std::vector<std::string>> GetBytes(std::string_view path, JsonStorageFormat format,
```
PragmaTwice: Why `vector<string>` instead of `vector<size_t>`?

src/types/json.h
```diff
@@ -217,6 +217,42 @@ struct JsonValue {
...
+                                              int max_nesting_depth = std::numeric_limits<int>::max()) const {
+    std::vector<std::string> results;
+    try {
+      jsoncons::jsonpath::json_query(value, path, [&](const std::string & /*path*/, const jsoncons::json &origin) {
+        if (!origin.empty()) {
+          Status s;
```
PragmaTwice: Is `s` really useful here? (I have to ask again)

src/types/json.h
```diff
@@ -217,6 +217,42 @@ struct JsonValue {
...
+            jsoncons::json_options options;
+            options.max_nesting_depth(max_nesting_depth);
+            jsoncons::compact_json_string_encoder encoder{buffer, options};
+            origin.dump(encoder, ec);
+          } else if (format == JsonStorageFormat::CBOR) {
+            jsoncons::cbor::cbor_options options;
+            options.max_nesting_depth(max_nesting_depth);
+            jsoncons::cbor::basic_cbor_encoder<jsoncons::string_sink<std::string>> encoder{buffer, options};
+            origin.dump(encoder, ec);
+          }
```
PragmaTwice: Please use `Dump` and `DumpCBOR`.

src/types/json.h
```diff
@@ -217,6 +217,42 @@ struct JsonValue {
...
+          if (ec) {
+            throw std::system_error(ec);
+          }
```
PragmaTwice: Please remove it.

tests/gocase/unit/type/json/json_test.go
```diff
@@ -627,6 +627,17 @@ func TestJson(t *testing.T) {
...
+
+	t.Run("JSON.DEBUG MEMORY basics", func(t *testing.T) {
+		var result1 = make([]interface{}, 0)
+		result1 = append(result1, int64(4))
+		require.NoError(t, rdb.Do(ctx, "JSON.SET", "a", "$", `{"a":"foo", "nested": {"a": "hello"}, "nested2": {"a": 31}}`).Err())
+		require.Equal(t, rdb.Do(ctx, "JSON.DEBUG", "MEMORY", "a", "$.a").Val(), result1)
+		var result2 = make([]interface{}, 0)
+		result2 = append(result2, int64(4), int64(6), int64(2))
+		require.Equal(t, rdb.Do(ctx, "JSON.DEBUG", "MEMORY", "a", "$..a").Val(), result2)
+		require.ErrorIs(t, rdb.Do(ctx, "JSON.DEBUG", "MEMORY", "not_exists", "$").Err(), redis.Nil)
```
PragmaTwice: Could you add more test cases? e.g. for a valid `$`, or a number/array/object, or a non-existent path.

src/types/json.h
jackyyyyyssss: This function is used to traverse the internal structure of JSON and will loop through the callback function multiple times inside

PragmaTwice: I mean the variable `s`.

src/types/json.h
jackyyyyyssss: Because currently ArrayOfBulkStrings only supports string types, if modified to a template, it cannot meet the writing requirements of std:: initializer_list. Do you suggest that I add another method, std: vector<size_t>?

src/types/json.h
```diff
@@ -217,6 +217,42 @@ struct JsonValue {
     return results;
   }
 
+  StatusOr<std::vector<std::string>> GetBytes(std::string_view path, JsonStorageFormat format,
+                                              int max_nesting_depth = std::numeric_limits<int>::max()) const {
+    std::vector<std::string> results;
+    try {
+      jsoncons::jsonpath::json_query(value, path, [&](const std::string & /*path*/, const jsoncons::json &origin) {
+        if (!origin.empty()) {
```
PragmaTwice: What's the meaning of this empty check?
Could you think about whether an empty JSON array/object will occupy storage space?

src/types/json.h
jackyyyyyssss: Dump and DumpCBOR cannot meet the JSON requirements for passing values. Do you suggest modifying this method to meet these requirements

src/types/json.h
PragmaTwice: You can use a loop to dump RESP output. It doesn't matter whether you use ArrayOfBulkStrings or not.

src/types/json.h
jackyyyyyssss: i see

tests/gocase/unit/type/json/json_test.go
jackyyyyyssss: Okay, I will add all the testcases

src/types/json.h
jackyyyyyssss: ok thanks

src/types/json.h
jackyyyyyssss: Empty arrays and nulls actually take up space, and the test cases have been supplemented. Please see if there are any areas that can be optimized

src/commands/cmd_json.cc
```diff
@@ -45,6 +45,14 @@ std::string OptionalsToString(const Connection *conn, Optionals<T> &opts) {
   return str;
 }
 
+std::string SizeToString(const std::vector<std::size_t> &elems) {
+  std::string result = "*" + std::to_string(elems.size()) + CRLF;
```
PragmaTwice: You can use `MultiLen` here.

src/commands/cmd_json.cc
```diff
@@ -624,6 +632,37 @@ class CommandJsonMSet : public Commander {
...
+      path = args_[3];
+    } else if (args_.size() > 4) {
+      return {Status::RedisExecErr, "The number of arguments is more than expected"};
+    }
+
+    std::vector<std::size_t> results;
+    auto s = json.DebugMemory(args_[2], path, &results);
+
+    if (s.IsNotFound()) {
+      *output = conn->NilString();
```
PragmaTwice: It's wrong. For nonexistent key, the result is `0`.

Please confirm with redis-stack and redis-cli by yourself.

src/types/json.h
```diff
@@ -217,6 +217,32 @@ struct JsonValue {
...
+        }
+        results.emplace_back(buffer.size());
+      });
+    } catch (const jsoncons::jsonpath::jsonpath_error &e) {
+      return {Status::NotOK, e.what()};
+    }
+    if (!s) return {Status::NotOK, s.Msg()};
+    if (results.size() == 0) {
+      results.emplace_back(0);
+    }
```
PragmaTwice: It's wrong.

Please use redis-stack and redis-cli to confirm the correct result for this case.

src/commands/cmd_json.cc
jackyyyyyssss: Thank you again. The latest version of Redis Stack returns all cases, and I will make the necessary modifications later
image

src/types/json.h
```diff
@@ -217,6 +217,29 @@ struct JsonValue {
     return results;
   }
 
+  StatusOr<std::vector<size_t>> GetBytes(std::string_view path, JsonStorageFormat format,
+                                         int max_nesting_depth = std::numeric_limits<int>::max()) const {
+    std::vector<size_t> results;
+    Status s;
+    try {
+      jsoncons::jsonpath::json_query(value, path, [&](const std::string & /*path*/, const jsoncons::json &origin) {
```
PragmaTwice, line 225:
```suggestion
      jsoncons::jsonpath::json_query(value, path, [&](const std::string & /*path*/, const jsoncons::json &origin) {
        if (!s) return;
```

src/commands/cmd_json.cc
```diff
@@ -647,6 +689,7 @@ REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandJsonSet>("json.set", 4, "write", 1, 1
                         MakeCmdAttr<CommandJsonStrAppend>("json.strappend", -3, "write", 1, 1, 1),
                         MakeCmdAttr<CommandJsonStrLen>("json.strlen", -2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandJsonMGet>("json.mget", -3, "read-only", 1, -2, 1),
-                        MakeCmdAttr<CommandJsonMSet>("json.mset", -4, "write", 1, -3, 3), );
+                        MakeCmdAttr<CommandJsonMSet>("json.mset", -4, "write", 1, -3, 3),
+                        MakeCmdAttr<CommandJsonDebug>("json.debug", -3, "read-only", 1, 2, 1));
```
PragmaTwice:
```suggestion
                        MakeCmdAttr<CommandJsonDebug>("json.debug", -3, "read-only", 2, 2, 1));
```

src/types/json.h
jackyyyyyssss: Hello, I have two questions I would like to ask. 1. When does the status initialization fail and needs to be returned in advance? 2. Currently, obtaining the byte is still inaccurate, and the actual value should be larger than the encoded string. size(). Does this meet the project's standards？

PragmaTwice: There's no initialization failure.

We should check it since the lambda function can be executed more than once.

src/commands/cmd_json.cc
```diff
@@ -652,14 +652,15 @@ class CommandJsonDebug : public Commander {
...
-      return Status::OK();
+    if (s.IsNotFound()) {
+      if (args_.size() == 3) {
+        *output = redis::Integer(0);
+        return Status::OK();
+      } else {
+        *output = SizeToString(results);
+        return Status::OK();
+      }
     }
```
git-hulk:
```suggestion
    if (s.IsNotFound()) {
      if (args_.size() == 3) {
        *output = redis::Integer(0);
      } else {
        *output = SizeToString(results);
      }
      return Status::OK();
    }
```

PragmaTwice (commented):
hmm review comment still doesn't get solved yet

jackyyyyyssss:
> hmm review comment still doesn't get solved yet
I'm sorry I'm a bit clumsy, may I ask where I'm referring to
image
image

PragmaTwice:
Ahh sorry I checked the diff again. It's fixed.

PragmaTwice (approved):

Issue #2310, opened 2024-05-14
Improve consistency and isolation semantics by adding Context parameter to DB API


PokIsemaine:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

The current DB API may use multiple read operations or have nested calls. These read operations do not use fixed snapshots, which may cause different snapshot data to be read during a single operation, causing inconsistency.

### Solution

Referring to the current LatestSnapshot and GetOptions, we can add a Context parameter to each DB API, through which the API can pass a definite snapshot.

After a few simple attempts, I found that changes often have a ripple effect, requiring multiple modules to change their APIs at the same time, which can result in a huge PR that is difficult to break down into multiple smaller PRs. I'm going to try to give a draft PR in the near future to get a rough idea of what to do, and then gradually refine the changes to each module.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PokIsemaine:
Need some help:

```
struct Context {
  explicit Context(engine::Storage *storage) : storage_(storage), snapshot_(storage->GetDB()->GetSnapshot()) {}

  engine::Storage *storage_ = nullptr;
  const rocksdb::Snapshot *snapshot_ = nullptr;
  rocksdb::WriteBatchWithIndex* batch_ = nullptr;

  Context() = default;
  rocksdb::ReadOptions GetReadOptions();
  const rocksdb::Snapshot *GetSnapShot();
};
```
This is the general idea: pass in the `Context` to the Database API, use a fixed snapshot when calling, and this snapshot **will not change during the entire calling process**. When we need to read our own data for the current operation, we use `WriteBatchWithIndex` + `GetFromBatchAndDB` to obtain the data batch=>db(snapshot=ctx.snapshot).

Most operations nowadays use WriteBatch. Is there any way to copy the operation sequence from `WriteBatch` to `ctx.WriteBatchWithIndex`? In this way, I only need to modify the last `Write` part of storage, instead of modifying `ctx.WriteBatchWithIndex` when the DB API modifies `WriteBatch`.

mapleFU:
> Most operations nowadays use WriteBatch. Is there any way to copy the operation sequence from WriteBatch to ctx.WriteBatchWithIndex? In this way, I only need to modify the last Write part of storage, instead of modifying ctx.WriteBatchWithIndex when the DB API modifies WriteBatch.

`rocksdb` has the relation below:

```
[...]

PR #2332, opened 2024-05-25
refactor: Improve consistency and isolation semantics by adding Context parameter to DB API

PokIsemaine:
issue: https://github.com/apache/kvrocks/issues/2310
This is just a draft PR to demonstrate ideas, so there are a lot of imperfections.

# **A. How:**
image

Add a Context parameter to each DB API. The Context contains the fixed snapshot during the entire call process and the resulting operations (stored using rocksdb::WriteBatchWithIndex).
**Start of call:** Construct `Context`, use snapshot for fixed call, and pass it to DB API as a parameter

**During the call:**

Read operation:
  - Use rocksdb::WriteBatchWithIndex's `GetFromBatchAndDB` to read data. Only two parts of data can be seen in one call: the write operation data in this call stored in `rocksdb::WriteBatchWithIndex`; the snapshot is `Context.snapshot` DB data at the time. (`NewIterator` and `MultiGet` use the same method)
  - You will not see data from other concurrent API calls.
  
Write operations:
  - Kvrocks current write operation: `GetBatchBase` gets a `WriteBatch`, then adds operations to `WriteBatch`, and then writes the operations in `WriteBatch` to DB.
  - Modify the `writeToDB` part of Write and implement a `WriteBatch::Handler` (tentatively named `WriteBatchIndexer`) for iterating the operations in `WriteBatch` and appending them to `WriteBatchWithIndex` of `Context`.

# **B. Possible codes to pay attention to:**

**storage:**
  - Structure of Context
  - writeToDB and Get
  - batch_indexer.h: WriteBatchIndexer

**redis_db:** As an example, other APIs are derived from the redis_db API

**cmd_xx:** Construct a new Context when executing

# **C. Description: This section is used to explain the modifications that you may feel strange.**

**CI part:** I temporarily added the `-v` parameter to observe the running of golang integration tests in GitHub Action in more detail

**Test part:**
  - C++ unit test: I added `engine::Context` to `TestBase` to simplify test initialization. During the test process, I used `SetLatestSnapshot` of `Context` to get the latest `snapshot` and clear the `batch` in `Context`. This is because some tests A large number of loops are used. If the same ctx is used all the time, the data of all loops will be accumulated on it, and lead to extremely long time consuming. This is unnecessary because for the test case we do not require multiple APIs to be guaranteed to read the same `snapshot`.
  - Go integration test: I added Sleep to the BlockCommand test of list and zset, because I found that sometimes the execution order of multiple client requests would be inconsistent with the order of code writing due to fluctuations, resulting in infinite waiting or errors. As a result, for reference to other parts of the code and for convenience, I added some sleep to ensure the order.

**TODO:** TODO: ctx in the implementation code means that I am not sure whether the ctx setting here is correct. Should it come from the upper parameters or create a Context myself? Where is the boundary of the Context. I'm not familiar with the search, slot, etc. parts yet.

# **D. Improvement: The following are the points that I currently consider need to be improved.**

- `Context` should be able to replace the original `LatestSnapshot`, `GetOptions`, and merge part of the API's `ReadOptions`
- Improve comments and code
- Correctness verification:
  - Unit test for WriteBatch::Handler
  - jepsen, sync point, and script tools verify the isolation level (expected isolation level of snapshot) and possible concurrency exceptions
Performance testing, especially testing of DeleteRange related operations (if not ideal, try disabling it)

# **E. suggestion:**
* Because it is a draft PR, it may be more efficient to discuss rationality first and then discuss optimization details.
* When discussing, you can use A-E to indicate which part is being discussed


PokIsemaine:
I'll be moving forward with this PR these days, and since it's been a long time, I'll start by resolving the current conflicts and synchronizing the code with the latest unstable.

PokIsemaine:
NOTE: 
This is a relatively large PR that involves most of the functional modules. I tried to split them into smaller PRs but failed because the changes are interrelated and propagate among each other. 

Therefore, to better advance this work, I believe we need to conduct the review gradually and multiple times. At the start of each review, I will post a `"REVIEW X"` comment to inform which parts are being reviewed and highlight any points that need attention or discussion. When this review is complete, please reply to the comment with `"REVIEW X ok"` and I will start the next review. If you have any questions or suggestions during this period, please include the prefix `"REVIEW X"` to indicate which stage it belongs to. 

Although there are still some `TODO` items, I plan to discuss and refine them during the review process as you become more familiar with the background of these `TODOs`. 

If you think this approach is feasible, I will soon start the first review, change the PR from `Draft` to `Review` status, and then switch it back to Draft status for modifications. 
The overall process is: `REVIEW 1 => DRAFT modification => REVIEW 1 ok => REVIEW 2 => DRAFT modification => REVIEW 2 ok...`.

PokIsemaine:
**REVIEW 1:**
Notes: The main goal of `REVIEW 1` is to help everyone get familiar with the changes that have been made. These modifications will serve as a template for other parts, so I will select the core, stable, and simple parts to review. Before the review, please browse the previous explanations, comments, and related issues(https://github.com/apache/kvrocks/issues/2310).
* Most APIs have been added with the `engine::Context` parameter. Please note the APIs that do not have this parameter added, as they may have been overlooked.
* Many APIs originally had the `ReadOptions` parameter. Adding another `Context` parameter might overlap with `ReadOptions` in functionality, so consider whether these two parameters should be merged. For example, it might be necessary to use both the `batch` of `Context` and the complete `ReadOptions` settings (not just `Snapshot`). In order to understand everyone's thoughts, I have not modified it directly. If you think it should be merged, please provide suggestions.
* Some APIs use `LatestSnapshot`, and it's necessary to consider whether this is mandatory. If not, consider using context directly to get the specified snapshot. I haven't modified it directly in order to understand what everyone thinks. If you think it should be replaced with `Context`, please provide suggestions.
Files to review:
```
storage/storage.h
storage/storage.cc
storage/redis_db.h
storage/redis_db.cc

commands/cmd_string.cc
types/redis_string.h
types/redis_string.cc

cppunit/test_base.h
cppunit/types/string_test.cc
```

Currently, we haven't assigned reviewers for this PR. If it's convenient, I would like to kindly request @mapleFU , @git-hulk , and @PragmaTwice  to assist with the review. I understand this may take some time, so please feel free to do it at your convenience. Your feedback and suggestions will be greatly appreciated. Thank you very much for your help.

src/storage/storage.cc
```diff
@@ -604,16 +610,19 @@ rocksdb::Status Storage::Get(const rocksdb::ReadOptions &options, rocksdb::Colum
...
+                             rocksdb::ColumnFamilyHandle *column_family, const rocksdb::Slice &key,
+                             rocksdb::PinnableSlice *value) {
   rocksdb::Status s;
   if (is_txn_mode_ && txn_write_batch_->GetWriteBatch()->Count() > 0) {
     s = txn_write_batch_->GetFromBatchAndDB(db_.get(), options, column_family, key, value);
+  } else if (ctx.batch) {
+    s = ctx.batch->GetFromBatchAndDB(db_.get(), options, column_family, key, value);
   } else {
     s = db_->Get(options, column_family, key, value);
   }
```
PragmaTwice: Seems snapshot inside `ctx` is not used here?

PragmaTwice: There's a snapshot in `ctx`, and also maybe another snapshot in `options`.

How do you handle? i.e. which snapshot is used?

We should consider seriously to avoid any confusion here.

src/storage/storage.h
```diff
@@ -358,8 +365,43 @@ class Storage {
...
+struct Context {
+  engine::Storage *storage = nullptr;
+  const rocksdb::Snapshot *snapshot = nullptr;
+  std::unique_ptr<rocksdb::WriteBatchWithIndex> batch = nullptr;
+
+  rocksdb::ReadOptions GetReadOptions();
+  const rocksdb::Snapshot *GetSnapShot();
+  void SetLatestSnapshot();
+
+  explicit Context(engine::Storage *storage) : storage(storage), snapshot(storage->GetDB()->GetSnapshot()) {}
```
PragmaTwice: Will the snapshot be used in every case? If not does it increase the cost of db operations?

PragmaTwice (commented):
Thank you very much for your contribution! 

This PR is quite large, so we may need a longer time to gradually familiarize ourselves with the changes and provide our suggestions : )

src/storage/storage.cc
PragmaTwice: > Many APIs originally had the ReadOptions parameter. Adding another Context parameter might overlap with ReadOptions in functionality, so consider whether these two parameters should be merged. For example, it might be necessary to use both the batch of Context and the complete ReadOptions settings (not just Snapshot). In order to understand everyone's thoughts, I have not modified it directly. If you think it should be merged, please provide suggestions.

Sorry I've checked your note now.

I think it's better to merge these two parameter.

PragmaTwice:
> Some APIs use LatestSnapshot, and it's necessary to consider whether this is mandatory. If not, consider using context directly to get the specified snapshot. I haven't modified it directly in order to understand what everyone thinks. If you think it should be replaced with Context, please provide suggestions.

It can be removed if the Context class handles them well.

src/storage/storage.h
```diff
@@ -358,8 +365,43 @@ class Storage {
 
   rocksdb::WriteOptions default_write_opts_ = rocksdb::WriteOptions();
 
-  rocksdb::Status writeToDB(const rocksdb::WriteOptions &options, rocksdb::WriteBatch *updates);
+  rocksdb::Status writeToDB(engine::Context &ctx, const rocksdb::WriteOptions &options, rocksdb::WriteBatch *updates);
   void recordKeyspaceStat(const rocksdb::ColumnFamilyHandle *column_family, const rocksdb::Status &s);
 };
 
+struct Context {
+  engine::Storage *storage = nullptr;
```
mapleFU, line 382: You can denote should it be nullptr

src/storage/storage.h
```diff
@@ -358,8 +365,43 @@ class Storage {
 
   rocksdb::WriteOptions default_write_opts_ = rocksdb::WriteOptions();
 
-  rocksdb::Status writeToDB(const rocksdb::WriteOptions &options, rocksdb::WriteBatch *updates);
+  rocksdb::Status writeToDB(engine::Context &ctx, const rocksdb::WriteOptions &options, rocksdb::WriteBatch *updates);
   void recordKeyspaceStat(const rocksdb::ColumnFamilyHandle *column_family, const rocksdb::Status &s);
 };
 
+struct Context {
```
mapleFU, line 381: Anyway `Context` is so important, so maybe some comment should be better here, including:

1. API
2. ThreadSafety
3. Limitations

src/storage/storage.h
mapleFU: We can add a todo for changing it to lazily get the context

src/storage/storage.h
```diff
@@ -358,8 +365,43 @@ class Storage {
...
+  engine::Storage *storage = nullptr;
+  const rocksdb::Snapshot *snapshot = nullptr;
+  std::unique_ptr<rocksdb::WriteBatchWithIndex> batch = nullptr;
+
+  rocksdb::ReadOptions GetReadOptions();
```
mapleFU: const method here?

src/storage/storage.cc
PokIsemaine: Recently, I reconsidered merging `Context` and `ReadOptions` and made some small attempts to encapsulate the complete `ReadOptions` into Context while removing the ReadOptions parameter from the API. I discovered another point that needs attention: in fact, the context we need to propagate between multiple API calls only includes `Batch` and `Snapshot`. For other settings of `ReadOptions`, they should not be propagated as context.

For example: API X uses `DefaultScanOptions` to set `ReadOptions`, and at the same time, it calls API Y, while API Y only uses the default `ReadOptions` (except for `Snapshot`). One common situation is that API Y directly uses the settings of `DefaultScanOptions` because for default `ReadOptions`, we might habitually use the existing `ReadOptions` in ctx and mistakenly assume it to be the default `ReadOptions`. Actually, we need to reset `ReadOptions` for each API and use the Snapshot in Context.

If not merged, it can explicitly isolate the `ReadOptions` of each API, forcing the creation of new `ReadOptions`.

What are your thoughts and suggestions? @PragmaTwice

PokIsemaine: The snapshot property in options is set by the caller, and it gets a fixed snapshot from the ctx.

mapleFU (commented):
Great work! I'll take a carefully pass tonight!

src/storage/storage.cc
PokIsemaine: This does tend to miss or create ambiguity though, perhaps mandatory synchronization needs to be added to the relatively low-level API?
```
rocksdb::ReadOptions options_with_snapshot = options;
options_with_snapshot = ctx.GetSnapshot();
```
I'm not sure if that's a problem at the moment.

Another route would be to try to force the snapshot setting on the calling side.

src/storage/storage.h
PokIsemaine: Okay, I plan to add some necessary comments after REVIEW 1, as the design on Context may be unstable now.

src/storage/storage.cc
PokIsemaine: There is no merging in the new commit, but a DCHECK is added to check the ambiguity of ctx.snapshot and read_otpions.snapshot. The caller is required to ensure that these two settings are consistent when using them, and a reminder is given in the comments.

src/storage/storage.h
PokIsemaine: The unnecessary Context() = default constructor is now removed, and only the Context(engine::Storage *storage) constructor can be used. If the incoming storage is nullptr, an error will be reported directly during storage->GetDB().

src/storage/storage.h
PokIsemaine: The API has been redesigned. The original GetReadOptions is non-const because the snapshot may be modified silently. Now the constructor Context() = default is removed and only Context(storage) is used to ensure that there is a certain snapshot initially. At the same time, DCHECK is added to the underlying API of GET to check the consistency with the snpashot setting in read_options. Also added Comment as reminder

src/search/indexer.cc
```diff
@@ -275,17 +281,20 @@ Status IndexUpdater::UpdateHnswVectorIndex(std::string_view key, const kqir::Val
   auto storage = indexer->storage;
   auto hnsw = HnswIndex(search_key, vector, storage);
 
+  // TODO: ctx?
```
mapleFU: What's the meaning of this todo?

src/storage/storage.h
```diff
@@ -358,8 +365,49 @@ class Storage {
...
 
-  rocksdb::Status writeToDB(const rocksdb::WriteOptions &options, rocksdb::WriteBatch *updates);
+  rocksdb::Status writeToDB(engine::Context &ctx, const rocksdb::WriteOptions &options, rocksdb::WriteBatch *updates);
   void recordKeyspaceStat(const rocksdb::ColumnFamilyHandle *column_family, const rocksdb::Status &s);
 };
 
+/// Context passes fixed snapshot and batche between APIs
+/// Limitations: Performing a large number of writes on the same Context may reduce performance.
+/// Please choose to use the same Context or create a new Context based on the actual situation.
+/// Context does not provide thread safety guarantees and is generally only passed as a parameter between APIs.
```
mapleFU: LGTM except the style
```suggestion
/// Context passes fixed snapshot and batche between APIs
///
/// Limitations: Performing a large number of writes on the same Context may reduce performance.
/// Please choose to use the same Context or create a new Context based on the actual situation.
///
/// Context does not provide thread safety guarantees and is generally only passed as a parameter between APIs.
```

src/storage/storage.h
mapleFU: > We can add a todo for changing it to lazily get the context

Would you mind adding a todo here?

src/storage/storage.h
```diff
@@ -358,8 +365,49 @@ class Storage {
...
+struct Context {
+  engine::Storage *storage = nullptr;
+  /// Snapshot should be specified instead of nullptr when used,
+  /// and should be consistent with snapshot in ReadOptions to avoid ambiguity.
+  /// Normally it will be fixed to the latest Snapshot when the Context is constructed
+  const rocksdb::Snapshot *snapshot = nullptr;
+  std::unique_ptr<rocksdb::WriteBatchWithIndex> batch = nullptr;
+
+  /// GetReadOptions returns a ReadOptions whose snapshot is specified by Context
+  [[nodiscard]] rocksdb::ReadOptions GetReadOptions() const;
```
mapleFU: This api is good, still `Storage` supports some api, like:

```c++
  rocksdb::ReadOptions DefaultScanOptions() const;
  rocksdb::ReadOptions DefaultMultiGetOptions() const;
```

We may wraps the same api for the Context

src/storage/storage.h
```diff
@@ -358,8 +365,49 @@ class Storage {
...
+  void SetLatestSnapshot();
```
mapleFU: The interface is good but could we naming it `ResetLatestSnapshot` ( since it release and get a new snapshot)

src/cluster/slot_migrate.cc
```diff
@@ -1235,7 +1238,9 @@ Status SlotMigrator::sendSnapshotByRawKV() {
   read_options.snapshot = slot_snapshot_;
   rocksdb::Slice prefix_slice(prefix);
   read_options.iterate_lower_bound = &prefix_slice;
-  engine::DBIterator iter(storage_, read_options);
+  // TODO: ctx
```
mapleFU: why a todo here?

src/cluster/slot_migrate.cc
```diff
@@ -1235,7 +1238,9 @@ Status SlotMigrator::sendSnapshotByRawKV() {
...
+  engine::Context ctx(storage_);
+  engine::DBIterator iter(ctx, storage_, read_options);
```
mapleFU: It's a bit weird taht read_options and iter might have different snapshot 🤔

Would we change to:

```
engine::Context ctx(storage_);
rocksdb::ReadOptions read_options = ctx->DefaultScanOptions();
```

So that we can use same scan-options here

src/cluster/slot_migrate.cc
```diff
@@ -1259,7 +1264,7 @@ Status SlotMigrator::sendSnapshotByRawKV() {
 
     GET_OR_RET(batch_sender.Put(storage_->GetCFHandle(ColumnFamilyID::Metadata), iter.Key(), iter.Value()));
 
-    auto subkey_iter = iter.GetSubKeyIterator();
+    auto subkey_iter = iter.GetSubKeyIterator(ctx);
```
mapleFU: `iter` contains a iterator, what's the considering to re-add a ctx here?

src/storage/scripting.h
```diff
@@ -26,6 +26,7 @@
 #include "lua.hpp"
 #include "server/redis_connection.h"
 #include "status.h"
+#include "storage.h"
```
mapleFU: Maybe not an issue but maybe we can use forward declaration for this

src/storage/storage.cc
```diff
@@ -530,6 +531,7 @@ Status Storage::RestoreFromCheckpoint() {
 }
 
 bool Storage::IsEmptyDB() {
+  // TODO ctx?
```
mapleFU: Hmm I think `Storage` itself shouldn't require on `Ctx`

src/search/indexer.cc
PokIsemaine: There are instructions at the beginning of this PR
> C. Description: This section is used to explain the modifications that you may feel strange 
> TODO: TODO: ctx in the implementation code means that I am not sure whether the ctx setting here is correct. Should it come from the upper parameters or create a Context myself? Where is the boundary of the Context. I'm not familiar with the search, slot, etc. parts yet.

src/storage/storage.cc
```diff
@@ -660,46 +676,56 @@ void Storage::MultiGet(const rocksdb::ReadOptions &options, rocksdb::ColumnFamil
...
   // Put replication id logdata at the end of write batch
   if (replid_.length() == kReplIdLength) {
     updates->PutLogData(ServerLogData(kReplIdLog, replid_).Encode());
   }
 
+  if (ctx.batch == nullptr) {
+    ctx.batch = std::make_unique<rocksdb::WriteBatchWithIndex>();
+  }
+  WriteBatchIndexer handle(this, ctx.batch.get());
+  auto s = updates->Iterate(&handle);
```
mapleFU: 👍

src/search/indexer.cc
PokIsemaine: I personally plan to discuss these TODOs when reviewing the corresponding parts. Of course, you can also give some suggestions and opinions by the way, but in this way I am not arbitrarily sure of the progress of your review and whether to enter the next stage of review.

mapleFU (commented):
This does better than I ever expected! You did a nice job, some syntax need to update a little

Perhaps I think a `Context` without reading snapshot and set batch can also be provided, the interface might like:

```
struct Context {
  static Context nonTransactionCtx(Storage* storage);
};
```

This interface might be confusing and breaks the transaction's design. However, some operations are background operations like `DeleteRange` or just migrations, and it doesn't require a strong transactional interface. Currently `Context` api for them would be a little difficult, so I may think a non-txn wrapper for background ops is ok

Other part LGTM!

src/cluster/slot_migrate.cc
PokIsemaine: This is actually a passive modification: Storage::NewIterator=> SubKeyIterator ctor => GetSubKeyIterator
```
rocksdb::Iterator *Storage::NewIterator(engine::Context &ctx, const rocksdb::ReadOptions &options,
                                        rocksdb::ColumnFamilyHandle *column_family) {
  DCHECK_EQ(ctx.snapshot->GetSequenceNumber(), options.snapshot->GetSequenceNumber());
  auto iter = db_->NewIterator(options, column_family);
  if (is_txn_mode_ && txn_write_batch_->GetWriteBatch()->Count() > 0) {
    return txn_write_batch_->NewIteratorWithBase(column_family, iter, &options);
  } else if (ctx.batch && ctx.batch->GetWriteBatch()->Count() > 0) {
    return ctx.batch->NewIteratorWithBase(column_family, iter, &options); // <== HERE
  }
  return iter;
}
```

src/storage/storage.h
PokIsemaine: I don’t quite understand what “changing it to lazily get the context” means and why we need to do this. If possible, could you show it directly or commit the modification?

mapleFU: Getting a snapshot, would acquire a global "pinned" version in RocksDB, and affect data reclaiming. Maybe we can delaying the time to get the snapshot when first acquire a read 🤔

PokIsemaine:
**Update:**
- Removed `LatestSnapshot` and `GetOptions`, replacing them with `Context`.
- Added `NoTransactionContext` to `Context` along with some encapsulated APIs to obtain commonly used `ReadOptions`.
- The `Context` uses a guard in its construction, destruction, and `ResetLatestSnapshot` methods, as a data race issue with the `storage` was discovered in the replication test cases.
- Cleared all TODOs related to `Context`.
    - In `slotmigrator`, the majority of TODOs suggest that `Context` is unnecessary, as there is already a setting for `slot_snapshot_`, which has low transaction requirements. Moreover, writing a large amount of data to `context batch` could severely impact performance.
    - In the TODO for the `search` module: It seems that the basic functionality of the `search` section has been completed, so I attempted to add `Context`. The main goal was to determine the `Context` during the construction of `ExecutorContext`, allowing multiple `ExecutorNodes` on the same execution plan tree to use the same `Context`.
- Added a test for `WriteBatchIndexer`


**Review 2:**
- Modifications under `types`  
- `search` module  
- Places where `no_txn_ctx` is used

PokIsemaine:
About the latest commit https://github.com/apache/kvrocks/pull/2332/commits/822712da596effe6a3c29c4e7d6aa11fe8c648bc
There was a difference in the behavior of `Speedb` and `RocksDB` that caused stream `AutoClaim` to fail in some tests. Since I'm not very familiar with `Speedb`, I didn't find exactly what caused the difference. 
image

I referred to the related PR of `AutoClaim` and found that it has been refactored recently and tried to add the key type check. The test currently passed.
https://github.com/apache/kvrocks/pull/2373
https://github.com/apache/kvrocks/pull/2384
If there may be issues with this modification, please let me know. Thank you very much.

git-hulk:
The speedb is no longer active in the public repository after acquiring, maybe we can remove it. What do you think? @apache/kvrocks-committers

PokIsemaine:
image
Occasional problems in CI are caused by https://github.com/apache/kvrocks/issues/2473. Temporarily use sleep to make the test pass. Remove sleep after fixing it.

PokIsemaine:
Currently, the main focus is on syncing with the unstable branch. 

If it's convenient, I suggest starting the review and discussion as soon as possible, because the PR will become more complex as we progress.

For less critical improvements and optimizations, we can temporarily mark them as TODOs to be addressed in future iterations. I believe we should prioritize completing a basic functional version first.

Thank you again for taking the time to provide guidance and suggestions. :kissing_cat:

mapleFU:
Thanks for the efforts:

```
Go integration test: I added Sleep to the BlockCommand test of list and zset, because I found that sometimes the execution order of multiple client requests would be inconsistent with the order of code writing due to fluctuations, resulting in infinite waiting or errors. As a result, for reference to other parts of the code and for convenience, I added some sleep to ensure the order.
```

Actually I didn't fully get what does this part means, can you explain it with some examples?

PokIsemaine:
> Thanks for the efforts:
> 
> ```
> Go integration test: I added Sleep to the BlockCommand test of list and zset, because I found that sometimes the execution order of multiple client requests would be inconsistent with the order of code writing due to fluctuations, resulting in infinite waiting or errors. As a result, for reference to other parts of the code and for convenience, I added some sleep to ensure the order.
> ```
> 
> Actually I didn't fully get what does this part means, can you explain it with some examples?

This was a previous judgment, but there was a misunderstanding. Now I think these issues are also caused by https://github.com/apache/kvrocks/issues/2473

src/storage/batch_indexer.h
```diff
@@ -0,0 +1,83 @@
...
+#include <rocksdb/db.h>
+#include <rocksdb/slice.h>
+#include <rocksdb/utilities/write_batch_with_index.h>
+
+#include <string>
+#include <vector>
+
+#include "storage.h"
+
+// WriteBatchIndexer traverses the operations in WriteBatch and append to the specified WriteBatchWithIndex
```
mapleFU:
```suggestion
// WriteBatchIndexer traverses the operations in WriteBatch and appends to the specified WriteBatchWithIndex
```

src/storage/batch_indexer.h
```diff
@@ -0,0 +1,83 @@
...
+  rocksdb::Status SingleDeleteCF(uint32_t column_family_id, const rocksdb::Slice& key) override {
+    return dest_batch_->SingleDelete(storage_->GetCFHandle(static_cast<ColumnFamilyID>(column_family_id)), key);
+  }
+
+  void SingleDelete(const rocksdb::Slice& key) override { dest_batch_->SingleDelete(key); }
+
+  rocksdb::Status DeleteRangeCF(uint32_t column_family_id, const rocksdb::Slice& begin_key,
+                                const rocksdb::Slice& end_key) override {
+    // The latest snapshot (default ReadOptions) needs to be used here.
+    // If an old snapshot is used here and writebatchwithindex is still being built,
```
mapleFU: "and writebatchwithindex is still being built" What does this means here?

src/storage/batch_indexer.h
```diff
@@ -0,0 +1,83 @@
...
+    // the newly added keys after the snapshot may not be obtained.
+    rocksdb::ReadOptions read_options;
+    auto cf_handle = storage_->GetCFHandle(static_cast<ColumnFamilyID>(column_family_id));
+    std::unique_ptr<rocksdb::Iterator> it(storage_->GetDB()->NewIterator(read_options, cf_handle));
```
mapleFU: Better setting a lower_bound and upper_bound, this would help skipping iterating some key-values

src/storage/batch_indexer.h
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  void Merge(const rocksdb::Slice& key, const rocksdb::Slice& value) override { dest_batch_->Merge(key, value); }
+
+  void LogData(const rocksdb::Slice& blob) override { dest_batch_->PutLogData(blob); }
+
+ private:
+  engine::Storage* storage_;
+  rocksdb::WriteBatchWithIndex* dest_batch_;
+};
```
mapleFU: nit: newline after this

src/storage/batch_indexer.h
```diff
@@ -0,0 +1,83 @@
...
+#include <string>
+#include <vector>
+
+#include "storage.h"
+
+// WriteBatchIndexer traverses the operations in WriteBatch and append to the specified WriteBatchWithIndex
+class WriteBatchIndexer : public rocksdb::WriteBatch::Handler {
+ public:
+  explicit WriteBatchIndexer(engine::Storage* storage, rocksdb::WriteBatchWithIndex* dest_batch)
+      : storage_(storage), dest_batch_(dest_batch) {}
```
mapleFU: adding `DCHECK` the `storage_` and `dest_batch_` is not nullptr here?

src/storage/storage.h
```diff
@@ -361,8 +368,75 @@ class Storage {
...
 
+/// Context passes fixed snapshot and batche between APIs
+///
+/// Limitations: Performing a large number of writes on the same Context may reduce performance.
+/// Please choose to use the same Context or create a new Context based on the actual situation.
+///
+/// Context does not provide thread safety guarantees and is generally only passed as a parameter between APIs.
+struct Context {
+  engine::Storage *storage = nullptr;
+  /// Snapshot should be specified instead of nullptr when used,
```
mapleFU: I guess this is conflict with `NonTransactionContext`

src/storage/storage.cc
```diff
@@ -636,21 +648,31 @@ void Storage::recordKeyspaceStat(const rocksdb::ColumnFamilyHandle *column_famil
...
 }
 
-void Storage::MultiGet(const rocksdb::ReadOptions &options, rocksdb::ColumnFamilyHandle *column_family,
-                       const size_t num_keys, const rocksdb::Slice *keys, rocksdb::PinnableSlice *values,
-                       rocksdb::Status *statuses) {
+void Storage::MultiGet(engine::Context &ctx, const rocksdb::ReadOptions &options,
+                       rocksdb::ColumnFamilyHandle *column_family, const size_t num_keys, const rocksdb::Slice *keys,
+                       rocksdb::PinnableSlice *values, rocksdb::Status *statuses) {
+  if (ctx.is_txn_mode) {
+    DCHECK_EQ(ctx.snapshot->GetSequenceNumber(), options.snapshot->GetSequenceNumber());
```
mapleFU, line 675: We'd better check `options.snapshot != nullptr` before checking equality

src/storage/storage.h
```diff
@@ -361,8 +368,75 @@ class Storage {
...
+  /// and should be consistent with snapshot in ReadOptions to avoid ambiguity.
+  /// Normally it will be fixed to the latest Snapshot when the Context is constructed
+  const rocksdb::Snapshot *snapshot = nullptr;
+  std::unique_ptr<rocksdb::WriteBatchWithIndex> batch = nullptr;
+
+  bool is_txn_mode = true;
```
mapleFU, line 392: Nit: we can adding some comments for `is_txn_mode`

mapleFU (dismissed):
LGTM % some nits! Thanks!

mapleFU:
@PragmaTwice @git-hulk This pr is ready % some nits and comments, and it's ready for review, would you mind take a look?

src/storage/batch_indexer.h
PokIsemaine: I might need to rethink this, as I may have been disrupted by an interim test.

**Previous thoughts:**
We can perceive operations on the database from two places: the database that specifies the `snapshot and` the current `WriteBatchWithIndex`.

Assuming the following operations now occur:

```
OP1: snapshot1(ctx) => snapshot2
PUT(k1,a), PUT(k2,b), PUT(k3,c), PUT(k4,d)

OP2: snapshot1(ctx)
DeleteRange(k2, k3)
```

`snapshot1` cannot perceive the existence of keys k1-k4 in the `snapshot2` database (after OP1 is completed), and thus it cannot iterate over these keys via an iterator to convert `DeleteRange` into individual `Delete`. 

On the other hand, the `WriteBatchWithIndex` is still incomplete; `batch_indexer` is there to construct the `WriteBatchWithIndex`, but now it depends on it as well.

**Current thoughts:**
We don't need a complete `WriteBatchWithIndex`, the `dest_batch_` should already have the `PUT` operations that can be detected. 

I'll give it a try later.

src/storage/batch_indexer.h
PokIsemaine: Ok, how about adding DefaultScanOptions?

mapleFU: Both ok to me 🤔

I need correct myself that "skipping iterating some key-values" is not right, the right wording might be "skipping lock some sst". Adding in `DefaultScanOptions(begin, end)` is also ok to me

src/storage/batch_indexer.h
mapleFU: From my aspect, `snapshot` is for read, and `batch_indexer` is for write. OP1 writing to global area, and when construct snapshot2, it might read new operations from the same area.

PokIsemaine: > it might read new operations from the same area.

What's the meaning?

src/storage/batch_indexer.h
PokIsemaine: I mean using both like other operations.

DefaultScanOptions + lower_bound + upper_bound

For example
```
rocksdb::ReadOptions read_options = storage_->DefaultScanOptions();
read_options.iterate_lower_bound = &begin_key;
read_options.iterate_upper_bound = &end_key;
```

src/storage/batch_indexer.h
mapleFU: Assuming concurrent txn x2 write to the k2.5 ( writing to the area that snapshot2 can read) and DeleteRange would remove it if a new snapshot is created? 🤔

src/storage/batch_indexer.h
mapleFU: Sure, if we have lower bound and upper_bound, then we'd better adding them

git-hulk:
> @PragmaTwice @git-hulk This pr is ready % some nits and comments, and it's ready for review, would you mind take a look?

Sure, will retake another pass tomorrow.

src/storage/batch_indexer.h
PokIsemaine: Do you mean that if snapshot adopts the latest snapshot (default setting), OP2 will be aware of k2.5 and delete it?

```
OP1: snapshot1(ctx) => snapshot2
PUT(k1,a), PUT(k2,b), PUT(k3,c), PUT(k4,d)
                                              OPa:snapshot2(ctx_a)
                                              PUT(k2.5)
OP2: snapshot1(ctx)
DeleteRange(k2, k3) // The latest snapshot is used when converting from DeleteRange to Delete
```


I think this does happen, and that's actually what this PR is trying to fix: reading inconsistent snapshots.

Previously I thought that the circular dependency of WriteBatchWithIndex (as mentioned above) caused this to be unfixable and compromised by using the default settings, but now it seems that it should be fixable.

mapleFU: Yes, so I think "snapshot1 cannot perceive the existence of keys k1-k4 in the snapshot2 database (after OP1 is completed), and thus it cannot iterate over these keys via an iterator to convert DeleteRange into individual Delete." would be weird, lol

src/storage/batch_indexer.h
mapleFU: FYI: https://github.com/facebook/rocksdb/wiki/Iterator#iterating-upper-bound-and-lower-bound

src/storage/storage.h
```diff
@@ -361,8 +368,78 @@ class Storage {
 
   rocksdb::WriteOptions default_write_opts_ = rocksdb::WriteOptions();
 
-  rocksdb::Status writeToDB(const rocksdb::WriteOptions &options, rocksdb::WriteBatch *updates);
+  rocksdb::Status writeToDB(engine::Context &ctx, const rocksdb::WriteOptions &options, rocksdb::WriteBatch *updates);
   void recordKeyspaceStat(const rocksdb::ColumnFamilyHandle *column_family, const rocksdb::Status &s);
 };
 
+/// Context passes fixed snapshot and batche between APIs
```
git-hulk:
```suggestion
/// Context passes fixed snapshot and batch between APIs
```

src/storage/storage.h
```diff
@@ -361,8 +368,78 @@ class Storage {
...
+
+  /// TODO: Change it to defer getting the context, and the snapshot is pinned after the first read operation
+  explicit Context(engine::Storage *storage) : storage(storage) {
+    auto guard = storage->ReadLockGuard();
+    snapshot = storage->GetDB()->GetSnapshot();  // NOLINT
+  }
+  ~Context() {
+    if (storage) {
+      auto guard = storage->WriteLockGuard();
+      if (storage->GetDB()) {
```
git-hulk: Should check if the snapshot is null here?

src/storage/storage.h
```diff
@@ -225,32 +227,37 @@ class Storage {
   Status ApplyWriteBatch(const rocksdb::WriteOptions &options, std::string &&raw_batch);
   rocksdb::SequenceNumber LatestSeqNumber();
 
-  [[nodiscard]] rocksdb::Status Get(const rocksdb::ReadOptions &options, const rocksdb::Slice &key, std::string *value);
-  [[nodiscard]] rocksdb::Status Get(const rocksdb::ReadOptions &options, rocksdb::ColumnFamilyHandle *column_family,
+  [[nodiscard]] rocksdb::Status Get(engine::Context &ctx, const rocksdb::ReadOptions &options,
```
git-hulk, line 230: Would it be better to provide an API group to allow us to omit the options(get from the context)? I see many places are using the options from the context.

src/storage/storage.cc
```diff
@@ -1224,4 +1261,31 @@ bool Storage::ReplDataManager::FileExists(Storage *storage, const std::string &d
...
+  return read_options;
+}
+
+[[nodiscard]] rocksdb::ReadOptions Context::DefaultMultiGetOptions() const {
+  rocksdb::ReadOptions read_options = storage->DefaultMultiGetOptions();
+  if (is_txn_mode) read_options.snapshot = snapshot;
+  return read_options;
+}
+
+void Context::ResetLatestSnapshot() {
```
git-hulk: It should be `RefreshLatestSnapshot`? since it will assign a new one after releasing the old one.

src/storage/storage.cc
```diff
@@ -1224,4 +1261,31 @@ bool Storage::ReplDataManager::FileExists(Storage *storage, const std::string &d
...
+  auto guard = storage->WriteLockGuard();
+  if (snapshot) {
+    storage->GetDB()->ReleaseSnapshot(snapshot);
+  }
+  snapshot = storage->GetDB()->GetSnapshot();
+  batch->Clear();
```
git-hulk: Need to check if the batch is null?

src/storage/storage.cc
```diff
@@ -587,15 +588,20 @@ Status Storage::GetWALIter(rocksdb::SequenceNumber seq, std::unique_ptr<rocksdb:
...
-rocksdb::Status Storage::Get(const rocksdb::ReadOptions &options, rocksdb::ColumnFamilyHandle *column_family,
-                             const rocksdb::Slice &key, std::string *value) {
+rocksdb::Status Storage::Get(engine::Context &ctx, const rocksdb::ReadOptions &options,
+                             rocksdb::ColumnFamilyHandle *column_family, const rocksdb::Slice &key,
+                             std::string *value) {
+  DCHECK_EQ(ctx.snapshot->GetSequenceNumber(), options.snapshot->GetSequenceNumber());
   rocksdb::Status s;
   if (is_txn_mode_ && txn_write_batch_->GetWriteBatch()->Count() > 0) {
     s = txn_write_batch_->GetFromBatchAndDB(db_.get(), options, column_family, key, value);
+  } else if (ctx.batch) {
```
git-hulk: Do we need to check if `is_txn_mode` is true here?

src/search/hnsw_indexer.h
```diff
@@ -41,16 +41,17 @@ struct HnswNode {
 
   HnswNode(NodeKey key, uint16_t level);
 
-  StatusOr<HnswNodeFieldMetadata> DecodeMetadata(const SearchKey& search_key, engine::Storage* storage) const;
+  StatusOr<HnswNodeFieldMetadata> DecodeMetadata(engine::Context& ctx, const SearchKey& search_key,
```
git-hulk: We can remove the `storage` parameter here?

git-hulk:
@PokIsemaine Sorry for the late review, a few comments inline.

src/search/hnsw_indexer.h
PokIsemaine: storage can be obtained through ctx

src/storage/storage.cc
PokIsemaine: Ah thanks, I missed the modification here

src/storage/storage.h
PokIsemaine: Do you mean to put ReadOptions in Context? 

Here are the ideas I mentioned before
> Recently, I reconsidered merging Context and ReadOptions and made some small attempts to encapsulate the complete ReadOptions into Context while removing the ReadOptions parameter from the API. I discovered another point that needs attention: in fact, the context we need to propagate between multiple API calls only includes Batch and Snapshot. For other settings of ReadOptions, they should not be propagated as context.

> For example: API X uses DefaultScanOptions to set ReadOptions, and at the same time, it calls API Y, while API Y only uses the default ReadOptions (except for Snapshot). One common situation is that API Y directly uses the settings of DefaultScanOptions because for default ReadOptions, we might habitually use the existing ReadOptions in ctx and mistakenly assume it to be the default ReadOptions. Actually, we need to reset ReadOptions for each API and use the Snapshot in Context.

> If not merged, it can explicitly isolate the ReadOptions of each API, forcing the creation of new ReadOptions.

Do you have any suggestions?

src/search/hnsw_indexer.h
git-hulk: oh, yes. What I want to say is to remove this storage in this and following methods since we can get the storage via the context.

src/storage/storage.h
git-hulk: @PokIsemaine Thanks for your reference. That makes sense to keep the current implementation.

mapleFU:
Unrelated to this pr: I think we can also add a server side conf to allow disable `is_txn_mode`, this is useful when we have some performance issue with DeleteRange or make something wrong. We can merge this first and add this in a separate patch

PokIsemaine:
> Unrelated to this pr: I think we can also add a server side conf to allow disable `is_txn_mode`, this is useful when we have some performance issue with DeleteRange or make something wrong. We can merge this first and add this in a separate patch

I agree, this PR has a wider impact and we should provide the option to turn it on and off. 

In subsequent observations, in addition to performance and errors, memory usage also needs attention, because we have an additional `WriteBatchWithIndex`.

mapleFU (approved):
+1, but I think we need fast collect all of us' approve since it's really huge
@PragmaTwice @git-hulk

git-hulk (approved):

mapleFU:
I'll draft to merge this patch if no negative comment this week

mapleFU:
Will merge this after ci pass

git-hulk:
> Will merge this after ci pass

I'm good with this.

PR #2334, opened 2024-05-27
Improve RESP handling code in replication

PragmaTwice:
See changes.

### DIFF
--- a/src/commands/error_constants.h
+++ b/src/commands/error_constants.h
@@ -43,5 +43,7 @@ inline constexpr const char *errValueIsNotFloat = "value is not a valid float";
 inline constexpr const char *errNoMatchingScript = "NOSCRIPT No matching script. Please use EVAL";
 inline constexpr const char *errUnknownOption = "unknown option";
 inline constexpr const char *errUnknownSubcommandOrWrongArguments = "Unknown subcommand or wrong number of arguments";
+inline constexpr const char *errWrongNumArguments = "ERR wrong number of arguments";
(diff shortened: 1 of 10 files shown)


git-hulk (dismissed):

src/commands/error_constants.h
```diff
@@ -43,5 +43,7 @@ inline constexpr const char *errValueIsNotFloat = "value is not a valid float";
 inline constexpr const char *errNoMatchingScript = "NOSCRIPT No matching script. Please use EVAL";
 inline constexpr const char *errUnknownOption = "unknown option";
 inline constexpr const char *errUnknownSubcommandOrWrongArguments = "Unknown subcommand or wrong number of arguments";
+inline constexpr const char *errWrongNumArguments = "ERR wrong number of arguments";
```
mapleFU, line 46: Should we remove ERR here? ( Seems some str has Err, some are not..)

PragmaTwice: I keep it for convenience, since it should be included in both use sites.

git-hulk: We cannot remove the ERR prefix since it may introduce a compatibility issue.

> Seems some str has Err, some are not..

It's an old implementation issue, Redis uses ERR/LOADING/NOPROTO to identify different kinds of errors, but we didn't take care of this in the original implementation. So it's really hard to fix them in a short time.

refer: https://redis.io/docs/latest/develop/reference/protocol-spec/

mapleFU (dismissed):

git-hulk (dismissed):

git-hulk (approved):

PR #2340, opened 2024-05-30
fix(json): int64 overflow in JSON numop

jackyyyyyssss:
Json_val. NumOp only considers cases where the value of the double type is out of bounds, and does not consider cases where the int64 type is out of bounds. If the int64 type is out of bounds, it will cause the lambda function to exit abnormally and not catch. This exception will cause the kvrocks to crash
27319d3236f39a914025911dac65157
ff70eb032f10ad79a3a379996df76fc

### DIFF
--- a/src/common/bit_util.h
+++ b/src/common/bit_util.h
@@ -41,13 +41,7 @@ inline size_t RawPopcount(const uint8_t *p, int64_t count) {
 
   return bits;
 }
-inline bool Int64OperationOverFlow(int64_t a, int64_t b, int64_t result, uint8_t operation) {
-  if (operation == 1) {
-    return __builtin_add_overflow(a, b, &result);
-  } else {
-    return __builtin_mul_overflow(a, b, &result);
-  }
-}
+
 
 template <typename T = void>
 inline int ClzllWithEndian(uint64_t x) {
--- a/src/common/bit_util.h
+++ b/src/common/bit_util.h
@@ -42,7 +42,6 @@ inline size_t RawPopcount(const uint8_t *p, int64_t count) {
   return bits;
 }
 
-
 template <typename T = void>
 inline int ClzllWithEndian(uint64_t x) {
   if constexpr (IsLittleEndian()) {
--- a/src/common/bit_util.h
+++ b/src/common/bit_util.h
@@ -41,6 +41,13 @@ inline size_t RawPopcount(const uint8_t *p, int64_t count) {
 
   return bits;
 }
+inline bool Int64OperationOverFlow(int64_t a, int64_t b, int64_t result, uint8_t operation) {
+  if (operation == 1) {
+    return __builtin_add_overflow(a, b, &result);
+  } else {
+    return __builtin_mul_overflow(a, b, &result);
+  }
+}
 
 template <typename T = void>
 inline int ClzllWithEndian(uint64_t x) {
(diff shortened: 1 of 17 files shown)


PragmaTwice:
Thank you for your great finding.

src/types/redis_json.cc
```diff
@@ -452,6 +452,11 @@ rocksdb::Status Json::numop(JsonValue::NumOpEnum op, const std::string &user_key
   if (!number_res || !number_res.GetValue().value.is_number()) {
     return rocksdb::Status::InvalidArgument("should be a number");
   }
+
+  if (!number_res.GetValue().value.is_int64() && !number_res.GetValue().value.is_double()) {
+    return rocksdb::Status::InvalidArgument("number out of range");
+  }
```
PragmaTwice: I think it can be fixed in the JsonValue::NumOp. Here we can only fix for `value`, not the number inside the current json.

jackyyyyyssss: Thank you. Summarized all the situations to see if there were any omissions

src/common/bit_util.h
```diff
@@ -41,6 +41,13 @@ inline size_t RawPopcount(const uint8_t *p, int64_t count) {
 
   return bits;
 }
+inline bool Int64OperationOverFlow(int64_t a, int64_t b, int64_t result, uint8_t operation) {
+  if (operation == 1) {
+    return __builtin_add_overflow(a, b, &result);
+  } else {
+    return __builtin_mul_overflow(a, b, &result);
+  }
+}
```
PragmaTwice: This function looks very weird. What's `operation`? What's `1` here? Why `mul` is executed if `operation` is not `1`?

Please remove it.

PragmaTwice (changes requested):
Please just add a catch block. Try not to add any meaningless weird code.

src/common/bit_util.h
PragmaTwice: Also, this function is totally wrong since the result is passed by value.

jackyyyyyssss:
> Please just add a catch block. Try not to add any meaningless weird code.

Simple catch doesn't seem to work, some overflow cases can return results that don't meet expectations without throwing exceptions  
For example, 9223372036854775806<std:: numericlimits:: max()  +20 is a negative number and will not throw an exception

PragmaTwice:
> > Please just add a catch block. Try not to add any meaningless weird code.
> 
> Simple catch doesn't seem to work, some overflow cases can return results that don't meet expectations without throwing exceptions For example, 9223372036854775806<std:: numericlimits:: max() +20 is a negative number and will not throw an exception

It can be actually well defined. It follows the [two's complement](https://en.m.wikipedia.org/wiki/Two's_complement). So generally we can leave it as is until we make a big number implementation available.

jackyyyyyssss:
> So generally we can leave it as is until we make a big number implementation available.

What do you mean by directly adding a catch (const jsoncons:: json_runtime_error<std:: runtime_error>e){
Return {Status:: NotOK, e.what()};
}Is that enough for this ？

PragmaTwice:
> > So generally we can leave it as is until we make a big number implementation available.
> 
> What do you mean by directly adding a catch (const jsoncons:: json_runtime_error<std:: runtime_error>e){ Return {Status:: NotOK, e.what()}; }Is that enough for this ？

Yeah, just preventing kvrocks from crash is enough.

PragmaTwice:
Sorry I looked into this problem a little bit, and seems it's a design issue inside the NumOp method.

I opened a PR to address it (#2345).

But anyway thank you very much for your great finding and effort.

jackyyyyyssss:
> Sorry I looked into this problem a little bit, and seems it's a design issue inside the NumOp method.
> 
> I opened a PR to address it (#2345).
> 
> But anyway thank you very much for your great finding and effort.

You have made the same changes to the visual and redis stack processing, and have learned a lot

PR #2343, opened 2024-05-30
chore(storage): remove FindKeyRangeWithPrefix and fix DeleteRange

PragmaTwice:
The method `FindKeyRangeWithPrefix` seems useless since we can always construct a key range quickly, without knowing the exact first key and last key in the range.

### DIFF
--- a/src/common/string_util.cc
+++ b/src/common/string_util.cc
@@ -377,4 +377,14 @@ std::string EscapeString(std::string_view s) {
   return str;
 }
 
+std::string StringNext(std::string s) {
+  for (auto iter = s.rbegin(); iter != s.rend(); ++iter) {
+    if (*iter != char(0xff)) {
+      (*iter)++;
+      break;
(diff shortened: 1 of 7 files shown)


git-hulk (approved):

src/common/string_util.cc
```diff
@@ -377,4 +377,14 @@ std::string EscapeString(std::string_view s) {
...
 }
 
+std::string StringNext(std::string s) {
+  for (auto iter = s.rbegin(); iter != s.rend(); ++iter) {
+    if (*iter != char(0xff)) {
+      (*iter)++;
+      break;
+    }
+  }
+  return s;
```
mapleFU, line 387: should we append a '0x00' if `s == return s`?

PragmaTwice: I think the semantics is different between appending 0x00 and increasing last char by 1.

PR #2347, opened 2024-06-01
fix(config): avoid rewriting the config file if it's unnecessary

git-hulk:
The server start will rewrite the config file to persist namespace/token pairs
if the namespace replication is enabled. But it's unnecessary if the namespace replication
is disabled or no tokens were loaded from the configuration file because the purpose of
this rewrite is to remove tokens from the config file and write to the Database.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -885,11 +885,20 @@ Status Config::Set(Server *srv, std::string key, const std::string &value) {
     if (!s.IsOK()) return s.Prefixed("invalid value");
   }
 
+  auto origin_value = field->ToString();
   auto s = field->Set(value);
   if (!s.IsOK()) return s.Prefixed("failed to set new value");
 
   if (field->callback) {
-    return field->callback(srv, key, value);
+    s = field->callback(srv, key, value);
+    if (!s.IsOK()) {
+      // rollback the value if the callback failed
+      auto set_status = field->Set(origin_value);
+      if (!set_status.IsOK()) {
+        return set_status.Prefixed("failed to rollback the value");
+      }
+    }
+    return s;
   }
 
   return Status::OK();
--- a/src/server/namespace.cc
+++ b/src/server/namespace.cc
@@ -51,35 +51,46 @@ bool Namespace::IsAllowModify() const {
   return config->HasConfigFile() || config->repl_namespace_enabled;
 }
 
+Status Namespace::loadFromDB(std::map<std::string, std::string>* db_tokens) const {
+  std::string value;
+  auto s = storage_->Get(rocksdb::ReadOptions(), cf_, kNamespaceDBKey, &value);
+  if (!s.ok()) {
+    if (s.IsNotFound()) return Status::OK();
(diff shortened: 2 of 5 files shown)


src/config/config.cc
```diff
@@ -885,11 +885,20 @@ Status Config::Set(Server *srv, std::string key, const std::string &value) {
...
+  auto origin_value = field->ToString();
   auto s = field->Set(value);
   if (!s.IsOK()) return s.Prefixed("failed to set new value");
 
   if (field->callback) {
-    return field->callback(srv, key, value);
+    s = field->callback(srv, key, value);
+    if (!s.IsOK()) {
+      // rollback the value if the callback failed
+      auto setStatus = field->Set(origin_value);
```
git-hulk: This place is a bug fix, we should rollback the configuration value if it's failed in the callback function.

PragmaTwice:
It can fix the problem when there's no namespace in kvrocks except the default one.

But what if there are some namespaces? We also need to consider this situation.

src/server/namespace.cc
```diff
@@ -51,36 +51,47 @@ bool Namespace::IsAllowModify() const {
...
+
 Status Namespace::LoadAndRewrite() {
   auto config = storage_->GetConfig();
   // Namespace is NOT allowed in the cluster mode, so we don't need to rewrite here.
   if (config->cluster_enabled) return Status::OK();
 
-  // Load from the configuration file first
-  tokens_ = config->load_tokens;
+  std::map<std::string, std::string> db_tokens;
+  auto s = loadFromDB(&db_tokens);
```
PragmaTwice, line 75: Why not just put `tokens_` into `loadFromDB` here?

PragmaTwice: Ahh need to merge.

git-hulk: Yes, it will be merged only if it doesn’t exist in the config file.

git-hulk:
> But what if there are some namespaces? We also need to consider this situation.

`Rewrite()` will persist them if there are any namespaces(except the default), and then remove them from the config file. So it should be good for this scenario.

PragmaTwice (dismissed):
Sounds good to me.

src/config/config.cc
```diff
@@ -885,11 +885,20 @@ Status Config::Set(Server *srv, std::string key, const std::string &value) {
...
 
   if (field->callback) {
-    return field->callback(srv, key, value);
+    s = field->callback(srv, key, value);
+    if (!s.IsOK()) {
+      // rollback the value if the callback failed
+      auto set_status = field->Set(origin_value);
+      if (!set_status.IsOK()) {
+        return set_status.Prefixed("failed to rollback the value");
+      }
```
PragmaTwice, line 899: I think maybe `Set` has other side effects, so maybe here the rollback is a little dangerous.

e.g. there's a `x++` in `callback`, it will be triggered twice here.

PragmaTwice (dismissed):

PragmaTwice (approved):

Issue #2344, opened 2024-05-30
Docker images with Debian or Ubuntu


thomas-0816:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Since many companies run their production servers on Debian or Ubuntu, it would be nice to have a Debian or Ubuntu container on docker hub as well, for local testing.

e.g. I tested with a single stage build:

```
# image size is 114M compared to alpine build with 38M uncompressed

FROM debian:bookworm-slim

# chain commands for single layer
RUN apt-get -y update \
    && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-recommends install git build-essential cmake libtool \
        python3 libssl-dev autoconf automake ninja-build redis-tools ca-certificates \
    && git clone --depth 1 --branch v2.8.0 https://github.com/apache/kvrocks.git \
    && cd kvrocks/ \
    && ./x.py build -DENABLE_OPENSSL=ON -DPORTABLE=1 -DCMAKE_BUILD_TYPE=Release -j $(nproc) \
    && cp build/kvrocks /bin/kvrocks \
    && cd / \
    && rm -rf /kvrocks \
    && apt-get purge -y --auto-remove git build-essential cmake libtool python3 libssl-dev autoconf automake ninja-build \
    && apt-get clean \
    && rm -rf /tmp/* /var/lib/apt/lists/* /var/cache/apt/archives/*

EXPOSE 6666

HEALTHCHECK ...

ENTRYPOINT ["/bin/kvrocks"...]
```

Release tags could be 2.8.0-debian, 2.8.0-alpine, etc.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

git-hulk:
@thbley Thanks for your proposal.

@apache/kvrocks-committers What do you think about this? From my perspective, it did help when users wanted to debug the kvrocks.

enjoy-binbin:
Sound good to me.

aleksraiden:
I think, we can close this issue as done

git-hulk:
@aleksraiden Yes, closing.

PR #2348, opened 2024-06-01
build: Change base Docker image from Alpine to Debian

aleksraiden:
Change base Docker image from alpine 3.16 (as an too old system) into Debian - use bookworm-slim image. (see our issue too https://github.com/apache/kvrocks/issues/2344)

As we can't upgrade alpine, for continuous use latest rocksdb release and fix some mysterious (for me, see https://github.com/apache/kvrocks/pull/2327) bugs with jemalloc in alpine, I propagate an new Docker, based on latest Debian slim. 

Basically, nothing to change from original, just change an image and use apt instead of apk. Added only an apt clean command for cleanup system after upgrade. 

Result image has an ~ 120Mb size instead of ~40 Mb in alpine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ RUN ./x.py build -DENABLE_OPENSSL=ON -DPORTABLE=1 -DCMAKE_BUILD_TYPE=Release -j 
 
 FROM debian:bookworm-slim
 
-RUN DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get upgrade -y && apt-get clean
+RUN DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get upgrade -y && apt-get -y install openssl ca-certificates && apt-get clean
 
 RUN mkdir /var/run/kvrocks
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ RUN ./x.py build -DENABLE_OPENSSL=ON -DPORTABLE=1 -DCMAKE_BUILD_TYPE=Release -j 
 
 FROM debian:bookworm-slim
 
-RUN DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get upgrade -y && apt-get -y install openssl ca-certificates && apt-get clean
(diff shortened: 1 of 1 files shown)


aleksraiden:
Thanks to @thbley about the some fix and optimizing from https://github.com/apache/kvrocks/issues/2344

aleksraiden:
Result image has an ~ 120Mb size instead of ~40 Mb in alpine. I think that its OK.

git-hulk:
@aleksraiden Thanks for taking care of this.

aleksraiden:
Some troubles in healthcheck on docker, try to fix today or tomorrow

PragmaTwice:
> As we can't upgrade alpine

Hmm is there any issue for us about upgrading alpine?

PragmaTwice:
And also, do you have any suggestion to select between debian and ubuntu?

thomas-0816:
> And also, do you have any suggestion to select between debian and ubuntu?

To me both are quite good and work great in production for many companies.
Ubuntu lts versions have longer support lifetime, more recent versions of packages (e.g. gcc, openssl), Debian more stable, less recent versions. Key for alpine is size and security, but no glibc.

for package version comparison:
https://packages.ubuntu.com/
https://www.debian.org/distrib/packages
https://pkgs.alpinelinux.org/packages

aleksraiden:
Yeah, sorry, I have not any real experience with ubuntu, but over 10 y use Debian on all our servers, so use it as a matured server-side os.

In other hands, the idea of multiple images (like a community or non-official, in separate directory etc) with other base images my Ok and I can support this feature. 

In my mind, ideally - alpine 3.20+ but we cant switch due to some dependency. So - Debian looks like good solution at now.

aleksraiden:
> > As we can't upgrade alpine
> 
> Hmm is there any issue for us about upgrading alpine?

With @torwig we have a lot of internal tests and a core point is libexecinfo dependency

git-hulk (approved):

PragmaTwice (approved):

PR #2360, opened 2024-06-09
feat(cron): add support for "*/n" interval cronjob syntax

kinoute:
Currently KVRocks only supports integers and * in crontab format. I thought it would be nice to add support for the `*/n` interval syntax. I understand that we can already add multiple crontab entries for the same feature like this:

```
bgsave-cron 0 3 * * * 0 6 * * * (and more...)
```

But it would be convenient if we could do something like this instead:

```
bgsave-cron 0 */3 * * *
```

Again, I'm new to C++ so no offense taken if the code provided is not good at all for your standards. The "month" part was a pain to do and I'm not sure about it. I used https://crontab.guru/#0_*_*_*/4_* to calibrate my code/tests for this.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -492,7 +492,7 @@ profiling-sample-record-threshold-ms 100
 ################################## CRON ###################################
 
 # Compact Scheduler, auto compact at schedule time
-# Time expression format is the same as crontab (currently only support *, int and */n)
+# Time expression format is the same as crontab (currently only support *, int and */int)
 # e.g. compact-cron 0 3 * * * 0 4 * * *
 # would compact the db at 3am and 4am everyday
 # compact-cron 0 3 * * *
@@ -522,7 +522,7 @@ compaction-checker-range 0-7
 # Kvrocks doesn't store the key number directly. It needs to scan the DB and
 # then retrieve the key number by using the dbsize scan command.
 # The Dbsize scan scheduler auto-recalculates the estimated keys at scheduled time.
-# Time expression format is the same as crontab (currently only support *, int and */n)
+# Time expression format is the same as crontab (currently only support *, int and */int)
 # e.g. dbsize-scan-cron 0 * * * *
 # would recalculate the keyspace infos of the db every hour.
 
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -492,7 +492,7 @@ profiling-sample-record-threshold-ms 100
 ################################## CRON ###################################
 
 # Compact Scheduler, auto compact at schedule time
-# time expression format is the same as crontab(currently only support * and int)
+# Time expression format is the same as crontab (currently only support *, int and */n)
 # e.g. compact-cron 0 3 * * * 0 4 * * *
 # would compact the db at 3am and 4am everyday
 # compact-cron 0 3 * * *
@@ -515,14 +515,14 @@ compaction-checker-range 0-7
 # force-compact-file-min-deleted-percentage 10
 
 # Bgsave scheduler, auto bgsave at scheduled time
-# time expression format is the same as crontab(currently only support * and int)
+# Time expression format is the same as crontab (currently only support *, int and */n)
 # e.g. bgsave-cron 0 3 * * * 0 4 * * *
 # would bgsave the db at 3am and 4am every day
 
 # Kvrocks doesn't store the key number directly. It needs to scan the DB and
 # then retrieve the key number by using the dbsize scan command.
 # The Dbsize scan scheduler auto-recalculates the estimated keys at scheduled time.
-# Time expression format is the same as crontab (currently only support * and int)
+# Time expression format is the same as crontab (currently only support *, int and */n)
 # e.g. dbsize-scan-cron 0 * * * *
 # would recalculate the keyspace infos of the db every hour.
 
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -515,7 +515,7 @@ compaction-checker-range 0-7
 # force-compact-file-min-deleted-percentage 10
 
 # Bgsave scheduler, auto bgsave at scheduled time
-# Time expression format is the same as crontab (currently only support *, int and */n)
+# Time expression format is the same as crontab (currently only support *, int and */int)
 # e.g. bgsave-cron 0 3 * * * 0 4 * * *
 # would bgsave the db at 3am and 4am every day
 
--- a/src/common/cron.cc
+++ b/src/common/cron.cc
@@ -63,11 +63,19 @@ bool Cron::IsTimeMatch(const tm *tm) {
     return false;
   }
   for (const auto &st : schedulers_) {
-    bool minuteMatch = (st.minute == -1 || tm->tm_min == st.minute || (st.minute > 0 && tm->tm_min % st.minute == 0));
-    bool hourMatch = (st.hour == -1 || tm->tm_hour == st.hour || (st.hour > 0 && tm->tm_hour % st.hour == 0));
-    bool mdayMatch = (st.mday == -1 || tm->tm_mday == st.mday);
-    bool monthMatch = (st.month == -1 || (tm->tm_mon + 1) == st.month);
-    bool wdayMatch = (st.wday == -1 || tm->tm_wday == st.wday);
+    bool minuteMatch = (st.minute == -1 || tm->tm_min == st.minute ||
+                        (st.minute > 0 && st.minute <= 59 && tm->tm_min % st.minute == 0) ||
+                        (st.minute == 0 && tm->tm_min == 0));
+    bool hourMatch = (st.hour == -1 || tm->tm_hour == st.hour ||
+                      (st.hour > 0 && st.hour <= 23 && tm->tm_hour % st.hour == 0) ||
+                      (st.hour == 0 && tm->tm_hour == 0));
+    bool mdayMatch = (st.mday == -1 || tm->tm_mday == st.mday ||
+                      (st.mday > 0 && st.mday <= 31 && tm->tm_mday % st.mday == 0) ||
+                      (st.mday == 0 && tm->tm_mday == 1));
+    bool monthMatch = (st.month == -1 || tm->tm_mon == st.month ||
+                       (st.month > 0 && st.month <= 12 && (tm->tm_mon - 1) % st.month == 0));
+    bool wdayMatch = (st.wday == -1 || tm->tm_wday == st.wday ||
+                      (st.wday >= 0 && st.wday <= 6 && tm->tm_wday % st.wday == 0));
 
     if (minuteMatch && hourMatch && mdayMatch && monthMatch && wdayMatch) {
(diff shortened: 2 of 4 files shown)


kvrocks.conf
```diff
@@ -515,14 +515,14 @@ compaction-checker-range 0-7
 # force-compact-file-min-deleted-percentage 10
 
 # Bgsave scheduler, auto bgsave at scheduled time
-# time expression format is the same as crontab(currently only support * and int)
+# Time expression format is the same as crontab (currently only support *, int and */n)
```
```suggestion
# Time expression format is the same as crontab (currently only support *, int and */int)
```

src/common/cron.cc
```diff
@@ -58,9 +63,21 @@ bool Cron::IsTimeMatch(const tm *tm) {
     return false;
   }
   for (const auto &st : schedulers_) {
-    if ((st.minute == -1 || tm->tm_min == st.minute) && (st.hour == -1 || tm->tm_hour == st.hour) &&
-        (st.mday == -1 || tm->tm_mday == st.mday) && (st.month == -1 || (tm->tm_mon + 1) == st.month) &&
-        (st.wday == -1 || tm->tm_wday == st.wday)) {
+    bool minuteMatch = (st.minute == -1 || tm->tm_min == st.minute ||
+                        (st.minute > 0 && st.minute <= 59 && tm->tm_min % st.minute == 0) ||
+                        (st.minute == 0 && tm->tm_min == 0));
```
PragmaTwice: `st.minute == 0 && tm->tm_min == 0` seems useless since it's covered in `tm->tm_min == st.minute`.

`st.minute > 0 && st.minute <= 59 && tm->tm_min % st.minute == 0`: from my understanding, I think we need to check `st.minute_interval` to see if it's `true`?

src/common/cron.cc
```diff
@@ -84,20 +101,33 @@ StatusOr<Scheduler> Cron::convertToScheduleTime(const std::string &minute, const
...
+  std::smatch match;
+  if (std::regex_match(param, match, interval_regex)) {
+    int interval = std::stoi(match[1].str());
+    if (interval >= lower_bound && interval <= upper_bound) {
+      is_interval = true;
+      return interval;
+    } else {
+      return {Status::NotOK, "interval value out of bounds"};
+    }
+  }
```
PragmaTwice:
```suggestion
  if (HasPrefix(param, "*/")) {
    auto s = ParseInt<int>(param.substr(2), {lower_bound, upper_bound}, 10);
    if (!s) {
      return std::move(s).Prefixed(fmt::format("malformed cron token `{}`", param));
    }

    is_interval = true;
    return *s;
  }
```

kinoute:
@PragmaTwice Thanks for your suggestions, I updated the code

src/common/cron.cc
```diff
@@ -84,20 +99,30 @@ StatusOr<Scheduler> Cron::convertToScheduleTime(const std::string &minute, const
...
   if (param == "*") {
     return -1;
   }
 
+  // Check for interval syntax (*/n)
+  if (util::HasPrefix(param, "*/")) {
+    auto s = ParseInt<int>(param.substr(2), {lower_bound, upper_bound}, 10);
+    if (!s) {
+      return std::move(s).Prefixed(fmt::format("malformed cron token `{}`", param));
+    }
```
PragmaTwice:
```suggestion
    if (!s || *s == 0) {
      return std::move(s).Prefixed(fmt::format("malformed cron token `{}`", param));
    }
```

src/common/cron.cc
```diff
@@ -58,9 +63,19 @@ bool Cron::IsTimeMatch(const tm *tm) {
     return false;
   }
   for (const auto &st : schedulers_) {
-    if ((st.minute == -1 || tm->tm_min == st.minute) && (st.hour == -1 || tm->tm_hour == st.hour) &&
-        (st.mday == -1 || tm->tm_mday == st.mday) && (st.month == -1 || (tm->tm_mon + 1) == st.month) &&
-        (st.wday == -1 || tm->tm_wday == st.wday)) {
+    bool minute_match = (st.minute == -1 || tm->tm_min == st.minute ||
+                         (st.minute > 0 && st.minute <= 59 && tm->tm_min % st.minute == 0 && st.minute_interval));
```
PragmaTwice:
```suggestion
    auto match = [](int current, int val, bool interval) {
      if (val == -1) return true;
      if (interval) return current % val == 0;
      return val == current;
    };
    bool minute_match = match(tm->tm_min, st.minute, st.minute_interval);
```

src/common/cron.cc
```diff
@@ -58,9 +63,19 @@ bool Cron::IsTimeMatch(const tm *tm) {
...
+    bool hour_match = (st.hour == -1 || tm->tm_hour == st.hour ||
+                       (st.hour > 0 && st.hour <= 23 && tm->tm_hour % st.hour == 0 && st.hour_interval));
+    bool mday_match = (st.mday == -1 || tm->tm_mday == st.mday ||
+                       (st.mday > 0 && st.mday <= 31 && tm->tm_mday % st.mday == 0 && st.mday_interval) ||
+                       (st.mday == 0 && tm->tm_mday == 1));
+    bool month_match = (st.month == -1 || tm->tm_mon == st.month ||
+                        (st.month > 0 && st.month <= 12 && (tm->tm_mon - 1) % st.month == 0 && st.month_interval));
+    bool wday_match = (st.wday == -1 || tm->tm_wday == st.wday ||
+                       (st.wday >= 0 && st.wday <= 6 && tm->tm_wday % st.wday == 0 && st.wday_interval));
```
PragmaTwice: ditto

src/common/cron.cc
PragmaTwice: the integer cannot be 0 here to prevent division by zero.

kinoute:
@PragmaTwice Thanks, updated again the code

src/common/cron.cc
```diff
@@ -57,10 +62,21 @@ bool Cron::IsTimeMatch(const tm *tm) {
...
+  };
+
   for (const auto &st : schedulers_) {
-    if ((st.minute == -1 || tm->tm_min == st.minute) && (st.hour == -1 || tm->tm_hour == st.hour) &&
-        (st.mday == -1 || tm->tm_mday == st.mday) && (st.month == -1 || (tm->tm_mon + 1) == st.month) &&
-        (st.wday == -1 || tm->tm_wday == st.wday)) {
+    bool minute_match = match(tm->tm_min, st.minute, st.minute_interval);
+    bool hour_match = match(tm->tm_hour, st.hour, st.hour_interval);
+    bool mday_match = match(tm->tm_mday, st.mday, st.mday_interval);
+    bool month_match = match(tm->tm_mon - 1, st.month, st.month_interval);
```
PragmaTwice:
```suggestion
    bool month_match = match(tm->tm_mon + 1, st.month, st.month_interval);
```

kinoute: But If I do that, my tests for the month day will fail. So I guess my tests are bad then? According to crontab.guru, a crontab like `0 0 1 */4 *` will result in the following:

```
next at 2024-09-01 00:00:00
then at 2025-01-01 00:00:00
then at 2025-05-01 00:00:00
then at 2025-09-01 00:00:00
then at 2026-01-01 00:00:00
```

https://crontab.guru/#0_0_1_*/4_*

PragmaTwice: For `tm->tm_mon - 1`, the non-interval case will be wrong.

src/common/cron.cc
```diff
@@ -57,10 +62,21 @@ bool Cron::IsTimeMatch(const tm *tm) {
       tm->tm_mon == last_tm_.tm_mon && tm->tm_wday == last_tm_.tm_wday) {
     return false;
   }
+
+  auto match = [](int current, int val, bool interval) {
+    if (val == -1) return true;
+    if (interval) return current % val == 0;
+    return val == current;
+  };
```
PragmaTwice:
```suggestion
  auto match = [](int current, int val, bool interval, int min_val) {
    if (val == -1) return true;
    if (interval) return (current - min_val) % val == 0;
    return val == current;
  };
```

src/common/cron.cc
```diff
@@ -57,10 +62,21 @@ bool Cron::IsTimeMatch(const tm *tm) {
...
+
   for (const auto &st : schedulers_) {
-    if ((st.minute == -1 || tm->tm_min == st.minute) && (st.hour == -1 || tm->tm_hour == st.hour) &&
-        (st.mday == -1 || tm->tm_mday == st.mday) && (st.month == -1 || (tm->tm_mon + 1) == st.month) &&
-        (st.wday == -1 || tm->tm_wday == st.wday)) {
+    bool minute_match = match(tm->tm_min, st.minute, st.minute_interval);
+    bool hour_match = match(tm->tm_hour, st.hour, st.hour_interval);
+    bool mday_match = match(tm->tm_mday, st.mday, st.mday_interval);
+    bool month_match = match(tm->tm_mon - 1, st.month, st.month_interval);
+    bool wday_match = match(tm->tm_wday, st.wday, st.wday_interval);
```
PragmaTwice:
```suggestion
    bool minute_match = match(tm->tm_min, st.minute, st.minute_interval, 0);
    bool hour_match = match(tm->tm_hour, st.hour, st.hour_interval, 0);
    bool mday_match = match(tm->tm_mday, st.mday, st.mday_interval, 1);
    bool month_match = match(tm->tm_mon + 1, st.month, st.month_interval, 1);
    bool wday_match = match(tm->tm_wday, st.wday, st.wday_interval, 0);
```

PragmaTwice: You can try other algorithm.

But be sure to make it correct for all cases.

kinoute:
I added more tests for non interval syntax and renamed the `min_val` parameter

src/common/cron.cc
```diff
@@ -57,10 +62,21 @@ bool Cron::IsTimeMatch(const tm *tm) {
...
+    return val == current;
+  };
+
   for (const auto &st : schedulers_) {
-    if ((st.minute == -1 || tm->tm_min == st.minute) && (st.hour == -1 || tm->tm_hour == st.hour) &&
-        (st.mday == -1 || tm->tm_mday == st.mday) && (st.month == -1 || (tm->tm_mon + 1) == st.month) &&
-        (st.wday == -1 || tm->tm_wday == st.wday)) {
+    bool minute_match = match(tm->tm_min, st.minute, st.minute_interval, 0);
+    bool hour_match = match(tm->tm_hour, st.hour, st.hour_interval, 0);
+    bool mday_match = match(tm->tm_mday, st.mday, st.mday_interval, 0);
```
PragmaTwice:
```suggestion
    bool mday_match = match(tm->tm_mday, st.mday, st.mday_interval, 1);
```

Should it be 1?

kinoute: @PragmaTwice Fixed

PragmaTwice (approved):

git-hulk (approved):

PR #2362, opened 2024-06-11
chore(error): unify error by adding Redis error codes for Status

git-hulk:
This PR introduces new error codes to identify Redis error prefixes,
and forces using status to unify Redis errors. You MUST convert
it to Status to send a Redis error response.


git-hulk:
By the way, I found some error messages aren't aligned with Redis, but I don't prefer fixing them in this PR. Will submit another PR to resolve this, as well as to reduce some of the redundant status codes.

src/cluster/cluster.cc
```diff
@@ -821,26 +821,26 @@ Status Cluster::CanExecByMySelf(const redis::CommandAttributes *attributes, cons
     int cur_slot = GetSlotIdFromKey(cmd_tokens[i]);
     if (slot == -1) slot = cur_slot;
     if (slot != cur_slot) {
-      return {Status::RedisExecErr, "CROSSSLOT Attempted to access keys that don't hash to the same slot"};
+      return {redis::ErrorKind::CrossSlot, "Attempted to access keys that don't hash to the same slot"};
```
torwig: Is it possible to have `using redis::ErrorKind;` to write less code in such places?

git-hulk: @torwig Do you mean to use `use redis::ErrorKind` to avoid the `redis::ErrorKind` prefix?

src/common/redis_error.h
```diff
@@ -0,0 +1,58 @@
...
+ *
+ */
+
+#pragma once
+
+#include <map>
+
+namespace redis {
+
+enum class ErrorKind {
```
caipengbo: Using `RedisError` ?

git-hulk: ErrorKind => RedisError? If yes, perhaps `Error` is better name since it was inside `redis` namespace now.

caipengbo: Yeah, I think the suffix of `Kind` is a bit redundant :)

git-hulk: But it's an error kind instead of error indeed, I prefer keeping this suffix.

caipengbo: OK, got it.

src/cluster/cluster.cc
torwig: @git-hulk Yes, to avoid typing the prefix.

PragmaTwice:
There are several points for the current changes:
- We need to also add ErrorKind to StatusOr, otherwise it's not consistent between Status and StatusOr, which violates our design of StatusOr.
- ErrorKind::None is weird and seems useless. Removing useless state is always a good thing to prevent unnecessary complex.
- ErrorKind is not always useful in Status, since not all sink of Status goes to RESP error.

So I suggest maybe we can just merge ErrorKind to Status::Code, which seems a solution that can solve these issues.

src/cluster/cluster.cc
git-hulk: OK, that makes sense.

git-hulk:
> We need to also add ErrorKind to StatusOr, otherwise it's not consistent between Status and StatusOr, which violates our design of StatusOr.

Sure, would add this.

> ErrorKind::None is weird and seems useless. Removing useless state is always a good thing to prevent unnecessary complex.

None is used to represent the error without the prefix.

> ErrorKind is not always useful in Status, since not all sink of Status goes to RESP error.

Yes, not all status errors can be mapped to the RESP error, and I think it's still important to identify the status error via code. That's why I used the RedisError to identify them.

src/commands/cmd_key.cc
```diff
@@ -508,15 +507,12 @@ class CommandSort : public Commander {
...
-        *output = redis::Error("One or more scores can't be converted into double");
-        break;
+        return {Status::RedisErrorNoPrefix, "One or more scores can't be converted into double"};
       case Database::SortResult::LIMIT_EXCEEDED:
-        *output = redis::Error("The number of elements to be sorted exceeds SORT_LENGTH_LIMIT = " +
-                               std::to_string(SORT_LENGTH_LIMIT));
-        break;
+        return {Status::RedisErrorNoPrefix,
+                "The number of elements to be sorted exceeds SORT_LENGTH_LIMIT = " + std::to_string(SORT_LENGTH_LIMIT)};
       case Database::SortResult::DONE:
```
PragmaTwice, line 516: There's really no prefix? Or just because we forget to add the prefix?

src/commands/cmd_replication.cc
```diff
@@ -230,7 +230,7 @@ class CommandFetchMeta : public Commander {
       std::string files;
       auto s = engine::Storage::ReplDataManager::GetFullReplDataInfo(srv->storage, &files);
       if (!s.IsOK()) {
-        s = util::SockSend(repl_fd, redis::Error("can't create db checkpoint"), bev);
+        s = util::SockSend(repl_fd, redis::Error({Status::RedisErrorNoPrefix, "can't create db checkpoint"}), bev);
```
PragmaTwice, line 233: If here it's because we forget to add the prefix, we should fix it.

src/commands/cmd_server.cc
```diff
@@ -630,9 +630,9 @@ class CommandDebug : public Commander {
       } else if (protocol_type_ == "verbatim") {  // verbatim string
         *output = conn->VerbatimString("txt", "verbatim string");
       } else {
-        *output = redis::Error(
-            "Wrong protocol type name. Please use one of the following: "
-            "string|integer|double|array|set|bignum|true|false|null|attrib|verbatim");
+        return {Status::RedisErrorNoPrefix,
+                "Wrong protocol type name. Please use one of the following: "
```
PragmaTwice, line 634: Ditto

src/commands/error_constants.h
```diff
@@ -40,10 +40,10 @@ inline constexpr const char *errLimitOptionNotAllowed =
 inline constexpr const char *errZSetLTGTNX = "GT, LT, and/or NX options at the same time are not compatible";
 inline constexpr const char *errScoreIsNotValidFloat = "score is not a valid float";
 inline constexpr const char *errValueIsNotFloat = "value is not a valid float";
-inline constexpr const char *errNoMatchingScript = "NOSCRIPT No matching script. Please use EVAL";
+inline constexpr const char *errNoMatchingScript = "No matching script. Please use EVAL";
 inline constexpr const char *errUnknownOption = "unknown option";
 inline constexpr const char *errUnknownSubcommandOrWrongArguments = "Unknown subcommand or wrong number of arguments";
 inline constexpr const char *errWrongNumArguments = "ERR wrong number of arguments";
```
PragmaTwice: How about this one?

PragmaTwice:
https://github.com/apache/kvrocks/blob/84a3559be87ef623d9a2cbca51018be7218835ce/src/cluster/replication.cc#L1003

Forget to change?

src/commands/cmd_key.cc
git-hulk: Not sure about this. I found we have some places that didn't align with Redis, but I suggest revising and fixing them in another PR: https://github.com/apache/kvrocks/pull/2362#issuecomment-2165834533

src/commands/cmd_replication.cc
git-hulk: Same as the above comment.

git-hulk:
> By the way, I found some error messages aren't aligned with Redis, but I don't prefer fixing them in this PR. Will submit another PR to resolve this, as well as to reduce some of the redundant status codes.



> errRestoringBackup

Yes, good catch, fixed now.

src/cluster/replication.cc
```diff
@@ -1000,15 +1000,15 @@ Status ReplicationThread::parseWriteBatch(const std::string &batch_string) {
 }
 
 bool ReplicationThread::isRestoringError(std::string_view err) {
-  return err == std::string(RESP_PREFIX_ERROR) + redis::errRestoringBackup;
+  return err == redis::Error({Status::RedisLoading, redis::errRestoringBackup});
```
```suggestion
  return err == std::string(RESP_PREFIX_ERROR) + redis::StatusToRedisError({Status::RedisLoading, redis::errRestoringBackup});
```

git-hulk: It's the same as Redis::Error

PragmaTwice: Ahh, let's remove CRLF in StatusToRedisError.

PragmaTwice: CRLF should not be in StatusToRedisError. It will make the semantics of StatusToRedisError very weird.

src/server/redis_reply.cc
```diff
@@ -20,15 +20,37 @@
...
 namespace redis {
 
 void Reply(evbuffer *output, const std::string &data) { evbuffer_add(output, data.c_str(), data.length()); }
 
 std::string SimpleString(const std::string &data) { return "+" + data + CRLF; }
 
-std::string Error(const std::string &err) { return "-" + err + CRLF; }
+std::string Error(const Status &s) { return RESP_PREFIX_ERROR + StatusToRedisError(s); }
+
+std::string StatusToRedisError(const Status &s) {
```
PragmaTwice:
```suggestion
std::string StatusToRedisErrorMsg(const Status &s) {
```
Maybe this name is better.

PragmaTwice: Then we can distinguish between redis error message and resp errors.
- redis error message -> `<PREFIX> <message..>`
- resp error -> `- <redis error message> CRLF`

git-hulk: done

src/cluster/replication.cc
git-hulk: done

PragmaTwice (approved):

Issue #2329, opened 2024-05-25
A new version of search key and metadata encoding


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

AS IS: (InernalKey encoding)
(`key` is omitted, actually it's InternalKey `namespace | (slot num) | index name`)
```
search data type metadata:

key -> metadata flag | expire | version | size | on_data_type (HASH or JSON)

prefixes encoding:

key | version | PREFIXES -> prefix num | prefix1 prefix2 ...

tag field metadata encoding:

key | version | TAG_FIELD_META | field name -> field flag | separator | case sensitive

tag field index encoding:

key | version | TAG_FIELD | field name | tag | key -> (nil)

numeric field metadata encoding:

key | version | NUM_FIELD_META | field name -> field flag

numeric field index encoding:

key | version | NUM_FIELD | field name | ordered floating number | key -> (nil)

field flag:

|              8 bit              |
|---------------------------------|
|  noindex: 1bit | reserved: 7bit |
```

TO BE: (new encoding)
```
search data type metadata:

ns | INDEX_META | index name ->  index flag | on_data_type (HASH or JSON)

prefixes encoding:

ns | PREFIXES | index name -> prefix num | prefix1 prefix2 ...

field alias encoding:

ns | FIELD_ALIAS | index name | alias name -> field name

tag field metadata encoding:

ns | FIELD_META | index name | field name -> field flag | separator | case sensitive

numeric field metadata encoding:

ns | FIELD_META | index name | field name -> field flag

tag field index encoding:

ns | FIELD | index name | field name | tag | key -> (nil)

numeric field index encoding:

ns | FIELD | index name | field name | ordered floating number | key -> (nil)

field flag:

|              8 bit                                 |
|----------------------------------------------------|
|  noindex: 1bit | field type: 4bit | reserved: 3bit |

field type:
TAG     1
NUMERIC 2
```

benefits:
- we can iterate all index metadata via a rocksdb iterator
- remove useless fields in index metadata
- field name is unique regardless of field type
- prefix compression can work better
- field type stored in metadata instead of key

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #2368, opened 2024-06-18
feat(search): Add HNSW encoding index & insertion/deletion algorithm

Beihao-Zhou:
Implement Proposal at #2316 

## Encoding
```
HNSW vector field metadata encoding:

ns | FIELD_META | index name | field name -> field flag | vector_type | dimension | distance_metric | initial_cap | m | ef_construction | ef_runtime | epsilon | num_levels

HNSW node index encoding:

ns | FIELD | index name | field name | level | NODE | key -> num_neighbours | vector dimension | [vector...]

HNSW edge index encoding:

ns | FIELD | index name | field name | level | EDGE |  key1 | key2 -> (nil)
```
Reference for other index encoding: #2329

## Future steps

- [ ] Add the plan operator and the corresponding executor
- [ ] Add expression node (i.e. SQL/RediSearch parsers) for vector search
- [ ] Modify some passes (eg. index_selection) to convert the expression node to plan operator
- [ ] Improve `HnswIndex` construction (#2398)


src/search/hnsw_indexer.h
```diff
@@ -0,0 +1,395 @@
...
+
+    while (std::getline(iss, num, ',')) {
+      try {
+        double value = std::stod(num);
+        vector.push_back(value);
+      } catch (const std::invalid_argument& ia) {
+        throw std::runtime_error("Invalid number in vector string: " + num);
+      }
+    }
+  }
```
PragmaTwice: It appears that our main issue here is not implementing a value system in KQIR, but relying on strings to represent values of all field types. 

Initially, I avoided introducing it for quick implementation at the outset. However, it now seems essential.

In HASH type vector fields, data is encoded in binary format; whereas in JSON, it is stored as an array of numbers. (https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/vectors/#storing-vectors-in-hashes)

I'll work on this later.

src/search/hnsw_indexer.h
```diff
@@ -0,0 +1,395 @@
...
+      for (auto i = 0; i < dim; i++) {
+        dist += left.vector[i] * right.vector[i];
+        norma += left.vector[i] * right.vector[i];
+        normb += left.vector[i] * right.vector[i];
+      }
+      auto similarity = dist / std::sqrt(norma * normb);
+      return 1.0 - similarity;
+    }
+    default:
+      // throw error
```
PragmaTwice: In Kvrocks we don't use exception, instead you can use `StatusOr` or `Status` for error handling.

You can refer to other source files like `index.cc` for how to use them.

src/search/hnsw_indexer.h
Beihao-Zhou: I've been exploring the possibility of storing distances in metadata to enhance computation efficiency, but given that these distances are always calculated between random vectors, I haven't yet discovered an effective work-around to reduce the number of compute. 

Thanks for pointing out the different data formats! If that's the case, I'm wondering if adopting the binary format to encode the vector for both JSON and HASH would be a good idea? This can save space and improve performance compared to storing them as raw string for now, but has some overhead when the vector is passed in as an array in JSON. Let me know if you prefer this solution or not :))

PragmaTwice: Now `kqir::Value` is available, and `IndexUpdater::UpdateXXXIndex` can take `kqir::Value` as input.

So you can:
- Add a new template argument to the base class (`std::variant`) of `kqir::Value` (e.g. `NumericArray`)
- Modify `FieldValueRetriever::Retrieve` to handle the case while the input field type is `HnswVectorFieldMetadata`
- Use it without parsing (you now can get `std::vector<double>` directly from `kqir::Value`)

For encoding part, we cannot change the existing encoding of JSON and HASH data types. But you can store the vector in binary format in the HNSW index encoding.

https://github.com/apache/kvrocks/blob/unstable/src/search/value.h

PragmaTwice: - Add a new template argument to the base class (std::variant) of kqir::Value (e.g. NumericArray)
- Modify FieldValueRetriever::Retrieve to handle the case while the input field type is HnswVectorFieldMetadata

Done in #2371.

src/search/search_encoding.h
```diff
@@ -76,6 +77,24 @@ enum class IndexFieldType : uint8_t {
   TAG = 1,
 
   NUMERIC = 2,
+
+  VECTOR = 3,
+};
+
+enum class VectorType : uint8_t {
+  FLOAT32 = 0,
+  FLOAT64 = 1,
```
PragmaTwice:
```suggestion
  FLOAT64 = 0,
```

maybe we can remove float32 if it's currently not supported.

Beihao-Zhou:
Hi @PragmaTwice , this PR is ready for review! :)

src/search/search_encoding.h
```diff
@@ -291,6 +359,92 @@ struct NumericFieldMetadata : IndexFieldMetadata {
...
+    }
+
+    GetFixed8(input, (uint8_t *)(&vector_type));
+    GetFixed16(input, &dim);
+    GetFixed8(input, (uint8_t *)(&distance_metric));
+    GetFixed32(input, &initial_cap);
+    GetFixed16(input, &m);
+    GetFixed32(input, &ef_construction);
+    GetFixed32(input, &ef_runtime);
+    GetDouble(input, (double *)(&epsilon));
```
PragmaTwice:
```suggestion
    GetDouble(input, &epsilon);
```

Do we need to cast it here?

src/search/indexer.cc
```diff
@@ -246,7 +243,7 @@ Status IndexUpdater::UpdateTagIndex(std::string_view key, const kqir::Value &ori
 Status IndexUpdater::UpdateNumericIndex(std::string_view key, const kqir::Value &original, const kqir::Value &current,
                                         const SearchKey &search_key, const NumericFieldMetadata *num) const {
   CHECK(original.IsNull() || original.Is<kqir::Numeric>());
-  CHECK(original.IsNull() || original.Is<kqir::Numeric>());
+  CHECK(current.IsNull() || current.Is<kqir::Numeric>());
```
PragmaTwice, line 246: Good catch!

src/search/hnsw_indexer.h
```diff
@@ -0,0 +1,107 @@
...
+  StatusOr<HnswNodeFieldMetadata> DecodeMetadata(const SearchKey& search_key, engine::Storage* storage) const;
+  void PutMetadata(HnswNodeFieldMetadata* node_meta, const SearchKey& search_key, engine::Storage* storage,
+                   ObserverOrUniquePtr<rocksdb::WriteBatchBase>& batch) const;
+  void DecodeNeighbours(const SearchKey& search_key, engine::Storage* storage);
+
+  // For testing purpose
+  Status AddNeighbour(const NodeKey& neighbour_key, const SearchKey& search_key, engine::Storage* storage,
+                      ObserverOrUniquePtr<rocksdb::WriteBatchBase>& batch) const;
+  Status RemoveNeighbour(const NodeKey& neighbour_key, const SearchKey& search_key, engine::Storage* storage,
+                         ObserverOrUniquePtr<rocksdb::WriteBatchBase>& batch) const;
```
PragmaTwice: Maybe we can just use `rocksdb::WriteBatchBase *` here instead of `ObserverOrUniquePtr<rocksdb::WriteBatchBase>` since we don't need to care the ownership.

src/search/hnsw_indexer.cc
```diff
@@ -0,0 +1,514 @@
...
+
+        // Re-evaluate the neighbours for the candidate node
+        candidate_node.DecodeNeighbours(search_key, storage);
+        auto candidate_node_neighbour_vec_items =
+            GET_OR_RET(DecodeNodesToVectorItems(candidate_node.neighbours, level, search_key, storage, metadata));
+        candidate_node_neighbour_vec_items.push_back(inserted_vector_item);
+        auto sorted_neighbours = GET_OR_RET(SelectNeighbors(candidate_vec, candidate_node_neighbour_vec_items, level));
+
+        bool inserted_node_is_selected = std::find(sorted_neighbours.begin(), sorted_neighbours.end(),
+                                                   inserted_vector_item) != sorted_neighbours.end();
```
PragmaTwice: If it's sorted maybe we can use `std::binary_search` here.

src/search/hnsw_indexer.h
```diff
@@ -0,0 +1,107 @@
...
+#include "search/indexer.h"
+#include "search/search_encoding.h"
+#include "search/value.h"
+#include "storage/storage.h"
+
+namespace redis {
+
+class HnswIndex;
+
+struct Node {
```
PragmaTwice: `Node` seems to generic. I think we can rename it to something like `HnswNode`.

src/search/hnsw_indexer.cc
```diff
@@ -0,0 +1,514 @@
...
+    auto entry_point_vector = VectorItem(entry_point_key, std::move(entry_node_metadata.vector), metadata);
+    auto dist = GET_OR_RET(ComputeSimilarity(target_vector, entry_point_vector));
+
+    explore_heap.push(std::make_pair(dist, entry_point_vector));
+    result_heap.push(std::make_pair(dist, std::move(entry_point_vector)));
+    visited.insert(entry_point_key);
+  }
+
+  while (!explore_heap.empty()) {
+    auto [dist, current_vector] = explore_heap.top();
```
PragmaTwice, line 302:
```suggestion
    auto [dist, current_vector] = std::move(explore_heap.top());
```

src/search/indexer.cc
```diff
@@ -269,6 +266,32 @@ Status IndexUpdater::UpdateNumericIndex(std::string_view key, const kqir::Value
...
 }
 
+Status IndexUpdater::UpdateHnswVectorIndex(std::string_view key, const kqir::Value &original,
+                                           const kqir::Value &current, const SearchKey &search_key,
+                                           HnswVectorFieldMetadata *vector) const {
+  CHECK(original.IsNull() || original.Is<kqir::NumericArray>());
+  CHECK(current.IsNull() || current.Is<kqir::NumericArray>());
+
+  auto storage = indexer->storage;
+  auto hnsw = HnswIndex(search_key, vector, storage);
```
PragmaTwice, line 276: HnswIndex is a relatively heavy class (since it includes `mt19937`, `sizeof(mt19937) ~= 5000`), so I'm wondering if we can optimize to avoid constructing it every time.

Although we don't need to optimize in this PR.

PragmaTwice (commented):
The rest looks good to me. Thank you!

Also there are still some clang-tidy issues that need to be fixed.

src/search/hnsw_indexer.cc
```diff
@@ -0,0 +1,514 @@
...
+StatusOr<std::vector<VectorItem>> HnswIndex::DecodeNodesToVectorItems(const std::vector<NodeKey>& node_keys,
+                                                                      uint16_t level, const SearchKey& search_key,
+                                                                      engine::Storage* storage,
+                                                                      const HnswVectorFieldMetadata* metadata) {
+  std::vector<VectorItem> vector_items;
+  vector_items.reserve(node_keys.size());
+
+  for (const auto& neighbour_key : node_keys) {
+    Node neighbour_node(neighbour_key, level);
+    auto neighbour_metadata = GET_OR_RET(neighbour_node.DecodeMetadata(search_key, storage));
```
git-hulk: should we ignore the neighbour node if it's not found?

src/search/search_encoding.h
Beihao-Zhou: Great catch! Changed :)

src/search/hnsw_indexer.h
Beihao-Zhou: Double-checked the `ObserverOrUniquePtr` class and agree, code changed 👍

src/search/hnsw_indexer.cc
Beihao-Zhou: `SelectNeighbors` returns `sorted_neighbours` (`std::vector<VectorItem>`) based on distance, while `VectorItem` overloads `operator<` based on its `key`. So I guess `std::binary_search` wouldn't work here since `sorted_neighbours` is sorted by distance instead of keys. 

I updated `sorted_neighbours` into `sorted_neighbours_by_distance` to make it more clear. Let me know if I misunderstand anything here :)

src/search/hnsw_indexer.h
Beihao-Zhou: Done :)

src/search/hnsw_indexer.cc
Beihao-Zhou: Done :)

Beihao-Zhou: `clang-tidy` throws no-effect error, change it back :(
Ref: https://github.com/Beihao-Zhou/kvrocks/actions/runs/9862851656/job/27234342104#step:7:1391

src/search/hnsw_indexer.cc
Beihao-Zhou: I agree, it's probably more robust. Code updated :)

src/search/indexer.cc
Beihao-Zhou: Ahh I wasn't aware of that, and thanks for pointing it out!

One way I could think of to avoid heavy construction is to make all instances of HnswIndex share a single `mt19937` instance. 

We can implement a `RandomGenerator` using Singleton Pattern as below:

```
class RandomGenerator {
public:
    static RandomGenerator& getInstance() {
        static RandomGenerator instance;
        return instance;
    }

    std::mt19937& getGenerator() {
        return generator_;
    }

private:
    RandomGenerator() {
        std::random_device rd;
        generator_ = std::mt19937(rd());
    }
    // delete copy & move constructor
    std::mt19937 generator_;
};

struct HnswIndex {
    HnswIndex(...) {
        generator_ = &RandomGenerator::getInstance().getGenerator();
    }
};
```

I'll open an issue for this, and see if anyone has better solutions. :)

src/search/hnsw_indexer.cc
PragmaTwice: Hmm seems here the return type of `top` is const reference so it cannot be moved.

src/search/indexer.cc
PragmaTwice: Yeah we can make it a static member. Maybe we can remove the mt19937 member in every instance since `move` cannot work on stack but only heap memory.

For static local variables I don't prefer them because they will introduce some mutexes.

PragmaTwice (commented):
The code looks fine to me.

But there are some issues in CI:
- one unit test case failed in macOS arm64,
- some memory issues (likely use-after-free) reported by ASan/TSan.

Could you try to investigate them?

src/search/search_encoding.h
```diff
@@ -291,6 +359,92 @@ struct NumericFieldMetadata : IndexFieldMetadata {
   bool IsSortable() const override { return true; }
 };
 
+struct HnswVectorFieldMetadata : IndexFieldMetadata {
+  VectorType vector_type;
+  uint16_t dim;
+  DistanceMetric distance_metric;
+
+  uint32_t initial_cap = 500000;
+  uint16_t m = 16;
```
Yangsx-1: It's not a good variable name.

Yangsx-1 (commented):
I wonder if there is any heuristic logic to avoid isolated cluster?

src/search/hnsw_indexer.h
```diff
@@ -0,0 +1,107 @@
...
+#include "storage/storage.h"
+
+namespace redis {
+
+class HnswIndex;
+
+struct HnswNode {
+  using NodeKey = std::string;
+  NodeKey key;
+  uint16_t level;
```
mapleFU:
```suggestion
  uint16_t level{};
```

src/search/search_encoding.h
```diff
@@ -291,6 +359,92 @@ struct NumericFieldMetadata : IndexFieldMetadata {
...
+    GetFixed32(input, &ef_runtime);
+    GetDouble(input, &epsilon);
+    GetFixed16(input, &num_levels);
+    return rocksdb::Status::OK();
+  }
+};
+
+struct HnswNodeFieldMetadata {
+  uint16_t num_neighbours;
+  std::vector<double> vector;
```
mapleFU, line 417: Just a minor question, would we support more types, like fp16, fp32 in the future? And do we need considering some forward compatibility?

src/search/search_encoding.h
```diff
@@ -291,6 +359,92 @@ struct NumericFieldMetadata : IndexFieldMetadata {
...
+    PutDouble(dst, epsilon);
+    PutFixed16(dst, num_levels);
+  }
+
+  rocksdb::Status Decode(Slice *input) override {
+    if (auto s = IndexFieldMetadata::Decode(input); !s.ok()) {
+      return s;
+    }
+
+    if (input->size() < 1 + 2 + 1 + 4 + 2 + 4 + 4 + 8 + 2) {
```
mapleFU: Emmm this is a bit hacking...

src/search/hnsw_indexer.cc
```diff
@@ -0,0 +1,521 @@
...
+
+  auto has_other_nodes_at_level = [&](uint16_t level, std::string_view skip_key) -> bool {
+    auto prefix = search_key.ConstructHnswLevelNodePrefix(level);
+    util::UniqueIterator it(storage, storage->DefaultScanOptions(), ColumnFamilyID::Search);
+    it->Seek(prefix);
+
+    Slice node_key;
+    Slice node_key_dst;
+    while (it->Valid() && it->key().starts_with(prefix)) {
+      node_key = Slice(it->key().ToString().substr(prefix.size()));
```
PragmaTwice: `it->key().ToString().substr(prefix.size())` is a temporary string object and will be destroyed after this line, and then `node_key` will be dangling.

That's probably the reason of the CI failure.

src/search/hnsw_indexer.cc
```diff
@@ -0,0 +1,521 @@
...
+
+StatusOr<HnswIndex::NodeKey> HnswIndex::DefaultEntryPoint(uint16_t level) const {
+  auto prefix = search_key.ConstructHnswLevelNodePrefix(level);
+  util::UniqueIterator it(storage, storage->DefaultScanOptions(), ColumnFamilyID::Search);
+  it->Seek(prefix);
+
+  Slice node_key;
+  Slice node_key_dst;
+  if (it->Valid() && it->key().starts_with(prefix)) {
+    node_key = Slice(it->key().ToString().substr(prefix.size()));
```
PragmaTwice: ditto

src/search/hnsw_indexer.cc
Beihao-Zhou: Thank you soooo much!! Really saved me from the persisting problem!! Will make all changes tmr :))

src/search/indexer.cc
Beihao-Zhou: Gotta, then we could potentially have something like
```
#include <random>

class HnswIndex {
// ...
private:
    // ...
    static std::mt19937 generator_;  // Static member variable
};

std::mt19937 HnswIndex::generator_{std::random_device{}()};
```

src/search/search_encoding.h
Beihao-Zhou: These metadata variable names are the same as [redis protocol](https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/vectors/). 

Also when browsing [related articles](https://www.pinecone.io/learn/series/faiss/hnsw/) regarding HNSW algorithm, it seems like it's a convention to use `m` or `m_L` referred to `max_num_outgoing_edges`. 

I feel like it might be good to be compatible with the protocol & convention, but still add comments beside them. Let me know if you have better ideas on how we can name it! <3

Beihao-Zhou:
> I wonder if there is any heuristic logic to avoid isolated cluster?

@Yangsx-1 Good question but not yet, the plan for the PR is just to implement the hnsw construction first so didn't think ahead that much. Also I saw KQIR cannot be enabled on cluster mode yet[1], so also didn't take this into the scope of this PR. 

[1] [KQIR: a query engine for Apache Kvrocks that supports both SQL and RediSearch queries](https://kvrocks.apache.org/blog/kqir-query-engine/#relationship-with-other-features)

src/search/search_encoding.h
Beihao-Zhou: Yeah, we want to support more data types like fp16 and fp32 in the future, just like Redis does.

To handle different data types, each `HnswIndex` has a consistent data type managed by a `VectorType` field in `HnswVectorFieldMetadata`. We can pass this field to `HnswNodeFieldMetadata` to handle encoding and decoding for different types.

Here might be how we can tweak the `HnswNodeFieldMetadata` to support different data types:

```
struct HnswNodeFieldMetadata {
    // ...
    VectorType vector_type;
    std::vector<double> vector; // Store different float types as double (probably have better solutions in the future)

    void Encode(std::string *dst) const {
        // ...
        for (double element : vector) {
            switch (vector_type) {
                case VectorType::FLOAT16:
                    PutFloat16(dst, static_cast<float16_t>(element)); // Implement PutFloat16 for float16
                    break;
                case VectorType::FLOAT32:
                    PutFloat32(dst, static_cast<float32_t>(element));
                    break;
                case VectorType::FLOAT64:
                    PutDouble(dst, element);
                    break;
            }
        }
    }
    
    // Similar as Decode()
}
```

There is definitely room for improvement (e.g., static dispatch), but this gives an idea of how we can ensure encode/decode correctness with RocksDB.

src/search/search_encoding.h
Beihao-Zhou: Hmm I changed it as below, no quite sure if that's the best solution. Let me know if you got better idea!
```
    constexpr size_t required_size = sizeof(uint8_t) + sizeof(uint16_t) + sizeof(uint8_t) + sizeof(uint32_t) +
                                     sizeof(uint16_t) + sizeof(uint32_t) + sizeof(uint32_t) + sizeof(uint64_t) +
                                     sizeof(uint16_t);

    if (input->size() < required_size)
      //...
```

Beihao-Zhou:
> The code looks fine to me.
> 
> But there are some issues in CI:
> 
> * one unit test case failed in macOS arm64,
> * some memory issues (likely use-after-free) reported by ASan/TSan.
> 
> Could you try to investigate them?

@git-hulk @PragmaTwice 
The issue was caused by the `ComputeSimilarity` calculates the distance between `VectorItem` based on the `HnswVectorMetadata`. In the unit test, I initialized `VectorItem` where its `vector` size less than `metadata->dim`, so looping through the vector causes memory leak. 

I changed the code with one `VectorItem::Create` to do this validation early. Let me know if this still looks good to you <3

CR: https://github.com/apache/kvrocks/pull/2368/commits/224141f8b3eea9f07a5cc2ce7fd5c2027824aa72
Successful workflow: https://github.com/Beihao-Zhou/kvrocks/actions/runs/9914854795

PragmaTwice (approved):

PragmaTwice:
Awesome. Thank you for your contribution!

PragmaTwice:
Hi @Beihao-Zhou , could you also open a tracking issue to track all issues and PRs for vector search in Kvrocks?

Beihao-Zhou:
> Hi @Beihao-Zhou , could you also open a tracking issue to track all issues and PRs for vector search in Kvrocks?

Sure, will do that later today <3

PR #2373, opened 2024-06-20
feat(stream): add support of the `XAUTOCLAIM` command

### LINKED ISSUE #2181: Add support of the XAUTOCLAIM command
### Search before asking

### Motivation

This issue wants to implement [XAUTOCLAIM](https://redis.io/commands/xautoclaim/) command like redis.

### Solution

_No response_

LindaSummer:
# Issue
close #2181

# Proposed Changes
- implement `XAUTOCLAIM` command
- add test cases following [Redis tests](https://github.com/redis/redis/blob/7.4/tests/unit/type/stream-cgroups.tcl#L692)

~~# Comment~~
~~1. go Redis client [checks `count` parameter](https://github.com/redis/go-redis/blob/v9.5.3/stream_commands.go#L330) before passing to server, so `count == 0` test case can't be created in go test.~~
~~2. go Redis client has [no delete ids](https://github.com/redis/go-redis/blob/v9/command.go#L1741) in response, so related test case can't be created.~~

### DIFF
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -367,11 +367,11 @@ class CommandAutoClaim : public Commander {
     key_name_ = GET_OR_RET(parser.TakeStr());
     group_name_ = GET_OR_RET(parser.TakeStr());
     consumer_name_ = GET_OR_RET(parser.TakeStr());
-    auto parse_int_status = parser.TakeInt<uint64_t>();
-    if (!parse_int_status.IsOK()) {
+    if (auto parse_status = parser.TakeInt<uint64_t>(); !parse_status.IsOK()) {
       return {Status::RedisParseErr, "Invalid min-idle-time argument for XAUTOCLAIM"};
+    } else {
+      options_.min_idle_time_ms = parse_status.GetValue();
     }
-    options_.min_idle_time_ms = parse_int_status.GetValue();
 
     auto start_str = GET_OR_RET(parser.TakeStr());
     if (!start_str.empty() && start_str.front() == '(') {
@@ -381,9 +381,9 @@ class CommandAutoClaim : public Commander {
     if (!options_.exclude_start && start_str == "-") {
       options_.start_id = StreamEntryID::Minimum();
     } else {
-      auto parse_range_status = ParseRangeStart(start_str, &options_.start_id);
-      if (!parse_range_status.IsOK()) {
-        return parse_range_status;
+      auto parse_status = ParseRangeStart(start_str, &options_.start_id);
+      if (!parse_status.IsOK()) {
+        return parse_status;
       }
     }
 
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -367,26 +367,23 @@ class CommandAutoClaim : public Commander {
     key_name_ = GET_OR_RET(parser.TakeStr());
     group_name_ = GET_OR_RET(parser.TakeStr());
     consumer_name_ = GET_OR_RET(parser.TakeStr());
-    {
-      auto s = parser.TakeInt<uint64_t>();
-      if (!s.IsOK()) {
-        return {Status::RedisParseErr, "Invalid min-idle-time argument for XAUTOCLAIM"};
-      }
-      options_.min_idle_time_ms = s.GetValue();
+    auto parse_int_status = parser.TakeInt<uint64_t>();
+    if (!parse_int_status.IsOK()) {
+      return {Status::RedisParseErr, "Invalid min-idle-time argument for XAUTOCLAIM"};
     }
-    {
-      auto start_str = GET_OR_RET(parser.TakeStr());
-      if (!start_str.empty() && start_str.front() == '(') {
-        options_.exclude_start = true;
-        start_str = start_str.substr(1);
-      }
-      if (!options_.exclude_start && start_str == "-") {
-        options_.start_id = StreamEntryID::Minimum();
-      } else {
-        auto s = ParseRangeStart(start_str, &options_.start_id);
-        if (!s.IsOK()) {
-          return s;
-        }
+    options_.min_idle_time_ms = parse_int_status.GetValue();
+
+    auto start_str = GET_OR_RET(parser.TakeStr());
+    if (!start_str.empty() && start_str.front() == '(') {
+      options_.exclude_start = true;
+      start_str = start_str.substr(1);
+    }
+    if (!options_.exclude_start && start_str == "-") {
+      options_.start_id = StreamEntryID::Minimum();
+    } else {
+      auto parse_range_status = ParseRangeStart(start_str, &options_.start_id);
+      if (!parse_range_status.IsOK()) {
+        return parse_range_status;
       }
     }
 
@@ -429,7 +426,6 @@ class CommandAutoClaim : public Commander {
     output->append(redis::MultiLen(result.entries.size()));
     for (const auto &item : result.entries) {
       if (options_.just_id) {
-        // output->append(redis::MultiLen(1));
         output->append(redis::BulkString(item.key));
       } else {
         output->append(redis::MultiLen(2));
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -18,6 +18,7 @@
  *
  */
 
+#include <algorithm>
 #include <limits>
 #include <memory>
 #include <stdexcept>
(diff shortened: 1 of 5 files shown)


PragmaTwice:
> go Redis client checks count parameter before passing to server, so count == 0 test case can't be created in go test.
> go Redis client has no delete ids in response, so related test case can't be created.

Maybe we can use `redis.Do(ctx, "XAUTOCLAIM", ...)` to create test cases.

LindaSummer:
> > go Redis client checks count parameter before passing to server, so count == 0 test case can't be created in go test.
> > go Redis client has no delete ids in response, so related test case can't be created.
> 
> Maybe we can use `redis.Do(ctx, "XAUTOCLAIM", ...)` to create test cases.

Thanks for your suggestions. Got it! I will add missing cases and fix CI issues today. 😊

LindaSummer:
Hi @PragmaTwice ,
  Thanks for your suggestions.
  I have added missing test cases and the missing header `algorithm` in Darwin build.

  Plus, I find that there is warning during Darwin Clang build.
  Maybe we should make a new PR to fix it.
```
/Users/runner/work/kvrocks/kvrocks/src/commands/cmd_stream.cc:145:36: warning: local variable 'result' will be copied despite being returned by name [-Wreturn-std-move]
        if (!result.IsOK()) return result;
                                   ^~~~~~
/Users/runner/work/kvrocks/kvrocks/src/commands/cmd_stream.cc:145:36: note: call 'std::move' explicitly to avoid copying
        if (!result.IsOK()) return result;
                                   ^~~~~~
                                   std::move(result)
```

LindaSummer:
Hi @PragmaTwice ,

I find that I can run GitHub Actions in my own repo first.
I have run CI in my own repo and fix all build issues now.

Thanks very much.

Best Regards,
Edward

src/commands/cmd_stream.cc
```diff
@@ -358,6 +360,101 @@ class CommandXClaim : public Commander {
...
 };
 
+class CommandAutoClaim : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 1);
+    key_name_ = GET_OR_RET(parser.TakeStr());
+    group_name_ = GET_OR_RET(parser.TakeStr());
+    consumer_name_ = GET_OR_RET(parser.TakeStr());
+    {
```
torwig: Why?

src/commands/cmd_stream.cc
```diff
@@ -358,6 +360,101 @@ class CommandXClaim : public Commander {
...
+      auto s = parser.TakeInt<uint64_t>();
+      if (!s.IsOK()) {
+        return {Status::RedisParseErr, "Invalid min-idle-time argument for XAUTOCLAIM"};
+      }
+      options_.min_idle_time_ms = s.GetValue();
+    }
+    {
```
torwig: Why? What is the benefit of this additional indentation? If you have this extra block, maybe you need to extract it to a separate function. Otherwise, I think it looks weird and shifts the code to the right for no reason, adding cognitive complexity.

src/commands/cmd_stream.cc
```diff
@@ -358,6 +360,101 @@ class CommandXClaim : public Commander {
...
+  }
+
+ private:
+  Status sendResults(Connection *conn, const StreamAutoClaimResult &result, std::string *output) const {
+    output->append(redis::MultiLen(3));
+    output->append(redis::BulkString(result.next_claim_id));
+    output->append(redis::MultiLen(result.entries.size()));
+    for (const auto &item : result.entries) {
+      if (options_.just_id) {
+        // output->append(redis::MultiLen(1));
```
torwig: This commented line can be removed.

src/types/redis_stream.cc
```diff
@@ -511,6 +511,162 @@ rocksdb::Status Stream::ClaimPelEntries(const Slice &stream_name, const std::str
...
+  if (has_next_entry) {
+    std::string tmp_group_name;
+    StreamEntryID entry_id = groupAndEntryIdFromPelInternalKey(iter->key(), tmp_group_name);
+    result->next_claim_id = entry_id.ToString();
+  } else {
+    result->next_claim_id = StreamEntryID::Minimum().ToString();
+  }
+
+  result->entries = std::move(pending_entries);
+  result->deleted_ids.clear();
```
torwig, line 666: Can we reserve here the expected number of elements that will be appended?

src/commands/cmd_stream.cc
LindaSummer: Hi @torwig ,

The result `s` in this block is type of `StatusOr<uint64t>` and the next block's `s` is type of `Status`.
I think they should both represent a “status” meaning and can share one variable name.
So, I chose to use a new sub block to limit the scope of `s`.

I will use two precise variable names rather than sharing same variable name with shadow style.

Best Regards,
Edward

src/commands/cmd_stream.cc
LindaSummer: Hi @torwig ,

The reason of using a new block is for sharing same variable name with the above scope.
I will change these two variables into two precise names to reduce the redundant block.

Best Regards,
Edward

src/commands/cmd_stream.cc
LindaSummer: Hi @torwig ,

Sorry for missing clean up this code in self-checking.
I will remove them.

Best Regards,
Edward

src/types/redis_stream.cc
LindaSummer: Hi @torwig ,

Of course. I will update it and reserve the capacity first then do the `delete_ids` generation.

Best Regards,
Edward

LindaSummer:
Hi @torwig ,

I have updated related code.
Please take a look.
Thanks very much.

Best Regards,
Edward

src/commands/cmd_stream.cc
```diff
@@ -358,6 +360,97 @@ class CommandXClaim : public Commander {
...
 };
 
+class CommandAutoClaim : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 1);
+    key_name_ = GET_OR_RET(parser.TakeStr());
+    group_name_ = GET_OR_RET(parser.TakeStr());
+    consumer_name_ = GET_OR_RET(parser.TakeStr());
+    auto parse_int_status = parser.TakeInt<uint64_t>();
```
torwig: I got your point, it's a valid reason. Having two variables is also a good decision. You can also try something like:
`if parse_status = ...; condition {`
I think C++17 offers this feature and there shouldn't be any shadowing.

trend-edward-xu: Hi @torwig,
Thanks for your suggestions.
I have updated the syntax with the `if-initialized` statement.

Best Regards,
Edward

LindaSummer: Hi @torwig ,

Thanks for your suggestions.
I have updated the syntax with the `if-initialized` statement.

Best Regards,
Edward

src/types/redis_stream.cc
```diff
@@ -511,6 +511,163 @@ rocksdb::Status Stream::ClaimPelEntries(const Slice &stream_name, const std::str
...
+
+      pending_entries.emplace_back(std::move(entry));
+      --count;
+
+      if (penl_entry.consumer_name != consumer_name) {
+        ++total_claimed_count;
+        claimed_consumer_entity_count[penl_entry.consumer_name] += 1;
+        penl_entry.consumer_name = consumer_name;
+        penl_entry.last_delivery_time_ms = now_ms;
+        penl_entry.last_delivery_count += 1;
```
torwig: According to the documentation:

> Lastly, claiming a message with XAUTOCLAIM also increments the attempted deliveries count for that message, unless the JUSTID option has been specified (which only delivers the message ID, not the message itself). Messages that cannot be processed for some reason - for example, because consumers systematically crash when processing them - will exhibit high attempted delivery counts that can be detected by monitoring.

Additionally, you can double-check the behavior via Redis.

LindaSummer: Hi @torwig ,
Thanks for your correction.

I have read the document and use [`extended-form-of-xpending`](https://redis.io/docs/latest/commands/xpending/#extended-form-of-xpending) to check its behavior on Redis.

Its behavior is same as description, `justid` option will not increment the counter.

I also checked the [Redis source code](https://github.com/redis/redis/blob/7.4/src/t_stream.c#L3468). 
It skipped the incrementation when `justid` enabled.

https://github.com/redis/redis/blob/7.4/src/t_stream.c#L3468-L3470

I have updated the code following Redis code.

Best Regards,
Edward

torwig (approved):

PragmaTwice:

PR #2374, opened 2024-06-21
fix(json):  JSON.ARRTRIM key no exists 

jackyyyyyssss:
image
JSON.ARRTRIM   key no exists  not return null


torwig (approved):

git-hulk (approved):

git-hulk:
@jackyyyyyssss Nice catch!

PR #2377, opened 2024-06-27
refactor(cron): add CronPattern and implement more cron syntax

PragmaTwice:
New supported syntax:

```
1,2,3,4-7,8
```

macos11 is disabled in CI, since these jobs cannot be started normally.

### DIFF
--- a/src/common/cron.h
+++ b/src/common/cron.h
@@ -79,6 +79,10 @@ struct CronPattern {
         }
       }
 
+      if (results.empty()) {
+        return {Status::NotOK, "invalid cron expression"};
+      }
+
       return CronPattern{results};
     }
   }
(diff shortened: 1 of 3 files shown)


git-hulk (dismissed):

git-hulk (dismissed):

src/common/cron.h
```diff
@@ -23,23 +23,128 @@
...
-  bool wday_interval = false;
+  struct Interval {
+    int interval;
+  };                                                         // */n
+  struct Any {};                                             // *
+  using Numbers = std::vector<std::variant<Number, Range>>;  // 1,2,3-6,7
+
+  std::variant<Numbers, Interval, Any> val;
+
+  static StatusOr<CronPattern> Parse(std::string_view str, std::tuple<int, int> minmax) {
```
torwig, line 45: I just wondering why std::tuple and not std::pair since we expect only 2 values. + We already use std::pair in this file.

torwig (dismissed):

src/common/cron.h
PragmaTwice: The underlying parameter type is tuple in ParseInt, and I don't want to change it now.

git-hulk (dismissed):

torwig (dismissed):

git-hulk (approved):

PR #2383, opened 2024-06-29
feat(config): add compaction-checker-cron to extend original compaction-checker-range

PragmaTwice:
```
# The hour range that compaction checker would be active
# e.g. compaction-checker-range 0-7 means compaction checker would be worker between
# 0-7am every day.
# WARNING: this config option is deprecated and will be removed,
# please use compaction-checker-cron instead
# compaction-checker-range 0-7

# The time pattern that compaction checker would be active
# Time expression format is the same as crontab (supported cron syntax: *, n, */n, `1,3-6,9,11`)
# e.g. compaction-checker-cron * 0-7 * * * means compaction checker would be worker between
# 0-7am every day.
compaction-checker-cron * 0-7 * * *
```

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -497,6 +497,13 @@ profiling-sample-record-threshold-ms 100
 # would compact the db at 3am and 4am everyday
 # compact-cron 0 3 * * *
 
+# The hour range that compaction checker would be active
+# e.g. compaction-checker-range 0-7 means compaction checker would be worker between
+# 0-7am every day.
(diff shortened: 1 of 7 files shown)


git-hulk (dismissed):

PragmaTwice:
It need to be picked to release 2.9.0 to fix a bug in 8547cfb48cb30dc3ad26214980e2c6a61c0e609a.

git-hulk (approved):

PR #2384, opened 2024-06-29
refactor(stream): change the encoding of stream consumer group

Yangsx-1:
The encoding way of stream consumer group is not efficient, this PR changes the encoding according to this [discussion](https://github.com/apache/kvrocks-website/pull/222#issuecomment-2130783692).


PragmaTwice (dismissed):

PragmaTwice:
Should `StreamEntryID::Maximum` be `{UINT64_MAX - 1, ..}` to avoid the `-1` case here?

Yangsx-1:
Can we merge this PR?

git-hulk:
@Yangsx-1 You can merge PR once the CI passed

git-hulk (dismissed):

PragmaTwice:
There is one issue that needs to be considered: after the modification, the case where `EID MS` is -1 needs to be excluded when traversing (iterating) all sub-keys (`EID MS|EID SEQ`).

As I mentioned [here](https://github.com/apache/kvrocks/pull/2384#issuecomment-2200221083).

Yangsx-1:
> There is one issue that needs to be considered: after the modification, the case where `EID MS` is -1 needs to be excluded when traversing (iterating) all sub-keys (`EID MS|EID SEQ`).
> 
> As I mentioned [here](https://github.com/apache/kvrocks/pull/2384#issuecomment-2200221083).

It's been changed now.

git-hulk (dismissed):

PragmaTwice (dismissed):

tests/gocase/unit/type/stream/stream_test.go
```diff
@@ -306,7 +306,7 @@ var streamTests = func(t *testing.T, enabledRESP3 string) {
 		require.Len(t, rdb.XRange(ctx, "vipstream", "(1-0", "(42-42").Val(), 1)
 		require.ErrorContains(t, rdb.XRange(ctx, "vipstream", "(-", "+").Err(), "ERR")
 		require.ErrorContains(t, rdb.XRange(ctx, "vipstream", "-", "(+").Err(), "ERR")
-		require.ErrorContains(t, rdb.XRange(ctx, "vipstream", "(18446744073709551615-18446744073709551615", "+").Err(), "ERR")
+		require.ErrorContains(t, rdb.XRange(ctx, "vipstream", "(18446744073709551614-18446744073709551615", "+").Err(), "ERR")
```
PragmaTwice, line 309: What will happen if it's `18446744073709551615` instead of `18446744073709551614` here?

Yangsx-1: There would be a nil answer rather than an error.

PragmaTwice:
Hi @torwig, could you please review this PR when you have some time?

torwig:
@PragmaTwice Sure. I'll review it today.

torwig:
@Yangsx-1 @PragmaTwice Since we have `UINT64_MAX` as a delimiter, should it be an invalid `seq` of the stream entry ID?
Should we consider changing code [here](https://github.com/apache/kvrocks/blob/unstable/src/types/redis_stream_base.cc#L40), [here](https://github.com/apache/kvrocks/blob/unstable/src/types/redis_stream_base.cc#L182) and [here](https://github.com/apache/kvrocks/blob/unstable/src/types/redis_stream_base.cc#L219)?

Yangsx-1:
> @Yangsx-1 @PragmaTwice Since we have `UINT64_MAX` as a delimiter, should it be an invalid `seq` of the stream entry ID? Should we consider changing code [here](https://github.com/apache/kvrocks/blob/unstable/src/types/redis_stream_base.cc#L40), [here](https://github.com/apache/kvrocks/blob/unstable/src/types/redis_stream_base.cc#L182) and [here](https://github.com/apache/kvrocks/blob/unstable/src/types/redis_stream_base.cc#L219)?

Yes, i'll change these code later.

torwig (approved):

PragmaTwice (approved):

PR #2386, opened 2024-06-30
fix(.asf.yaml): branch protection cannot be created

PragmaTwice:
See changes.


PragmaTwice:
Since now the protection on unstable branch is removed unexpectedly, I need to merge it to recover as soon as possible.

PR #2387, opened 2024-06-30
feat(stream): add support of XPENDING command

### LINKED ISSUE #1734: Add support of the XPENDING command

Yangsx-1:
Close #1734


src/commands/cmd_stream.cc
```diff
@@ -814,6 +815,98 @@ class CommandXInfo : public Commander {
...
+    output->append(redis::MultiLen(3 + results.consumer_infos.size()));
+    output->append(redis::Integer(results.pending_number));
+    output->append(redis::BulkString(results.smallest_id.ToString()));
+    output->append(redis::BulkString(results.greatest_id.ToString()));
+    for (const auto &entry : results.consumer_infos) {
+      output->append(redis::MultiLen(1));
+      output->append(redis::MultiLen(2));
+      output->append(redis::BulkString(entry.first));
+      output->append(redis::BulkString(std::to_string(entry.second)));
+    }
```
git-hulk: This looks wrong according to Redis: https://github.com/redis/redis/blob/a9267137ee5cbd0908d3844e9d57284e93d85b72/src/t_stream.c#L2985

```suggestion
    output->append(redis::MultiLen(results.consumer_infos.size()));
    for (const auto &entry : results.consumer_infos) {
      output->append(redis::MultiLen(2));
      output->append(redis::BulkString(entry.first));
      output->append(redis::BulkString(std::to_string(entry.second)));
    }
```

src/types/redis_stream_base.h
```diff
@@ -239,6 +239,36 @@ struct StreamAutoClaimResult {
...
+  bool with_consumer = false;
+
+  std::string consumer;
+  std::string stream_name;
+  std::string group_name;
+};
+
+struct StreamGetPendingEntryResult {
+  uint64_t pending_number;
+  StreamEntryID smallest_id;
```
git-hulk: Do you think if it's a good idea to rename `smallest_id` and `greatest_id` to `first_entry_id` and `last_entry_id` to keep consistent with other places?

src/types/redis_stream_base.h
```diff
@@ -239,6 +239,36 @@ struct StreamAutoClaimResult {
...
+  StreamEntryID greatest_id;
+  std::vector<std::pair<std::string, int>> consumer_infos;
+};
+
+struct StreamGetExtPendingEntryResult {
```
git-hulk: Redis named as `StreamNACK` which seems a more reasonable name for the message queue: https://github.com/redis/redis/blob/a9267137ee5cbd0908d3844e9d57284e93d85b72/src/t_stream.c#L3025

src/types/redis_stream_base.h
Yangsx-1: I'll rename it.

git-hulk (dismissed):

src/types/redis_stream.cc
```diff
@@ -1707,4 +1707,90 @@ rocksdb::Status Stream::SetId(const Slice &stream_name, const StreamEntryID &las
...
+        last_entry_id = entry_id;
+      }
+      StreamPelEntry pel_entry = decodeStreamPelEntryValue(iter->value().ToString());
+      const std::string &consumer_name = pel_entry.consumer_name;
+
+      if (options.with_consumer && options.consumer != consumer_name) {
+        continue;
+      }
+
+      ext_results.push_back({entry_id, pel_entry.last_delivery_time_ms, pel_entry.last_delivery_count, consumer_name});
```
torwig: @Yangsx-1 Do we have filtering by the IDLE time?
image

Yangsx-1: Sorry for the ignorance, added now.

torwig (dismissed):
LGTM.
Ideally, it would be great to add the test case for the filtering by IDLE time.

Yangsx-1:
> LGTM. Ideally, it would be great to add the test case for the filtering by IDLE time.

Will add it in another PR.  :)

PragmaTwice:
After https://github.com/apache/kvrocks/pull/2384, I think this command need to be refactored.

So it's better to review and merge #2384 before this PR.

PragmaTwice:
https://github.com/apache/kvrocks/pull/2384 is merged, could you refactor this PR to avoid iterating the whole stream subkeys?

Yangsx-1:
> #2384 is merged, could you refactor this PR to avoid iterating the whole stream subkeys?

Sure, will do it later.

src/types/redis_stream.cc
```diff
@@ -1698,4 +1698,91 @@ rocksdb::Status Stream::SetId(const Slice &stream_name, const StreamEntryID &las
   return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
 }
 
+rocksdb::Status Stream::GetPendingEntries(StreamPendingOptions &options, StreamGetPendingEntryResult &pending_infos,
```
torwig, line 1701: Can we pass output parameters by pointers?

src/types/redis_stream.cc
```diff
@@ -1698,4 +1698,91 @@ rocksdb::Status Stream::SetId(const Slice &stream_name, const StreamEntryID &las
...
+
+    const std::string &consumer_name = pel_entry.consumer_name;
+
+    if (options.with_consumer && options.consumer != consumer_name) {
+      continue;
+    }
+
+    ext_results.push_back({entry_id, pel_entry.last_delivery_time_ms, pel_entry.last_delivery_count, consumer_name});
+
+    if (options.with_count) {
```
torwig, line 1761: Could you please elaborate on this if-statement? If `count` is specified, we will not increment the counter. How will we break the loop in line 1738 if the counter remains 0?

src/types/redis_stream.cc
Yangsx-1: We can, but why? I'm not fully understand the necessity of using pointer.

src/types/redis_stream.cc
Yangsx-1: Sorry, that's my fault. I think we should use two different variables to record summary result count and extend result count. And it is revised now.

src/types/redis_stream.cc
torwig: @Yangsx-1 It's not a requirement from my side :) I just saw some conventions that output parameters go at the end of the list and are passed by a pointer. Maybe it has its roots in C. However, references are safer since you can't dereference a null pointer.

torwig (approved):

PragmaTwice (approved):

Issue #2355, opened 2024-06-07
Support passing a slot range in CLUSTERX MIGRATE


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

In slot migration, many users would like to be able to pass a slot range to migrate in batches.

### Solution

--

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PokIsemaine:
I'd like to learn kvrocks' cluster design by solving this problem, if it's suitable please arrange it for me. At the same time, if you have relevant information, please list it. Thank you very much!

git-hulk:
@PokIsemaine Sure, thank you. Free feel to ping @PragmaTwice @git-hulk @caipengbo for help or confirm anything.

caipengbo:
@PokIsemaine I suggest that before you start writing code, outline your solution and let's discuss it together.

PokIsemaine:
First of all, I would like to discuss the use of commands, and noting that `CLUSTERX SETSLOT` has similar needs, I decided to refer to the following PR and issues.
https://github.com/apache/kvrocks/issues/1412
https://github.com/apache/kvrocks/pull/1414

The current form of `CLUSTERX MIGRATE` is as follows
```
CLUSTERX MIGRATE $SLOT_ID $DEST_NODE_ID [$SYNC_FLAG] [$TIMEOUT]
```
To be compatible with the previous command, we've only modified the form of content that SLOT_ID can accept, allowing it to accept a single slot or a range of slotrange, or both.
```
CLUSTERX MIGRATE 1 $dest_node_id
CLUSTERX MIGRATE "3 5 9 20-30" $dest_node_id
```
Do you think this is reasonable? If it makes sense, I'll continue to think about the specific design and implementation.


> In addition, in the process of learning cluster, I found that the https://kvrocks.apache.org/docs/cluster/ documentation is not very user-friendly, for example, it does not tell CLUSTERX MIGRATE how to use it, command line arguments, etc., which may need to be improved in the future

caipengbo:
I agree with this way.

PokIsemaine:
The next thing I'd like to discuss is a more specific design: how do we migrate when we have multiple slot ranges?
```
CLUSTERX MIGRATE "3 5 9 20-30" $dest_node_id
```
In the current implementation, migration-related functions, including `SlotMigrationJob`, are implemented for a single slot.

[...]

PR #2389, opened 2024-07-01
feat(cluster): support migrate slot range

PokIsemaine:
issues: https://github.com/apache/kvrocks/issues/2355

This draft PR demonstrates how to support migrating slot ranges.

# What I Did

1 Migration job - 1 slot range:

* I encapsulated a `SlotRange` structure and changed the migration-related class members from a single slot to a slot range.
* Reference: https://github.com/apache/kvrocks/issues/412. Slot migration includes the following phases: start migration, migrate existing data, migrate incremental data, and end migration. In each modified phase, the entire slot range must be completed before moving to the next phase.

# TODO
TODO represents the items I hope to discuss.

1. Support multiple slot ranges:
   * Current situation: Only one migration job can be performed at a time, and each migration job corresponds to a slot range.
   * Possible modification: Allow multiple migration jobs, migrating sequentially or in parallel.

2. Perform multiple migrations consecutively but do not immediately use `setslot` to update the topology, referring to the example in TestSlotRangeMigrate:
```go
	t.Run("MIGRATE - Repeat migration cases, but does not immediately update the topology via setslot", func(t *testing.T) {
		// Disjoint
		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", "114-116", id1).Val())
		waitForMigrateSlotRangeState(t, rdb0, "114-116", SlotMigrationStateSuccess)
		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", "117-118", id1).Val())
		waitForMigrateSlotRangeState(t, rdb0, "117-118", SlotMigrationStateSuccess)
		require.Equal(t, "OK", rdb0.Do(ctx, "clusterx", "migrate", "112-113", id1).Val())
		waitForMigrateSlotRangeState(t, rdb0, "112-113", SlotMigrationStateSuccess)

		errMsg := "Can't migrate slot which has been migrated"
		// TODO: Migrating 112-113, but 114-118 is covered and cannot be detected.
		// require.ErrorContains(t, rdb0.Do(ctx, "clusterx", "migrate", "114-116", id1).Err(), errMsg)
		// require.ErrorContains(t, rdb0.Do(ctx, "clusterx", "migrate", "117-118", id1).Err(), errMsg)

		// Intersection
		require.ErrorContains(t, rdb0.Do(ctx, "clusterx", "migrate", "112", id1).Err(), errMsg)
		require.ErrorContains(t, rdb0.Do(ctx, "clusterx", "migrate", "112-112", id1).Err(), errMsg)
		require.ErrorContains(t, rdb0.Do(ctx, "clusterx", "migrate", "113", id1).Err(), errMsg)
		require.ErrorContains(t, rdb0.Do(ctx, "clusterx", "migrate", "113-113", id1).Err(), errMsg)

		// Subset
		require.ErrorContains(t, rdb0.Do(ctx, "clusterx", "migrate", "112-113", id1).Err(), errMsg)
		require.ErrorContains(t, rdb0.Do(ctx, "clusterx", "migrate", "112-120", id1).Err(), errMsg)
		require.ErrorContains(t, rdb0.Do(ctx, "clusterx", "migrate", "110-112", id1).Err(), errMsg)
	})
```
This situation also seems to exist in the original single slot migration, and I am not sure if such operations are reasonable.
```go
migrate A => migrate B => setslot A&B
migrate A => migrate B => migrate A (expected to fail, but allowed to pass) => setslot A&B

// slotrange A-C
migrate A-B => migrate C => setslot A-C
migrate A-B => migrate C => migrate B (expected error, but allowed to pass) => setslot A-C
```

3. More precise migration failure slot range:
   The current implementation determines the entire slot range to fail and cleans it up, and the user later migrates the entire slot range again.
   Do we want to support a more precise failure range? For example, `[start_slot-fail_slot), [fail_slot, end]`. Users can check the status with commands such as `cluster info` and then re-migrate the failed slot range by themselves. (Personally, I think it's a bit cumbersome and error-prone)

# Miscellaneous
Other suggestions are welcome, such as more testing, code optimization, better user interaction, etc.


src/cluster/cluster_defs.h
```diff
@@ -33,11 +34,36 @@ enum {
...
+    if (start == end) return fmt::format("{}", start);
+    return fmt::format("{}-{}", start, end);
+  }
+
+  bool operator==(const SlotRange &rhs) const { return start == rhs.start && end == rhs.end; }
+  bool operator!=(const SlotRange &rhs) const { return !(*this == rhs); }
+
+  int start;
+  int end;
+};
```
git-hulk:
```suggestion
};

```

src/cluster/cluster_defs.h
```diff
@@ -33,11 +34,36 @@ enum {
...
 inline constexpr const char *errClusterNoInitialized = "The cluster is not initialized";
 inline constexpr const char *errInvalidClusterNodeInfo = "Invalid cluster nodes info";
 inline constexpr const char *errInvalidImportState = "Invalid import state";
 
-using SlotRange = std::pair<int, int>;
+struct SlotRange {
+  SlotRange(int start, int end) : start(start), end(end) {}
+  SlotRange() : start(-1), end(-1) {}
+  bool IsValid() const {
+    return start >= 0 && start < kClusterSlots && end >= 0 && end < kClusterSlots && start <= end;
```
git-hulk: start <= end && end < kClusterSlots can deduce start < kClusterSlots
```suggestion
    return start >= 0 && end >= 0 && start <= end && end < kClusterSlots;
```

src/cluster/cluster_defs.h
```diff
@@ -33,11 +34,36 @@ enum {
...
+  }
+
+  bool Contain(int slot) const { return IsValid() && slot >= start && slot <= end; }
```
git-hulk: Should be `Contains`

src/cluster/cluster_defs.h
```diff
@@ -33,11 +34,36 @@ enum {
...
+
+  bool CheckIntersection(const SlotRange &rhs) const {
+    return IsValid() && rhs.IsValid() && end >= rhs.start && rhs.end >= start;
+  }
+
+  std::string String() const {
+    if (!IsValid()) return "empty";
```
git-hulk: I'm wondering if `-1` will be better than the "empty" literal since it can be parsed back to an invalid slot.

src/cluster/cluster_defs.h
```diff
@@ -33,11 +34,36 @@ enum {
...
+struct SlotRange {
+  SlotRange(int start, int end) : start(start), end(end) {}
+  SlotRange() : start(-1), end(-1) {}
+  bool IsValid() const {
+    return start >= 0 && start < kClusterSlots && end >= 0 && end < kClusterSlots && start <= end;
+  }
+
+  bool Contain(int slot) const { return IsValid() && slot >= start && slot <= end; }
+
+  bool CheckIntersection(const SlotRange &rhs) const {
```
git-hulk: HasOverlap?

git-hulk (commented):
@PokIsemaine Thanks for your impressive contributions. I will take another pass while it's ready to be reviewed.

src/storage/redis_db.cc
```diff
@@ -534,6 +534,20 @@ rocksdb::Status Database::ClearKeysOfSlot(const rocksdb::Slice &ns, int slot) {
...
+    return rocksdb::Status::Aborted("It is not in cluster mode");
+  }
+
+  std::string prefix = ComposeSlotKeyPrefix(ns, slot_range.start);
+  std::string prefix_end = ComposeSlotKeyPrefix(ns, slot_range.end + 1);
+  auto s = storage_->DeleteRange(prefix, prefix_end);
+  if (!s.ok()) {
+    return s;
+  }
+  return rocksdb::Status::OK();
```
caipengbo:
```suggestion
  return storage_->DeleteRange(prefix, prefix_end);
```

src/cluster/cluster_defs.h
caipengbo: The semantics of range is preferably `[start,end)`.

src/storage/redis_db.cc
caipengbo: Remove the old `ClearKeysOfSlot()`

src/cluster/slot_import.cc
```diff
@@ -21,57 +21,60 @@
...
   std::lock_guard<std::mutex> guard(mutex_);
   if (import_status_ == kImportStart) {
     // return ok if the same slot is importing
-    if (import_slot_ == slot) {
+    if (import_slot_range_ == slot_range) {
       return Status::OK();
     }
-    return {Status::NotOK, fmt::format("only one importing slot is allowed, current slot is: {}", import_slot_)};
+    return {Status::NotOK,
+            fmt::format("only one importing slot is allowed, current slot is: {}", import_slot_range_.String())};
```
caipengbo:
```suggestion
            fmt::format("only one importing job is allowed, current importing: {}", import_slot_range_.String())};
```

src/cluster/slot_import.cc
```diff
@@ -21,57 +21,60 @@
...
   return Status::OK();
 }
 
-Status SlotImport::Success(int slot) {
+Status SlotImport::Success(const SlotRange &slot_range) {
   std::lock_guard<std::mutex> guard(mutex_);
-  if (import_slot_ != slot) {
-    return {Status::NotOK, fmt::format("mismatch slot, importing slot: {}, but got: {}", import_slot_, slot)};
+  if (import_slot_range_ != slot_range) {
+    return {Status::NotOK, fmt::format("mismatch slot, importing slot: {}, but got: {}", import_slot_range_.String(),
```
caipengbo:
```suggestion
    return {Status::NotOK, fmt::format("mismatch slots, importing slots: {}, but got: {}", import_slot_range_.String(),
```

src/cluster/slot_import.cc
```diff
@@ -21,57 +21,60 @@
...
   return Status::OK();
 }
 
-Status SlotImport::Fail(int slot) {
+Status SlotImport::Fail(const SlotRange &slot_range) {
   std::lock_guard<std::mutex> guard(mutex_);
-  if (import_slot_ != slot) {
-    return {Status::NotOK, fmt::format("mismatch slot, importing slot: {}, but got: {}", import_slot_, slot)};
+  if (import_slot_range_ != slot_range) {
+    return {Status::NotOK, fmt::format("mismatch slot, importing slot: {}, but got: {}", import_slot_range_.String(),
```
caipengbo:
```suggestion
    return {Status::NotOK, fmt::format("mismatch slots, importing slots: {}, but got: {}", import_slot_range_.String(),
```

src/cluster/cluster_defs.h
PokIsemaine: I referenced the `SetSlotRanges` implementation, and it should be following `[start, end]`.
https://github.com/apache/kvrocks/blob/7ed512e835201abe3f68f3c795dabb13acdc6733/src/cluster/cluster.cc#L120
Should they be harmonized, especially since `migrate` and `setslot` tend to appear together when the user is using them.
If you change it to `[start, end)` then `SetSlotRanges` should be changed synchronously too, what do you think to do?

caipengbo: OK, keep it that way.

src/cluster/slot_import.cc
PokIsemaine: Thanks for the advice. Actually, if you search for `".String()"`, there are many similar issues.
image

In the beginning, I tried changing "slot" to "slots", but encountered some difficulties. On one hand, when `start == end`, `String()` will return a single `slot` instead of `[start,end]`, which causes ambiguity with the term "slots" in the message. On the other hand, some expected values in the test cases also need to be changed from "slot" to "slots". So I temporarily abandoned these modifications.

```cpp
std::string String() const {
  if (!IsValid()) return "-1";
  if (start == end) return fmt::format("{}", start);  // <== HERE
  return fmt::format("{}-{}", start, end);
}
```

Regarding the ambiguity issue, we could probably change it to "slot(s)" instead of "slot" or "slots" to be as compatible as possible with both representations. If you think that's OK, I'll change them (including some cases in the test cases).

caipengbo: Thanks for the deep thought. If you think it is not trouble, better modify it :)

PokIsemaine:
@git-hulk @caipengbo 

If it's convenient, I would like to discuss some matters related to the TODO list:

First, regarding item 2 on the TODO list, I find this situation a bit confusing because I am not sure if this is the intended design of kvrocks. If possible, please determine whether this is a normal situation.

Second, concerning items 1 and 3 on the TODO list, these are enhancements. Please evaluate whether these features are needed and if they should be included in this PR.

caipengbo:
@PokIsemaine 
> First, regarding item 2 on the TODO list, I find this situation a bit confusing because I am not sure if this is the intended design of kvrocks. If possible, please determine whether this is a normal situation.

This is indeed the case at the moment. The migrated slots is not checked when the migration is performed, and I think they should be checked.

> Second, concerning items 1 and 3 on the TODO list, these are enhancements. Please evaluate whether these features are needed and if they should be included in this PR.

I think these are good improvements, but we can add them in the subsequent PR. This PR is enough.

PokIsemaine:
@caipengbo Ok, thanks for the suggestion. So for the second TODO, do we try to fix it at this PR or create another separate issue to track it?

caipengbo:
> for the second TODO, do we try to fix it at this PR or create another separate issue to track it?

Let's fix it in this PR. @PokIsemaine

git-hulk:
@PokIsemaine I think this PR is ready to review now?

PokIsemaine:
> @PokIsemaine I think this PR is ready to review now?

You can start the review now :)

The `clang-tidy` check in the CI doesn't seem to be working, so please try to re-run it. This works in my local and fork repository CI.

PragmaTwice:
Could you fix these clang-tidy issues? Please make sure the code can be successfully compiled.
```
/home/runner/work/kvrocks/kvrocks/build/_deps/fmt-src/include/fmt/base.h:1619:3: error: static_assert failed due to requirement 'formattable_pointer' "Formatting of non-void pointers is disallowed." [clang-diagnostic-error]
Suppressed 21102 warnings (21069 in non-user code, 32 NOLINT, 1 with check filters).
Found compiler error(s).
  static_assert(formattable_pointer,
  ^             ~~~~~~~~~~~~~~~~~~~
/home/runner/work/kvrocks/kvrocks/build/_deps/fmt-src/include/fmt/base.h:2002:20: note: in instantiation of function template specialization 'fmt::detail::make_arg<true, fmt::context, double (double) noexcept, 0>' requested here
  return {{detail::make_arg<NUM_ARGS <= detail::max_packed_args, Context>(
                   ^
/home/runner/work/kvrocks/kvrocks/build/_deps/fmt-src/include/fmt/format.h:4357:28: note: in instantiation of function template specialization 'fmt::make_format_args<fmt::context, double (double) noexcept, std::basic_string<char>, 2UL, 0UL, 223ULL, 0>' requested here
  return vformat(fmt, fmt::make_format_args(args...));
                           ^
/home/runner/work/kvrocks/kvrocks/src/cluster/slot_migrate.cc:394:33: note: in instantiation of function template specialization 'fmt::format<double (&)(double) noexcept, std::basic_string<char> &>' requested here
    return {Status::NotOK, fmt::format("failed to iterate keys of slot {}: {}", slot, err_str)};
                                ^
/home/runner/work/kvrocks/kvrocks/src/cluster/slot_migrate.cc:393:66: error: use of undeclared identifier 'slot' [clang-diagnostic-error]
    LOG(ERROR) << "[migrate] Failed to iterate keys of slot " << slot << ": " << err_str;
                                                                 ^
/home/runner/work/kvrocks/kvrocks/src/cluster/slot_migrate.cc:394:81: error: use of undeclared identifier 'slot' [clang-diagnostic-error]
    return {Status::NotOK, fmt::format("failed to iterate keys of slot {}: {}", slot, err_str)};
                                                                                ^
```

git-hulk:
> > @PokIsemaine I think this PR is ready to review now?
> 
> You can start the review now :)
> 
> The `clang-tidy` check in the CI doesn't seem to be working, so please try to re-run it. This works in my local and fork repository CI.

Thank you! I will take another pass in one or two days.

src/cluster/slot_migrate.cc
```diff
@@ -74,20 +74,20 @@ SlotMigrator::SlotMigrator(Server *srv)
...
-Status SlotMigrator::PerformSlotMigration(const std::string &node_id, std::string &dst_ip, int dst_port, int slot_id,
-                                          SyncMigrateContext *blocking_ctx) {
+Status SlotMigrator::PerformSlotRangeMigration(const std::string &node_id, std::string &dst_ip, int dst_port,
+                                               const SlotRange &slot_range, SyncMigrateContext *blocking_ctx) {
+  // TODO: concurrent migration, multiple migration jobs
   // Only one slot migration job at the same time
-  int16_t no_slot = -1;
-  if (!migrating_slot_.compare_exchange_strong(no_slot, static_cast<int16_t>(slot_id))) {
-    return {Status::NotOK, "There is already a migrating slot"};
+  SlotRange no_slot_range = {-1, -1};
```
git-hulk: no_slot_range => empty_slot_range

src/cluster/slot_migrate.cc
```diff
@@ -346,6 +351,7 @@ Status SlotMigrator::sendSnapshotByCmd() {
   auto iter = util::UniqueIterator(storage_->GetDB()->NewIterator(read_options, cf_handle));
 
   // Seek to the beginning of keys start with 'prefix' and iterate all these keys
+  auto current_slot = slot_range.start;
```
git-hulk, line 354: Seems this assignment can be removed.

git-hulk (commented):
This PR generally looks good to me, left two comments. To see if other maintainers have further comments.

src/cluster/slot_migrate.cc
PokIsemaine: This also informs the following error handling, considering the case where the first iter fails and does not update the `current_slot`.

```
  if (auto s = iter->status(); !s.ok()) {
    auto err_str = s.ToString();
    LOG(ERROR) << "[migrate] Failed to iterate keys of slot " << current_slot << ": " << err_str;
    return {Status::NotOK, fmt::format("failed to iterate keys of slot {}: {}", current_slot, err_str)};
  }
```

git-hulk: @PokIsemaine Got it, makes sense. Thanks for your explain.

git-hulk (dismissed):

src/cluster/cluster_defs.h
```diff
@@ -33,11 +34,34 @@ enum {
...
+inline constexpr const char *errSlotRangeInvalid = "Slot range is invalid";
 inline constexpr const char *errInvalidClusterVersion = "Invalid cluster version";
 inline constexpr const char *errSlotOverlapped = "Slot distribution is overlapped";
 inline constexpr const char *errNoMasterNode = "The node isn't a master";
 inline constexpr const char *errClusterNoInitialized = "The cluster is not initialized";
 inline constexpr const char *errInvalidClusterNodeInfo = "Invalid cluster nodes info";
 inline constexpr const char *errInvalidImportState = "Invalid import state";
 
-using SlotRange = std::pair<int, int>;
+struct SlotRange {
```
mapleFU, line 48: Perhaps a comment for inclusive is better here

src/cluster/cluster.cc
```diff
@@ -214,7 +214,7 @@ Status Cluster::SetClusterNodes(const std::string &nodes_str, int64_t version, b
   if (!migrated_slots_.empty()) {
     for (const auto &[slot, _] : migrated_slots_) {
       if (slots_nodes_[slot] != myself_) {
-        auto s = srv_->slot_migrator->ClearKeysOfSlot(kDefaultNamespace, slot);
+        auto s = srv_->slot_migrator->ClearKeysOfSlotRange(kDefaultNamespace, {slot, slot});
```
mapleFU: Nit: a helper function for single range is ok here

src/storage/batch_extractor.cc
```diff
@@ -279,15 +281,17 @@ rocksdb::Status WriteBatchExtractor::DeleteCF(uint32_t column_family_id, const S
     std::string user_key;
     std::tie(ns, user_key) = ExtractNamespaceKey<std::string>(key, is_slot_id_encoded_);
 
-    if (slot_id_ >= 0 && static_cast<uint16_t>(slot_id_) != GetSlotIdFromKey(user_key)) {
+    auto key_slot_id = GetSlotIdFromKey(user_key);
+    if (slot_range_.IsValid() && !slot_range_.Contains(key_slot_id)) {
```
mapleFU, line 285: Is `Contains` enough here?

src/storage/batch_extractor.cc
```diff
@@ -279,15 +281,17 @@ rocksdb::Status WriteBatchExtractor::DeleteCF(uint32_t column_family_id, const S
...
       return rocksdb::Status::OK();
     }
 
     command_args = {"DEL", user_key};
   } else if (column_family_id == static_cast<uint32_t>(ColumnFamilyID::PrimarySubkey)) {
     InternalKey ikey(key, is_slot_id_encoded_);
     std::string user_key = ikey.GetKey().ToString();
-    if (slot_id_ >= 0 && static_cast<uint16_t>(slot_id_) != GetSlotIdFromKey(user_key)) {
+    auto key_slot_id = GetSlotIdFromKey(user_key);
+    if (slot_range_.IsValid() && !slot_range_.Contains(key_slot_id)) {
```
mapleFU, line 294: ditto?

mapleFU (dismissed):
+1 for we with some nits

src/storage/batch_extractor.cc
PokIsemaine: Do you think there are any inappropriate examples? Could you please elaborate?

mapleFU: Oh I got this since invalid always return !contains, ignoring this

src/cluster/cluster.cc
PokIsemaine: Do you mean to keep `ClearKeysOfSlot`?

src/cluster/slot_migrate.h
```diff
@@ -190,9 +191,9 @@ class SlotMigrator : public redis::Database {
   UniqueFD dst_fd_;
 
   MigrationType migration_type_ = MigrationType::kRedisCommand;
-  std::atomic<int16_t> forbidden_slot_ = -1;
-  std::atomic<int16_t> migrating_slot_ = -1;
-  int16_t migrate_failed_slot_ = -1;
+  std::atomic<SlotRange> forbidden_slot_range_ = SlotRange{-1, -1};
+  std::atomic<SlotRange> slot_range_ = SlotRange{-1, -1};
```
mapleFU, line 197: should we add a `static_assert(slot_range_.is_lock_free())` to force `slot_range` cas not uses mutex?

src/cluster/slot_migrate.h
```diff
@@ -55,8 +55,9 @@ enum class SlotMigrationStage { kNone, kStart, kSnapshot, kWAL, kSuccess, kFaile
 enum class KeyMigrationResult { kMigrated, kExpired, kUnderlyingStructEmpty };
 
 struct SlotMigrationJob {
-  SlotMigrationJob(int slot_id, std::string dst_ip, int dst_port, int speed, int pipeline_size, int seq_gap)
-      : slot_id(static_cast<int16_t>(slot_id)),
+  SlotMigrationJob(const SlotRange &slot_range, std::string dst_ip, int dst_port, int speed, int pipeline_size,
+                   int seq_gap)
+      : slot_range(slot_range),
```
mapleFU: this is a bit weird so maybe rename the argument to `slow_range_in`?

PokIsemaine: I don't quite understand the meaning of `slow_range_in`.

src/cluster/cluster.cc
mapleFU: Nope, I mean maybe two helper functions:

```
// Get [slot_id, slot_id] 
SlotRange::GetPoint(int slot_id)
SlotRange::GetInvalid();
```

I think they would better for understanding

src/cluster/slot_migrate.h
PokIsemaine: It seems like this can't be checked at compile time? `is_always_lock_free`?
https://godbolt.org/z/8qeW4M6he

src/cluster/slot_migrate.h
mapleFU: `slot_range(slot_range)` is a bit weird since input arg and member shares the same naming. I think rename input argument to another name is a good way for this problem

src/cluster/slot_migrate.h
mapleFU: Yeah, I think a compile time `is_always_lock_free` is better here, currently two integer will not meet any problem but If future somebody change this to more than 8B this should be take care of

src/cluster/slot_migrate.h
PokIsemaine: ok, does `slow_range_in` here mean there is `max_speed`? I wanted to confirm if you actually meant to express `slot_range_in` but made a typo. Because I didn’t react right away.

src/cluster/cluster.cc
PokIsemaine: I will add a static function `GetPoint` in `SlotRange`. For `GetInvaild`, I feel that it may not be suitable if it is used to replace `{-1, -1}`, because sometimes `{-1, -1}` represents a wrong SlotRange, and sometimes represents an empty SlotRange, depending on the context.

git-hulk (approved):

git-hulk:
Will merge this PR if no further comments and CI is passed.

caipengbo (commented):
LGTM

mapleFU:
Nice work, thanks!

Issue #2292, opened 2024-05-05
Introduce the key encoding version


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

The current key encoding has some historical legacy issues, and we need to adjust it through some means while ensuring compatibility. 

To do this, first we can persist a key encoding version in storage and read it during loading.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

VasuDevrani:
I went through the codebase, can you confirm if i got it right? 
- the key encoding logic (encoding and decoding) is implemented in the `InternalKey` class 
- can be fixed by introducing a new variable for encoding version, a function to read it from storage during loading process, update `InternalKey` class to handle different versions

PragmaTwice:
> I went through the codebase, can you confirm if i got it right?
> 
> * the key encoding logic (encoding and decoding) is implemented in the `InternalKey` class
> * can be fixed by introducing a new variable for encoding version, a function to read it from storage during loading process, update `InternalKey` class to handle different versions

Yeah you are right.

PR #2402, opened 2024-07-11
feat(hash): Support hash field expiration

### LINKED ISSUE #2269: Add support of hash field expiration
### Search before asking

### Motivation

Refer to https://github.com/redis/redis/pull/13172, Redis has supported this feature.

### Solution

TBD

jjz921024:
This pr want to close #2269 

1. Encoding 
To implement hash expiration by field level, we encode the expiration time into the first 8 bytes of value. 
For example:
    ```
                         +-------------------------+
    key|version|field => | expire (8 byte) | value |
                         +-------------------------+
    ```
    And we use the second bit of the flags in metadata to indicate whether encode the expiration into value.
    If all the fields in the hash have no encode expiration time, the `flags` will be `1 0 0 0 | 0 0 1 0`. 
    If any field in the hash is encoded into the expiration time, the `flags` will be `1 1 0 0 | 0 0 1 0`. 

2. Flags conversion
For the [HEXPIRE](https://redis.io/docs/latest/commands/hexpire/) command, it will set an expiration on one or more fields.
If we execute hexpire command on a hash key that has not been set to expire, this will set the secode bit of `flags` and encode expiration for all fields (fields that not appear in the command will be set to 0)

3. Add commands
According to https://redis.io/docs/latest/commands   add a series of related commands:
    - `HEXPIRE` , `HPEXPIRE` , `HEXPIREAT` , `HPEXPIREAT`
    - `HEXPIRETIME` , `HPEXPIRETIME` , `HTTL` , `HPTTL`
    - `HPERSIST`

### DIFF
--- a/src/commands/cmd_hash.cc
+++ b/src/commands/cmd_hash.cc
@@ -23,6 +23,7 @@
 #include "error_constants.h"
 #include "scan_base.h"
 #include "server/server.h"
+#include "time_util.h"
 #include "types/redis_hash.h"
 
 namespace redis {
@@ -444,6 +445,238 @@ class CommandHRandField : public Commander {
   bool no_parameters_ = true;
 };
 
+class CommandFieldExpireBase : public Commander {
+ protected:
+  Status commonParse(const std::vector<std::string> &args, int start_idx) {
+    CommandParser parser(args, start_idx);
+    std::string_view expire_flag, num_flag;
+    uint64_t fields_num = 0;
+    while (parser.Good()) {
+      if (parser.EatEqICaseFlag("FIELDS", num_flag)) {
+        fields_num = GET_OR_RET(parser.template TakeInt<uint64_t>());
+        break;
+      } else if (parser.EatEqICaseFlag("NX", expire_flag)) {
+        field_expire_type_ = HashFieldExpireType::NX;
+      } else if (parser.EatEqICaseFlag("XX", expire_flag)) {
+        field_expire_type_ = HashFieldExpireType::XX;
+      } else if (parser.EatEqICaseFlag("GT", expire_flag)) {
+        field_expire_type_ = HashFieldExpireType::GT;
+      } else if (parser.EatEqICaseFlag("LT", expire_flag)) {
+        field_expire_type_ = HashFieldExpireType::LT;
+      } else {
+        return parser.InvalidSyntax();
+      }
+    }
+
+    auto remains = parser.Remains();
+    auto size = args.size();
+    if (remains != fields_num) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+
+    for (size_t i = size - remains; i < size; i++) {
+      fields_.emplace_back(args_[i]);
+    }
+
+    return Status::OK();
+  }
+
+  Status expireFieldExecute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) {
+    if (!srv->storage->GetConfig()->hash_field_expiration) {
+      return {Status::RedisExecErr, "field expiration feature is disabled"};
+    }
+
+    std::vector<int8_t> ret;
+    redis::Hash hash_db(srv->storage, conn->GetNamespace());
+    auto s = hash_db.ExpireFields(ctx, args_[1], expire_, fields_, field_expire_type_, &ret);
+    if (!s.ok()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    *output = redis::MultiLen(ret.size());
+    for (const auto &i : ret) {
+      output->append(redis::Integer(i));
+    }
+
+    return Status::OK();
+  }
+
+  Status ttlExpireExecute(engine::Context &ctx, Server *srv, Connection *conn, std::vector<int64_t> &ret) {
+    redis::Hash hash_db(srv->storage, conn->GetNamespace());
+    auto s = hash_db.TTLFields(ctx, args_[1], fields_, &ret);
+    if (!s.ok()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+    return Status::OK();
+  }
+
+  uint64_t expire_ = 0;
+  HashFieldExpireType field_expire_type_ = HashFieldExpireType::None;
+  std::vector<Slice> fields_;
+};
+
+class CommandHExpire : public CommandFieldExpireBase {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto parse_result = ParseInt<uint64_t>(args[2], 10);
+    if (!parse_result) return {Status::RedisParseErr, errValueNotInteger};
+
+    expire_ = *parse_result * 1000 + util::GetTimeStampMS();
+    return CommandFieldExpireBase::commonParse(args, 3);
+  }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    return expireFieldExecute(ctx, srv, conn, output);
+  }
+};
+
+class CommandHExpireAt : public CommandFieldExpireBase {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto parse_result = ParseInt<uint64_t>(args[2], 10);
+    if (!parse_result) return {Status::RedisParseErr, errValueNotInteger};
+
+    expire_ = *parse_result * 1000;
+    return CommandFieldExpireBase::commonParse(args, 3);
+  }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    return expireFieldExecute(ctx, srv, conn, output);
+  }
+};
+
+class CommandHPExpire : public CommandFieldExpireBase {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto parse_result = ParseInt<uint64_t>(args[2], 10);
+    if (!parse_result) return {Status::RedisParseErr, errValueNotInteger};
+
+    expire_ = *parse_result + util::GetTimeStampMS();
+    return CommandFieldExpireBase::commonParse(args, 3);
+  }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    return expireFieldExecute(ctx, srv, conn, output);
+  }
+};
+
+class CommandHPExpireAt : public CommandFieldExpireBase {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto parse_result = ParseInt<uint64_t>(args[2], 10);
+    if (!parse_result) return {Status::RedisParseErr, errValueNotInteger};
+
+    expire_ = *parse_result;
+    return CommandFieldExpireBase::commonParse(args, 3);
+  }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    return expireFieldExecute(ctx, srv, conn, output);
+  }
+};
+
+class CommandHExpireTime : public CommandFieldExpireBase {
+ public:
+  Status Parse(const std::vector<std::string> &args) override { return CommandFieldExpireBase::commonParse(args, 2); }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    std::vector<int64_t> ret;
+    auto s = ttlExpireExecute(ctx, srv, conn, ret);
+    if (!s.IsOK()) {
+      return {Status::RedisExecErr, s.Msg()};
+    }
+    auto now = util::GetTimeStampMS();
+    *output = redis::MultiLen(ret.size());
+    for (const auto &ttl : ret) {
+      if (ttl > 0) {
+        output->append(redis::Integer((now + ttl) / 1000));
+      } else {
+        output->append(redis::Integer(ttl));
+      }
+    }
+    return Status::OK();
+  }
+};
+
+class CommandHPExpireTime : public CommandFieldExpireBase {
+ public:
+  Status Parse(const std::vector<std::string> &args) override { return CommandFieldExpireBase::commonParse(args, 2); }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    std::vector<int64_t> ret;
+    auto s = ttlExpireExecute(ctx, srv, conn, ret);
+    if (!s.IsOK()) {
+      return {Status::RedisExecErr, s.Msg()};
+    }
+    auto now = util::GetTimeStampMS();
+    *output = redis::MultiLen(ret.size());
+    for (const auto &ttl : ret) {
+      if (ttl > 0) {
+        output->append(redis::Integer(now + ttl));
+      } else {
+        output->append(redis::Integer(ttl));
+      }
+    }
+    return Status::OK();
+  }
+};
+
+class CommandHTTL : public CommandFieldExpireBase {
+ public:
+  Status Parse(const std::vector<std::string> &args) override { return CommandFieldExpireBase::commonParse(args, 2); }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    std::vector<int64_t> ret;
+    auto s = ttlExpireExecute(ctx, srv, conn, ret);
+    if (!s.IsOK()) {
+      return {Status::RedisExecErr, s.Msg()};
+    }
+    *output = redis::MultiLen(ret.size());
+    for (const auto &ttl : ret) {
+      output->append(redis::Integer(ttl > 0 ? ttl / 1000 : ttl));
+    }
+    return Status::OK();
+  }
+};
+
+class CommandHPTTL : public CommandFieldExpireBase {
+ public:
+  Status Parse(const std::vector<std::string> &args) override { return CommandFieldExpireBase::commonParse(args, 2); }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    std::vector<int64_t> ret;
+    auto s = ttlExpireExecute(ctx, srv, conn, ret);
+    if (!s.IsOK()) {
+      return {Status::RedisExecErr, s.Msg()};
+    }
+    *output = redis::MultiLen(ret.size());
+    for (const auto &ttl : ret) {
+      output->append(redis::Integer(ttl));
+    }
+    return Status::OK();
+  }
+};
+
+class CommandHPersist : public CommandFieldExpireBase {
+ public:
+  Status Parse(const std::vector<std::string> &args) override { return CommandFieldExpireBase::commonParse(args, 2); }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    std::vector<int8_t> ret;
+    redis::Hash hash_db(srv->storage, conn->GetNamespace());
+    auto s = hash_db.PersistFields(ctx, args_[1], fields_, &ret);
+    if (!s.ok()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    *output = redis::MultiLen(ret.size());
+    for (const auto &i : ret) {
+      output->append(redis::Integer(i));
+    }
+    return Status::OK();
+  }
+};
+
 REDIS_REGISTER_COMMANDS(Hash, MakeCmdAttr<CommandHGet>("hget", 3, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandHIncrBy>("hincrby", 4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandHIncrByFloat>("hincrbyfloat", 4, "write", 1, 1, 1),
@@ -460,6 +693,15 @@ REDIS_REGISTER_COMMANDS(Hash, MakeCmdAttr<CommandHGet>("hget", 3, "read-only", 1
                         MakeCmdAttr<CommandHGetAll>("hgetall", 2, "read-only slow", 1, 1, 1),
                         MakeCmdAttr<CommandHScan>("hscan", -3, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandHRangeByLex>("hrangebylex", -4, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandHRandField>("hrandfield", -2, "read-only slow", 1, 1, 1), )
+                        MakeCmdAttr<CommandHRandField>("hrandfield", -2, "read-only slow", 1, 1, 1),
+                        MakeCmdAttr<CommandHExpire>("hexpire", -6, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandHExpireAt>("hexpireat", -6, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandHExpireTime>("hexpiretime", -5, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandHPExpire>("hpexpire", -6, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandHPExpireAt>("hpexpireat", -6, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandHPExpireTime>("hpexpiretime", -5, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandHPersist>("hpersist", -5, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandHTTL>("httl", -5, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandHPTTL>("hpttl", -5, "read-only", 1, 1, 1), )
 
 }  // namespace redis
--- a/src/config/config.h
+++ b/src/config/config.h
@@ -177,6 +177,9 @@ struct Config {
 
   bool skip_block_cache_deallocation_on_close = false;
 
+  // whether to enable hash field expiration feature
+  bool hash_field_expiration = false;
+
   struct RocksDB {
     int block_size;
     bool cache_index_and_filter_blocks;
--- a/src/storage/compact_filter.cc
+++ b/src/storage/compact_filter.cc
@@ -28,6 +28,7 @@
 #include "db_util.h"
 #include "time_util.h"
 #include "types/redis_bitmap.h"
+#include "types/redis_hash.h"
 
 namespace engine {
 
@@ -132,7 +133,9 @@ bool SubKeyFilter::Filter([[maybe_unused]] int level, const Slice &key, const Sl
     return false;
   }
 
-  return IsMetadataExpired(ikey, metadata) || (metadata.Type() == kRedisBitmap && redis::Bitmap::IsEmptySegment(value));
+  return IsMetadataExpired(ikey, metadata) ||
+         (metadata.Type() == kRedisBitmap && redis::Bitmap::IsEmptySegment(value)) ||
+         (metadata.Type() == kRedisHash && redis::Hash::IsFieldExpired(cached_metadata_, value));
 }
 
 }  // namespace engine
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -140,7 +140,7 @@ rocksdb::Status Database::Expire(engine::Context &ctx, const Slice &user_key, ui
     if (!s.ok()) return s;
     redis::Hash hash_db(storage_, namespace_);
     if (hash_db.GetValidFieldCount(ctx, ns_key, hash_metadata) == 0) {
-      return rocksdb::Status::NotFound("no element found");
+      return rocksdb::Status::NotFound("no elements");
     }
   }
   if (metadata.expire == timestamp) return rocksdb::Status::OK();
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -106,7 +106,7 @@ rocksdb::Status Database::GetMetadata(engine::Context &ctx, RedisTypes types, co
     s = hash_metadata.Decode(*raw_value);
     if (!s.ok()) return s;
     redis::Hash hash_db(storage_, namespace_);
-    if (!hash_db.ExistValidField(ctx, ns_key, hash_metadata)) {
+    if (hash_db.GetValidFieldCount(ctx, ns_key, hash_metadata) == 0) {
       return rocksdb::Status::NotFound("no element found");
     }
   }
@@ -139,7 +139,7 @@ rocksdb::Status Database::Expire(engine::Context &ctx, const Slice &user_key, ui
     s = hash_metadata.Decode(value);
     if (!s.ok()) return s;
     redis::Hash hash_db(storage_, namespace_);
-    if (!hash_db.ExistValidField(ctx, ns_key, hash_metadata)) {
+    if (hash_db.GetValidFieldCount(ctx, ns_key, hash_metadata) == 0) {
       return rocksdb::Status::NotFound("no element found");
     }
   }
@@ -229,7 +229,7 @@ rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &k
       s = hash_metadata.Decode(rocksdb::Slice(pin_values[i].data(), pin_values[i].size()));
       if (!s.ok()) continue;
       redis::Hash hash_db(storage_, namespace_);
-      if (hash_db.ExistValidField(ctx, slice_keys[i], hash_metadata)) {
+      if (hash_db.GetValidFieldCount(ctx, slice_keys[i], hash_metadata) > 0) {
         if (!s.ok()) return s;
         *deleted_cnt += 1;
       }
@@ -268,7 +268,7 @@ rocksdb::Status Database::TTL(engine::Context &ctx, const Slice &user_key, int64
     s = hash_metadata.Decode(value);
     if (!s.ok()) return s;
     redis::Hash hash_db(storage_, namespace_);
-    if (!hash_db.ExistValidField(ctx, ns_key, hash_metadata)) {
+    if (hash_db.GetValidFieldCount(ctx, ns_key, hash_metadata) == 0) {
       *ttl = -2;
       return rocksdb::Status::OK();
     }
@@ -335,7 +335,7 @@ rocksdb::Status Database::Keys(engine::Context &ctx, const std::string &prefix, 
         s = hash_metadata.Decode(iter->value());
         if (!s.ok()) continue;
         redis::Hash hash_db(storage_, namespace_);
-        if (!hash_db.ExistValidField(ctx, iter->key(), hash_metadata)) {
+        if (hash_db.GetValidFieldCount(ctx, iter->key(), hash_metadata) == 0) {
           continue;
         }
       }
@@ -427,7 +427,7 @@ rocksdb::Status Database::Scan(engine::Context &ctx, const std::string &cursor, 
         s = hash_metadata.Decode(iter->value());
         if (!s.ok()) continue;
         redis::Hash hash_db(storage_, namespace_);
-        if (!hash_db.ExistValidField(ctx, iter->key(), hash_metadata)) {
+        if (hash_db.GetValidFieldCount(ctx, iter->key(), hash_metadata) == 0) {
           continue;
         }
       }
@@ -727,7 +727,7 @@ rocksdb::Status Database::existsInternal(engine::Context &ctx, const std::vector
         s = hash_metadata.Decode(value);
         if (!s.ok()) return s;
         redis::Hash hash_db(storage_, namespace_);
-        if (!hash_db.ExistValidField(ctx, key, hash_metadata)) {
+        if (hash_db.GetValidFieldCount(ctx, key, hash_metadata) == 0) {
           continue;
         }
       }
@@ -753,7 +753,7 @@ rocksdb::Status Database::typeInternal(engine::Context &ctx, const Slice &key, R
     s = hash_metadata.Decode(value);
     if (!s.ok()) return s;
     redis::Hash hash_db(storage_, namespace_);
-    if (hash_db.ExistValidField(ctx, key, hash_metadata)) {
+    if (hash_db.GetValidFieldCount(ctx, key, hash_metadata) > 0) {
       *type = metadata.Type();
     } else {
       *type = kRedisNone;
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -231,7 +231,7 @@ rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &k
       redis::Hash hash_db(storage_, namespace_);
       if (hash_db.ExistValidField(ctx, slice_keys[i], hash_metadata)) {
         if (!s.ok()) return s;
-    *deleted_cnt += 1;
+        *deleted_cnt += 1;
       }
     } else {
       *deleted_cnt += 1;
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -96,7 +96,21 @@ rocksdb::Status Database::GetMetadata(engine::Context &ctx, RedisTypes types, co
   auto s = GetRawMetadata(ctx, ns_key, raw_value);
   *rest = *raw_value;
   if (!s.ok()) return s;
-  return ParseMetadataWithStats(types, rest, metadata);
+
+  s = ParseMetadataWithStats(types, rest, metadata);
+  if (!s.ok()) return s;
+
+  // if type is hash, we still need to check if the all of fields expired.
+  if (metadata->Type() == kRedisHash) {
+    HashMetadata hash_metadata(false);
+    s = hash_metadata.Decode(*raw_value);
+    if (!s.ok()) return s;
+    redis::Hash hash_db(storage_, namespace_);
+    if (!hash_db.ExistValidField(ctx, ns_key, hash_metadata)) {
+      return rocksdb::Status::NotFound("no element found");
+    }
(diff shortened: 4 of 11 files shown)


PragmaTwice:
> And we use the second bit of the flags in metadata to indicate whether encode the expiration into value.

As we specified in https://kvrocks.apache.org/community/data-structure-on-rocksdb#encoding-version-and-data-type, the first and second (and maybe more bits from the MSB) of `flags` is used to represent the "encoding version". `flags` `1 1 0 0 | 0 0 1 0` indicates that the encoding version is 3.

So it should not be used for field expiration. (Also, not all data types need the field expiration feature.)

Instead, you can add a new enum value of data type, or add a new field in the metadata of hash.

git-hulk:
@jjz921024 Thanks for your contributions. As @PragmaTwice mentioned, we cannot put the expired flag in the version field and we also need to adopt the SubkeyFilter in compact_filter.cc to recycle expired subkeys.

jjz921024:
> we also need to adopt the SubkeyFilter in compact_filter.cc to recycle expired subkeys.

@git-hulk Thanks for you reminding. I have added this feature in latest commit.

About `encoding`, my previous thought was to use each bit for a different meaning, rather than as an incremental version.
For example:
  - For all type, the first bit indicate use 64bit or 32bit encoding
  - For hash type, the second bit of the `flags` to indicate whether encode the expiration into value (not affect for other types)

I see this issues #2292. Maybe we should finish it first?

PragmaTwice:
> For all type, the first bit indicate use 64bit or 32bit encoding
For hash type, the second bit of the flags to indicate whether encode the expiration into value (not affect for other types)

No. It's too ad hoc and will introduce additional maintanance effort. (the encoding of HASH will be inconsistent with other types, which brings huge understanding, maintanance cost and additional code logic)

And also, it breaks our whole design of "encoding version", e.g. HASH cannot have encoding version larger than 1.


You can consider these options, as I mentioned before.
> Instead, you can add a new enum value of data type, or add a new field in the metadata of hash.

---

-1 for this change if no further commits.

jjz921024:
@PragmaTwice Got it, i will change it.

src/commands/cmd_hash.cc
```diff
@@ -429,6 +430,238 @@ class CommandHRandField : public Commander {
   bool no_parameters_ = true;
 };
 
+class CommandFiledExpireBase : public Commander {
```
torwig: Small typo. `CommandFiledExpireBase` (Fi**le**d) => `CommandFieldExpireBase`.

src/types/redis_hash.cc
```diff
@@ -418,4 +473,206 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+  buf.append(*value);
+  value->assign(buf.data(), buf.size());
+  return rocksdb::Status::OK();
+}
+
+bool Hash::isMeetCondition(HashFieldExpireType type, uint64_t new_expire, uint64_t old_expire) {
+  if (type == HashFieldExpireType::None) return true;
+  if (type == HashFieldExpireType::NX && old_expire == 0) return true;
+  if (type == HashFieldExpireType::XX && old_expire != 0) return true;
+  // if a filed has no associated expiration, we treated it expiration is infinite
```
torwig: Typo in the same word `filed`.

src/storage/redis_db.cc
```diff
@@ -578,9 +579,20 @@ rocksdb::Status SubKeyScanner::Scan(RedisType type, const Slice &user_key, const
       break;
     }
     InternalKey ikey(iter->key(), storage_->IsSlotIdEncoded());
+    auto value = iter->value().ToString();
+    // filter expired hash feild
```
torwig: Typo in the same word. `feild`. I also tend to make typos in the word `field` :)

src/storage/redis_metadata.h
```diff
@@ -203,6 +204,8 @@ class Metadata {
 class HashMetadata : public Metadata {
  public:
   explicit HashMetadata(bool generate_version = true) : Metadata(kRedisHash, generate_version) {}
+
+  bool IsEncodedFieldExpire() const;
```
torwig: For better understanding, I would rename this function to `IsEncodedFieldHasTTL`/`IsEncodedFieldContainsTTL`.

src/storage/redis_db.cc
```diff
@@ -578,9 +579,20 @@ rocksdb::Status SubKeyScanner::Scan(RedisType type, const Slice &user_key, const
...
+    if (type == kRedisHash && (static_cast<HashMetadata*>(&metadata))->IsEncodedFieldExpire()) {
+      uint64_t expire = 0;
+      rocksdb::Slice data(value.data(), value.size());
+      GetFixed64(&data, &expire);
+      value = data.ToString();
```
torwig: Can we move this assignment after the if-statement since we don't use it if the field has expired?

torwig:
@ jjz921024 Please let me know if I'm wrong. If the data structure contains at least 1 key that can be expired, we set a special bit to true and later check it.

src/types/redis_hash.cc
```diff
@@ -45,7 +46,13 @@ rocksdb::Status Hash::Size(const Slice &user_key, uint64_t *size) {
   HashMetadata metadata(false);
   rocksdb::Status s = GetMetadata(Database::GetOptions{}, ns_key, &metadata);
   if (!s.ok()) return s;
-  *size = metadata.size;
+  if (!metadata.IsEncodedFieldExpire()) {
+    *size = metadata.size;
+  } else {
+    std::vector<FieldValue> field_values;
+    GetAll(user_key, &field_values, HashFetchType::kOnlyKey);
```
torwig: Would it be more effective to **count** available keys (a dedicated function) instead of loading them into memory and getting their size? If we have millions of elements, for example.

src/types/redis_hash.cc
```diff
@@ -58,7 +65,10 @@ rocksdb::Status Hash::Get(const Slice &user_key, const Slice &field, std::string
   rocksdb::ReadOptions read_options;
   read_options.snapshot = ss.GetSnapShot();
   std::string sub_key = InternalKey(ns_key, field, metadata.version, storage_->IsSlotIdEncoded()).Encode();
-  return storage_->Get(read_options, sub_key, value);
+  s = storage_->Get(read_options, sub_key, value);
+  if (!s.ok()) return s;
+  uint64_t expire = 0;
+  return decodeFieldValue(metadata, value, expire);
```
torwig: Is it OK, that `decodeFieldValue` doesn't assign something to `value` if the data structure doesn't have the `METADATA_HASH_FIELD_EXPIRE_MASK` flag?

src/types/redis_hash.h
```diff
@@ -63,9 +65,16 @@ class Hash : public SubKeyScanner {
                        std::vector<std::string> *values = nullptr);
   rocksdb::Status RandField(const Slice &user_key, int64_t command_count, std::vector<FieldValue> *field_values,
                             HashFetchType type = HashFetchType::kOnlyKey);
+  rocksdb::Status ExpireFields(const Slice &user_key, uint64_t expire_ms, const std::vector<Slice> &fields, 
+                               HashFieldExpireType type, bool is_persist, std::vector<int8_t> *ret);
+  rocksdb::Status TTLFields(const Slice &user_key, const std::vector<Slice> &fields, std::vector<int64_t> *ret);
+  static bool IsExpiredField(Metadata &metadata, const Slice &value);
```
torwig: `IsExpiredField` => `IsFieldExpired`

src/types/redis_hash.h
```diff
@@ -63,9 +65,16 @@ class Hash : public SubKeyScanner {
...
 
  private:
   rocksdb::Status GetMetadata(Database::GetOptions get_options, const Slice &ns_key, HashMetadata *metadata);
+  static rocksdb::Status decodeFieldValue(const HashMetadata &metadata, std::string *value, uint64_t &expire);
```
torwig: `decodeFieldValue` => `decodeFieldAndTTL`.

src/types/redis_hash.h
```diff
@@ -63,9 +65,16 @@ class Hash : public SubKeyScanner {
...
+  static rocksdb::Status encodeValueExpire(std::string *value, uint64_t expire);
```
torwig: `encodeValueExpire` => `encodeFieldAndTTL`.

src/types/redis_hash.h
```diff
@@ -63,9 +65,16 @@ class Hash : public SubKeyScanner {
...
+  static bool isMeetCondition(HashFieldExpireType type, uint64_t new_expire, uint64_t old_expire);
```
torwig: Can we change the name of this function to describe more clearly what it checks? What question does it answer?

src/commands/cmd_hash.cc
git-hulk: Good catch!

jjz921024: @torwig thank for you review, I will fix it :)

src/types/redis_hash.cc
jjz921024: done

src/storage/redis_db.cc
jjz921024: done

src/storage/redis_db.cc
jjz921024: done

src/types/redis_hash.h
jjz921024: done

src/storage/redis_metadata.h
jjz921024: I prefer `IsEncodedFieldHasTTL` because it has fewer letters than `IsEncodedFieldContainsTTL`  :)

src/types/redis_hash.h
jjz921024: done

src/types/redis_hash.h
jjz921024: done

src/types/redis_hash.cc
jjz921024: Yes. Inside the `decodeFieldValue`, if doesn't have `METADATA_HASH_FIELD_EXPIRE_MASK` flag will return directly and don't do anything.

jjz921024:
> @ jjz921024 Please let me know if I'm wrong. If the data structure contains at least 1 key that can be expired, we set a special bit to true and later check it.

Yes. Due to the expiration time is encoded in the field, the time complexity of some commands changes from O(1) to O(n), such as `type` and `hlen`. If we can check this flag before execution, then the some commands complexity will remains the same when
 this feature is not used

jjz921024:
> > For all type, the first bit indicate use 64bit or 32bit encoding
> > For hash type, the second bit of the flags to indicate whether encode the expiration into value (not affect for other types)
> 
> No. It's too ad hoc and will introduce additional maintanance effort. (the encoding of HASH will be inconsistent with other types, which brings huge understanding, maintanance cost and additional code logic)
> 
> And also, it breaks our whole design of "encoding version", e.g. HASH cannot have encoding version larger than 1.
> 
> You can consider these options, as I mentioned before.
> 
> > Instead, you can add a new enum value of data type, or add a new field in the metadata of hash.
> 
> -1 for this change if no further commits.

@PragmaTwice Hi, I fix it about `encoding version`. and add a new field in the metadata of hash to indicate whether the expiration time of field is encoded. please review. thank

src/storage/redis_metadata.h
```diff
@@ -202,7 +203,12 @@ class Metadata {
 
 class HashMetadata : public Metadata {
  public:
+  bool is_ttl_field_encoded;
```
PragmaTwice: You can add a new enum type like:
```
enum class SubkeyEncoding {
  RAW = 0;
  WITH_TTL = 1;
};
```

Add then add a `SubkeyEncoding` field into `HashMetadata` instead of a `bool`.

Also, please don't forget to give it a default value.

src/storage/redis_db.cc
```diff
@@ -650,6 +662,17 @@ rocksdb::Status Database::typeInternal(const Slice &key, RedisType *type) {
   if (!s.ok()) return s;
   if (metadata.Expired()) {
     *type = kRedisNone;
+  } else if (metadata.Type() == kRedisHash && (static_cast<HashMetadata*>(&metadata))->is_ttl_field_encoded) {
```
PragmaTwice: `Metadata` can never be `HashMetadata` here, it's wrong C++ code.

src/storage/redis_metadata.h
jjz921024: thank for you hint. I had updated it.

src/storage/redis_db.cc
jjz921024: I've fix here. please review

src/storage/redis_metadata.h
```diff
@@ -200,9 +200,22 @@ class Metadata {
   static uint64_t generateVersion();
 };
 
+enum class HashFieldEncoding : uint8_t {
```
```suggestion
enum class HashSubkeyEncoding : uint8_t {
```

src/storage/redis_metadata.cc
```diff
@@ -334,6 +334,29 @@ bool Metadata::IsEmptyableType() const {
 
 bool Metadata::Expired() const { return ExpireAt(util::GetTimeStampMS()); }
 
+bool HashMetadata::IsTTLFieldEncoded() const { return field_encoding == HashFieldEncoding::WITH_TTL; }
+
+void HashMetadata::Encode(std::string *dst) const {
+  Metadata::Encode(dst);
+  if (IsTTLFieldEncoded()) {
+    PutFixed8(dst, uint8_t(field_encoding));
+  }
```
```suggestion
  PutFixed8(dst, uint8_t(field_encoding));
```

src/storage/redis_metadata.cc
```diff
@@ -334,6 +334,29 @@ bool Metadata::IsEmptyableType() const {
 
 bool Metadata::Expired() const { return ExpireAt(util::GetTimeStampMS()); }
 
+bool HashMetadata::IsTTLFieldEncoded() const { return field_encoding == HashFieldEncoding::WITH_TTL; }
```
```suggestion
bool HashMetadata::IsFieldExpirationEnabled() const { return field_encoding == HashFieldEncoding::WITH_TTL; }
```

src/types/redis_hash.cc
```diff
@@ -418,4 +472,208 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+
+rocksdb::Status Hash::encodeFieldAndTTL(std::string *value, uint64_t expire) {
+  std::string buf;
+  PutFixed64(&buf, expire);
+  buf.append(*value);
+  value->assign(buf.data(), buf.size());
+  return rocksdb::Status::OK();
+}
+
+bool Hash::isMeetCondition(HashFieldExpireType type, uint64_t new_expire, uint64_t old_expire) {
```
PragmaTwice: could you rename the function to a reasonable name?

src/types/redis_hash.cc
```diff
@@ -418,4 +472,208 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+  }
+
+  std::string bytes;
+  metadata.Encode(&bytes);
+  batch->Put(metadata_cf_handle_, ns_key, bytes);
+
+  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+rocksdb::Status Hash::TTLFields(const Slice &user_key, const std::vector<Slice> &fields, std::vector<int64_t> *ret) {
```
PragmaTwice: please return an error status if the current hash object doesn't support field expiration.

src/types/redis_hash.cc
```diff
@@ -418,4 +472,208 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+      ret->emplace_back(-1);
+    } else {
+      ret->emplace_back(int64_t(expire - now));
+    }
+  }
+  return rocksdb::Status::OK();
+}
+
+bool Hash::IsFieldExpired(Metadata &metadata, const Slice &value) {
+  if (!(static_cast<HashMetadata *>(&metadata))->IsFieldExpirationEnabled()) {
```
PragmaTwice: do you ever think about what will happen if metadata is actually not a HashMetadata?

PragmaTwice (changes requested):

src/commands/cmd_hash.cc
```diff
@@ -429,6 +430,228 @@ class CommandHRandField : public Commander {
...
+ protected:
+  Status commonParse(const std::vector<std::string> &args, int start_idx) {
+    int idx = start_idx;
+    CommandParser parser(args, idx);
+    std::string_view expire_flag, num_flag;
+    uint64_t fields_num = 0;
+    while (parser.Good()) {
+      if (parser.EatEqICaseFlag("FIELDS", num_flag)) {
+        fields_num = GET_OR_RET(parser.template TakeInt<uint64_t>());
+        idx += 2;
```
git-hulk: We don't need to record the index here, it's good to use `parse.Good()` to check if there are remaining arguments to be consumed.

src/types/redis_hash.cc
```diff
@@ -418,4 +472,208 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status Hash::ExpireFields(const Slice &user_key, uint64_t expire_ms, const std::vector<Slice> &fields,
```
git-hulk: can remove `is_persist` by checking if `expire_ms` is equal to 0?

src/types/redis_hash.cc
```diff
@@ -418,4 +472,208 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+      ret->emplace_back(-2);
+      continue;
+    }
+
+    InternalKey sub_ikey(ns_key, fields[i], metadata.version, storage_->IsSlotIdEncoded());
+
+    // expire with a pass time
+    if (expire_ms <= now && !is_persist) {
+      batch->Delete(sub_ikey.Encode());
+      ret->emplace_back(2);
```
git-hulk, line 548: Should be `-2`？

src/types/redis_hash.cc
```diff
@@ -418,4 +472,208 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+  for (size_t i = 0; i < keys.size(); i++) {
+    if (!statuses_vector[i].ok() && !statuses_vector[i].IsNotFound()) return statuses_vector[i];
+
+    // no such field exists
+    if (statuses_vector[i].IsNotFound()) {
+      ret->emplace_back(-2);
+      continue;
+    }
+
+    InternalKey sub_ikey(ns_key, fields[i], metadata.version, storage_->IsSlotIdEncoded());
```
git-hulk: Why not use `subkeys[i]` directly?

src/types/redis_hash.cc
jjz921024: Do you mean `HEXPIRE` or `HTTL` a not hash object？I will fix this bug

src/types/redis_hash.cc
jjz921024: Yes. i will change it

src/types/redis_hash.cc
jjz921024: According to https://redis.io/docs/latest/commands/hexpire/ 

>Array reply. For each field:
Integer reply: -2 if no such field exists in the provided hash key, or the provided key does not exist.
Integer reply: 0 if the specified NX | XX | GT | LT condition has not been met.
Integer reply: 1 if the expiration time was set/updated.
Integer reply: 2 when HEXPIRE/HPEXPIRE is called with 0 seconds/milliseconds or when HEXPIREAT/HPEXPIREAT is called with a past Unix time in seconds/milliseconds.

expire with a pass time should return `2`? please let me know if I'm wrong.

git-hulk: Got it, thanks.

src/types/redis_hash.cc
jjz921024: Yes. This function name is not clear enough to convey the intention. Let me think of a better name.

src/types/redis_hash.cc
jjz921024: Thanks for your tip. I'll change this.

src/commands/cmd_hash.cc
jjz921024: Thank for you hint. I replace it with `parser.Remains()`, please review.

src/types/redis_hash.cc
jjz921024: done

src/types/redis_hash.cc
jjz921024: I tried the behavior on redis. Maybe we need to distinguish the `HEXPIREAT` with 0 and `HPERSIST`.

For example:
- ` HEXPIREAT k 0 FIELDS 1 f ` will return `2` (equivalent expire a pass time)
- `HPERSIS k FIELDS 1 f` will return `1` (expiration was removed)

So, we need a flag to distinguish them?

src/storage/redis_metadata.h
```diff
@@ -200,9 +200,22 @@ class Metadata {
   static uint64_t generateVersion();
 };
 
+enum class HashSubkeyEncoding : uint8_t {
+  RAW = 0,
```
git-hulk: RAW => VALUE_ONLY,
WITH_TTL => VALUE_WITH_TTL

src/types/redis_hash.h
git-hulk: Agree with what @torwig mentioned, for:

`decodeFieldAndTTL` should be separated into two methods: `decodeExpireFromValue` and `checkIfFieldExpired`, and `encodeFieldAndTTL` can rename to `encodeExpireToValue`. @torwig What do you think?

src/storage/compact_filter.cc
```diff
@@ -131,6 +132,17 @@ bool SubKeyFilter::Filter(int level, const Slice &key, const Slice &value, std::
...
+    Slice bytes(cached_metadata_);
+    if (!hash_metadata.Decode(bytes).ok()) {
+      LOG(ERROR) << "[compact_filter/subkey] Failed to get hash_metadata"
+                 << ", namespace: " << ikey.GetNamespace() << ", key: " << ikey.GetKey() << ", err: " << s.Msg();
+      return false;
+    }
+    return redis::Hash::IsFieldExpired(hash_metadata, value);
+  }
+
   return IsMetadataExpired(ikey, metadata) || (metadata.Type() == kRedisBitmap && redis::Bitmap::IsEmptySegment(value));
```
git-hulk: Can check if Metadata is expired first

src/types/redis_hash.cc
```diff
@@ -418,4 +472,208 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status Hash::ExpireFields(const Slice &user_key, uint64_t expire_ms, const std::vector<Slice> &fields,
+                                   HashFieldExpireType type, bool is_persist, std::vector<int8_t> *ret) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
```
git-hulk, line 496: Need to add the lock guard to prevent data race for the user key.

src/types/redis_hash.cc
```diff
@@ -418,4 +472,208 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+    sub_keys[i] = InternalKey(ns_key, field, metadata.version, storage_->IsSlotIdEncoded()).Encode();
+    keys.emplace_back(sub_keys[i]);
+  }
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisHash);
+  batch->PutLogData(log_data.Encode());
+
+  // expire special field
+  std::vector<rocksdb::PinnableSlice> values_vector;
```
git-hulk: Don't bring the type into the variable naming, values_vector should be `values`, same as the `statuses`

src/types/redis_hash.cc
git-hulk: I think we should distinguish this in Commands instead of DB implementation.

src/types/redis_hash.cc
jjz921024: done

src/storage/redis_metadata.h
jjz921024: done

src/types/redis_hash.cc
jjz921024: done

src/types/redis_hash.cc
jjz921024: Yes. In redis source code. The hpersist command is a independent function. not reuse `hexpireGenericCommand`.

https://github.com/redis/redis/blob/unstable/src/t_hash.c#L3235-L3236

src/storage/compact_filter.cc
jjz921024: done

jjz921024:
@git-hulk Hi, I had renamed the `decodeFieldAndTTL` to `decodeExpireFromValue`, `encodeFieldAndTTL` to `encodeExpireToValue`.  And rewrite the `IsFieldExpired()` to make it clearer. please review.

src/types/redis_hash.cc
jjz921024: @torwig Yes. maybe we should add a method that check if there is any unexpired field in the hash, and avoid query all the fields

src/commands/cmd_hash.cc
```diff
@@ -429,6 +430,234 @@ class CommandHRandField : public Commander {
...
+        output->append(redis::Integer((now + ttl) / 1000));
+      } else {
+        output->append(redis::Integer(ttl));
+      }
+    }
+    return Status::OK();
+  }
+};
+
+class CommandHPExpireTime : public CommandFieldExpireBase {
```
git-hulk, line 601: The command HExpireTime/HTTL/HPTTL doesn't support NX/XX/LT/GT flags. By the way,  it should be good to remove the ttlExpireExecute/expireFieldExecute and only add a reply helper function like `redis::IntArray`. @torwig @PragmaTwice What do you think?

src/storage/redis_db.cc
```diff
@@ -655,6 +680,22 @@ rocksdb::Status Database::typeInternal(const Slice &key, RedisType *type) {
   if (!s.ok()) return s;
   if (metadata.Expired()) {
     *type = kRedisNone;
+  } else if (metadata.Type() == kRedisHash) {
+    HashMetadata hash_metadata(false);
```
git-hulk, line 752: Can add a `Count` method for the HashDB to avoid storing keys in memory?

src/types/redis_hash.cc
git-hulk: What @torwig means is adding a method called `Count` to count the number of keys instead of storing keys in memory.

src/types/redis_hash.cc
```diff
@@ -418,4 +469,272 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+  }
+
+  std::string bytes;
+  metadata.Encode(&bytes);
+  batch->Put(metadata_cf_handle_, ns_key, bytes);
+
+  return storage_->Write(storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+rocksdb::Status Hash::PersistFields(const Slice &user_key, const std::vector<Slice> &fields, std::vector<int8_t> *ret) {
```
git-hulk: I do not quite understand why we need to copy `ExpireFields` to implement the persist fields.

src/types/redis_hash.cc
```diff
@@ -418,4 +469,272 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+
+    auto iter = util::UniqueIterator(storage_, read_options);
+    for (iter->Seek(prefix_key); iter->Valid() && iter->key().starts_with(prefix_key); iter->Next()) {
+      InternalKey sub_ikey(iter->key(), storage_->IsSlotIdEncoded());
+      auto value = iter->value().ToString();
+      if (field_set.find(sub_ikey.GetSubKey().ToStringView()) == field_set.end()) {
+        encodeExpireToValue(&value, 0);
+        batch->Put(sub_ikey.Encode(), value);
+      }
+    }
```
git-hulk: It's not good to do this because it will be too slow if a hash has a lot of fields. Even having the old hash key not allow to expire fields is better than the current implementation. cc @PragmaTwice @torwig

jjz921024:
Modification:
1.  In `hlen` command, count the number of fields at hash expiration enable,  replace the previous method to avoid loading all of fields into memory.
2. For most `Generic`  type commands (eg: ttl, type, copy, expire, scan...). If the type is hash, we still need to check each fields. If all of fields was expired, the GetMetadata should return not found.
3. In `del` command, if delete a hash object that all of fields expired, so this hash object should be treated as empty and should not affect the deleted_cnt.

src/storage/redis_db.cc
jjz921024: done

src/types/redis_hash.cc
jjz921024: Sorry, I don't understand what you meat. Persist a field is equivalent to setting the expiration of the field to 0.

src/types/redis_hash.cc
jjz921024: Maybe we can add a option in config file, that lets users decide whether to allow old hash to be converted to the new format?
If the user clearly knows that their hash is not large and wants to support field expiration on existing hash object. It is very useful.

git-hulk: I do not prefer doing this since it will introduce an extra maintenance burden. It makes sense only to allow to expire fields in the new hash key from my side.

PragmaTwice: Yeah, please don't perform the convertion automatically .

PragmaTwice: also we should to provide an option for users to let them decide if the newly created hash object is in the new encoding.

jjz921024: @git-hulk @PragmaTwice Hi, I have modified it. Don't encode field ttl on existing hash object. please review.

src/commands/cmd_hash.cc
```diff
@@ -429,6 +430,234 @@ class CommandHRandField : public Commander {
...
+        return parser.InvalidSyntax();
+      }
+    }
+
+    auto remains = parser.Remains();
+    auto size = args.size();
+    if (remains != fields_num) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+
```
PragmaTwice, line 476: Is `FIELDS 2 NX x y` a valid input?

src/storage/redis_metadata.cc
```diff
@@ -334,6 +334,25 @@ bool Metadata::IsEmptyableType() const {
...
+}
+
+rocksdb::Status HashMetadata::Decode(Slice *input) {
+  if (auto s = Metadata::Decode(input); !s.ok()) {
+    return s;
+  }
+
+  if (input->size() >= 1 && !GetFixed8(input, reinterpret_cast<uint8_t *>(&field_encoding))) {
+    return rocksdb::Status::InvalidArgument(kErrMetadataTooShort);
+  }
```
```suggestion
  if (input->size() >= 1) {
    if (!GetFixed8(input, reinterpret_cast<uint8_t *>(&field_encoding))) {
      return rocksdb::Status::InvalidArgument(kErrMetadataTooShort);
    }
    if (uint8_t(field_encoding) > 1) {
      return rocksdb::Status::InvalidArgument("unexpected subkey encoding version");
    }
  }
```

src/commands/cmd_hash.cc
jjz921024: No, the input `FIELDS 2 NX x y` will get  `ERR wrong number of arguments`. But the `FIELDS 2 NX x` is a valid input. it means is to perform on a field called `NX` and `x`.

src/types/redis_hash.cc
```diff
@@ -418,4 +502,278 @@ rocksdb::Status Hash::RandField(const Slice &user_key, int64_t command_count, st
...
+  rocksdb::Status s = GetMetadata(GetOptions{ss.GetSnapShot()}, ns_key, &metadata);
+  if (!s.ok() && !s.IsNotFound()) return s;
+  if (s.IsNotFound()) {
+    ret->resize(fields.size(), -2);
+    return rocksdb::Status::OK();
+  }
+
+  // we don't support encode ttl on existing hash object
+  if (!metadata.IsFieldExpirationEnabled()) {
+    return rocksdb::Status::NotSupported("can't expire fields on existing hash object");
```
PragmaTwice:
```suggestion
    return rocksdb::Status::NotSupported("can't expire fields on hash object whose field expiration feature is disabled");
```

What is "existing hash object"? IMHO it's a very vague message.

PragmaTwice (commented):
As I mentioned before, we can add a config option (e.g. `hash-field-expiration yes/no`) for users to decide if they want to enable field expiration.

Also our testing can benefit from this option since we can cover the old hash encoding.

src/commands/cmd_hash.cc
PragmaTwice: Ahh I see.

jjz921024:
> As I mentioned before, we can add a config option (e.g. `hash-field-expiration yes/no`) for users to decide if they want to enable field expiration.
> 
> Also our testing can benefit from this option since we can cover the old hash encoding.

@PragmaTwice Thank for you tip, I have finished this feature. please review.

src/storage/redis_db.cc
```diff
@@ -558,10 +611,24 @@ rocksdb::Status SubKeyScanner::Scan(RedisType type, const Slice &user_key, const
...
   if (!s.ok()) return s;
 
+  // for hash type, we should filter expired field if encoding is with_ttl
+  bool is_encoding_field_ttl = false;
+  if (metadata.Type() == kRedisHash && !rest.empty()) {
+    uint8_t field_encoding = 0;
+    if (!GetFixed8(&rest, &field_encoding)) {
+      return rocksdb::Status::InvalidArgument();
+    }
+    if (field_encoding == 1) {
```
PragmaTwice: Please use `HashSubkeyType` instead of hard-coding (`1`) here.

tests/gocase/unit/type/hash/hash_test.go
```diff
@@ -936,3 +944,235 @@ func TestHashWithAsyncIODisabled(t *testing.T) {
...
+func TestExpireFieldOnExistingHash(t *testing.T) {
+	srv := util.StartServer(t, map[string]string{
+		"hash-field-expiration": "no",
+	})
+	defer srv.Close()
+
+	rdb := srv.NewClient()
+	defer func() { require.NoError(t, rdb.Close()) }()
+
+	ctx := context.Background()
```
PragmaTwice: Just one test case is not enough.

For all existing test function e.g.
```
func TestXXX() { ... }
```
we can change them to:
```
func TestXXXImpl(field_expiration_enabled) {
  startServer({"hash-field-expiration": field_expiration_enabled})
  ... // original code
}

func TestXXX() {
  TestXXXImpl(true)
  TestXXXImpl(false)
}
```

PragmaTwice (changes requested):
See comments : )

src/storage/redis_db.cc
jjz921024: Thank for you point it up. I had fixed it.

jjz921024:
@git-hulk @PragmaTwice @torwig Sorry to bother your. I have changed the code as review required. May I ask your review it again? I think this is a very useful feature and hope it can be merge to kvrocks. 
If there are any changes, please let me know. Thank.

jjz921024:
I had fixed the gofmt error and passed all CI in my fork repo.

https://github.com/jjz921024/kvrocks/actions/runs/10337719500

src/config/config.h
```diff
@@ -167,6 +167,9 @@ struct Config {
   int json_max_nesting_depth = 1024;
   JsonStorageFormat json_storage_format = JsonStorageFormat::JSON;
 
+  // hash
```
torwig: The variable name is self-explanatory so you can remove this comment. Otherwise, it would be best if you elaborated on the goal of this variable/option.

PragmaTwice:
Seems we need to solve the conflict first?

src/config/config.h
jjz921024: Yes. I had polished the comment of here

jjz921024:
> Seems we need to solve the conflict first?

rebase this pr on the latest commit of unstable branch.

jjz921024:
Seems there are failure in CI.  I will check it up

src/storage/redis_metadata.cc
```diff
@@ -334,6 +334,30 @@ bool Metadata::IsEmptyableType() const {
...
+rocksdb::Status HashMetadata::Decode(Slice *input) {
+  if (auto s = Metadata::Decode(input); !s.ok()) {
+    return s;
+  }
+
+  if (input->size() >= 1) {
+    if (!GetFixed8(input, reinterpret_cast<uint8_t *>(&field_encoding))) {
+      return rocksdb::Status::InvalidArgument(kErrMetadataTooShort);
+    }
+    if (uint8_t(field_encoding) > 1) {
```
PragmaTwice: we should not hardcode `1` here.

jjz921024: Yes. fixed

git-hulk:
@jjz921024 I will take another pass in a few days, thanks for your contribution.

ltagliamonte:
@jjz921024 @git-hulk @PragmaTwice I'm looking forward to this feature, is there anything I can help with to help here?

PragmaTwice:
> @jjz921024 @git-hulk @PragmaTwice I'm looking forward to this feature, is there anything I can help with to help here?

Thank you for your input. I'll take a look at this weekend.

jjz921024:
rebase this branch on the HEAD

src/storage/redis_db.cc
```diff
@@ -96,7 +96,21 @@ rocksdb::Status Database::GetMetadata(engine::Context &ctx, RedisTypes types, co
...
+  // if type is hash, we still need to check if the all of fields expired.
+  if (metadata->Type() == kRedisHash) {
+    HashMetadata hash_metadata(false);
+    s = hash_metadata.Decode(*raw_value);
+    if (!s.ok()) return s;
+    redis::Hash hash_db(storage_, namespace_);
+    if (!hash_db.ExistValidField(ctx, ns_key, hash_metadata)) {
+      return rocksdb::Status::NotFound("no element found");
+    }
+  }
```
PragmaTwice: This will make almost all operation of HASH inefficient, we should consider how to do this in another way.

PragmaTwice: It requires at least O(N) IO accesses for every HASH operation. And if you think about it: not all operations require to check that all elements are expired before the actual operation.

jjz921024: It seems that only `hset` and `hmset` does not need to check the expire of the field？But we don't know the the command type in the `GetMetadata()`

aleksraiden:
Awesome work, @jjz921024

Maybe we need some benchmark - before include this feature and after? Тo test the impact of new logic on basic HSET's command (and prevent or minimize it)?

jjz921024:
@git-hulk @ltagliamonte Hi, I don't know how to commit code on the new pr #2717 , so I just keep working here.

For the first point you mentioned on https://github.com/apache/kvrocks/pull/2717#issuecomment-2585068688, I had finish. please review.  

And the second point, do you mean that the `ExistValidField()` function should contain the following code? This code is repeated many times. I think the `ExistValidField()` is a dedicated function to check if it has any valid fields.
```
HashMetadata hash_metadata(false);
s = hash_metadata.Decode(rocksdb::Slice(pin_values[i].data(), pin_values[i].size()));
if (!s.ok()) continue;
redis::Hash hash_db(storage_, namespace_);
```

jjz921024:
> Awesome work, @jjz921024
> 
> Maybe we need some benchmark - before include this feature and after? Тo test the impact of new logic on basic HSET's command (and prevent or minimize it)?

@aleksraiden Hi, After optimization of https://github.com/apache/kvrocks/pull/2402/commits/473db0c9ec5cf9f38379eda0d11f01e61ba5f90a, the expire fields should have no effect on HSET command. Should we add some bench test for read command of hash type (like hgetall)?

aleksraiden:
> > Awesome work, @jjz921024
> > Maybe we need some benchmark - before include this feature and after? Тo test the impact of new logic on basic HSET's command (and prevent or minimize it)?
> 
> @aleksraiden Hi, After optimization of [473db0c](https://github.com/apache/kvrocks/commit/473db0c9ec5cf9f38379eda0d11f01e61ba5f90a), the expire fields should have no effect on HSET command. Should we add some bench test for read command of hash type (like hgetall)?

I have a test by standard redis-bench yesterday (before your commit) - and at quick look is there NO any impact to performance. But standard test looks only to GET/HSET as I see. 

Great job!

git-hulk:
> @git-hulk @ltagliamonte Hi, I don't know how to commit code on the new pr https://github.com/apache/kvrocks/pull/2717 , so I just keep working here.

@jjz921024 Perhaps you can suggest your change in the new PR due to this PR is overload for discussing. What do you think?

> I have a test by standard redis-bench yesterday (before your commit) - and at quick look is there NO any impact to performance. But standard test looks only to GET/HSET as I see.

@aleksraiden Great!

jjz921024:
> Perhaps you can suggest your change in the new PR due to this PR is overload for discussing. What do you think?

I'm pleasure to do it and I will try my best to finish this feature if i can.

git-hulk:
> I'm pleasure to do it and I will try my best to finish this feature if i can.

@jjz921024 Thank you! We will take care of PR #2717 to keep the original commit information while merging.

PragmaTwice:
Refer to #3436.

PR #2406, opened 2024-07-13
chore(resp): extract RESP functions from Connection

PragmaTwice:
Currently, many RESP functions are tightly coupled with the Connection class type, which is unnecessary.


torwig:
@PragmaTwice Yes, the RESP functions are more about the Redis protocol than the connection between a client and the server.

git-hulk (approved):

torwig (approved):

mapleFU (approved):

Issue #2426, opened 2024-07-15
Tracking Issue: Support vector search in KQIR


Beihao-Zhou:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

This includes:
- [x] Implementing vector indexing algorithm, i.e. HNSW & Adding vector indexing support for KQIR Indexer - #2368
- [x] Plan operator & executor for vector indexing - #2434 
- [x] Supporting relevant syntax parsing in front-end languages (SQL, RediSearch query) - #2450 
- [x] Add Optimization Pass, including Index Selection & KnnExpr transformation - #2466 
- [x] Add command to support Hnsw index creation - #2477  
- [x] Documentation on Kvrocks Search - https://github.com/apache/kvrocks-website/pull/241

### Further Steps
- [ ] Improvement for HNSW construction - #2398 
- [ ] Create a const version of HnswIndex that takes a const pointer to HnswVectorFieldMetadata - https://github.com/apache/kvrocks/pull/2434#discussion_r1687707134
- [ ] Support Hybrid Query
- [ ] Improve consistency

### Solution

- #2287
- #2316 

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PragmaTwice:
Would you like to add the encoding format of vector index to https://kvrocks.apache.org/community/kvrocks-search-index-encoding/ when you have time?

PR #2434, opened 2024-07-18
feat(search): Hnsw Vector Search Plan Operator & Executor

Beihao-Zhou:
This PR is part of #2426 

## Summary

- [x] "Pure" KNN Plan Operator + Executor (i.e. search on HNSW graph directly on disk)
- [x] "Pure" Range Query Plan Operator + Executor
  1. KNN search to get `k` nearest candidates as a mini-batch initialization, ensuring each one of them is within the expected range
        - If there has been candidates out of the range, then return `End`;
  2. Expanding results from step 1 by fetching their unvisited neighbours
        - Return results which are within the range and treat them as the starting candidates for the next batch
 
    Ref: Idea from [Pgvector Hnsw Scan Code - bool hnswgettuple(IndexScanDesc scan, ScanDirection dir)](https://github.com/pgvector/pgvector/compare/hnsw#diff-731a6344b65391d9a4dd0683f349111d030f5f7f57ed1de6fe407b6028a03b3bR110)
    
## Next
-  Hybrid Query with Filter (Potential Solution)
      1. KNN query as usual
      2. Post-process the results by selecting only those results returned by the index that matches the query filter 
      
    Note that since both "Pure" Range Query Executor and "Pure" KNN Search Executor are based on disk, "Pure" Range Query Executor cannot be reused for hybrid query. I plan to add the new hybrid query executor after implementing the `ir` expression for HNSW (e.g. reuse`FilterExecutor` for `HnswRangeQueryExpr`?). Also, I currently don't have a clear idea on how to continue fetching more nearest vectors after some vectors are filtered for hybrid queries. So for hybrid query, I'll make changes in future PRs after the expression layer is more clear to me. 
          
## Resources
- [Redis Search and Query - Vectors](https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/vectors/)
- [HQANN: Efficient and Robust Similarity Search for Hybrid Queries with Structured and Unstructured Constraints](https://arxiv.org/pdf/2207.07940)
- [Filtered − DiskANN: Graph Algorithms for Approximate Nearest Neighbor Search with Filters](https://harsha-simhadri.org/pubs/Filtered-DiskANN23.pdf)
- [Pgvector Similar Discussion](https://github.com/pgvector/pgvector/issues/259#issuecomment-1796976820)


Beihao-Zhou:
Hiii @PragmaTwice, once we discussed the potential solutions for vector search query executor on Slack, this PR is attempting to solve it. I'm wondering if the solution above sounds good to you as an initial implementation, and whether you see any room for improvement. Thanks!

PragmaTwice:
Yeah I think it's good to be an initial implementation. We can gradually refactor it to a more iterative way.

src/search/ir_plan.h
```diff
@@ -96,6 +97,25 @@ struct TagFieldScan : FieldScan {
...
+
+  HnswVectorFieldKnnScan(std::unique_ptr<FieldRef> field, kqir::NumericArray vector, uint16_t k)
+      : FieldScan(std::move(field)), vector(std::move(vector)), k(k) {}
+
+  std::string_view Name() const override { return "HnswVectorFieldKnnScan"; };
+  std::string Content() const override { return kqir::MakeValue<kqir::NumericArray>(vector).ToString(); };
+  std::string Dump() const override {
+    return fmt::format("hnsw-vector-knn-scan {}, {}", field->name,
+                       kqir::MakeValue<kqir::NumericArray>(vector).ToString());
+  }
```
PragmaTwice:
```suggestion
  std::string Content() const override {
    return fmt::format("[{}], {}", util::StringJoin(vector, [](auto v) { return std::to_string(v); }), k);
  };
  std::string Dump() const override {
    return fmt::format("hnsw-vector-knn-scan {}, {}", field->name, Content());
  }
```

Beihao-Zhou:
This PR is ready for review! @PragmaTwice 
I probably want to add the hybrid executor after implementing the IR expressions (See the Next section in the first conversation block) if this sounds good to you. <3

PragmaTwice:
> This PR is ready for review! @PragmaTwice I probably want to add the hybrid executor after implementing the IR expressions (See the Next section in the first conversation block) if this sounds good to you. <3

Yeah we can then add new nodes in IR and support these new field scan operators in optimizer passes.

e.g. for RediSearch query `@year:[2020 2022] @v:[VECTOR_RANGE ...]`, it's possible to generate a query plan that looks like `filter (2020 <= @year and @year <= 2022) from (hnsw-vector-range-scan @v, ...)`.

As for which query plan is the best, we can leave it to the cost model to determine.

src/search/executors/hnsw_vector_field_knn_scan_executor.h
```diff
@@ -0,0 +1,76 @@
...
+
+// TODO(Beihao): Add DB context to improve consistency and isolation - see #2332
+struct HnswVectorFieldKnnScanExecutor : ExecutorNode {
+  HnswVectorFieldKnnScan *scan;
+  redis::LatestSnapShot ss;
+  bool initialized = false;
+
+  IndexInfo *index;
+  redis::SearchKey search_key;
+  redis::HnswVectorFieldMetadata field_metadata;
```
PragmaTwice, line 45: seems here we can just keep a pointer e.g. `const redis::HnswVectorFieldMetadata *field_metadata`.

PragmaTwice (commented):
The rest looks good to me. Thank you!

src/search/executors/hnsw_vector_field_knn_scan_executor.h
Beihao-Zhou: Oh this was because `HnswIndex` constructor does not support `const HnswVectorFieldMetadata*` . 
```
HnswIndex(const SearchKey& search_key, HnswVectorFieldMetadata* vector, engine::Storage* storage);
```

`HnswIndex` needs to modify `HnswVectorFieldMetadata*`  in other functions, so I did copy here. Do you have ideas on how we can improve it?

PragmaTwice: Let me think, if the metadata is modified in a certain vector query, it seems that the global metadata of the vector field (in `server->index_mgr->index_map`) will not change. Will this cause any problems?

Beihao-Zhou: `HnswVectorFieldMetadata*` is only modified when the node is inserted/deleted in a higher layer than all other nodes, so `num_levels` in `HnswVectorFieldMetadata : IndexFieldMetadata` is modified. So the affected field is `server->index_mgr->index_map[<index_key>]->fields[<field>]->metadata`. 

In `indexer.cc`, we do 
```
auto *metadata = iter->second.metadata.get();
if (auto vector = dynamic_cast<HnswVectorFieldMetadata *>(metadata)) {
    GET_OR_RET(UpdateHnswVectorIndex(key, original, current, search_key, vector));
}
```
So `*metadata` changes correspondingly. 

But in this PR, since we don't modify `metadata`, I guess it will not cause actual problems by copying (maybe with consistency issue but this would be fixed later aligning with #2310 , and the expensive copy because of the large size of `HnswIndex` but I can fix that right after the PR if the solution using `static` member variable sounds good to you <3).

PragmaTwice: I see what you mean.

It's good for now, but we can enhance it later. For example, we could create a const version of HnswIndex that takes a const pointer to HnswVectorFieldMetadata.

PragmaTwice (dismissed):

src/search/executors/hnsw_vector_field_knn_scan_executor.h
Beihao-Zhou: That sounds good!
I wanted to overload with a const version, but there are some nested calls also asking for `HnswVectorFieldMetadata*` as a parameter, so eventually didn't implement it to avoid making this PR look too messy. I'll take a note in tracking issue #2426 and improve it in the future.

PragmaTwice (approved):

PR #2444, opened 2024-07-24
feat(config): allow to change the max length of the bulk string

### LINKED ISSUE #2399: Extend limits of max value size to > 1Gb
### Search before asking

### Motivation

I like kvrocks and I'm interested in storing values that are larger than 256M, rocks db supports larger values and I'd like to take advantage of that capability.

### Solution

I don't yet know where the limitations are and would appreciate clues so that I can investigate further.

git-hulk:
This closes https://github.com/apache/kvrocks/issues/2399.

Currently, the proto's max length of the bulk string is hard coded to 512MiB, and it's impossible to create a string value more than that. And Redis allows us to change this value by the configuration `proto-max-bulk-len`, so we also just add this to align the behavior.


torwig (approved):

kvrocks.conf
```diff
@@ -61,6 +61,11 @@ cluster-enabled no
 # Default: no
 repl-namespace-enabled no
 
+# By default, the max length of bulk string is limited to 512MB. If you want to
+# change this limit to a different value(must >= 1MiB), you can use the following configuration.
+#
+# proto-max-bulk-len 536870912
```
PragmaTwice, line 67: Should it be in MB or KB?

git-hulk: Yes, I had thought about this. However the configuration's name didn't indicate the unit, so I think it will be better to use bytes.

PragmaTwice: You can use `ParseSizeAndUnit` [here](https://github.com/apache/kvrocks/blob/unstable/src/common/parse_util.cc#L35), or you can rename it to proto-max-bulk-len-mb.

git-hulk: Currently, the config rewrite will change it to the number of bytes. I have created an issue to support this: https://github.com/apache/kvrocks/issues/2445, and we can disallow using `unit` as the suffix of the configuration.

ghost (approved):

PragmaTwice (approved):

Issue #2414, opened 2024-07-14
Support more script flags in lua script shebang


PokIsemaine:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

https://redis.io/docs/latest/develop/interact/programmability/lua-api/#script_flags
https://redis.io/docs/latest/develop/interact/programmability/eval-intro/#eval-flags
https://redis.io/docs/latest/develop/interact/programmability/functions-intro/#function-flags

It seems that only the `name` flag of redis function is supported now, and we can support the parsing of some other flags. For example `no-writes`, `allow-cross-slot-keys`, etc.

On the one hand, there are already existing requirements issue: https://github.com/apache/kvrocks/issues/1884 , https://github.com/apache/kvrocks/issues/2133

On the other hand, for the optimization function of Lua script in the future, we may need to support some special flags like dragonflydb to enable optimization. Relevant information:

https://www.dragonflydb.io/docs/managing-dragonfly/scripting

https://www.dragonflydb.io/blog/leveraging-power-of-lua-scripting

https://github.com/apache/kvrocks/issues/2162

### Solution

I'm thinking about how to implement it. Generally speaking, the process of parsing and obtaining flag should be more general.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

PokIsemaine:
I plan to provide support for `no-writes`, `no-cluster`, `allow-cross-slot-keys` in the next few days.

PokIsemaine:
Next I plan to support the `disable-atomicity` flag, which is from dragonflydb, and scripts with `disable-atomicity` are allowed to execute non-atomistically.

* The script will use a separate instance of lua, similar to the current read-only script.
* Commands in a script will collectively use a Context (waiting for the https://github.com/apache/kvrocks/pull/2332 to complete), which provides some isolation (about the snapshot isolation level).
* When this non-atomic script is executed, other non-atomic scripts and common commands (except `MULTI/EXEC`) can also be executed at the same time.

Do you have any suggestions or discussions that you would like to discuss? For example, whether the flag makes sense, and some suggestions for implementation.

As far as the examples provided by dragonfly are concerned, I think `read-only` will solve most of the requirements, and `disable-atomicity` is the more permissive version.
```
#!lua flags=disable-atomicity,allow-undeclared-keys
local cursor = "0"
repeat
[...]

Issue #2133, opened 2024-03-02
Add support for `flags` parameter of `redis.register_function`


i18nsite:
### Search before asking

https://redis.io/docs/interact/programmability/functions-intro/#function-flags

<img width="1199" alt="image" src="https://github.com/apache/kvrocks/assets/145643935/4a6def33-27e1-4e38-ac63-4922369d4290">

### Motivation

compatibility

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

VasuDevrani:
@PragmaTwice 
- `flags` parameter in `redis.register_function`
- will allow `no-writes` flag 
- do we need to support other flags? like `allow-oom`, `admin`

Possible implementation:
1. `RedisRegisterFunction` parses `flags` table and checks for `no-writes` [src/storage/scripting.cc](https://github.com/apache/kvrocks/blob/unstable/src/storage/scripting.cc)
2. if invalid flag, raise error, else store as constant (or something else)
3. check function flags before execution `FunctionCall` [src/storage/scripting.cc](https://github.com/apache/kvrocks/blob/unstable/src/storage/scripting.cc)

```cc
if (read_only && !(flags & redis::kFunctionNoWrites)) {
  return {Status::NotOK, "Can not execute a function with write flag using fcall_ro."};
}
```

PR #2446, opened 2024-07-25
feat(script): Support script flags of Eval script and Function

### LINKED ISSUE #1884: add support for allow-cross-slot-keys for lua function
### Search before asking

### Motivation

lua function can read/write cross slot keys 

### Solution

https://developer.redis.com/create/redis-functions/
like redis
image

PokIsemaine:
fix: https://github.com/apache/kvrocks/issues/1884, https://github.com/apache/kvrocks/issues/2133
a part of: https://github.com/apache/kvrocks/issues/2414

This PR is designed to support three script flags within Eval Script and Function: `no-writes`, `no-cluster`, and `allow-cross-slot-keys`.

**Implementation Details**
Focus on `ScriptRunCtx`. I use it to pass context between Lua and C++. Before executing the Lua script, `SaveOnRegistry` stores the parsed flags in `REGISTRY_SCRIPT_RUN_CTX_NAME`. During the execution of the Lua script, `GetFromRegistry` retrieves the flags of the currently executing script. After the script execution ends, it should be set to `nil`.

For APIs like `EVAL`, `SCRIPT LOAD`, and `FUNCTION LOAD`, the flags parsed from the Eval Script will be stored in Lua's global variable `f_<sha>_flags_`. The flags parsed by FUNCTION `register_function()` will be stored in the Lua global variable `_registered_flags_<funcname>`.

**Testing**
* Distinguish between the default values with and without `#!` in the EVAL Script.
* Function’s `name=` and EVAL Script’s `flags=` which will not appear simultaneously.
* FUNCTION requires `#!lua` and `name=`.
* Even with `allow-cross-slot-keys`, pre-defined Keys are still prohibited from crossing Slots. 
Different slots on different nodes are still prohibited from crossing Slots. However, slots on the same node are allowed to cross slots when this flag is enabled (generally not recommended).
* The `read_only` of `EVAL_RO`, `EVALSHA_RO`, and `FCALL_RO` will be added to the script flags, and even if the `no-writes` flag is not set, checks will still be performed.

**Note**
Since I am not very familiar with Lua, the rationality and safety of related operations need to be reviewed.


mapleFU:
Just ping me if ready for code review

src/storage/scripting.cc
```diff
@@ -127,11 +127,6 @@ void LoadFuncs(lua_State *lua, bool read_only) {
   lua_pushcfunction(lua, RedisStatusReplyCommand);
   lua_settable(lua, -3);
 
-  /* redis.read_only */
-  lua_pushstring(lua, "read_only");
-  lua_pushboolean(lua, read_only);
-  lua_settable(lua, -3);
```
PragmaTwice, line 133: It's used for users to check if the script is running on a read-only lua vm (we have one global lua vm and several per-worker lua vms, and the latter can only run read-only scripts).

src/storage/scripting.cc
```diff
@@ -243,6 +238,9 @@ int RedisRegisterFunction(lua_State *lua) {
 
   // set this function to global
   std::string name = lua_tostring(lua, 1);
+  if (argc == 3) {
+    lua_setglobal(lua, (REDIS_LUA_REGISTER_FUNC_FLAGS_PREFIX + name).c_str());
+  }
```
PragmaTwice: what happens if `argc > 3`?

src/storage/scripting.cc
PokIsemaine: Yes, I noticed this and used the `read_only` parameter in the function to get a read-only lua vm.
```
auto lua = read_only ? conn->Owner()->Lua() : srv->Lua();
```
For checks during script execution, I store flags in `ScriptRunCtx`, which supports more flags, not just `read-only`.
I'll go into more detail tonight and you can start reviewing it tonight or tomorrow.

src/storage/scripting.cc
```diff
@@ -1470,8 +1488,73 @@ Status CreateFunction(Server *srv, const std::string &body, std::string *sha, lu
...
 
+[[nodiscard]] Status ExtractLibNameFromShebang(const std::string &shebang, std::string &libname) {
+  static constexpr const char *shebang_prefix = "#!lua";
+  static constexpr const char *shebang_libname_prefix = "name=";
+
+  if (!util::HasPrefix(shebang, shebang_prefix)) {
+    return {Status::NotOK, "Expect shebang prefix \"#!lua\" at the beginning of the first line"};
+  }
+
+  if (auto pos = shebang.find(shebang_libname_prefix, strlen(shebang_prefix)); pos != std::string::npos) {
```
PragmaTwice: here we need to consider more to prevent invalid input, e.g. `#!luaname=..`, `#!lua xxxname=..`, or valid input with a key that's currently undefined, e.g. `#!lua name=a key1=b`.

src/storage/scripting.cc
```diff
@@ -1470,8 +1488,73 @@ Status CreateFunction(Server *srv, const std::string &body, std::string *sha, lu
...
+
+  return {Status::NotOK, "Expect a library name in the Shebang statement"};
+}
+
+[[nodiscard]] Status ExtractFlagsFromShebang(const std::string &shebang, uint64_t &flags) {
+  static constexpr const char *shebang_prefix = "#!lua";
+  static constexpr const char *shebang_flags_prefix = "flags=";
+
+  if (auto pos = shebang.find(shebang_flags_prefix, strlen(shebang_prefix)); pos != std::string::npos) {
+    auto flags_content = shebang.substr(pos + strlen(shebang_flags_prefix));
```
PragmaTwice: ditto

src/storage/scripting.cc
```diff
@@ -409,13 +384,42 @@ Status FunctionCall(redis::Connection *conn, const std::string &name, const std:
...
+        return {Status::NotOK, "Unexpected function flag: " + flag};
+      }
+      lua_pop(lua, 1);
+    }
+  }
+  lua_pop(lua, 1);
+
+  ScriptRunCtx script_run_ctx;
+  script_run_ctx.flags = function_flags;
+  SaveOnRegistry(lua, REGISTRY_SCRIPT_RUN_CTX_NAME, &script_run_ctx);
```
PragmaTwice: I think we can parse these flags in function register phase instead of calling phase.

PragmaTwice: also `SaveOnRegistry` seems not needed then.

src/storage/scripting.cc
```diff
@@ -1470,8 +1488,73 @@ Status CreateFunction(Server *srv, const std::string &body, std::string *sha, lu
...
+        return s;
+      }
+      script_flags |= shebang_flags;
+    } else {
+      // scripts without #! can run commands that access keys belonging to different cluster hash slots,
+      // but ones with #! inherit the default flags, so they cannot.
+      script_flags |= ScriptFlags::kScriptAllowCrossSlotKeys;
+    }
+  }
+  srv->CacheScriptFlags(*sha, script_flags);
```
PragmaTwice: why is it stored in server instead of lua global variables?

we already have a cache mechanism for `CreateFunction` [here](https://github.com/apache/kvrocks/blob/236ec7a3257a5b03d6006e06b672f64f5337bdcf/src/storage/scripting.cc#L620), i.e. if the script is loaded in the lua vm (recall that we have multiple vms) `CreateFunction` will not be called.

PragmaTwice (commented):
I think inside the lua vm we can have such global variables:
```
f_<sha>_flags  ->  <flags (encoded to integer)>
__redis_registered_flags_<func name>  ->  <flags (encoded to integer)>
```
and we can construct ctx quickly when the lua function is called.

src/storage/scripting.cc
PokIsemaine: I think both are possible. If it is stored in server, different Lua vm can share it. There is no need to create a global variable for each vm.

> if the script is loaded in the lua vm (recall that we have multiple vms) CreateFunction will not be called.

I don’t quite understand this. If the script has been loaded, then CreateFunction has been called for the first time and the flag has also been stored. That is, CreateFunction will be triggered at least once

Can you give a more detailed example or point out the key code directly? Maybe I've overlooked something.

src/storage/scripting.cc
PokIsemaine: > also `SaveOnRegistry` seems not needed then.

If removed, how to obtain flags when the Lua script `redis.call` is executed?

src/storage/scripting.cc
PragmaTwice: I think we can try to avoid adding more members to class Server if possible. (it's better for maintanace that it can only be accessed in the lua vm rather than almost all the codebase.)

Also in lua vm we can ensure it's accessed from just one thread (since it's not thread-safe), and it will be deleted after the lua vm is reset.

And I think it's better to keep consistence between lua functions and scripts.

src/storage/scripting.cc
PragmaTwice: Ahh I see. You can keep it then.

PragmaTwice: Don't forget to remove it after the function call is finished.

src/storage/scripting.cc
PokIsemaine: Okay, that sounds good, I'll try to redesign it

src/server/server.cc
```diff
@@ -1682,6 +1682,16 @@ Status Server::ScriptSet(const std::string &sha, const std::string &body) const
   return storage->WriteToPropagateCF(func_name, body);
 }
 
+void Server::CacheScriptFlags(const std::string &sha, uint64_t flags) { script_flags_cache_.try_emplace(sha, flags); }
+
+[[nodiscard]] Status Server::GetScriptFlags(const std::string &sha, uint64_t &flags) const {
+  if (script_flags_cache_.count(sha)) {
+    flags = script_flags_cache_.at(sha);
```
mapleFU:
```suggestion
  auto iter = script_flags_cache_.find(sha);
  if (iter != script_flags_cache_.end()) {
    flags = iter->second;
```

src/server/server.cc
```diff
@@ -1682,6 +1682,16 @@ Status Server::ScriptSet(const std::string &sha, const std::string &body) const
   return storage->WriteToPropagateCF(func_name, body);
 }
 
+void Server::CacheScriptFlags(const std::string &sha, uint64_t flags) { script_flags_cache_.try_emplace(sha, flags); }
+
+[[nodiscard]] Status Server::GetScriptFlags(const std::string &sha, uint64_t &flags) const {
```
mapleFU: Using `StatusOr<uint64_t>` or `uint64_t* flags`?

src/server/server.h
```diff
@@ -341,6 +343,8 @@ class Server {
   std::mutex last_random_key_cursor_mu_;
 
   std::atomic<lua_State *> lua_;
+  // The cache of flag is cached when the script is created and cleared when the script is flushed.
```
mapleFU: Maybe we should denote this is sha -> Flag, and changing value to `ScriptFlags`?

src/storage/scripting.h
```diff
@@ -101,4 +103,66 @@ void SHA1Hex(char *digest, const char *script, size_t len);
...
 int RedisMathRandomSeed(lua_State *l);
 
+/// ScriptFlags turn on/off constraints or indicate properties in Eval scripts and functions
+///
+/// Note: The default for Eval scripts are different than the default for functions(default is 0).
+/// As soon as Redis sees the #! comment, it'll treat the script as if it declares flags, even if no flags are defined,
+/// it still has a different set of defaults compared to a script without a #! line.
+/// Another difference is that scripts without #! can run commands that access keys belonging to different cluster hash
+/// slots, but ones with #! inherit the default flags, so they cannot.
+enum ScriptFlags : uint64_t {
```
mapleFU: enum class?

src/storage/scripting.cc
```diff
@@ -1470,8 +1488,73 @@ Status CreateFunction(Server *srv, const std::string &body, std::string *sha, lu
...
+      return {Status::NotOK, "Expect a valid library name in the Shebang statement"};
+    }
+    return Status::OK();
+  }
+
+  return {Status::NotOK, "Expect a library name in the Shebang statement"};
+}
+
+[[nodiscard]] Status ExtractFlagsFromShebang(const std::string &shebang, uint64_t &flags) {
+  static constexpr const char *shebang_prefix = "#!lua";
```
mapleFU: string_view can be used?

src/storage/scripting.cc
```diff
@@ -1470,8 +1488,73 @@ Status CreateFunction(Server *srv, const std::string &body, std::string *sha, lu
...
+        std::any_of(libname.begin(), libname.end(), [](char v) { return !std::isalnum(v) && v != '_'; })) {
+      return {Status::NotOK, "Expect a valid library name in the Shebang statement"};
+    }
+    return Status::OK();
+  }
+
+  return {Status::NotOK, "Expect a library name in the Shebang statement"};
+}
+
+[[nodiscard]] Status ExtractFlagsFromShebang(const std::string &shebang, uint64_t &flags) {
```
mapleFU: Output using StatusOr or uint64_t* flags?

src/storage/scripting.cc
```diff
@@ -409,13 +384,42 @@ Status FunctionCall(redis::Connection *conn, const std::string &name, const std:
...
+    for (int i = 1; i <= n; ++i) {
+      lua_pushnumber(lua, i);
+      lua_gettable(lua, -2);
+      std::string flag = lua_tostring(lua, -1);
+      if (flag == "no-writes") {
+        function_flags |= kScriptNoWrites;
+      } else if (flag == "allow-oom") {
+        return {Status::NotSupported, "allow-oom is not supported yet"};
+      } else if (flag == "allow-stale") {
+        return {Status::NotSupported, "allow-stale is not supported yet"};
```
mapleFU: @PragmaTwice Don't familiar with Lua but would RAII for pushnumber required here?

src/storage/scripting.cc
PokIsemaine: Oh thanks. I forgot about `StatusOr`

src/storage/scripting.h
PokIsemaine: I understand that `enum class` has better type safety, but when used here, it is more like a bitwise `uint64_t`. If using `enum class` requires the use of `static_cast`, it feels troublesome. Do you have any better suggestions?

src/server/server.h
PokIsemaine: Although this area may need to be redesigned, it is still worth thinking about. 

I think we need to rename `ScriptFlags=>ScriptFlagType` and add `using ScriptFlags = uint64_t;`.
`ScriptFlags` is composed of `ScriptFlagType` using bit operations. This is more clear. What do you think?

mapleFU: LGTM

src/storage/scripting.cc
```diff
@@ -409,13 +390,23 @@ Status FunctionCall(redis::Connection *conn, const std::string &name, const std:
...
     if (!s) return s;
 
     lua_getglobal(lua, (REDIS_LUA_REGISTER_FUNC_PREFIX + name).c_str());
   }
 
+  ScriptFlags function_flags = read_only ? ScriptFlagType::kScriptNoWrites : 0;
+  lua_getglobal(lua, (REDIS_LUA_REGISTER_FUNC_FLAGS_PREFIX + name).c_str());
+  if (lua_isstring(lua, -1)) {
+    function_flags |= GET_OR_RET(ParseInt<ScriptFlags>(lua_tostring(lua, -1)));
+  }
```
PragmaTwice: As I mentioned before, I think we can just store the flags (as an integer) to `REDIS_LUA_REGISTER_FUNC_FLAGS_PREFIX + name` in `RedisRegisterFunction`, instead of parsing it every time in `FunctionCall`.

src/storage/scripting.cc
```diff
@@ -635,6 +643,15 @@ Status EvalGenericCommand(redis::Connection *conn, const std::string &body_or_sh
     lua_getglobal(lua, funcname);
   }
 
+  ScriptRunCtx current_script_run_ctx;
+  lua_getglobal(lua, fmt::format(REDIS_LUA_FUNC_SHA_FLAGS, funcname + 2).c_str());
+  if (lua_isstring(lua, -1)) {
+    current_script_run_ctx.flags = GET_OR_RET(ParseInt<ScriptFlags>(lua_tostring(lua, -1)));
+  }
+  lua_pop(lua, 1);
```
PragmaTwice: ditto

src/storage/scripting.cc
PokIsemaine: Lua does not support direct storage of uint64_t(**unsigned** integer) and requires static_cast, so I chose to save string

PokIsemaine: The string here is the string form of uint64_t (parsing result). The parsing process has been completed during Function Load instead of in Function Call.

PokIsemaine: If there is any deviation in my understanding, please point it out, thanks.

PragmaTwice: hmm I think maybe signed integer is fine here?

PokIsemaine: Yes, I think using signed integer is also an option. Actually, I tried using signed integer at first, but a strange feeling prompted me to switch to string. I suddenly felt uncomfortable looking at static_cast. 

This is just my very subjective feeling, HaHa. If you mind, I will change it to store as signed integers later.

PokIsemaine:
image

The converting integer part uses the `lua_tointegerx` API, which actually belongs to lua 5.2. `luaJIT` is compatible with lua 5.1 and also extends this API, but lua 5.1 itself does not support it.
I use this API because I want to detect the case where the conversion to integer fails, I don't use `lua_tointeger` because on failure it returns 0, which is indistinguishable from a true 0.
Two possible solutions:
1. Upgrade lua to 5.2
2. Just use `lua_tointeger` and the conversion is successful by default. The flags stored to lua here are controlled, generated by kvrocks and illegal flags should be detected during parsing. I think we can ignore the detection during conversion.

What do you think, or is there a better solution? @PragmaTwice

PragmaTwice:
I think maybe we can just use `lua_isinteger` to check before calling  `lua_tointeger`?

PokIsemaine:
> I think maybe we can just use `lua_isinteger` to check before calling `lua_tointeger`?

This API seems to be supported by lua 5.3

src/cluster/cluster.cc
```diff
@@ -849,6 +849,21 @@ Status Cluster::CanExecByMySelf(const redis::CommandAttributes *attributes, cons
     return {Status::RedisClusterDown, "Hash slot not served"};
   }
 
+  bool cross_slot_ok = false;
+  if (script_run_ctx) {
+    if (script_run_ctx->current_slot != -1 && script_run_ctx->current_slot != slot) {
+      if (getNodeIDBySlot(script_run_ctx->current_slot) != getNodeIDBySlot(slot)) {
+        return {Status::RedisMoved, fmt::format("{} {}:{}", slot, slots_nodes_[slot]->host, slots_nodes_[slot]->port)};
+      }
```
PragmaTwice, line 857: Seems here we'll ignore all other checks below?

src/storage/scripting.cc
```diff
@@ -427,6 +418,22 @@ Status FunctionCall(redis::Connection *conn, const std::string &name, const std:
     lua_pop(lua, 2);
   }
 
+  SaveOnRegistry<void>(lua, REGISTRY_SCRIPT_RUN_CTX_NAME, nullptr);
```
PragmaTwice: Maybe we can define a new function e.g. `RemoveFromRegistry` for this purpose?

src/cluster/cluster.cc
PokIsemaine: `RedisMoved` is prohibited for scripts. Non-script calls to `CanExecByMySelf` will not go to this branch.

`RedisGenericCommand`
```cpp
auto s = srv->cluster->CanExecByMySelf(attributes, args, conn, script_run_ctx);
if (!s.IsOK()) {
    if (s.Is<Status::RedisMoved>()) {
        PushError(lua, "Script attempted to access a non local key in a cluster node script");
    } else {
        PushError(lua, redis::StatusToRedisErrorMsg(s).c_str());
    }
    return raise_error ? RaiseError(lua) : 1;
}
```

PokIsemaine: > allow-cross-slot-keys 
> Different slots on different nodes are still prohibited from crossing Slots. However, slots on the same node are allowed to cross slots when this flag is enabled (generally not recommended).

PragmaTwice: Ahh I see.

git-hulk (approved):
LGTM, this PR is in good shape.

PragmaTwice (commented):
LGMT except that I'm not so familiar with the impl of `Cluster::CanExecByMySelf`.

cc @mapleFU @git-hulk could you check it when you have time?

PragmaTwice (approved):

git-hulk:
> LGMT except that I'm not so familiar with the impl of `Cluster::CanExecByMySelf`.
> 
> cc @mapleFU @git-hulk could you check it when you have time?

Sure, I have checked and looks good from my side.

git-hulk:
@mapleFU To see if you have further comments? or we can merge this PR to avoid pending too long.

mapleFU (approved):

PR #2448, opened 2024-07-26
feat(config): implement integer config option with unit

### LINKED ISSUE #2445: Support parsing the configuration with unit type
### Search before asking

### Motivation

Currently, some configurations are in units of bytes which are not intuitive for human reading. So we would like to introduce a new config type like `IntWithUnitField` or `BytesField` to mitigate this issue.

For example, the following configurations are allowed:

```
rocksdb.block_size 16384
```

or

```
rocksdb.block_size 16k
```

and it will be rewritten to human readable format if possible.

### Solution

_No response_

PragmaTwice:
It closes #2445.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -63,8 +63,9 @@ repl-namespace-enabled no
 
 # By default, the max length of bulk string is limited to 512MB. If you want to
 # change this limit to a different value(must >= 1MiB), you can use the following configuration.
+# It can be just an integer (e.g. 10000000), or an integer followed by a unit (e.g. 12M, 7G, 2T).
 #
-# proto-max-bulk-len 536870912
+# proto-max-bulk-len 512M
 
 # Persist the cluster nodes topology in local file($dir/nodes.conf). This configuration
 # takes effect only if the cluster mode was enabled.
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -65,7 +65,7 @@ repl-namespace-enabled no
 # change this limit to a different value(must >= 1MiB), you can use the following configuration.
 # It can be just an integer (e.g. 10000000), or an integer followed by a unit (e.g. 12M, 7G, 2T).
 #
-# proto-max-bulk-len 512M
+# proto-max-bulk-len 536870912
 
 # Persist the cluster nodes topology in local file($dir/nodes.conf). This configuration
 # takes effect only if the cluster mode was enabled.
--- a/tests/gocase/unit/config/config_test.go
+++ b/tests/gocase/unit/config/config_test.go
@@ -266,6 +266,12 @@ func TestChangeProtoMaxBulkLen(t *testing.T) {
 	require.NoError(t, err)
 	require.EqualValues(t, "2097152", vals["proto-max-bulk-len"])
 
+	// change to 100MB
+	require.NoError(t, rdb.ConfigSet(ctx, "proto-max-bulk-len", "100M").Err())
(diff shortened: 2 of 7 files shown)


tests/gocase/unit/config/config_test.go
```diff
@@ -266,6 +266,12 @@ func TestChangeProtoMaxBulkLen(t *testing.T) {
 	require.NoError(t, err)
 	require.EqualValues(t, "2097152", vals["proto-max-bulk-len"])
 
+	// change to 100MB
+	require.NoError(t, rdb.ConfigSet(ctx, "proto-max-bulk-len", "100M").Err())
+	vals, err = rdb.ConfigGet(ctx, "proto-max-bulk-len").Result()
+	require.NoError(t, err)
+	require.EqualValues(t, "100M", vals["proto-max-bulk-len"])
```
git-hulk: For the config get command, Redis always returns an integer instead of the value with unit, for the sake of many monitoring tools will use this value as an integer.

PragmaTwice: I don't think there can be 2 different `ToString()` for two purposes. As it can make the config part hard to maintain.

Users can control whether it is pure integer or integer with unit.

kvrocks.conf
```diff
@@ -63,8 +63,9 @@ repl-namespace-enabled no
 
 # By default, the max length of bulk string is limited to 512MB. If you want to
 # change this limit to a different value(must >= 1MiB), you can use the following configuration.
+# It can be just an integer (e.g. 10000000), or an integer followed by a unit (e.g. 12M, 7G, 2T).
 #
-# proto-max-bulk-len 536870912
+# proto-max-bulk-len 512M
```
```suggestion
# proto-max-bulk-len 536870912
```

tests/gocase/unit/config/config_test.go
PragmaTwice: also I think we should keep the original format instead of converting the value in `CONFIG GET`.

git-hulk: Yes, I agree we should not have two `ToString()` for this purpose. But perhaps we can convert the value to string this in Gonfig::Get if it's an integer field. What do you think?

> Users can control whether it is pure integer or integer with unit.

It may introduce complexity for changing the configuration, users need to know this context if the monitoring/exporter is using the configuration value with unit.

PragmaTwice: >  Redis always returns an integer instead of the value with unit

Redis doesn't support such config format (integer with unit) at all, so I don't think we should align with Redis.

> convert the value to string this in Gonfig::Get if it's an integer field

Sorry I cannot get your point. Could you explain it?

git-hulk: > Redis doesn't support such config format (integer with unit) at all, so I don't think we should align with Redis.


https://github.com/redis/redis/blob/unstable/redis.conf#L2111 Redis also supports the value with the unit. The main issue is it needs to workaround in the monitoring tool if uses the value returns by `config::get`. 

> Sorry I cannot get your point. Could you explain it?

My initial thought is to use `std::string()` instead of `ToString()` if the field type is InterFieldWithUnit.

PragmaTwice: > My initial thought is to use std::string() instead of ToString() if the field type is InterFieldWithUnit.

IMHO it's even worse than two virtual function inside ConfigField in the aspect of software engineering...

git-hulk: I see, might we also add another virtual function for the config dump purpose.

git-hulk (approved):

PR #2450, opened 2024-07-28
feat(search): implement vector query for sql/redisearch parser & transformer

Beihao-Zhou:
This PR is part of #2426 

## Summary

- [x] Hnsw Vector Redis Query Parser & Transformer (i.e. KNN & Range)
- [x] Hnsw Vector SQL Query Parser & Transformer (i.e. KNN & Range)
- [x] Handle vector input (including blob and string representation)
   
## References
- [Redis Combined queries](https://redis.io/docs/latest/develop/interact/search-and-query/query/combined/#numeric-filter)
- [Pgvector SQL query](https://github.com/pgvector/pgvector?tab=readme-ov-file#querying)


src/search/ir.h
```diff
@@ -336,15 +391,22 @@ struct LimitClause : Node {
   std::string Content() const override { return fmt::format("{}, {}", offset, count); }
 
   std::unique_ptr<Node> Clone() const override { return std::make_unique<LimitClause>(*this); }
+  size_t GetOffset() const { return offset; }
+
+  size_t GetCount() const { return count; }
```
PragmaTwice:
```suggestion

  size_t Offset() const { return offset; }
  size_t Count() const { return count; }
```

src/search/ir.h
```diff
@@ -229,6 +229,61 @@ struct NumericCompareExpr : BoolAtomExpr {
   }
 };
 
+struct VectorLiteral : Literal {
+  std::string data;
```
PragmaTwice: Hmm should it be a `std::vector<double>` instead of `std::string`?

src/search/redis_query_transformer.h
```diff
@@ -141,11 +139,29 @@ struct Transformer : ir::TreeTransformer {
         } else {
           return std::make_unique<ir::AndExpr>(std::move(exprs));
         }
+      } else if (Is<VectorRange>(query)) {
+        auto vector = GET_OR_RET(GetParam(query->children[2]));
+        return std::make_unique<VectorRangeExpr>(std::make_unique<FieldRef>(field),
+                                                 GET_OR_RET(NumberOrParam(query->children[1])),
+                                                 std::make_unique<VectorLiteral>(std::move(vector)));
```
PragmaTwice: here the vector is in binary format, but in SQL it's text format.

I think we need to unify them for further processing.

Beihao-Zhou: Oh right, this is the todo item left for this PR. I realized I forgot to leave some comments or notes about it—my bad. I’m currently wrapping up the ParseArray functions into parse_util.h. This update should make them more abstract and reusable. I’m aiming to push the new revision soon!

Beihao-Zhou:
@PragmaTwice Ready for review now! 
I'm not sure if there are better solutions for the unified abstraction to handle vector parsing. Let me know if you see any room of improvement! Thx! <3

src/search/redis_query_parser.h
```diff
@@ -30,6 +30,11 @@ namespace redis_query {
 
 using namespace peg;
 
+struct VectorRangeKey : string<'V', 'E', 'C', 'T', 'O', 'R', '_', 'R', 'A', 'N', 'G', 'E'> {};
+struct KnnKey : string<'K', 'N', 'N'> {};
```
PragmaTwice:
```suggestion
struct VectorRangeToken : string<'V', 'E', 'C', 'T', 'O', 'R', '_', 'R', 'A', 'N', 'G', 'E'> {};
struct KnnToken : string<'K', 'N', 'N'> {};
```

src/search/redis_query_transformer.h
```diff
@@ -53,7 +53,32 @@ StatusOr<std::unique_ptr<parse_tree::node>> ParseToTree(Input&& in) {
 struct Transformer : ir::TreeTransformer {
   explicit Transformer(const ParamMap& param_map) : TreeTransformer(param_map) {}
 
+  StatusOr<std::unique_ptr<VectorLiteral>> Transform2Vector(const TreeNode& node) {
+    std::string vector_str = GET_OR_RET(GetParam(node));
+
+    auto vector_chars = GET_OR_RET(Binary2Chars(vector_str));
+    std::vector<double> values = GET_OR_RET(CharsToVector<double>(vector_chars));
```
PragmaTwice: I think we can just create a function to convert the binary data to `std::vector<double>` since it's more intuitive.

Also in the SemaChecker we need to check if the length of the vector is valid.

src/search/sql_transformer.h
```diff
@@ -57,7 +58,42 @@ StatusOr<std::unique_ptr<parse_tree::node>> ParseToTree(Input&& in) {
...
+      vector_str = GET_OR_RET(GetParam(node));
+    } else {
+      return {Status::NotOK, "encountered invalid vector input"};
+    }
+
+    if (vector_str.size() < 2 || vector_str[0] != '[' || vector_str.back() != ']') {
+      return {Status::NotOK, "encountered invalid vector input without brackets"};
+    }
+
+    std::vector<double> values = GET_OR_RET(ParseFloatArray(vector_str.substr(1, vector_str.size() - 2)));
```
PragmaTwice: I believe we should standardize the parameter format here, i.e. regardless of the frontend language used, values of vector parameters should be represented in binary format.

This will make it easier to understand for users.

src/search/sql_parser.h
```diff
@@ -41,7 +41,11 @@ struct NumericAtomExpr : WSPad<sor<NumberOrParam, Identifier>> {};
 struct NumericCompareOp : sor<string<'!', '='>, string<'<', '='>, string<'>', '='>, one<'=', '<', '>'>> {};
 struct NumericCompareExpr : seq<NumericAtomExpr, NumericCompareOp, NumericAtomExpr> {};
 
-struct BooleanAtomExpr : sor<HasTagExpr, NumericCompareExpr, WSPad<Boolean>> {};
+struct VectorCompareOp : sor<string<'<', '-', '>'>> {};
+struct VectorCompareExpr : seq<WSPad<Identifier>, VectorCompareOp, WSPad<StringOrParam>> {};
```
PragmaTwice: I'm wondering if we can introduce a new syntax of vector literals here, e.g.
```c++
struct VectorLiteral 
  : seq<WSPad<one<'['>>, Number, star<seq<WSPad<one<','>>>, Number>, WSPad<one<']'>>> {};
```

And in the transformer you can collect all children node (typed `Number`) inside the `VectorLiteral` to construct a `std::vector<double>`.

Beihao-Zhou: This looks good to me. Perhaps supporting both text and binary formats for **SQL AND Redis** would be more flexible, allowing users to choose the format that best suits their needs? 
```
struct BinaryOrVectorL : sor<StringL, VectorLiteral> {};
struct VectorCompareExpr : seq<WSPad<Identifier>, VectorCompareOp, WSPad<BinaryOrVectorL>> {};
```

PragmaTwice: It looks good, but I personally prefer not to use binary format in the syntax because it is more suited for programming interfaces and is not easily readable.

Beihao-Zhou: I also personally prefer to only support text, but we need to document it in the future, as this is different from Redis vector search which only supports binary format. 
Will implement `VectorLiteral` and release the next revision very soon!

@PragmaTwice I'm also considering whether we should stick to Redis's protocol, which requires vectors to be passed in as parameters, or if we should parse VectorLiterals directly, like "*=>[KNN 10 @doc_embedding [1,2,3]]". Personally, I prefer the latter approach because it makes the query transform layer more intuitive. Let me know it this looks good to you

PragmaTwice: For RediSearch part, we need to be compatible to the original implementation in Redis Stack. (which means we only need to support vector parameters in binary format, without any custom syntax).

But for SQL part, we can add our own syntax while trying to be similar to the syntax of MySQL. (so that we can support both vector parameters in binary format and also the syntax of vector literals).

There are many bad design in RediSearch and we can avoid them in SQL part (e.g. no vector literals).

Beihao-Zhou: I see I see, finally get your point, thanks a lot!!

PragmaTwice: Also for the documentation, yeah we really need a separate documentation page for users to learn how to use Kvrocks Search and what the difference is between RediSearch and Kvrocks Search.

Currently it's fine since we haven't release Kvrocks Search. Also I didn't find any spare time to create the page.

Beihao-Zhou: I could probably help with that after getting done most part of vector search, will take note of that on the tracking issues.

src/search/common_transformer.h
```diff
@@ -105,6 +105,46 @@ struct TreeTransformer {
 
     return result;
   }
+
+  template <typename T = double>
+  static StatusOr<std::vector<T>> Binary2Vector(std::string_view str) {
```
PragmaTwice, line 110: Hmm I tried some simple queries on Redis Stack, e.g.
```
FT.SEARCH documents "*=>[KNN 10 @doc_embedding $BLOB]" PARAMS 2 BLOB "\x12\xa9\xf5\x6c" SORTBY __vector_score DIALECT 4
```

Seems here the parameter `"\x12\xa9\xf5\x6c"` is a redis-cli level string literal and it will be converted to binary format  (i.e. a 4-byte binary string) before it is sent to the redis server.

So maybe in the server side we can just do some bit-level casts, e.g. `std::bit_cast<double>(first_8_bytes_of_str)`

src/search/sql_parser.h
```diff
@@ -41,7 +41,12 @@ struct NumericAtomExpr : WSPad<sor<NumberOrParam, Identifier>> {};
 struct NumericCompareOp : sor<string<'!', '='>, string<'<', '='>, string<'>', '='>, one<'=', '<', '>'>> {};
 struct NumericCompareExpr : seq<NumericAtomExpr, NumericCompareOp, NumericAtomExpr> {};
 
-struct BooleanAtomExpr : sor<HasTagExpr, NumericCompareExpr, WSPad<Boolean>> {};
+struct VectorCompareOp : sor<string<'<', '-', '>'>> {};
```
PragmaTwice: seems `sor` here is useless?

Beihao-Zhou: This was initially designed to accommodate future operators like `<=>` for cosine distance metrics (see [pgvector](https://github.com/pgvector/pgvector?tab=readme-ov-file#querying)). But upon reconsideration, I think it's probably unnecessary since distance metrics are typically specified during index creation, which doesn't allow users to alter them during queries. 

Thanks for pointing this out; I'll remove the redundant code. <3

src/search/common_transformer.h
Beihao-Zhou: Thanks for helping test it; will make the changes!
One thing is `std::bit_cast<T>` is supported until c++20, I may have to use `memcpy` instead

src/search/ir.h
```diff
@@ -229,6 +229,63 @@ struct NumericCompareExpr : BoolAtomExpr {
...
+  std::unique_ptr<NumericLiteral> range;
+  std::unique_ptr<VectorLiteral> vector;
+
+  VectorRangeExpr(std::unique_ptr<FieldRef> &&field, std::unique_ptr<NumericLiteral> &&range,
+                  std::unique_ptr<VectorLiteral> &&vector)
+      : field(std::move(field)), range(std::move(range)), vector(std::move(vector)) {}
+
+  std::string_view Name() const override { return "VectorRangeExpr"; }
+  std::string Dump() const override {
+    return fmt::format("{} vector_range {} {}", field->Dump(), range->Dump(), vector->Dump());
```
PragmaTwice:
```suggestion
    return fmt::format("{} <-> {} < {}", field->Dump(), vector->Dump(), range->Dump());
```

src/search/ir.h
```diff
@@ -229,6 +229,63 @@ struct NumericCompareExpr : BoolAtomExpr {
...
+  }
+
+  std::unique_ptr<Node> Clone() const override {
+    return std::make_unique<VectorRangeExpr>(Node::MustAs<FieldRef>(field->Clone()),
+                                             Node::MustAs<NumericLiteral>(range->Clone()),
+                                             Node::MustAs<VectorLiteral>(vector->Clone()));
+  }
+};
+
+struct VectorSearchExpr : BoolAtomExpr {
```
PragmaTwice: Would "VectorKnnExpr" be a more suitable name here?

src/search/ir.h
```diff
@@ -336,15 +393,22 @@ struct LimitClause : Node {
...
   enum Order { ASC, DESC } order = ASC;
   std::unique_ptr<FieldRef> field;
+  std::unique_ptr<VectorLiteral> vector = nullptr;
 
   SortByClause(Order order, std::unique_ptr<FieldRef> &&field) : order(order), field(std::move(field)) {}
+  SortByClause(std::unique_ptr<FieldRef> &&field, std::unique_ptr<VectorLiteral> &&vector)
+      : field(std::move(field)), vector(std::move(vector)) {}
 
   static constexpr const char *OrderToString(Order order) { return order == ASC ? "asc" : "desc"; }
+  bool IsKnn() const { return vector != nullptr; }
```
PragmaTwice:
```suggestion
  bool IsVectorField() const { return vector != nullptr; }
```

src/search/ir.h
```diff
@@ -356,6 +420,10 @@ struct SortByClause : Node {
   std::unique_ptr<Node> Clone() const override {
     return std::make_unique<SortByClause>(order, Node::MustAs<FieldRef>(field->Clone()));
   }
+
+  std::unique_ptr<FieldRef> GetFieldRef() { return std::move(field); }
+
+  std::unique_ptr<VectorLiteral> GetVectorLiteral() { return std::move(vector); }
```
PragmaTwice: usually "Get" implies that we don't modify the original object, but here `field` and `vector` will be nullptr after moving.

src/search/ir_sema_checker.h
```diff
@@ -97,6 +97,42 @@ struct SemaChecker {
...
+
+        if (meta->distance_metric == redis::DistanceMetric::COSINE && (v->range->val < 0 || v->range->val > 2)) {
+          return {Status::NotOK, "range has to be between 0 and 2 for cosine distance metric"};
+        }
+
+        if (v->vector->values.size() != meta->dim) {
+          return {Status::NotOK,
+                  fmt::format("vector should be of size `{}` for field `{}`", meta->dim, v->field->name)};
+        }
+      }
```
PragmaTwice, line 162: also in ORDER BY we need to check if the vector literal is provided when the field is typed vector.

otherwise `ORDER BY vector_field LIMIT n` is valid.

src/search/ir.h
Beihao-Zhou: I see, will change it to `TakeFieldRef` and `TakeVectorLiteral`.

src/search/ir_sema_checker.h
Beihao-Zhou: Double checked this case with `select f4 from ia order by f4 limit 5`. In the case that query only gives a field, it will go through
```
if (auto v = dynamic_cast<SortByClause *>(node)) {
// ...
      if (!iter->second.IsSortable()) {
        return {Status::NotOK, fmt::format("field `{}` is not sortable", v->field->name)};
      } 
    } 
```

So this wouldn't be a problem for vector type in particular, as `SortByClause` will invalidate all non-numeric literal.

In the case that  `ORDER BY vector_field <-> [3...] LIMIT n`, the syntax is guaranteed to be correct by parser.

PragmaTwice: > In the case that `ORDER BY vector_field <-> [3...] LIMIT n`, the syntax is guaranteed to be correct by parser.

Yeah but seems it will be rejected by SemaChecker.

Could you add this example into test cases of SemaChecker to confirm it?

PragmaTwice: Ahh I've read the code in Transformer. 

Seems it will be transformed into `VectorKnnExpr` rather than an `ORDER BY` with a vector literal?

Then I can get the idea, but I have several comments:
- Such a transformation may be better to put into an optimization pass instead of Transformer in the frontend,
- If we really want to put it inside the Transformer, maybe we can delete `vector` member in `SortByClause` (since it becomes a useless member after Transformer).

Beihao-Zhou: Gotcha! I didn't think of putting it into an optimization pass, but it does make more sense to me; will implement this solution.

Beihao-Zhou: Removed the transformation in the frontend; will add it in the next PR which will handle everything related to pass for hnsw vector search.

src/search/ir.h
```diff
@@ -356,6 +425,10 @@ struct SortByClause : Node {
   std::unique_ptr<Node> Clone() const override {
     return std::make_unique<SortByClause>(order, Node::MustAs<FieldRef>(field->Clone()));
   }
+
+  std::unique_ptr<FieldRef> TakeFieldRef() { return std::move(field); }
+
+  std::unique_ptr<VectorLiteral> TakeVectorLiteral() { return std::move(vector); }
```
PragmaTwice: I prefer to delete them if they're currently not used.

src/search/ir_sema_checker.h
```diff
@@ -50,6 +50,9 @@ struct SemaChecker {
         GET_OR_RET(Check(v->query_expr.get()));
         if (v->limit) GET_OR_RET(Check(v->limit.get()));
         if (v->sort_by) GET_OR_RET(Check(v->sort_by.get()));
+        if (v->sort_by && v->sort_by->IsVectorField() && !v->limit) {
+          return {Status::NotOK, "invalid knn search clause without limit"};
```
PragmaTwice:
```suggestion
          return {Status::NotOK, "expect a LIMIT clause for vector field to construct a KNN search"};
```

src/search/ir.h
```diff
@@ -229,6 +229,63 @@ struct NumericCompareExpr : BoolAtomExpr {
...
+  std::unique_ptr<NumericLiteral> k;
+  std::unique_ptr<VectorLiteral> vector;
+
+  VectorKnnExpr(std::unique_ptr<FieldRef> &&field, std::unique_ptr<NumericLiteral> &&k,
+                std::unique_ptr<VectorLiteral> &&vector)
+      : field(std::move(field)), k(std::move(k)), vector(std::move(vector)) {}
+
+  std::string_view Name() const override { return "VectorKnnExpr"; }
+  std::string Dump() const override {
+    return fmt::format("{} <-> {} knn {}", field->Dump(), vector->Dump(), k->Dump());
```
PragmaTwice:
```suggestion
    return fmt::format("KNN k={}, {} <-> {}", k->Dump(), field->Dump(), vector->Dump());
```

src/search/ir.h
```diff
@@ -336,18 +393,30 @@ struct LimitClause : Node {
...
   static constexpr const char *OrderToString(Order order) { return order == ASC ? "asc" : "desc"; }
+  bool IsVectorField() const { return vector != nullptr; }
 
   std::string_view Name() const override { return "SortByClause"; }
-  std::string Dump() const override { return fmt::format("sortby {}, {}", field->Dump(), OrderToString(order)); }
+  std::string Dump() const override {
+    if (!IsVectorField()) {
+      return fmt::format("sortby {}, {}", field->Dump(), OrderToString(order));
+    }
+    return fmt::format("sortby {} <-> {}", field->Dump(), vector->Dump());
```
PragmaTwice, line 418: Is ASC DESC useful in KNN search?

src/search/ir_sema_checker.h
```diff
@@ -60,8 +63,21 @@ struct SemaChecker {
         return {Status::NotOK, fmt::format("field `{}` not found in index `{}`", v->field->name, current_index->name)};
       } else if (!iter->second.IsSortable()) {
         return {Status::NotOK, fmt::format("field `{}` is not sortable", v->field->name)};
+      } else if (auto meta = iter->second.MetadataAs<redis::HnswVectorFieldMetadata>();
+                 (meta != nullptr) != v->IsVectorField()) {
```
PragmaTwice:
```suggestion
      } else if (auto is_vector = iter->second.MetadataAs<redis::HnswVectorFieldMetadata>() != nullptr;
                 is_vector != v->IsVectorField()) {
```

PragmaTwice: current code is a bit difficult to read

src/search/ir_sema_checker.h
```diff
@@ -50,6 +50,9 @@ struct SemaChecker {
         GET_OR_RET(Check(v->query_expr.get()));
         if (v->limit) GET_OR_RET(Check(v->limit.get()));
         if (v->sort_by) GET_OR_RET(Check(v->sort_by.get()));
+        if (v->sort_by && v->sort_by->IsVectorField() && !v->limit) {
```
PragmaTwice, line 53: maybe here we also need to consider if the current field is a `noindex` field? (i.e. no index is constructed for such a field, just used in filtering)

src/search/ir_sema_checker.h
PragmaTwice: Great!

src/search/ir.h
Beihao-Zhou: Nope, KNN executors always return the result from the nearest node to farthest.

PragmaTwice: hmm then maybe we can try `std::variant<Order, std::unique_ptr<VectorLiteral>>` to prevent misunderstanding since order will be useless for vector fields?

src/search/sql_parser.h
```diff
@@ -84,7 +89,8 @@ struct Limit : string<'l', 'i', 'm', 'i', 't'> {};
 
 struct WhereClause : seq<Where, QueryExpr> {};
 struct AscOrDesc : sor<Asc, Desc> {};
-struct OrderByClause : seq<OrderBy, WSPad<Identifier>, opt<WSPad<AscOrDesc>>> {};
+struct OrderByExpr : sor<WSPad<VectorCompareExpr>, seq<WSPad<Identifier>, opt<WSPad<AscOrDesc>>>> {};
```
PragmaTwice: Also here seems it's allowed to pass an ASC/DESC after `VectorCompareExpr`. We can either prevent it in sql::Transformer or change the syntax definition here.

src/search/ir_sema_checker.h
Beihao-Zhou: Just curious, is a `NOINDEX` vector field the only type of index that completely prohibits all query operations, such as KNN and range queries? For other field types, like tag and numeric filters, queries simply revert to a full scan right?

PragmaTwice: I think we can allow range queries for noindex vector fields. But for KNN we don't need to support it for noindex fields.

For tag and numeric (and also vector), it depends on the query itself and the index selection pass to determine the query plan. e.g. if we have a noindex numeric field `a` and non-noindex numeric field `b` in `idx`,  the query plan of `select * from idx where a > 1 and b > 2` can be `filter a > 1 in (range-query b)`.

src/search/sql_parser.h
Beihao-Zhou: This is functionally correct since `opt<WSPad<AscOrDesc>>` is wrapped into`seq<WSPad<Identifier>, [...]>` but it is unreadable, so made it to the following and add corresponding test
```
struct SortableFieldExpr : seq<WSPad<Identifier>, opt<AscOrDesc>> {};
struct OrderByExpr : sor<WSPad<VectorCompareExpr>, WSPad<SortableFieldExpr>> {};
```

PragmaTwice (approved):
Looks good to me now. Thank you!

src/search/sql_parser.h
PragmaTwice: Ahh got it.

PR #2452, opened 2024-07-28
[WIP]feat(config): Add Support for maxmemory Configuration

### LINKED ISSUE #2284: Memory limits on connections
### Search before asking

### Motivation

The lack of memory constraints in current Kvrocks can lead to OOM, especially when deployed in container environments. 

### Solution

We should provide mechanisms to limit memory and avoid OOM. 

In the connection dimension:
- avoid large `WriteBatch`
- limit connection output buffer, support `client-output-buffer-limit` and `maxmemory-clients`

AntiTopQuark:
This PR will close https://github.com/apache/kvrocks/issues/2284 .

This PR introduces support for the `maxmemory` configuration, allowing users to set a maximum memory limit for the server. Key changes include:

- **Configuration Update**: Added `maxmemory` option in `kvrocks.conf`, allowing users to define the maximum memory size limit (in bytes) that the server can use. This limit is checked only when a connection is established.
- **Command Attributes**: Updated `commander.h` to include attributes for the `maxmemory` configuration.
- **Configuration Parsing**: Modified `config.cc` and `config.h` to parse and store the `maxmemory` configuration value.
- **Memory Info**: Enhanced `server.cc` to report the `maxmemory` configuration and calculate memory usage as a percentage of the defined limit.
- **Memory Limit Check**: Implemented `CheckMemorySizeLimit` function in `server.cc` to ensure memory usage does not exceed the defined `maxmemory`.
- **Connection Handling**: Updated `worker.cc` to check memory size limit before establishing a new TCP connection, logging a warning if the memory limit is exceeded.

These changes add robustness to the server's memory management, preventing it from using more memory than allocated.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -63,6 +63,7 @@ repl-namespace-enabled no
 
 # By default, the max length of bulk string is limited to 512MB. If you want to
 # change this limit to a different value(must >= 1MiB), you can use the following configuration.
+# It can be just an integer (e.g. 10000000), or an integer followed by a unit (e.g. 12M, 7G, 2T).
 #
 # proto-max-bulk-len 536870912
 
--- a/src/common/bit_util.h
+++ b/src/common/bit_util.h
@@ -20,6 +20,14 @@
 
 #pragma once
 
+#include <cstddef>
+#include <cstdint>
+#include <limits>
+
+#include "encoding.h"
+#include "status.h"
+#include "type_util.h"
+
 namespace util {
 
 /* Count number of bits set in the binary array pointed by 's' and long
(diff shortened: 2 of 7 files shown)


git-hulk:
@AntiTopQuark What you did in this PR did not match issue #2284, which intended to control the output buffer limit instead of the process memory. The current implementation also cannot promise maximum memory usage since we have many background threads and existing connections might request more memory than that.

AntiTopQuark:
> @AntiTopQuark What you did in this PR did not match issue #2284, which intended to control the output buffer limit instead of the process memory. The current implementation also cannot promise maximum memory usage since we have many background threads and existing connections might request more memory than that.

I apologize for the oversight. I will revise the code to align with the specified requirements.

git-hulk:
@AntiTopQuark No worry, thanks for your efforts.

PR #2466, opened 2024-08-04
feat(search): Hnsw Vector Search Optimizaton Pass

Beihao-Zhou:
This PR is part of #2426 

## Summary

- Add optimization pass to transform `sort by` to `knn search expr` for SQL query
- Use cost model to prioritize KNN search and range search
- Add vector range search in `Filter` executor in case that vector range search acts as the prefilter for KNN search
- Support prefilters in redis query transformer

### DIFF
--- a/src/search/passes/transfer_to_knn.h
+++ b/src/search/passes/transfer_to_knn.h
@@ -32,27 +32,15 @@ struct TransferSortByToKnnExpr : Visitor {
   std::unique_ptr<Node> Visit(std::unique_ptr<SearchExpr> node) override {
     node = Node::MustAs<SearchExpr>(Visitor::Visit(std::move(node)));
 
+    // TODO: allow hybrid query
     if (node->sort_by && node->sort_by->IsVectorField() && node->limit) {
-      std::vector<std::unique_ptr<QueryExpr>> exprs;
-      auto knn_expr = std::make_unique<VectorKnnExpr>(Node::MustAs<FieldRef>(node->sort_by->TakeFieldRef()),
-                                                      Node::MustAs<VectorLiteral>(node->sort_by->TakeVectorLiteral()),
-                                                      node->limit->Count());
-      if (auto b = Node::As<BoolLiteral>(std::move(node->query_expr))) {
-        if (b->val) exprs.push_back(std::move(knn_expr));
-      } else {
-        exprs.push_back(std::move(node->query_expr));
-        exprs.push_back(std::move(knn_expr));
+      if (auto b = dynamic_cast<BoolLiteral*>(node->query_expr.get()); b && b->val) {
+        node->query_expr = std::make_unique<VectorKnnExpr>(
+            Node::MustAs<FieldRef>(node->sort_by->TakeFieldRef()),
+            Node::MustAs<VectorLiteral>(node->sort_by->TakeVectorLiteral()), node->limit->Count());
+        node->sort_by.reset();
+        node->limit.reset();
       }
-
-      if (exprs.empty()) {
-        node->query_expr = std::make_unique<BoolLiteral>(false);
-      } else if (exprs.size() == 1) {
-        node->query_expr = std::move(exprs[0]);
-      } else {
-        node->query_expr = std::make_unique<AndExpr>(std::move(exprs));
-      }
-      node->sort_by.reset();
-      node->limit.reset();
     }
 
     return node;
--- a/src/search/passes/transfer_to_knn.h
+++ b/src/search/passes/transfer_to_knn.h
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#pragma once
+
+#include <memory>
+
+#include "search/ir.h"
+#include "search/ir_pass.h"
+#include "search/ir_plan.h"
+
+namespace kqir {
+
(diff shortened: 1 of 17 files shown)


src/search/passes/transfer_to_knn.h
```diff
@@ -0,0 +1,62 @@
...
+      std::vector<std::unique_ptr<QueryExpr>> exprs;
+      auto knn_expr = std::make_unique<VectorKnnExpr>(Node::MustAs<FieldRef>(node->sort_by->TakeFieldRef()),
+                                                      Node::MustAs<VectorLiteral>(node->sort_by->TakeVectorLiteral()),
+                                                      node->limit->Count());
+      if (auto b = Node::As<BoolLiteral>(std::move(node->query_expr))) {
+        if (b->val) exprs.push_back(std::move(knn_expr));
+      } else {
+        exprs.push_back(std::move(node->query_expr));
+        exprs.push_back(std::move(knn_expr));
+      }
```
PragmaTwice: Seems here we cannot guarantee that the semantics of LIMIT clause is satisfied, e.g.
```sql
SELECT * FROM idx WHERE num > 1 ORDER BY vec <-> [...] LIMIT 10;
```
for this SQL statement, users want 10 records where `num > 1` (unless there's less than 10 records satisfying `num > 1`), 

but after transformation it's `(KNN(10) vec <-> [...]) AND num > 1`, which implies the number of records may be less than 10 even if there's more than 10 records satisfying `num > 1`, since it equals to filtering `num > 1` from `KNN(10)` query (firstly get 10 records from KNN search and then filter records of `num <= 1` out of these 10).

PragmaTwice: And in the RediSearch query part, I think it's same, i.e. we cannot simply treat prefilter KNN search as AND expressions since it requries a specific number of records (`k`) -- and it breaks the semantics of `AND`.

So if we cannot currently come up with a good solution, we can prohibit any KNN search with prefilter (in transformer or sema checker) and leave it to further PRs.

PragmaTwice (commented):
The rest looks good to me!

Also if you have time maybe you can also add [some golang test cases](https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/search/search_test.go) for vector search, after this PR is merged.

src/search/passes/transfer_to_knn.h
Beihao-Zhou: Thanks for pointing it out! Will prohibit hybrid query from this PR then. Also open to any suggestions from the community here <3

Beihao-Zhou: Ready for review again! Prohibit hybrid query in transformer for Redis and semantic check for SQL.

src/search/passes/transfer_to_knn.h
```diff
@@ -0,0 +1,50 @@
...
+
+#include <memory>
+
+#include "search/ir.h"
+#include "search/ir_pass.h"
+#include "search/ir_plan.h"
+
+namespace kqir {
+
+struct TransferSortByToKnnExpr : Visitor {
```
PragmaTwice:
```suggestion
struct SortByWithLimitToKnnExpr : Visitor {
```

src/search/passes/transfer_to_knn.h
```diff
@@ -0,0 +1,50 @@
...
+  std::unique_ptr<Node> Visit(std::unique_ptr<SearchExpr> node) override {
+    node = Node::MustAs<SearchExpr>(Visitor::Visit(std::move(node)));
+
+    // TODO: allow hybrid query
+    if (node->sort_by && node->sort_by->IsVectorField() && node->limit) {
+      if (auto b = dynamic_cast<BoolLiteral*>(node->query_expr.get()); b && b->val) {
+        node->query_expr = std::make_unique<VectorKnnExpr>(
+            Node::MustAs<FieldRef>(node->sort_by->TakeFieldRef()),
+            Node::MustAs<VectorLiteral>(node->sort_by->TakeVectorLiteral()), node->limit->Count());
```
PragmaTwice: How about `limit->offset` here? Seems it's ignored.

Beihao-Zhou: Oh thanks for your reminder! I'll wrap a Limit executor for all Knn scan then, passing `offset + count` as `k` to `KnnScan`.

Beihao-Zhou: Done <3

tests/cppunit/ir_pass_test.cc
```diff
@@ -111,18 +111,35 @@ TEST(IRPassTest, Manager) {
             "select * from a where (and x <= 1, y >= 2, z != 3)");
 }
 
+TEST(IRPassTest, SortByWithLimitToKnnExpr) {
+  SortByWithLimitToKnnExpr tsbtke;
+
+  ASSERT_EQ(tsbtke.Transform(*Parse("select a from b order by embedding <-> [3.6] limit 5"))->Dump(),
+            "select a from b where KNN k=5, embedding <-> [3.600000] limit 0, 5");
+  ASSERT_EQ(tsbtke.Transform(*Parse("select a from b where false order by embedding <-> [3,1,2] limit 5"))->Dump(),
+            "select a from b where false sortby embedding <-> [3.000000, 1.000000, 2.000000] limit 0, 5");
```
PragmaTwice, line 120: I'm wondering if this query is valid and has a correct execution result 🤔

Beihao-Zhou: For `"select a from b where false order by embedding <-> [3,1,2] limit 5"`, this will be set as `noop` in `LowerToPlan`. 
Ref: https://github.com/apache/kvrocks/pull/2466#discussion_r1705921930

tests/cppunit/ir_pass_test.cc
```diff
@@ -111,18 +111,35 @@ TEST(IRPassTest, Manager) {
...
+
 TEST(IRPassTest, LowerToPlan) {
   LowerToPlan ltp;
 
   ASSERT_EQ(ltp.Transform(*Parse("select * from a"))->Dump(), "project *: full-scan a");
   ASSERT_EQ(ltp.Transform(*Parse("select * from a limit 1"))->Dump(), "project *: (limit 0, 1: full-scan a)");
   ASSERT_EQ(ltp.Transform(*Parse("select * from a where false"))->Dump(), "project *: noop");
   ASSERT_EQ(ltp.Transform(*Parse("select * from a where false limit 1"))->Dump(), "project *: noop");
+  ASSERT_EQ(ltp.Transform(*Parse("select * from a where false order by embedding <-> [3,1,2] limit 5"))->Dump(),
+            "project *: noop");
```
Beihao-Zhou, line 135: Referring to this test case

PragmaTwice: Ahh sorry, got it now. Thank you!

PragmaTwice (approved):

PR #2472, opened 2024-08-06
feat(command): Implement the new command POLLUPDATES for polling updates by sequence

### LINKED ISSUE #2469: Proposal: add a new command for polling new updates by sequence
### Search before asking

### Motivation

Currently, RocksDB provides an API `GetUpdatesSince` to allow us to poll the write batched by the sequence number. And Kvrocks is now depending on this mechanism to implement the partial sync(PSYNC). Except for that, the official migration tool `kvrocks2redis` is also using it to fetch new updates after parsing the entire DB, but it requires running alongside the DB dir.  In some scenarios like CDC(Change Stream Capture) also has this requirement, but it’ll be too troublesome if it requires running an agent alongside each Kvrocks node.

As far as I know, some users also have this similar requirement[1]. So I propose to add a new command for this purpose:

```bash
POLLUPDATES <Sequence Number> [MAX <N>] [STRICT] [FORMAT <RAW>]
```

- `Sequence Number` represents the start sequence of the polling operation and it’s a required argument.
- `MAX` represents the maximum number of items that can be retrieved, it’s an optional argument and uses `16` as the default value if it’s missing
- `STRICT` is set means the update’s sequence MUST be exactly equal to the sequence number, it’s an optional argument. `GetUpdatesSince` will return the first available sequence if the sequence number is non-existent, so we allow users to specify if required to match the input sequence number.

And we also can extend more arguments like TIMEOUT/MIN, etc..

[1] https://www.revenuecat.com/blog/engineering/how-we-replicate-kvrocks-dataset/

git-hulk:
As proposed in issue #2469, we would like to add a new command for
polling updates from Kvrocks. The main purpose is to allow to implement
features like CDC(Change Stream Capture) without running an agent
alongside Kvrocks instances to make it easier to operate.

The following is the command format:

```shell
POLLUPDATES <Sequence Number> [MAX <N>] [STRICT] [FORMAT <RAW>]
```

- `Sequence Number` represents the start sequence of the polling operation and it’s a required argument.
- `MAX` represents the maximum number of items that can be retrieved, it’s an optional argument and uses `16` as the default value if it’s missing
- `STRICT` is set means the update’s sequence MUST be exactly equal to the sequence number, it’s an optional argument. `GetUpdatesSince` will return the first available sequence if the sequence number is non-existent, so we allow users to specify if required to match the input sequence number.

The output contains the following sections:

- last_sequence
- updates
- format
- next_sequence

For example, we assume the DB's latest sequence is 100 and we send the
command: `POLLUPDATES 90 MAX 3 FORMAT RAW`, it will return the following
response:

```shell
"latest_sequence"
100
"updates"
  {batch-0}
  {batch-1}
  {batch-2}
"next_sequence"
93
```

This will close #2469

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -1301,7 +1301,7 @@ class CommandPollUpdates : public Commander {
       next_sequence = batch.sequence + batch.writeBatchPtr->Count() - 1;
     }
 
-    *output = conn->Map(std::map<std::string, std::string>{
+    *output = conn->Map({
         {redis::BulkString("latest_sequence"), redis::Integer(srv->storage->LatestSeqNumber())},
         {redis::BulkString("updates"), std::move(updates)},
         {redis::BulkString("next_sequence"), redis::Integer(next_sequence)},
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -1265,6 +1265,64 @@ class CommandDump : public Commander {
   }
 };
 
+class CommandPollUpdates : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 1);
+    sequence_ = GET_OR_RET(parser.TakeInt<uint64_t>());
+
+    while (parser.Good()) {
+      if (parser.EatEqICase("MAX")) {
+        max_ = GET_OR_RET(parser.TakeInt<int64_t>(NumericRange<int64_t>{1, 1000}));
+      } else if (parser.EatEqICase("STRICT")) {
+        is_strict_ = true;
+      } else if (parser.EatEqICase("FORMAT")) {
+        auto format = GET_OR_RET(parser.TakeStr());
+        if (util::EqualICase(format, "RAW")) {
+          format_ = Format::Raw;
+        } else {
+          return {Status::RedisParseErr, "invalid FORMAT option, only support RAW"};
+        }
+      } else {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+    }
+    return Status::OK();
+  }
+
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    // sequence + 1 is for excluding the current sequence to avoid getting duplicate updates
+    auto batches = GET_OR_RET(srv->PollUpdates(sequence_ + 1, max_, is_strict_));
+
+    *output = redis::MultiLen(8);
+    *output += redis::BulkString("latest_sequence");
+    *output += redis::Integer(srv->storage->LatestSeqNumber());
+    *output += redis::BulkString("format");
+    *output += redis::BulkString("RAW");
+    *output += redis::BulkString("updates");
+    *output += redis::MultiLen(batches.size());
+    uint64_t next_sequence = sequence_;
+    for (const auto &batch : batches) {
+      *output += redis::BulkString(util::StringToHex(batch.writeBatchPtr->Data()));
+      // It might contain more than one sequence in a batch
+      next_sequence = batch.sequence + batch.writeBatchPtr->Count() - 1;
+    }
+    *output += redis::BulkString("next_sequence");
+    *output += redis::Integer(next_sequence);
+    return Status::OK();
+  }
+
+ private:
+  enum class Format {
+    Raw,
+  };
+
+  uint64_t sequence_ = -1;
+  bool is_strict_ = false;
+  int64_t max_ = 16;
+  Format format_ = Format::Raw;
+};
+
 REDIS_REGISTER_COMMANDS(MakeCmdAttr<CommandAuth>("auth", 2, "read-only ok-loading", 0, 0, 0),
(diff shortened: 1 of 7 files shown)


src/commands/cmd_server.cc
```diff
@@ -1265,6 +1265,64 @@ class CommandDump : public Commander {
...
+    *output += redis::BulkString("updates");
+    *output += redis::MultiLen(batches.size());
+    uint64_t next_sequence = sequence_;
+    for (const auto &batch : batches) {
+      *output += redis::BulkString(util::StringToHex(batch.writeBatchPtr->Data()));
+      // It might contain more than one sequence in a batch
+      next_sequence = batch.sequence + batch.writeBatchPtr->Count() - 1;
+    }
+    *output += redis::BulkString("next_sequence");
+    *output += redis::Integer(next_sequence);
```
PragmaTwice: Could we use RESP dict format here?

src/commands/cmd_server.cc
```diff
@@ -1265,6 +1265,64 @@ class CommandDump : public Commander {
...
+
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    // sequence + 1 is for excluding the current sequence to avoid getting duplicate updates
+    auto batches = GET_OR_RET(srv->PollUpdates(sequence_ + 1, max_, is_strict_));
+
+    *output = redis::MultiLen(8);
+    *output += redis::BulkString("latest_sequence");
+    *output += redis::Integer(srv->storage->LatestSeqNumber());
+    *output += redis::BulkString("format");
+    *output += redis::BulkString("RAW");
```
PragmaTwice: I don't think we need to repeat the content in the input.

src/commands/cmd_server.cc
PragmaTwice, line 1311: ditto

src/commands/cmd_server.cc
git-hulk: I wrote back the format field since I think it'll be convenient for parsing the batch format without knowing the request context.

src/commands/cmd_server.cc
git-hulk: Do you mean that we can replace `MultiLen` with the MAP header?

caipengbo:
Do we need to provide oldest sequence number?

git-hulk:
> Do we need to provide oldest sequence number?

AFAIK, we don't have a fast way to get this "oldest" sequence number for now?

caipengbo:
> we don't have a fast way to get this "oldest" sequence number for now?

Pass start sequence 0 and end sequcence MAX, and the first batch of sequence obtained is the oldest one.

git-hulk:
> > we don't have a fast way to get this "oldest" sequence number for now?
> 
> Pass start sequence 0 and end sequcence MAX, and the first batch of sequence obtained is the oldest one.

Yes, it can work by doing this but requires another iterator operation to get this. And I think we should not do this unless we did need this information. I guess your purpose is to know the oldest sequence if you want to start to consume from the oldest sequence? if yes, it now can use 0 as the sequence number without the STRICT flag, and it will be always started from the oldest batch.

src/commands/cmd_server.cc
PragmaTwice: yeah

src/commands/cmd_server.cc
PragmaTwice: Users MUST know the context because they should know they've sent a POLLUPDATES command.

caipengbo:
> I guess your purpose is to know the oldest sequence if you want to start to consume from the oldest sequence? if yes, it now can use 0 as the sequence number without the STRCT flag, and it will be always started from the oldest batch.

Yeah, it's what I need.  Got it :)

src/commands/cmd_server.cc
caipengbo: It is possible that the user did not specify a format in command.

src/commands/cmd_server.cc
git-hulk: Sounds good.

src/commands/cmd_server.cc
git-hulk: Yes, I believe users should know the request context. My main point is it'll be convenient for parsing response if we can identify the format directly.

caipengbo (dismissed):
LGTM

src/commands/cmd_server.cc
```diff
@@ -1265,6 +1265,61 @@ class CommandDump : public Commander {
...
+    // sequence + 1 is for excluding the current sequence to avoid getting duplicate updates
+    auto batches = GET_OR_RET(srv->PollUpdates(sequence_ + 1, max_, is_strict_));
+    std::string updates = redis::MultiLen(batches.size());
+    for (const auto &batch : batches) {
+      updates += redis::BulkString(util::StringToHex(batch.writeBatchPtr->Data()));
+      // It might contain more than one sequence in a batch
+      next_sequence = batch.sequence + batch.writeBatchPtr->Count() - 1;
+    }
+
+    *output = conn->Map(std::map<std::string, std::string>{
```
```suggestion
    *output = conn->Map({
```

caipengbo (approved):

PR #2476, opened 2024-08-07
chore(storage): Remove speedb due to lack of active

### LINKED ISSUE #2475: Remove  speedb due to it's lack of active in the public repository
The speedb is no longer active in the public repository after the acquisition, maybe we can remove it. What do you think? @apache/kvrocks-committers

_Originally posted by @git-hulk in https://github.com/apache/kvrocks/issues/2332#issuecomment-2270418939_

git-hulk:
This closes #2475


src/storage/storage.h
```diff
@@ -48,7 +48,6 @@
 
 enum class StorageEngineType : uint16_t {
   RocksDB,
-  Speedb,
```
mapleFU, line 51: So we don't need a `1` for legacy speedb here?

mapleFU (approved):

src/storage/storage.h
git-hulk: Yes, I guess no. We added this to allow awareness of the storage type so that we can tune the performance for different storage types, but we haven't used it for now.

PragmaTwice (approved):

PR #2477, opened 2024-08-08
feat(search): add support of HNSW vector fields for FT.CREATE

Beihao-Zhou:
This PR is part of #2426 

## Summary

- Add support of `ft.create` for Hnsw index
- Add go test cases for vector search


PragmaTwice (dismissed):
LGTM. Thank you! You can try to run `gofmt` to fix the CI failure.

PragmaTwice (dismissed):

tests/gocase/unit/search/search_test.go
```diff
@@ -88,12 +104,30 @@ func TestSearch(t *testing.T) {
...
+					require.Equal(t, "12.000000, 13.000000, 14.000000", fields[3])
+					require.Equal(t, "b", fields[4])
+					require.Equal(t, "22", fields[5])
+				} else if fields[2] == "b" {
+					require.Equal(t, "22", fields[3])
+					require.Equal(t, "c", fields[4])
+					require.Equal(t, "12.000000, 13.000000, 14.000000", fields[5])
+				} else {
+					require.Fail(t, "not started with c or b")
+				}
```
PragmaTwice: maybe we can read them into a map and then do some checks so that we don't need to care the order.

Beihao-Zhou: good idea! I've pushed new code change

PragmaTwice (approved):

PR #2479, opened 2024-08-09
feat(command): add category enum for every command

PragmaTwice:
Sometimes just checking the prefix is not enough, so we can add a category to command attributes.

### DIFF
--- a/src/commands/commander.h
+++ b/src/commands/commander.h
@@ -67,6 +67,31 @@ enum CommandFlags : uint64_t {
   kCmdNoDBSizeCheck = 1ULL << 12,  // "no-dbsize-check" flag
 };
 
+enum class CommandCategory : uint8_t {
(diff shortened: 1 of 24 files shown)


git-hulk (approved):

src/commands/commander.h
```diff
@@ -67,6 +67,31 @@ enum CommandFlags : uint64_t {
...
+  Unknown = 0,
+  Bit,
+  BloomFilter,
+  Cluster,
+  Function,
+  Geo,
+  Hash,
+  HLL,
+  JSON,
+  Key,
```
mapleFU, line 80: "Key" is weird here since most methods would have a "Key"?

PragmaTwice: Key is for key-related commands, same as in redis.

PR #2484, opened 2024-08-10
feat(search): allow to omit `ON HASH|JSON` in FT.CREATE

PragmaTwice:
The default data type is HASH, same as in Redis.


git-hulk:
LGTM, we can add a test case to avoid unexpected changes on this?

PragmaTwice:
> LGTM, we can add a test case to avoid unexpected changes on this?

Done.

git-hulk (approved):

PR #2488, opened 2024-08-12
feat: use non-root user in Docker image

kinoute:


git-hulk (approved):

git-hulk:
It's great to use the user `kvrocks` instead of `root` from the security proposition.

aleksraiden (approved):

Issue #2481, opened 2024-08-09
Benchmark for vector indexing and searching


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

We can try to use this repo: https://github.com/qdrant/vector-db-benchmark

After some simple patching:
```patch
diff --git a/engine/clients/redis/configure.py b/engine/clients/redis/configure.py
index a5e6fe8..b2b8af0 100644
--- a/engine/clients/redis/configure.py
+++ b/engine/clients/redis/configure.py
@@ -7,6 +7,7 @@ from redis.commands.search.field import (
     TextField,
     VectorField,
 )
+from redis.commands.search.indexDefinition import IndexDefinition, IndexType
 
 from benchmark.dataset import Dataset
 from engine.base_client.configure import BaseConfigurator
@@ -81,7 +82,7 @@ class RedisConfigurator(BaseConfigurator):
                 name="vector",
                 algorithm="HNSW",
                 attributes={
-                    "TYPE": "FLOAT32",
+                    "TYPE": "FLOAT64",
                     "DIM": dataset.config.vector_size,
                     "DISTANCE_METRIC": self.DISTANCE_MAPPING[dataset.config.distance],
                     **self.collection_params.get("hnsw_config", {}),
@@ -97,8 +98,13 @@ class RedisConfigurator(BaseConfigurator):
             ]
         for conn in conns:
             search_namespace = conn.ft()
+            idef = IndexDefinition(
+                prefix=[''],
+                index_type=IndexType.HASH
+            )
+            idef.args = idef.args[:-2]
             try:
-                search_namespace.create_index(fields=index_fields)
+                search_namespace.create_index(fields=index_fields, definition=idef)
             except redis.ResponseError as e:
                 if "Index already exists" not in str(e):
                     raise e
diff --git a/engine/clients/redis/upload.py b/engine/clients/redis/upload.py
index cd4b888..0e29533 100644
--- a/engine/clients/redis/upload.py
+++ b/engine/clients/redis/upload.py
@@ -57,7 +57,7 @@ class RedisUploader(BaseUploader):
             cls.client.hset(
                 str(idx),
                 mapping={
-                    "vector": np.array(vec).astype(np.float32).tobytes(),
+                    "vector": np.array(vec).astype(np.float64).tobytes(),
                     **payload,
                     **geopoints,
                 },

```

[...]

PR #2491, opened 2024-08-16
feat(search): Support mutex to guarantee index update thread safety

### LINKED ISSUE #2489: Introduce Mutex Map in IndexInfo to Ensure Thread Safety During Index Construction
### Search before asking

### Motivation

In the context of index, the absence of a locking mechanism could lead to issues when multiple threads attempt to modify shared resources simultaneously. 

[1] Original issue raised from HNSW #2481 
[2] Marked TODO
https://github.com/apache/kvrocks/blob/3408318934cb68c0c1f2ffa937f19f02346d0672/src/server/redis_connection.cc#L522-L528

We might want to introduce locks to solve the problem. 

### Solution 1: Mutex (Straight-forward) 
We can introduce a mutex map within each IndexInfo structure. This mutex map would contain a mutex for each field that requires protection against concurrent access. 

#### Con
- Blocking if multiple connections update the same index

### Solution 2: Queue 
We can store the `Update` task for each `IndexInfo` field in a queue, and have a background thread to schedule tasks in order asynchronously. I'm not quite sure about how many number of threads would be appropriate, but can discuss further if everyone leans towards this solution. 

cc @PragmaTwice 

Beihao-Zhou:
Closes #2489

### Summary
This PR fixes the concurrency issue mentioned in #2481, however, this is now pretty slow for uploading operation and bounded by the computational resources from my local laptop. I'll set up the same instance capacity as mentioned in the [post](https://qdrant.tech/benchmarks/) and benchmark.

### DIFF
--- a/src/search/index_info.h
+++ b/src/search/index_info.h
@@ -23,6 +23,7 @@
 #include <algorithm>
 #include <map>
 #include <memory>
+#include <mutex>
 #include <string>
 #include <utility>
 
@@ -52,10 +53,12 @@ struct FieldInfo {
 
 struct IndexInfo {
   using FieldMap = std::map<std::string, FieldInfo>;
+  using MutexMap = std::map<std::string, std::mutex>;
 
   std::string name;
   redis::IndexMetadata metadata;
   FieldMap fields;
+  mutable MutexMap field_mutexes;
   redis::IndexPrefixes prefixes;
   std::string ns;
 
@@ -63,10 +66,15 @@ struct IndexInfo {
       : name(std::move(name)), metadata(std::move(metadata)), ns(std::move(ns)) {}
 
   void Add(FieldInfo &&field) {
-    const auto &name = field.name;
+    auto name = field.name;
     field.index = this;
     fields.emplace(name, std::move(field));
+    field_mutexes.emplace(std::piecewise_construct, std::make_tuple(name), std::make_tuple());
   }
+
+  void LockField(const std::string &field_name) const { field_mutexes.at(field_name).lock(); }
+
+  void UnLockField(const std::string &field_name) const { field_mutexes.at(field_name).unlock(); }
 };
 
 struct IndexMap : std::map<std::string, std::unique_ptr<IndexInfo>> {
--- a/src/search/indexer.cc
+++ b/src/search/indexer.cc
@@ -304,19 +304,23 @@ Status IndexUpdater::UpdateIndex(const std::string &field, std::string_view key,
     return {Status::NotOK, "No such field to do index updating"};
   }
 
+  info->LockField(field);
+
   auto *metadata = iter->second.metadata.get();
   SearchKey search_key(info->ns, info->name, field);
+  Status status;
   if (auto tag = dynamic_cast<TagFieldMetadata *>(metadata)) {
-    GET_OR_RET(UpdateTagIndex(key, original, current, search_key, tag));
+    status = UpdateTagIndex(key, original, current, search_key, tag);
   } else if (auto numeric [[maybe_unused]] = dynamic_cast<NumericFieldMetadata *>(metadata)) {
-    GET_OR_RET(UpdateNumericIndex(key, original, current, search_key, numeric));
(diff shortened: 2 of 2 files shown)


src/search/index_info.h
```diff
@@ -52,21 +53,28 @@ struct FieldInfo {
 
 struct IndexInfo {
   using FieldMap = std::map<std::string, FieldInfo>;
+  using MutexMap = std::map<std::string, std::mutex>;
```
PragmaTwice, line 56: Maybe we can put mutexes into `IndexUpdater`s instead of `IndexInfo`s?

Beihao-Zhou: Sure, this makes more sense <3

Beihao-Zhou: While trying to move the exact same piece of code into `IndexUpdater`s, I encountered the following errors:
```
‘constexpr std::pair<_T1, _T2>::pair(const std::pair<_T1, _T2>&) [with _T1 = const std::__cxx11::basic_string<char>; _T2 = std::mutex]’ is implicitly deleted because the default definition would be ill-formed:
  314 |       constexpr pair(const pair&) = default;    ///< Copy constructor
```
I suspect that this is caused by `IndexUpdater`s are usually created on stacks, which causes `std::map` doing some underlying copies and deletions, while `IndexInfo`s are allocated on heap so doesn't have the issue. 
I tried to wrap `std::mutex` into a `unique_ptr` but the compiler will report the same error. 

I'm not quite sure what I can do at this point, may leave it as it is, and try to benchmark first. But feel free to let me know if you think there is anything I missed, or any workaround you can think of. @PragmaTwice  <3

Beihao-Zhou:
Tried the benchmark framework today, which took more than one hour to upload, but didn't even finish. I thought there was something wrong and killed the process, but after double-checking the redis benchmark [here](https://qdrant.tech/benchmarks/), seems like it also took redis 1.5 hours. 
Will try it again these two days, but one call-out is that kvrocks definitely doesn't fully utilize the compute resources. 

Screen Shot 2024-08-18 at 5 51 38 PM

jonathanc-n:
@Beihao-Zhou Do you know if theres a place where you can configure your compute for kvrocks?

Beihao-Zhou:
> @Beihao-Zhou Do you know if theres a place where you can configure your compute for kvrocks?

Thanks a lot for pointing this out! I thought I was bottlenecked by lock itself so didn't try different config options, will look into them!

src/search/index_info.h
PragmaTwice: I think we can try `map<.., unique_ptr<mutex>>` to avoid such problem.

Beihao-Zhou:
Apology for the late update! Was busy with relocation for the past two weeks.  

Quick Update:
I tried to benchmark the default dataset upload operation, but it took around 4.5 hours and didn't even finish...
I can't quite estimate how long it will take, but notice that only one thread will use 100% CPU during the majority time. I'm not quite sure if this is because we used mutex here. 
I'll try to tune CPU usage in the config and run the benchmark again these two days; however, let me know if you have any better ideas! @PragmaTwice

src/search/index_info.h
PragmaTwice: Sorry for the delayed response.

I checked the related code, and I think there's two methods to solve it:
1. keep mutex maps inside `GlobalIndexer`, and add a raw pointer into `IndexUpdater` to access them;
2. follow the current way and modify methods like `void Add(IndexUpdater updater)` to `Add(std::unqiue_ptr<IndexUpdater>)`.

Beihao-Zhou: All good all good, thanks for sharing your solution! 
Might be a bit busy these two weeks, will be back to work on it during mid October.

PragmaTwice:
I'll take this task and submit a patch for it : )

Beihao-Zhou:
> I'll take this task and submit a patch for it : )

Thanks a ton!! Was too distracted by work these days.... My apology 😭

Issue #2422, opened 2024-07-15
Improve support for Redis connection management commands


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

https://redis.io/docs/latest/commands/?group=connection

There's many `CLIENT` subcommands that we can evaluate and see if we can support in Kvrocks, like `CLIENT ID/KILL/INFO/LIST/CACHING...`.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

furkan-bilgin:
Can I be assigned this issue?

git-hulk:
@furkan-bilgin Assigned, thank you!

PR #2495, opened 2024-08-18
feat: add CLIENT PAUSE and CLIENT REPLY subcommands

furkan-bilgin:
This PR is a part of #2422

I have added two subcommands for the command `CLIENT`:
- `CLIENT PAUSE`: Pauses processing of either `WRITE` or `ALL` commands until after `timeout` milliseconds.
- `CLIENT REPLY`: Stops server from sending replies to the client, starting; immediately if `OFF`, or skips one if `SKIP`. Starts sending replies if `ON`.


src/commands/cmd_server.cc
```diff
@@ -477,7 +477,53 @@ class CommandClient : public Commander {
...
+      if (args.size() != 3 && args.size() != 4) {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+
+      pause_timeout_ms_ = atoi(args[2].c_str());
+
+      if (args.size() == 3) {
+        pause_type_ = kPauseAll;
+      } else {
+        if (!strcasecmp(args[3].c_str(), "all")) {
```
git-hulk: Can use the CommandParser to parse command arguments.

src/server/redis_connection.cc
```diff
@@ -396,6 +413,13 @@ void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
     auto cmd_name = attributes->name;
     auto cmd_flags = attributes->GenerateFlags(cmd_tokens);
 
+    // Pause the processing of only the write commands, and push them back
+    // to a list that adds them back to the queue to process them when we unpause
+    if (srv_->GetCommandPauseType() == kPauseWrite && (cmd_flags & kCmdWrite)) {
```
git-hulk: I didn't check the Redis codebase, but it looks like the entire connection will be suspended like `PAUSE ALL` once the write command happens according to the documentation: https://redis.io/docs/latest/commands/client-pause/

> WRITE: Clients are only blocked if they attempt to execute a write command.

But from the current implementation, it will let read commands go through and it might cause unmatched responses on the client side.

git-hulk:
@furkan-bilgin Thanks for your contribution. Could you please help to pass the CI lint according to https://github.com/apache/kvrocks/actions/runs/10441248380/job/28912245323?pr=2495

src/server/redis_connection.cc
furkan-bilgin: [Looking at the Redis codebase](https://github.com/redis/redis/blob/3264deb24e3b01ad7b2777bc68b9d91f0f41aa3b/src/server.c#L4186), it seems to me that the server allows read commands to pass through when we are in `PAUSE WRITE` mode, and `PAUSE ALL` is when we defer all. Let me know if I got it right.
Also slaves should be kept exempt from this as per Redis, so I'll be working on that.

git-hulk: @furkan-bilgin Yes, what I mean is that the client should be pending once it occurred the WRITE command. However, the current implementation will let READ commands go through. [Redis](https://github.com/redis/redis/blob/3264deb24e3b01ad7b2777bc68b9d91f0f41aa3b/src/server.c#L4186) will pause the client once found the WRITE command.

furkan-bilgin: Alright, now I got it. Pushed the new commit.

src/commands/cmd_server.cc
furkan-bilgin: Done!

git-hulk:
@furkan-bilgin Looks good if you can add some Go test cases to cover this. And another point is that we should disable the `EV_READ` for connection while pausing?

PragmaTwice (commented):
Some tests should be added to demonstrate the functionality of the software as it includes changes in the critical execution path.

PR #2498, opened 2024-08-20
feat: support limit WriteBatch size,client outbuf limit and maxmemory clients 

AntiTopQuark:
relate to #2284 

- Added the ability to limit the size of WriteBatch operations in RocksDB to control memory usage effectively.
- Introduced client-output-buffer-limit configurations to manage client disconnections based on output buffer usage, enhancing stability, especially in high-throughput scenarios.
- Implemented a new maxmemory-clients configuration to enable client eviction based on overall memory usage, preventing potential OOM issues due to excessive client connections.

### DIFF
--- a/src/config/config_type.h
+++ b/src/config/config_type.h
@@ -65,8 +65,8 @@ class ConfigField {
   virtual std::string ToString() const = 0;
   virtual std::string ToStringForRewrite() const { return ToString(); }
   virtual Status Set(const std::string &v) = 0;
-  virtual Status ToNumber(int64_t *n) const { return {Status::NotOK, "not supported"}; }
-  virtual Status ToBool(bool *b) const { return {Status::NotOK, "not supported"}; }
+  virtual Status ToNumber([[maybe_unused]] int64_t *n) const { return {Status::NotOK, "not supported"}; }
+  virtual Status ToBool([[maybe_unused]] bool *b) const { return {Status::NotOK, "not supported"}; }
 
   ConfigType GetConfigType() const { return config_type; }
   bool IsMultiConfig() const { return config_type == ConfigType::MultiConfig; }
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -80,7 +80,7 @@ void Connection::Close() {
 
 void Connection::Detach() { owner_->DetachConnection(this); }
 
-void Connection::OnRead(struct bufferevent *bev) {
+void Connection::OnRead([[maybe_unused]] struct bufferevent *bev) {
   is_running_ = true;
   MakeScopeExit([this] { is_running_ = false; });
 
@@ -99,7 +99,7 @@ void Connection::OnRead(struct bufferevent *bev) {
   }
 }
 
-void Connection::OnWrite(bufferevent *bev) {
+void Connection::OnWrite([[maybe_unused]] bufferevent *bev) {
   if (IsFlagEnabled(kCloseAfterReply) || IsFlagEnabled(kCloseAsync)) {
     Close();
   }
--- a/src/server/worker.cc
+++ b/src/server/worker.cc
@@ -100,13 +100,14 @@ Worker::~Worker() {
   lua::DestroyState(lua_);
 }
 
-void Worker::TimerCB(int, int16_t events) {
+void Worker::TimerCB(int, [[maybe_unused]] int16_t events) {
   auto config = srv->GetConfig();
   if (config->timeout == 0) return;
   KickoutIdleClients(config->timeout);
 }
 
-void Worker::newTCPConnection(evconnlistener *listener, evutil_socket_t fd, sockaddr *address, int socklen) {
+void Worker::newTCPConnection(evconnlistener *listener, evutil_socket_t fd, [[maybe_unused]] sockaddr *address,
+                              [[maybe_unused]] int socklen) {
   int local_port = util::GetLocalPort(fd);  // NOLINT
   DLOG(INFO) << "[worker] New connection: fd=" << fd << " from port: " << local_port << " thread #" << tid_;
 
@@ -186,7 +187,8 @@ void Worker::newTCPConnection(evconnlistener *listener, evutil_socket_t fd, sock
   }
 }
 
-void Worker::newUnixSocketConnection(evconnlistener *listener, evutil_socket_t fd, sockaddr *address, int socklen) {
+void Worker::newUnixSocketConnection(evconnlistener *listener, evutil_socket_t fd, [[maybe_unused]] sockaddr *address,
+                                     [[maybe_unused]] int socklen) {
   DLOG(INFO) << "[worker] New connection: fd=" << fd << " from unixsocket: " << srv->GetConfig()->unixsocket
              << " thread #" << tid_;
   event_base *base = evconnlistener_get_base(listener);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -239,6 +239,7 @@ add_library(kvrocks_objs OBJECT ${KVROCKS_SRCS})
 target_include_directories(kvrocks_objs PUBLIC src src/common src/vendor ${PROJECT_BINARY_DIR} ${Backtrace_INCLUDE_DIR})
 target_compile_features(kvrocks_objs PUBLIC cxx_std_17)
 target_compile_options(kvrocks_objs PUBLIC -Wall -Wpedantic -Wsign-compare -Wreturn-type -fno-omit-frame-pointer)
+target_compile_options(kvrocks_objs PUBLIC -Werror=unused-parameter)
 target_compile_options(kvrocks_objs PUBLIC -Werror=unused-result)
 
 # disable unused-variable check on GCC < 8 due to the structure bindings
--- a/src/cli/main.cc
+++ b/src/cli/main.cc
@@ -47,7 +47,7 @@
 
 Server *srv = nullptr;
 
-extern "C" void SignalHandler(int sig) {
+extern "C" void SignalHandler([[maybe_unused]] int sig) {
   if (srv && !srv->IsStopped()) {
     LOG(INFO) << "Bye Bye";
     srv->Stop();
--- a/src/cli/signal_util.h
+++ b/src/cli/signal_util.h
@@ -30,7 +30,7 @@
 
 #include "version_util.h"
 
-extern "C" inline void SegvHandler(int sig, siginfo_t *info, void *secret) {
+extern "C" inline void SegvHandler(int sig, [[maybe_unused]] siginfo_t *info, [[maybe_unused]] void *secret) {
   LOG(ERROR) << "Ooops! Apache Kvrocks " << PrintVersion << " got signal: " << strsignal(sig) << " (" << sig << ")";
   auto trace = cpptrace::generate_trace();
   trace.print(LOG(ERROR));
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -353,7 +353,7 @@ Status Cluster::ImportSlotRange(redis::Connection *conn, const SlotRange &slot_r
       conn->SetImporting();
       myself_->importing_slot_range = slot_range;
       // Set link error callback
-      conn->close_cb = [object_ptr = srv_->slot_import.get(), slot_range](int fd) {
+      conn->close_cb = [object_ptr = srv_->slot_import.get(), slot_range]([[maybe_unused]] int fd) {
         auto s = object_ptr->StopForLinkError();
         if (!s.IsOK()) {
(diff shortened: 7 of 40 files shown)


src/server/worker.cc
```diff
@@ -446,6 +452,13 @@ Status Worker::Reply(int fd, const std::string &reply) {
...
   if (iter != conns_.end()) {
     iter->second->SetLastInteraction();
+    if (iter->second->CheckClientReachOBufLimits(reply.size())) {
+      srv->stats.IncrReachOutbufLimitDisconnections();
+      // unlock before calling FreeConnectionByID, otherwise it will cause nested deadlock
+      lock.unlock();
+      FreeConnectionByID(iter->first, iter->second->GetID());
+      return {Status::NotOK, "connection reach output buffer limits"};
+    }
     redis::Reply(iter->second->Output(), reply);
```
caipengbo: The reply will be added to libevent's evbuffer. Will the data accumulate in the evbuffer?

src/server/server.cc
```diff
@@ -2115,3 +2122,40 @@ AuthResult Server::AuthenticateUser(const std::string &user_password, std::strin
...
+  }
+  if (mem < config_->max_memory_clients) {
+    return;
+  }
+
+  std::vector<redis::Connection *> conns;
+  for (const auto &t : worker_threads_) {
+    auto worker_conns = t->GetWorker()->GetConnections();
+    for (const auto &iter : worker_conns) {
+      conns.push_back(iter.second);
```
caipengbo: Are these thread-safe? Is it possible that the connections has been released?

src/server/worker.cc
AntiTopQuark: Do you mean that publishing a large number of small messages caused the module consuming the `evbuffer` to not process them immediately, leading to memory bloat?

src/server/server.cc
AntiTopQuark: I referred to the implementation of the `void Server::decreaseWorkerThreads(size_t delta)` function, which is also called without locking.
https://github.com/AntiTopQuark/kvrocks/blob/43f14db05f73d02a6fefc95347fd5c79781f85a5/src/server/server.cc#L1895

src/server/worker.cc
caipengbo: Yes, I think the evbuffer is the client's final output buffer.

src/server/server.cc
caipengbo: @git-hulk I think there's something wrong here. Am I missing something?

caipengbo: In `MigrateConnection`, the connection cannot be released.

src/server/worker.cc
AntiTopQuark: done, thank you

src/server/server.cc
AntiTopQuark: done, thank you

src/config/config_type.h
```diff
@@ -338,3 +338,119 @@ class IntWithUnitField : public ConfigField {
...
+
+struct ClientOutputBufferLimitConfig {
+  uint64_t hard_limit_bytes = 0;
+  uint64_t soft_limit_bytes = 0;
+  uint64_t soft_limit_seconds = 0;
+};
+
+constexpr int CLIENT_TYPE_OBUF_COUNT = 3;
+
+class ClientOutputBufferLimitFiled : public ConfigField {
```
caipengbo: field

src/server/server.cc
caipengbo: The lock will not work here, and these connections may still be released when you access them below.

One of my own practices is to `maxmemory-clients/worker`, estimate the `maxmemory` of each worker, and then record (update in real time) the memory of each connection in each worker, which can be more timely and without locking.

src/server/worker.cc
```diff
@@ -446,6 +452,13 @@ Status Worker::Reply(int fd, const std::string &reply) {
   auto iter = conns_.find(fd);
   if (iter != conns_.end()) {
     iter->second->SetLastInteraction();
+    if (iter->second->CheckClientReachOBufLimits(reply.size() + evbuffer_get_length(iter->second->Output()))) {
```
caipengbo: This reply may become large during generation, such as the `KEYS` command.

src/config/config_type.h
AntiTopQuark: done, thank you

src/server/server.cc
AntiTopQuark: I referred to the implementation of the KickoutIdleClients() function. In FreeConnectionByID(), if the connection (conn) has already been released, no further action will be taken.

Additionally, may I ask if I can assume that, in most cases, the memory usage across different workers in the system is almost uniform, and that there is rarely a situation where memory usage is uneven?

src/server/worker.cc
AntiTopQuark: There are many commands that return results by using the `operator+= ` operation on the output string. Are you suggesting that every time `operator+=` is used, we should monitor and check the memory usage? Wouldn't this cause significant performance issues? Is it acceptable to make extensive changes to the code related to many commands?

src/server/server.cc
caipengbo: Yes, we usually assume that the worker is balanced. In fact, I do this, it is just a trade-off

src/server/worker.cc
caipengbo: There are not many such dangerous commands. I don't think the overhead per check will be too high (or we can check once every N times).

src/config/config_type.h
```diff
@@ -338,3 +338,119 @@ class IntWithUnitField : public ConfigField {
...
+    std::vector<std::string> tokens;
+    std::string token;
+    std::istringstream token_stream(str);
+    while (std::getline(token_stream, token, delim)) {
+      tokens.push_back(token);
+    }
+    return tokens;
+  }
+
+  static Status capacityToInt(const std::string &str, uint64_t &capacity) {
```
torwig: We have something similar here src/common/parse_util.cc -> ParseSizeAndUnit(). Can it serve your purpose here?

src/server/redis_connection.cc
```diff
@@ -97,6 +103,9 @@ void Connection::OnRead(struct bufferevent *bev) {
   if (IsFlagEnabled(kCloseAsync)) {
     Close();
   }
+  if (IsReachOBufLimit()) {
```
torwig: We can expand the letter `O` to `Output` and have something like `IsOutputBufferLimitReached`.

src/server/redis_connection.cc
```diff
@@ -559,4 +652,86 @@ void Connection::ResetMultiExec() {
   DisableFlag(Connection::kMultiExec);
 }
 
+bool Connection::CheckClientReachOBufLimits(size_t reply_bytes) {
```
torwig: The same here: `O` -> `Output`.

src/server/redis_connection.cc
```diff
@@ -559,4 +652,86 @@ void Connection::ResetMultiExec() {
...
+  assert(reply_bytes < SIZE_MAX - (1024 * 64));
+  if (reply_bytes == 0 && GetClientType() != kTypeSlave) {
+    return false;
+  }
+  int client_type = 0;
```
torwig: Maybe we can extract this logic to a separate function, introducing descriptive constants for 0, 1, 2.

src/server/redis_connection.cc
```diff
@@ -559,4 +652,86 @@ void Connection::ResetMultiExec() {
...
+      }
+    }
+  } else {
+    SetObufSoftLimitReachedTime(0);
+  }
+  return hard || soft;
+}
+
+std::string Connection::CheckClientReachOBufLimits(const std::string &msg) {
+  auto memSize = msg.size() + GetOutputBuffer().capacity() + evbuffer_get_length(Output());
```
torwig: `memSize` -> `mem_size`

src/server/redis_connection.cc
```diff
@@ -559,4 +652,86 @@ void Connection::ResetMultiExec() {
...
+  if (CheckClientReachOBufLimits(memSize)) {
+    SetReachOBufLimit(true);
+    return "";
+  }
+  return msg;
+}
+
+size_t Connection::GetConnectionMemoryUsed() const {
+  size_t total_memory = sizeof(*this);  // 包含所有成员变量的静态内存大小
```
torwig: Could you please translate the comment into English?

src/server/redis_connection.h
```diff
@@ -180,6 +172,15 @@ class Connection : public EvbufCallbackBase<Connection> {
   std::set<std::string> watched_keys;
   std::atomic<bool> watched_keys_modified = false;
 
+  bool CheckClientReachOBufLimits(size_t reply_bytes);
```
torwig: In all functions of this "family" we can expand  a single `O` letter to `Output`

src/server/redis_connection.h
```diff
@@ -180,6 +172,15 @@ class Connection : public EvbufCallbackBase<Connection> {
...
+  std::string CheckClientReachOBufLimits(const std::string &msg);
+  void SetObufSoftLimitReachedTime(int64_t time) { obuf_soft_limit_reached_time_ = time; }
+  int64_t GetObufSoftLimitReachedTime() const { return obuf_soft_limit_reached_time_; }
+  inline std::string &GetOutputBuffer() { return output_buffer_; }
+  size_t GetConnectionMemoryUsed() const;
+  void SetReachOBufLimit(bool reach) { is_reach_obuf_limit_ = reach; }
+  bool IsReachOBufLimit() const { return is_reach_obuf_limit_; }
```
torwig: `IsReachOBufLimit` -> `IsOutputBufLimitReached`.

src/server/server.cc
```diff
@@ -839,6 +839,11 @@ void Server::cron() {
       cleanupExitedWorkerThreads(false);
     }
 
+    // check if we need to evict connections every 10s
+    if (counter != 0 && counter % 100 == 0 && config_->max_memory_clients > 0) {
```
torwig: Can we make the condition `config_->max_memory_clients > 0` the first one to benefit from fast evaluation if this option is disabled in the config?

src/server/worker.cc
```diff
@@ -102,8 +102,13 @@ Worker::~Worker() {
...
 void Worker::TimerCB(int, int16_t events) {
   auto config = srv->GetConfig();
-  if (config->timeout == 0) return;
-  KickoutIdleClients(config->timeout);
+  if (config->timeout != 0) {
+    KickoutIdleClients(config->timeout);
+  }
+
+  if (config->IsClientOutputBufferLimitsEnabled()) {
+    KickoutReachOBufLimitsClients();
```
torwig: The same: `O` -> `Output`

torwig (commented):
@AntiTopQuark Thank you for your rigorous work.
I left some comments.

torwig:
@AntiTopQuark Ideally extract each feature in a separate PR as they are independent as far as I understand.
in a separate PR. It's not **strongly** required but it would make your PRs more granular and [maybe] smaller in size :)

torwig:
@AntiTopQuark Running `./x.py format` should make the linter happy.

git-hulk:
To be honest, I'm worried about the performance and maintenance burden of this implementation. Maybe we should separate it into following PRs to make it easier to merge and review:

- Allow to control the maximum size of the write batch
- Allow to control the maximum size of connection input/output buffer(just check the size of libevent buffer)
- Allow to control the maximum size of all clients(need a clear implementation proposal first)

What do you think? @torwig @caipengbo @mapleFU @PragmaTwice

caipengbo:
> Maybe we should separate it into following PRs to make it easier to merge and review

I agree with it.

src/server/worker.cc
AntiTopQuark: Hi, I noticed that the operator += and append are used quite frequently. I need to convert all `redis::xxx` to `GET_OR_RET(conn->xxx())` to check whether the output buffer size meets the expectations each time. Even if I change it to check the client output buffer every 10 times, I suspect this might have a significant impact on performance.

image
image


My idea is to create multiple member functions for conn to replace the original redis::xxxx() functions, so that the memory size check can be performed each time an output is made. If the check fails, an error code will be thrown.

```c++
StatusOr<std::string> Connection::SimpleString(const std::string &msg) {
  auto res = redis::SimpleString(msg);
  if (CheckClientReachOutputBufferLimits(res)) {
    return {Status::ReachClientOutputBufferLimit, "client reached output buffer limits"};
  }
  return res;
}

-   *output += redis::SimpleString("Capacity");
-   *output += redis::SimpleString("Size");
+  *output += GET_OR_RET((conn->SimpleString("Capacity"));
+  *output += GET_OR_RET((conn->SimpleString("Size"));
```

PR #2499, opened 2024-08-20
fix(config):  wrong rocksdb config name max_sub_compactions

ChrisZMF:
Configuration `item rocksdb.max_sub_compactions` should be `rocksdb.max_subcompactions`.
Try set `rocksdb.max_sub_compactions`:
image
Fix it as `rocksdb.max_subcompactions`, try again:
image

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -311,7 +311,7 @@ max-backup-keep-hours 24
 # Default: 16
 max-bitmap-to-string-mb 16
 
-# Whether to enable SCAN-like cursor compatible with Redis. 
+# Whether to enable SCAN-like cursor compatible with Redis.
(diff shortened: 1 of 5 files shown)


tisonkun (dismissed):
LGTM. But seems no regression tests to guard this manner ..

git-hulk (approved):
LGTM, good catch.

ChrisZMF:
> LGTM. But seems no regression tests to guard this manner ..

Unit tests can cover this scenario. The reason why unit test was ineffective before is that callback was not registered for the item, which means `config set` command did not actually execute in rocksdb. See
image

mapleFU:
We can first checkin, and adding a cfg testing, and everyone we add a new config, a go-side cfg test should be added to cover the case?

PR #2506, opened 2024-08-25
feat(config): add `txn_context_enabled` to allow to enable the transaction feature

### LINKED ISSUE #2504: Provide a configuration to turn off transactional Context
### Search before asking

### Motivation

https://github.com/apache/kvrocks/pull/2332 may have a wide range of impacts.
Users may need a configuration to avoid these impacts.

### Solution

_No response_

PokIsemaine:
fix: https://github.com/apache/kvrocks/issues/2504

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -336,8 +336,18 @@ json-max-nesting-depth 1024
 json-storage-format json
 
 # Whether to enable transactional mode engine::Context.
-# Default: yes
-txn-context-enabled yes
+#
+# If enabled, is_txn_mode in engine::Context will be set properly,
+# which is expected to improve the consistency of commands.
+# If disabled, is_txn_mode in engine::Context will be set to false,
+# making engine::Context equivalent to engine::Storage.
+#
+# NOTE: This is an experimental feature. If you find errors, performance degradation,
+# excessive memory usage, excessive disk I/O, etc. after enabling it, please try disabling it.
+# At the same time, we welcome feedback on related issues to help iterative improvements.
+# 
+# Default: no
+txn-context-enabled no
 
 ################################## TLS ###################################
 
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -335,6 +335,10 @@ json-max-nesting-depth 1024
 # Default: json
 json-storage-format json
 
+# Whether to enable transactional mode engine::Context.
+# Default: yes
+txn-context-enabled yes
+
 ################################## TLS ###################################
 
 # By default, TLS/SSL is disabled, i.e. `tls-port` is set to 0.
--- a/src/config/config.h
+++ b/src/config/config.h
@@ -169,7 +169,7 @@ struct Config {
   JsonStorageFormat json_storage_format = JsonStorageFormat::JSON;
 
   // transactional mode engine::Context
-  bool txn_context_enabled = true;
+  bool txn_context_enabled = false;
 
   struct RocksDB {
     int block_size;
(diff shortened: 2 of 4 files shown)


kvrocks.conf
```diff
@@ -335,6 +335,10 @@ json-max-nesting-depth 1024
 # Default: json
 json-storage-format json
 
+# Whether to enable transactional mode engine::Context.
```
torwig, line 338: Can you please expand this comment by explaining what this mode will change if enabled/disabled?
Why it's enabled by default?

PokIsemaine: Okay, I've explained where the changes occurred and I'll add them to the config file.

enabled by default: This is a experimental feature that is expected to enhance the consistency of kvrocks. Enabled by default in the hope of collecting more feedback.

torwig: @PokIsemaine Two points:
1. Users usually don't read the source code so we need to explain them in the config file or other docs. The key question for users is "What benefit do I gain if I enable the feature?"
2. IMHO,  if the feature **is** experimental. it should **not** be enabled by default.

PokIsemaine: Thank you for your explanation, disabled by default is more user friendly.

Aside from that, I'd like to respectfully seek your advice. For experimental features, does the community have a standard experimental process? For example, are there guidelines on when a feature should graduate to become a formal one, when it should be discontinued, and how we can quickly gather feedback?

From a developer's perspective, this feature has broad implications and could affect subsequent development. I'm concerned that some issues may not surface until later, making it difficult and time-consuming to either discontinue or refine the feature down the line. Could you provide any suggestions or best practices on this matter?

src/config/config.h
```diff
@@ -168,6 +168,9 @@ struct Config {
   int json_max_nesting_depth = 1024;
   JsonStorageFormat json_storage_format = JsonStorageFormat::JSON;
 
+  // transactional mode engine::Context
```
mapleFU:
```suggestion
  // Enable transactional mode in engine::Context
```

mapleFU (approved):

kvrocks.conf
torwig: @PokIsemaine I'm not sure we have a flow for experimental features. Additionally, it's hard to force users to test the experimental feature on their valuable production data :)

torwig (approved):

git-hulk (approved):

kvrocks.conf
```diff
@@ -335,6 +335,20 @@ json-max-nesting-depth 1024
...
+# which is expected to improve the consistency of commands.
+# If disabled, is_txn_mode in engine::Context will be set to false,
+# making engine::Context equivalent to engine::Storage.
+#
+# NOTE: This is an experimental feature. If you find errors, performance degradation,
+# excessive memory usage, excessive disk I/O, etc. after enabling it, please try disabling it.
+# At the same time, we welcome feedback on related issues to help iterative improvements.
+# 
+# Default: no
+txn-context-enabled no
```
PragmaTwice, line 350: maybe we can add some go cases for "yes" mode.

PokIsemaine: Yeah, we need to test, but "some" may actually be almost all.
Is it possible to add a "yes" mode process to the GitHub Actions?

PragmaTwice: You can just change the go cases to make some test functions like `TestXX` work on both mode, e.g.
```
TestXXAllMode() {
  TestXX(yes)
  TestXX(no)
}
```

Currently it is not required to include all test cases. Just some basic tests to ensure it can works.

PokIsemaine: Okay, how about adding these tests in another PR? I need to go out recently, so I may not be able to update immediately. You can merge this PR first, and then I will create an issue to track it and solve it when I have free time.

mapleFU:
Will merge this, lets' adding test in another pr. So sorry for delaying

PR #2507, opened 2024-08-26
chore: Enable -Werror=unused-parameter flag

### LINKED ISSUE #2266: Treat `unused-parameter` warnings as errors in CMake

c8ef:
- resolve: #2266
- supersede: #2313

### DIFF
--- a/src/vendor/rand.cc
+++ b/src/vendor/rand.cc
@@ -106,11 +106,6 @@ constexpr void SEED(const T x0, const T x1, const T x2) {
   SET3(x, x0, x1, x2), SET3(a, A0, A1, A2), c = C;
 }
 
-template <typename T>
-constexpr T HiBit(const T x) {
-  return (1L << (2 * N - 1));
-}
-
 static void Next() {
   uint32_t p[2], q[2], r[2], carry0 = 0, carry1 = 0;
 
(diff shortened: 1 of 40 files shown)


src/vendor/rand.cc
```diff
@@ -106,11 +106,6 @@ constexpr void SEED(const T x0, const T x1, const T x2) {
   SET3(x, x0, x1, x2), SET3(a, A0, A1, A2), c = C;
 }
 
-template <typename T>
-constexpr T HiBit(const T x) {
```
c8ef, line 110: This function appears to be unused.

mapleFU: this is vendored ( pasted from other project )

Maybe we can disable lint for "vendor" dir?

c8ef: Even though it's in the vendored directory, I believe it's more of a modern C++ re-implementation based on the comment below. Therefore, removing the unused function should be acceptable, at least NFC.

https://github.com/apache/kvrocks/blob/5ecf023bcf0a92e23982fef1710a53f495cfef9a/src/vendor/rand.cc#L44-L47

c8ef:
I don't know why [this test](https://github.com/apache/kvrocks/actions/runs/10562102251/job/29261542437#logs) failed. I believe the entire patch should be NFC. Other CI failures have already been fixed in the latest patch. Please allow the CI to run.

git-hulk (approved):

torwig (approved):

PR #2508, opened 2024-08-28
feat: support limit WriteBatch size

AntiTopQuark:
relate to https://github.com/apache/kvrocks/issues/2284

Added the ability to limit the size of WriteBatch operations in RocksDB to control memory usage effectively.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -953,5 +953,12 @@ rocksdb.rate_limiter_auto_tuned yes
 # Default: yes
 # rocksdb.avoid_unnecessary_blocking_io yes
 
+# Specifies the maximum size in bytes for a write batch in RocksDB.
+# If set to 0, there is no size limit for write batches.
+# This option can help control memory usage and manage large WriteBatch operations more effectively.
+# 
+# Default: 0 (no size limit)
+# rocksdb.write_options.write_batch_max_bytes 0
+
 ################################ NAMESPACE #####################################
 # namespace.test change.me
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -253,7 +253,7 @@ Config::Config() {
       {"rocksdb.write_options.low_pri", true, new YesNoField(&rocks_db.write_options.low_pri, false)},
       {"rocksdb.write_options.memtable_insert_hint_per_batch", true,
        new YesNoField(&rocks_db.write_options.memtable_insert_hint_per_batch, false)},
-      {"rocksdb.write_options.write_batch_max_bytes", true,
+      {"rocksdb.write_options.write_batch_max_bytes", false,
        new IntField(&rocks_db.write_options.write_batch_max_bytes, 0, 0, INT_MAX)},
 
       /* rocksdb read options */
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -253,6 +253,8 @@ Config::Config() {
       {"rocksdb.write_options.low_pri", true, new YesNoField(&rocks_db.write_options.low_pri, false)},
       {"rocksdb.write_options.memtable_insert_hint_per_batch", true,
        new YesNoField(&rocks_db.write_options.memtable_insert_hint_per_batch, false)},
+      {"rocksdb.write_options.write_batch_max_bytes", true,
+       new IntField(&rocks_db.write_options.write_batch_max_bytes, 0, 0, INT_MAX)},
 
       /* rocksdb read options */
       {"rocksdb.read_options.async_io", false, new YesNoField(&rocks_db.read_options.async_io, true)},
--- a/src/search/hnsw_indexer.cc
+++ b/src/search/hnsw_indexer.cc
@@ -49,11 +49,16 @@ StatusOr<HnswNodeFieldMetadata> HnswNode::DecodeMetadata(engine::Context& ctx, c
   return metadata;
 }
 
-void HnswNode::PutMetadata(HnswNodeFieldMetadata* node_meta, const SearchKey& search_key, engine::Storage* storage,
-                           rocksdb::WriteBatchBase* batch) const {
+Status HnswNode::PutMetadata(HnswNodeFieldMetadata* node_meta, const SearchKey& search_key, engine::Storage* storage,
+                             rocksdb::WriteBatchBase* batch) const {
   std::string updated_metadata;
   node_meta->Encode(&updated_metadata);
-  batch->Put(storage->GetCFHandle(ColumnFamilyID::Search), search_key.ConstructHnswNode(level, key), updated_metadata);
+  auto s = batch->Put(storage->GetCFHandle(ColumnFamilyID::Search), search_key.ConstructHnswNode(level, key),
+                      updated_metadata);
+  if (!s.ok()) {
+    return {Status::NotOK, s.ToString()};
+  }
+  return Status::OK();
 }
 
 void HnswNode::DecodeNeighbours(engine::Context& ctx, const SearchKey& search_key) {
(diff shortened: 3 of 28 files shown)


src/config/config.cc
```diff
@@ -253,6 +253,8 @@ Config::Config() {
       {"rocksdb.write_options.low_pri", true, new YesNoField(&rocks_db.write_options.low_pri, false)},
       {"rocksdb.write_options.memtable_insert_hint_per_batch", true,
        new YesNoField(&rocks_db.write_options.memtable_insert_hint_per_batch, false)},
+      {"rocksdb.write_options.write_batch_max_bytes", true,
```
git-hulk: Can this configuration be modified inflight? if yes, we can change `true` to `false`.

git-hulk:
@AntiTopQuark Thanks for your efforts. Would you mind adding a Go test case for this?

src/config/config.cc
AntiTopQuark: You are right, thanks～

AntiTopQuark:
> @AntiTopQuark Thanks for your efforts. Would you mind adding a Go test case for this?

Sure, I will complete this task later.

AntiTopQuark:
While adding cases, it was found that there were no corresponding error messages, so error codes were added and reported in many places involving batch processing

git-hulk:
> While adding cases, it was found that there were no corresponding error messages, so error codes were added and reported in many places involving batch processing

Yes, that makes sense.

src/search/indexer.cc
```diff
@@ -226,13 +226,19 @@ Status IndexUpdater::UpdateTagIndex(engine::Context &ctx, std::string_view key,
   for (const auto &tag : tags_to_delete) {
     auto index_key = search_key.ConstructTagFieldData(tag, key);
 
-    batch->Delete(cf_handle, index_key);
+    auto rocker_status = batch->Delete(cf_handle, index_key);
```
git-hulk: Why not name it to `s` to keep consistent with others.

src/search/indexer.cc
```diff
@@ -226,13 +226,19 @@ Status IndexUpdater::UpdateTagIndex(engine::Context &ctx, std::string_view key,
...
+    if (!rocker_status.ok()) {
+      return {Status::NotOK, rocker_status.ToString()};
+    }
   }
 
   for (const auto &tag : tags_to_add) {
     auto index_key = search_key.ConstructTagFieldData(tag, key);
 
-    batch->Put(cf_handle, index_key, Slice());
+    auto rocker_status = batch->Put(cf_handle, index_key, Slice());
```
git-hulk: Same as above.

src/types/redis_stream.cc
```diff
@@ -1619,11 +1672,12 @@ uint64_t Stream::trim(engine::Context &ctx, const std::string &ns_key, const Str
...
   }
 
-  if (ret > 0) {
+  if (delete_cnt > 0) {
     metadata->max_deleted_entry_id = entryIDFromInternalKey(last_deleted);
   }
 
-  return ret;
+  return rocksdb::Status::OK();
+  ;
```
git-hulk:
```suggestion
  return rocksdb::Status::OK();
```

git-hulk: Can use `StatusOr` instead of introducing a new input reference.

src/types/redis_zset.cc
```diff
@@ -83,14 +84,17 @@ rocksdb::Status ZSet::Add(engine::Context &ctx, const Slice &user_key, ZAddFlags
           old_score_bytes.append(it->member);
           std::string old_score_key =
               InternalKey(ns_key, old_score_bytes, metadata.version, storage_->IsSlotIdEncoded()).Encode();
-          batch->Delete(score_cf_handle_, old_score_key);
+          s = batch->Delete(score_cf_handle_, old_score_key);
+          if (!s.ok() && !s.IsNotFound()) return s;
```
git-hulk: IsNotFound should never happen in batch->Delete and batch->Put?

src/types/redis_zset.cc
```diff
@@ -101,18 +105,21 @@ rocksdb::Status ZSet::Add(engine::Context &ctx, const Slice &user_key, ZAddFlags
     }
     std::string score_bytes;
     PutDouble(&score_bytes, it->score);
-    batch->Put(member_key, score_bytes);
+    s = batch->Put(member_key, score_bytes);
+    if (!s.ok() && !s.IsNotFound()) return s;
```
git-hulk: same as above

src/types/redis_zset.cc
AntiTopQuark: done, thanks

src/search/indexer.cc
AntiTopQuark: done

src/search/indexer.cc
AntiTopQuark: done

src/types/redis_zset.cc
AntiTopQuark: done

src/types/redis_stream.cc
AntiTopQuark: I tried it and I can't use `StatusOr<T>` as the return value because `rocksdb::Status` and `Status` are different and `rocksdb::Status` lacks methods like `GetValue()`.

git-hulk: I see. Can you change `&delete_cnt` to `*delete_cnt`?

git-hulk:
@AntiTopQuark One comment inline, rest are good to me. Could anyone also have a look at this PR? @PragmaTwice @torwig @caipengbo @mapleFU

kvrocks.conf
```diff
@@ -953,5 +953,12 @@ rocksdb.rate_limiter_auto_tuned yes
 # Default: yes
 # rocksdb.avoid_unnecessary_blocking_io yes
 
+# Specifies the maximum size in bytes for a write batch in RocksDB.
+# If set to 0, there is no size limit for write batches.
+# This option can help control memory usage and manage large WriteBatch operations more effectively.
+# 
+# Default: 0 (no size limit)
```
caipengbo:
```suggestion
# Default: 0
```

It's the same thing as above. Personal opinion, it doesn't really matter.

caipengbo (dismissed):
LGTM

kvrocks.conf
AntiTopQuark: done, thank u

caipengbo (dismissed):

caipengbo (dismissed):

git-hulk (dismissed):

src/search/hnsw_indexer.cc
```diff
@@ -75,11 +80,16 @@ void HnswNode::DecodeNeighbours(engine::Context& ctx, const SearchKey& search_ke
...
+    return {Status::NotOK, rocket_s.ToString()};
+  }
   HnswNodeFieldMetadata node_metadata = GET_OR_RET(DecodeMetadata(ctx, search_key));
   node_metadata.num_neighbours++;
-  PutMetadata(&node_metadata, search_key, ctx.storage, batch);
+  auto s = PutMetadata(&node_metadata, search_key, ctx.storage, batch);
+  if (!s.IsOK()) {
+    return s;
+  }
   return Status::OK();
```
mapleFU:
```suggestion
  return PutMetadata(&node_metadata, search_key, ctx.storage, batch);
```

src/search/hnsw_indexer.cc
```diff
@@ -93,7 +103,10 @@ Status HnswNode::RemoveNeighbour(engine::Context& ctx, const NodeKey& neighbour_
 
   HnswNodeFieldMetadata node_metadata = GET_OR_RET(DecodeMetadata(ctx, search_key));
   node_metadata.num_neighbours--;
-  PutMetadata(&node_metadata, search_key, ctx.storage, batch);
+  auto redis_status = PutMetadata(&node_metadata, search_key, ctx.storage, batch);
```
mapleFU: RETURN_NOT_OK or just return this?

mapleFU (changes requested):
This pr LGTM. But it can be shorter with macros

src/search/hnsw_indexer.cc
AntiTopQuark: done

src/search/hnsw_indexer.cc
AntiTopQuark: done

mapleFU (approved):
Also cc @PragmaTwice

Do we need some macro like `RETURN_NOT_OK`, `RETURN_NOT_OK_FROM_ROCKSDB`?

git-hulk:
> Also cc @PragmaTwice
> 
> Do we need some macro like `RETURN_NOT_OK`, `RETURN_NOT_OK_FROM_ROCKSDB`?

@mapleFU @AntiTopQuark `RETURN_IF_ERR` is also good for this?

mapleFU:
I think it's good. Another problem is that we may need type cast between `Status` and `rocksdb::Status`. We can make a new issue here.

git-hulk:
Yes, let's discuss this in new issue.

PragmaTwice:
> Also cc @PragmaTwice
> 
> Do we need some macro like `RETURN_NOT_OK`, `RETURN_NOT_OK_FROM_ROCKSDB`?

Nope. They can be done by GET_OR_RET.

mapleFU:
> Nope. They can be done by GET_OR_RET.

So `GET_OR_RET` can also handle the case for pure "Status"?

Besides, what about err from rocksdb?

PR #2513, opened 2024-08-30
feat: support maxmemory-clients config

AntiTopQuark:
relate to https://github.com/apache/kvrocks/issues/2284
Implemented a new maxmemory-clients configuration to enable client eviction based on overall memory usage, preventing potential OOM issues due to excessive client connections.

### DIFF
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -563,7 +563,7 @@ void Connection::ResetMultiExec() {
 }
 
 size_t Connection::GetConnectionMemoryUsed() const {
-  size_t total_memory = sizeof(*this);  // 包含所有成员变量的静态内存大小
+  size_t total_memory = sizeof(*this);
 
   total_memory += name_.capacity();
   total_memory += ns_.capacity();
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -52,6 +52,7 @@ Connection::Connection(bufferevent *bev, Worker *owner)
   create_time_ = now;
   last_interaction_ = now;
   output_buffer_.clear();
+  slave_output_buffer_.clear();
 }
 
 Connection::~Connection() {
@@ -369,9 +370,9 @@ static bool IsCmdForIndexing(const CommandAttributes *attr) {
 }
 
 void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
-  output_buffer_.clear();
+  GetOutputBuffer().clear();
   const Config *config = srv_->GetConfig();
-  std::string &reply = output_buffer_;
+  std::string &reply = GetOutputBuffer();
   std::string password = config->requirepass;
 
   while (!to_process_cmds->empty()) {
@@ -571,6 +572,7 @@ size_t Connection::GetConnectionMemoryUsed() const {
   total_memory += addr_.capacity();
   total_memory += last_cmd_.capacity();
   total_memory += output_buffer_.capacity();
+  total_memory += slave_output_buffer_.capacity();
   total_memory += evbuffer_get_length(Output()) + evbuffer_get_length(Input());
 
   for (const auto &channel : subscribe_channels_) {
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -51,6 +51,7 @@ Connection::Connection(bufferevent *bev, Worker *owner)
   int64_t now = util::GetTimeStamp();
   create_time_ = now;
   last_interaction_ = now;
+  output_buffer_.clear();
 }
 
 Connection::~Connection() {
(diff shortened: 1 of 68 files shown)


src/server/server.cc
```diff
@@ -1136,12 +1141,13 @@ void Server::GetStatsInfo(std::string *info) {
...
   string_stream << "sync_partial_ok:" << stats.psync_ok_count << "\r\n";
   string_stream << "sync_partial_err:" << stats.psync_err_count << "\r\n";
+  string_stream << "evicted_clients:" << stats.stat_evictedclients << "\r\n";
 
   auto db_stats = storage->GetDBStats();
   string_stream << "keyspace_hits:" << db_stats->keyspace_hits << "\r\n";
   string_stream << "keyspace_misses:" << db_stats->keyspace_misses << "\r\n";
 
-  {
+  {https://www.anyknew.com/go/10112884
```
PragmaTwice: hmm

AntiTopQuark: I'm sorry for the mistake

AntiTopQuark: I'll review the code again shortly.

caipengbo:
I want to discuss how we can implement this feature. My idea is to count client memory within each worker (using `libevent`'s callback to count `evbuffer`) to avoid locking across workers.  @AntiTopQuark @PragmaTwice @git-hulk @mapleFU @torwig

git-hulk:
> I want to discuss how we can implement this feature. My idea is to count client memory within each worker (using `libevent`'s callback to count `evbuffer`) to avoid locking across workers. @AntiTopQuark @PragmaTwice @git-hulk @mapleFU @torwig

Yes, as mentioned in [previous comment](https://github.com/apache/kvrocks/pull/2498#issuecomment-2311996885). It would be better to limit the connection output first by checking the libevent buffer. And for the `maxmemory-clients`, I think we don't need to count every byte, just checking the libevent input and output buffer is good to me as well.

AntiTopQuark:
> > I want to discuss how we can implement this feature. My idea is to count client memory within each worker (using `libevent`'s callback to count `evbuffer`) to avoid locking across workers. @AntiTopQuark @PragmaTwice @git-hulk @mapleFU @torwig
> 
> Yes, as mentioned in [previous comment](https://github.com/apache/kvrocks/pull/2498#issuecomment-2311996885). It would be better to limit the connection output first by checking the libevent buffer. And for the `maxmemory-clients`, I think we don't need to count every byte, just checking the libevent input and output buffer is good to me as well.

I have modified the code and moved the `evictionClients` function into `Worker`.
On another note, I believe that `maxmemory-clients` is used to limit the total size of connections. If you want to limit the size of the output buffer, you can use the client-output-buffer-limit configuration option. I will submit a PR for this in the future.

mapleFU:
>  want to discuss how we can implement this feature. My idea is to count client memory within each worker (using libevent's callback to count evbuffer) to avoid locking across workers. 

Yeah I think holding a lock is harmful here, I can accept some untimely memory change, but global checking is harmful...

src/server/redis_connection.cc
```diff
@@ -559,4 +562,37 @@ void Connection::ResetMultiExec() {
   DisableFlag(Connection::kMultiExec);
 }
 
+size_t Connection::GetConnectionMemoryUsed() const {
+  size_t total_memory = sizeof(*this);  // 包含所有成员变量的静态内存大小
```
PragmaTwice: please dont include chinese comments.

src/server/redis_connection.cc
```diff
@@ -559,4 +562,37 @@ void Connection::ResetMultiExec() {
...
+  for (const auto &channel : subscribe_shard_channels_) {
+    total_memory += channel.capacity();
+  }
+  for (const auto &cmd : multi_cmds_) {
+    total_memory += cmd.capacity();
+  }
+
+  if (saved_current_command_) {
+    total_memory += saved_current_command_->GetMemoryUsage();
+  }
```
PragmaTwice, line 593: The code seems hard to maintain and also not like a precise estimation of memory usage.

I'm wondering if it's really necessary to limit the connection memory usage..

PragmaTwice:
I believe implementing an elegant solution to track and restrict client connection memory usage is challenging and could reduce code maintainability. 

Instead, I suggest exploring alternative approaches, like limiting the number of clients.

src/server/redis_connection.cc
AntiTopQuark: done

AntiTopQuark:
> I believe implementing an elegant solution to track and restrict client connection memory usage is challenging and could reduce code maintainability.
> 
> Instead, I suggest exploring alternative approaches, like limiting the number of clients.



> I believe implementing an elegant solution to track and restrict client connection memory usage is challenging and could reduce code maintainability.
> 
> Instead, I suggest exploring alternative approaches, like limiting the number of clients.

When I looked at Redis previously, it uses the maxclients configuration option to limit the number of clients, and maxmemory-clients to limit the total memory used by connections. If you only want to limit the output buffer, you can use client-output-buffer-limit to restrict the usage per connection.

AntiTopQuark:
@PragmaTwice  @git-hulk  Hi, should we continue pushing this PR forward, or can it be closed?

Issue #2425, opened 2024-07-15
Support count-min sketch data structure and commands


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Redis Stack already supports this probabilistic data structure, and we can support it too.

https://redis.io/docs/latest/develop/data-types/probabilistic/count-min-sketch/
https://redis.io/docs/latest/commands/?group=cms

### Solution

Please share your design proposal through issue/discussion before submitting a PR.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

jonathanc-n:
Can I be assigned this issue?

mapleFU:
@jonathanc-n FYI, you can first add a proposal, like [1][2], then after discussion, you can draft the website work and in the mean time working on implementation

[1] https://github.com/apache/kvrocks/discussions/2438
[2] https://github.com/apache/kvrocks/discussions/2449
[3] https://github.com/apache/kvrocks-website/pull/207

Tangruilin:
> [@jonathanc-n](https://github.com/jonathanc-n) FYI, you can first add a proposal, like [1][2], then after discussion, you can draft the website work and in the mean time working on implementation
> 
> [1] [#2438](https://github.com/apache/kvrocks/discussions/2438) [2] [#2449](https://github.com/apache/kvrocks/discussions/2449) [3] [apache/kvrocks-website#207](https://github.com/apache/kvrocks-website/pull/207)

https://github.com/apache/kvrocks/discussions/3404
here is my proposal

PR #2524, opened 2024-09-07
feat: support count min sketch data structure and most commands

jonathanc-n:
references issue: #2425 

The commands, IncrBy, Info, InitByDim, InitByProb, and Query have been made.

The merge command will still need working on, and go integration tests will be added for the future. I will also probably add a cache similar to the hyperloglog for in-memory operations.


src/commands/commander.h
```diff
@@ -73,6 +73,7 @@ enum class CommandCategory : uint8_t {
   Bit,
   BloomFilter,
   Cluster,
+  CMS,
```
mapleFU, line 76: Would this better being CMSketh? Or `CountMinSketch`?

src/types/cms.h
```diff
@@ -0,0 +1,82 @@
...
+ * under the License.
+ *
+ */
+
+#pragma once
+
+#include <assert.h>
+
+#include <atomic>
+#include <iostream>
```
mapleFU: Why?

src/types/cms.h
```diff
@@ -0,0 +1,82 @@
...
+#include <vector>
+
+#include "vendor/murmurhash2.h"
+
+class CMSketch {
+ public:
+  explicit CMSketch(uint32_t width = 0, uint32_t depth = 0, uint64_t counter = 0, std::vector<uint32_t> array = {})
```
mapleFU: Why using default args?

src/types/cms.h
```diff
@@ -0,0 +1,82 @@
...
+      : width_(width),
+        depth_(depth),
+        counter_(counter),
+        array_(array.empty() ? std::vector<uint32_t>(width * depth, 0) : std::move(array)) {}
+
+  ~CMSketch() = default;
```
mapleFU: Remove this?

src/types/cms.h
```diff
@@ -0,0 +1,82 @@
...
+
+  static CMSketch* NewCMSketch(size_t width, size_t depth) { return new CMSketch(width, depth); }
```
mapleFU: Why call new explicitly? Shouldn't a unique_ptr being returned in this case?

src/types/cms.h
```diff
@@ -0,0 +1,82 @@
...
+
+  static void CMSDimFromProb(double error, double delta, uint32_t& width, uint32_t& depth);
```
mapleFU: Would you mind encapsulate `(width, depth)` to a struct, and just return the struct?

src/types/cms.h
```diff
@@ -0,0 +1,82 @@
...
+
```
mapleFU, line 46: `const char* item, size_t item_len` -> can this be change to std::string_view

src/types/cms.h
```diff
@@ -0,0 +1,82 @@
...
+  int CMSMergeParams(const MergeParams& params);
+
+  uint64_t& GetCounter() { return counter_; }
+  std::vector<uint32_t>& GetArray() { return array_; }
+
+  const uint64_t& GetCounter() const { return counter_; }
+  const std::vector<uint32_t>& GetArray() const { return array_; }
+
+  size_t GetWidth() const { return width_; }
+  size_t GetDepth() const { return depth_; }
```
mapleFU: Why do they shift between u32 and size_t?

src/types/cms.h
```diff
@@ -0,0 +1,82 @@
...
+    CMSketch* dest;
+    size_t num_keys;
+    std::vector<const CMSketch*> cms_array;
+    std::vector<long long> weights;
+  };
+
+  int CMSMergeParams(const MergeParams& params);
+
+  uint64_t& GetCounter() { return counter_; }
+  std::vector<uint32_t>& GetArray() { return array_; }
```
mapleFU, line 63: Would a set array being better here?

mapleFU (commented):
I didn't go through the detail, just some style nits firstly

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+  LOG(INFO) << error << delta;
+  width = std::ceil(2 / error);
+  depth = std::ceil(std::log10(delta) / std::log10(0.5));
+}
+
+size_t CMSketch::IncrBy(const char* item, size_t item_len, size_t value) {
+  size_t min_count = std::numeric_limits<size_t>::max();
+
+  for (size_t i = 0; i < depth_; ++i) {
+    uint32_t hash = HllMurMurHash64A(item, static_cast<int>(item_len), i);
```
mapleFU: I don't know getting u32 from u64 hash a good idea...

jonathanc-n: @mapleFU  tyty for the review, i'll be revising it tmr.

src/commands/commander.h
jonathanc-n: I think CMS is fine, CountMinSketch would be too long. HyperLogLog is also written as HLL here, so it is consistent.

jonathanc-n:
Was there a change to GetOptions recently? After the merge, it seems to be behaving different

jonathanc-n:
Was there a change made to GetOptions recently? There seems to be a problem after the merge.

mapleFU:
> Was there a change made to GetOptions recently? There seems to be a problem after the merge.

Yes, the operations to db requires an options here.

src/types/cms.h
jonathanc-n: No, for a cms, it is storing counters and not unique objects.

jonathanc-n:
I believe this pr should be good for further review

mapleFU:
Nice, I'll take a round tonight

src/commands/cmd_cms.cc
```diff
@@ -0,0 +1,168 @@
...
+#include "server/server.h"
+
+namespace redis {
+
+/// CMS.INCRBY key item increment [item increment ...]
+class CommandCMSIncrBy final : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    if ((args_.size() - 2) % 2 != 0) {
+      return Status::RedisTryAgain;
```
mapleFU: ?

src/commands/cmd_cms.cc
```diff
@@ -0,0 +1,168 @@
...
+    }
+    redis::CMS cms(srv->storage, conn->GetNamespace());
+    engine::Context ctx(srv->storage);
+    rocksdb::Status s;
+    std::unordered_map<std::string, uint64_t> elements;
+    for (size_t i = 2; i < args_.size(); i += 2) {
+      std::string key = args_[i];
+      uint64_t value = 0;
+      try {
+        value = std::stoull(args_[i + 1]);
```
mapleFU: we have `ParseInt` in `parse_util.h`?

src/commands/cmd_cms.cc
```diff
@@ -0,0 +1,168 @@
...
+};
+
+/// CMS.INFO key
+class CommandCMSInfo final : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    redis::CMS cms(srv->storage, conn->GetNamespace());
+    engine::Context ctx(srv->storage);
+    rocksdb::Status s;
+    std::unordered_map<std::string, uint64_t> elements;
```
mapleFU: never used?

src/commands/cmd_cms.cc
```diff
@@ -0,0 +1,168 @@
...
+    if (s.IsNotFound()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    if (!s.ok() && !s.IsNotFound()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    *output = redis::Array({redis::BulkString("width"), redis::Integer(ret[0]), redis::BulkString("depth"),
+                            redis::Integer(ret[1]), redis::BulkString("count"), redis::Integer(ret[2])});
```
mapleFU: Can you use a struct for `ret` rather than a `vector`?

src/commands/cmd_cms.cc
```diff
@@ -0,0 +1,168 @@
...
+
+/// CMS.INITBYDIM key width depth
+class CommandCMSInitByDim final : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    redis::CMS cms(srv->storage, conn->GetNamespace());
+    engine::Context ctx(srv->storage);
+    rocksdb::Status s;
+    uint64_t width = std::stoull(args_[2]);
+    uint64_t depth = std::stoull(args_[3]);
```
mapleFU: ParseInt

src/commands/cmd_cms.cc
```diff
@@ -0,0 +1,168 @@
...
+
+/// CMS.INITBYPROB key error probability
+class CommandCMSInitByProb final : public Commander {
+ public:
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    redis::CMS cms(srv->storage, conn->GetNamespace());
+    engine::Context ctx(srv->storage);
+    rocksdb::Status s;
+    double error = std::stod(args_[2]);
+    double delta = std::stod(args_[3]);
```
mapleFU: ParseFloat

src/storage/redis_metadata.h
```diff
@@ -335,3 +336,15 @@ class HyperLogLogMetadata : public Metadata {
 
   EncodeType encode_type = EncodeType::DENSE;
 };
+
+class CountMinSketchMetadata : public Metadata {
+ public:
+  uint32_t width;
+  uint32_t depth;
+  uint64_t counter;
```
mapleFU: Should this have default value?

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+size_t CMSketch::IncrBy(std::string_view item, size_t value) {
+  size_t min_count = std::numeric_limits<size_t>::max();
+
+  for (size_t i = 0; i < depth_; ++i) {
+    uint64_t hash = HllMurMurHash64A(item.data(), static_cast<int>(item.size()), i);
+    size_t loc = (hash % width_) + (i * width_);
+    array_[loc] += value;
+    if (array_[loc] < value) {
+      array_[loc] = UINT32_MAX;
+    }
```
mapleFU: can this changed to check overflow?

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+
+
+CMSketch::CMSketchDimensions CMSketch::CMSDimFromProb(double error, double delta) {
+  CMSketchDimensions dims;
+  dims.width = std::ceil(2 / error);
+  dims.depth = std::ceil(std::log10(delta) / std::log10(0.5));
+  return dims;
+}
+
+size_t CMSketch::IncrBy(std::string_view item, size_t value) {
```
mapleFU: why value is a size_t here?

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+#include <algorithm>
+#include <cmath>
+#include <cstdint>
+#include <vector>
+
+
+CMSketch::CMSketchDimensions CMSketch::CMSDimFromProb(double error, double delta) {
+  CMSketchDimensions dims;
+  dims.width = std::ceil(2 / error);
+  dims.depth = std::ceil(std::log10(delta) / std::log10(0.5));
```
mapleFU, line 37: Should we also try to adding a verify to check the width and depth not too large or too small ?

src/types/redis_cms.cc
```diff
@@ -0,0 +1,167 @@
...
+rocksdb::Status CMS::GetMetadata(engine::Context &ctx, const Slice &ns_key,
+                                 CountMinSketchMetadata *metadata) {
+  return Database::GetMetadata(ctx, {kRedisCountMinSketch}, ns_key, metadata);
+}
+
+rocksdb::Status CMS::IncrBy(engine::Context &ctx, const Slice &user_key, const std::unordered_map<std::string, uint64_t> &elements) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+  CountMinSketchMetadata metadata{};
```
mapleFU, line 39: If not exists, does this mean getting a `<0, 0, 0>` pair? What does it means?

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+  return dims;
+}
+
+size_t CMSketch::IncrBy(std::string_view item, size_t value) {
+  size_t min_count = std::numeric_limits<size_t>::max();
+
+  for (size_t i = 0; i < depth_; ++i) {
+    uint64_t hash = HllMurMurHash64A(item.data(), static_cast<int>(item.size()), i);
+    size_t loc = (hash % width_) + (i * width_);
```
mapleFU: can you extract the loc to a helper function?

mapleFU (commented):
I would take a look at redis sketch later

src/commands/cmd_cms.cc
jonathanc-n: should I make this part with a status parse instead of putting it in execute?

src/types/redis_cms.cc
jonathanc-n: Yeah, you would get a 0,0,0 pair

jonathanc-n:
@mapleFU do the changes seem right? sorry for the bother.

src/commands/cmd_cms.cc
mapleFU: When arg mismatch, try again never solving this, it should report input is invalid

Besides, the arg-parsing framework can well handle this?

jonathanc-n: @mapleFU alright should be good!

jonathanc-n:
@PragmaTwice Is the workflow able to run again?

mapleFU:
> do the changes seem right? sorry for the bother.

Sorry for delaying, I'll take a pass

src/storage/redis_metadata.h
```diff
@@ -335,3 +336,15 @@ class HyperLogLogMetadata : public Metadata {
 
   EncodeType encode_type = EncodeType::DENSE;
 };
+
+class CountMinSketchMetadata : public Metadata {
+ public:
+  uint32_t width;
+  uint32_t depth;
+  uint64_t counter = 0;
+  std::vector<uint32_t> array;
```
mapleFU, line 345: @git-hulk @PragmaTwice 

I'm thinking about handling the `array`. Since regarding the cmsketch as a string is ok to me, but parsing it to an `std::vector` is a bit weird to me. Can we take a `Buffer` and do zero copying when doing this?

Like:
```
LoadMetadata from string. When storing, forcing LittleEndian in the array
When reading from metadata, report error if not length enough, and hold a sliced buffer on underlying data
When read/write, using memcpy to avoid unaligned access.
```

mapleFU:
@jonathanc-n would you mind fix the lint?

src/storage/redis_metadata.cc
```diff
@@ -329,7 +329,8 @@ bool Metadata::ExpireAt(uint64_t expired_ts) const {
 bool Metadata::IsSingleKVType() const { return Type() == kRedisString || Type() == kRedisJson; }
 
 bool Metadata::IsEmptyableType() const {
-  return IsSingleKVType() || Type() == kRedisStream || Type() == kRedisBloomFilter || Type() == kRedisHyperLogLog;
+  return IsSingleKVType() || Type() == kRedisStream || Type() == kRedisBloomFilter || Type() == kRedisHyperLogLog ||
```
mapleFU: Wait, isn't a CMSketch a Single-KV Type?

src/types/cms.cc
mapleFU: Why did this being solved? Would you mind point out the code resolve this problem?

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+      return -1;
+    }
+  }
+
+  if (cms_count < 0) {
+    return -1;
+  }
+
+  return 0;
+}
```
mapleFU: fmt: `\n` after this

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+  dims.width = std::ceil(2 / error);
+  dims.depth = std::ceil(std::log10(delta) / std::log10(0.5));
+  return dims;
+}
+
+size_t CMSketch::IncrBy(std::string_view item, uint32_t value) {
+  size_t min_count = std::numeric_limits<size_t>::max();
+
+  for (size_t i = 0; i < depth_; ++i) {
+    uint64_t hash = HllMurMurHash64A(item.data(), static_cast<int>(item.size()), i);
```
mapleFU: Please, a uint64_t as hash might overflows uint32_t, making `array_` likely to randomly set uint32 max

mapleFU: What about using xxhash32?

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+  counter_ += value;
+  return min_count;
+}
+
+size_t CMSketch::Query(std::string_view item) const {
+  size_t min_count = std::numeric_limits<size_t>::max();
+
+  for (size_t i = 0; i < depth_; ++i) {
+    uint64_t hash = HllMurMurHash64A(item.data(), static_cast<int>(item.size()), i);
+    min_count = std::min(min_count, static_cast<size_t>(array_[(hash % width_) + (i * width_)]));
```
mapleFU: Why didn't `GetLocation` used here?

src/types/cms.h
```diff
@@ -0,0 +1,89 @@
...
+ *
+ */
+
+#pragma once
+
+#include <memory>
+#include <vector>
+
+#include "server/redis_reply.h"
+#include "vendor/murmurhash2.h"
```
mapleFU: why murmurhash2 included here?

src/types/cms.h
```diff
@@ -0,0 +1,89 @@
...
+
+class CMSketch {
+ public:
+  explicit CMSketch(uint32_t width, uint32_t depth, uint64_t counter, std::vector<uint32_t> array = {})
+      : width_(width),
+        depth_(depth),
+        counter_(counter),
+        array_(array.empty() ? std::vector<uint32_t>(width * depth, 0) : std::move(array)) {}
+
+  static std::unique_ptr<CMSketch> NewCMSketch(uint32_t width, int32_t depth) {
```
mapleFU: Ever used?

mapleFU:
I'm outing these days, so maybe late reply

src/types/cms.cc
mapleFU: Let me change the wording here, can we assure:
1. the cmsketch memory used less than 50MB
2.the cmsketch memory used greater than 0

src/storage/redis_metadata.cc
jonathanc-n: Oh, sorry about that, i was thinking due to the different variables, that it would be instead put under emptyable type. But after review it makes more sense as a singlekvtype

jonathanc-n:
@git-hulk @PragmaTwice I think a workflow can be run again with the changes

jonathanc-n:
@PragmaTwice Would you be able to review this pr?

src/types/redis_cms.h
```diff
@@ -0,0 +1,50 @@
...
+  rocksdb::Status InitByDim(engine::Context &ctx, const Slice &user_key, uint32_t width, uint32_t depth);
+  rocksdb::Status InitByProb(engine::Context &ctx, const Slice &user_key, double error, double delta);
+  rocksdb::Status Query(engine::Context &ctx, const Slice &user_key, const std::vector<std::string> &elements,
+                        std::vector<uint32_t> &counters);
+
+ private:
+  [[nodiscard]] rocksdb::Status GetMetadata(engine::Context &ctx, const Slice &ns_key,
+                                            CountMinSketchMetadata *metadata);
+
+  // TODO (jonathanc-n)
```
mapleFU: todo what?

jonathanc-n: @mapleFU the mergeuserkeys function

mapleFU:
```suggestion
  // TODO (jonathanc-n)
  // [[nodiscard]] rocksdb::Status mergeUserKeys(engine::Context &ctx, const std::vector<Slice> &user_keys,
  //                                            std::vector<std::string> *register_segments);  
```

Can you do like this?

src/types/cms.h
```diff
@@ -0,0 +1,85 @@
...
+ *
+ */
+
+#pragma once
+
+#include <memory>
+#include <vector>
+
+#include "server/redis_reply.h"
+#include "xxhash.h"
```
mapleFU: I didn't see xxhash being used in header...

src/types/cms.h
```diff
@@ -0,0 +1,85 @@
...
+
+class CMSketch {
+ public:
+  explicit CMSketch(uint32_t width, uint32_t depth, uint64_t counter, std::vector<uint32_t> array = {})
```
mapleFU: Please not use default argument, which might being misused when we program later

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+    min_count = std::min(min_count, static_cast<size_t>(array_[GetLocation(hash, i)]));
+  }
+  return min_count;
+}
+
+int CMSketch::Merge(CMSketch* dest, size_t quantity, const std::vector<const CMSketch*>& src,
+                    const std::vector<long long>& weights) {
+  if (checkOverflow(dest, quantity, src, weights) != 0) {
+    return -1;
+  }
```
mapleFU: Should we use `StatusOr<>` rather than `-1` code here?

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+
+  for (size_t i = 0; i < dest->GetDepth(); ++i) {
+    for (size_t j = 0; j < dest->GetWidth(); ++j) {
+      int64_t item_count = 0;
+      for (size_t k = 0; k < quantity; ++k) {
+        item_count += static_cast<int64_t>(src[k]->array_[(i * dest->GetWidth()) + j]) * weights[k];
```
mapleFU: `GetLocation * weights[k]`?

src/types/cms.cc
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  return 0;
+}
+
+int CMSMergeParams(const CMSketch::MergeParams& params) {
+  return CMSketch::Merge(params.dest, params.num_keys, params.cms_array, params.weights);
+}
+
+int CMSketch::checkOverflow(CMSketch* dest, size_t quantity, const std::vector<const CMSketch*>& src,
```
mapleFU: Also using `StatusOr` here?

mapleFU (commented):
General LGTM
Would you mind also add the go test, for:
1. do operations on non-created
2. Basic query
3. Merge
?

src/types/cms.cc
jonathanc-n: No, i believe they dont do the same

src/types/cms.cc
jonathanc-n: I would rather avoid using statusor on the implementation layer, dealing with future errors gets difficult to track.

src/storage/redis_metadata.h
PragmaTwice: Yeah it can be like how we handle the JSON data structure.

PragmaTwice: But here I'm more interested in whether it's better to be a single key or putting the array in subkeys.

jonathanc-n:
@mapleFU something seems to be a little odd here, I ran clang-format on all changed files and its running into formatting issues

tests/gocase/unit/cms/cms_test.go
```diff
@@ -0,0 +1,245 @@
...
+		res := rdb.Do(ctx, "cms.incrby", "nonexistent_cms", "apple", "5")
+		require.Error(t, res.Err())
+	})
+
+	// Query for non-existent element
+	t.Run("cms.query - Query Non-Existent Element", func(t *testing.T) {
+		require.NoError(t, rdb.Do(ctx, "DEL", "cmsA").Err())
+		res := rdb.Do(ctx, "cms.initbydim", "cmsA", "100", "10")
+		require.NoError(t, res.Err())
+		require.Equal(t, "OK", res.Val())
```
mapleFU, line 176: Can we also query a non-exists key like "cms.incrby - Increment Non-Existent CMS"

tests/gocase/unit/cms/cms_test.go
```diff
@@ -0,0 +1,245 @@
...
+		require.NoError(t, res.Err())
+		countSlice, ok := res.Val().([]interface{})
+		require.True(t, ok, "Expected cms.query to return a slice")
+		require.Len(t, countSlice, 1, "Expected cms.query to return a single count")
+		count, ok := countSlice[0].(string)
+		require.True(t, ok, "Expected count to be a string")
+		require.Equal(t, "0", count, "Non-existent element should return count '0'")
+	})
+
+	// Merging CMS structures
```
mapleFU, line 189: Can we also merge on unexists key? Includling uninitialized dest or sources

mapleFU (commented):
@jonathanc-n would you mind me add some comments and edit it directly after test is added?

jonathanc-n:
@mapleFU Test cases are implemented! Ty for the test suggestions.

jonathanc-n:
@mapleFU Does this seem fine to move forward?

src/commands/cmd_cms.cc
```diff
@@ -0,0 +1,266 @@
...
+      if (!parse_result) {
+        return {Status::RedisParseErr, errValueNotInteger};
+      }
+      uint64_t value = *parse_result;
+      elements[key] = value;
+    }
+
+    s = cms.IncrBy(ctx, args_[1], elements);
+    if (s.IsNotFound()) {
+      return {Status::RedisExecErr, "Key not found"};
```
mapleFU, line 64: Not Found handling is different from `CMS.INFO`?

src/commands/cmd_cms.cc
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    *output = redis::SimpleString("OK");
+    return Status::OK();
+  }
+
+ private:
+  std::string destination_;  // Changed to std::string
+  int num_keys_;
+  std::vector<std::string> src_keys_;  // Changed to std::vector<std::string>
```
mapleFU: Actually I don't understanding why change to std::string...

src/types/redis_cms.h
```diff
@@ -0,0 +1,48 @@
...
+ public:
+  explicit CMS(engine::Storage *storage, const std::string &ns) : Database(storage, ns) {}
+
+  rocksdb::Status IncrBy(engine::Context &ctx, const Slice &user_key,
+                         const std::unordered_map<std::string, uint64_t> &elements);
+  rocksdb::Status Info(engine::Context &ctx, const Slice &user_key, CMSketch::CMSInfo *ret);
+  rocksdb::Status InitByDim(engine::Context &ctx, const Slice &user_key, uint32_t width, uint32_t depth);
+  rocksdb::Status InitByProb(engine::Context &ctx, const Slice &user_key, double error, double delta);
+  rocksdb::Status Query(engine::Context &ctx, const Slice &user_key, const std::vector<std::string> &elements,
+                        std::vector<uint32_t> &counters);
```
mapleFU, line 50:
```suggestion
                        std::vector<uint32_t> *counters);
```

src/types/redis_cms.cc
```diff
@@ -0,0 +1,264 @@
...
+  rocksdb::Status s = GetMetadata(ctx, ns_key, &metadata);
+
+  if (s.IsNotFound()) {
+    return rocksdb::Status::NotFound();
+  }
+  if (!s.ok() && !s.IsNotFound()) {
```
mapleFU: Why `!s.IsNotFound()` here?

src/types/redis_cms.cc
```diff
@@ -0,0 +1,264 @@
...
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisCountMinSketch);
+  batch->PutLogData(log_data.Encode());
+
+  CMSketch cms(metadata.width, metadata.depth, metadata.counter, metadata.array);
+
+  if (elements.empty()) {
+    return rocksdb::Status::OK();
+  }
```
mapleFU: I don't understand why it would be empty: https://github.com/RedisBloom/RedisBloom/blob/b5e674f61e84f9636b26d2005fe5ea50230ecb79/src/rm_cms.c#L120

src/types/redis_cms.cc
```diff
@@ -0,0 +1,264 @@
...
+
+  for (auto &element : elements) {
+    cms.IncrBy(element.first, element.second);
```
mapleFU: We don't check "INCRBY overflow"?

src/types/redis_cms.cc
```diff
@@ -0,0 +1,264 @@
...
+}
+
+rocksdb::Status CMS::Info(engine::Context &ctx, const Slice &user_key, CMSketch::CMSInfo *ret) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+  CountMinSketchMetadata metadata{};
+  rocksdb::Status s = GetMetadata(ctx, ns_key, &metadata);
+
+  if (!s.ok() || s.IsNotFound()) {
```
mapleFU: I don't understand why `!s.ok()` can be cast to `NotFound`

src/types/redis_cms.cc
```diff
@@ -0,0 +1,264 @@
...
+rocksdb::Status CMS::InitByDim(engine::Context &ctx, const Slice &user_key, uint32_t width, uint32_t depth) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+  CountMinSketchMetadata metadata{};
+
+  rocksdb::Status s = GetMetadata(ctx, ns_key, &metadata);
+
+  if (!s.IsNotFound()) {
+    return s;
```
mapleFU, line 114: Shouldn't we return "key already exists" when `s.ok()`. If we follow the current way, it will return ok when init.

src/types/redis_cms.cc
```diff
@@ -0,0 +1,264 @@
...
+  return storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+};
+
+rocksdb::Status CMS::InitByProb(engine::Context &ctx, const Slice &user_key, double error, double delta) {
+  if (error <= 0 || error >= 1) {
+    return rocksdb::Status::InvalidArgument("Error must be between 0 and 1 (exclusive).");
+  }
+  if (delta <= 0 || delta >= 1) {
+    return rocksdb::Status::InvalidArgument("Delta must be between 0 and 1 (exclusive).");
+  }
```
mapleFU, line 139: Can we count a `(dim, ...)` then call `InitByDim`?

src/types/redis_cms.cc
```diff
@@ -0,0 +1,264 @@
...
+
+  std::string encoded_metadata;
+  dest_metadata.Encode(&encoded_metadata);
+  batch->Put(metadata_cf_handle_, dest_ns_key, encoded_metadata);
+
+  return storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+rocksdb::Status CMS::Query(engine::Context &ctx, const Slice &user_key, const std::vector<std::string> &elements,
+                           std::vector<uint32_t> &counters) {
```
mapleFU, line 224: Should we clear the counters firstly?

src/types/cms.cc
mapleFU: Why? So you think integers are better to tracking than Explicit Status code?

mapleFU (commented):
I found some unsolved issue marked as "solved", I spend some time on tracking them...

src/storage/redis_metadata.h
mapleFU: > But here I'm more interested in whether it's better to be a single key or putting the array in subkeys.

Actually I think this in single key is ok since we merely "only query metadata" if not calling "info"

src/types/cms.h
```diff
@@ -0,0 +1,84 @@
...
+                   const std::vector<long long>& weights);
+
+  struct MergeParams {
+    CMSketch* dest;
+    size_t num_keys;
+    std::vector<const CMSketch*> cms_array;
+    std::vector<long long> weights;
+  };
+
+  int CMSMergeParams(const MergeParams& params);
```
mapleFU: Ever used?

src/types/redis_cms.cc
```diff
@@ -0,0 +1,264 @@
...
+  std::vector<CMSketch> src_cms_objects;
+  src_cms_objects.reserve(num_sources);
+  std::vector<const CMSketch *> src_cms_pointers;
+  src_cms_pointers.reserve(num_sources);
+  std::vector<long long> weights_long;
+  weights_long.reserve(num_sources);
+
+  for (size_t i = 0; i < num_sources; ++i) {
+    std::string src_ns_key = AppendNamespacePrefix(src_keys[i]);
+    LOG(INFO) << "Dest Key: " << dest_ns_key << " | Source Key: " << src_ns_key;
```
mapleFU: Is this neccessary for INFO?

src/types/redis_cms.cc
```diff
@@ -0,0 +1,264 @@
...
+    return dest_status;
+  }
+
+  CMSketch dest_cms(dest_metadata.width, dest_metadata.depth, dest_metadata.counter, dest_metadata.array);
+
+  std::vector<CMSketch> src_cms_objects;
+  src_cms_objects.reserve(num_sources);
+  std::vector<const CMSketch *> src_cms_pointers;
+  src_cms_pointers.reserve(num_sources);
+  std::vector<long long> weights_long;
```
mapleFU: Should we uses `int64_t` for this?

src/types/cms.cc
mapleFU: You're right, this avoid calculation the hash location here. I'm thinking that would `GetLocation -> GetLocationForHash` would being a less confusing naming

src/types/cms.cc
jonathanc-n: Oh sorry, I wasn't aware of what Status was able to do.

jonathanc-n:
@mapleFU Alright, thanks for the reviews, I put the changes in for it.

src/types/cms.cc
```diff
@@ -0,0 +1,130 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#include "cms.h"
+
+#include <xxhash.h>
```
mapleFU: wonder isn't xxhash include with `"`?

mapleFU (commented):
Rest LGTM. I'll update some comment before approve these two days

src/types/redis_cms.cc
```diff
@@ -0,0 +1,251 @@
...
+
+rocksdb::Status CMS::Info(engine::Context &ctx, const Slice &user_key, CMSketch::CMSInfo *ret) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+  CountMinSketchMetadata metadata{};
+  rocksdb::Status s = GetMetadata(ctx, ns_key, &metadata);
+
+  if (!s.ok()) {
+    return rocksdb::Status::NotFound();
```
mapleFU: ditto, !ok doesn't means NotFound

jonathanc-n:
@mapleFU Srorry about that, fixed the lint.

mapleFU:
No problem, we're almost their

src/storage/redis_metadata.h
PragmaTwice: What's the limit of the size of such an array? Do we need to consider to split it?

src/types/redis_cms.cc
```diff
@@ -0,0 +1,252 @@
...
+    return rocksdb::Status::InvalidArgument("Delta must be between 0 and 1 (exclusive).");
+  }
+  CMSketch::CMSketchDimensions dim = CMSketch::CMSDimFromProb(error, delta);
+  auto s = InitByDim(ctx, user_key, dim.width, dim.depth);
+  if (!s.ok()) {
+    return s;
+  }
+
+  size_t memory_used = dim.width * dim.depth * sizeof(uint32_t);
+  const size_t max_memory = 50 * 1024 * 1024;
```
mapleFU: Actually I think this should be put in `InitByDim` before allocating the memory or put to storage

src/storage/redis_metadata.h
mapleFU: https://github.com/apache/kvrocks/pull/2524/files#r1779978212

@PragmaTwice The problem is that, the stored value would always be fetched together when required. It's merely to "partition querying" / "updating" since it's a prob structure

We can limit it to 50M or less?

src/types/cms.h
```diff
@@ -0,0 +1,81 @@
...
+  uint32_t GetDepth() const { return depth_; }
+
+ private:
+  uint32_t width_;
+  uint32_t depth_;
+  uint64_t counter_;
+  std::vector<uint32_t> array_;
+
+  static int checkOverflow(CMSketch* dest, size_t quantity, const std::vector<const CMSketch*>& src,
+                           const std::vector<long long>& weights);
```
PragmaTwice: Why don't we use `this` and make it a member function here? instead of `dest`.

src/types/cms.h
```diff
@@ -0,0 +1,81 @@
...
+  size_t Query(std::string_view item) const;
+
+  struct MergeParams {
+    CMSketch* dest;
+    size_t num_keys;
+    std::vector<const CMSketch*> cms_array;
+    std::vector<long long> weights;
+  };
+
+  static Status Merge(const MergeParams& params);
```
PragmaTwice: Maybe we can just use these fields of MergeParams as parameters? i.e. `Merge(size_t num_keys, ...)`.

mapleFU:
I'm a bit tired today and will revisit this tomorrow

src/storage/redis_metadata.h
PragmaTwice: I don't know if it's good to have one rocksdb key with a 50M-bytes value.

I checked the code and seems we'll access `d` elements (once for every row of `w` elements) in the `d*w` array.
Then I think maybe it's good to keep it in one key if the total size is small enough, e.g. 1M.

And if it's larger, we can move them into `w/k` key-values (one column as one key), and calculate how many keys we need to retrieve.

mapleFU: > I don't know if it's good to have one rocksdb key with a 50M-bytes value.

The default engine doesn't handle this well. There're some blog engine which could do this, and we can enable the blob in kvrocks here.

Perhaps we can limit the size to 1MB firstly? 🤔

PragmaTwice: Looks good to me.

Also yeah we should not put the array inside the metadata.

mapleFU: I may try a round tonight

src/types/cms.h
```diff
@@ -0,0 +1,82 @@
...
+  size_t IncrBy(std::string_view item, uint32_t value);
+
+  size_t Query(std::string_view item) const;
+
+  struct MergeParams {
+    CMSketch* dest;
+    size_t num_keys;
+    std::vector<const CMSketch*> cms_array;
+    std::vector<uint32_t> weights;
+  };
```
PragmaTwice: seems useless now

src/storage/redis_metadata.h
jonathanc-n: I'll adjust it to 1MB limit

jonathanc-n: So do we want to hold the array in a buffer on storage, and do operations on that?

src/types/redis_cms.cc
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  size_t memory_used = dim.width * dim.depth * sizeof(uint32_t);
+  const size_t max_memory = 1 * 1024 * 1024;
+
+  if (memory_used == 0) {
+    return rocksdb::Status::InvalidArgument("Memory usage must be greater than 0.");
+  }
+  if (memory_used > max_memory) {
+    return rocksdb::Status::InvalidArgument("Memory usage exceeds 50MB.");
```
PragmaTwice:
```suggestion
    return rocksdb::Status::InvalidArgument("Memory usage exceeds 1MB.");
```

src/types/redis_cms.cc
```diff
@@ -0,0 +1,256 @@
...
+  std::string ns_key = AppendNamespacePrefix(user_key);
+
+  size_t memory_used = width * depth * sizeof(uint32_t);
+  const size_t max_memory = 1 * 1024 * 1024;
+
+  if (memory_used == 0) {
+    return rocksdb::Status::InvalidArgument("Memory usage must be greater than 0.");
+  }
+  if (memory_used > max_memory) {
+    return rocksdb::Status::InvalidArgument("Memory usage exceeds 50MB.");
```
PragmaTwice:
```suggestion
    return rocksdb::Status::InvalidArgument("Memory usage exceeds 1MB.");
```

mapleFU:
I just reach home and have some minor updates, would continue tomorrow. Sorry for delaying

src/types/redis_cms.cc
```diff
@@ -204,11 +191,15 @@ rocksdb::Status CMS::MergeUserKeys(engine::Context &ctx, const Slice &user_key,
 
     CMSketch src_cms(src_metadata.width, src_metadata.depth, src_metadata.counter, src_metadata.array);
     src_cms_objects.emplace_back(std::move(src_cms));
-    src_cms_pointers.push_back(&src_cms_objects.back());
 
     weights_long.push_back(static_cast<uint32_t>(src_weights[i]));
   }
-
+  // Initialize the destination CMS with the source CMSes after initializations
+  // since vector might resize and reallocate memory.
```
mapleFU, line 199: Previously this might cause memory issue

src/types/redis_cms.cc
```diff
@@ -165,33 +151,34 @@ rocksdb::Status CMS::MergeUserKeys(engine::Context &ctx, const Slice &user_key,
   }
 
   std::string dest_ns_key = AppendNamespacePrefix(user_key);
-  LockGuard guard(storage_->GetLockManager(), dest_ns_key);
```
mapleFU: This might cause deadlock

jonathanc-n:
@mapleFU Were you gonna make your own changes to the code?

mapleFU:
Yes I'm back from holiday and working, would continue.

PR #2530, opened 2024-09-09
test: add go test cases for `txn-context-enabled` config

### LINKED ISSUE #2514: Add go test cases for `txn_context_enabled` config
### Search before asking

### Motivation

Make sure the `txn_context_enabled` config is available

### Solution

_No response_

PokIsemaine:
fix: https://github.com/apache/kvrocks/issues/2514

I added a config test of `txn-context-enabled` to the not particularly time-consuming go test cases

Also, I've noticed that some of the tests have `resp3-enabled` related tests, which require multiple layers of nesting and branching if combined with `txn-context-enabled`.

To alleviate this, I tried adding `util. GenerateConfigsMatrix` to generate possible config combinations, which you learn about by using `TestGenerateConfigsMatrix` in `config_test.go`. At the moment, only `YesNo` configs are supported, but I think this will provide some convenience for config-related testing, and we can support more types later if needed

### DIFF
--- a/tests/gocase/util/configs.go
+++ b/tests/gocase/util/configs.go
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package util
+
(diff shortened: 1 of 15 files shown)


tests/gocase/util/configs.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		return fmt.Errorf("invalid option for yes/no config")
+	default:
+		return fmt.Errorf("unsupported config type")
+	}
+	return nil
+}
+
+// / GenerateConfigsMatrix generates all possible combinations of config options
+func GenerateConfigsMatrix(configOptions []ConfigOptions) ([]KvrocksServerConfigs, error) {
```
mapleFU, line 52: So this only return false when options is error?

tests/gocase/util/configs.go
```diff
@@ -0,0 +1,88 @@
...
+	helper = func(configs []ConfigOptions, currentIndex int, currentConfig map[string]string) error {
+		if currentIndex == len(configOptions) {
+			configsMatrix = append(configsMatrix, currentConfig)
+			return nil
+		}
+
+		currentConfigBackup := make(KvrocksServerConfigs, len(currentConfig))
+		for k, v := range currentConfig {
+			currentConfigBackup[k] = v
+		}
```
mapleFU, line 66: Copying this might be heavy weight but lgtm since it's just testing...

mapleFU (approved):

mapleFU:
Retrigger CI, would merge if all ci passed

git-hulk (approved):

Issue #2478, opened 2024-08-08
Add more tests for stream consumer group


Yangsx-1:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

We should add more test cases for stream consumer group, especially for the time related information such as idle time and inactive time, and other information in XINFO command.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

jonathanc-n:
@PragmaTwice @git-hulk What is the syntax for isolating go tests for testing?

jonathanc-n:
@Yangsx-1 Are there any tests you would like to add in particular? Checking certain functionality?

git-hulk:
> @PragmaTwice @git-hulk What is the syntax for isolating go tests for testing?

What do you mean by isolating go tests? Are you asking how to run the specified test cases?

jonathanc-n:
@git-hulk Yes, specified test cases.

Yangsx-1:
> @Yangsx-1 Are there any tests you would like to add in particular? Checking certain functionality?

I think we should pay attention to time related variables, such as idel time and inactive time in XINFO CONSUMERS, and some counter variables such as pending number, last_delivered_id, entries_read, lag and so on.

Yangsx-1:
> @git-hulk Yes, specified test cases.

You can run command like this: `./x.py test go ./build -run ^TestStreamOffset$ github.com/apache/incubator-kvrocks/tests/gocase/unit/type/stream -v`, and you can change `TestStreamOffset` when you want to test other go functions.

git-hulk:
> @git-hulk Yes, specified test cases.

You can run with `-test.run`. For example, I want to run the test case `TestFullSyncReplication`, then you should enter the dir `tests/gocase/integration/replication` and run `go test -test.run "TestFullSyncReplication" -workspace /tmp -binPath {Your Kvrocks Binary Path}`.

Remember to replace `{Your Kvrocks Binary Path}` with your owner Kvrocks binary path.

jonathanc-n:
@Yangsx-1 @git-hulk Thank you guys

PR #2533, opened 2024-09-14
chore(tests): add basic tests for the stream consumer group 

jonathanc-n:
references issue #2478 

@Yangsx-1 I am creating these two test cases, just to make sure that these are in the direction that you wanted. I will more test cases later.

test 1 - Focuses on idle time and pending messages

test 2 - Focuses on consumer existence, inactive time, and verifies that consumers will be properly removed from the group


jonathanc-n:
Workflow should be good I believe

git-hulk (approved):

PragmaTwice (approved):

Issue #2518, opened 2024-09-02
The DUMP command does not support the BITMAP type


VIVALXH:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

DUMP command is supported from version 2.9, but not support the BITMAP TYPE, will it be supported in the future?

Doc：
https://kvrocks.apache.org/docs/supported-commands/#server-commands

Partial code ：
https://github.com/apache/kvrocks/blob/0329170d60d67a5f0f62ee565a1a14f0bbfb81da/src/storage/rdb.cc#L719

Example：
![image](https://github.com/user-attachments/assets/ba496eac-4047-4ddb-be83-096da0678561)

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #2535, opened 2024-09-14
feat: Bitmap type DUMP/RESTORE command support

poipoiPIO:
Changes requested by #2518 

## Summary
This PR extends dump and restore operations to provide support for bitmap-type objects using existing string object serialisation and deserialisation routines.

### DIFF
--- a/src/storage/rdb.cc
+++ b/src/storage/rdb.cc
@@ -789,8 +789,8 @@ Status RDB::SaveObject(const std::string &key, const RedisType type) {
   } else if (type == kRedisBitmap) {
     std::string value;
     redis::Bitmap bitmap_db(storage_, ns_);
-    Config* config = storage_->GetConfig();
-    uint32_t max_btos_size = static_cast<uint32_t>(config->max_bitmap_to_string_mb) * MiB; 
+    Config *config = storage_->GetConfig();
+    uint32_t max_btos_size = static_cast<uint32_t>(config->max_bitmap_to_string_mb) * MiB;
     auto s = bitmap_db.GetString(ctx, key, max_btos_size, &value);
     if (!s.ok() && !s.IsNotFound()) {
       return {Status::RedisExecErr, s.ToString()};
--- a/src/storage/rdb.cc
+++ b/src/storage/rdb.cc
@@ -405,7 +405,7 @@ StatusOr<int> RDB::loadRdbType() {
 }
 
 StatusOr<RedisObjValue> RDB::loadRdbObject(int type, [[maybe_unused]] const std::string &key) {
-  if (type == RDBTypeString || type == RDBTypeBitmap) {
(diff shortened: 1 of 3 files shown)


git-hulk:
@poipoiPIO thanks for your contribution.

src/storage/rdb.cc
```diff
@@ -402,7 +405,7 @@ StatusOr<int> RDB::loadRdbType() {
 }
 
 StatusOr<RedisObjValue> RDB::loadRdbObject(int type, [[maybe_unused]] const std::string &key) {
-  if (type == RDBTypeString) {
+  if (type == RDBTypeString || type == RDBTypeBitmap) {
```
git-hulk: The bitmap type is an internal represent type in Kvrocks, and it should be regarded as a string type. I think we can remove `RDBTypeBitmap`? also, the number `6` is now used by [RDB_TYPE_MODULE_PRE_GA](https://github.com/redis/redis/blob/unstable/src/rdb.h#L60C9-L60C31), reuse this number might cause the unexpected error while restoring rdb from Redis.

src/storage/rdb.cc
```diff
@@ -728,6 +731,8 @@ Status RDB::SaveObjectType(const RedisType type) {
     robj_type = RDBTypeSet;
   } else if (type == kRedisZSet) {
     robj_type = RDBTypeZSet2;
+  } else if (type == kRedisBitmap) {
+    robj_type = RDBTypeBitmap;
```
git-hulk: As mentioned in the above comment, we can convert the robj_type to RDBTypeString here.

src/storage/rdb.cc
poipoiPIO: Oh I see, thank you for the suggestion! Will do!

git-hulk (approved):

torwig (approved):

git-hulk:
@poipoiPIO Thank you.

poipoiPIO:
> @poipoiPIO Thank you.

Thank you for reviewing!

src/storage/rdb.cc
```diff
@@ -781,6 +784,16 @@ Status RDB::SaveObject(const std::string &key, const RedisType type) {
     }
 
     return SaveHashObject(field_values);
+  } else if (type == kRedisBitmap) {
+    std::string value;
+    redis::Bitmap bitmap_db(storage_, ns_);
+    Config *config = storage_->GetConfig();
+    uint32_t max_btos_size = static_cast<uint32_t>(config->max_bitmap_to_string_mb) * MiB;
+    auto s = bitmap_db.GetString(ctx, key, max_btos_size, &value);
```
mapleFU, line 792: @git-hulk would this meet the same issue like lsb to msb?

Issue #2401, opened 2024-07-11
Confusing error message while failed to parse the cluster slots 


spihiker:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

v2.9.0

### Minimal reproduce step

 CLUSTERX SETNODES "86f7e437faa5a7fce15d1ddcb9eaeaea377667b8 172.20.84.154 6666 master - 0-5460 3c363836cf4e16666669a25da280a1865c2d2874 172.20.84.154 6669 slave 86f7e437faa5a7fce15d1ddcb9eaeaea377667b8" 1
(error) ERR Slot is out of range

### What did you expect to see?

normal as redis.

### What did you see instead?

no 

### Anything Else?

no

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PragmaTwice:
Could you try to add a `\n`?

```
CLUSTERX SETNODES "86f7e437faa5a7fce15d1ddcb9eaeaea377667b8 172.20.84.154 6666 master - 0-5460 \n  3c363836cf4e16666669a25da280a1865c2d2874 172.20.84.154 6669 slave 86f7e437faa5a7fce15d1ddcb9eaeaea377667b8" 1
```

spihiker:
After add a \n.It can run normally.Thank you.

git-hulk:
@PragmaTwice @spihiker This error message is very confusing for users, I have changed the issue title to track and improve it.

jonathanc-n:
@PragmaTwice @git-hulk Would this be more of a parser change vs. a changed error messsage? Is using '\n' the proper way to do it?

jonathanc-n:
This can be closed

git-hulk:
@jonathanc-n Thank you!

PR #2538, opened 2024-09-14
fix: add more check in cluster node parsing

jonathanc-n:
references #2401 

Added non confusing message when multiple nodes are added without a newline separator

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -716,21 +716,6 @@ Status Cluster::parseClusterNodes(const std::string &nodes_str, ClusterNodes *no
     return {Status::ClusterInvalidInfo, errInvalidClusterNodeInfo};
   }
 
-  if (nodes_info.size() == 1) {
-    std::regex node_id_regex(R"(\b[a-fA-F0-9]{40}\b)");
-    auto begin = std::sregex_iterator(nodes_info[0].begin(), nodes_info[0].end(), node_id_regex);
-    auto end = std::sregex_iterator();
-    size_t node_id_count = std::distance(begin, end);
-
-    if (node_id_count > 1) {
-      return {Status::ClusterInvalidInfo, "Invalid nodes definition: Missing newline between node entries."};
-    }
-
-    if (nodes_info[0].find(' ') == std::string::npos) {
-      return {Status::ClusterInvalidInfo, "Invalid nodes definition: Node entry is improperly formatted."};
-    }
-  }
-
   for (const auto &entry : nodes_info) {
     if (entry.find(' ') == std::string::npos) {
       return {Status::ClusterInvalidInfo,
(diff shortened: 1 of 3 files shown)


PragmaTwice (commented):
Could you add a test case for this patch?

jonathanc-n:
Test case is added + the node checking is more flexible

jonathanc-n:
@PragmaTwice Are you able to see what exactly is causing the fail on the error, I don't seem to be able to see the reason for it.

git-hulk:
@jonathanc-n #2401 is caused by missing a newline between nodes, and omitting the newline at the end should be allowed.

jonathanc-n:
@git-hulk Alright, it should be fully working now.

git-hulk:
@jonathanc-n You can enhance this inside [parseClusterNodes](https://github.com/apache/kvrocks/blob/2a0c57ac7fc4499a210bc452884025a0010cb470/src/cluster/cluster.cc#L712)

jonathanc-n:
@PragmaTwice @git-hulk This should be good I believe, I moved the changes to cluster.cc.

git-hulk:
@jonathanc-n It would be better to check if the field is node id in line 778? https://github.com/apache/kvrocks/blob/2a0c57ac7fc4499a210bc452884025a0010cb470/src/cluster/cluster.cc#L778

The current solution is too tricky from my perspective.

jonathanc-n:
@git-hulk I simplified as much as I thought I could for the issue to not persist. I believe it should be fine.

jonathanc-n:
@git-hulk @PragmaTwice Pretty sure this should be good, it passed the workflow. Unless you see some changes that should be made.

src/cluster/cluster.cc
```diff
@@ -716,6 +716,13 @@ Status Cluster::parseClusterNodes(const std::string &nodes_str, ClusterNodes *no
     return {Status::ClusterInvalidInfo, errInvalidClusterNodeInfo};
   }
 
+  for (const auto &entry : nodes_info) {
+    if (entry.find(' ') == std::string::npos) {
+      return {Status::ClusterInvalidInfo,
+              "Invalid nodes definition: Each node entry must contain fields separated by spaces."};
+    }
+  }
+
```
```suggestion
```

It has been checked at Line 730.

src/cluster/cluster.cc
```diff
@@ -776,6 +783,13 @@ Status Cluster::parseClusterNodes(const std::string &nodes_str, ClusterNodes *no
     for (unsigned i = 5; i < fields.size(); i++) {
       std::vector<std::string> ranges = util::Split(fields[i], "-");
       if (ranges.size() == 1) {
+        std::regex node_id_regex(R"(\b[a-fA-F0-9]{40}\b)");
```
git-hulk: It should be good to only check if `fields[i]` is a node id?

jonathanc-n:
@git-hulk Alright, should be fine now, sorry about that.

git-hulk:
@jonathanc-n No worry, thanks for your great contribution.

git-hulk (approved):

Issue #2534, opened 2024-09-14
Support for Cuckoo filter data structure and commands


jonathanc-n:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

One of the remaining data structures that are in redis that are not implemented in Kvrocks. 

Functionality: https://redis.io/docs/latest/develop/data-types/probabilistic/cuckoo-filter/ 
Commands: https://redis.io/docs/latest/commands/?group=cf 

Tracking
- [ ] CF.ADD command
- [ ] CF.ADDNX command
- [ ] CF.COUNT command
- [ ] CF.DEL command
- [ ] CF.EXISTS command
- [ ] CF.INFO command
- [ ] CF.INSERT command
- [ ] CF.INSERTNX command
- [ ] CF.LOADCHUNK command
- [ ] CF.MEXISTS command
- [ ] CF.RESERVE command 
- [ ] CF.SCANDUMP command

### Solution

I will create a design proposal through issue/discussion before submitting a PR.

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

geonove:
@jonathanc-n Are you still working on this?

jonathanc-n:
@geonove No I am not, feel free to work on it!

LiuQhahah:
Hi All,
I want to claim this task
Thanks,
Qiang

git-hulk:
@LiuQhahah Assigned, thank you!

PragmaTwice:
Could I ask if @geonove is working on it or not?

geonove:
You can take it, I am working on something else at the moment. I'll come back for some other command to implement :)

nagisa-kunhah:
@jihuayu Hi, for the `CF.EXISTS` command, I asked about taking this part a few months ago in #3359, and @LiuQhahah said I could go ahead.

Would it be okay if I continue and implement this part based on the `LiuQhahah/cuckoo-filter` branch?

jihuayu:
@nagisa-kunhah All right. Go ahead!

PR #2545, opened 2024-09-18
feat: support cuckoo filter and most commands

jonathanc-n:
references issue: #2534 

Added support for all commands except for SCANDUMP and LOADCHUNK.

The go cases are currently being worked on.

### DIFF
--- a/src/storage/redis_metadata.cc
+++ b/src/storage/redis_metadata.cc
@@ -329,7 +329,7 @@ bool Metadata::ExpireAt(uint64_t expired_ts) const {
 bool Metadata::IsSingleKVType() const { return Type() == kRedisString || Type() == kRedisJson; }
 
 bool Metadata::IsEmptyableType() const {
-  return IsSingleKVType() || Type() == kRedisStream || Type() == kRedisBloomFilter || Type() == kRedisHyperLogLog;
+  return IsSingleKVType() || Type() == kRedisStream || Type() == kRedisBloomFilter || Type() == kRedisHyperLogLog || Type() == kRedisCuckooFilter;
 }
 
 bool Metadata::Expired() const { return ExpireAt(util::GetTimeStampMS()); }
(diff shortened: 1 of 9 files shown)


torwig:
Running `./x.py format` should make the linter happy.

src/types/redis_cuckoo.cc
```diff
@@ -0,0 +1,487 @@
...
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+  CuckooFilterMetadata metadata{};
+  rocksdb::Status s = GetMetadata(ctx, ns_key, &metadata);
+  if (!s.ok() && !s.IsNotFound()) {
+    return s;
+  }
+
+  if (s.IsNotFound()) {
+    metadata.capacity = 1024;
```
torwig: Perhaps, you can extract this logic into a separate function with a descriptive name (e.g. `createDefaultCuckooFilterMetadata` or any other you find appropriate). I see you repeat this piece of code so it's an ideal candidate for a separate function.

src/types/redis_cuckoo.cc
```diff
@@ -0,0 +1,487 @@
...
+    initial_filter.num_buckets = metadata.num_buckets;
+    initial_filter.data.resize(metadata.num_buckets * metadata.bucket_size, CUCKOO_NULLFP);
+    metadata.filters.push_back(initial_filter);
+  }
+
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisCuckooFilter);
+  batch->PutLogData(log_data.Encode());
+
+  CuckooFilter cf(metadata.capacity, metadata.bucket_size, metadata.max_iterations, metadata.expansion,
```
torwig, line 97: Maybe you can pass just `metadata` instead of 9 fields as parameters?

torwig:
@jonathanc-n Thank you for this contribution. Great job!

PragmaTwice:
Thanks for your contribution!

However, I cannot find a proposal to state the design details and storage format of the new data structure.

May I ask you provide such a proposal before we continue to review this patch?

jonathanc-n:
@PragmaTwice it is mentioned in #2539, I added some extra information to clarify some things.

src/types/redis_cuckoo.cc
jonathanc-n: @torwig I did try to get pass this problem, but it introduces many type dependencies and errors  if redis metadata is included in the cuckoo.h file where the cf is initialized (SubCF is used here). Passing 9 parameters doesn't look good, but I believe it would be better than involving a lot of dependencies between files.

src/types/cuckoo.cc
```diff
@@ -0,0 +1,396 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+/*
+This project uses Redis, which is licensed under your choice of the
+Redis Source Available License 2.0 (RSALv2) or the Server Side Public License v1 (SSPLv1).
```
PragmaTwice, line 20: As an ASF project, we CANNOT use any code under these non-open-source licenses, which includes RSAL and SSPL.

src/storage/redis_metadata.cc
```diff
@@ -495,3 +496,71 @@ rocksdb::Status HyperLogLogMetadata::Decode(Slice *input) {
...
+  PutFixed16(dst, max_iterations);
+  PutFixed32(dst, num_buckets);
+  PutFixed16(dst, expansion);
+  PutFixed16(dst, num_filters);
+  PutFixed64(dst, num_items);
+  PutFixed64(dst, num_deletes);
+  for (const auto &filter : filters) {
+    PutFixed16(dst, filter.bucket_size);
+    PutFixed64(dst, filter.num_buckets);
+    dst->append(filter.data.begin(), filter.data.end());
```
PragmaTwice, line 513: I doubt the efficiency of this approach: everytime we read/update the filters, we need to retrieve/put all of these filters.

Please refer to the design of bloom filters to see if you can put filters to multiple key values.

jonathanc-n:
@PragmaTwice Sorry, about this implementation, are you able to close this pr. I might come back to it with a completely different code implementation.

PragmaTwice:
Sure.

geonove:
@jonathanc-n Are you still working on this? If not can I have a try at it?

PR #2547, opened 2024-09-20
fix(rdb): incorrect type conversion during RDB loads the int8 encoding

### LINKED ISSUE #2546: Inconsistent behavior when using RESTORE with dumped negative integers
### Search before asking

### Version

2.9.0

### Minimal reproduce step

```
127.0.0.1:6666> set k -1
OK
127.0.0.1:6666> get k
"-1"
127.0.0.1:6666> dump k
"\x00\xc0\xff\x06\x00\x9a~\x16\xa4p]\xa1V"
127.0.0.1:6666> restore k1 0 "\x00\xc0\xff\x06\x00\x9a~\x16\xa4p]\xa1V"
OK
127.0.0.1:6666> get k1
"255"
127.0.0.1:6666> get k
"-1"
127.0.0.1:6666> set k2 -2
OK
127.0.0.1:6666> get k2
"-2"
127.0.0.1:6666> dump k2
"\x00\xc0\xfe\x06\x00$FVjY\x0c\xb3\xa2"
127.0.0.1:6666> restore k22 0 "\x00\xc0\xfe\x06\x00$FVjY\x0c\xb3\xa2"
OK
127.0.0.1:6666> get k2
"-2"
127.0.0.1:6666> get k22
"254"
```

 The value remains negative after RESTORE.

### What did you see instead?

when restoring dumped negative integers, the values are restored as positive numbers (255 for -1 and 254 for -2). 

### Anything Else?

Possibly a problem with the following code?

https://github.com/apache/kvrocks/blob/2a0c57ac7fc4499a210bc452884025a0010cb470/src/storage/rdb.cc#L158-L160

fstd2:
Resolve https://github.com/apache/kvrocks/issues/2546

### DIFF
--- a/src/storage/rdb.cc
+++ b/src/storage/rdb.cc
@@ -157,7 +157,7 @@ StatusOr<std::string> RDB::loadEncodedString() {
     unsigned char buf[4] = {0};
     if (len == RDBEncInt8) {
       auto next = GET_OR_RET(stream_->ReadByte());
-      return std::to_string(static_cast<int>(next));
+      return std::to_string(static_cast<int8_t>(next));
     } else if (len == RDBEncInt16) {
       GET_OR_RET(stream_->Read(reinterpret_cast<char *>(buf), 2));
       auto value = static_cast<uint16_t>(buf[0]) | (static_cast<uint16_t>(buf[1]) << 8);
--- a/tests/gocase/unit/dump/dump_test.go
+++ b/tests/gocase/unit/dump/dump_test.go
@@ -167,3 +167,29 @@ func TestDump_Bitset(t *testing.T) {
 	require.NoError(t, rdb.RestoreReplace(ctx, restoredKey, 0, serialized).Err())
 	require.Equal(t, rdb.Get(ctx, key).Val(), rdb.Get(ctx, restoredKey).Val())
 }
+
+func TestDump_IntegerEncoding(t *testing.T) {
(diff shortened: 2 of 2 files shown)


src/storage/rdb.cc
```diff
@@ -157,7 +157,7 @@ StatusOr<std::string> RDB::loadEncodedString() {
     unsigned char buf[4] = {0};
     if (len == RDBEncInt8) {
       auto next = GET_OR_RET(stream_->ReadByte());
-      return std::to_string(static_cast<int>(next));
+      return std::to_string(static_cast<int8_t>(next));
```
torwig, line 160: Why `int` if the length is 8?

torwig: Did you try `uint8_t`? However, I'm not sure that it's the right type.

git-hulk: It should be correct to use int8_t here since the encoding is int8. Cast it to `int` will drop the sign bit.

git-hulk:
@fstd2 It would be great if you could add a Go test case to prevent breaking in the future.

fstd2:
> @fstd2 It would be great if you could add a Go test case to prevent breaking in the future.

I'm a beginner, don't know how to do..

git-hulk:
@fstd2 No worry, I will do this later if you don't know how to. Thank you!

git-hulk (approved):

PragmaTwice (approved):

PR #2548, opened 2024-09-20
feat: Add the ability to checkpoint an existing server, and spawn a read-only server on that view.

nathanlo99:
It would be nice to be able to spawn a read-only server on a snapshot of a given server.

This PR implements this feature, and introduces the flag `--snapshot-dir` to manage the snapshot.


src/cli/main.cc
```diff
@@ -46,6 +48,8 @@
 #include "version_util.h"
 
 Server *srv = nullptr;
+std::optional<std::string> checkpoint_location = std::nullopt;
+bool keep_checkpoint = false;
```
PragmaTwice: We can move them to CLIOptions rather than global variables.

src/cli/main.cc
```diff
@@ -116,6 +128,32 @@ static void InitGoogleLog(const Config *config) {
...
 
+static void CreateCheckpoint(Config &config, const std::string &checkpoint_location) {
+  // The Storage destructor deletes anything at the checkpoint_dir, so we need to make sure it's empty in case the user
+  // happens to use a checkpoint folder name which matches the default (checkpoint/)
+  const std::string old_checkpoint_dir = std::exchange(config.checkpoint_dir, "");
+  const auto checkpoint_dir_guard =
+      MakeScopeExit([&config, &old_checkpoint_dir] { config.checkpoint_dir = old_checkpoint_dir; });
+
+  engine::Storage storage(&config);
+  if (const auto s = storage.Open(kDBOpenModeForReadOnly); !s.IsOK()) {
```
git-hulk: Can return the error as a Status instead of an exception.

nathanlo99:
```
--- FAIL: TestBitmap (846.21s)
    --- FAIL: TestBitmap/SETBIT/GETBIT/BITCOUNT/BITPOS_boundary_check_(type_bitmap) (150.90s)
        bitmap_test.go:204: 
            	Error Trace:	/Users/runner/work/kvrocks/kvrocks/tests/gocase/unit/type/bitmap/bitmap_test.go:204
            	            	read tcp 127.0.0.1:50435->127.0.0.1:50429: i/o timeout
            	Test:       	TestBitmap/SETBIT/GETBIT/BITCOUNT/BITPOS_boundary_check_(type_bitmap)
```

The above check failure _seems_ to be an unrelated i/o timeout; is there a way to rerun the checks?

git-hulk:
> ```
> --- FAIL: TestBitmap (846.21s)
>     --- FAIL: TestBitmap/SETBIT/GETBIT/BITCOUNT/BITPOS_boundary_check_(type_bitmap) (150.90s)
>         bitmap_test.go:204: 
>             	Error Trace:	/Users/runner/work/kvrocks/kvrocks/tests/gocase/unit/type/bitmap/bitmap_test.go:204
>             	Error:      	Received unexpected error:
>             	            	read tcp 127.0.0.1:50435->127.0.0.1:50429: i/o timeout
>             	Test:       	TestBitmap/SETBIT/GETBIT/BITCOUNT/BITPOS_boundary_check_(type_bitmap)
> ```
> 
> The above check failure _seems_ to be an unrelated i/o timeout; is there a way to rerun the checks?

Yes, it's a flaky test and not related to this PR.

nathanlo-hrt:
@PragmaTwice @git-hulk could you take another look at this? It seems the previous opportunity to merge was missed because new changes made it to the unstable branch.

src/config/config.cc
```diff
@@ -159,6 +159,8 @@ Config::Config() {
        new IntField(&force_compact_file_min_deleted_percentage, 10, 1, 100)},
       {"db-name", true, new StringField(&db_name, "change.me.db")},
       {"dir", true, new StringField(&dir, kDefaultDir)},
+      {"snapshot-dir", false, new StringField(&snapshot_dir, "")},
+      {"keep-snapshot", false, new YesNoField(&keep_snapshot, false)},
```
PragmaTwice: We can add them into `kvrocks.conf`, and add some comments for these new options.

src/cli/main.cc
```diff
@@ -170,8 +197,27 @@ int main(int argc, char *argv[]) {
...
+      LOG(INFO) << "Removing snapshot dir: " << config.snapshot_dir;
+      std::filesystem::remove_all(config.snapshot_dir);
+    }
+  });
+
   engine::Storage storage(&config);
-  s = storage.Open();
+  const bool read_only = use_snapshot;
+  const DBOpenMode open_mode = read_only ? kDBOpenModeForReadOnly : kDBOpenModeDefault;
+  s = storage.Open(open_mode);
```
PragmaTwice: Since now they are just normal options, I think we'd better to move these logic to config.cc or storage.cc, just like other options, to make it more maintainable.

WDYT?

PragmaTwice: Also the `CreateSnapshot` function, it need to be moved out of `cli/main.cc` since its funtionality is out of CLI scope.

nathanlo99: Thanks, this makes sense: this is done now

src/config/config.cc
```diff
@@ -190,6 +188,7 @@ Config::Config() {
        new IntField(&force_compact_file_min_deleted_percentage, 10, 1, 100)},
       {"db-name", true, new StringField(&db_name, "change.me.db")},
       {"dir", true, new StringField(&dir, kDefaultDir)},
+      {"snapshot-dir", false, new StringField(&snapshot_dir, "")},
```
```suggestion
      {"snapshot-dir", true, new StringField(&snapshot_dir, "")},
```
I think it should be read-only.

src/storage/storage.cc
```diff
@@ -75,12 +76,46 @@ const int64_t kIORateLimitMaxMb = 1024000;
...
+
 Storage::Storage(Config *config)
     : backup_creating_time_secs_(util::GetTimeStamp<std::chrono::seconds>()),
       env_(rocksdb::Env::Default()),
       config_(config),
       lock_mgr_(16),
       db_stats_(std::make_unique<DBStats>()) {
+  if (config->snapshot_dir != "") {
+    if (const auto s = CreateSnapshot(*config, config->snapshot_dir); !s.IsOK()) {
+      throw std::runtime_error(fmt::format("Failed to create snapshot: {}", s.Msg()));
```
PragmaTwice: Hmm I think we need to move them to Storage::Open, also the throw can be replaced then.

src/storage/storage.cc
```diff
@@ -75,12 +76,46 @@ const int64_t kIORateLimitMaxMb = 1024000;
 
 using rocksdb::Slice;
 
+static Status CreateSnapshot(Config &config, const std::string &snapshot_location) {
```
mapleFU: Actually I don't really understand how this works, seems this create a checkpoint when opening the file? Do you need create it on server and consume it on cli?

mapleFU:
Would you mind also add example for how you'd like use this? ( better on doc of cli or doc of project?)

src/storage/storage.cc
nathanlo99: > Do you need create it on server and consume it on cli?

I'm not sure i understand the second part of your question

> Actually I don't really understand how this works, seems this create a checkpoint when opening the file?

As for how this works, it will read the data from the `dir` passed by the user, then use RocksDB snapshot functionality to make a read-only copy of that data into the `snapshot-dir` directory, then spawn a read-only server on that copy.

This is helpful for my team to debug a running kvrocks app without accidentally changing the values of the data or closing the existing read-write server.

src/storage/storage.cc
nathanlo99: Done; had to do a bit more state management to avoid infinite recursion, since `CreateSnapshot` also calls `Storage::Open()`

nathanlo99:
apologies for the linter back and forth: i'm unable to build this locally with the same set-up as the linter uses, so a lot of these issues fall through the cracks

src/cli/main.cc
```diff
@@ -171,7 +171,8 @@ int main(int argc, char *argv[]) {
 #endif
 
   engine::Storage storage(&config);
-  s = storage.Open();
+  s = storage.Open(kDBOpenModeDefault);
```
PragmaTwice: Hmm I think we don't need to change it?

nathanlo99:
@mapleFU I'm not sure how to edit the docs, but one example workflow is the following.

Start a read-write Kvrocks server writing data to `~/kvrocks-dir`, for the purposes of writing important data.
```
$ kvrocks [...] --dir ~/kvrocks-dir
```

Say the user realizes some of the important data is incorrect: they'd like to debug their script and run it on a copy of the data, possibly changing it. They'd like to do this without fear of corrupting or changing the original server's copy of the data:
```
$ mkdir ~/debug-dir
$ kvrocks [...] --dir ~/kvrocks-dir --snapshot-dir ~/debug-dir
```

4player90 (approved):

PR #2553, opened 2024-09-23
fix: format changes

jonathanc-n:
I ran ./x.py format and ran into some small changes. They looked fine to push, and I left out config format changes.


jonathanc-n:
Never mind, this pr seems to messing with the original Lint/Check code process. Maybe I can open an issue to check why these changed lines from ./x.py format cause an error when clang-format is run. Anyone have thoughts on this?

PragmaTwice:
To avoid such inconsistency, please use clang-format 14 which is the same version as in CI pipeline.

Closed.

PragmaTwice:
To avoid such inconsistency, please use clang-format 14 which is the same version as in CI pipeline.

Closed.

Issue #2418, opened 2024-07-15
Add key expiration support to KQIR


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Currently, KQIR cannot handle expired keys or remove them from the index after they expire.

We need to add this support.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

jonathanc-n:
Can I be assigned this?

jonathanc-n:
@PragmaTwice What would be most ideal for handling expiring keys, lazy expiration, active expiration, etc.?

jonathanc-n:
I think I will go with a lazy expiration strategy

PR #2556, opened 2024-09-25
feat: Added key expiration support

jonathanc-n:
references: #2418 

Added key expiration support tag, and numeric data for KQIR.


jonathanc-n:
@PragmaTwice Are you able to see if this solution is logically sound, I am struggling to create a test case right now.

jonathanc-n:
@PragmaTwice One thing noted for this pr, is that it is not fully removing the key from index from being searched. After deletion is called, the key and data is deleted however when calling the search function again, the key is still looked for and will call a notfound error

jonathanc-n:
@PragmaTwice Is there a particular way you would want key expiration to be handled?

PR #2564, opened 2024-09-30
feat(search): support the FT.TAGVALS command

jonathanc-n:
Added FT.TagVals for search + tests


src/search/index_manager.h
```diff
@@ -272,6 +273,62 @@ struct IndexManager {
 
     return Status::OK();
   }
-};
 
+  Status FieldValues(engine::Context &ctx, std::string_view index_name, std::string_view input_tag,
```
PragmaTwice: We should iterate through the index data, instead of redis keys.

jonathanc-n: @PragmaTwice Put in the fixes! thanks for the feedback :)

src/search/index_manager.h
```diff
@@ -272,6 +273,58 @@ struct IndexManager {
...
+
+    std::unordered_set<std::string> matching_values;
+    util::UniqueIterator index_iter(ctx, ctx.DefaultScanOptions(), ColumnFamilyID::Search);
+
+    auto index_key = SearchKey(ns, index_name, field_name);
+    std::string field_prefix;
+    index_key.PutNamespace(&field_prefix);
+    SearchKey::PutType(&field_prefix, SearchSubkeyType::FIELD);
+    index_key.PutIndex(&field_prefix);
+    PutSizedString(&field_prefix, field_name);
```
PragmaTwice, line 304: You can use `ConstructTagFieldData({}, {})` here.

src/search/index_manager.h
```diff
@@ -272,6 +273,58 @@ struct IndexManager {
 
     return Status::OK();
   }
-};
 
+  Status FieldValues(engine::Context &ctx, std::string_view index_name, std::string_view input_tag,
```
PragmaTwice:
```suggestion
  Status FieldValues(engine::Context &ctx, std::string_view index_name, std::string_view tag_field_name,
```

src/search/index_manager.h
```diff
@@ -272,6 +273,58 @@ struct IndexManager {
...
+      auto key = index_iter->key();
+
+      if (!key.starts_with(field_prefix)) {
+        break;
+      }
+
+      Slice key_slice = key;
+      key_slice.remove_prefix(field_prefix.size());
+
+      Slice field_value_slice;
```
PragmaTwice:
```suggestion
      Slice tag;
```

src/search/index_manager.h
```diff
@@ -272,6 +273,58 @@ struct IndexManager {
...
+      if (!GetSizedString(&key_slice, &field_value_slice)) continue;
+
+      matching_values.insert(field_value_slice.ToString());
```
PragmaTwice: Here we can make an optimization: if the tag is still the last one, we can skip it instead of insert it.

jonathanc-n: I have now implemented it, let me know if its fine

src/search/index_manager.h
```diff
@@ -272,6 +273,67 @@ struct IndexManager {
 
     return Status::OK();
   }
-};
 
+  Status FieldValues(engine::Context &ctx, std::string_view index_name, std::string_view input_tag,
```
PragmaTwice:
```suggestion
  StatusOr<std::unordered_set<std::string>> FieldValues(engine::Context &ctx, std::string_view index_name, std::string_view input_tag,
```

jonathanc-n: I'm a little confused, how come this is the implementation here, however in other parts of Kvrocks, it uses a ret value? When do I know when to use Status + *ret value compared to using StatusOr<T>?

PragmaTwice: It's preferred to use `StatusOr` instead of `Status` and pointer output.

src/commands/cmd_search.cc
```diff
@@ -488,6 +504,7 @@ REDIS_REGISTER_COMMANDS(Search,
                         MakeCmdAttr<CommandFTExplain>("ft.explain", -3, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandFTInfo>("ft.info", 2, "read-only", 0, 0, 0),
                         MakeCmdAttr<CommandFTList>("ft._list", 1, "read-only", 0, 0, 0),
-                        MakeCmdAttr<CommandFTDrop>("ft.dropindex", 2, "write exclusive no-multi no-script", 0, 0, 0));
+                        MakeCmdAttr<CommandFTDrop>("ft.dropindex", 2, "write exclusive no-multi no-script", 0, 0, 0),
+                        MakeCmdAttr<CommandFTTagVals>("ft.tagvals", 3, "read-only", 0, 0, 0));
```
PragmaTwice:
```suggestion
                        MakeCmdAttr<CommandFTTagVals>("ft.tagvals", 3, "read-only slow", 0, 0, 0));
```

PragmaTwice (commented):
The rest looks good to me.

src/search/index_manager.h
```diff
@@ -272,6 +273,66 @@ struct IndexManager {
 
     return Status::OK();
   }
-};
 
+  StatusOr<std::unordered_set<std::string>> FieldValues(engine::Context &ctx, std::string_view index_name,
+                                                        std::string_view input_tag, const std::string &ns) {
```
PragmaTwice:
```suggestion
                                                        std::string_view tag_field_name, const std::string &ns) {
```

jonathanc-n:
@PragmaTwice Commited to change. Should be good to run workflow

PragmaTwice:
Could you check the CI failure to see if it can be fixed?

jonathanc-n:
@PragmaTwice CI, should be fixed. It was just incorrect return values for type mismatch.

PragmaTwice (approved):

PR #2566, opened 2024-09-30
fix(stream): Fix XPENDING serialization issue

nathanlo-hrt:
Currently `XPENDING` doesn't properly create and serialize an array of the right length: the response to an `XPENDING` command is always an array of length 4, regardless of the number of consumer infos it has to send back.

The existing tests pass because they all happened to have exactly one consumer.


PragmaTwice (dismissed):

torwig:
Stream test time out after 30 minutes. Maybe there is a block or infinite wait (for example, waiting for a message that is never added).

torwig:
@nathanlo-hrt Thank you for this catch.

PragmaTwice:
> Stream test time out after 30 minutes. Maybe there is a block or infinite wait (for example, waiting for a message that is never added).

So.. is this another bug of XPENDING? Or we can just change the test case to fix it?

nathanlo-hrt:
I believe this is a bug in my changes to the test code; I'll put up a fix later today. Thanks for the swift reviews!

tests/gocase/unit/type/stream/stream_test.go
```diff
@@ -2083,7 +2086,7 @@ func TestStreamOffset(t *testing.T) {
 			Count:     1,
 			Lower:     "1-0",
 			Higher:    "1-0",
-			Consumers: map[string]int64{"myconsumer": 1},
+			Consumers: map[string]int64{consumerName: 1},
```
torwig, line 2119: Can we extract the `consumerName` to a variable like `streamName` and `groupName`?

torwig (dismissed):

nathanlo-hrt:
Found another issue while testing: Kvrocks currently deviates from Redis' behavior when there are no pending messages. I've added a special case for this and a test to cover the new code.

torwig:
@nathanlo-hrt Good job! You can run `./x.py format` to format the code and make the linter happy.

torwig:
```
Error: unit/type/stream/stream_test.go:2170:6: ineffectual assignment to err (ineffassign)
		r, err := rdb.XAck(ctx, streamName, groupName, "0-0").Result()
```

tests/gocase/unit/type/stream/stream_test.go
```diff
@@ -2125,9 +2128,78 @@ func TestStreamOffset(t *testing.T) {
...
+
+	t.Run("XPENDING on a consumer group with no pending messages", func(t *testing.T) {
+		streamName := "stream"
+		groupName := "group"
+		consumerName := "consumer"
+		messageID := "1-0"
+		
+		// Remove any existing data
+		require.NoError(t, rdb.Del(ctx, streamName).Err())
+		r, err := rdb.XAck(ctx, streamName, groupName, "0-0").Result()
```
torwig, line 2200: Check this error.

torwig (commented):
LGTM with minor comment.

PragmaTwice (approved):

Issue #2568, opened 2024-10-01
XINFO GROUPS fails to deduplicate; returns invalid `pending` 


nathanlo-hrt:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

unstable commit b8cf118bbe56984c7112c0929402f79d81d8a0d7

### Minimal reproduce step

```
127.0.0.1:3847> FLUSHALL
OK
127.0.0.1:3847> XGROUP CREATE stream1 group 0 MKSTREAM
OK
127.0.0.1:3847> XGROUP CREATE stream2 group 0 MKSTREAM
OK
127.0.0.1:3847> XREADGROUP GROUP group consumer STREAMS stream1 stream2 0 0
(nil)
127.0.0.1:3847> XINFO GROUPS stream2
1)  1) "name"
    2) "group"
    3) "consumers"
    4) (integer) 1
    5) "pending"
    6) (integer) 0
    7) "last-delivered-id"
    8) "0-0"
    9) "entries-read"
   10) (nil)
   11) "lag"
   12) (integer) 0
2)  1) "name"
    2) "group"
    3) "consumers"
    4) (integer) 0
    5) "pending"
    6) (integer) 1727818077688
    7) "last-delivered-id"
    8) "1727818077688-0"
    9) "entries-read"
   10) (integer) 0
   11) "lag"
   12) (integer) 0
```

### What did you expect to see?

The following output comes from Redis:
```
127.0.0.1:6379> FLUSHALL
OK
127.0.0.1:6379> XGROUP CREATE stream1 group 0 MKSTREAM
OK
127.0.0.1:6379> XGROUP CREATE stream2 group 0 MKSTREAM
OK
127.0.0.1:6379> XREADGROUP GROUP group consumer STREAMS stream1 stream2 0 0
1) 1) "stream1"
   2) (empty array)
2) 1) "stream2"
   2) (empty array)
127.0.0.1:6379> XINFO GROUPS stream2
1) 1) "name"
   2) "group"
   3) "consumers"
   4) (integer) 1
   5) "pending"
   6) (integer) 0
   7) "last-delivered-id"
   8) "0-0"
```

There are a few notable differences here:
1. `XREADGROUP GROUP` returns `stream1` and `stream2` as empty arrays instead of returning `nil` for the entire request
2. `XINFO GROUPS stream2` doesn't return a duplicated consumer group
[...]

PR #2569, opened 2024-10-02
fix(stream): change XREADGROUP response for empty read response

jonathanc-n:
references #2568 

Changed XReadGroup response for empty responses to match redis behaviour.

### DIFF
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -31,7 +31,6 @@
 #include "status.h"
 #include "time_util.h"
 #include "types/redis_stream.h"
-#include "types/redis_string.h"
 
 namespace redis {
 
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -31,6 +31,7 @@
 #include "status.h"
 #include "time_util.h"
 #include "types/redis_stream.h"
+#include "types/redis_string.h"
 
 namespace redis {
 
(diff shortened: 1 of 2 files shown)


torwig:
@jonathanc-n Should we change existing tests or add extra tests for this case if they don't exist?

jonathanc-n:
@torwig changes in this particular pr are negligible so probably doesn't need a test, but for the other half of the issue I believe there should be a test.

torwig:
@jonathanc-n How to be sure that bug was fixed? Exactly, write a test :)
 
There is a classic flow for fixing a bug:
1. Write a test that fails.
2. Fix a bug.
3. Assert that the test passes -> it will prevent any regression.

jonathanc-n:
I see, thank you for letting me know! :)

jonathanc-n:
@torwig The test is added

torwig (dismissed):

jonathanc-n:
@torwig I made some changes to the stream, hopefully it passes all the checks now

src/commands/cmd_stream.cc
```diff
@@ -31,6 +31,7 @@
 #include "status.h"
 #include "time_util.h"
 #include "types/redis_stream.h"
+#include "types/redis_string.h"
```
PragmaTwice: seems we actually don't need it.

jonathanc-n: sorry bout that, thought i would need it for the bulk string

torwig (approved):

PragmaTwice (approved):

jonathanc-n:
@PragmaTwice Do you know why this might be failing? I am not too sure myself

PragmaTwice:
It keeps failing even after multiple retries.

I'll look into it soon.

Issue #2421, opened 2024-07-15
Support time series data structure and commands like Redis


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

RedisTimeSeries is a redis module used to operate and query time series data, giving redis basic time series database capabilities.

As Apache Kvrocks is characterized by being compatible with the Redis protocol and commands, we also hope to provide temporal data processing capabilities that are compatible with RedisTimeSeries.

This task is to implement the time series data structure and its commands on Kvrocks. Note: Since Kvrocks is an on-disk database based on RocksDB, the implementation will be quite different from Redis.

https://redis.io/docs/data-types/timeseries/
https://kvrocks.apache.org/community/data-structure-on-rocksdb

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

Beihao-Zhou:
Since this was a GSOC project, I drafted a [TS data structure proposal](https://beihao-rebecca.notion.site/TimeSeries-Proposal-6c5d2f9bc0624a50a209cb8f517109a8) before. Hope this can help give a good starting point!

LindaSummer:
Hi @Beihao-Zhou ,

I have read your proposal and have one concern on `Secondary Keys & Values > Label Index`.
The key in proposal is in `[label_key]|[label_value]|[key]` format.
It may need a full iteration scan when purging a key.
Making key as the first prefix part like `[key]|[label_key]|[label_value]` may reduce the complexity on cleanup.

I'm very interested in time series topic and willing to make some contribution in this feature. 😄

Best Regards,
Edward

Beihao-Zhou:
> Hi @Beihao-Zhou ,
> 
> I have read your proposal and have one concern on `Secondary Keys & Values > Label Index`. The key in proposal is in `[label_key]|[label_value]|[key]` format. It may need a full iteration scan when purging a key. Making key as the first prefix part like `[key]|[label_key]|[label_value]` may reduce the complexity on cleanup.
> 
> I'm very interested in time series topic and willing to make some contribution in this feature. 😄
> 
> Best Regards, Edward

The secondary indexing I proposed is mainly for`MRANGE`, which will do some filtering based on label. So for this command, the secondary index is pretty similar to tag field index in KQIR (#2329). If a key is purged, then I think constructing the corresponding secondary index key and deleting it would be enough?

[...]

PR #2573, opened 2024-10-03
feat(ts): initialize metadata encoding for time series

jonathanc-n:
references: #2421

This uses @Beihao-Zhou 's proposal:
https://beihao-rebecca.notion.site/TimeSeries-Proposal-6c5d2f9bc0624a50a209cb8f517109a8

Time Series Main Data (TS.INFO) - in Metadata Column:
flags | expire | version | size | num_samples | total_memory | first_ts | last_ts | retention_time | chunk_count | chunk_size | chunk_type | duplicate_policy (8 bit enum) | source_key
size wasn't used to represent num_samples here because they are inherently different

Primary Subkey (Data point key):
ns | TYPE | key | version | timestamp => value

Secondary Subkey (Label Index Key):
ns | TYPE | label_key | label_value | key => null

Secondary Subkey (Aggregation Rule Key):
ns | TYPE | key | version | dest_key => aggregator | bucket_duration

- The primary sub key were implemented for the regular operations
- The Label Index subkey is for fast search/aggregation commands over labels.
- The aggregation index sub key was implemented for rule related commands
Is to be noted I didn't include the value range index (which was also suggested by beihao) from the proposal. I looked into how much the command was being used on redis, and it wasn't worth the excess storage that would be taken up. 

Let me know if there are any changes or formatting changes that should be done! (I am not confident that the separate file was done properly)

A new search column will be added as it allows for isolated search and aggregation over time series data.


jonathanc-n:
@PragmaTwice @Beihao-Zhou @LindaSummer Feel free to leave changes!

LindaSummer:
Hi @jonathanc-n ,

For improving review efficiency, we could enable our fork repo's github actions.

It will be triggered once we push to our own fork repo and after our own repo's github action passed, we can push our changes to the PR branch.

This can accelerate our own development process and reduce review time. 😊

Best Regards,
Edward

jonathanc-n:
Took out column temporarily as it was failing tests. Will put it back in the full implementation.

Beihao-Zhou:
Thanks! Would it be possible to add details about the encoding and outline the next steps in this PR, similar to how it was done in [this PR](https://github.com/apache/kvrocks/pull/2368)? I think it would help make everything clearer. :)

Beihao-Zhou:
Also can we move the encoding under folder `types` instead of `storage`? Maybe similar to https://github.com/apache/kvrocks/blob/0a43bade1a4c3c885947fdd42eacefac934bf1bd/src/types/redis_stream_base.h

jonathanc-n:
@Beihao-Zhou I could, but I assumed that creating the time_encoding in storage is fine as it needs its own search and querying capabilities over subkeys unlike the stream data.

jonathanc-n:
tests pass

PR #2576, opened 2024-10-04
fix(stream): fix `XINFO` group metadata key overlapping.

### LINKED ISSUE #2568: XINFO GROUPS fails to deduplicate; returns invalid `pending`

LindaSummer:
# Issue

Fix #2568 

# Root cause

Here is the group metadata subkey.
https://github.com/apache/kvrocks/blob/0a43bade1a4c3c885947fdd42eacefac934bf1bd/src/types/redis_stream.cc#L172-L181

Here is the consumer metadata subkey.
https://github.com/apache/kvrocks/blob/0a43bade1a4c3c885947fdd42eacefac934bf1bd/src/types/redis_stream.cc#L222-L233

The enum `StreamConsumerGroupMetadata` is 1 and `StreamConsumerMetadata` is 2.

Here is group info key scanning.
https://github.com/apache/kvrocks/blob/0a43bade1a4c3c885947fdd42eacefac934bf1bd/src/types/redis_stream.cc#L1343-L1348

When search the whole version from `StreamConsumerGroupMetadata`, the `StreamConsumerMetadata` will be included.

Here is an example below.

```
0b5f5f6e616d6573706163650000000773747265616d32311cf46497d771f3ffffffffffffffff01 # lower bound

0b5f5f6e616d6573706163650000000773747265616d32311cf46497d771f3ffffffffffffffff01000000000000000567726f7570 # consumer group metadata
0b5f5f6e616d6573706163650000000773747265616d32311cf46497d771f3ffffffffffffffff02000000000000000567726f75700000000000000008636f6e73756d6572 # consumer metadata


0b5f5f6e616d6573706163650000000773747265616d32311cf46497d771f4ffffffffffffffff01 # upper bound
```

So, we should filter the key we got when searching through the whole version.

By the way, there is another bug.

In encoding, we use `UINT64_MAX` as bytes rather than literal string for `subkey_type_delimiter`.
```
encode: PutFixed64(&sub_key, UINT64_MAX);

search: std::string subkey_type_delimiter = std::to_string(UINT64_MAX);
```

It should also be fixed if I understand correctly.

# Proposed Changes

- Use `identifySubkeyType` to filter metadata key for searching.
- Fix `subkey_type_delimiter` encoding bug.
- Add this issue to unit test.

### DIFF
--- a/src/types/redis_stream.cc
+++ b/src/types/redis_stream.cc
@@ -1340,7 +1340,8 @@ rocksdb::Status Stream::GetGroupInfo(engine::Context &ctx, const Slice &stream_n
   rocksdb::Status s = GetMetadata(ctx, ns_key, &metadata);
   if (!s.ok()) return s;
 
-  std::string subkey_type_delimiter = std::to_string(UINT64_MAX);
+  std::string subkey_type_delimiter;
+  PutFixed64(&subkey_type_delimiter, UINT64_MAX);
   PutFixed8(&subkey_type_delimiter, (uint8_t)StreamSubkeyType::StreamConsumerGroupMetadata);
   std::string next_version_prefix_key =
       InternalKey(ns_key, subkey_type_delimiter, metadata.version + 1, storage_->IsSlotIdEncoded()).Encode();
@@ -1355,6 +1356,10 @@ rocksdb::Status Stream::GetGroupInfo(engine::Context &ctx, const Slice &stream_n
 
   auto iter = util::UniqueIterator(ctx, read_options, stream_cf_handle_);
   for (iter->SeekToFirst(); iter->Valid(); iter->Next()) {
+    if (identifySubkeyType(iter->key()) != StreamSubkeyType::StreamConsumerGroupMetadata) {
+      continue;
+    }
+
     std::string group_name = groupNameFromInternalKey(iter->key());
     StreamConsumerGroupMetadata cg_metadata = decodeStreamConsumerGroupMetadataValue(iter->value().ToString());
     CheckLagValid(metadata, cg_metadata);
@@ -1390,6 +1395,10 @@ rocksdb::Status Stream::GetConsumerInfo(
 
   auto iter = util::UniqueIterator(ctx, read_options, stream_cf_handle_);
   for (iter->SeekToFirst(); iter->Valid(); iter->Next()) {
+    if (identifySubkeyType(iter->key()) != StreamSubkeyType::StreamConsumerMetadata) {
+      continue;
+    }
+
     std::string consumer_name = consumerNameFromInternalKey(iter->key());
     StreamConsumerMetadata c_metadata = decodeStreamConsumerMetadataValue(iter->value().ToString());
     std::pair<std::string, StreamConsumerMetadata> tmp_item(consumer_name, c_metadata);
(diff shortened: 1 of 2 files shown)


nathanlo-hrt:
thanks for taking a look at this!

src/types/redis_stream.cc
```diff
@@ -1340,7 +1340,8 @@ rocksdb::Status Stream::GetGroupInfo(engine::Context &ctx, const Slice &stream_n
   rocksdb::Status s = GetMetadata(ctx, ns_key, &metadata);
   if (!s.ok()) return s;
 
-  std::string subkey_type_delimiter = std::to_string(UINT64_MAX);
+  std::string subkey_type_delimiter;
+  PutFixed64(&subkey_type_delimiter, UINT64_MAX);
```
PragmaTwice, line 1344: Good catch!

@git-hulk That's seems a very serious bug in stream group which leads to wrong data encoding.
Should we consider to cancel the current release vote (2.10.0) for it?

PragmaTwice (approved):

mapleFU (approved):

jonathanc-n:
@LindaSummer Nice change!

Yangsx-1:
@LindaSummer Good catch!!!

Issue #2555, opened 2024-09-23
Tracking Issues: Add support for RediSearch Commands


jonathanc-n:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

references the tracker issue: #2064 

### FT Commands Implementation Progress

- [x] **FT._LIST**: Returns a list of all existing indexes
- [ ] **FT.AGGREGATE**: Run a search query on an index and perform aggregate transformations on the results
- [ ] **FT.ALIASADD**: Adds an alias to the index
- [ ] **FT.ALIASDEL**: Deletes an alias from the index
- [ ] **FT.ALIASUPDATE**: Adds or updates an alias to the index
#2577 
- [ ] **FT.ALTER**: Adds a new field to the index
- [ ] **FT.CONFIG GET**: Retrieves runtime configuration options
- [ ] **FT.CONFIG SET**: Sets runtime configuration options
- [x] **FT.CREATE**: Creates an index with the given spec
- [ ] **FT.CURSOR DEL**: Deletes a cursor
- [ ] **FT.CURSOR READ**: Reads from a cursor
- [ ] **FT.DICTADD**: Adds terms to a dictionary
- [ ] **FT.DICTDEL**: Deletes terms from a dictionary
- [ ] **FT.DICTDUMP**: Dumps all terms in the given dictionary
- [x] **FT.DROPINDEX**: Deletes the index
- [x] **FT.EXPLAIN**: Returns the execution plan for a complex query
- [ ] **FT.EXPLAINCLI**: Returns the execution plan for a complex query
- [x] **FT.INFO**: Returns information and statistics on the index
- [ ] **FT.PROFILE**: Performs a `FT.SEARCH` or `FT.AGGREGATE` command and collects performance information
- [x] **FT.SEARCH**: Searches the index with a textual query, returning either documents or just ids
- [ ] **FT.SPELLCHECK**: Performs spelling correction on a query, returning suggestions for misspelled terms
- [ ] **FT.SYNDUMP**: Dumps the contents of a synonym group
- [ ] **FT.SYNUPDATE**: Creates or updates a synonym group with additional terms
- [x] **FT.TAGVALS**: Returns the distinct tags indexed in a Tag field
#2564 

FT.SEARCHSQL and FT.EXPLAINSQL has also been additionally implemented in Kvrocks

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

[...]

PR #2577, opened 2024-10-04
feat(search): Added Alias Commands to Search

jonathanc-n:
references #2555 

Added alias commands to search with unit tests in Go. (FT.ALIASADD, FT.ALIASDEL, FT.ALIASUPDATE)

Just wanted to see if this seemed fine before adding it to the actual alias functionality of the other search commands. Will create the alias functionality immediately if this PR gets merged.

Alias Subkey 
ns | type | index | alias_name => index_name

### DIFF
--- a/src/commands/cmd_search.cc
+++ b/src/commands/cmd_search.cc
@@ -471,8 +471,49 @@ class CommandFTList : public Commander {
 class CommandFTDrop : public Commander {
   Status Execute(Server *srv, Connection *conn, std::string *output) override {
     const auto &index_name = args_[1];
+    engine::Context ctx(srv->storage);
 
-    GET_OR_RET(srv->index_mgr.Drop(index_name, conn->GetNamespace()));
+    GET_OR_RET(srv->index_mgr.Drop(ctx, index_name, conn->GetNamespace()));
+
+    output->append(SimpleString("OK"));
+
+    return Status::OK();
+  };
+};
+
+class CommandFTAliasAdd : public Commander {
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    const auto &alias_name = args_[1];
+    const auto &index_name = args_[2];
+    engine::Context ctx(srv->storage);
+
+    GET_OR_RET(srv->index_mgr.AddAlias(ctx, alias_name, index_name, conn->GetNamespace()));
+    output->append(SimpleString("OK"));
+
+    return Status::OK();
+  };
+};
+
+class CommandFTAliasDel : public Commander {
+  Status Execute(Server *srv, Connection *conn, std::string *output) override {
+    const auto &alias_name = args_[1];
+    engine::Context ctx(srv->storage);
+
+    GET_OR_RET(srv->index_mgr.DelAlias(ctx, alias_name, conn->GetNamespace()));
+
+    output->append(SimpleString("OK"));
+
+    return Status::OK();
+  };
+};
+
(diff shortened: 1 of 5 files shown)


jonathanc-n:
Is somebody able to resolve the conflict, its just the previous command is just in the same place. When I try to retrieve from origin the previous commits aren't showing up right now, sorry about that

PragmaTwice (changes requested):
Thanks for your contribution.

However FIELD_ALIAS is not for this purpose. It's for alias of index fields.

Also we should see the encoding design in your proposal, but I think now it's not consistent with the implementation.

jonathanc-n:
@PragmaTwice I decided to switch it up the serialization on the implementation, I can include it on the pr. As for the field_alias, you can use this implementation to use index fields to look and see if the alias matches when it receives a non matching index in a command (which will be implemented in a GetAlias function).

PragmaTwice:
`FT.ALIAS*` is for alias of **indexes**.

FIELD_ALIAS is for alias of **fields**.

**Indexes** is not **fields**, and one **index** can includes several **fields**.

Your implementation is quite confusing, if you really want to add alias of indexes, the encoding should be something like
```
ns | INDEX_ALIAS | alias => index
```

If you put index name before alias, how do you quickly map from alias to the index name before you know the original index?

I'd suggest that figure out what you want to do and how to do before writing some random code.

jonathanc-n:
@PragmaTwice Sorry about that, I didn't know there was aliases for the fields. I wanted to make use of the SearchKey for search fields, and saw that it had an index field and thought I could include it for loading up index information. When implementing the add/delete/update commands I would fetch with the index field in the search key being empty. Would this make it slower?

Sorry again about the ignorance, I just started learning computer science in my last year of highschool and was excited about databases. After a couple months I tried to learn about Kvrocks and make contributions since I was pretty excited about the database itself. I will try harder to make better contributions in the future

PragmaTwice:
Besides ADD, UPDATE and DELETE operations of alias, the most important part of aliasing is to find the corresponding index by the user-provided alias in query commands (e.g. FT.SEARCH, FT.SEARCHSQL).

So we need to consider how to efficiently identify the alias and find the corresponding index in the query (in a very little runtime overhead). e.g. for index `a`, `b` and their aliases (`c` -> index `a`, `d` -> index `b`), we can put a key like this:
```
// alias
ns | INDEX_ALIAS | c -> a
ns | INDEX_ALIAS | d ->b

// index metadata (already exists)
ns | INDEX_META | a -> ...
ns | INDEX_META | b -> ...
```

And when we saw a query like `select * from <idx>`, we need to check if `<idx>` is a real index name or it's an alias, e.g.
- `select * from a`: `a` is a real index name.
- `select * from d`: `d` is an alias and the real index is `b`.
- `select * from e`: `e` doesn't exist.

For index names and its metadata, we will load all of them into the memory from start instead of getting it from rocksdb every time we need it to achieve better performance. 

So I think for alias we should also load from start.

In your design:
> Alias Subkey
> ns | type | *index* | alias_name => *index_name*

The index name seems repeated in both keys and values, which is not so useful.

jonathanc-n:
@PragmaTwice While the load commands happen do you think it'd be better to create a global alias map in index_manager? 
There would be two places for aliases to be stored:
1. A global alias map where it is 'alias -> index' for quick alias to index lookups.
2. A alias array in IndexInfo for deleting all aliases that are associated with the index quickly if ft.dropindex is called.

PR #2589, opened 2024-10-11
fix: some rocksdb properties are counted multiple times

sryanyuan:
Some rocksdb properties, such as `num_snapshots/num_running_flushes/num_running_compactions`, are read directly from DB stats.  However, using `GetAggregatedIntProperty` can cause the property values to be counted multiple times, resulting in a final value that is the original value multiplied by the number of column families. This can be misleading when troubleshooting performance issues, especially when compared to the rocksdb configuration field 'max_background_jobs'.


torwig (approved):

torwig:
@sryanyuan Thank you for this fix.

git-hulk (approved):
Good catch

PR #2590, opened 2024-10-11
ci: add arm64 linux pipeline

PragmaTwice:
Add arm64 linux pipeline via ASF self-hosted runners.


tisonkun (dismissed):
This is cool! We may use it for the release workflow later.

git-hulk (dismissed):
Awesome!

PragmaTwice:
https://github.com/apache/kvrocks/actions/runs/11296291234/job/31435412588?pr=2590

The CI job via ASF self-hosted runner is cancelled. Not sure why.

> The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.
The self-hosted runner: asf-arm-trzxd-rh9np lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.


cc @gmcdonald

aleksraiden:
As I see, arm64 now public in Github - https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

aleksraiden:
All test passed, maybe are ok to merge?

.github/workflows/kvrocks.yaml
```diff
@@ -212,11 +216,15 @@ jobs:
           echo "NPROC=$(sysctl -n hw.ncpu)" >> $GITHUB_ENV
           echo "CMAKE_EXTRA_DEFS=-DOPENSSL_ROOT_DIR=/usr/local/opt/openssl" >> $GITHUB_ENV
       - name: Setup Linux
-        if: ${{ startsWith(matrix.os, 'ubuntu') }}
+        if: ${{ startsWith(matrix.os, 'ubuntu') || matrix.arm_linux }}
```
tisonkun, line 219:
```suggestion
        if: ${{ startsWith(matrix.os, 'ubuntu') }}
```

"Ubuntu Arm64" falls into `startsWith(matrix.os, 'ubuntu') == true`?

tisonkun (approved):

.github/workflows/kvrocks.yaml
aleksraiden: @PragmaTwice

PragmaTwice: ahh you are right. missed it.

PR #2592, opened 2024-10-12
feat(config): support retrieving configuration fields using glob patterns

sryanyuan:
It can be difficult to remember the full configuration field, so providing a keyword to retrieve the configuration value, as is done in Redis, can be more convenient.


src/vendor/globmatch.cc
```diff
@@ -0,0 +1,138 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "globmatch.h"
+
+// NOLINTBEGIN
+/* Glob-style pattern matching. */
+int stringmatchlen(const char *pattern, int patternLen, const char *string, int stringLen, int nocase) {
```
PragmaTwice: Seems we already have one called `StringMatch` in https://github.com/apache/kvrocks/blob/unstable/src/common/string_util.cc .

sryanyuan: ok, i'll use this one

sryanyuan: fixed

src/config/config.cc
```diff
@@ -867,7 +868,7 @@ Status Config::Load(const CLIOptions &opts) {
 void Config::Get(const std::string &key, std::vector<std::string> *values) const {
   values->clear();
   for (const auto &iter : fields_) {
-    if (key == "*" || util::ToLower(key) == iter.first) {
+    if (util::StringMatchLen(key.c_str(), key.size(), iter.first.c_str(), iter.first.size(), 1)) {
```
```suggestion
    if (util::StringMatch(key, iter.first, 1)) {
```

sryanyuan: ok

sryanyuan: Done

PragmaTwice (approved):
Looks good to me.

git-hulk (approved):

PR #2598, opened 2024-10-14
feat(worker): Allow passing a socket FD to dup and listen on

nathanlo-hrt:
When spawning Kvrocks as a part of a script, it's sometimes helpful to first reserve a port with a socket, and pass that socket to Kvrocks as a child process. 

This change implements this change:
1. Adds a config item `socket-fd` which defaults to -1 (no fd specified).
2. If `socket-fd` is provided, will skip checking that the supplied port is available (since it will already be grabbed by the provided fd). Instead, it will check that the supplied fd is indeed a socket associated to the provided port.
3. If `socket-fd` is provided, `Worker`'s will `dup` the supplied fd and listen to that instead of trying to make their own socket from scratch.
4. Close `socket-fd` on shut-down.

As a bonus, removed some unneeded indirect includes and cleaned up some surrounding code.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -110,6 +110,7 @@ Config::Config() {
       {"daemonize", true, new YesNoField(&daemonize, false)},
       {"bind", true, new StringField(&binds_str_, "")},
       {"port", true, new UInt32Field(&port, kDefaultPort, 1, PORT_LIMIT)},
+      {"socket-fd", true, new IntField(&socket_fd, -1, -1, 1 << 16)},
 #ifdef ENABLE_OPENSSL
       {"tls-port", true, new UInt32Field(&tls_port, 0, 0, PORT_LIMIT)},
       {"tls-cert-file", false, new StringField(&tls_cert_file, "")},
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -20,6 +20,16 @@ bind 127.0.0.1
 # unixsocket /tmp/kvrocks.sock
 # unixsocketperm 777
 
+# Allows a parent process to open a socket and pass its FD down to kvrocks as a child 
+# process. Useful to reserve a port and prevent race conditions.
+# 
+# PLEASE NOTE: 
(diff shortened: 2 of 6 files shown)


PragmaTwice:
The implementation looks good.

But could you elabrate more about the use case? i.e. in which cases it's better than connecting to tcp/unix sockets? Maybe you can attach a shell script or something to illustrate it.

nathanlo-hrt:
Currently, the Kvrocks port listening implementation has a race condition: since Kvrocks checks whether the port is in use before actually grabbing exclusive use of the port, it's possible to spawn two Kvrocks servers listening on the same ports.

The minimal repro of this is to run `build/kvrocks --port 8888 & build/kvrocks --port 8888` a few times: eventually both will succeed and you'll be able to issue commands to port 8888 and cause undefined behavior.

As such, I have a script like below which reserves a port, and passes the socket's FD to spawn Kvrocks:
```py
import socket

s = socket.create_server(...)  # Grab a port which is guaranteed to be free
port = s.getsockname()[1]
socket_fd = s.fileno()
subprocess.run(f"./kvrocks --port {port} --socket-fd {socket_fd}")

# Now, Kvrocks can safely duplicate and use the FD `socket_fd` without fear of other Kvrocks instances trampling it
```

src/config/config.cc
```diff
@@ -110,6 +110,7 @@ Config::Config() {
       {"daemonize", true, new YesNoField(&daemonize, false)},
       {"bind", true, new StringField(&binds_str_, "")},
       {"port", true, new UInt32Field(&port, kDefaultPort, 1, PORT_LIMIT)},
+      {"socket-fd", true, new IntField(&socket_fd, -1, -1, 1 << 16)},
```
PragmaTwice, line 113: We need to add `socket-fd` into `kvrocks.conf` and add some comments to describe it.

Also we need to note to users that: if it's specified, `bind` and `tls*` cannot work.

PragmaTwice (approved):

git-hulk (approved):

PR #2599, opened 2024-10-14
fix: Better checking for prefix matches

nathanlo-hrt:
Currently a match pattern like `*ab*` is silently accepted, because we only check whether the last character is *, so SCAN just looks for keys beginning with the literal `*ab`. 

However this is misleading; this PR makes this check more precise.


torwig:
@nathanlo-hrt Thank you for your contribution. Ideally, is to add a dedicated test for the case(s) you highlighted.

PragmaTwice (dismissed):
LGTM

torwig (dismissed):

PragmaTwice (approved):

git-hulk (approved):

Issue #2602, opened 2024-10-16
Make configurable compression for L0-L1


paragor:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

I have high write rate, and data have good compression (coef 8-10). But in kvrocks - L0 and L1 dont compress data. And i have a very big write/read rate for compaction L0-L1.
may be move to config this logic?

```
    // For the first two levels, it may contain the frequently accessed data,
    // so it'd be better to use uncompressed data to save the CPU.
    std::string compression_levels = "kNoCompression:kNoCompression";
    auto db = srv->storage->GetDB();
    for (size_t i = 2; i < db->GetOptions().compression_per_level.size(); i++) {
      compression_levels += ":";
      compression_levels += compression_option;
    }
    return srv->storage->SetOptionForAllColumnFamilies("compression_per_level", compression_levels);
  };
```

### Solution

add setting in config
```
 rocksdb.enable_compression_for_l0_and_l1 yes/NO
 ``` 

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

**Comment by PragmaTwice** [comment]
Maybe we can have an option called `rocksdb.nocompression_level`, 
and `rocksdb.nocompression_level n` means that the first `n` levels don't need to compress.

The default value can be 2.

**Comment by PragmaTwice** [comment]
Maybe we can have an option called `rocksdb.nocompression_level`, 

and `rocksdb.nocompression_level n` means that the first `n` levels don't need to compress.

**Comment by paragor** [comment]
```
    int compression_level;
    int nocompression_level;
```
funny naming :)

i think about something more verbose like
```
nocompression_for_first_levels
```

PragmaTwice:
Good to me.

Also we can/should explain to users in comments.

aleksraiden:
Good to me too, we also have an heavy write rate system

PR #2605, opened 2024-10-17
feat(config): add `rocksdb.compression_start_level` to allow configure levels without compression

paragor:
references: https://github.com/apache/kvrocks/issues/2602

add `rocksdb.compression_start_level` to allow configure levels without compression

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -814,6 +814,15 @@ rocksdb.compression_level 32767
 # Default: 2 MB
 rocksdb.compaction_readahead_size 2097152
 
+# Disable compression for first n levels.
+# By default disabled for the first two levels, because it may contain the
+# frequently accessed data, so it'd be better to use uncompressed data to
+# save the CPU.
+# Value: [0, 7] (upper boundary is rocksdb levels count)
+#
+# Default: 2
+rocksdb.nocompression_for_first_n_levels 2
+
 # he limited write rate to DB if soft_pending_compaction_bytes_limit or
 # level0_slowdown_writes_trigger is triggered.
 
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -814,14 +814,14 @@ rocksdb.compression_level 32767
 # Default: 2 MB
 rocksdb.compaction_readahead_size 2097152
 
-# Disable compression for first n levels.
-# By default disabled for the first two levels, because it may contain the
-# frequently accessed data, so it'd be better to use uncompressed data to
-# save the CPU.
-# Value: [0, 7] (upper boundary is rocksdb levels count)
+# Enable compression from n levels of LSM-tree.
+# By default compression is disabled for the first two levels (L0 and L1),
+# because it may contain the frequently accessed data, so it'd be better
+# to use uncompressed data to save the CPU.
+# Value: [0, 7) (upper boundary is kvrocks maximum levels number)
 #
 # Default: 2
-rocksdb.nocompression_for_first_n_levels 2
+rocksdb.compression_start_level 2
 
 # he limited write rate to DB if soft_pending_compaction_bytes_limit or
 # level0_slowdown_writes_trigger is triggered.
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -211,6 +211,9 @@ Config::Config() {
        new EnumField<rocksdb::CompressionType>(&rocks_db.compression, compression_types,
                                                rocksdb::CompressionType::kNoCompression)},
       {"rocksdb.compression_level", true, new IntField(&rocks_db.compression_level, 32767, INT_MIN, INT_MAX)},
+      // can't be more than rocksdb levels count
+      {"rocksdb.nocompression_for_first_n_levels", false,
+       new IntField(&rocks_db.nocompression_for_first_n_levels, 2, 0, 7)},
       {"rocksdb.block_size", true, new IntField(&rocks_db.block_size, 16384, 0, INT_MAX)},
       {"rocksdb.max_open_files", false, new IntField(&rocks_db.max_open_files, 8096, -1, INT_MAX)},
       {"rocksdb.write_buffer_size", false, new IntField(&rocks_db.write_buffer_size, 64, 0, 4096)},
@@ -381,12 +384,11 @@ void Config::initFieldCallback() {
     return srv->storage->SetOptionForAllColumnFamilies(TrimRocksDbPrefix(k), v);
   };
   auto set_compression_type_cb = [](Server *srv, [[maybe_unused]] const std::string &k,
-                                    const std::string &v) -> Status {
+                                    [[maybe_unused]] const std::string &v) -> Status {
     if (!srv) return Status::OK();
-
     std::string compression_option;
     for (auto &option : engine::CompressionOptions) {
-      if (option.name == v) {
+      if (option.type == srv->GetConfig()->rocks_db.compression) {
         compression_option = option.val;
         break;
       }
@@ -395,15 +397,23 @@ void Config::initFieldCallback() {
       return {Status::NotOK, "Invalid compression type"};
     }
 
-    // For the first two levels, it may contain the frequently accessed data,
-    // so it'd be better to use uncompressed data to save the CPU.
-    std::string compression_levels = "kNoCompression:kNoCompression";
-    auto db = srv->storage->GetDB();
-    for (size_t i = 2; i < db->GetOptions().compression_per_level.size(); i++) {
-      compression_levels += ":";
-      compression_levels += compression_option;
+    const int nocompression_for_first_n_levels = srv->GetConfig()->rocks_db.nocompression_for_first_n_levels;
+    const int num_levels = srv->storage->GetDB()->GetOptions().num_levels;
+    if (nocompression_for_first_n_levels > num_levels) {
+      return {Status::NotOK, "nocompression_for_first_n_levels must <= rocksdb levels count"};
     }
-    return srv->storage->SetOptionForAllColumnFamilies("compression_per_level", compression_levels);
+    std::vector<std::string> compression_per_level_builder;
+    compression_per_level_builder.reserve(num_levels);
+
+    for (int i = 0; i < nocompression_for_first_n_levels; i++) {
+      compression_per_level_builder.emplace_back("kNoCompression");
+    }
+    for (int i = nocompression_for_first_n_levels; i < num_levels; i++) {
+      compression_per_level_builder.emplace_back(compression_option);
+    }
+    const std::string compression_per_level = util::StringJoin(
+        compression_per_level_builder, [](const auto &s) -> decltype(auto) { return s; }, ":");
+    return srv->storage->SetOptionForAllColumnFamilies("compression_per_level", compression_per_level);
   };
 #ifdef ENABLE_OPENSSL
   auto set_tls_option = [](Server *srv, [[maybe_unused]] const std::string &k, [[maybe_unused]] const std::string &v) {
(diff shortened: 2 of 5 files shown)


kvrocks.conf
```diff
@@ -794,6 +794,14 @@ rocksdb.compression_level 32767
 # Default: 2 MB
 rocksdb.compaction_readahead_size 2097152
 
+# Disable compression for first n levels.
+# By default disabled for the first two levels, because it may contain the
+# frequently accessed data, so it'd be better to use uncompressed data to
+# save the CPU.
+#
+# Default: 2
+rocksdb.nocompression_for_first_levels 2
```
PragmaTwice: After some rethinking, maybe it can be `nocompression_for_first_n_levels`, since `first` usually refers to just one element.

paragor: ok, i will fix it

paragor: fixed

PragmaTwice:
From the CI log, we need to add `vector.reserve()` before the loop containing `vector.push_back()`.

paragor:
> From the CI log, we need to add `vector.reserve()` before the loop containing `vector.push_back()`.

thank you, didnt see it.
fixed

src/config/config.cc
```diff
@@ -382,14 +383,20 @@ void Config::initFieldCallback() {
...
-      compression_levels += ":";
-      compression_levels += compression_option;
+    const auto config = srv->GetConfig();
+    std::vector<std::string> compression_levels_builder;
+    compression_levels_builder.reserve(db->GetOptions().compression_per_level.size());
+
+    for (int i = 0; i < config->rocks_db.nocompression_for_first_n_levels; i++) {
+      compression_levels_builder.emplace_back("kNoCompression");
+    }
+    for (size_t i = config->rocks_db.nocompression_for_first_n_levels;
```
PragmaTwice: We need to check at the beginning of the callback that nocompression_for_first_n_levels <= compression_per_level.size().

Also you can make a variable for compression_per_level.size().

src/config/config.cc
```diff
@@ -201,6 +201,8 @@ Config::Config() {
        new EnumField<rocksdb::CompressionType>(&rocks_db.compression, compression_types,
                                                rocksdb::CompressionType::kNoCompression)},
       {"rocksdb.compression_level", true, new IntField(&rocks_db.compression_level, 32767, INT_MIN, INT_MAX)},
+      {"rocksdb.nocompression_for_first_n_levels", false,
+       new IntField(&rocks_db.nocompression_for_first_n_levels, 2, 0, 7)},
```
PragmaTwice: I'm wondering why it's 7 here? maybe we can add some comments.

paragor: kvrocks support only 7 levels of leveled compaction, ok i stay a comment

src/config/config.cc
paragor: i was thinking that boundary check is unnecessary, because kvrocks has hardcoded 7 lvles of compactions. but it is ok, i add it

paragor: fixed

src/config/config.cc
paragor: fixed

src/config/config.cc
```diff
@@ -368,12 +371,11 @@ void Config::initFieldCallback() {
...
   };
   auto set_compression_type_cb = [](Server *srv, [[maybe_unused]] const std::string &k,
-                                    const std::string &v) -> Status {
+                                    [[maybe_unused]] const std::string &v) -> Status {
     if (!srv) return Status::OK();
-
     std::string compression_option;
     for (auto &option : engine::CompressionOptions) {
-      if (option.name == v) {
+      if (option.type == srv->GetConfig()->rocks_db.compression) {
```
Beihao-Zhou: Just wondering if this change here is just to make the comparison condition more concrete, or it has to be changed to make `nocompression_for_first_levels` work? Thank you!

paragor: previously it compare value that come to callback. callback was used once when "compression" option set. Now callback call twice - for set "compression" and for set "nocompression_for_first_n_levels", so value may be name of compression or levels count that no need compression.

if i make two separate callback, it will have dublicate code, because this two options closely related.

paragor:
@PragmaTwice what do you think about rename nocompression_for_first_n_levels  -> start_compression_level?

src/config/config.cc
```diff
@@ -395,15 +397,23 @@ void Config::initFieldCallback() {
...
+    if (nocompression_for_first_n_levels > compression_per_level_size) {
+      return {Status::NotOK, "nocompression_for_first_n_levels must <= rocksdb levels count"};
+    }
+    std::vector<std::string> compression_per_level_builder;
+    compression_per_level_builder.reserve(compression_per_level_size);
+
+    for (int i = 0; i < nocompression_for_first_n_levels; i++) {
+      compression_per_level_builder.emplace_back("kNoCompression");
+    }
+    for (int i = nocompression_for_first_n_levels; i < compression_per_level_size; i++) {
```
mapleFU: should we use `num_levels`?

src/storage/storage.cc
```diff
@@ -171,7 +171,7 @@ rocksdb::Options Storage::InitRocksDBOptions() {
   options.wal_compression = config_->rocks_db.wal_compression;
   // only compress levels >= 2
```
mapleFU: should fix the comment here

mapleFU (commented):
So AFAIK, just like https://github.com/apache/kvrocks/pull/2607 . You'd like to enable compression on L1 ( or even L0 ) for write intensive case?

mapleFU:
Perhaps, if you're a heavy write system, maybe you can enable different compression in different level, like lz4 in low level and zstd in high level...

wanghenshui:
maybe we could specific compression setting by level? user could mark l0/l1 as snappy or lz4 compression style

l0/l1 nocompression is common type of usage,  use or not, 

rocksdb.nocompression_for_first_n_levels  indicate user could make l2/l3... no compression, user may not know about rocksdb stuff, make them confused about what senerio to use it

instead, let user specifc what compression style level used, maybe more clear.

paragor:
> So AFAIK, just like #2607 . You'd like to enable compression on L1 ( or even L0 ) for write intensive case?

yes, write 128MB compressed data on disk is better, than write uncompressed 1 GB data on disk. Now cpu usage is only 1/8 cores. so i need able to load cpu and decrease disk usage.

it is not general case, but in my case it will help me to use disks with fewer throughput and as a consequence of ec2 limitations - cheaper ec2 instance

paragor:
> maybe we could specific compression setting by level? user could mark l0/l1 as snappy or lz4 compression style

I think this is a good point
Something similar to the more advanced settings in kvrocks. By default, the "compression" settings are used, but if the new setting "compression_by_levels" is specified, it takes precedence, and the "compression" setting is ignored.

mapleFU:
Maybe we should first check this in, would you mind fix the comments firstly?

src/config/config.cc
paragor: fixed

src/storage/storage.cc
paragor: fixed

src/config/config.cc
paragor: i dont know how make this code better, if you have solution, plz share with me :)
i never write in c++ before

kvrocks.conf
```diff
@@ -814,6 +814,15 @@ rocksdb.compression_level 32767
 # Default: 2 MB
 rocksdb.compaction_readahead_size 2097152
 
+# Disable compression for first n levels.
+# By default disabled for the first two levels, because it may contain the
+# frequently accessed data, so it'd be better to use uncompressed data to
+# save the CPU.
+# Value: [0, 7] (upper boundary is rocksdb levels count)
```
mapleFU:
```suggestion
# Value: [0, 7] (upper boundary is kvrocks maximum levels count)
```

Would above being better?

src/config/config.cc
mapleFU: I think rather than a comment, we can use a `constexpr` valuable like `KVROCKS_MAX_LSM_LEVEL = 7`, and use it in config and here

src/config/config.cc
mapleFU: This is a bit weird to reuse the same function...@git-hulk do you have any better idea here? There're 2 configs about compression and would set the system once...

git-hulk: I have no better solution for now because both two configurations will affect the same rocksdb option. It's weird while compared to other configurations, but it looks reasonable.

mapleFU: Should we separate it to:

```
set_compression_method_callback
set_no_compression_level_callback
```

?

src/config/config.cc
paragor: where should i declare const? in config.h or storage.h?

src/config/config.cc
git-hulk: @mapleFU Do you mean this?

```C++
rocksdb::Status set_rocksdb_compression(CompressionType compression, int nocompression_for_first_n_levels) {
   ...
}

rocksdb::Status set_compression_method_callback() {
   ...
   set_rocksdb_compression(compression, nocompression_for_first_n_levels)
}

rocksdb::Status set_no_compression_level_callback() {
  ...
  set_rocksdb_compression(compression, nocompression_for_first_n_levels)
}

```

If yes, I'm good with it.

src/config/config.cc
paragor: oh i see config.h best place for it

mapleFU: Ok to be in config.h

PragmaTwice:
> @PragmaTwice what do you think about rename nocompression_for_first_n_levels -> start_compression_level?

Nice. But seems compression_start_level or compression_begin_level is better to me.

src/config/config.cc
paragor: i'am do it in separate commit, but i dont see any benefits. validation of fields is done already in `IntField` and `EnumField<rocksdb::CompressionType>`, so no need validate options again. 
set_db_option_cb used more then 1 times for example

kvrocks.conf
paragor: fixed

paragor:
> > @PragmaTwice what do you think about rename nocompression_for_first_n_levels -> start_compression_level?
> 
> Nice. But seems compression_start_level or compression_begin_level is better to me.

renamed

src/config/config.cc
paragor: fixed

git-hulk (approved):

mapleFU (approved):
LGTM, thanks!

src/config/config.cc
```diff
@@ -105,6 +105,37 @@ std::string TrimRocksDbPrefix(std::string s) {
...
+  }
+  if (compression_option.empty()) {
+    return {Status::NotOK, "Invalid compression type"};
+  }
+
+  if (compression_start_level >= KVROCKS_MAX_LSM_LEVEL) {
+    return {Status::NotOK, "compression_start_level must <= rocksdb levels count"};
+  }
+  std::vector<std::string> compression_per_level_builder;
+  compression_per_level_builder.reserve(KVROCKS_MAX_LSM_LEVEL);
```
PragmaTwice, line 126: I'm wondering if `db->GetOptions().compression_per_level.size()` equals to `KVROCKS_MAX_LSM_LEVEL`?

PragmaTwice: Also is compression level equals to LSM level?

PragmaTwice: Ahh it's right. Just a confused `level` between `compression_start_level` and `compression_level`..

PragmaTwice (approved):

PragmaTwice (approved):

src/config/config.cc
```diff
@@ -105,6 +105,37 @@ std::string TrimRocksDbPrefix(std::string s) {
...
+      compression_option = option.val;
+      break;
+    }
+  }
+  if (compression_option.empty()) {
+    return {Status::NotOK, "Invalid compression type"};
+  }
+
+  if (compression_start_level >= KVROCKS_MAX_LSM_LEVEL) {
+    return {Status::NotOK, "compression_start_level must <= rocksdb levels count"};
```
mapleFU, line 123: Oh I catch a problem here, should we weeken to `!(compression_start_level > KVROCKS_MAX_LSM_LEVEL)`?

mapleFU: Oh here we should just change content to `return {Status::NotOK, "compression_start_level must < rocksdb levels count"};`

Issue #2606, opened 2024-10-17
Add abilty to compress WAL


paragor:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

I have heavy write, and instance have a lot of free cpu, but have not enough disk throughput. So i think i need able to configure WAL compression. 

### Solution

https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log-(WAL)
https://github.com/facebook/rocksdb/wiki/WAL-Compression
https://github.com/facebook/rocksdb/blob/787730c859d83fd82b3087ae2edf189f8f2e698f/include/rocksdb/options.h#L1365

Currently supported only ZSTD.
It is not dynamic option - can be set only on open DB.

rocksdb already supported wal compression, just need to add option `rocksdb.wal_compression` Yes/No.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #2607, opened 2024-10-17
feat(config): add `rocksdb.wal_compression` to allow enable wal compression

paragor:
issue: https://github.com/apache/kvrocks/issues/2606

add `rocksdb.wal_compression` to allow enable wal compression
only zstd supported, so field YesOrNo

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -708,6 +708,15 @@ rocksdb.max_background_flushes -1
 # Default: 2
 rocksdb.max_subcompactions 2
 
+# If enabled WAL records will be compressed before they are written. Only
+# ZSTD (= kZSTD) is supported (until streaming support is adapted for other
+# compression types). Compressed WAL records will be read in supported
+# versions (>= RocksDB 7.4.0 for ZSTD) regardless of this setting when
+# the WAL is read.
+#
+# Default is no
+rocksdb.wal_compression no
+
 # In order to limit the size of WALs, RocksDB uses DBOptions::max_total_wal_size
 # as the trigger of column family flush. Once WALs exceed this size, RocksDB
 # will start forcing the flush of column families to allow deletion of some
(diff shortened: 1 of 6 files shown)


kvrocks.conf
```diff
@@ -708,6 +708,15 @@ rocksdb.max_background_flushes -1
...
 rocksdb.max_subcompactions 2
 
+# If enabled WAL records will be compressed before they are written. Only
+# ZSTD (= kZSTD) is supported (until streaming support is adapted for other
+# compression types). Compressed WAL records will be read in supported
+# versions (>= RocksDB 7.4.0 for ZSTD) regardless of this setting when
+# the WAL is read.
+#
+# Default is no
+rocksdb.wal_compression no
```
PragmaTwice, line 729: I think it can be `none` and `zstd` instead of `no` and `yes` to leave some upgrading space.

paragor: ok, i will fix it

paragor: fixed

PragmaTwice (approved):

git-hulk (approved):

mapleFU:
By the way, may I ask that is this only for write-heavy workload(maybe with long and compress-able key-values)?

paragor:
> By the way, may I ask that is this only for write-heavy workload(maybe with long and compress-able key-values)?

i have a lot key-value (uncompressed size in [256B, 8kB]) with good compression rate (value is json, where sum length of json's keys is bigger then json's values).
i have ~100-200MB/s write, wal and compaction load, so on aws i need use aws ec2 instances with enough cpu (see ec2 instance throughput limits https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-optimized.html). 

finally i have a lot of free cpu. this pull request can help me decrease write load on disk and i can pickup cheaper ec2

mapleFU:
I see, this make sense

PR #2608, opened 2024-10-17
feat(keys, scan): Support arbitrary glob patterns

nathanlo-hrt:
Kvrocks currently only supports prefix matching (glob patterns like `ab*`).
This change implements arbitrary glob patterns for `KEYS` and `SCAN MATCH [pattern]`.


tests/gocase/unit/keyspace/keyspace_test.go
```diff
@@ -77,6 +74,37 @@ func TestKeyspace(t *testing.T) {
 		require.Equal(t, []string{"foo_a", "foo_b", "foo_c", "key_x", "key_y", "key_z"}, keys)
 	})
 
+	t.Run("KEYS with non-trivial patterns", func(t *testing.T) {
+		require.NoError(t, rdb.FlushDB(ctx).Err())
+		for _, key := range []string{"aa", "aab", "aabb", "ab", "abb"} {
+			require.NoError(t, rdb.Set(ctx, key, "hello", 0).Err())
+		}
+
+		keys := rdb.Keys(ctx, "a*").Val()
```
PragmaTwice, line 96: Could you add a case to ensure that `aa` just matches `aa`, and `aa*` can match `aa`, `aab`, ..?

src/common/glob.h
```diff
@@ -0,0 +1,158 @@
...
+  // the pattern consists only of '*', and the string is empty.
+  return string.empty() && std::all_of(pattern.begin(), pattern.end(), [](char c) { return c == '*'; });
+}
+
+// Given a glob [pattern] and a string [string], return true iff the string matches the glob.
+// If [ignore_case] is true, the match is case-insensitive.
+constexpr inline bool GlobMatches(std::string_view glob, std::string_view str, bool ignore_case = false) {
+  bool skip_longer_matches = false;
+  return GlobMatchImpl(glob, str, ignore_case, &skip_longer_matches);
+}
```
PragmaTwice: Are we able to reuse the glob impl in string_util.h? If not it's fine to impl it here.

nathanlo-hrt: Thanks, I changed `StringMatch` and added some performance optimizations and guards against exponential complexity attacks `KEYS *?*?*?*?*?*?*?*...` which used to completely max out the CPU usage.

tests/cppunit/string_util_test.cc
```diff
@@ -84,6 +84,119 @@ TEST(StringUtil, HasPrefix) {
...
+  EXPECT_TRUE(util::StringMatch("[a]", "a"));
+  EXPECT_FALSE(util::StringMatch("[a]", "b"));
+  EXPECT_FALSE(util::StringMatch("[^a]", "a"));
+  EXPECT_TRUE(util::StringMatch("[^a]", "b"));
+  EXPECT_TRUE(util::StringMatch("[ab]", "a"));
+  EXPECT_TRUE(util::StringMatch("[ab]", "b"));
+  EXPECT_FALSE(util::StringMatch("[ab]", "c"));
+  EXPECT_TRUE(util::StringMatch("[^ab]", "c"));
+  EXPECT_TRUE(util::StringMatch("[a-c]", "b"));
+  EXPECT_FALSE(util::StringMatch("[a-c]", "d"));
```
PragmaTwice, line 151: I'm wondering what will happen if user inputs something like `[a`, `\` or `[a-]` i.e. wrong glob patterns.

Maybe we can do some syntax check for glob patterns? So that users can get some good error messages instead of matching nothing. although it isn't required to be added in this patch.

nathanlo-hrt: Added some code to validate globs and return error messages on invalid globs

src/storage/redis_db.cc
```diff
@@ -366,6 +368,10 @@ rocksdb::Status Database::Scan(engine::Context &ctx, const std::string &cursor,
 
       if (metadata.Expired()) continue;
       std::tie(std::ignore, user_key) = ExtractNamespaceKey<std::string>(iter->key(), storage_->IsSlotIdEncoded());
+
+      if (!util::StringMatch(suffix_glob, user_key.substr(prefix.size()))) {
+        continue;
+      }
       keys->emplace_back(user_key);
       cnt++;
```
PragmaTwice, line 376: I'm wondering that, if nothing got matched in this limit, e.g. the limit is 10, and in this 10 keys no key is matched, but the 12th key is matched, will a valid cursor be returned so that users can use it to continue the scan?

Also could we add a test case to ensure that?

nathanlo-hrt: it seems like `cnt` is only incremented when we actually add a key to the result, so the loop will continue until no more keys match the prefix, or `limit` keys are inserted into the result vector

PragmaTwice: Hmmm I think we'd better refactor the logic, so that the SCAN can be quick, instead of a long-time blocking scan. E.g. we can scan a fixed maximum numbers of keys even if the limit/cnt is not reach. It's fine to return even zero matched key to users.
Or maybe we can confirm the logic in Redis?

PragmaTwice: I'm not sure if we can just merge this first and then plan to refactor to a "quick" scan.

WDYT? cc @mapleFU @git-hulk

nathanlo99:
just to say it, it would be nice to have the typos script run as a precommit hook or lint: it's a bit unideal to get the workflow approved by a reviewer just to find out you've introduced a typo in your commit somewhere

src/storage/redis_db.cc
git-hulk: Sorry for missing this comment. This sounds good to me after taking a rough review.

> Hmmm I think we'd better refactor the logic, so that the SCAN can be quick, instead of a long-time blocking scan. E.g. we can scan a fixed maximum numbers of keys even if the limit/cnt is not reach. It's fine to return even zero matched key to users.
Or maybe we can confirm the logic in Redis?

Yes, Redis will set a max iteration number(10*count) to avoid blocking too long.

Refer: https://github.com/redis/redis/blob/611c950293ae34dcef148ec62c9dd9626d7dc9e3/src/db.c#L1212

git-hulk:
Hi @nathanlo-hrt Seems the Go test is broken, would you mind fixing it to make it mergeable?

PragmaTwice (approved):

git-hulk (approved):

src/storage/redis_db.cc
jishengming1: In this code, we replace `user_key.substr(prefix.size())` with `std::string_view(user_key).substr(prefix.size())`, which seems to offer better performance, right?

PR #2611, opened 2024-10-19
feat(conn): move time-consuming ops out of ExecuteCommand for context

PragmaTwice:
The scope (position of destruction) is important for Context, so we should move time-consuming ops like `FeedMonitorConns` out of `ExecuteCommand` so that the context can be destructed before these ops. (especially after we move key locking into ExecuteCommands)


PragmaTwice:
cc @PokIsemaine

git-hulk (approved):
LGTM

PokIsemaine (approved):
LGTM

PR #2619, opened 2024-10-23
feat(cmd): avoid to use CommandAttributes::flags directly

PragmaTwice:
We should use `CommandAttributes::GenerateFlags` instead of accessing the `flags` member directly.


mapleFU:
Is this to avoid not calling `flag_gen` fn if having?

PragmaTwice:
> Is this to avoid not calling `flag_gen` fn if having?

Exactly.

mapleFU:
What about making it private to avoid some explicit calling? ( Current method lgtm but I think it's easy to use it directly next time)

PragmaTwice:
> What about making it private to avoid some explicit calling? ( Current method lgtm but I think it's easy to use it directly next time)

Done.

PragmaTwice:
Also I plan to make key_range and other related members private in the next PR.

src/commands/commander.h
```diff
@@ -182,9 +228,11 @@ struct CommandAttributes {
   // commander object generator
   CommanderFactory factory;
 
+  uint64_t InitialFlags() const { return flags_; }
```
mapleFU, line 231: Minor question: why `InitialFlags()` in some cases?

PragmaTwice: Some times we cannot get the cmd args, so maybe the initial flags is useful. Also I currently don't want to touch some legacy sources.

src/commands/commander.h
```diff
@@ -257,23 +312,21 @@ inline uint64_t ParseCommandFlags(const std::string &description, const std::str
   return flags;
 }
 
+template <typename T>
+auto MakeCmdAttr(const std::string &name, int arity, const std::string &description, NoKeyInThisCommand no_key,
+                 const AdditionalFlagGen &flag_gen = {}) {
+  CommandAttributes attr(name, arity, CommandCategory::Unknown, ParseCommandFlags(description, name), flag_gen, no_key,
```
mapleFU, line 318: So NO_KEY has becoming a different ctor fn?

mapleFU (approved):

src/commands/commander.h
PragmaTwice: Yeah, `0,0,0` is not clear enough.

Issue #2626, opened 2024-10-28
Optimization of the "OK" String


RiversJin:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

In the kvrocks codebase, we often see code like the following, representing a successfully executed action:
```c++
*output = redis::SimpleString("OK");
```
Due to the design of C++ and the limitations of std::string, although we know the returned string is always "+OK\r\n", the compiler cannot fully optimize this (https://godbolt.org/z/nr69s9z83).

Perhaps we could define "+OK\r\n" as a constant string to save on the overhead of repeated string concatenations. While this is a very short string, I believe this adjustment could provide a slight performance boost in theory.

Additionally, defining a constant string, rather than hardcoding "OK" throughout the code, may also improve code readability and maintainability.

I hope this suggestion is helpful. If there is support for this idea, I would be glad to submit a PR to implement this change.

Thank you for your consideration!

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

mapleFU:
This is ok to me since "OK" would be frequently used, but I don't know what would other frequently used str like

RiversJin:
> This is ok to me since "OK" would be frequently used, but I don't know what would other frequently used str like这对我来说很好,因为"OK"会经常使用,但我不知道其他经常使用的字符串会是什么

From the code perspective, it seems that "OK" is used most frequently, while others, like "PONG," appear occasionally. However, based on the principle of simplicity, I think caching the "OK" string alone should be sufficient.

PR #2627, opened 2024-10-28
feat(resp): optimize simple string "OK" construction

RiversJin:
… pre-allocated memory

By pre-allocating memory, the speed rate of simple strings can be improved. Additionally, using a predefined "+OK\r\n" string eliminates the need for string concatenation when returning "OK".

It solves #2626

### DIFF
--- a/src/server/redis_reply.cc
+++ b/src/server/redis_reply.cc
@@ -37,7 +37,14 @@ namespace redis {
 
 void Reply(evbuffer *output, const std::string &data) { evbuffer_add(output, data.c_str(), data.length()); }
 
-std::string SimpleString(const std::string &data) { return "+" + data + CRLF; }
+std::string SimpleString(std::string_view data) {
+  std::string res;
+  res.reserve(data.size() + 3);  // 1 for '+', 2 for CRLF
+  res += RESP_PREFIX_SIMPLE_STRING;
+  res += data;
+  res += CRLF;
+  return res;
+}
 
 std::string Error(const Status &s) { return RESP_PREFIX_ERROR + StatusToRedisErrorMsg(s) + CRLF; }
 
--- a/src/server/redis_reply.h
+++ b/src/server/redis_reply.h
@@ -24,6 +24,7 @@
 
 #include <map>
 #include <string>
+#include <string_view>
 #include <vector>
 
 #include "rocksdb/status.h"
@@ -39,7 +40,9 @@ namespace redis {
 enum class RESP { v2, v3 };
 
 void Reply(evbuffer *output, const std::string &data);
-std::string SimpleString(const std::string &data);
+std::string SimpleString(std::string_view data);
+
+static const inline std::string RESP_OK = RESP_PREFIX_SIMPLE_STRING "OK" CRLF;
 
 std::string Error(const Status &s);
 std::string StatusToRedisErrorMsg(const Status &s);
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -62,7 +62,7 @@ Status FeedSlaveThread::Start() {
     sigaddset(&mask, SIGHUP);
     sigaddset(&mask, SIGPIPE);
     pthread_sigmask(SIG_BLOCK, &mask, &omask);
-    auto s = util::SockSend(conn_->GetFD(), redis::SimpleString("OK"), conn_->GetBufferEvent());
+    auto s = util::SockSend(conn_->GetFD(), redis::RESP_OK, conn_->GetBufferEvent());
     if (!s.IsOK()) {
       LOG(ERROR) << "failed to send OK response to the replica: " << s.Msg();
       return;
--- a/src/cluster/sync_migrate_context.cc
+++ b/src/cluster/sync_migrate_context.cc
@@ -66,7 +66,7 @@ void SyncMigrateContext::TimerCB(int, [[maybe_unused]] int16_t events) {
 
 void SyncMigrateContext::OnWrite(bufferevent *bev) {
   if (migrate_result_) {
-    conn_->Reply(redis::SimpleString("OK"));
+    conn_->Reply(redis::RESP_OK);
   } else {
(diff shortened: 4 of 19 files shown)


src/server/redis_reply.cc
```diff
@@ -37,9 +37,23 @@ namespace redis {
 
 void Reply(evbuffer *output, const std::string &data) { evbuffer_add(output, data.c_str(), data.length()); }
 
-std::string SimpleString(const std::string &data) { return "+" + data + CRLF; }
+std::string SimpleString(const std::string &data) {
```
mapleFU: Can the input also be a `std::string_view`?

RiversJin: Of course! In fact, I would really like to switch to using std::string_view as the parameter type. However, I saw that const string& is widely used in the code, so I assumed it was a stylistic choice and didn't make the change. If using the former is acceptable, I’ll make the adjustment later.

mapleFU: > However, I saw that const string& is widely used in the code, so I assumed it was a stylistic choice and didn't make the change.

Previously `+` for string_view would be a bit trick, but now there is no reason not shift to that. For sso string the performance might be same but for long string `string_view` would have better performance

PragmaTwice: For some cases (e.g. we need to pass a null-terminated string to a C api), `std::string` is required since `std::string_view` will not ensure that the string is terminated by `\0`.
In other cases, it can be replaced by std::string_view if the lifetime can be maintained well (no dangling happen).

PragmaTwice: However, previously I was considering to introduce "rope" (a data structure) to build `concat`ed strings in a relatively cheaper way, but it may require many additional work.

mapleFU: > However, previously I was considering to introduce "rope" (a data structure) to build `concat`ed strings in a relatively cheaper way, but it may require many non-trivial work.

It's ok for some network buffer but would be to complex for `SimpleString`...

src/server/redis_reply.cc
```diff
@@ -37,9 +37,23 @@ namespace redis {
...
+  res += RESP_PREFIX_SIMPLE_STRING;
+  res += data;
+  res += CRLF;
+  return res;
+}
 
-std::string Error(const Status &s) { return RESP_PREFIX_ERROR + StatusToRedisErrorMsg(s) + CRLF; }
+std::string Error(const Status &s) {
+  std::string res;
+  res.reserve(s.Msg().size() + 3);  // 1 for '-', 2 for CRLF
```
mapleFU: I wonder that `s.Msg()` might not be really the final string generation when `redisErrorPrefixMapping` in `StatusToRedisErrorMsg` contains the specific error message.

mapleFU (commented):
Rest LGTM

src/server/redis_reply.cc
RiversJin: Yes, you're absolutely right; I overlooked the possibility that the text could come directly from redisErrorPrefixMapping. However, performing a hash lookup here to confirm the reserved length might be a bit heavy.

Or, do you think it’s a good idea to directly use std::max(16, s.Msg().size() + 3)? In most cases, the minimum allocation unit by the memory allocator is often two pointer lengths, and for the contents within redisErrorPrefixMapping, 16 bytes should be sufficient.

mapleFU: > However, performing a hash lookup here to confirm the reserved length might be a bit heavy.

As the hash table is small and anyway it would be built 😅, I think construct error string firstly is ok...

> do you think it’s a good idea to directly use std::max(16, s.Msg().size() + 3)

It would be good if it's a critical path, but I don't think the error path worth this, since the current code is clear to understand and Error path might not a bottleneck...

RiversJin: Ok, I've reverted this modification.

mapleFU (dismissed):

mapleFU:
Would you mind fix the lint?

RiversJin:
> Would you mind fix the lint?

Sure but ...  
From the logs, this is not just a lint issue, but rather due to setting RESP_OK as the type std::string_view. Some function parameters accept const string& instead of std::string_view. If we change everything to accept std::string_view, I think it might require significant modifications...

Since std::string_view cannot be implicitly converted to std::string, and I want to implement RESP_OK as a constexpr variable, I changed it to use constexpr char[].

mapleFU:
>  Since std::string_view cannot be implicitly converted to std::string, and I want to implement RESP_OK as a constexpr variable, I changed it to use constexpr char[].

This not fixes the problem, we can either use `const static inline std::string`, or just cast once on "Reply". `const char[]` to `std::string` also applies copy...

RiversJin:
> > Since std::string_view cannot be implicitly converted to std::string, and I want to implement RESP_OK as a constexpr variable, I changed it to use constexpr char[].由于 std::string_view 无法隐式转换为 std::string,而且我想将 RESP_OK 实现为 constexpr 变量,因此我将其改为使用 constexpr char[]。
> 
> This not fixes the problem, we can either use `const static inline std::string`, or just cast once on "Reply". `const char[]` to `std::string` also applies copy...这并不能解决问题,我们可以使用 `const static inline std::string` ，或者只在"Reply"上进行一次转换。 `const char[]` 对 `std::string` 也适用于复制...

Ok u're right, I've done

src/server/redis_reply.h
```diff
@@ -39,7 +40,9 @@ namespace redis {
 enum class RESP { v2, v3 };
 
 void Reply(evbuffer *output, const std::string &data);
-std::string SimpleString(const std::string &data);
+std::string SimpleString(std::string_view data);
+
+static inline std::string RESP_OK = RESP_PREFIX_SIMPLE_STRING "OK" CRLF;
```
mapleFU: Should also being const?

RiversJin: sorry hhh

mapleFU (approved):

PragmaTwice (approved):

Issue #2521, opened 2024-09-04
Improve the code style by introducing macros


git-hulk:
Followup code improvements mentioned in PR #2508. Currently, we check and return error if the status is not ok in many many places, which can be improved by introducing macros.

The naming candidates can be:

- RETURN_NOT_OK
- RETURN_IF_ERROR

and for converting the rocksdb status if there is an error:

- RETURN_NOT_OK_FROM_ROCKSDB
- RETURN_IF_ERR_FROM_ROCKSDB

_Originally posted by @mapleFU in https://github.com/apache/kvrocks/pull/2508#pullrequestreview-2276461263_

git-hulk:
@AntiTopQuark Would you like to take this issue?

AntiTopQuark:
> @AntiTopQuark Would you like to take this issue?

I’d be happy to take on this issue.  ：）

PragmaTwice:
The first macro is duplicated with the current GET_OR_RET macro.

PragmaTwice:
For the second I think it depends on the context, e.g. rocksdb::Status -> Status? rocksdb::Status -> rocksdb::Status?

We can extend GET_OR_RET for the latter, and the former doesn't require new macro.

git-hulk:
> For the second I think it depends on the context, e.g. rocksdb::Status -> Status? rocksdb::Status -> rocksdb::Status?
> 
> We can extend GET_OR_RET for the latter, and the former doesn't require new macro.

Yes, my bad. GET_OR_RET allows ignoring the return value, so the former one is unnecessary.

PR #2629, opened 2024-10-30
chore: Improve the code style by introducing macros 

AntiTopQuark:
Relate Issue: https://github.com/apache/kvrocks/issues/2521

Introduced two macros for handling error codes: `RETURN_IF_ERROR` and `RETURN_IF_ERR_FROM_ROCKSDB`. 

I haven't modified all the code yet, so could the project committer please review this code first to see if it’s acceptable? 
If approved, I'll go through and update the other files as well.


PragmaTwice:
`GET_OR_RET` is already a generalized macro for both `Status` and `StatusOr<T>`.

You can refer to my comment https://github.com/apache/kvrocks/issues/2521#issuecomment-2335096655.

src/common/status.h
```diff
@@ -376,3 +377,21 @@ struct [[nodiscard]] StatusOr {
...
+  ({                                                                                 \
+    auto&& status = (expr);                                                          \
+    if (!status.ok()) {                                                              \
+      if (with_status_string) {                                                      \
+        return {error_code, fmt::format(format_str, status.ToString())};             \
+      } else {                                                                       \
+        return {error_code, fmt::format(format_str)};                                \
+      }                                                                              \
+    }                                                                                \
+  })
```
PragmaTwice: Several points:
- Many code of this macro can be implemented by a ctor `Status(rocksdb::Status&&)`, we don't need write them in the macro;
- After such ctor is impl, this macro is useless since it also can be handled by `GET_OR_RET`;
- This macro doesn't explicit specify the output type, which can trigger some weird issues (the return type can actually be not a `Status`, but `rocksdb::Status`, or even `pair<int, string>`).

PragmaTwice (changes requested):
see above.

PragmaTwice:
You can check #2630 about how to implement `rocksdb::Status` forwarding without introducing a new macro.

src/common/status.h
```diff
@@ -397,3 +397,16 @@ bool StatusIsOK(const T& v) {
     if (!StatusIsOK(status)) return std::forward<decltype(status)>(status); \
     std::forward<decltype(status)>(status);                                 \
   }))
+
+// NOLINTNEXTLINE
+#define RETURN_IF_ERROR(...)                                                \
```
PragmaTwice, line 402: Hmmm I don't think you've got my point. As I've said, this macro is useless and can be replaced by GET_OR_RET. Actually it's function is a subset of GET_OR_RET.

src/common/status.h
```diff
@@ -397,3 +397,16 @@ bool StatusIsOK(const T& v) {
...
+  ({                                                                        \
+    auto&& status = (__VA_ARGS__);                                          \
+    if (!StatusIsOK(status)) return std::forward<decltype(status)>(status); \
+  })
+
+// NOLINTNEXTLINE
+#define RETURN_IF_ERR_FROM_ROCKSDB(s, code, msg) \
```
PragmaTwice, line 409: As I already said, we only need a function/method/ctor/operator T, instead of a macro. If we can complete some task via function/method, please never introduce a new macro.

src/common/status.h
AntiTopQuark: > Hmmm I don't think you've got my point. As I've said, this macro is useless and can be replaced by GET_OR_RET. Actually it's function is a subset of GET_OR_RET.

Firstly, Thank you very much for reviewing my submitted PR and providing valuable feedback! I truly appreciate your input and am grateful for the time you’ve spent helping me improve the code.


This macro can indeed be replaced by `GET_OR_RET`.

src/common/status.h
AntiTopQuark: > As I already said, we only need a function/method/ctor/operator T, instead of a macro. If we can complete some task via function/method, please never introduce a new macro.

Regarding the comment, after careful consideration, I realize that I may not have fully grasped your intention. Specifically, I feel that without introducing a new macro, it becomes challenging to directly return a new Status object.

AntiTopQuark:
I apologize, but upon reflection, I believe this PR offers minimal value. Additionally, due to my current busy schedule and limited energy, I am unable to continue this work. I apologize once again.


cc @PragmaTwice

PR #2630, opened 2024-10-31
feat(status): support rocksdb::Status fowarding in GET_OR_RET

PragmaTwice:
So now `Get_OR_RET` supports `StatusOr<T>`, `Status` and `rocksdb::Status`.

For conversions:
- `StatusOr<T>` and `Status` are designed to be converted mutually and implicitly.
- `rocksdb::Status` -> `Status` is currently not supported, it can be supported by ctors or functions.

### DIFF
--- a/src/common/status.h
+++ b/src/common/status.h
@@ -29,6 +29,7 @@
 #include <type_traits>
 #include <utility>
 
+#include "rocksdb/status.h"
 #include "type_util.h"
 
 class [[nodiscard]] Status {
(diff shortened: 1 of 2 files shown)


git-hulk (approved):

src/common/status.h
```diff
@@ -369,10 +370,30 @@ struct [[nodiscard]] StatusOr {
...
+}
+
 // NOLINTNEXTLINE
-#define GET_OR_RET(...)                                         \
-  ({                                                            \
-    auto&& status = (__VA_ARGS__);                              \
-    if (!status) return std::forward<decltype(status)>(status); \
-    std::forward<decltype(status)>(status);                     \
-  }).GetValue()
+#define GET_OR_RET(...)                                                     \
```
mapleFU, line 394: So this is `RET_NOT_OK` for `rocksdb::Status`?

PragmaTwice: Same as for `Status`.

PR #2643, opened 2024-11-03
fix(cluster): should remove the master if it's not node in the cluster

### LINKED ISSUE #2618: updating the topology information in kvrocks does not change the replication relationships
### Discussed in https://github.com/apache/kvrocks/discussions/2616

<sup>Originally posted by **quan-kuang** October 22, 2024</sup>
I have already removed the slave node using the setnodes command, but I can still see the instance information through the info
企业微信截图_17295785146549
How should I remove this master-slave relationship?

</div>

git-hulk:
Currently, the replica won't remove the master replication
while it's not a node in the cluster, which is an unexpected behavior
for the living master node in the cluster.

This fixes #2618.

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -75,9 +75,7 @@ Status Cluster::SetNodeId(const std::string &node_id) {
   }
 
   // Set replication relationship
-  if (myself_) return SetMasterSlaveRepl();
-
-  return Status::OK();
+  return SetMasterSlaveRepl();
 }
 
 // The reason why the new version MUST be +1 of current version is that,
@@ -204,11 +202,8 @@ Status Cluster::SetClusterNodes(const std::string &nodes_str, int64_t version, b
   }
 
   // Set replication relationship
-  if (myself_) {
-    s = SetMasterSlaveRepl();
-    if (!s.IsOK()) {
-      return s.Prefixed("failed to set master-replica replication");
-    }
+  if (auto s = SetMasterSlaveRepl(); !s.IsOK()) {
+    return s.Prefixed("failed to set master-replica replication");
   }
 
   // Clear data of migrated slots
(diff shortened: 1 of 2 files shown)


mapleFU:
When would "master is not a node in cluster" happens?

git-hulk:
> When would "master is not a node in cluster" happens?

oh, this issue is caused by the replica instead of the master was removed in the cluster, but its replication is still living.

caipengbo (approved):

src/cluster/cluster.cc
```diff
@@ -204,11 +202,8 @@ Status Cluster::SetClusterNodes(const std::string &nodes_str, int64_t version, b
...
-  if (myself_) {
-    s = SetMasterSlaveRepl();
-    if (!s.IsOK()) {
-      return s.Prefixed("failed to set master-replica replication");
-    }
+  if (auto s = SetMasterSlaveRepl(); !s.IsOK()) {
+    return s.Prefixed("failed to set master-replica replication");
   }
 
   // Clear data of migrated slots
```
mapleFU, line 209: should here check `myself_ == nullptr` and skip checking for:

```
  if (!migrated_slots_.empty()) {
    engine::Context ctx(srv_->storage);
    for (const auto &[slot, _] : migrated_slots_) {
      if (slots_nodes_[slot] != myself_) {
        auto s = srv_->slot_migrator->ClearKeysOfSlotRange(ctx, kDefaultNamespace, SlotRange::GetPoint(slot));
        if (!s.ok()) {
          LOG(ERROR) << "failed to clear data of migrated slots: " << s.ToString();
        }
      }
    }
  }
```

mapleFU (approved):

src/cluster/cluster.cc
git-hulk: I prefer keeping the previous behavior here since it should clear the keys of migrating slots once removed from the cluster.

PR #2657, opened 2024-11-11
fix(stream): add `KeyRangeGen` for `XRead` and `XReadGroup`

### LINKED ISSUE #2625: Key range of XREAD/XREADGROUP is not correct
### Search before asking

### Version

unstable

### Minimal reproduce step

https://github.com/apache/kvrocks/blob/5fccd57c66bc8f0e4138924fe6a0148931e5650f/src/commands/cmd_stream.cc#L1891-L1892

We need to define a KeyRangeGen function to make it produce correct key range.

--

### What did you see instead?

--

### Anything Else?

_No response_

LindaSummer:
# Issue

Close #2625 .

# Proposed Changes

- Refactor `Parse` function for `XRead` and `XReadGroup`.
- Add `KeyRangeGen` for `XRead` and `XReadGroup`.

### DIFF
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -38,7 +38,7 @@ namespace {
 CommandKeyRange ParseStreamReadRange(const std::vector<std::string> &args) {
   // assert here we must have a stream in args since it has been parsed.
   auto stream_keyword_iter =
-      std::find_if(args.cbegin(), args.cend(), [](const std::string &arg) { return util::ToLower(arg) == "streams"; });
+      std::find_if(args.cbegin(), args.cend(), [](const std::string &arg) { return util::EqualICase(arg, "streams"); });
   int stream_pos = static_cast<int>(std::distance(args.cbegin(), stream_keyword_iter));
   int stream_size = static_cast<int>(args.size() - stream_pos) / 2;
 
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -35,10 +35,10 @@
 namespace redis {
 namespace {
 // for XRead and XReadGroup stream range parse.
-CommandKeyRange ParseStreamReadRange(const std::vector<std::string> &args) {
+CommandKeyRange ParseStreamReadRange(const std::vector<std::string> &args, uint32_t start_offset) {
   // assert here we must have a stream in args since it has been parsed.
-  auto stream_keyword_iter =
-      std::find_if(args.cbegin(), args.cend(), [](const std::string &arg) { return util::EqualICase(arg, "streams"); });
+  auto stream_keyword_iter = std::find_if(std::next(args.cbegin(), start_offset), args.cend(),
+                                          [](const std::string &arg) { return util::EqualICase(arg, "streams"); });
   int stream_pos = static_cast<int>(std::distance(args.cbegin(), stream_keyword_iter));
   int stream_size = static_cast<int>(args.size() - stream_pos) / 2;
 
@@ -1420,7 +1420,9 @@ class CommandXRead : public Commander,
     bufferevent_enable(bev, EV_READ);
   }
 
-  static const inline CommandKeyRangeGen keyRangeGen = ParseStreamReadRange;
+  static const inline CommandKeyRangeGen keyRangeGen = [](const std::vector<std::string> &args) {
+    return ParseStreamReadRange(args, 0);
+  };
 
  private:
   std::vector<std::string> streams_;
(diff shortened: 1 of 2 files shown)


src/commands/cmd_stream.cc
```diff
@@ -1715,6 +1684,27 @@ class CommandXReadGroup : public Commander,
     bufferevent_enable(bev, EV_READ);
   }
 
+  static CommandKeyRangeGen KeyRangeGen() {
+    static constexpr auto keyRangeGenFunc = [](const std::vector<std::string> &args) {
+      auto parsed_args = *parse(args);
```
PragmaTwice: Hmm actually from my side I don't think we need to reuse the code of `Parse()`, since:
- many fields in `parsed_args` is useless in key range gen, so we made lots of useless string copies.
- in key range gen we can assume that the syntax of the input args is valid, so we can omit many checks.

src/commands/cmd_stream.cc
```diff
@@ -1715,6 +1684,27 @@ class CommandXReadGroup : public Commander,
     bufferevent_enable(bev, EV_READ);
   }
 
+  static CommandKeyRangeGen KeyRangeGen() {
+    static constexpr auto keyRangeGenFunc = [](const std::vector<std::string> &args) {
```
PragmaTwice:
```suggestion
  static CommandKeyRange KeyRangeGen(const std::vector<std::string> &args) {
```

Also we don't need to make a function that returns a function here.

src/commands/cmd_stream.cc
LindaSummer: Hi @PragmaTwice ,

Thanks for your suggestions! 😊

I have reverted the original `Parse` function and add a simple search function for parsing stream key counts.

Best Regards,
Edward

src/commands/cmd_stream.cc
LindaSummer: Hi @PragmaTwice ,

Got it!

I have changed it to a static variable. 😊

Best Regards,
Edward

src/commands/cmd_stream.cc
```diff
@@ -33,6 +33,22 @@
 #include "types/redis_stream.h"
 
 namespace redis {
+namespace {
+// for XRead and XReadGroup stream range parse.
+CommandKeyRange ParseStreamReadRange(const std::vector<std::string> &args) {
+  // assert here we must have a stream in args since it has been parsed.
+  auto stream_keyword_iter =
+      std::find_if(args.cbegin(), args.cend(), [](const std::string &arg) { return util::ToLower(arg) == "streams"; });
```
PragmaTwice:
```suggestion
  auto stream_keyword_iter =
      std::find_if(args.cbegin(), args.cend(), [](const std::string &arg) { return util::EqualICase(arg, "streams"); });
```

LindaSummer: Hi @PragmaTwice ,

Thanks very much for your suggestion! 😊

I have replaced the comparison operator to `util::EqualICase`.

Best Regards,
Edward

src/commands/cmd_stream.cc
```diff
@@ -33,6 +33,22 @@
 #include "types/redis_stream.h"
 
 namespace redis {
+namespace {
+// for XRead and XReadGroup stream range parse.
+CommandKeyRange ParseStreamReadRange(const std::vector<std::string> &args) {
+  // assert here we must have a stream in args since it has been parsed.
+  auto stream_keyword_iter =
+      std::find_if(args.cbegin(), args.cend(), [](const std::string &arg) { return util::EqualICase(arg, "streams"); });
```
PragmaTwice: For XREADGROUP, I think we need to find from `begin + 4`, to avoid a group or consumer named `streams`.

refer to https://redis.io/docs/latest/commands/xreadgroup/ .

LindaSummer: Hi @PragmaTwice ,

Thanks very much for your review! 😊

I have updated the `ParseStreamReadRange` with a begin offset.

Best Regards,
Edward

PragmaTwice (commented):
The code looks good to me.

Would you like to write a go test case that calls redis command `COMMAND GETKEYS XREAD/XREADGROUP ...` to confirm the returned keys is correct?

You can add a new test like https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/command/command_test.go#L125.

LindaSummer:
> The code looks good to me.
> 
> 
> 
> Would you like to write a go test case that calls redis command `COMMAND GETKEYS XREAD/XREADGROUP ...` to confirm the returned keys is correct?
> 
> 
> 
> You can add a new test like https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/command/command_test.go#L125.

Of course! 

I will add unittest later today.😄

Best Regards,
Edward

LindaSummer:
Hi @PragmaTwice ,

I have added related unit tests for different combinations of `XREAD` and `XREADGROUP`. 😊

Best Regards,
Edward

PragmaTwice (approved):

PR #2658, opened 2024-11-11
feat(config): make rocksdb.dump_malloc_stats configurable

nathanlo-hrt:
This causes a lot of logging: I believe this should be configurable.


src/config/config.cc
```diff
@@ -258,6 +258,7 @@ Config::Config() {
                                                rocksdb::CompressionType::kNoCompression)},
       {"rocksdb.wal_ttl_seconds", true, new IntField(&rocks_db.wal_ttl_seconds, 3 * 3600, 0, INT_MAX)},
       {"rocksdb.wal_size_limit_mb", true, new IntField(&rocks_db.wal_size_limit_mb, 16384, 0, INT_MAX)},
+      {"rocksdb.dump_malloc_stats", false, new YesNoField(&rocks_db.dump_malloc_stats, true)},
```
```suggestion
      {"rocksdb.dump_malloc_stats", true, new YesNoField(&rocks_db.dump_malloc_stats, true)},
```

If it's not read-only, we need to provide a callback.

PragmaTwice (approved):

PR #2661, opened 2024-11-14
fix(cmd): args should be parsed before retrieving keys in COMMAND GETKEYS

PragmaTwice:
For commands that CommandKeyRangeGen or CommandKeyRangeVecGen is used, args SHOULD be parsed before we call the key range gen function to avoid undefined behaviors.


git-hulk (dismissed):
Good catch

git-hulk (approved):

PR #2662, opened 2024-11-14
fix(replication): slave blocks until keepalive timer is reached when master is gone without fin/rst notification

sryanyuan:
If the master is lost, the replication thread will block until the keepalive timer is reached when receiving full-sync SST files.  At the same time, if we execute a 'clusterx setnodes' command, it will hold an exclusive lock until the replication thread is stopped. This will cause all other worker threads to block.

My solution is to enable the socket read timeout on the file descriptor receiving SST files, if a timeout occurs and the replication thread is marked as stopped, the receiving action will be broken.


src/cluster/replication.cc
```diff
@@ -770,14 +770,16 @@ Status ReplicationThread::parallelFetchFile(const std::string &dir,
           if (this->stop_flag_) {
             return {Status::NotOK, "replication thread was stopped"};
           }
+          int sock_timeout_ms = 3000;
```
torwig: I think it's better to move it to the top of the file and make it const/constexpr.

sryanyuan: ok, updated

src/cluster/replication.cc
```diff
@@ -770,14 +770,16 @@ Status ReplicationThread::parallelFetchFile(const std::string &dir,
...
           ssl_st *ssl = nullptr;
 #ifdef ENABLE_OPENSSL
           if (this->srv_->GetConfig()->tls_replication) {
             ssl = SSL_new(this->srv_->ssl_ctx.get());
           }
           auto exit = MakeScopeExit([ssl] { SSL_free(ssl); });
 #endif
-          int sock_fd = GET_OR_RET(util::SockConnect(this->host_, this->port_, ssl).Prefixed("connect the server err"));
+          int sock_fd = GET_OR_RET(util::SockConnect(this->host_, this->port_, ssl, sock_timeout_ms, sock_timeout_ms)
```
PragmaTwice: - Why have the connection timeout and the timeout both been set to the same value?
- Also why is the timeout set to 3000 here?
- Should we make this configurable?

src/cluster/replication.cc
```diff
@@ -874,6 +876,12 @@ Status ReplicationThread::fetchFile(int sock_fd, evbuffer *evbuf, const std::str
     UniqueEvbufReadln line(evbuf, EVBUFFER_EOL_CRLF_STRICT);
     if (!line) {
       if (auto s = util::EvbufferRead(evbuf, sock_fd, -1, ssl); !s) {
+        if (errno == EAGAIN || errno == EWOULDBLOCK) {
+          if (stop_flag_) {
+            return {Status::NotOK, "replication thread was stopped"};
+          }
+          continue;
+        }
```
PragmaTwice: I think you can notice that to enter this block, the status must be OK (i.e. `if (auto s = ...; !s)`).

It means that, `evbuffer_read` must have a positive integer as a return value to indicate that the operation is successful. So I highly suspect that the errno will not be either of these two.

sryanyuan: Our QA team tested and it do resolve the issue (On X86-64 centos).

It depends on the return value of evbuffer_read, if it returns 0, it means the connection is EOF, or -1 if error occurred. So timeout will always return Status with NotOK and errno will be EAGAIN or EWOULDBLOCK as manual says:

if no data has been transferred and the timeout has been reached then -1 is returned with errno set to EAGAIN or EWOULDBLOCK, or EINPROGRESS (for [connect](https://linux.die.net/man/2/connect)(2)) just as if the socket was specified to be nonblocking.

src/cluster/replication.cc
git-hulk: I suggest separating those two default values due to their backoff interval is different. For the TCP `syn` packet loss retry in Linux is `1s`, `2s`, `4s`, `8s`, so a value like `3200ms` will be better than `3000ms` because it would retry twice if it's loss due to high traffic load.

And for the read/write backoff is related to RTO, but it's generally around 200ms in the internal network, so its backoff would be `200ms`, `400ms`, `800ms`, `1600ms`, and it will be better to set it to a value like `3100ms`.

src/cluster/replication.cc
sryanyuan: The other errors (not timeout), just returned as previous

sryanyuan: To enter this block, evbuffer_read should returns a non-positive integer to indicate the operation is not successful, so it only enters when the read operation fails.

If it fails, check the errno then we can retry the operation if timed out or quit if replication thread is stopped.

src/cluster/replication.cc
sryanyuan: > I suggest separating those two default values due to their backoff interval is different. For the TCP `syn` packet loss retry in Linux is `1s`, `2s`, `4s`, `8s`, so a value like `3200ms` will be better than `3000ms` because it would retry twice if it's loss due to high traffic load.
> 
> And for the read/write backoff is related to RTO, but it's generally around 200ms in the internal network, so its backoff would be `200ms`, `400ms`, `800ms`, `1600ms`, so it will be better to set it to a value like `3100ms`.

updated

src/cluster/replication.cc
PragmaTwice: Ahh you are right, sorry. It looks good to me now.

src/cluster/replication.cc
sryanyuan: > * Why have the connection timeout and the timeout both been set to the same value?
> * Also why is the timeout set to 3000 here?
> * Should we make this configurable?

I will make timeout config fields configurable, and using the default value suggested by hulk :)

kvrocks.conf
```diff
@@ -174,6 +174,20 @@ slave-read-only yes
...
+# If the master is unreachable before connecting, not having a timeout may block future
+# 'clusterx setnodes' commands because the replication thread is blocked on connect.
+slave-fullsync-connect-timeout 3100
+
+# Change the default timeout in milliseconds for socket recv during fullsync.
+# The default value is 3200, and 0 means no timeout.
+#
+# If the master is unreachable when fetching SST files, not having a timeout may block
+# future 'clusterx setnodes' commands because the replication thread is blocked on recv.
+slave-fullsync-recv-timeout 3200
```
PragmaTwice: There's also another socket connection in `ReplicationThread::CallbacksStateMachine`. Could we unify these connection timeout settings in replication and name this config option `replication-..-timeout`? cc @git-hulk

PragmaTwice: For connection timeout I think it's good to share the config. Not sure if it's fine for read/write timeout. WDYT?

kvrocks.conf
```diff
@@ -174,6 +174,20 @@ slave-read-only yes
 # By default the priority is 100.
 slave-priority 100
 
+# Change the default timeout in milliseconds for socket connect during fullsync.
+# The default value is 3100, and 0 means no timeout.
+#
+# If the master is unreachable before connecting, not having a timeout may block future
+# 'clusterx setnodes' commands because the replication thread is blocked on connect.
+slave-fullsync-connect-timeout 3100
```
mapleFU: Nit: we need to denote it's microsec / millsecs or other in comment

PragmaTwice: Yeah we can rename it with a `-ms` suffix.

sryanyuan: > Yeah we can rename it with a `-ms` suffix.

ok

kvrocks.conf
sryanyuan: The main replication thread is also using a blocking socket for receiving data, so it may block if the master is unreachable. I will test this case and modify it if the situation exists.

git-hulk: > There's also another socket connection in ReplicationThread::CallbacksStateMachine. Could we unify these connection timeout settings in replication and name this config option replication-..-timeout? cc @git-hulk

Yes, it would be better to use a unify connection timeout for replication sockets.

sryanyuan: The main replication thread seems to be using a blocking socket wrapped by bufferevent for receiving data, so it may block if master is unreachable. I will test this case and modify it if the situation exists.

git-hulk: The replication socket is a blocking socket, but it won't block forever if we have set the timeout when connecting.

mapleFU:
By the way, can we add a test with master removed and about 5s' connection timeout in go integration?

src/cluster/replication.cc
```diff
@@ -874,6 +877,12 @@ Status ReplicationThread::fetchFile(int sock_fd, evbuffer *evbuf, const std::str
     UniqueEvbufReadln line(evbuf, EVBUFFER_EOL_CRLF_STRICT);
     if (!line) {
       if (auto s = util::EvbufferRead(evbuf, sock_fd, -1, ssl); !s) {
+        if (errno == EAGAIN || errno == EWOULDBLOCK) {
```
git-hulk: This check should be unnecessary since the replication socket is in blocking mode, and it will never return the EAGAIN which is for the non-blocking socket.

kvrocks.conf
sryanyuan: I unified the connection timeout config. Please have a look.

src/cluster/replication.cc
sryanyuan: > This check should be unnecessary since the replication socket is in blocking mode, and it will never return the EAGAIN which is for the non-blocking socket.

According to the man page:

SO_RCVTIMEO and SO_SNDTIMEO
Specify the receiving or sending timeouts until reporting an error. The argument is a struct timeval. If an input or output function blocks for this period of time, and data has been sent or received, the return value of that function will be the amount of data transferred; **if no data has been transferred and the timeout has been reached then -1 is returned with errno set to EAGAIN or EWOULDBLOCK, or EINPROGRESS (for [connect](https://linux.die.net/man/2/connect)(2)) just as if the socket was specified to be nonblocking.** If the timeout is set to zero (the default) then the operation will never timeout. Timeouts only have effect for system calls that perform socket I/O (e.g., [read](https://linux.die.net/man/2/read)(2), [recvmsg](https://linux.die.net/man/2/recvmsg)(2), [send](https://linux.die.net/man/2/send)(2), [sendmsg](https://linux.die.net/man/2/sendmsg)(2)); timeouts have no effect for [select](https://linux.die.net/man/2/select)(2), [poll](https://linux.die.net/man/2/poll)(2), [epoll_wait](https://linux.die.net/man/2/epoll_wait)(2), and so on.

sryanyuan: We should check the timeout error because if the master sends SST files with low rate, then it can sleep a long time to send the next SST file.

git-hulk (dismissed):

sryanyuan:
We can't distinguish whether the result is EOF or error if EvbufferRead returns an error. When the underlying I/O syscall returns EOF, the errno will not be set. So, I added a new EOF status to break the loop if the connection is EOF.

https://github.com/apache/kvrocks/pull/2662/commits/434f1480476458875aa2c70fcac78d62f409efd9

@git-hulk @PragmaTwice please have a look

git-hulk (dismissed):

src/common/status.h
```diff
@@ -75,6 +75,9 @@ class [[nodiscard]] Status {
     // Search
     NoPrefixMatched,
     TypeMismatched,
+
+    // Network
+    NetworkEOF,
```
```suggestion
    EndOfFile,
    TryAgain,
```

src/common/io_util.cc
```diff
@@ -515,7 +515,8 @@ StatusOr<int> EvbufferRead(evbuffer *buf, evutil_socket_t fd, int howmuch, [[may
   if (int ret = evbuffer_read(buf, fd, howmuch); ret > 0) {
     return ret;
   } else {
-    return {Status::NotOK, fmt::format("failed to read from socket: {}", strerror(errno))};
+    return ret == 0 ? Status{Status::NetworkEOF, "EOF"}
+                    : Status{Status::NotOK, fmt::format("failed to read from socket: {}", strerror(errno))};
   }
```
PragmaTwice:
```c++
  if (int ret = evbuffer_read(buf, fd, howmuch); ret > 0) {
    return ret;
  } else if (ret == 0) {
    return {Status::EndOfFile, "EOF encountered while reading from socket"};
  } else {
    return {(errno == EAGAIN || errno == EWOULDBLOCK ? Status::TryAgain, Status::NotOK), 
            fmt::format("failed to read from socket: {}", strerror(errno))};
  }
```

src/common/io_util.cc
```diff
@@ -515,7 +515,8 @@ StatusOr<int> EvbufferRead(evbuffer *buf, evutil_socket_t fd, int howmuch, [[may
   if (int ret = evbuffer_read(buf, fd, howmuch); ret > 0) {
```
PragmaTwice, line 520: Also we need to add it for the SSL part above.

sryanyuan: ok, i will modify it later

src/cluster/replication.cc
```diff
@@ -874,6 +876,12 @@ Status ReplicationThread::fetchFile(int sock_fd, evbuffer *evbuf, const std::str
     UniqueEvbufReadln line(evbuf, EVBUFFER_EOL_CRLF_STRICT);
     if (!line) {
       if (auto s = util::EvbufferRead(evbuf, sock_fd, -1, ssl); !s) {
+        if (s.GetCode() != Status::NetworkEOF && (errno == EAGAIN || errno == EWOULDBLOCK)) {
```
```suggestion
        if (s.Is<Status::TryAgain>())) {
```

src/common/io_util.cc
```diff
@@ -502,7 +502,11 @@ StatusOr<int> EvbufferRead(evbuffer *buf, evutil_socket_t fd, int howmuch, [[may
       howmuch = BUFFER_SIZE;
     }
     if (howmuch = SSL_read(ssl, tmp, howmuch); howmuch <= 0) {
-      return {Status::NotOK, fmt::format("failed to read from SSL connection: {}", fmt::streamed(SSLError(howmuch)))};
+      if (howmuch == 0) {
+        return {Status::EndOfFile, "EOF encountered while reading from SSL connection"};
+      }
+      return {(errno == EWOULDBLOCK || errno == EAGAIN) ? Status::TryAgain : Status::NotOK,
+              fmt::format("failed to read from SSL connection: {}", fmt::streamed(SSLError(howmuch)))};
```
PragmaTwice:
```suggestion
      int err = SSL_get_error(ssl, howmuch);
      if (err == SSL_ERROR_ZERO_RETURN) {
        return {Status::EndOfFile, "EOF encountered while reading from SSL connection"};
      }
      return {(err == SSL_ERROR_WANT_READ) ? Status::TryAgain : Status::NotOK,
              fmt::format("failed to read from SSL connection: {}", fmt::streamed(SSLError(howmuch)))};
```

PragmaTwice (dismissed):

git-hulk (dismissed):

src/config/config.h
```diff
@@ -105,6 +105,8 @@ struct Config {
   bool slave_serve_stale_data = true;
   bool slave_empty_db_before_fullsync = false;
   int slave_priority = 100;
+  int replication_connect_timeout_ms = 3200;
```
torwig: Connect is 3100 and receive is 3200. It may not be a big issue here but for convenience. 
BTW, you can make them constants since you repeated them twice (and made an error here).

src/config/config.h
```diff
@@ -105,6 +105,8 @@ struct Config {
...
+  int replication_recv_timeout_ms = 3100;
```
PragmaTwice:
```suggestion
  int replication_connect_timeout_ms = 3100;
  int replication_recv_timeout_ms = 3200;
```

torwig (approved):

PragmaTwice (approved):

PR #2664, opened 2024-11-15
refactor: initialize google logger after parsing options

aleksraiden:
At startup we need to init a logger after fully initialising config and parsing command line override options. I think this way is clearer and in some cases Google Logger can't apply options from flag after calling google::InitGoogleLogging.

### DIFF
--- a/src/cli/main.cc
+++ b/src/cli/main.cc
@@ -116,9 +116,6 @@ static void InitGoogleLog(const Config *config) {
 int main(int argc, char *argv[]) {
   srand(static_cast<unsigned>(util::GetTimeStamp()));
 
-  google::InitGoogleLogging("kvrocks");
-  auto glog_exit = MakeScopeExit(google::ShutdownGoogleLogging);
-
   evthread_use_pthreads();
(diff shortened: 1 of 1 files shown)


mapleFU:
>  and in some cases Google Logger can't apply options from flag after calling google::InitGoogleLogging.

Just curious is there any doc for this behavior?

aleksraiden:
> > and in some cases Google Logger can't apply options from flag after calling google::InitGoogleLogging.
> 
> Just curious is there any doc for this behavior?

I'm not a familiar with c+/glog, but I interpret this moment from docs (https://google.github.io/glog/0.8.0/flags/#using-command-line-parameters-and-environment-variables):

`
You can also modify flag values in your program by modifying global variables FLAGS_*. Most settings start working immediately after you update FLAGS_*. The exceptions are the flags related to destination files. For instance, you might want to set FLAGS_log_dir before calling google::InitGoogleLogging.
`

torwig (approved):

git-hulk (approved):

PragmaTwice (approved):

PR #2683, opened 2024-12-01
feat: skip block cache deallocation to make shutdown fast

wanghenshui:
when blockcache size is too big, skip dtor could save shutdown time

ref https://groups.google.com/g/rocksdb/c/xQ_o9jWoFqg

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -644,8 +644,8 @@ migrate-batch-size-kb 16
 migrate-batch-rate-limit-mb 16
 
 
-# Specifies fast shutdown mode
-# If set to yes, we skip blockcache dtor to speed up shutdown
+# If it is set to yes, kvrocks will skip the deallocation of block cache
+# while closing the database to speed up the shutdown
 #
 # Default: no
 # skip-block-cache-deallocation-on-close no
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -643,6 +643,13 @@ migrate-batch-size-kb 16
 # Default: 16M
 migrate-batch-rate-limit-mb 16
 
+
+# Specifies fast shutdown mode
+# If set to 1, we skip blockcache dtor to speed up shutdown
+#
+# Default: 0
+# fast_shutdown 0
+
 ################################ ROCKSDB #####################################
 
 # Specify the capacity of column family block cache. A larger block cache
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -648,7 +648,7 @@ migrate-batch-rate-limit-mb 16
 # If set to 1, we skip blockcache dtor to speed up shutdown
 #
 # Default: 0
-# fast_shutdown 0
+# skip-block-cache-deallocation-on-close 0
 
 ################################ ROCKSDB #####################################
 
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -645,10 +645,10 @@ migrate-batch-rate-limit-mb 16
 
 
 # Specifies fast shutdown mode
-# If set to 1, we skip blockcache dtor to speed up shutdown
+# If set to yes, we skip blockcache dtor to speed up shutdown
 #
-# Default: 0
-# skip-block-cache-deallocation-on-close 0
+# Default: no
+# skip-block-cache-deallocation-on-close no
 
 ################################ ROCKSDB #####################################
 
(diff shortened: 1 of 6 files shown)


torwig:
@wanghenshui You can run `./x.py format` before committing changes to format the code.

git-hulk (dismissed):

src/storage/storage.cc
```diff
@@ -98,6 +98,7 @@ void Storage::CloseDB() {
   db_->SyncWAL();
   rocksdb::CancelAllBackgroundWork(db_.get(), true);
   for (auto handle : cf_handles_) db_->DestroyColumnFamilyHandle(handle);
+  shared_block_cache_->DisownData();
```
mapleFU: Should we add another flag for this? Since I don't know CloseDb in testing or other case would checking leak

wanghenshui: ok then.

`fast_shutdown`?

mapleFU: this is ok to me

kvrocks.conf
```diff
@@ -643,6 +643,13 @@ migrate-batch-size-kb 16
 # Default: 16M
 migrate-batch-rate-limit-mb 16
 
+
+# Specifies fast shutdown mode
+# If set to 1, we skip blockcache dtor to speed up shutdown
+#
+# Default: 0
+# fast_shutdown 0
```
```suggestion
# skip-block-cache-deallocation-on-close no
```

PragmaTwice (commented):
Also we need to add a go test case for it.

mapleFU (commented):
`PrepareRestoreDB()` also call `CloseDB`, should it prevent from unmanage memory?

wanghenshui:
> `PrepareRestoreDB()` also call `CloseDB`, should it prevent from unmanage memory?

yes. i 'll make a single function from `CloseDB`, not inside it

kvrocks.conf
```diff
@@ -643,6 +643,13 @@ migrate-batch-size-kb 16
 # Default: 16M
 migrate-batch-rate-limit-mb 16
 
+
+# Specifies fast shutdown mode
+# If set to yes, we skip blockcache dtor to speed up shutdown
```
```suggestion
# If it is set to yes, kvrocks will skip the deallocation of block cache
# while closing the database to speed up the shutdown
```

PragmaTwice (dismissed):

tests/gocase/unit/config/config_test.go
```diff
@@ -320,3 +320,25 @@ func TestGenerateConfigsMatrix(t *testing.T) {
...
+	require.EqualValues(t, "yes", val["skip-block-cache-deallocation-on-close"])
+
+	// default value "no"
+	srv1 := util.StartServer(t, map[string]string{})
+	defer srv1.Close()
+
+	rdb = srv1.NewClient()
+	val = rdb.ConfigGet(ctx, "skip-block-cache-deallocation-on-close").Val()
+	require.EqualValues(t, "no", val["skip-block-cache-deallocation-on-close"])
+}
```
git-hulk:
```suggestion
func TestGetConfigSkipBlockCacheDeallocationOnClose(t *testing.T) {
	srv := util.StartServer(t, map[string]string{
		"skip-block-cache-deallocation-on-close": "yes",
	})
	defer srv.Close()

	ctx := context.Background()
	rdb := srv.NewClient()
	defer func() { require.NoError(t, rdb.Close()) }()
	val := rdb.ConfigGet(ctx, "skip-block-cache-deallocation-on-close").Val()
	require.EqualValues(t, "yes", val["skip-block-cache-deallocation-on-close"])

	// default value "no"
	srv1 := util.StartServer(t, map[string]string{})
	defer srv1.Close()

	rdb = srv1.NewClient()
	val = rdb.ConfigGet(ctx, "skip-block-cache-deallocation-on-close").Val()
	require.EqualValues(t, "no", val["skip-block-cache-deallocation-on-close"])
}

```

git-hulk (dismissed):

src/storage/storage.cc
```diff
@@ -88,6 +88,13 @@ Storage::Storage(Config *config)
 Storage::~Storage() {
   DestroyBackup();
   CloseDB();
+  SkipBlockCacheDeallocationOnClose();
```
mapleFU: nit: you can change to `TrySkip`, since it may not stop this

wanghenshui: ok

git-hulk (approved):

mapleFU (approved):
General LGTM but maybe in the future we can have better method for this, I think this flag can be set only when `cli/main.cc`. But this is also ok to me

PR #2685, opened 2024-12-02
fix(worker): memory leak when Worker::AddConnection() failed

weim0000:
when new connection on the worker, if AddConnection function return the status is not OK, we need call 'delete conn' before 'return'， avoid memory leak

auto conn = new redis::Connection(bev, this);
s = AddConnection(conn);
if (!s.IsOK()) {
  conn->Close();
  **delete conn;**
  return;
}

### DIFF
--- a/src/server/worker.cc
+++ b/src/server/worker.cc
@@ -178,6 +178,7 @@ void Worker::newTCPConnection(evconnlistener *listener, evutil_socket_t fd, [[ma
       LOG(WARNING) << "Failed to send error response to socket: " << s.Msg();
     }
     conn->Close();
+    delete conn;
     return;
   }
 


torwig:
@weim0000 Thank you for your contribution. Please, run `./x.py format` to format the code accordingly.

PragmaTwice (approved):

PragmaTwice (changes requested):
Ahh sorry, I've checked the `Connection::Close()` and it will call `Worker::FreeConnection()` to `delete` itself.

So I think it should not be a problem here. cc @git-hulk  

And thus `delete` after `Close()` will cause a double-free.

weim0000:
> Ahh sorry, I've checked the `Connection::Close()` and it will call `Worker::FreeConnection()` to `delete` itself.
> 
> So I think it should not be a problem here. cc @git-hulk
> 
> And thus `delete` after `Close()` will cause a double-free.

oh, yes, you are right;  It is no problem, I'm sorry, I will close the pr

git-hulk:
Yes, the connection is registered in worker, so need to remove from the worker first and then free the connection.

PR #2686, opened 2024-12-02
fix(string): reset the value of expired key for SETRANGE cmd

weim0000:
reset the value of expired key to prevent the following parsing and calculation.
`127.0.0.1:6003> set k v
OK
127.0.0.1:6003> expire k 1
(integer) 1

after 2 seconds

127.0.0.1:6003> get k
(nil)
127.0.0.1:6003> setrange k 1 aa
(integer) 10
127.0.0.1:6003> get k
**(nil)**
127.0.0.1:6003> exists k
**(integer) 0**
127.0.0.1:6003>`

1 set value with expire time.
2 wait until the key expires
3 use setrange to this key. The cmd be successful, but get k return nil

**What did you expect to see?**

127.0.0.1:6003> set k v
OK
127.0.0.1:6003> expire k 1
(integer) 1
127.0.0.1:6003> get k
(nil)
127.0.0.1:6003> setrange k 0 aa
(integer) 10
127.0.0.1:6003> get k
**aa**
127.0.0.1:6003> exists k
**(integer) 1**
127.0.0.1:6003>

### DIFF
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -65,7 +65,9 @@ rocksdb::Status String::getRawValue(engine::Context &ctx, const std::string &ns_
 
   Metadata metadata(kRedisNone, false);
   Slice slice = *raw_value;
-  return ParseMetadataWithStats({kRedisString}, &slice, &metadata);
+  s = ParseMetadataWithStats({kRedisString}, &slice, &metadata);
+  if (!s.ok()) raw_value->clear();
+  return s;
 }
 
 rocksdb::Status String::getValueAndExpire(engine::Context &ctx, const std::string &ns_key, std::string *value,
@@ -297,7 +299,6 @@ rocksdb::Status String::SetRange(engine::Context &ctx, const std::string &user_k
     }
 
     Metadata metadata(kRedisString, false);
-    raw_value.clear();
     metadata.Encode(&raw_value);
   }
 
@@ -330,7 +331,6 @@ rocksdb::Status String::IncrBy(engine::Context &ctx, const std::string &user_key
   if (!s.ok() && !s.IsNotFound()) return s;
   if (s.IsNotFound()) {
     Metadata metadata(kRedisString, false);
-    raw_value.clear();
     metadata.Encode(&raw_value);
   }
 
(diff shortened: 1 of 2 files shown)


git-hulk:
@weim0000 Thanks for your fix, would you mind adding a test case to cover this? https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/type/strings/strings_test.go

weim0000:
> @weim0000 Thanks for your fix, would you mind adding a test case to cover this? https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/type/strings/strings_test.go

Ok, no problem. I will add the test cases ASAP

tests/cppunit/types/string_test.cc
```diff
@@ -250,6 +250,15 @@ TEST_F(RedisStringTest, SetRange) {
   string_->Get(*ctx_, key_, &value);
   EXPECT_EQ(16, value.size());
   auto s = string_->Del(*ctx_, key_);
+
+  string_->SetEX(*ctx_, key_, "test-value", 1);
+  sleep(2);
```
git-hulk, line 255:
```suggestion
  time.Sleep(2 * time.Second);
```

torwig: @git-hulk It's a C++ test not Go :)

git-hulk: My bad 😢

src/types/redis_string.cc
```diff
@@ -297,6 +297,7 @@ rocksdb::Status String::SetRange(engine::Context &ctx, const std::string &user_k
     }
 
     Metadata metadata(kRedisString, false);
+    raw_value.clear();
```
git-hulk: I think it'd be better to clear the raw_value inside `getRawValue` if the return status isn't ok. Because we call `getRawValue` many times and it's hard to keep having an eye on this. @torwig @PragmaTwice @mapleFU What do you think?

torwig: @git-hulk Valid point. Since the value has expired, we should return "nothing" (empty) to the caller of `getRawValue`.

PragmaTwice: Good to me. Actually refactor them to StatusOr can solve the problem, but I don't have much time to do that.

git-hulk (approved):

torwig (approved):

git-hulk:
@weim0000 Thanks for your fix.

PR #2688, opened 2024-12-06
feat(config): add the `partition_filters` option to disable the partitioned filters

### LINKED ISSUE #2676: Allow to disable the index partitioning in kvrocks.conf
### Discussed in https://github.com/apache/kvrocks/discussions/2675

<sup>Originally posted by **bartek-siudeja** November 22, 2024</sup>
kvrocks is quite impressive! Thanks for doing this. 

We are trying to build a KV store with extremely strict latency/throughput requirements. We are working with really large servers, with lots of memory. Our values are relatively big and keys are random and small. So it would make sense to store all indexes and big bloom filters in memory. It seems this can be achieved using `cache_index_and_filter_blocks: no` and `max_open_files: -1`. However, kvrocks enables partitioning (at least for filters, not sure about indexes):
https://github.com/apache/kvrocks/blob/2.10/src/storage/storage.cc#L132
And partitioned filters seem to always live inside block_cache, so they can be evicted.

Would it be possible to expose the partitioning option in the config file? Or maybe have a "meta option" to force filters/indexes to stay in memory. Along the same line bloom filter size could be increased from 10 to say 15 bits to really avoid hitting the disk at any cost. Really only when we are pretty much sure key is there, and value can be retrieved. But this option is also hardcoded (a line above filter partitioning).</div>

cxljs:
It closes #2676.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -131,8 +131,8 @@ Status SetRocksdbCompression(Server *srv, const rocksdb::CompressionType compres
   for (size_t i = compression_start_level; i < KVROCKS_MAX_LSM_LEVEL; i++) {
     compression_per_level_builder.emplace_back(compression_option);
   }
-  const std::string compression_per_level =
-      util::StringJoin(compression_per_level_builder, [](const auto &s) -> decltype(auto) { return s; }, ":");
+  const std::string compression_per_level = util::StringJoin(
+      compression_per_level_builder, [](const auto &s) -> decltype(auto) { return s; }, ":");
   return srv->storage->SetOptionForAllColumnFamilies("compression_per_level", compression_per_level);
 };
 
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -131,8 +131,8 @@ Status SetRocksdbCompression(Server *srv, const rocksdb::CompressionType compres
   for (size_t i = compression_start_level; i < KVROCKS_MAX_LSM_LEVEL; i++) {
     compression_per_level_builder.emplace_back(compression_option);
   }
-  const std::string compression_per_level = util::StringJoin(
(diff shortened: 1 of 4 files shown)


src/config/config.cc
```diff
@@ -131,8 +131,8 @@ Status SetRocksdbCompression(Server *srv, const rocksdb::CompressionType compres
   for (size_t i = compression_start_level; i < KVROCKS_MAX_LSM_LEVEL; i++) {
     compression_per_level_builder.emplace_back(compression_option);
   }
-  const std::string compression_per_level = util::StringJoin(
-      compression_per_level_builder, [](const auto &s) -> decltype(auto) { return s; }, ":");
+  const std::string compression_per_level =
+      util::StringJoin(compression_per_level_builder, [](const auto &s) -> decltype(auto) { return s; }, ":");
```
PragmaTwice: Seems we need to revert this change to make CI pass.

Note that in the CI the clang-format version is 14. Different versions may lead to different format results.

cxljs: Thanks, I'm doing it.

PragmaTwice (commented):
Please avoid introducing unrelated changes.

Also please ensure the version of clang-format you are using is 14. You can download it from [here](https://github.com/llvm/llvm-project/releases/tag/llvmorg-14.0.0).

PragmaTwice (changes requested):
See above.

cxljs:
> Please avoid introducing unrelated changes.
> 
> Also please ensure the version of clang-format you are using is 14. You can download it from [here](https://github.com/llvm/llvm-project/releases/tag/llvmorg-14.0.0).

thanks, I fixed it.

PragmaTwice (approved):
LGTM. Thank you for your contribution!

git-hulk (approved):

PR #2689, opened 2024-12-10
refactor: reduce memory copying during incremental synchronization

RiversJin:
Optimized the logic for handling Psync incremental data on replica nodes, reducing an unnecessary data copy and lowering the loop complexity in the corresponding logic.

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -30,13 +30,16 @@
 #include <atomic>
 #include <csignal>
 #include <future>
+#include <memory>
 #include <string>
+#include <string_view>
 #include <thread>
 
 #include "commands/error_constants.h"
 #include "event_util.h"
 #include "fmt/format.h"
 #include "io_util.h"
+#include "rocksdb/write_batch.h"
 #include "rocksdb_crc32c.h"
 #include "scope_exit.h"
 #include "server/redis_reply.h"
@@ -123,8 +126,8 @@ void FeedSlaveThread::loop() {
     // iter_ would be always valid here
     auto batch = iter_->GetBatch();
     if (batch.sequence != curr_seq) {
-      LOG(ERROR) << "Fatal error encountered, WAL iterator is discrete, some seq might be lost"
-                 << ", sequence " << curr_seq << " expected, but got " << batch.sequence;
+      LOG(ERROR) << "Fatal error encountered, WAL iterator is discrete, some seq might be lost" << ", sequence "
+                 << curr_seq << " expected, but got " << batch.sequence;
       Stop();
       return;
     }
@@ -470,8 +473,7 @@ ReplicationThread::CBState ReplicationThread::replConfReadCB(bufferevent *bev) {
   // on unknown option: first try without announce ip, if it fails again - do nothing (to prevent infinite loop)
   if (isUnknownOption(line.View()) && !next_try_without_announce_ip_address_) {
     next_try_without_announce_ip_address_ = true;
-    LOG(WARNING) << "The old version master, can't handle ip-address, "
-                 << "try without it again";
+    LOG(WARNING) << "The old version master, can't handle ip-address, " << "try without it again";
     // Retry previous state, i.e. send replconf again
     return CBState::PREV;
   }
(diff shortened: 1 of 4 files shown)


src/cluster/replication.cc
```diff
@@ -576,31 +576,35 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
...
+        if (bulk_string == "ping") {
           return CBState::AGAIN;
         }
+
+        rocksdb::WriteBatch batch(std::move(bulk_string));
+
+        auto s = storage_->ReplicaApplyWriteBatch(&batch);
+        if (!s.IsOK()) {
+          LOG(ERROR) << "[replication] CRITICAL - Failed to write batch to local, " << s.Msg() << ". batch: 0x"
+                     << util::StringToHex(bulk_string);
```
mapleFU: used-after-move?

RiversJin: Oops... 😅 It's been updated to use batch.Data().

src/cluster/replication.cc
```diff
@@ -576,31 +578,35 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
...
-            if (!s.IsOK()) {
-              LOG(ERROR) << "[replication] CRITICAL - failed to parse write batch 0x" << util::StringToHex(bulk_string)
-                         << ": " << s.Msg();
-              return CBState::RESTART;
-            }
-          }
-          evbuffer_drain(input, incr_bulk_len_ + 2);
-          incr_state_ = Incr_batch_size;
-        } else {
+        char* bulk_data = reinterpret_cast<char *>(evbuffer_pullup(input, static_cast<ssize_t>(incr_bulk_len_ + 2)));
```
mapleFU:
```suggestion
        const char* bulk_data = reinterpret_cast<const char *>(evbuffer_pullup(input, static_cast<ssize_t>(incr_bulk_len_ + 2)));
```

?

mapleFU (dismissed):

src/cluster/replication.cc
```diff
@@ -576,31 +578,35 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
...
+        std::string bulk_string = std::string(bulk_data, incr_bulk_len_);
+        evbuffer_drain(input, incr_bulk_len_ + 2);
```
mapleFU, line 588: Not related to this issue directly

I'm not familiar with `evbuffer` api, but would

```c++
std::string bulk_string(0, incr_bulk_len_);
evbuffer_remove(input, bulk_string.data(), incr_bulk_len_ + 2);
```

Being ok since it avoid adjust the `input` internal? (The bad things is that std::string would zero-initialize itself, which introducing a round of copying 😅) @git-hulk @PragmaTwice

mapleFU: oh let's not optimize this if no benchmark shows it's better...

src/cluster/replication.cc
```diff
@@ -576,31 +578,35 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
       }
       case Incr_batch_data:
         // Read bulk data (batch data)
-        if (incr_bulk_len_ + 2 <= evbuffer_get_length(input)) {  // We got enough data
-          bulk_data = reinterpret_cast<char *>(evbuffer_pullup(input, static_cast<ssize_t>(incr_bulk_len_ + 2)));
-          std::string bulk_string = std::string(bulk_data, incr_bulk_len_);
-          // master would send the ping heartbeat packet to check whether the slave was alive or not,
-          // don't write ping to db here.
```
PragmaTwice: Why remove it?

src/cluster/replication.cc
RiversJin: ok

src/cluster/replication.cc
RiversJin: a minor mistake... already reverted.

PragmaTwice (dismissed):
Looks good to me. Thank you!

PragmaTwice:
You can run `./x.py format` from the root dir of the repo to format code via `clang-format`.

Additional, please ensure `clang-format` exists and is in version 14. (download from [here](https://github.com/llvm/llvm-project/releases/tag/llvmorg-14.0.0))

Read https://kvrocks.apache.org/community/contributing for more information.

PragmaTwice:
Also, although it's not mandatory, it's good to make your further changes small additional commits instead of amending to the existing commit, to make reviewers easy to know what you change after a round of reviewing.

RiversJin:
> Also, although it's not mandatory, it's good to make your further changes small additional commits instead of amending to the existing commit, to make reviewers easy to know what you change after a round of reviewing.

ok, fixed. I was a bit surprised to find that different versions of clang-format can produce different results under the same configuration file...

mapleFU (approved):

PragmaTwice (approved):

PragmaTwice:

PR #2691, opened 2024-12-11
test(integration): add integration test for master lost during syncing sst files.

### LINKED ISSUE #2671: Add more test cases about replication and cluster
By the way, can we add a test with master removed and about 5s' connection timeout in go integration?

_Originally posted by @mapleFU in https://github.com/apache/kvrocks/issues/2662#issuecomment-2477964921_

LindaSummer:
# Issue

Close #2671 

# Proposed Changes

Add integration test for losing master during SST file syncing.

Here is the steps.

1. Create master and slave instances.
2. Apply master and slave setnodeid.
3. Create a slow tcp proxy for master.
4. Setnodes to master and write data to master.
5. Setnodes to slave.
6. Wait for one second and close tcp proxy.
7. Try setnodes in slave and get duration.

### DIFF
--- a/tests/gocase/integration/replication/replication_test.go
+++ b/tests/gocase/integration/replication/replication_test.go
@@ -552,3 +552,60 @@ func TestFullSyncReplication(t *testing.T) {
 		require.Equal(t, "bar", slaveClient.Get(ctx, "foo").Val())
 	})
 }
+
+func TestSlaveLostMaster(t *testing.T) {
+	// integration test for #2662 and #2671
+	ctx := context.Background()
+
+	masterSrv := util.StartServer(t, map[string]string{
+		"cluster-enabled":               "yes",
+		"max-replication-mb":            "1",
+		"rocksdb.compression":           "no",
+		"rocksdb.write_buffer_size":     "1",
+		"rocksdb.target_file_size_base": "1",
+	})
+	defer func() { masterSrv.Close() }()
+	masterClient := masterSrv.NewClient()
+	defer func() { require.NoError(t, masterClient.Close()) }()
+	masterNodeID := "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx00"
+	require.NoError(t, masterClient.Do(ctx, "clusterx", "SETNODEID", masterNodeID).Err())
+
+	replicaSrv := util.StartServer(t, map[string]string{
+		"cluster-enabled":                "yes",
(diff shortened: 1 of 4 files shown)


LindaSummer:
Hi,

I'm not sure my case is correctly covered the scenario in related issue.

I only got about 1 second delay in my own machine.

Please let me know if there is any problem about the test case. 😊

Best Regards,
Edward

PragmaTwice (approved):
Thank you for your contribution! See if some others have comments.

tests/gocase/integration/replication/replication_test.go
```diff
@@ -552,3 +552,60 @@ func TestFullSyncReplication(t *testing.T) {
...
+	}
+
+	require.NoError(t, replicaClient.Do(ctx, "clusterx", "SETNODES", clusterNodesInfo, "1").Err())
+
+	time.Sleep(2 * time.Second)
+	cancelProxy()
+	start := time.Now()
+	require.NoError(t, replicaClient.Do(ctx, "clusterx", "SETNODES", unexistNodesInfo, "2").Err())
+	duration := time.Since(start)
+	require.Less(t, duration, time.Second*6)
```
mapleFU, line 610: Why the time out is 5000/5100 but the duration be 6 here 🤔

git-hulk (approved):
LGTM, great work.

tests/gocase/integration/replication/replication_test.go
LindaSummer: Hi @mapleFU ,

The time duration may not be a precise value due to network.

So I reserved 1 second to reduce flaky failure. 😊

Best Regards,
Edward

mapleFU: What would a duration really looks like? Would it:
1. Force greater than 5?
2. Always less than 6? (I guess this is not guranteed ...)

LindaSummer: Hi @mapleFU ,

Yes, I thought it should be in a [5, 6 or longer] range.

But I only got about one second in my own environment and make me confused and not sure whether the case full covers the issue. 🤔

Maybe we can set the upper bound to a conservative value just for integration test. 😊

Best Regards,
Edward

mapleFU: > But I only got about one second in my own environment and make me confused and not sure whether the case full covers the issue. 🤔

Thanks! I'll dive in my env later, we can first move forward

PR #2703, opened 2024-12-31
feat(search): add basic infra for text fields (without indexing)

PragmaTwice:
In this PR, the basic infrastructure for text fields are added, including:
- redisearch / sql parser support for text query
- text filtering support in filter executor
- text field metadata / management
- text query node in IR & sema checking

Note that this PR doesn't bring text indexing (the key algorithm for full-text search).


dmazzella:
Any progress on full-text support?

PragmaTwice:
> Any progress on full-text support?

hmm currently i don't have much time for this feature.

PR #2716, opened 2025-01-08
fix(cluster): should stop the migration if it's changed to the slave role

RiversJin:
In cluster mode, if a master-slave failover occurs, it seems that slot migration is not properly stopped. However, the slave of command has corresponding update logic. So, this is a bug, right?

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -237,6 +237,9 @@ Status Cluster::SetMasterSlaveRepl() {
     return Status::OK();
   }
 
+  bool is_slave = srv_->IsSlave();
+  bool is_cluster_enabled = srv_->GetConfig()->cluster_enabled;
+
   if (myself_->role == kClusterMaster) {
     // Master mode
     auto s = srv_->RemoveMaster();
@@ -244,8 +247,11 @@ Status Cluster::SetMasterSlaveRepl() {
       return s.Prefixed("failed to remove master");
     }
     LOG(INFO) << "MASTER MODE enabled by cluster topology setting";
-    srv_->slot_migrator->SetStopMigrationFlag(false);
-    LOG(INFO) << "Change server role to master, restart migration task";
+    if (is_slave && is_cluster_enabled) {
+      // Slave -> Master
+      srv_->slot_migrator->SetStopMigrationFlag(false);
+      LOG(INFO) << "Change server role to master, stop migration task";
+    }
     return Status::OK();
   }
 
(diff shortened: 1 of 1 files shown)


git-hulk:
@RiversJin Thanks for raising this issue. You're right that we should stop the migration while becoming to the slave role.

src/cluster/cluster.cc
```diff
@@ -237,23 +237,40 @@ Status Cluster::SetMasterSlaveRepl() {
...
+
   if (myself_->role == kClusterMaster) {
     // Master mode
     auto s = srv_->RemoveMaster();
     if (!s.IsOK()) {
       return s.Prefixed("failed to remove master");
     }
     LOG(INFO) << "MASTER MODE enabled by cluster topology setting";
-  } else if (nodes_.find(myself_->master_id) != nodes_.end()) {
+    if (is_slave && is_cluster_enabled) {
```
```suggestion
    if (is_slave && is_cluster_enabled && srv_->slot_migrator) {
```

src/cluster/cluster.cc
```diff
@@ -237,23 +237,40 @@ Status Cluster::SetMasterSlaveRepl() {
...
+      // Slave -> Master
+      srv_->slot_migrator->SetStopMigrationFlag(false);
+      LOG(INFO) << "Change server role to master, stop migration task";
```
git-hulk: Wrong log message, it should restart the migration task instead of the stop.

src/cluster/cluster.cc
```diff
@@ -237,23 +237,40 @@ Status Cluster::SetMasterSlaveRepl() {
...
-    std::shared_ptr<ClusterNode> master = nodes_[myself_->master_id];
+    std::shared_ptr<ClusterNode> master = it->second;
     auto s = srv_->AddMaster(master->host, master->port, false);
     if (!s.IsOK()) {
       LOG(WARNING) << "SLAVE OF " << master->host << ":" << master->port
                    << " wasn't enabled by cluster topology setting, encounter error: " << s.Msg();
       return s.Prefixed("failed to add master");
     }
-    LOG(INFO) << "SLAVE OF " << master->host << ":" << master->port << " enabled by cluster topology setting";
+    if (!is_slave && is_cluster_enabled) {
```
```suggestion
    if (!is_slave && is_cluster_enabled && srv_->slot_migrator) {
```

src/cluster/cluster.cc
git-hulk: The slot_migrator might be null pointer while starting the server.

src/cluster/cluster.cc
RiversJin: oops... fixed

caipengbo (approved):

git-hulk (approved):

PragmaTwice (approved):

PR #2721, opened 2025-01-14
feat: support histograms for command latency statistics

rabunkosar-dd:
This PR adds an option to build kvrocks with histogram support to breakdown the command latencies. Current statistics are based on the total latency and the command count, giving averages. For our use case, where tail latencies are very impactful, this was not sufficient to track the general performance of the instances.

It also adds a config parameter to define the bucket boundaries. I will also follow up with a PR to add similar support in kvrocks_exporter (and support additive histograms for prometheus). 

The buckets are not additive and there is an implicit bucket that tracks values beyond the defined values (inf bucket). 

~~I added a build paramater (ENABLE_HISTOGRAMS), when set to ON will compile the support into the binary.~~
The feature is enabled if the config parameter defines a non-empty list of buckets that are coma separated

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,6 @@ option(ENABLE_STATIC_LIBSTDCXX "link kvrocks with static library of libstd++ ins
 option(ENABLE_LUAJIT "enable use of luaJIT instead of lua" ON)
 option(ENABLE_OPENSSL "enable openssl to support tls connection" OFF)
 option(ENABLE_IPO "enable interprocedural optimization" ON)
-option(ENABLE_HISTOGRAMS "enable histograms to view the operation latencies" OFF)
 set(SYMBOLIZE_BACKEND "" CACHE STRING "symbolization backend library for cpptrace (libbacktrace, libdwarf, or empty)")
 set(PORTABLE 0 CACHE STRING "build a portable binary (disable arch-specific optimizations)")
 # TODO: set ENABLE_NEW_ENCODING to ON when we are ready
@@ -288,11 +287,6 @@ if(ENABLE_IPO)
         message(WARNING "IPO is not supported: ${ipo_output}")
     endif()
 endif()
-
-if(ENABLE_HISTOGRAMS)
-    target_compile_definitions(kvrocks_objs PUBLIC ENABLE_HISTOGRAMS)
-endif()
-
 
 # kvrocks main target
 add_executable(kvrocks src/cli/main.cc)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,7 @@ option(ENABLE_STATIC_LIBSTDCXX "link kvrocks with static library of libstd++ ins
 option(ENABLE_LUAJIT "enable use of luaJIT instead of lua" ON)
 option(ENABLE_OPENSSL "enable openssl to support tls connection" OFF)
 option(ENABLE_IPO "enable interprocedural optimization" ON)
+option(ENABLE_HISTOGRAMS "enable histograms to view the operation latencies" OFF)
 set(SYMBOLIZE_BACKEND "" CACHE STRING "symbolization backend library for cpptrace (libbacktrace, libdwarf, or empty)")
 set(PORTABLE 0 CACHE STRING "build a portable binary (disable arch-specific optimizations)")
 # TODO: set ENABLE_NEW_ENCODING to ON when we are ready
@@ -287,6 +288,11 @@ if(ENABLE_IPO)
         message(WARNING "IPO is not supported: ${ipo_output}")
     endif()
 endif()
+
+if(ENABLE_HISTOGRAMS)
+    target_compile_definitions(kvrocks_objs PUBLIC ENABLE_HISTOGRAMS)
+endif()
+
 
 # kvrocks main target
 add_executable(kvrocks src/cli/main.cc)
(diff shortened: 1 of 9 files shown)


git-hulk:
@rabunkosar-dd Thanks for your contribution. You use `./x.py format` to format your source codes after installing the clang-format-14

git-hulk:
> @rabunkosar-dd Thanks for your contribution. You use `./x.py format` to format your source codes after installing the clang-format-14

CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ option(ENABLE_STATIC_LIBSTDCXX "link kvrocks with static library of libstd++ ins
 option(ENABLE_LUAJIT "enable use of luaJIT instead of lua" ON)
 option(ENABLE_OPENSSL "enable openssl to support tls connection" OFF)
 option(ENABLE_IPO "enable interprocedural optimization" ON)
+option(ENABLE_HISTOGRAMS "enable histograms to view the operation latencies" OFF)
```
git-hulk: Perhaps we can remove this build gate and allow to disable it by the configuration way. For example, we can disable this feature if the bucket is set to empty.

rabunkosar-dd: Good suggestion, I did this to make it less intrusive initially and get your feedback. I can make it config driven only

src/stats/stats.cc
```diff
@@ -26,7 +26,7 @@
 #include "fmt/format.h"
 #include "time_util.h"
 
-Stats::Stats() {
+Stats::Stats(Config *config) : config_(config) {
```
git-hulk: Can we just pass the buckets to Stats instead of the entire `Config`.

git-hulk (commented):
Looks good to me, one comment inline.

src/config/config.cc
```diff
@@ -755,6 +756,24 @@ void Config::initFieldCallback() {
...
 #endif
+      {"histogram-bucket-boundaries",
+       [this]([[maybe_unused]] Server *srv, [[maybe_unused]] const std::string &k, const std::string &v) -> Status {
+         std::vector<std::string> buckets = util::Split(v, ",");
+         histogram_bucket_boundaries.clear();
+         if (buckets.size() < 1) {
+           return Status::OK();
+         }
+         std::transform(buckets.begin(), buckets.end(), std::back_inserter(histogram_bucket_boundaries),
+                        [](const std::string &val) { return std::stod(val); });
```
PragmaTwice: `stod` will throw exceptions while it fails to parse the string to number. We can use `ParseInt` in `parse_util.cc` instead.

src/stats/stats.cc
```diff
@@ -26,7 +26,7 @@
 #include "fmt/format.h"
 #include "time_util.h"
 
-Stats::Stats() {
+Stats::Stats(std::vector<double> bucket_boundaries) : bucket_boundaries(bucket_boundaries) {
```
PragmaTwice:
```suggestion
Stats::Stats(std::vector<double> bucket_boundaries) : bucket_boundaries(std::move(bucket_boundaries)) {
```

src/stats/stats.h
```diff
@@ -22,12 +22,15 @@
...
 #include <unistd.h>
 
+#include <algorithm>
 #include <atomic>
 #include <map>
 #include <shared_mutex>
 #include <string>
 #include <vector>
 
+#include "config/config.h"
```
PragmaTwice: can be removed.

git-hulk (dismissed):

git-hulk:
@rabunkosar-dd Could you please fix the lint issue?

danishgh (approved):

git-hulk (dismissed):
LGTM

src/stats/stats.h
```diff
@@ -43,6 +45,13 @@ enum StatsMetricFlags {
 
 constexpr int STATS_METRIC_SAMPLES = 16;  // Number of samples per metric
 
+// Experimental part to support histograms for cmd statistics
+struct CommandHistogram {
+  std::vector<std::shared_ptr<std::atomic<uint64_t>>> buckets;
```
PragmaTwice: It seems the atomic numbers inside the vector are not shared. Is it necessary to use `shared_ptr` here?

rabunkosar-dd: I used that as a workaround, since vector will not support atomic directly.

PragmaTwice: There are basically two things:

> since vector will not support atomic directly.

That's not totally correct. You can use `std::vector<std::atomic>`, but then some methods of `vector` will lead to compilation errors due to that `atomic` doesn't have copy ctor and copy assign op, e.g. `push_back`. Hence an obvious method is to avoid `push_back`, since we already know the size of the vector (i.e. make the vector a fixed-size one), then it can pass the compilation, with `vector<atomic>` being used.

> I used that as a workaround

If you want to wrap it inside a smart pointer, I recommend `std::unique_ptr`.

rabunkosar-dd: I would prefer using the pointer, will make the change

git-hulk (approved):

PragmaTwice (approved):

PragmaTwice:

rabunkosar-dd:
> Thank you for your contribution!

Thanks for the feedbacks and patience @PragmaTwice and @git-hulk

PR #2724, opened 2025-01-17
perf(cluster): add upper bound for scan during migration

RiversJin:
By setting an upper bound in advance during the scan, it can prevent RocksDB from having to scan through a block of tombstone data before stopping in certain situations.

### DIFF
--- a/src/storage/redis_metadata.cc
+++ b/src/storage/redis_metadata.cc
@@ -159,9 +159,7 @@ std::string ComposeSlotKeyPrefix(const Slice &ns, int slotid) {
   return output;
 }
 
-std::string ComposeSlotKeyUpperBound(const Slice &ns, int slotid) {
-  return ComposeSlotKeyPrefix(ns, slotid + 1);
-}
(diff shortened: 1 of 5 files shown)


src/storage/redis_metadata.cc
```diff
@@ -159,6 +159,19 @@ std::string ComposeSlotKeyPrefix(const Slice &ns, int slotid) {
...
+  std::string output;
+
+  PutFixed8(&output, static_cast<uint8_t>(ns.size()));
+  output.append(ns.data(), ns.size());
+
+  // Is't ok cause slotid <= HASH_SLOTS_SIZE(16384), uint16_t <= 65535
+  // no overflow
+  PutFixed16(&output, static_cast<uint16_t>(slotid + 1));
+
+  return output;
```
```suggestion
  return ComposeSlotKeyPrefix(ns, slotid + 1);
```

danishgh (approved):

git-hulk (approved):

PragmaTwice (approved):

Issue #2563, opened 2024-09-30
Hoist key mutexes from every command impls to Connection::ExecuteCommand


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

We can remove all key-level mutex locking operations in command implementation and handle them uniformly in `Connection::ExecuteCommand`. 

In this way, we can handle transaction support of scripts, index, etc. more easily.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PokIsemaine:
I am trying to move the operation of the lock key to `ExecuteCommands`(not `ExecuteCommand`), because I realized that the index might also need to be locked. 

If `cmd_flags` contains the `kCmdWrite` flag, I will use `attributes->ForEachKeyRange` to get all the keys that need to be locked and lock these keys together with `MultiLockGuard`.

What I want to confirm is whether the keys obtained in this way do indeed correspond to those that need to be locked, and if there are any redundancies or omissions?

Furthermore, after some attempts, I found that the locking for certain methods depends on the method's parameters. Since the locks are now uniformly applied within `ExecuteCommands`, we may need to unlock them based on the inverse conditions.
```
rocksdb::Status String::MSet(engine::Context &ctx, const std::vector<StringPair> &pairs, uint64_t expire_ms,
                             bool lock) {
  // Data race, key string maybe overwrite by other key while didn't lock the keys here,
  // to improve the set performance
  std::optional<MultiLockGuard> guard;
  if (lock) {
    std::vector<std::string> lock_keys;
    lock_keys.reserve(pairs.size());
    for (const StringPair &pair : pairs) {
      std::string ns_key = AppendNamespacePrefix(pair.key);
      lock_keys.emplace_back(std::move(ns_key));
    }
    guard.emplace(storage_->GetLockManager(), lock_keys);
  }
  

rocksdb::Status ZSet::RangeByRank(engine::Context &ctx, const Slice &user_key, const RangeRankSpec &spec,
                                  MemberScores *mscores, uint64_t *removed_cnt) {
  if (mscores) mscores->clear();

  uint64_t cnt = 0;
  if (!removed_cnt) removed_cnt = &cnt;
  *removed_cnt = 0;

  std::string ns_key = AppendNamespacePrefix(user_key);

  std::optional<LockGuard> lock_guard;
[...]

PR #2731, opened 2025-01-24
feat(script, function): remove global lock for EVAL, EVALSHA and FCALL

### LINKED ISSUE #2612: Transaction support for scripting
### Search before asking

### Motivation

- Pass context into scripting and use one context for all commands in one script
- Key locking improvement (#2563)
- Remove global lock for EVAL command

### Solution

--

PragmaTwice:
It closes #2612.

After #2563, kvrocks have gained the chance to remove global locks for lua scripting (as well as lua functions). In this PR, we:
- remove the server-level global lua vm; just keep these worker-level vms;
- remove exclusive flag for EVAL, EVALSHA and FCALL;
- dispatch all lua commands to worker-local vms.

It's expected that these commands will be more efficient after this patch, since the global lock is removed.

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -113,7 +113,6 @@ Server::Server(engine::Storage *storage, Config *config)
   AdjustOpenFilesLimit();
   slow_log_.SetMaxEntries(config->slowlog_max_len);
   perf_log_.SetMaxEntries(config->profiling_sample_record_max_len);
-  lua_ = lua::CreateState();
 }
 
 Server::~Server() {
@@ -134,8 +133,6 @@ Server::~Server() {
   }
   cleanupExitedWorkerThreads(true /* force */);
   CleanupExitedSlaves();
-
-  lua::DestroyState(lua_);
 }
 
 // Kvrocks threads list:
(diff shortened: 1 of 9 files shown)


src/server/server.cc
```diff
@@ -1013,16 +1010,12 @@ void Server::GetClientsInfo(std::string *info) {
...
   int64_t rss = Stats::GetMemoryRSS();
-  int memory_lua = lua_gc(lua_, LUA_GCCOUNT, 0) * 1024;
   std::string used_memory_rss_human = util::BytesToHuman(rss);
-  std::string used_memory_lua_human = util::BytesToHuman(memory_lua);
 
   std::ostringstream string_stream;
   string_stream << "# Memory\r\n";
   string_stream << "used_memory_rss:" << rss << "\r\n";
   string_stream << "used_memory_rss_human:" << used_memory_rss_human << "\r\n";
-  string_stream << "used_memory_lua:" << memory_lua << "\r\n";
```
git-hulk: We should keep the Lua VM memory usage section. The `used_memory_lua` can be the sum value of the worker's Lua VM memory.

PragmaTwice: Added.

git-hulk (dismissed):

git-hulk (approved):

PR #2736, opened 2025-01-25
chore(server): check the arity should happen before the flag generate

git-hulk:
Currently, we check the number of arguments after generating the flag and the authentication. Move it before them can fastly return the error if the argument number is wrong. Also, it will make safer to implement the flag gen function.

### DIFF
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -397,8 +397,15 @@ void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
 
     const auto &attributes = current_cmd->GetAttributes();
     auto cmd_name = attributes->name;
-    auto cmd_flags = attributes->GenerateFlags(cmd_tokens);
 
+    int tokens = static_cast<int>(cmd_tokens.size());
+    if (!attributes->CheckArity(tokens)) {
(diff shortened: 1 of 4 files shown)


PragmaTwice:
After #2732 is merged, we can remove these arg size checks inside flag gen functions.

src/server/redis_connection.cc
```diff
@@ -397,8 +397,15 @@ void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
 
     const auto &attributes = current_cmd->GetAttributes();
     auto cmd_name = attributes->name;
-    auto cmd_flags = attributes->GenerateFlags(cmd_tokens);
```
PragmaTwice, line 400: We should also check this call in scripting, not only here.

git-hulk: Good point, done.

git-hulk:
> After https://github.com/apache/kvrocks/pull/2732 is merged, we can remove these arg size checks inside flag gen functions.

It should be good to keep the argument number checked inside the flag gen functions. We can avoid depending on the order between CheckArity and FlagGen functions.

PragmaTwice (dismissed):

PragmaTwice (approved):

PR #2740, opened 2025-01-25
feat(script): allow blocking commands in scripting

PragmaTwice:
We refactored blocking commands to be aware whether it's executed from scripting, and avoid to block if it is.

So that now these blocking commands can be executed in lua scripts.


PragmaTwice:
Ready for review now.

git-hulk (approved):
LGTM

Issue #2558, opened 2024-09-27
Tracking Issues: support t-digest data structure and commands


LindaSummer:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Redis Stack already supports this probabilistic data structure, and we can support it too.

- #2423 

This includes:

- [x] #2741
- [ ] #2789 
- [x] #2791 
- [x] #2792 
- [x] #2793 
- [x] #2794 
- [x] #2804
- [x] #2805
- [ ] #2807 
- [x] #3062 
- [x] #3063 
- [x] #3064 
- [x] #3065 
- [x] #3066 
- [x] #2813

### Solution

- #2542 

### Are you willing to submit a PR?

- [X] I'm willing to submit a PR!

LindaSummer:
Hi @PragmaTwice ,

Sorry for delay. 

This issue's implementation is still in progress. 😊

I will try my best to implement and test the first fundamental commands in one month.

Best Regards,
Edward

LindaSummer:
Hi @PragmaTwice ,

  Sorry that I may need some more time to finish the first PR of t-digest feature. 

  Major part of `create`, `add` and `quantile` are nearly finished and I'm now trying to do self-testing and add unit tests. 😊

  Here is my draft PR https://github.com/LindaSummer/kvrocks/pull/1 in my own repo.
  I'll try my best to add more tests and make it ready to be reviewed. 😊

Best Regards,
Edward

yunxiao3:
Hello @LindaSummer,  I am very interested in your project, and it seems that there is a lot of development tasks that need to be done. If possible, could you assign me some sub-tasks or areas where I can help?  I’d like to contribute to the growth of the kvrocks community.

LindaSummer:
> Hello [<img alt="" width="16" height="16" src="https://avatars.githubusercontent.com/u/14014471?u=ddefac063166b1a2f14250b595de2a3e8412b9db&amp;v=4&amp;size=80">@LindaSummer](https://github.com/LindaSummer), I am very interested in your project, and it seems that there is a lot of development tasks that need to be done. If possible, could you assign me some sub-tasks or areas where I can help? I’d like to contribute to the growth of the kvrocks community.

Hi @yunxiao3 ,

Thanks for your interests in this topic! 😊

You can try to pick up one command in current issue list or create a new issue for implementation. 😊

I will implement this #2791 command first in recent days.
[...]

PR #2741, opened 2025-01-25
feat(tdigest): add tdigest algorithm and storage encoding implementations

LindaSummer:
# Issue

- #2558

# Proposal

- #2542 

# Proposed Changes

- add tdigest core algorithm support.
- add encoding and decoding method as designed.
- add core implementations for `create`, `add` and `quantile`.
- add basic unit tests for `create`, `add` and `quantile`.

# TODO List

- add more unit tests for corner cases and benchmark.
- find a more stable way to evaluate the error of quantile.
- refactor the `DummyCentroids` to an iteration way for efficiency.
- integrate the core implementations to commands and add integration tests.


LindaSummer:
Hi Team,

Sorry for delaying this feature's implementation.

I kept the commit log since we will squash merging.
Maybe it can help new-commers if they meet similar problems as me for a new feature. 😄

Please give me some suggestions and help take a review.

There are still many parts need to be improved, and I will be very happy to try my best for any improvement. 😊

Best Regards,
Edward

src/storage/redis_metadata.cc
```diff
@@ -497,3 +499,40 @@ rocksdb::Status HyperLogLogMetadata::Decode(Slice *input) {
...
+  PutFixed64(dst, total_observations);
+  PutFixed64(dst, merge_times);
+}
+
+rocksdb::Status TDigestMetadata::Decode(Slice *input) {
+  if (auto s = Metadata::Decode(input); !s.ok()) {
+    return s;
+  }
+
+  if (input->size() < sizeof(uint64_t) * 8) {
```
```suggestion
  if (input->size() < sizeof(uint64_t) * 8 + sizeof(double) + 2) {
```

src/storage/storage.h
```diff
@@ -70,9 +70,10 @@ enum class ColumnFamilyID : uint32_t {
   Propagate,
   Stream,
   Search,
+  TDigest,
 };
 
-constexpr uint32_t kMaxColumnFamilyID = static_cast<uint32_t>(ColumnFamilyID::Search);
+constexpr uint32_t kMaxColumnFamilyID = static_cast<uint32_t>(ColumnFamilyID::TDigest);
```
PragmaTwice: I think we can use `PrimarySubkey` to store subkeys instead of a new column family?

src/types/tdigest.h
```diff
@@ -0,0 +1,173 @@
...
+
+  // find centroid contains the index
+  double weight_sum = 0;
+  auto iter = td.Begin();
+  for (; iter->Valid(); iter->Next()) {
+    auto centroid = iter->GetCentroid();
+    if (!centroid) {
+      return centroid.ToStatus();
+    }
+    weight_sum += centroid->weight;
```
PragmaTwice:
```suggestion
    auto centroid = GET_OR_RET(iter->GetCentroid());
    weight_sum += centroid.weight;
```

PragmaTwice: And other places.

PragmaTwice: or, `weight_sum += GET_OR_RET(iter->GetCentroid()).weight;` if you like.

src/types/tdigest.cc
```diff
@@ -0,0 +1,439 @@
...
+  digest.Merge(others);
+
+  return digest.DumpCentroids();
+}
+StatusOr<CentroidsWithDelta> TDigestMerge(const std::vector<double>& buffer, const CentroidsWithDelta& centroid_list) {
+  TDigest digest{centroid_list.delta};
+  digest.Reset(centroid_list);
+  digest.Add(buffer);
+  return digest.DumpCentroids();
+}
```
PragmaTwice: We can leave an empty line in the end.

src/types/tdigest.cc
```diff
@@ -0,0 +1,439 @@
...
+  return {
+      .centroids = std::move(centroids),
+      .delta = impl_->Delta(),
+      .min = impl_->Min(),
+      .max = impl_->Max(),
+      .total_weight = impl_->TotalWeight(),
+  };
+}
+
+void TDigest::Add(std::vector<double> items) { impl_->MergeInput(items); }
```
PragmaTwice: Should the parameter be a reference type?

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,429 @@
...
+  double Min() const { return meta_data_.minimum; }
+  double Max() const { return meta_data_.maximum; }
+  uint64_t Size() const { return meta_data_.merged_nodes; }
+
+ private:
+  const TDigestMetadata& meta_data_;
+  std::vector<Centroid> centroids_;
+};
+
+uint64_t constexpr kMaxBufferSize = 1 * 1024;  // 1k doubles
```
mapleFU: 1k doubles means a 8KB buffer size? Or should we rename this to `kMaxElements` or something else?

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,429 @@
...
+    LockGuard guard(storage_->GetLockManager(), ns_key);
+
+    auto status = GetMetaData(ctx, ns_key, &metadata);
+    if (!status.ok()) {
+      return status;
+    }
+
+    auto batch = storage_->GetWriteBatchBase();
+    WriteBatchLogData log_data(kRedisTDigest);
+    status = batch->PutLogData(log_data.Encode());
```
mapleFU: Can we just use RETURN_NOT_OK to simplify the code?

src/storage/redis_metadata.cc
LindaSummer: Hi @PragmaTwice ,

Thanks for your review! 😊
I have fixed this bug in new commit.

Best Regards,
Edward

src/storage/storage.h
LindaSummer: Hi @PragmaTwice ,

Got it! 😄

I have changed the column family to `PrimarySubkey`.

Best Regards,
Edward

src/types/tdigest.h
LindaSummer: Hi @PragmaTwice ,

Thanks for your advice! 😄

I have replaced this block with `GET_OR_RET`.

Best Regards,
Edward

src/types/tdigest.cc
LindaSummer: Hi @PragmaTwice ,

Got it!

I have added a new endline. 😄

Best Regards,
Edward

src/storage/redis_metadata.cc
```diff
@@ -497,3 +499,40 @@ rocksdb::Status HyperLogLogMetadata::Decode(Slice *input) {
...
+  PutFixed64(dst, total_observations);
+  PutFixed64(dst, merge_times);
+}
+
+rocksdb::Status TDigestMetadata::Decode(Slice *input) {
+  if (auto s = Metadata::Decode(input); !s.ok()) {
+    return s;
+  }
+
+  if (input->size() < sizeof(uint64_t) * 8 + sizeof(double) + 2) {
```
```suggestion
  if (input->size() < sizeof(uint64_t) * 8 + sizeof(double) * 2) {
```

typo?

PragmaTwice: Ahh sorry, my mistake in the previous suggestion.

src/types/tdigest.cc
LindaSummer: Hi @PragmaTwice ,

I have updated the public method as a const reference and the internal method tries to sort the vector, so I make it a pass by value. 😄

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Got it! Thanks for your suggestion! 😄

I have renamed the variable.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

I didn't find a function or macro named like `RETURN_NOT_OK`. Do you mean the `GET_OR_RET`?

I have updated all the status statement to an inline if statement. 😄

Please let me know if I misunderstand.

Best Regards,
Edward

src/storage/redis_metadata.cc
LindaSummer: Sorry that I forget to double check it. 😄

I have updated the statement.

Thanks very much for your warm help!

src/types/redis_tdigest.cc
PragmaTwice: GET_OR_RET can also be used here. But the current form is also fine.

src/types/tdigest.cc
PragmaTwice: Would you mind to introduce a newline at end for every new source file?

LindaSummer: Of course!

I will check all files and add each file a newline at end. 😄

LindaSummer: Hi @PragmaTwice ,

I have added endline for each file in this PR. 😄
Please let me know if I missed any file.

Best Regards,
Edward

src/storage/redis_metadata.h
```diff
@@ -91,9 +93,9 @@ enum RedisCommand {
   kRedisCmdLMove,
 };
 
-const std::vector<std::string> RedisTypeNames = {"none",   "string",    "hash",      "list",
-                                                 "set",    "zset",      "bitmap",    "sortedint",
-                                                 "stream", "MBbloom--", "ReJSON-RL", "hyperloglog"};
+const std::vector<std::string> RedisTypeNames = {"none",      "string",      "hash",      "list",   "set",
+                                                 "zset",      "bitmap",      "sortedint", "stream", "MBbloom--",
+                                                 "ReJSON-RL", "hyperloglog", "tdigest"};
```
```suggestion
                                                 "ReJSON-RL", "hyperloglog", "TDIS-TYPE"};
```

LindaSummer: Hi @PragmaTwice ,

Thanks for your suggestion! 😊

I have updated the type name of tdigest.

Best Regards,
Edward

src/types/tdigest.cc
```diff
@@ -0,0 +1,439 @@
...
+#include <algorithm>
+#include <iterator>
+#include <memory>
+#include <queue>
+
+#include "status.h"
+
+namespace {
+// scale function K0: linear function, as baseline
+struct ScalerK0 {
```
mapleFU: Seems this is unused?

src/types/tdigest.h
```diff
@@ -0,0 +1,148 @@
...
+  // merge with another centroid
+  void Merge(const Centroid& centroid) {
+    weight += centroid.weight;
+    mean += (centroid.mean - mean) * centroid.weight / weight;
+  }
+
+  std::string ToString() const { return fmt::format("centroid<mean: {}, weight: {}>", mean, weight); }
+
+  template <typename T>
+  explicit Centroid(T&& centroid) : mean(centroid.mean), weight(centroid.weight) {}
```
mapleFU: Why not just leave a copy ctor here?

src/types/tdigest.h
```diff
@@ -0,0 +1,148 @@
...
+
+  explicit Centroid() = default;
+  explicit Centroid(double mean, double weight) : mean(mean), weight(weight) {}
+};
+
+struct CentroidsWithDelta {
+  std::vector<Centroid> centroids;
+  uint64_t delta;
```
mapleFU, line 47: What does delta means here?

src/types/tdigest.h
```diff
@@ -0,0 +1,148 @@
...
+**/
+
+// a numerically stable lerp is unbelievably complex
+// but we are *approximating* the quantile, so let's keep it simple
+// reference:
+// https://github.com/apache/arrow/blob/27bbd593625122a4a25d9471c8aaf5df54a6dcf9/cpp/src/arrow/util/tdigest.cc#L38
+static inline double Lerp(double a, double b, double t) { return a + t * (b - a); }
+
+template <typename TD, typename Lerp>
+inline StatusOr<double> TDigestQuantile(TD&& td, double q, Lerp lerp) {
```
mapleFU: why using template for `TD`? Also why `Lerp` is an argument?

src/storage/redis_metadata.h
```diff
@@ -336,3 +338,27 @@ class HyperLogLogMetadata : public Metadata {
...
+
+  explicit TDigestMetadata(uint64_t compression, uint64_t capacity, bool generate_version = true)
+      : Metadata(kRedisTDigest, generate_version), compression(compression), capacity(capacity) {}
+  explicit TDigestMetadata(bool generate_version = true) : TDigestMetadata(0, 0, generate_version) {}
+  void Encode(std::string *dst) const override;
+  rocksdb::Status Decode(Slice *input) override;
+
+  uint64_t TotalNodes() const { return merged_nodes + unmerged_nodes; }
+
+  double Delta() const { return 1. / static_cast<double>(compression); }
```
mapleFU, line 364: So delta means compression here?

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,416 @@
...
+#include "storage/redis_db.h"
+#include "storage/redis_metadata.h"
+#include "types/tdigest.h"
+
+namespace redis {
+
+// It should be replaced by a iteration of the rocksdb iterator
+class DummyCentroids {
+ public:
+  DummyCentroids(const TDigestMetadata& meta_data, std::vector<Centroid> centroids)
```
mapleFU: why copying these centroids?

mapleFU (commented):
I'm still learning the concept of tdigests, just part of the review

src/types/tdigest.cc
LindaSummer: Hi @mapleFU ,

This implement follows Apache arrow. 
I find it is also [unused in arrow](https://github.com/apache/arrow/blob/27bbd593625122a4a25d9471c8aaf5df54a6dcf9/cpp/src/arrow/util/tdigest.cc#L53).

I'm not sure whether I need to keep it and try to follow the original implementation, so I kept it in code.
I guess it is reserved for scale function's benchmark.

Maybe it can be removed until we try to do some benchmark for the scale function. 😄

Best Regards,
Edward

src/types/tdigest.h
LindaSummer: Hi @mapleFU ,

Thanks for your suggestion! 😄
I have updated it to a default copy ctor.

Best Regards,
Edward

src/types/tdigest.h
LindaSummer: Hi @mapleFU ,

The delta should be the delta used in TDigest.
This struct represents the internal state of the TDigest algorithm.

Best Regards,
Edward

src/types/tdigest.h
LindaSummer: Hi @mapleFU ,

I want to make the `TD` an static implementation which can be evolved in future.
I try to reduce the virtual function, so I choose a template implementation.

The `Lerp` is an abstract function here, maybe it could be a dynamic runtime one like `std::function`.
So, I decide to use an argument here.

Please let me know if here can be improved. 😄

Best Regards,
Edward

src/storage/redis_metadata.h
LindaSummer: Hi @mapleFU ,

Yes, this delta should be the compression. 😊

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Thanks for your suggestion! 😊

I have refactored it to a const reference. 

Best Regards,
Edward

src/storage/redis_metadata.h
```diff
@@ -336,3 +338,27 @@ class HyperLogLogMetadata : public Metadata {
...
+  uint64_t compression;
+  uint64_t capacity;
+  uint64_t unmerged_nodes = 0;
+  uint64_t merged_nodes = 0;
+  uint64_t total_weight = 0;
+  uint64_t merged_weight = 0;
+  double minimum = std::numeric_limits<double>::infinity();
+  double maximum = -1 * std::numeric_limits<double>::infinity();
+  uint64_t total_observations = 0;
+  uint64_t merge_times = 0;
```
mapleFU: So statisitics like `merge_times` and `total_observations` is also serialized, do they used for `INFO` output or just for debugging?

src/storage/redis_metadata.h
```diff
@@ -336,3 +338,27 @@ class HyperLogLogMetadata : public Metadata {
 
   EncodeType encode_type = EncodeType::DENSE;
 };
+
+class TDigestMetadata : public Metadata {
+ public:
+  uint64_t compression;
```
mapleFU: would 64bits too large for compression argument?

src/types/tdigest.h
```diff
@@ -0,0 +1,147 @@
...
+ *
+ */
+
+#pragma once
+
+#include <fmt/format.h>
+
+#include <vector>
+
+#include "status.h"
```
mapleFU: common/status.h for full path?

src/types/tdigest.h
mapleFU: I mean would it able to give the different lerp function?

src/types/tdigest.h
```diff
@@ -0,0 +1,147 @@
...
+  // index happen to be in a unit weight centroid
+  if (centroid.weight == 1 && std::abs(diff) < 0.5) {
+    return centroid.mean;
+  }
+
+  // find adjacent centroids for interpolation
+  auto ci_left = iter->Clone();
+  auto ci_right = iter->Clone();
+  if (diff > 0) {
+    if (ci_right == td.End()) {
```
mapleFU, line 124: `Valid` and `End` api mixed in TDigestQuantile function, what should we use here?

src/types/tdigest.cc
```diff
@@ -0,0 +1,439 @@
...
+
+  void Merge(const std::vector<TDigest>& others);
+  void Add(const std::vector<double>& items);
+  void Reset(const CentroidsWithDelta& centroid_list);
+  void Reset();
+  CentroidsWithDelta DumpCentroids() const;
+
+ private:
+  class TDigestImpl;
+  std::unique_ptr<TDigestImpl> impl_;
```
mapleFU: Why use pimpl in just .cc ...

src/storage/redis_metadata.h
```diff
@@ -336,3 +338,27 @@ class HyperLogLogMetadata : public Metadata {
 
   EncodeType encode_type = EncodeType::DENSE;
 };
+
+class TDigestMetadata : public Metadata {
```
mapleFU, line 343: Do we have some validate function here?

src/storage/redis_metadata.h
```diff
@@ -336,3 +338,27 @@ class HyperLogLogMetadata : public Metadata {
...
+ public:
+  uint64_t compression;
+  uint64_t capacity;
+  uint64_t unmerged_nodes = 0;
+  uint64_t merged_nodes = 0;
+  uint64_t total_weight = 0;
+  uint64_t merged_weight = 0;
+  double minimum = std::numeric_limits<double>::infinity();
+  double maximum = -1 * std::numeric_limits<double>::infinity();
```
mapleFU: I'm afraid init double via this way is not a good way, what about just min max?

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,416 @@
...
+};
+
+uint64_t constexpr kMaxElements = 1 * 1024;  // 1k doubles
+
+std::optional<rocksdb::Status> TDigest::Create(engine::Context& ctx, const Slice& digest_name,
+                                               const TDigestCreateOptions& options) {
+  auto ns_key = AppendNamespacePrefix(digest_name);
+  auto capacity = options.compression * 6 + 10;
+  capacity = ((capacity < kMaxElements) ? capacity : kMaxElements);
+  TDigestMetadata metadata(options.compression, capacity);
```
mapleFU, line 122: Is there any validation for the TDigestCreateOptions or metadata? I'm afraid a too large compression is meaning less

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,416 @@
...
+      return status;
+    }
+
+    if (status.ok()) {
+      rocksdb::Slice buffer_slice = buffer_value;
+      for (uint64_t i = 0; i < metadata.unmerged_nodes; ++i) {
+        double tmp_value = std::numeric_limits<double>::quiet_NaN();
+        if (!GetDouble(&buffer_slice, &tmp_value)) {
+          return rocksdb::Status::Corruption(
+              fmt::format("metadata has {} records, but get {} failed", metadata.unmerged_nodes, i));
```
mapleFU: should we mention this is in buffer?

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,416 @@
...
+      return status;
+    }
+    centroids->emplace_back(centroid);
+  }
+
+  if (clean_after_dump_batch != nullptr) {
+    if (auto status = (*clean_after_dump_batch)->DeleteRange(cf_handle_, start_key, guard_key); !status.ok()) {
+      return status;
+    }
+  }
```
mapleFU: why not also drop them key-by-key?

src/types/redis_tdigest.h
```diff
@@ -0,0 +1,97 @@
...
+
+  rocksdb::ColumnFamilyHandle* cf_handle_;
+
+  rocksdb::Status appendBuffer(engine::Context& ctx, ObserverOrUniquePtr<rocksdb::WriteBatchBase>& batch,
+                               const std::string& ns_key, const std::vector<double>& inputs, TDigestMetadata* metadata);
+
+  rocksdb::Status dumpCentroidsAndBuffer(
+      engine::Context& ctx, const std::string& ns_key, const TDigestMetadata& metadata,
+      std::vector<Centroid>* centroids, std::vector<double>* buffer = nullptr,
+      ObserverOrUniquePtr<rocksdb::WriteBatchBase>* clean_after_dump_batch = nullptr);
```
mapleFU: Can we avoid default argument and add docstring here?

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,416 @@
...
+  std::string metadata_bytes;
+  metadata.Encode(&metadata_bytes);
+  if (auto status = batch->Put(metadata_cf_handle_, ns_key, metadata_bytes); !status.ok()) {
+    return status;
+  }
+
+  return storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+rocksdb::Status TDigest::Quantile(engine::Context& ctx, const Slice& digest_name, const std::vector<double>& qs,
```
mapleFU, line 188: What if "Buffer" not exists in this scenerio?

mapleFU (commented):
I'm on my tour out, would be a bit slowly, just part of the comment

src/types/tdigest.h
LindaSummer: Hi @mapleFU ,

Thanks for your suggestion! 😄

I have replaced `Lerp` template argument to `Lerp` function itself.
I think this function may not change frequently.

Best Regards,
Edward

src/storage/redis_metadata.h
LindaSummer: Hi @mapleFU ,

These two metrics are used in `INFO` command for compatibility with redis.

Best Regards,
Edward

src/storage/redis_metadata.h
LindaSummer: Hi @mapleFU ,

Thanks for your advice! 😊
I have updated compression and capacity to `uint32_t`.

Best Regards,
Edward

src/types/tdigest.h
LindaSummer: Hi @mapleFU ,

I have updated related headers to `common/status.h`.

Best Regards,
Edward

src/types/tdigest.h
LindaSummer: Hi @mapleFU ,

Thanks for your suggestion! This should be fixed for a clear semantic. 😊

I have updated the usage and implementation of `End`.

Best Regards,
Edward

src/types/tdigest.cc
LindaSummer: Hi @mapleFU ,

I referred Arrow's implementation and didn't notice that I can remove this p-impl. 😄

This has been refactored to a regular member.

Best Regards,
Edward

src/storage/redis_metadata.h
LindaSummer: Hi @mapleFU ,

Do you mean for validation the compression property?

I have added a limitation for it as suggested.

Best Regards,
Edward

src/storage/redis_metadata.h
LindaSummer: Hi @mapleFU ,

Got it! 😊

I have updated with `max` and `lowest`.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Thanks for your suggestion! 😊
I have set a limitation with 1000.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Got it! 😊
I have moved the messages to log.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Thanks for your suggestion! 😊
I have replaced it with key-by-key deletion.

Best Regards,
Edward

src/types/redis_tdigest.h
LindaSummer: Hi @mapleFU ,

Got it! 😊

I have removed default `nullptr_t` and add docstring for it.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

It will try to merge buffer before dumping centroids.

So, if there is no buffer, the merging will be skipped and continue calculating quantile with centroids.

Best Regards,
Edward

src/storage/redis_metadata.h
mapleFU: We can add a comment for that one

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,425 @@
...
+      return std::make_unique<Iterator>(centroids_.cend(), centroids_);
+    }
+    bool Next() {
+      if (Valid()) {
+        std::advance(iter_, 1);
+      }
+      return iter_ != centroids_.cend();
+    }
+    bool Prev() {
+      if (Valid()) {
```
mapleFU: assume it's at end, what's it expected behavior here?

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,425 @@
...
+    bool Valid() const { return iter_ != centroids_.cend(); }
+    StatusOr<Centroid> GetCentroid() const {
+      if (iter_ == centroids_.cend()) {
+        return {::Status::NotOK, "invalid iterator during decoding tdigest centroid"};
+      }
+      return *iter_;
+    }
+
+   private:
+    std::vector<Centroid>::const_iterator iter_;
```
mapleFU, line 89: Curious why we don't use integer index for that

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,425 @@
...
+
+  LockGuard guard(storage_->GetLockManager(), ns_key);
+  auto status = GetMetaData(ctx, ns_key, &metadata);
+  if (status.ok()) {
+    return {};
+  }
+
+  if (!status.IsNotFound()) {
+    return status;
+  }
```
mapleFU: Can we merge these two lines together?

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,425 @@
...
+
+  std::string metadata_bytes;
+  metadata.Encode(&metadata_bytes);
+  if (auto status = batch->Put(metadata_cf_handle_, ns_key, metadata_bytes); !status.ok()) {
+    return status;
+  }
+
+  return storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
```
mapleFU, line 148: newline?

src/types/redis_tdigest.cc
mapleFU: But still lock would be held and new log would be added?

mapleFU (commented):
General LGTM

src/storage/redis_metadata.h
LindaSummer: Hi @mapleFU ,

Got it!

I have added comments for these two properties.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Thanks for your suggestion! 😊

I have added a check before try to merge buffer into centroids.
It will skip this action if unmerged nodes is empty.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

At the beginning of tdigest centroids iteration, we have confirmed the value must be inside the range of `(min_, max_)`.
So, if iteration reaches the `cend()`, it means that the precheck logic doesn't cover all cases or the centroids may be corrupted.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

I plan to implement the next version of centroids iteration with rocksdb iterator, during iteration in rocksdb we can't directly use index.
So, I try to keep them with similar style. 😊

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Thanks for your suggestion! 😊

We need to return both status and the existence of same tdigest before creation.
I have refactored the `std::optional` with an explicit flag `exists` with docstring.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Got it! 😊
I have added a newline before next function.

LindaSummer:
Hi @mapleFU ,

Sorry to interrupt you again. 😊
Could you help take a review for latest update?

Have a nice day!

Best Regards,
Edward

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,427 @@
...
+      }
+      return std::make_unique<Iterator>(centroids_.cend(), centroids_);
+    }
+    bool Next() {
+      if (Valid()) {
+        std::advance(iter_, 1);
+      }
+      return iter_ != centroids_.cend();
+    }
+    bool Prev() {
```
mapleFU, line 74: Can you add a comment to explain the `Prev()` requires to be valid? This function call is ok to be but a bit weird. Since if the iter is the `cend()`, for `std::vector`, it's not "Valid" but it prev is valid?

I've read the caller in tdigest.cc, this interface is ok since it prevent to call `Prev()` if `!Valid()`, but this requires some comment

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,427 @@
...
+#include "db_util.h"
+#include "encoding.h"
+#include "status.h"
+#include "storage/redis_db.h"
+#include "storage/redis_metadata.h"
+#include "types/tdigest.h"
+
+namespace redis {
+
+// It should be replaced by a iteration of the rocksdb iterator
```
mapleFU: Can you add a todo and create an issue for this?

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,427 @@
...
+  PutDouble(&value, centroid.weight);
+  return value;
+}
+
+rocksdb::Status TDigest::decodeCentroidFromKeyValue(const rocksdb::Slice& key, const rocksdb::Slice& value,
+                                                    Centroid* centroid) const {
+  InternalKey ikey(key, storage_->IsSlotIdEncoded());
+  auto subkey = ikey.GetSubKey();
+  auto type_flg = static_cast<uint8_t>(SegmentType::kGuardFlag);
+  GetFixed8(&subkey, &type_flg);
```
mapleFU: Have we checked the return value of `Get{}` api here?

mapleFU (dismissed):
General LGTM. We can first move on and add a go unittests

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Thanks for your suggestion! 😊
I have added comment for this function.

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Got it! 😊
I have added checking logic for the decode logic.

Best Regards,
Edward

LindaSummer:
> General LGTM. We can first move on and add a go unittests

Hi @mapleFU ,

Thanks very much for your warm help and patience! 😊
Do I need to add command with go integration tests in this PR or create a new PR?

Best Regards,
Edward

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Got it!
I have added a `TODO` label for this comment and attach it to the tracking issue #2558 . 😊

Best Regards,
Edward

mapleFU:
Let's merge first and then add lots of go tests

src/types/redis_tdigest.cc
```diff
@@ -0,0 +1,442 @@
...
+    }
+    bool Next() {
+      if (Valid()) {
+        std::advance(iter_, 1);
+      }
+      return iter_ != centroids_.cend();
+    }
+
+    // The Prev function can only be called for item is not cend,
+    // because we must gurantee the iterator to be inside the valid range before iteration.
```
PragmaTwice:
```suggestion
    // because we must guarantee the iterator to be inside the valid range before iteration.
```

PragmaTwice (approved):

PragmaTwice:

PR #2750, opened 2025-01-28
feat(hash): add the support of HSETEXPIRE command (HSET + EXPIRE)

ltagliamonte-dd:
New Command HSETEX (HSET+EXPIRE):
```
HSETEX key TTL f1 v1 f2 v2 f3 v3...
```
Adding a command that sets an hash key and adds also TTL.
this reduces I/O and improves drastically the performances over pipelining or LUA functions.

This is a common request feature/cmd https://github.com/redis/redis/issues/2905

**From my local tests with 8 workers, 600 client connections, over 4 executions on the same dataset.** 
(branch, TTL, results):
kvrocks/unstable, NO TTL: 392,346,350,346 seconds -> avg -> **358.50 s**
kvrocks/unstable, 30d TTL HSET+EXPIRE pipeline: 481,532,483,525 seconds -> avg -> **505.25 s**
kvrocks/unstable, 30d TTL using function: 549,499,491,488 seconds -> avg -> **506.75 s**
kvrocks/setExpire, 30d TTL with HSETEX: 398,348,342,393 -> avg -> **370.25s**

### DIFF
--- a/src/commands/cmd_hash.cc
+++ b/src/commands/cmd_hash.cc
@@ -265,7 +265,7 @@ class CommandHSetEx : public Commander {
     if (args.size() < 5) {
       return {Status::RedisParseErr, errWrongNumOfArguments};
     }
-    ttl_ = GET_OR_RET(ParseInt<int64_t>(args[2], 10));
+    ttl_ = GET_OR_RET(ParseInt<uint64_t>(args[2], 10));
     for (size_t i = 3; i < args_.size(); i += 2) {
       field_values_.emplace_back(args_[i], args_[i + 1]);
     }
--- a/src/commands/cmd_hash.cc
+++ b/src/commands/cmd_hash.cc
@@ -262,10 +262,10 @@ class CommandHMSet : public Commander {
 class CommandHSetEx : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    if (args.size() < 5) {
-      return {Status::RedisParseErr, errWrongNumOfArguments};
-    }
     ttl_ = GET_OR_RET(ParseInt<uint64_t>(args[2], 10));
+    if ((args.size() - 3) % 2 != 0) {
+      return {Status::RedisParseErr, "Invalid number of arguments: field-value pairs must be complete"};
+    }
     for (size_t i = 3; i < args_.size(); i += 2) {
       field_values_.emplace_back(args_[i], args_[i + 1]);
     }
@@ -479,7 +479,7 @@ REDIS_REGISTER_COMMANDS(Hash, MakeCmdAttr<CommandHGet>("hget", 3, "read-only", 1
                         MakeCmdAttr<CommandHIncrBy>("hincrby", 4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandHIncrByFloat>("hincrbyfloat", 4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandHMSet>("hset", -4, "write", 1, 1, 1),
-                        MakeCmdAttr<CommandHSetEx>("hsetex", -4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandHSetEx>("hsetex", -5, "write", 1, 1, 1),
                         MakeCmdAttr<CommandHSetNX>("hsetnx", -4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandHDel>("hdel", -3, "write no-dbsize-check", 1, 1, 1),
                         MakeCmdAttr<CommandHStrlen>("hstrlen", 3, "read-only", 1, 1, 1),
--- a/src/commands/cmd_hash.cc
+++ b/src/commands/cmd_hash.cc
@@ -259,7 +259,7 @@ class CommandHMSet : public Commander {
   std::vector<FieldValue> field_values_;
 };
 
-class CommandHSetEx : public Commander {
+class CommandHSetExpire : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
     ttl_ = GET_OR_RET(ParseInt<uint64_t>(args[2], 10));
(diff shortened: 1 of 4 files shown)


src/types/redis_hash.h
```diff
@@ -56,7 +56,7 @@ class Hash : public SubKeyScanner {
   rocksdb::Status IncrByFloat(engine::Context &ctx, const Slice &user_key, const Slice &field, double increment,
                               double *new_value);
   rocksdb::Status MSet(engine::Context &ctx, const Slice &user_key, const std::vector<FieldValue> &field_values,
-                       bool nx, uint64_t *added_cnt);
+                       bool nx, uint64_t *added_cnt, uint64_t ttl = -1);
```
PragmaTwice:
```suggestion
                       bool nx, uint64_t *added_cnt, uint64_t expire = 0);
```
better to be 0 here.

src/commands/cmd_hash.cc
```diff
@@ -258,6 +259,36 @@ class CommandHMSet : public Commander {
   std::vector<FieldValue> field_values_;
 };
 
+class CommandHSetEx : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() < 5) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+    ttl_ = GET_OR_RET(ParseInt<int64_t>(args[2], 10));
```
PragmaTwice:
```suggestion
    ttl_ = GET_OR_RET(ParseInt<uint64_t>(args[2], 10));
```

src/types/redis_hash.h
ltagliamonte-dd: honestly I choose `-1` bc it turns out that in redis you can set negative ttl 🤷‍♂️ 
0 is fine with me.

src/commands/cmd_hash.cc
ltagliamonte-dd: addressed via 76b25447b94878bfc75da31b30b6f7bcad5fce4f

src/types/redis_hash.h
ltagliamonte-dd: addressed via 76b25447b94878bfc75da31b30b6f7bcad5fce4f

src/commands/cmd_hash.cc
```diff
@@ -448,6 +479,7 @@ REDIS_REGISTER_COMMANDS(Hash, MakeCmdAttr<CommandHGet>("hget", 3, "read-only", 1
                         MakeCmdAttr<CommandHIncrBy>("hincrby", 4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandHIncrByFloat>("hincrbyfloat", 4, "write", 1, 1, 1),
                         MakeCmdAttr<CommandHMSet>("hset", -4, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandHSetEx>("hsetex", -4, "write", 1, 1, 1),
```
PragmaTwice:
```suggestion
                        MakeCmdAttr<CommandHSetEx>("hsetex", -5, "write", 1, 1, 1),
```

src/commands/cmd_hash.cc
```diff
@@ -258,6 +259,36 @@ class CommandHMSet : public Commander {
   std::vector<FieldValue> field_values_;
 };
 
+class CommandHSetEx : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() < 5) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
```
PragmaTwice:
```suggestion
```

src/commands/cmd_hash.cc
```diff
@@ -258,6 +259,36 @@ class CommandHMSet : public Commander {
...
+    ttl_ = GET_OR_RET(ParseInt<uint64_t>(args[2], 10));
+    for (size_t i = 3; i < args_.size(); i += 2) {
+      field_values_.emplace_back(args_[i], args_[i + 1]);
+    }
```
PragmaTwice, line 271: We should considered invalid input like `HSETEX key ttl k1 v1 k2` and reject it.

ltagliamonte-dd: addressed via f9551a6071bd5612b7a4633ff82eaa78a8a21798

src/commands/cmd_hash.cc
ltagliamonte-dd: addressed via f9551a6071bd5612b7a4633ff82eaa78a8a21798

src/commands/cmd_hash.cc
ltagliamonte-dd: addressed via f9551a6071bd5612b7a4633ff82eaa78a8a21798

PragmaTwice (commented):
Looks good to me. cc @git-hulk

git-hulk (dismissed):
LGTM, it would be great if we can add a few test cases for this command in the further PR.

ltagliamonte-dd:
> LGTM, it would be great if we can add a few test cases for this command in the further PR.

writing test cases, will push here as soon i'm done

PragmaTwice:
Yeah it's better to include test cases in this PR.

ltagliamonte-dd:
@PragmaTwice @git-hulk pushed unit tests.
thanks for the amazing support!

git-hulk (dismissed):

PragmaTwice:
I just found one issue about the command name: HSETEX.

HSETEX can be HSET + EXPIRE, but it also can be HSET + HEXPIRE (with the hash field expiration).

Redis doesn't have this command (and, potentially will add this command in the future).

Dragonfly has this command (https://www.dragonflydb.io/docs/command-reference/hashes/hsetex), which means HSET + HEXPIRE (e.g. `HSETEX key ttl k1 v1` will add the ttl to only k1 field, instead of the whole key).

So if we add this, there's a potentially semantic inconsistence: if Redis add it in the future and the Redis one means HSET + HEXPIRE, we'll have a trouble, i.e. we will be incompatible to Redis, otherwise incompatible to previous version of Kvrocks. (and also confuse our users.)

cc @git-hulk 

Maybe we can rename it to HSETEXPIRE.

git-hulk:
> Dragonfly has this command (https://www.dragonflydb.io/docs/command-reference/hashes/hsetex), which means HSET + HEXPIRE (e.g. HSETEX key ttl k1 v1 will add the ttl to only k1 field, instead of the whole key).

@PragmaTwice @ltagliamonte-dd 

Good finding. I didn't notice Dragonfly had this command. `HSETEXPIRE` sounds good to me.

PragmaTwice:
> Good finding. I didn't notice Dragonfly had this command. HSETEXPIRE sounds good to me.

Great! Let's change it to HSETEXPIRE. cc @ltagliamonte-dd

ltagliamonte-dd:
@PragmaTwice @git-hulk Rename addressed via 6f28ae05e7679983703c566b9a64936feb014663

PragmaTwice (approved):

PR #2759, opened 2025-02-01
build: remove glibc-specific dependencies

err931:
Resolves #2758

PR #2760, opened 2025-02-01
ci: add an alpine-based build & test workflow

PragmaTwice:
This is a follow-up PR after #2759, thanks to @err931.

We want to add an alpine-based build and test workflow in github actions, to ensure that kvrocks can always build against alpine, and musl c.

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -508,6 +508,7 @@ jobs:
         run: |
           apk update
           apk add bash cmake curl git python3 wget make gcc g++ autoconf linux-headers py3-pip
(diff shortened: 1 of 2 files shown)


aleksraiden:
I think, could be better use a more accuracy version, like alpine:3.21

PragmaTwice:
> I think, could be better use a more accuracy version, like alpine:3.21

I think it's fine. For archlinux we also didn't pin the version (also it cannot be pinned). So that we can quickly know if it cannot be built in the latest version.

git-hulk (approved):

danishgh (approved):

danishgh (approved):

danishgh (approved):

Issue #2642, opened 2024-11-03
Optimize the execution time of go test cases


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

In current GitHub Action workflows, go test execution now often takes more than 30 minutes, which may affect our responsiveness and PR processing efficiency.

We need to analyze which test cases take too long and optimize their execution time without affecting the test results.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

lyang24:
Hi i would like to take a shot with this issue. I ran the go test locally do you think its a good idea to take a look at the tests that rans over 20s?

```
ok  	github.com/apache/kvrocks/tests/gocase/integration/cluster	26.165s
ok  	github.com/apache/kvrocks/tests/gocase/integration/replication	119.125s
ok  	github.com/apache/kvrocks/tests/gocase/integration/slotmigrate	95.155s
ok  	github.com/apache/kvrocks/tests/gocase/unit/config	27.430s
ok  	github.com/apache/kvrocks/tests/gocase/unit/type/list	49.528s
ok  	github.com/apache/kvrocks/tests/gocase/unit/scripting	20.098s
ok  	github.com/apache/kvrocks/tests/gocase/unit/type/bitmap	34.487s
ok  	github.com/apache/kvrocks/tests/gocase/unit/type/zset	21.468s
```

PragmaTwice:
Yeah you can also pick a random workflow run and check the log, e.g.
https://github.com/apache/kvrocks/actions/runs/13096047337/job/36538680033#step:17:55

You can see that execution time of lots of tests can exceed 100s. And it's similar to the list you give.

lyang24:
We can try run tests in parallel. Then we can identify the bottle neck cases to further improve

For example i added some debug logging on github.com/apache/kvrocks/tests/gocase/integration/replication looks like
TestReplicationWithLimitSpeed is the bottle neck
```
TestReplicationAnnounceIP took 2.147078875s
TestShouldNotReplicate took 2.154190125s
TestReplicationContinueRunning took 2.247801667s
TestReplicationChangePassword took 3.255705917s
TestReplicationWithMultiSlaves took 3.894940417s
TestReplicationShareCheckpoint took 4.605201458s
TestFullSyncReplication took 10.106044083s
TestShouldNotReplicate took 10.585230084s
[...]

PR #2763, opened 2025-02-03
ci: execute long running go tests in parallel

lyang24:
As we can see from ci
https://github.com/apache/kvrocks/actions/runs/13096047337/job/36538680033#step:17:55

2 long runing go tests are low hanging fruit to be run in parallel 
```
ok  	github.com/apache/kvrocks/tests/gocase/integration/replication	110.239s
ok  	github.com/apache/kvrocks/tests/gocase/integration/cluster	26.476s
```

results after parallel execution in local
```
ok  	github.com/apache/kvrocks/tests/gocase/integration/replication	29.888s
ok  	github.com/apache/kvrocks/tests/gocase/integration/cluster	12.576s
```
max parallel running tests are set to 4 to avoid resource constraints

Touches: https://github.com/apache/kvrocks/issues/2642


PragmaTwice:
https://github.com/apache/kvrocks/actions/runs/13107232513/job/36565873699?pr=2763#step:17:62

Interesting that there's a server crash. Don't know if it is because of parallal test execution.

```
Replication thread operation failed: thread #140295113012800 cannot be `join`ed: Resource deadlock avoided
```

The replication thread failed to join the thread while `Stop()` is called (due to "Resource deadlock avoided"), so that the dtor of `std::thread` leads to a `std::terminate` (while the dtor of ReplicationThread is called).

This indicates that, the ReplicationThread::Stop() is called from the replication thread itself (which is a deadlock). (but it looks impossible.)

cc @git-hulk

PragmaTwice:
I think it should be due to that parallal execution is introduced for slotmigrate test cases. Maybe we can revert it for slot migration first, and create an issue for this problem.

lyang24:
> . Don't know if it is because of parallal test execution.
> 
> ```
> Replication thread operation failed: thread #140295113012800 cannot be `join`ed: Resource deadlock avoided
> ```
> 
> The replication thread failed to join the thread while `Stop()` is called (due to "Resource deadlock avoided"), so that the dtor of `std::thread` leads to a `std::terminate` (while the dtor of ReplicationThread is called).
> 
> This indicates that, the ReplicationThread::Stop() is called from the replication thread itself (which is a deadlock). (but it looks impossible.)

i guess it may due to the parallelism consumed too much system resources, will try to limit parallelism to 4. If this fails too will revert slot migraiton.

torwig:
Do our runners have enough memory to spawn all instances of Kvrocks we need to execute those tests in parallel?

danishgh (approved):

git-hulk:
> This indicates that, the ReplicationThread::Stop() is called from the replication thread itself (which is a deadlock). (but it looks impossible.)

@PragmaTwice From logs, it seems the replication thread was joined(due to receiving a TERM signal) before scheduling the thread. And we didn't check if it's joinable before joining the thread.

```
W20250203 07:12:02.381725 140295312238144 replication.cc:351] Clean old synced checkpoint successfully
I20250203 07:12:02.382069 140295312238144 cluster.cc:271] Change server role to slave, stop migration task
I20250203 07:12:02.382079 140295312238144 cluster.cc:273] SLAVE OF 127.0.0.1:38657 enabled by cluster topology setting
I20250203 07:12:02.382861 140295113012800 main.cc:50] Signal Terminated (15) received, stopping the server
W20250203 07:12:02.383481 140295113012800 replication.cc:371] Replication thread operation failed: thread #140295113012800 cannot be `join`ed: Resource deadlock avoided
I20250203 07:12:02.383501 140295113012800 replication.cc:373] [replication] Stopped
E20250203 07:12:02.427797 140295445802240 signal_util.h:33] Ooops! Apache Kvrocks unstable (commit c7ab45d3) got signal: Aborted (6)
```

lyang24:
looking at the ci failure here should we increase the wait time in `TestReplicationShareCheckpoint` to 100s to avoid flakeness?

```
			return master.LogFileMatches(t, ".*Using current existing checkpoint.*")
		}, 50*time.Second, 100*time.Millisecond)
```

git-hulk:
> looking at the ci failure here should we increase the wait time in TestReplicationShareCheckpoint to 100s to avoid flakeness?

How about reducing the parallel to 2? I'm not sure how much resource can be allocated in GitHub Action Runner, guess it won't allocate 4 cores for each runner.

git-hulk (dismissed):
LGTM

I checked the test case [replication](https://github.com/apache/kvrocks/actions/runs/13173723439/job/36769314824?pr=2763#step:17:47) costs about 50~60s for now. I think it's good to merge.

tests/gocase/unit/type/list/list_test.go
```diff
@@ -1053,6 +1053,10 @@ func testList(t *testing.T, configs util.KvrocksServerConfigs) {
 		require.EqualValues(t, 5, rdb.LPush(ctx, key2, "ONE", "TWO", "THREE", "FOUR", "FIVE").Val())
 
 		t.Run(fmt.Sprintf("LMPOP test oneKey countSingle %s", direction), func(t *testing.T) {
+			start := time.Now()
```
torwig: Did we leave this debug/dev message intentionally?

eyang-2: good catch, this is the fall out of other testing i did

torwig (dismissed):

git-hulk (dismissed):

PragmaTwice (dismissed):

lyang24:
looks like slot migrate tests are unstable with parallel maybe i could drop the changes from this pr

PragmaTwice:
> looks like slot migrate tests are unstable with parallel maybe i could drop the changes from this pr

Yeah let's drop it.

.github/workflows/kvrocks.yaml
```diff
@@ -312,7 +312,7 @@ jobs:
             cp minica.pem tests/gocase/tls/cert/ca.crt
             GOCASE_RUN_ARGS="-tlsEnable"
           fi
-          ./x.py test go build $GOCASE_RUN_ARGS ${{ matrix.ignore_when_tsan}}
+          ./x.py test go build $GOCASE_RUN_ARGS --parallel 2 ${{ matrix.ignore_when_tsan}}
```
PragmaTwice: I think we can also revert this line.

.github/workflows/kvrocks.yaml
```diff
@@ -554,7 +554,7 @@ jobs:
         run: |
           export PATH=$PATH:$HOME/local/bin/
           GOCASE_RUN_ARGS=""
-          ./x.py test go build $GOCASE_RUN_ARGS
+          ./x.py test go build $GOCASE_RUN_ARGS --parallel 2
```
PragmaTwice: ditto

lyang24:
> looking at the ci failure here should we increase the wait time in `TestReplicationShareCheckpoint` to 100s to avoid flakeness?
> 
> ```
> 	require.Eventually(t, func() bool {
> 			return master.LogFileMatches(t, ".*Using current existing checkpoint.*")
> 		}, 50*time.Second, 100*time.Millisecond)
> ```

we might need to keep the parallelism to ensure `TestReplicationShareCheckpoint` has sufficient cpu

PragmaTwice (approved):

git-hulk (approved):

PR #2764, opened 2025-02-03
feat(rockdb): Allow customization of Rocksdb threads number

ltagliamonte-dd:
Allow customization of rocksdb low and high thread number

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -130,8 +130,8 @@ Status SetRocksdbCompression(Server *srv, const rocksdb::CompressionType compres
   for (size_t i = compression_start_level; i < KVROCKS_MAX_LSM_LEVEL; i++) {
     compression_per_level_builder.emplace_back(compression_option);
   }
-  const std::string compression_per_level = util::StringJoin(
-      compression_per_level_builder, [](const auto &s) -> decltype(auto) { return s; }, ":");
+  const std::string compression_per_level =
+      util::StringJoin(compression_per_level_builder, [](const auto &s) -> decltype(auto) { return s; }, ":");
   return srv->storage->SetOptionForAllColumnFamilies("compression_per_level", compression_per_level);
 };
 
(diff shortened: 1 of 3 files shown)


torwig:
@ltagliamonte-dd Could you please elaborate on the change? I see that we can set max_background_compactions and max_background_flushes. It's not enough? Or it's something different?

ltagliamonte-dd:
> @ltagliamonte-dd Could you please elaborate on the change? I see that we can set max_background_compactions and max_background_flushes. It's not enough? Or it's something different?

Hello @torwig, I'm very new to rocksdb, I may be wrong, but:
max_background_jobs/max_background_compactions/max_background_flushes are higher-level setting that controls the total number of concurrent background operations that RocksDB will perform.

SetBackgroundThreads controls the size of the thread pools separately for HIGH priority (typically compactions) and LOW priority (typically flushes) operations.

this is my understanding reading this [doc](https://github.com/facebook/rocksdb/wiki/RocksDB-Tuning-Guide#parallelism-options)

so for optimal perfomances you should configure both.

git-hulk:
@ltagliamonte-dd You could just control the compaction/flush thread number by configuring:

- rocksdb.max_background_compactions (default is LOW priority)
- rocksdb.max_background_flushes (default is HIGH priority)

But those two parameters might be deprecated in future versions. It will use 1/4 * max_background_jobs as the flush thread number if both of them are not set. See [GetBGJobLimits](https://github.com/facebook/rocksdb/blob/ce6065ef70fe8c55bdf74c2a4121c038a1f7da4b/db/db_impl/db_impl_compaction_flush.cc#L2935)

ltagliamonte-dd:
@git-hulk I may be reading this wrong, but where is it the correlation between this params (max_background_jobs/max_background_compactions/max_background_flushes) and the number of threads in rocksdb codebase?

In this example in the doc I see the setting of the threadpool size and max_background_jobs
https://github.com/facebook/rocksdb/wiki/Thread-Pool

git-hulk:
> where is it the correlation between this params (max_background_jobs/max_background_compactions/max_background_flushes) and the number of threads in rocksdb codebase?

The correlation, you could refer: [GetBGJobLimits](https://github.com/facebook/rocksdb/blob/ce6065ef70fe8c55bdf74c2a4121c038a1f7da4b/db/db_impl/db_impl_compaction_flush.cc#L2935)

danishgh (approved):

ltagliamonte-dd:
Closing in favor of using max_background_compactions and max_background_flushes has the same effect.

PR #2765, opened 2025-02-03
fix(conn): detect and drop HTTP requests in connection

PragmaTwice:
Refer to https://github.com/redis/redis/blob/870b6bd487ec4623af458d0a660b2854dee29d5a/src/server.c#L3997, we should do similiar HTTP request detection as in Redis.

### DIFF
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -385,10 +385,17 @@ void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
 
     auto cmd_s = Server::LookupAndCreateCommand(cmd_tokens.front());
     if (!cmd_s.IsOK()) {
+      auto cmd_name = cmd_tokens.front();
+      if (util::EqualICase(cmd_name, "host:") || util::EqualICase(cmd_name, "post")) {
+        LOG(WARNING) << "A likely HTTP request is detected in the RESP connection, indicating a potential "
+                        "Cross-Protocol Scripting attack. Connection aborted.";
+        owner_->FreeConnection(this);
+        break;
+      }
       if (is_multi_exec) multi_error_ = true;
       Reply(redis::Error(
           {Status::NotOK,
-           fmt::format("unknown command `{}`, with args beginning with: {}", cmd_tokens.front(),
+           fmt::format("unknown command `{}`, with args beginning with: {}", cmd_name,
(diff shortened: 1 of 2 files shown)


torwig:
@PragmaTwice Ideally is to add a Go test to verify the behavior.

PragmaTwice:
> @PragmaTwice Ideally is to add a Go test to verify the behavior.

Added.

src/server/redis_connection.cc
```diff
@@ -385,10 +385,17 @@ void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
 
     auto cmd_s = Server::LookupAndCreateCommand(cmd_tokens.front());
     if (!cmd_s.IsOK()) {
+      auto cmd_name = cmd_tokens.front();
+      if (util::EqualICase(cmd_name, "host:") || util::EqualICase(cmd_name, "post")) {
+        LOG(WARNING) << "A likely HTTP request is detected in the RESP connection, indicating a potential "
+                        "Cross-Protocol Scripting attack. Connection aborted.";
+        owner_->FreeConnection(this);
+        break;
```
git-hulk: I think we can `Close` the connection and return here. To prevent wrongly using the connection outside the while loop.

PragmaTwice: Done.

git-hulk (dismissed):

torwig (dismissed):

src/server/redis_connection.cc
```diff
@@ -385,10 +385,17 @@ void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
 
     auto cmd_s = Server::LookupAndCreateCommand(cmd_tokens.front());
     if (!cmd_s.IsOK()) {
+      auto cmd_name = cmd_tokens.front();
+      if (util::EqualICase(cmd_name, "host:") || util::EqualICase(cmd_name, "post")) {
+        LOG(WARNING) << "A likely HTTP request is detected in the RESP connection, indicating a potential "
+                        "Cross-Protocol Scripting attack. Connection aborted.";
+        Close();
```
git-hulk:
```suggestion
        EnableFlag(kCloseAsync);
```
Sorry for misleading. Should use the `kCloseAsync` flag to avoid use-heap-after-free error due to `ExecuteCommands` was called inside conn
s read callback.

git-hulk (approved):

PR #2783, opened 2025-02-10
fix(string): the server will be crashed when passing a negative offset in SETRANGE command

git-hulk:
Currently, we didn't check if the offset is a negative value and using it to index the value string will crash the server.

### DIFF
--- a/src/cli/main.cc
+++ b/src/cli/main.cc
@@ -109,6 +109,7 @@ static void InitGoogleLog(const Config *config) {
   } else {
     FLAGS_log_dir = config->log_dir + "/";
     if (config->log_retention_days != -1) {
+      google::EnableLogCleaner(std::chrono::hours(24) * config->log_retention_days);
     }
   }
 }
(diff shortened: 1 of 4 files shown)


torwig (dismissed):

PragmaTwice (dismissed):

src/cli/main.cc
```diff
@@ -109,7 +109,6 @@ static void InitGoogleLog(const Config *config) {
   } else {
     FLAGS_log_dir = config->log_dir + "/";
     if (config->log_retention_days != -1) {
-      google::EnableLogCleaner(std::chrono::hours(24) * config->log_retention_days);
```
torwig: We don't need this line?

git-hulk: Yes, I forget to revert this change. This line throws an error on my macOS when building. I workaround this by remove it temporially.

torwig (approved):

PragmaTwice (approved):

PR #2801, opened 2025-02-22
chore: remove redundant code in the method `Execute` of `slaveof` cmd

cxljs:
`slaveof` cmd cannot be executed in cluster mode, and we check it at the beginning of the `Execute()` .
So we don't need to check it again later.


torwig (approved):

git-hulk (approved):

git-hulk:
@fukua95 Thank you!

PR #2802, opened 2025-02-24
feat(config): allow to config the rocksdb's max compaction bytes

git-hulk:
`rocksdb.max_compaction_bytes` could be used to limit the max compaction bytes in a single compaction run to prevent too much write and space amplification.


PragmaTwice (approved):

torwig (approved):

git-hulk:
@PragmaTwice @torwig Thank you!

PR #2803, opened 2025-02-25
feat(tdigest): add TDIGEST.ADD command

### LINKED ISSUE #2792: TDigest: Implement ADD command for TDigest Algorithm
### Search before asking

### Motivation

We should implement `TDIGEST.ADD` command.

Redis command reference:

https://redis.io/docs/latest/commands/tdigest.add/

### Solution

Depends on #2791.

We can follow the implementation of #2791.

wyxxxcat:
# Issue
close: https://github.com/apache/kvrocks/issues/2792
# Proposed Changes
- implement tdigest.add command.
- add test
# Comment
...

### DIFF
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -131,6 +131,44 @@ class CommandTDigestInfo : public Commander {
   std::string key_name_;
 };
 
+class CommandTDigestAdd : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    key_name_ = args[1];
+
+    values_.reserve(args.size() - 2);
+    for (size_t i = 2; i < args.size(); i++) {
+      auto value = ParseFloat(args[i]);
+      if (!value) {
+        return {Status::RedisParseErr, errValueIsNotFloat};
+      }
+      values_.push_back(*value);
+    }
+
+    return Status::OK();
+  }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    TDigest tdigest(srv->storage, conn->GetNamespace());
+
+    auto s = tdigest.Add(ctx, key_name_, values_);
+    if (!s.ok()) {
+      if (s.IsNotFound()) {
+        return {Status::RedisExecErr, errKeyNotFound};
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    *output = redis::RESP_OK;
+    return Status::OK();
+  }
+
+ private:
+  std::string key_name_;
+  std::vector<double> values_;
+};
+
 REDIS_REGISTER_COMMANDS(TDigest, MakeCmdAttr<CommandTDigestCreate>("tdigest.create", -2, "write", 1, 1, 1),
-                        MakeCmdAttr<CommandTDigestInfo>("tdigest.info", 2, "read-only", 1, 1, 1));
+                        MakeCmdAttr<CommandTDigestInfo>("tdigest.info", 2, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandTDigestAdd>("tdigest.add", -3, "write", 1, 1, 1));
 }  // namespace redis
--- a/src/types/redis_tdigest.cc
+++ b/src/types/redis_tdigest.cc
@@ -147,7 +147,6 @@ rocksdb::Status TDigest::Create(engine::Context& ctx, const Slice& digest_name, 
 
 rocksdb::Status TDigest::Add(engine::Context& ctx, const Slice& digest_name, const std::vector<double>& inputs) {
   auto ns_key = AppendNamespacePrefix(digest_name);
-  LockGuard guard(storage_->GetLockManager(), ns_key);
 
   TDigestMetadata metadata;
   if (auto status = getMetaDataByNsKey(ctx, ns_key, &metadata); !status.ok()) {
--- a/tests/gocase/unit/type/tdigest/tdigest_test.go
+++ b/tests/gocase/unit/type/tdigest/tdigest_test.go
@@ -141,4 +141,66 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
 			require.EqualValues(t, 0, info.TotalCompressions)
 		}
 	})
+
+	t.Run("tdigest.add with different arguments", func(t *testing.T) {
+		keyPrefix := "tdigest_add_"
+
+		// Satisfy the number of parameters
+		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.ADD").Err(), errMsgWrongNumberArg)
+		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.ADD", keyPrefix+"key").Err(), errMsgWrongNumberArg)
+		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.ADD", keyPrefix+"key", "abc").Err(), "not a valid float")
+		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.ADD", keyPrefix+"nonexistent", "1.0").Err(), errMsgKeyNotExist)
+
+		// Test adding values to a key
+		key := keyPrefix + "test1"
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.CREATE", key, "compression", "100").Err())
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.ADD", key, "42.0").Err())
+
+		rsp := rdb.Do(ctx, "TDIGEST.INFO", key)
+		require.NoError(t, rsp.Err())
+		info := toTdigestInfo(t, rsp.Val())
+		require.EqualValues(t, 1, info.UnmergedNodes)
+		require.EqualValues(t, 1, info.Observations)
+
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.ADD", key, "1.0", "2.0", "3.0", "4.0", "5.0").Err())
+
+		rsp = rdb.Do(ctx, "TDIGEST.INFO", key)
+		require.NoError(t, rsp.Err())
+		info = toTdigestInfo(t, rsp.Val())
+		require.EqualValues(t, 6, info.Observations)
+
+		// Test adding values to a key with compression
+		key2 := keyPrefix + "test2"
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.CREATE", key2, "compression", "100").Err())
+
+		args := []interface{}{key2}
+		for i := 1; i <= 1000; i++ {
+			args = append(args, float64(i))
+		}
+		require.NoError(t, rdb.Do(ctx, append([]interface{}{"TDIGEST.ADD"}, args...)...).Err())
+
+		rsp = rdb.Do(ctx, "TDIGEST.INFO", key2)
+		require.NoError(t, rsp.Err())
+		info = toTdigestInfo(t, rsp.Val())
+		require.EqualValues(t, 1000, info.Observations)
+
+		// Test adding values to a key with compression and merge node
+		key3 := keyPrefix + "test3"
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.CREATE", key3, "compression", "10").Err())
+
+		args = []interface{}{key3}
+		for i := 1; i <= 100; i++ {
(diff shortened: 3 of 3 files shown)


src/commands/cmd_tdigest.cc
```diff
@@ -131,6 +131,54 @@ class CommandTDigestInfo : public Commander {
...
+
+    *output = redis::RESP_OK;
+    return Status::OK();
+  }
+
+ private:
+  std::string key_name_;
+  std::vector<double> values_;
+
+  static StatusOr<double> parseFloat(const std::string &str) {
```
mapleFU: Why not call https://github.com/apache/kvrocks/blob/e00fcf10621f242f6de4bf4602d79da8b9cfcc10/src/common/parse_util.h#L198 ?

src/commands/cmd_tdigest.cc
```diff
@@ -131,6 +131,54 @@ class CommandTDigestInfo : public Commander {
   std::string key_name_;
 };
 
+class CommandTDigestAdd : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() < 3) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+    key_name_ = args[1];
```
mapleFU, line 137: why store the key_name rather than just uses args[1]?

src/commands/cmd_tdigest.cc
```diff
@@ -131,6 +131,54 @@ class CommandTDigestInfo : public Commander {
...
+
+    for (size_t i = 2; i < args.size(); i++) {
+      auto value = parseFloat(args[i]);
+      if (!value) {
+        return {Status::RedisParseErr, errValueIsNotFloat};
+      }
+      values_.push_back(*value);
```
mapleFU, line 145: Reserve firstly since it's likely to success

mapleFU (commented):
General LGTM

tests/gocase/unit/type/tdigest/tdigest_test.go
```diff
@@ -141,4 +141,45 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
...
+	
+		key := keyPrefix + "test1"
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.CREATE", key, "compression", "100").Err())
+	
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.ADD", key, "42.0").Err())
+		
+		rsp := rdb.Do(ctx, "TDIGEST.INFO", key)
+		require.NoError(t, rsp.Err())
+		info := toTdigestInfo(t, rsp.Val())
+		require.EqualValues(t, 1, info.UnmergedNodes)
```
LindaSummer, line 162: If we want to test this property here, we'd better add a case that triggers the merge and test both the `Merged nodes`, `Unmerged nodes` and `Total compressions`. (The `Total compressions` is the times of merge triggered)
We can just add some simple test cases here and I will do a more complex one in `tdigest.quantile` command.

src/commands/cmd_tdigest.cc
```diff
@@ -131,6 +131,54 @@ class CommandTDigestInfo : public Commander {
...
+    } catch (const std::exception &) {
+      return {Status::RedisParseErr, errValueIsNotFloat};
+    }
+  }
+};
+
 REDIS_REGISTER_COMMANDS(TDigest, MakeCmdAttr<CommandTDigestCreate>("tdigest.create", -2, "write", 1, 1, 1),
-                        MakeCmdAttr<CommandTDigestInfo>("tdigest.info", 2, "read-only", 1, 1, 1));
+                        MakeCmdAttr<CommandTDigestInfo>("tdigest.info", 2, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandTDigestAdd>("tdigest.add", -3, "write", 2, -1, 1));
```
LindaSummer: It seems that the 5th argument is confused to be `-1` here.
It should be the `last_key` of command's key.

LindaSummer (commented):
Hi @wyxxxcat ,

Generally LGTM, left some comments.

Best Regards,
Edward

src/commands/cmd_tdigest.cc
wyxxxcat: thx,it's a good ida

src/commands/cmd_tdigest.cc
wyxxxcat: improve readability when calling `tdigest.Add` function

src/commands/cmd_tdigest.cc
wyxxxcat: thx,it's a good idea

src/commands/cmd_tdigest.cc
wyxxxcat: thanks for your suggestion
```
  // index of the last key in command tokens
  // in normal one-key commands, first key and last key index are both 1
  // -n stands for the n-th last index of the sequence, i.e. args.size() - n
  int last_key;
```
It looks like `last_key = -1` which can contain the entire sequence

tests/gocase/unit/type/tdigest/tdigest_test.go
wyxxxcat: thanks !
I will improve it

src/commands/cmd_tdigest.cc
```diff
@@ -131,6 +131,54 @@ class CommandTDigestInfo : public Commander {
   std::string key_name_;
 };
 
+class CommandTDigestAdd : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() < 3) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
```
PragmaTwice:
```suggestion
```

It will be checked in other places.

src/commands/cmd_tdigest.cc
PragmaTwice:
```suggestion
                        MakeCmdAttr<CommandTDigestAdd>("tdigest.add", -3, "write", 1, 1, 1));
```

src/commands/cmd_tdigest.cc
```diff
@@ -131,6 +131,54 @@ class CommandTDigestInfo : public Commander {
...
+  std::string key_name_;
+  std::vector<double> values_;
+
+  static StatusOr<double> parseFloat(const std::string &str) {
+    try {
+      return std::stod(str);
+    } catch (const std::exception &) {
+      return {Status::RedisParseErr, errValueIsNotFloat};
+    }
+  }
```
PragmaTwice: You can use `ParseFloat` in `parse_util.h` instead.

src/commands/cmd_tdigest.cc
PragmaTwice: The range is for Redis keys, which is `args[1]` here.

src/commands/cmd_tdigest.cc
wyxxxcat: thanks for your suggestions

src/commands/cmd_tdigest.cc
```diff
@@ -131,6 +131,44 @@ class CommandTDigestInfo : public Commander {
...
 };
 
+class CommandTDigestAdd : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    key_name_ = args[1];
+    values_.reserve(args.size());
+
+    for (size_t i = 2; i < args.size(); i++) {
+      auto value = ParseFloat<>(args[i]);
```
PragmaTwice:
```suggestion
      auto value = ParseFloat(args[i]);
```

I prefer to remove this `<>`.

LindaSummer:
Hi @wyxxxcat ,

You could enable your fork repo's Github Action and it will trigger ci pipeline inside your own repo without the upstream repo's permission.

It may accelerate your development and review process. 😁

Best Regards,
Edward

src/commands/cmd_tdigest.cc
```diff
@@ -131,6 +131,44 @@ class CommandTDigestInfo : public Commander {
   std::string key_name_;
 };
 
+class CommandTDigestAdd : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    key_name_ = args[1];
+
+    values_.reserve(args.size());
```
mapleFU: args.size() - 2?

src/types/redis_tdigest.cc
```diff
@@ -147,7 +147,6 @@ rocksdb::Status TDigest::Create(engine::Context& ctx, const Slice& digest_name,
 
 rocksdb::Status TDigest::Add(engine::Context& ctx, const Slice& digest_name, const std::vector<double>& inputs) {
   auto ns_key = AppendNamespacePrefix(digest_name);
-  LockGuard guard(storage_->GetLockManager(), ns_key);
```
mapleFU, line 150: May I ask the reason for not guard here?

wyxxxcat: deadlock problem will occur

src/commands/cmd_tdigest.cc
wyxxxcat: try to leave enough space

mapleFU: The maximum space for this is `args.size() - 2`, isn't it? It never reaches `args.size()`.

src/types/redis_tdigest.cc
mapleFU: Conflict with which lock 🤔?

PragmaTwice: I think it should be this one: https://github.com/apache/kvrocks/blob/e00fcf10621f242f6de4bf4602d79da8b9cfcc10/src/server/redis_connection.cc#L514.

Refer to https://github.com/apache/kvrocks/commit/1e25be6a4e20242c6f87c7b561f87666c6359437.

mapleFU: You're right, I forgot the change here!

mapleFU (approved):

mapleFU:
Will merge if no negative comments tomorrow

src/commands/cmd_tdigest.cc
```diff
@@ -131,6 +131,44 @@ class CommandTDigestInfo : public Commander {
...
+class CommandTDigestAdd : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    key_name_ = args[1];
+
+    values_.reserve(args.size() - 2);
+    for (size_t i = 2; i < args.size(); i++) {
+      auto value = ParseFloat(args[i]);
+      if (!value) {
+        return {Status::RedisParseErr, errValueIsNotFloat};
```
LindaSummer, line 143: I test this case in redis stack latest image, the error message is `error parsing val parameter`.

Maybe we could follow current redis's error message if possible.😊

tests/gocase/unit/type/tdigest/tdigest_test.go
```diff
@@ -141,4 +141,66 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
...
 		}
 	})
+
+	t.Run("tdigest.add with different arguments", func(t *testing.T) {
+		keyPrefix := "tdigest_add_"
+
+		// Satisfy the number of parameters
+		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.ADD").Err(), errMsgWrongNumberArg)
+		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.ADD", keyPrefix+"key").Err(), errMsgWrongNumberArg)
+		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.ADD", keyPrefix+"key", "abc").Err(), "not a valid float")
```
LindaSummer, line 151: Maybe we could follow current redis stack's error message to keep compatibility. 😊

LindaSummer (commented):
Hi @wyxxxcat ,

I left comment for error message compalibility with redis. I think it's not mandatory for tdigest command and current messages are reasonable, but will be better to keep compalibility.

Other parts are LGTM.

Best Regards,
Edward

tests/gocase/unit/type/tdigest/tdigest_test.go
wyxxxcat: thanks for your suggestions, I will pay attention to the following work

PragmaTwice:
Let's merge it first. Improvement can happen in further PRs.

PR #2811, opened 2025-02-28
feat(tdigest): add TDIGEST.MAX/MIN command

### LINKED ISSUE #2793: TDigest: Implement MIN MAX command for TDigest Algorithm
### Search before asking

### Motivation

Refer: #2558
We should implement TDIGEST.MIN TDIGEST.MAX command.

Redis command reference:

- https://redis.io/docs/latest/commands/tdigest.min/
- https://redis.io/docs/latest/commands/tdigest.max/

### Solution

Depends on #2791 

We can follow the implementation of #2791 for this command.

To write more comprehensive integration tests, it will be better to be implemented after #2792 .

wyxxxcat:
This closes #2793.

### DIFF
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -18,8 +18,6 @@
  *
  */
 
-#include <type_traits>
-
 #include "command_parser.h"
 #include "commander.h"
 #include "server/redis_reply.h"
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -18,6 +18,8 @@
  *
  */
 
+#include <type_traits>
+
 #include "command_parser.h"
 #include "commander.h"
 #include "server/redis_reply.h"
@@ -168,7 +170,55 @@ class CommandTDigestAdd : public Commander {
   std::vector<double> values_;
 };
 
+class CommandTDigestMinMax : public Commander {
+ public:
+  explicit CommandTDigestMinMax(bool is_min) : is_min_(is_min) {}
+
+  Status Parse(const std::vector<std::string> &args) override {
+    key_name_ = args[1];
+    return Status::OK();
+  }
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    TDigest tdigest(srv->storage, conn->GetNamespace());
+    TDigestMetadata metadata;
+    auto s = tdigest.GetMetaData(ctx, key_name_, &metadata);
+    if (!s.ok()) {
+      if (s.IsNotFound()) {
+        return {Status::RedisExecErr, errKeyNotFound};
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    if (metadata.total_observations == 0) {
+      *output = redis::BulkString("nan");
+      return Status::OK();
+    }
+
+    double value = is_min_ ? metadata.minimum : metadata.maximum;
+    *output = redis::BulkString(fmt::format("{}", value));
+    return Status::OK();
+  }
+
+ private:
+  std::string key_name_;
+  bool is_min_;
+};
+
+// Then replace the existing template implementation and type aliases with:
+class CommandTDigestMin : public CommandTDigestMinMax {
+ public:
+  CommandTDigestMin() : CommandTDigestMinMax(true) {}
+};
+
+class CommandTDigestMax : public CommandTDigestMinMax {
+ public:
+  CommandTDigestMax() : CommandTDigestMinMax(false) {}
+};
+
 REDIS_REGISTER_COMMANDS(TDigest, MakeCmdAttr<CommandTDigestCreate>("tdigest.create", -2, "write", 1, 1, 1),
                         MakeCmdAttr<CommandTDigestInfo>("tdigest.info", 2, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandTDigestAdd>("tdigest.add", -3, "write", 1, 1, 1));
+                        MakeCmdAttr<CommandTDigestAdd>("tdigest.add", -3, "write", 1, 1, 1),
+                        MakeCmdAttr<CommandTDigestMax>("tdigest.max", 2, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandTDigestMin>("tdigest.min", 2, "read-only", 1, 1, 1));
 }  // namespace redis
--- a/src/types/redis_tdigest.cc
+++ b/src/types/redis_tdigest.cc
@@ -161,6 +161,8 @@ rocksdb::Status TDigest::Add(engine::Context& ctx, const Slice& digest_name, con
 
   metadata.total_observations += inputs.size();
   metadata.total_weight += inputs.size();
+  metadata.maximum = std::max(metadata.maximum, *std::max_element(inputs.cbegin(), inputs.cend()));
+  metadata.minimum = std::min(metadata.minimum, *std::min_element(inputs.cbegin(), inputs.cend()));
 
   if (metadata.unmerged_nodes + inputs.size() <= metadata.capacity) {
     if (auto status = appendBuffer(ctx, batch, ns_key, inputs, &metadata); !status.ok()) {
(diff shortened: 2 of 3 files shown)


src/commands/cmd_tdigest.cc
```diff
@@ -168,7 +168,69 @@ class CommandTDigestAdd : public Commander {
...
+    auto s = tdigest.GetMetaData(ctx, key_name_, &metadata);
+    if (!s.ok()) {
+      if (s.IsNotFound()) {
+        return {Status::RedisExecErr, errKeyNotFound};
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    if (metadata.total_observations == 0) {
+      return {Status::RedisExecErr, errValueIsNotExists};
```
mapleFU: https://github.com/RedisBloom/RedisBloom/blob/f68a72e463b84c2298990295c6a3db4719dc5c7a/src/rm_tdigest.c#L386

What about?

src/commands/cmd_tdigest.cc
```diff
@@ -168,7 +168,69 @@ class CommandTDigestAdd : public Commander {
...
```
mapleFU: ditto

wyxxxcat: done

mapleFU (commented):
Code general LGTM, but suddenly I realized that `TDigest.Add` nan is invalid: https://github.com/RedisBloom/RedisBloom/blob/f68a72e463b84c2298990295c6a3db4719dc5c7a/src/rm_tdigest.c#L187

Would you mind also change that? ( Can be in another pr )

wyxxxcat:
> Code general LGTM, but suddenly I realized that `TDigest.Add` nan is invalid: https://github.com/RedisBloom/RedisBloom/blob/f68a72e463b84c2298990295c6a3db4719dc5c7a/src/rm_tdigest.c#L187
> 
> Would you mind also change that? ( Can be in another pr )

I will fix

mapleFU (dismissed):

src/commands/cmd_tdigest.cc
```diff
@@ -168,7 +168,71 @@ class CommandTDigestAdd : public Commander {
   std::vector<double> values_;
 };
```
mapleFU, line 170: By the way would you mind extract an:

```

struct MinMax {
   min, max
};
class CommandTDigestMinMaxImpl {
   StatusOr<MinMax> Execute(...)
};
```
? I just think this could eliminate some code

wyxxxcat: yes！I will pick it

src/types/redis_tdigest.cc
```diff
@@ -161,6 +161,8 @@ rocksdb::Status TDigest::Add(engine::Context& ctx, const Slice& digest_name, con
 
   metadata.total_observations += inputs.size();
   metadata.total_weight += inputs.size();
+  metadata.maximum = std::max(metadata.maximum, *std::max_element(inputs.cbegin(), inputs.cend()));
```
LindaSummer: Hi @wyxxxcat ,

The `metadata.minimum` and `metadata.maximum` should represent the centroids min and max.
It will be updated in [merge action](https://github.com/apache/kvrocks/blob/34007db3a5a74e7c773209a16c6d0a5cc09ef884/src/types/tdigest.cc#L238) and updated by [dumping after merging](https://github.com/apache/kvrocks/blob/34007db3a5a74e7c773209a16c6d0a5cc09ef884/src/types/tdigest.cc#L393).

Maybe we could just compute inplace or create a new property in metadata to keep the consistency of metadata and centroids. 😊

If we try to do this in buffer appending, we may need to refactor the merge logic to make sure these max min properties are maintained by just one place.

LindaSummer (commented):
Hi @wyxxxcat ,

Leave one comment and other parts are generally LGTM. 😊

Best Regards,
Edward

src/commands/cmd_tdigest.cc
LindaSummer: Hi @mapleFU ,

Thanks very much for this link.❤️

I had seeking for a long time for redis's native tdigest implementation during my development but didn't find it. 😆

Best Regards,
Edward

src/types/redis_tdigest.cc
mapleFU: I don't think this matters since `DumpCentroids()` is always called after Merge currently?

LindaSummer: Hi @mapleFU ,

Thanks for your update! 😊

Currently the min and max are updated after Merge with `DumpCentroids()`.

But `Add` will not always trigger the Merge action until the buffer is full.

So, if we try to update it in `Add` maybe we'd better remove the [updating of min&max after `DumpCentroids()`](https://github.com/apache/kvrocks/blob/34007db3a5a74e7c773209a16c6d0a5cc09ef884/src/types/redis_tdigest.cc#L282-L283).

Since when we try to merge, the buffer will always be locked, it is safe to update the value either in `Add` or `Merge`.

Best Regards,
Edward

mapleFU: To change the word here, what's the main purpose to set min max after merge🤔? Since update min max is free when calling Add if any io really happens. And `Quantile` will always merge before doing any real things?

Without updating min max in add, the `tdigest.min/max` will traverse the unmerged data buffer

Both way is ok to me here but I just don't know the point or trade-off here

LindaSummer: Hi @mapleFU ,

I agree on your view. 😊
Frankly this probably just because I implement the `Merge` function ahead of `Add`. 😄

So, I think move it to the `Add` function is reasonable. But we'd better remove the updating after Merge if we decide to maintain it in `Add`.

Update it in two places may not be a good practice.

Best Regards,
Edward

PragmaTwice: Hmm so I'm wondering if we should submit a PR to make these changes before this PR is merged?

LindaSummer: Hi @PragmaTwice ,

Got it! 
I'll create a patch today to refactor the min max updating logic to `Add`.😊

Best Regards,
Edward

mapleFU:
I would approve and merge after https://github.com/apache/kvrocks/pull/2811#discussion_r1975049547 is applied

PragmaTwice (commented):
Looks good to me. To see if @LindaSummer has any comment.

src/commands/cmd_tdigest.cc
```diff
@@ -168,7 +168,71 @@ class CommandTDigestAdd : public Commander {
...
+    auto s = tdigest.GetMetaData(ctx, key_name_, &metadata);
+    if (!s.ok()) {
+      if (s.IsNotFound()) {
+        return {Status::RedisExecErr, errKeyNotFound};
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    if (metadata.total_observations == 0) {
+      *output = redis::BulkString(fmt::format("{}", NAN));
```
LindaSummer: Hi @wyxxxcat ,

Maybe it could be a constant string. 😊

Best Regards,
Edward

src/commands/cmd_tdigest.cc
```diff
@@ -168,7 +168,71 @@ class CommandTDigestAdd : public Commander {
...
```
LindaSummer: same as above.😊

LindaSummer (commented):
Hi @wyxxxcat ,

Thanks for your effort!😊
Generally LGTM.

Left some minor nitpick.
Current code is also good to me.

If you don't mind, you could update it in next commit.

Best Regards,
Edward

src/commands/cmd_tdigest.cc
wyxxxcat: thanks, I will improve it in my free time

src/commands/cmd_tdigest.cc
```diff
@@ -168,7 +170,52 @@ class CommandTDigestAdd : public Commander {
   std::vector<double> values_;
 };
 
+enum MinMax { MAX, MIN };
+
+template <MinMax T>
```
mapleFU: We can make this simplier and just using argument? This is not a cpu-bounded class, and I guess the performance here is not matter

The compute function uses template because that's cpu bound and branching matters, here we can just take it easy

src/commands/cmd_tdigest.cc
```diff
@@ -18,6 +18,8 @@
  *
  */
 
+#include <type_traits>
+
```
mapleFU:
```suggestion
```

mapleFU (approved):

mapleFU:
Thanks all, merged!

By the way, meaning full commit message and not using same commit is appreciated

Issue #2806, opened 2025-02-26
Flaky test when terminating the server


git-hulk:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

```
I20250226 11:50:40.185284 140031354014976 main.cc:144] kvrocks unstable (commit b7005d93)
I20250226 11:50:40.212890 140031354014976 storage.cc:396] [storage] Success to load the data from disk: 12 ms
I20250226 11:50:40.213458 140031354014976 worker.cc:75] [worker] Listening on: 127.0.0.1:35315
I20250226 11:50:40.213650 140031354014976 worker.cc:75] [worker] Listening on: 127.0.0.1:35315
I20250226 11:50:40.213767 140031354014976 worker.cc:75] [worker] Listening on: 127.0.0.1:35315
I20250226 11:50:40.213874 140031354014976 worker.cc:75] [worker] Listening on: 127.0.0.1:35315
I20250226 11:50:40.213977 140031354014976 worker.cc:75] [worker] Listening on: 127.0.0.1:35315
I20250226 11:50:40.214076 140031354014976 worker.cc:75] [worker] Listening on: 127.0.0.1:35315
I20250226 11:50:40.214189 140031354014976 worker.cc:75] [worker] Listening on: 127.0.0.1:35315
I20250226 11:50:40.214285 140031354014976 worker.cc:75] [worker] Listening on: 127.0.0.1:35315
I20250226 11:50:40.214372 140031354014976 cluster.cc:688] The cluster nodes file /home/runner/work/kvrocks/kvrocks/tests/gocase/workspace/TestSlaveLostMaster-1740570640181-3921765877/nodes.conf is not found. Use CLUSTERX subcommands to specify it.
I20250226 11:50:40.214458 140031354014976 worker.cc:593] [worker] Thread #140031226283584 started
I20250226 11:50:40.214511 140031354014976 worker.cc:593] [worker] Thread #140031207409216 started
I20250226 11:50:40.214[55](https://github.com/apache/kvrocks/actions/runs/13541357113/job/37847608993?pr=2803#step:20:56)8 140031354014976 worker.cc:593] [worker] Thread #140031188534848 started
I20250226 11:50:40.214635 140031354014976 worker.cc:593] [worker] Thread #140031169660480 started
I20250226 11:50:40.214695 140031354014976 worker.cc:593] [worker] Thread #140031150786112 started
[...]

PR #2815, opened 2025-03-05
fix: the repl thread may join itself when the server is stopped

cxljs:
ref: #2806 

Hi @git-hulk , can you review the pr if you have free time?

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -242,20 +242,7 @@ Status Server::Start() {
   return Status::OK();
 }
 
-void Server::Stop() {
-  stop_ = true;
-
-  slaveof_mu_.lock();
-  if (replication_thread_) replication_thread_->Stop();
-  slaveof_mu_.unlock();
-
-  for (const auto &worker : worker_threads_) {
-    worker->Stop(0 /* immediately terminate  */);
-  }
-
-  rocksdb::CancelAllBackgroundWork(storage->GetDB(), true);
-  task_runner_.Cancel();
-}
+void Server::Stop() { stop_ = true; }
 
 void Server::Join() {
   if (auto s = util::ThreadJoin(cron_thread_); !s) {
@@ -853,6 +840,19 @@ void Server::cron() {
     CleanupExitedSlaves();
     recordInstantaneousMetrics();
   }
+
+  assert(stop_);
+
+  slaveof_mu_.lock();
+  if (replication_thread_) replication_thread_->Stop();
+  slaveof_mu_.unlock();
+
+  for (const auto &worker : worker_threads_) {
+    worker->Stop(0 /* immediately terminate  */);
+  }
+
+  rocksdb::CancelAllBackgroundWork(storage->GetDB(), true);
+  task_runner_.Cancel();
 }
 
 Server::InfoEntries Server::GetRocksDBInfo() {


git-hulk:
@fukua95 sure, thank you.

src/server/server.cc
```diff
@@ -853,6 +840,19 @@ void Server::cron() {
     CleanupExitedSlaves();
     recordInstantaneousMetrics();
   }
+
+  assert(stop_);
```
git-hulk, line 844: That's weird to move those codes from `Stop()` to `cron()`. I suspect the issue was caused by the server being stopped before the replication was started.

cxljs: Yes, it's weird. But I don't find a better place to impl the logic, so I read the Redis code and find Redis is also  impl the logic in its `serverCron()`.

cxljs: > I suspect the issue was caused by the server being stopped before the replication was started.

I think before the replication was started, the `ReplicationThread::stop_flag_` is true, so `ReplicationThread::Stop()` will not call the `join()`.

And according to the log:
```
I20250226 11:50:45.351392 140030983013952 main.cc:50] Signal Terminated (15) received, stopping the server
W20250226 11:50:45.351526 140030983013952 replication.cc:371] Replication thread operation failed: thread #140030983013952 cannot be `join`ed: Resource deadlock avoided
I20250226 11:50:45.351537 140030983013952 replication.cc:373] [replication] Stopped
```
`3952` (the repl thread) receive the signal -> call `Server::Stop()` -> join itself.

PragmaTwice (commented):
I'm wondering if it's possible to add a test case for this?

src/server/server.cc
git-hulk: > 3952 (the repl thread) receive the signal -> call Server::Stop() -> join itself.

The repl thread won't receive the terminated signal and stop the server.

cxljs: > The repl thread won't receive the terminated signal and stop the server.

Why do you have this conclusion?

cxljs:
> I'm wondering if it's possible to add a test case for this?

I test it, but I can't reproduce the bug.

In a multi-thread process, every thread have the chance to be interrupted by derive. When the repl thread be interrupted, it will implement the signal handle function, and the handle function will join the repl thread, which cause deadlock.
```
I20250226 11:50:45.351392 140030983013952 main.cc:50] Signal Terminated (15) received, stopping the server
W20250226 11:50:45.351526 140030983013952 replication.cc:371] Replication thread operation failed: thread #140030983013952 cannot be `join`ed: Resource deadlock avoided
I20250226 11:50:45.351537 140030983013952 replication.cc:373] [replication] Stopped
```

From the log, we know the xxx3952 is the repl thread ID, and it received the signal -> call `Server::Stop()` -> wants to join itself (xxx3952).

src/server/server.cc
git-hulk:
```
I20250226 11:50:45.351392 140030983013952 main.cc:50] Signal Terminated (15) received, stopping the server
```

Only the main thread will handle the signal. After receiving the terminated signal, the server->Stop() will be called, and it will then stop the replication thread afterward.

cxljs: https://man7.org/linux/man-pages/man7/signal.7.html#:~:text=A%20process%2Ddirected%20signal%20may%20be,which%20to%20deliver%20the%20signal.

PR #2819, opened 2025-03-07
feat: Add DELPREFIX to delete keys by prefix

### LINKED ISSUE #2628: Allow to delete keys by the prefix
### Search before asking

### Motivation

- Using Namespace in cluster mode to easily flush the keys .

### Solution

- As i knew namespace simply prepend the namespace prefix to the user key and remove it before retrieving. But it couldn't achieve with just simple key prefix.

we use the hash bracket like below, if our namespace is `foo`, and key is 1234, it will be successfully removed all the keys with same namespace.
{foo}:1234

iihimanshuu:
Closes #2628 🚀

Added DELPREFIX command to delete keys by prefix using DeleteRange.

Registered DELPREFIX in commander.cc to ensure Kvrocks recognizes it.

Now, users can run DELPREFIX prefix: to efficiently remove all matching keys.

### DIFF
--- a/src/commands/cmd_key.cc
+++ b/src/commands/cmd_key.cc
@@ -336,6 +336,30 @@ class CommandPersist : public Commander {
   }
 };
 
+class CommandDelPrefix : public Commander {
+ public:
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    if (args_.size() != 2) {
+      return {Status::RedisParseErr, "wrong number of arguments for 'DELPREFIX' command"};
+    }
+
+    std::string prefix = args_[1];
+    auto storage = srv->storage;
+    std::string start_key = prefix;
+    std::string end_key = prefix;
+    end_key.back()++;
+
+    rocksdb::WriteOptions write_options;
+    auto s = storage->GetDB()->DeleteRange(write_options, storage->GetCFHandle("data"), start_key, end_key);
+    if (!s.ok()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    *output = redis::SimpleString("OK");
+    return Status::OK();
+  }
+};
+
 class CommandDel : public Commander {
  public:
   Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
--- a/src/commands/cmd_key.cc
+++ b/src/commands/cmd_key.cc
@@ -31,14 +31,16 @@
 namespace redis {
 
 class CommandType : public Commander {
- public:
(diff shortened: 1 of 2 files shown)


src/commands/cmd_key.cc
```diff
@@ -336,6 +336,30 @@ class CommandPersist : public Commander {
...
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    if (args_.size() != 2) {
+      return {Status::RedisParseErr, "wrong number of arguments for 'DELPREFIX' command"};
+    }
+
+    std::string prefix = args_[1];
+    auto storage = srv->storage;
+    std::string start_key = prefix;
+    std::string end_key = prefix;
+    end_key.back()++;
```
torwig, line 376: Maybe I forgot something about C++, but could you please explain what the goal of this line is?

src/commands/cmd_key.cc
```diff
@@ -336,6 +336,30 @@ class CommandPersist : public Commander {
...
 
+class CommandDelPrefix : public Commander {
+ public:
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    if (args_.size() != 2) {
+      return {Status::RedisParseErr, "wrong number of arguments for 'DELPREFIX' command"};
+    }
+
+    std::string prefix = args_[1];
+    auto storage = srv->storage;
```
torwig, line 373: Why do we introduce a new variable just to use it in a single place?

src/commands/cmd_key.cc
git-hulk: Looks like this line was intended to calculate the boundary of the end key. If so, you could use `StringNext` in the common dir.

src/commands/cmd_key.cc
```diff
@@ -336,6 +336,30 @@ class CommandPersist : public Commander {
...
+    std::string start_key = prefix;
+    std::string end_key = prefix;
+    end_key.back()++;
+
+    rocksdb::WriteOptions write_options;
+    auto s = storage->GetDB()->DeleteRange(write_options, storage->GetCFHandle("data"), start_key, end_key);
```
torwig: Do we have a column family with the name "data"? 
Additionally, I'm a little confused here since I see the function signature:
`rocksdb::ColumnFamilyHandle *GetCFHandle(ColumnFamilyID id);`

PragmaTwice (commented):
Hi, thank you for your contribution!

Could you revert these unrelated code style changes?

iihimanshuu:
Sorry, sir. I'm currently learning Git and open source. I was exploring through the changes I made. Can I please create a new PR and keep it as clean as possible? Apologies for the trouble!

PR #2820, opened 2025-03-07
fix: access nullptr issue in LoadClusterNodes

RiversJin:
The sequence `LoadClusterNodes` -> `SetClusterNodes` -> `SetMasterSlaveRepl` -> `srv_->slot_migrator->SetStopMigrationFlag` causes a null pointer access during initialization. The fix involves moving the initialization of slot_migrator earlier to resolve this issue.

PS: I'sorry that the logic where `SetClusterNodes` accesses slot_migrator was my modification, meaning I introduced this bug. :P

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -176,14 +176,16 @@ Status Server::Start() {
   }
 
   if (config_->cluster_enabled) {
+    // Create objects used for slot migration
+    slot_migrator = std::make_unique<SlotMigrator>(this);
+
     if (config_->persist_cluster_nodes_enabled) {
       auto s = cluster->LoadClusterNodes(config_->NodesFilePath());
       if (!s.IsOK()) {
         return s.Prefixed("failed to load cluster nodes info");
       }
     }
-    // Create objects used for slot migration
-    slot_migrator = std::make_unique<SlotMigrator>(this);
+
     auto s = slot_migrator->CreateMigrationThread();
     if (!s.IsOK()) {
       return s.Prefixed("failed to create migration thread");


torwig (approved):

git-hulk:
@RiversJin, do you mean the null pointer may happen on this line? https://github.com/apache/kvrocks/blob/231b093c979ab80c11ac5433e52d43213dc94840/src/cluster/cluster.cc#L214C24-L214C37

mapleFU:
By the way, anyway we should return error/emit log rather than coredump on access nullptr? Can we add some guard in visiting the pointer?

RiversJin:
> `LoadClusterNodes` -> `SetClusterNodes` -> `SetMasterSlaveRepl` -> `srv_->slot_migrator->SetStopMigrationFlag`
https://github.com/apache/kvrocks/blob/231b093c979ab80c11ac5433e52d43213dc94840/src/cluster/cluster.cc#L252
@git-hulk  here

PragmaTwice (approved):

git-hulk:
> > `LoadClusterNodes` -> `SetClusterNodes` -> `SetMasterSlaveRepl` -> `srv_->slot_migrator->SetStopMigrationFlag`
> > https://github.com/apache/kvrocks/blob/231b093c979ab80c11ac5433e52d43213dc94840/src/cluster/cluster.cc#L252
> > 
> > @git-hulk  here

@RiversJin Seems it has been checked if the `migrator` is null before accessing? see: https://github.com/apache/kvrocks/blob/231b093c979ab80c11ac5433e52d43213dc94840/src/cluster/cluster.cc#L250C15-L250C28

RiversJin:
> > > `LoadClusterNodes` -> `SetClusterNodes` -> `SetMasterSlaveRepl` -> `srv_->slot_migrator->SetStopMigrationFlag`
> > > https://github.com/apache/kvrocks/blob/231b093c979ab80c11ac5433e52d43213dc94840/src/cluster/cluster.cc#L252
> > > 
> > > @git-hulk  here
> 
> Seems it has been checked if the `migrator` is null before accessing? see: https://github.com/apache/kvrocks/blob/231b093c979ab80c11ac5433e52d43213dc94840/src/cluster/cluster.cc#L250C15-L250C28

Oops, it seems I've mixed up the differences between the private code of my current company and the community code ..
However,according to the information above, there is another place that also relies on slot_migrator, so I'll first test locally to see if there is an issue with the current branch ...

git-hulk:
@RiversJin Yes, but I think this PR could also be merged since it would be better to move it before loading the cluster node.

git-hulk (approved):

PR #2822, opened 2025-03-07
feat(cmd): add iso8601 time format parameter to LASTSAVE command

### LINKED ISSUE #2817: LASTSAVE Command to Return Human-Readable Timestamp.
### Search before asking

### Motivation

While i was working with creating backups in the kvrocks i wanted to see the time when the last save was done, i ran the ``LASTSAVE`` command and it returned the time in the unix epoch timestamp.

### Solution

I’d like to propose an improvement to the ``LASTSAVE`` command by using the ``<ctime>`` library to convert the Unix timestamp into a human-readable local time format.

Currently, ``LASTSAVE`` returns the time in Unix epoch format, which requires users to manually convert it to their local time. This can be a bit time-consuming.

I have two solutions for this.
1.) Introduce a new command ``LASTSAVEH``
This command would return the last save time in a human-readable format (e.g., ``2025-03-06 21:54:51``). The existing ``LASTSAVE`` command would remain unchanged, continuing to return the Unix timestamp. 
2.) Modify ``LASTSAVE`` to return both formats
Instead of introducing a new command, we could modify ``LASTSAVE`` to return two values:
- The Unix timestamp (for backward compatibility).
- The formatted timestamp (for better readability).

I am happy to work on this enhancement, if you find my idea relevant kindly assign this to me. 😄 

SharonIV0x86:
This closes #2817.
 
This PR introduces a new parameter to the ``LASTSAVE`` command to return the time of last save to disk in a human readable format.

Usage
``LASTSAVE ISO8601``
``LASTSAVE iso8601``

``LASTSAVE`` also works for backward compatibility.

```bash
127.0.0.1:6666> LASTSAVE
(integer) 1741363334
127.0.0.1:6666> LASTSAVE iso8601
"2025-03-07T21:32:14+0530"
127.0.0.1:6666> LASTSAVE ISO8601
"2025-03-07T21:32:14+0530"
127.0.0.1:6666> 
```


src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
 
 class CommandLastSave : public Commander {
  public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() > 2) {
+      return {Status::RedisParseErr, "unknown extra subcommand"};
+    }
+    if (args.size() == 2) {
+      std::string fmt_arg = args[1];
```
torwig: You can try to use `EqualICase` (common->string_util.h).

SharonIV0x86: > You can try to use `EqualICase` (common->string_util.h).

Initially i came across ``parser.EatEqICase()`` but then i dropped the idea as i didnt know its correct usage so i though i'll take some feedback on its usage. 

Thanks for the review i'll use ``EqualICase`` as it seems a better fit in this case.

PragmaTwice (commented):
It doesn't seem to be a Redis feature. So I'm wondering if it will be useful in production envs?

SharonIV0x86:
> It doesn't seem to be a Redis feature. So I'm wondering if it will be useful in production envs?

You're right this isn't a redis feature. 

But instead of converting the UNIX timestamp manually, users can directly retrieve an easily interpretable format.

PragmaTwice:
> You're right this isn't a redis feature.
> 
> But instead of converting the UNIX timestamp manually, users can directly retrieve an easily interpretable format.

I've checked the issue you opened. I'm fine with this change.

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
 
 class CommandLastSave : public Commander {
  public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() > 2) {
+      return {Status::RedisParseErr, "unknown extra subcommand"};
```
```suggestion
      return {Status::RedisParseErr, "unknown extra arguments"};
```

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
...
+    }
+    if (args.size() == 2) {
+      std::string fmt_arg = args[1];
+      std::transform(fmt_arg.begin(), fmt_arg.end(), fmt_arg.begin(), ::tolower);
+      if (fmt_arg == "iso8601") {
+        format_spec_ = true;
+      } else {
+        return {Status::RedisParseErr, "unknown subcommand"};
```
```suggestion
        return {Status::RedisParseErr, "unknown arguments"};
```

src/commands/cmd_server.cc
SharonIV0x86: Thanks for review, i was just working on the changes requested by torwig right now, ill make this change as well.

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
...
 class CommandLastSave : public Commander {
  public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() > 2) {
+      return {Status::RedisParseErr, "unknown extra subcommand"};
+    }
+    if (args.size() == 2) {
+      std::string fmt_arg = args[1];
+      std::transform(fmt_arg.begin(), fmt_arg.end(), fmt_arg.begin(), ::tolower);
+      if (fmt_arg == "iso8601") {
```
```suggestion
      if (util::EqualICase(args[1], "iso8601")) {
```

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
...
+    }
+    return Status::OK();
+  }
   Status Execute([[maybe_unused]] engine::Context &ctx, Server *srv, [[maybe_unused]] Connection *conn,
                  std::string *output) override {
     int64_t unix_sec = srv->GetLastBgsaveTime();
-    *output = redis::Integer(unix_sec);
+    if (format_spec_) {
+      time_t raw_time = static_cast<time_t>(unix_sec);
+      struct tm *local_time = localtime(&raw_time);
```
PragmaTwice: We can use the thread-safe version `localtime_r`.

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
...
+      if (local_time) {
+        char buf[100];
+        strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%S%z", local_time);
+        *output = redis::BulkString(buf);
+      } else {
+        *output = redis::BulkString("Error converting time");
```
```suggestion
        *output = redis::BulkString("Failed to convert timestamp to local time");
```

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
...
-    *output = redis::Integer(unix_sec);
+    if (format_spec_) {
+      time_t raw_time = static_cast<time_t>(unix_sec);
+      struct tm local_time;
+      if (localtime_r(&raw_time, &local_time)) {
+        char buf[100];
+        strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%S%z", &local_time);
+        *output = redis::BulkString(buf);
+      } else {
+        *output = redis::BulkString("Failed to convert timestamp to local time");
```
```suggestion
        *output = redis::Error({Status::NotOK, "unable to convert timestamp to local time"});
```

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
...
-    *output = redis::Integer(unix_sec);
+    if (format_spec_) {
+      time_t raw_time = static_cast<time_t>(unix_sec);
+      struct tm local_time;
+      if (localtime_r(&raw_time, &local_time)) {
+        char buf[100];
+        strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%S%z", &local_time);
+        *output = redis::BulkString(buf);
+      } else {
+        *output = redis::Error({Status::NotOK, "unable to convert timestamp to local time"});
```
```suggestion
        return {Status::NotOK, "unable to convert timestamp to local time"};
```

PragmaTwice (commented):
Looks good. Could you add a go test case for your changes?

SharonIV0x86:
> Looks good. Could you add a go test case for your changes?

I am a bit confused about where exactly to add the test for ``LASTSAVE`` command, i guess it will make sense to add it in the ``gocase/unit/server/poll_updates_test.go`` as the ``LASTSAVE`` command is listed under the server section in the docs?

PragmaTwice:
You can add a new go source file under `tests/gocase/unit/server/`.

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
...
+    }
+
+    return Status::OK();
+  }
   Status Execute([[maybe_unused]] engine::Context &ctx, Server *srv, [[maybe_unused]] Connection *conn,
                  std::string *output) override {
     int64_t unix_sec = srv->GetLastBgsaveTime();
-    *output = redis::Integer(unix_sec);
+    if (format_spec_) {
+      time_t raw_time = static_cast<time_t>(unix_sec);
```
```suggestion
      auto raw_time = static_cast<time_t>(unix_sec);
```

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
...
+
+    return Status::OK();
+  }
   Status Execute([[maybe_unused]] engine::Context &ctx, Server *srv, [[maybe_unused]] Connection *conn,
                  std::string *output) override {
     int64_t unix_sec = srv->GetLastBgsaveTime();
-    *output = redis::Integer(unix_sec);
+    if (format_spec_) {
+      auto raw_time = static_cast<time_t>(unix_sec);
+      struct tm local_time;
```
```suggestion
      tm local_time{};
```

src/commands/cmd_server.cc
```diff
@@ -1006,12 +1008,42 @@ static uint64_t GenerateConfigFlag(uint64_t flags, const std::vector<std::string
...
+      if (localtime_r(&raw_time, &local_time)) {
+        char buf[100];
```
```suggestion
        char buf[64];
```

git-hulk:
Looks good. @SharonIV0x86 You could use the gofmt to format Go files to pass the CI.

git-hulk (approved):

PragmaTwice (approved):

PR #2823, opened 2025-03-08
feat: Enable to save slowlog in logfile

greatsharp:


aleksraiden:
It's a good idea, thanks!

src/stats/log_collector.cc
```diff
@@ -89,6 +101,37 @@ void LogCollector<T>::PushEntry(std::unique_ptr<T> &&entry) {
...
+    if (entry->args.size() > 0) {
+      for (const auto &arg: entry->args) {
+        cmd.append(arg).append(" ");
+      }
+      cmd.pop_back();
+    }
+    // json format
+    auto log = fmt::format("[slowlog] {{\"id\":{}, \"timestamp\":{}, \"duration\":{}, \"cmd\":\"{}\", "
+        "\"ip\":\"{}\", \"port\":{}, \"client_name\":\"{}\"}}",
+        entry->id, entry->time, entry->duration, cmd, entry->ip, entry->port, entry->client_name);
```
PragmaTwice: Hmm in other places we don't use JSON format in logs, so maybe we should consider if we want to use JSON format in log here.

greatsharp: 1. just want to show slowlog integrally in log file, distinguish with other runtime logs.
Snipaste_2025-03-09_10-13-27

2. if third-party software wants to gather slowlog from log file, json format maybe more friendlly.

torwig: @greatsharp I think it's already distinguished since the lines have the `[slowlog]` prefix. What @PragmaTwice might suggested was to introduce one more option to specify the slow log format (for example, "text" (default) or "json"). WDYT, @PragmaTwice ?

greatsharp: @aleksraiden @torwig @PragmaTwice @git-hulk .
Excuse me, should I change the json format to normal?

torwig: In my opinion there are two options:
1. Make a slow log in the same format as other logs.
2. Introduce a setting to choose the JSON format for the slow log records.

I would like to hear opinions on this.

git-hulk: Agree with @torwig, both of them are good to me.

torwig: @greatsharp, as far as I can see, you didn't change one of the options I mentioned. We can start with option 1; later, if there is demand, we can easily transform it into option 2 by adding one more setting.

greatsharp: @torwig
I agree with you, make it simple at first. 
I tried to submit code yesterday, but continuesly reject by github of code 403, let me try today.

greatsharp: @git-hulk @torwig check again.

torwig (approved):

Issue #2794, opened 2025-02-15
TDigest: Implement QUANTILE command for TDigest Algorithm


LindaSummer:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Refer: #2558
We should implement `TDIGEST.QUANTILE` command.

Redis command reference:

https://redis.io/docs/latest/commands/tdigest.quantile/

### Solution

Depends on #2791 #2792 .

Follow the implementation of #2791 for this command.

Add go integration tests for it.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

AntiTopQuark:
Hi, I would like to work on this command . Could you please assign the issue to me?

LindaSummer:
> Hi, I would like to work on this command . Could you please assign the issue to me?

Hi @AntiTopQuark ,

Thanks very much for your interests on this topic! 😁

This command depends on #2792 for integration tests.

Maybe we need to wait for this command to be finished.

Best Regards,
Edward

AntiTopQuark:
OK, I will wait unitl #2792  is merged, and then I will take it on.
Could you assign me the issue beforehand?

LindaSummer:
> OK, I will wait unitl [#2792](https://github.com/apache/kvrocks/issues/2792) is merged, and then I will take it on. Could you assign me the issue beforehand?

Hi @AntiTopQuark ,

Thanks for your interest on this topic! You can leave an assignment comment here. 😊

You may need take care on the lock of quantile function.

We need just lock the merge part and keep read part read-only.

We should confirm that we create a correct lock with other write actions.

You can refer the comment in #2799 PR's description. 

Best Regards,
Edward

SharonIV0x86:
@LindaSummer I was trying to work on this, but it seems like the code just freezes on this particular line when trying to lock the key before merging the unmerged nodes.
https://github.com/apache/kvrocks/blob/3e1b21b71dcb5d424c89204d05835a99e5d0fca3/src/types/redis_tdigest.cc#L193

Any guidance on this is appreciated.

LindaSummer:
> [@LindaSummer](https://github.com/LindaSummer?rgh-link-date=2025-03-19T09%3A58%3A20.000Z) I was trying to work on this, but it seems like the code just freezes on this particular line when trying to lock the key before merging the unmerged nodes.
> 
[...]

PR #2826, opened 2025-03-13
feat(tdigest): add support of TDIGEST.RESET command

### LINKED ISSUE #2804: TDigest: Implement `TDIGEST.RESET` command Algorithm
### Search before asking

### Motivation

Refer: #2558
We should implement `TDIGEST.RESET` command.

Redis command reference:
https://redis.io/docs/latest/commands/tdigest.reset

### Solution

Depends on #2791 #2792 .

Follow the implementation of https://github.com/apache/kvrocks/issues/2791 for this command.

Add go integration tests for it.

SharonIV0x86:
This closes #2804 

This PR ~is a Work In Progress and~ adds the support for the ``TDIGEST.RESET`` command https://redis.io/docs/latest/commands/tdigest.reset/.

I’m currently working on implementing the ``TDIGEST.RESET`` command. However, I noticed that there doesn’t seem to be a built-in way to delete centroids from a ``TDigest``. And in order to reset a digest we need to delete the centroids that its containing, correct me if im wrong. 
Also, to do this, one way i see is that we can manually delete them by iterating over the stored centroids and removing them one by one. 

Additionally there seems to be already a ``Reset()`` function defined.
https://github.com/apache/kvrocks/blob/b8db0e75553ccf6e02cde66c382d86a7d7ec5570/src/types/tdigest.cc#L117
And i am really confused about the two implementation of the ``TDigest`` in kvrocks, one in ``types/redis_tdigest.cc`` and one in ``type/tdigest.cc``. 
The existing ``TDigest`` commands uses the ``TDigest`` implementation from ``types/redis_tdigest.cc`` and it currently has no ``Reset()`` function defined. 

So any assistance or reviews on this PR are highly appreciated as i am new to the codebase.

To the best of my knowledge ``DummyCentroids`` class is being used as a class to iterate over centroids in ``Quantile`` function, maybe if its possible we can add some functionality in this class to delete a centroid.

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -377,7 +377,13 @@ Status Cluster::ImportSlotRange(redis::Connection *conn, const SlotRange &slot_r
           LOG(ERROR) << fmt::format("[import] Failed to stop importing slot(s) {}: {}", slot_range.String(), s.Msg());
         }
       };  // Stop forbidding writing slot to accept write commands
-      if (slot_range == srv_->slot_migrator->GetForbiddenSlotRange()) srv_->slot_migrator->ReleaseForbiddenSlotRange();
+      if (slot_range.HasOverlap(srv_->slot_migrator->GetForbiddenSlotRange())) {
+        // This approach assumes a shard only handles one migration task at a time.
+        // When executing the import logic, the absence of other outgoing migrations on this shard justifies safely
+        // removing the forbidden slot. A more robust solution would be required if concurrent slot migrations are
+        // supported in the future.
+        srv_->slot_migrator->ReleaseForbiddenSlotRange();
+      }
       LOG(INFO) << fmt::format("[import] Start importing slot(s) {}", slot_range.String());
       break;
     case kImportSuccess:
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -38,7 +38,6 @@ constexpr auto kInfoUnmergedWeight = "Unmerged weight";
 constexpr auto kInfoObservations = "Observations";
 constexpr auto kInfoTotalCompressions = "Total compressions";
 }  // namespace
-// chamba, sultanpur college.
 class CommandTDigestCreate : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -202,7 +202,22 @@ class CommandTDigestMinMax : public Commander {
   std::string key_name_;
   bool is_min_;
 };
+class CommandTDigestReset : public Commander {
+  Status Parse(const std::vector<std::string> &args) override {
+    key_name_ = args[1];
+    return Status::OK();
+  }
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    TDigest tdigest(srv->storage, conn->GetNamespace);
+    TDigestMetadata metadata;
+    auto s = tdigest.GetMetaData(ctx, key_name_, &metadata);
+    CentroidsWithDelta cd = tdigest.DumpCentroids();
+    tdigest.Reset(cd);
+  }
 
+ private:
+  srd::srring key_name_;
+}
 // Then replace the existing template implementation and type aliases with:
 class CommandTDigestMin : public CommandTDigestMinMax {
  public:
@@ -218,5 +233,6 @@ REDIS_REGISTER_COMMANDS(TDigest, MakeCmdAttr<CommandTDigestCreate>("tdigest.crea
                         MakeCmdAttr<CommandTDigestInfo>("tdigest.info", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandTDigestAdd>("tdigest.add", -3, "write", 1, 1, 1),
                         MakeCmdAttr<CommandTDigestMax>("tdigest.max", 2, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandTDigestMin>("tdigest.min", 2, "read-only", 1, 1, 1));
+                        MakeCmdAttr<CommandTDigestMin>("tdigest.min", 2, "read-only", 1, 1, 1)),
+                        MakeCmdAttr<CommandTDigestReset>("tdigest.reset", 2, "write", 1, 1, 1);
 }  // namespace redis
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -38,6 +38,7 @@ constexpr auto kInfoUnmergedWeight = "Unmerged weight";
 constexpr auto kInfoObservations = "Observations";
 constexpr auto kInfoTotalCompressions = "Total compressions";
 }  // namespace
+
 class CommandTDigestCreate : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -38,7 +38,7 @@ constexpr auto kInfoUnmergedWeight = "Unmerged weight";
 constexpr auto kInfoObservations = "Observations";
 constexpr auto kInfoTotalCompressions = "Total compressions";
 }  // namespace
-
+// chamba, sultanpur college.
 class CommandTDigestCreate : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
@@ -208,16 +208,29 @@ class CommandTDigestReset : public Commander {
     return Status::OK();
   }
   Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
-    TDigest tdigest(srv->storage, conn->GetNamespace);
+    TDigest tdigest(srv->storage, conn->GetNamespace());
     TDigestMetadata metadata;
     auto s = tdigest.GetMetaData(ctx, key_name_, &metadata);
-    CentroidsWithDelta cd = tdigest.DumpCentroids();
-    tdigest.Reset(cd);
+    if (!s.ok()) {
+      if (s.IsNotFound()) {
+        return {Status::RedisExecErr, errKeyNotFound};
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
+    if (metadata.total_observations == 0) {
+      *output = redis::BulkString("nan");
+      return Status::OK();
+    }
+    s = tdigest.Reset(ctx, key_name_);
+    if (s.ok()) {
+      return Status::OK();
+    }
+    return {Status::RedisExecErr, s.ToString()};
   }
 
  private:
-  srd::srring key_name_;
-}
+  std::string key_name_;
+};
 // Then replace the existing template implementation and type aliases with:
 class CommandTDigestMin : public CommandTDigestMinMax {
  public:
@@ -233,6 +246,6 @@ REDIS_REGISTER_COMMANDS(TDigest, MakeCmdAttr<CommandTDigestCreate>("tdigest.crea
                         MakeCmdAttr<CommandTDigestInfo>("tdigest.info", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandTDigestAdd>("tdigest.add", -3, "write", 1, 1, 1),
                         MakeCmdAttr<CommandTDigestMax>("tdigest.max", 2, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandTDigestMin>("tdigest.min", 2, "read-only", 1, 1, 1)),
-                        MakeCmdAttr<CommandTDigestReset>("tdigest.reset", 2, "write", 1, 1, 1);
+                        MakeCmdAttr<CommandTDigestMin>("tdigest.min", 2, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandTDigestReset>("tdigest.reset", 2, "write", 1, 1, 1));
 }  // namespace redis
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -38,6 +38,7 @@ constexpr auto kInfoUnmergedWeight = "Unmerged weight";
 constexpr auto kInfoObservations = "Observations";
 constexpr auto kInfoTotalCompressions = "Total compressions";
 }  // namespace
+// chamba, sultanpur college.
 class CommandTDigestCreate : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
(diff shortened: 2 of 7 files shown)


SharonIV0x86:
@PragmaTwice I updated the Reset function so that it resets the metadata fields and writes the new metadata back.

From what I understand, this should only clear the high-level info, removing references to the centroids without actually deleting them. The centroids should still exist until compaction eventually cleans them up.

But when I run ``TDIGEST.RESET``, it gets stuck in an infinite loop. I’m not sure why any ideas on what might be causing this?

PragmaTwice:
@LindaSummer Hi, could you help to review this PR when you have time?

LindaSummer:
> @LindaSummer Hi, could you help to review this PR when you have time?

Hi @PragmaTwice ,

Of course!
I will take a review today.😊

Best Regards,
Edward

LindaSummer:
> @PragmaTwice I updated the Reset function so that it resets the metadata fields and writes the new metadata back.
> 
> 
> 
> From what I understand, this should only clear the high-level info, removing references to the centroids without actually deleting them. The centroids should still exist until compaction eventually cleans them up.
> 
> 
> 
> But when I run ``TDIGEST.RESET``, it gets stuck in an infinite loop. I’m not sure why any ideas on what might be causing this?

Hi @SharonIV0x86 ,

Thanks very much for your effort!😊

If I understand correctly, the compaction is handled by rocksdb and we should delete related keys to free the reference of related keys.

In my subkey design, I use a flag prefix for centroid keys. It could be deleted by some method like deletebyprefix.

I'm now without a computer by my side, I would provide the function link of prefix deletion today after back home.😊

For the infinite loop issue, I will try to find the cause if you don't mind.😊

Best Regards,
Edward

SharonIV0x86:
@LindaSummer The looping issue is fixed and it works as expected.

SharonIV0x86:
@LindaSummer If possible also provide your implementation too, as you mentioned it can be used by some method like ``deletebyprefix`` but ig its not added to kvrocks yet. So ill compare yours and mine and whichever is better we'll go with it.

src/types/redis_tdigest.cc
```diff
@@ -238,7 +238,46 @@ rocksdb::Status TDigest::Quantile(engine::Context& ctx, const Slice& digest_name
...
+  metadata.merge_times = 0;
+
+  std::string metadata_bytes;
+  metadata.Encode(&metadata_bytes);
+
+  status = batch->Put(metadata_cf_handle_, ns_key, metadata_bytes);
+  if (!status.ok()) {
+    return status;
+  }
+
```
LindaSummer, line 270:
```suggestion

  auto start_key = internalSegmentGuardPrefixKey(metadata, ns_key, SegmentType::kBuffer);
  auto guard_key = internalSegmentGuardPrefixKey(metadata, ns_key, SegmentType::kGuardFlag);
  status = batch->DeleteRange(cf_handle_, start_key, guard_key);
  if (!status.ok()) {
    return status;
 }
```

Hi @SharonIV0x86 ,

It will be better to delete buffer and centroid keys. 😊
The [`DeleteRange`](https://github.com/facebook/rocksdb/wiki/DeleteRange) action is the one mentioned like `deletebyprefix`. It's a lightweight action and just add a range tombstone for keys. This will make rocksdb cleanup current data in future compaction.

Best Regards,
Edward

src/types/redis_tdigest.cc
```diff
@@ -238,7 +238,46 @@ rocksdb::Status TDigest::Quantile(engine::Context& ctx, const Slice& digest_name
...
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisTDigest);
+  status = batch->PutLogData(log_data.Encode());
+  if (!status.ok()) {
+    return status;
+  }
 
+  // metadata.compression = 0;
```
LindaSummer: Hi @SharonIV0x86 ,

It will be better to remove unused code comments. 😊

Best Regards,
Edward

LindaSummer (commented):
Hi @SharonIV0x86 ,

Left with some comments.

By the way, could you add some go integration test for this command to cover the new code in this PR if possible?

Best Regards,
Edward

src/types/redis_tdigest.cc
SharonIV0x86: Right! I wanted to do something exactly like that, but i was just no getting the idea how to do so. Thanks so much for this review, ill add this.

src/types/redis_tdigest.cc
SharonIV0x86: Will do it, i was just waiting for the review, also earlier i was also modifying the **compression** rate and **capacity**, but i realized that they should not be changed in a reset as they are being set by user when creating a tdigest.

https://github.com/apache/kvrocks/blob/b8db0e75553ccf6e02cde66c382d86a7d7ec5570/src/types/redis_tdigest.cc#L121

src/types/redis_tdigest.cc
```diff
@@ -273,6 +271,11 @@ rocksdb::Status TDigest::Reset(engine::Context& ctx, const Slice& digest_name) {
     return status;
   }
 
+  if (!status.ok()) return status;
+  auto start_key = internalSegmentGuardPrefixKey(metadata, ns_key, SegmentType::kBuffer);
+  auto guard_key = internalSegmentGuardPrefixKey(metadata, ns_key, SegmentType::kGuardFlag);
+  
+  status = batch->DeleteRange(cf_handle_, start_key, guard_key);
```
LindaSummer:
```suggestion
  status = batch->DeleteRange(cf_handle_, start_key, guard_key);
  if (!status.ok()) {
    return status;
  }
 
 status = storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
```

src/commands/cmd_tdigest.cc
```diff
@@ -202,7 +201,36 @@ class CommandTDigestMinMax : public Commander {
...
+    TDigest tdigest(srv->storage, conn->GetNamespace());
+    TDigestMetadata metadata;
+    auto s = tdigest.GetMetaData(ctx, key_name_, &metadata);
+    if (!s.ok()) {
+      if (s.IsNotFound()) {
+        return {Status::RedisExecErr, errKeyNotFound};
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
+    if (metadata.total_observations == 0) {
```
LindaSummer, line 220: Hi @SharonIV0x86 ,

I have tested this case on redis-stack docker.
It seems to respond with `OK`.
```
127.0.0.1:6389> tdigest.create hahah compression 1000
OK
127.0.0.1:6389> tdigest.reset hahah
OK
```

Could you help us double-check this behavior for this `nan` response on redis-stack and add a go test case?


Best Regards,
Edward

LindaSummer (commented):
Hi @SharonIV0x86 ,

Reviewed with some comments.  😊

Best Regards,
Edward

src/types/redis_tdigest.cc
SharonIV0x86: I am a bit confused with this, am i not supposed to write to the storage?
Should i remove this line?
```cpp
status = storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
```

src/commands/cmd_tdigest.cc
SharonIV0x86: Ill check with redis-stack once i set it up. 

Assuming the output you attached is from redis-stack. 
In the current implementation, when performing a reset on a empty tdigest "nan" is returned, i guess we can replace this with a OK response just like redis-stack.

src/types/redis_tdigest.cc
LindaSummer: Sorry for my mistake. This line should be kept.
We should add the status check before writing to storage.

SharonIV0x86:
@LindaSummer I cross-checked with the ``redis-stack`` and it returns ``OK`` for both the cases.
1. If digest is empty and you execute a reset: returns OK.
2. If digest is non-empty and you execute a reset: return OK.

Now this aligns with the functionality of ``redis-stack``. Also i added a go test case for this. Please review when you have time 😄

src/types/redis_tdigest.cc
```diff
@@ -238,7 +238,50 @@ rocksdb::Status TDigest::Quantile(engine::Context& ctx, const Slice& digest_name
...
+
+  std::string metadata_bytes;
+  metadata.Encode(&metadata_bytes);
+
+  status = batch->Put(metadata_cf_handle_, ns_key, metadata_bytes);
+  if (!status.ok()) {
+    return status;
+  }
+
+  if (!status.ok()) return status;
```
LindaSummer: it'll be better to keep same code style in same commit.😊

src/types/redis_tdigest.cc
```diff
@@ -238,7 +238,50 @@ rocksdb::Status TDigest::Quantile(engine::Context& ctx, const Slice& digest_name
...
+  auto start_key = internalSegmentGuardPrefixKey(metadata, ns_key, SegmentType::kBuffer);
+  auto guard_key = internalSegmentGuardPrefixKey(metadata, ns_key, SegmentType::kGuardFlag);
+  
+  status = batch->DeleteRange(cf_handle_, start_key, guard_key);
+  if (!status.ok()) return status;
```
LindaSummer: same as above.

src/types/redis_tdigest.cc
```diff
@@ -238,7 +238,50 @@ rocksdb::Status TDigest::Quantile(engine::Context& ctx, const Slice& digest_name
...
+  status = storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+  if (!status.ok()) return status;
```
LindaSummer: same as above.

tests/gocase/unit/type/tdigest/tdigest_test.go
```diff
@@ -257,4 +257,41 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
...
+		key := keyPrefix + "mydigest"
+		//adding some data to digest
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.ADD", key, "-84.3", "199.3", "343.34", "12.34").Err())
+		rsp := rdb.Do(ctx, "TDIGEST.MIN", key)
+		require.NoError(t, rsp.Err())
+		require.EqualValues(t, rsp.Val(), "-84.3")
+
+		//reset on a non existent key.
+		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.RESET", keyPrefix+"notexist").Err(), errMsgKeyNotExist)
+		{
```
LindaSummer: maybe we could add braces for all cases or none to keep same style.😊

tests/gocase/unit/type/tdigest/tdigest_test.go
```diff
@@ -257,4 +257,41 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
...
+			// reset on non-empty digest.
+			require.NoError(t, rdb.Do(ctx, "TDIGEST.RESET", key).Err())
+			// getting TDIGEST.INFO after an reset for testing metadata.
+			rsp = rdb.Do(ctx, "TDIGEST.INFO", key)
+			require.NoError(t, rsp.Err())
+			info := toTdigestInfo(t, rsp.Val())
+			require.GreaterOrEqual(t, int64(1024), info.Capacity)
```
LindaSummer: maybe we could call tdigest.info to get the capacity before reset and assert it keeps the original value.😊

LindaSummer (commented):
Hi @SharonIV0x86 ,

Thanks for your effort.😊
Generally LGTM.

Leave with some nitpick comments.

Best Regards,
Edward

SharonIV0x86:
@LindaSummer Made the changes, pls check now and lmk so i can proceed with opening a PR for modifying docs.

src/types/redis_tdigest.cc
```diff
@@ -266,20 +264,19 @@ rocksdb::Status TDigest::Reset(engine::Context& ctx, const Slice& digest_name) {
...
-  
-  status = batch->DeleteRange(cf_handle_, start_key, guard_key);
-  if (!status.ok()) return status;
-  status = storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
-  if (!status.ok()) return status;
 
+  if (auto status = batch->DeleteRange(cf_handle_, start_key, guard_key); !status.ok()) {
+    return status;
+  }
+  if (auto status = storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch()); !status.ok()) {
```
LindaSummer: We can directly return this status.

LindaSummer (commented):
Hi @SharonIV0x86 ,

Thanks for your contribution!

Left one comment.

Other parts are LGTM.

Best Regards,
Edward

src/cluster/cluster.cc
```diff
@@ -377,7 +377,13 @@ Status Cluster::ImportSlotRange(redis::Connection *conn, const SlotRange &slot_r
           LOG(ERROR) << fmt::format("[import] Failed to stop importing slot(s) {}: {}", slot_range.String(), s.Msg());
         }
       };  // Stop forbidding writing slot to accept write commands
-      if (slot_range == srv_->slot_migrator->GetForbiddenSlotRange()) srv_->slot_migrator->ReleaseForbiddenSlotRange();
+      if (slot_range.HasOverlap(srv_->slot_migrator->GetForbiddenSlotRange())) {
```
LindaSummer: Hi @SharonIV0x86 ,

It seems that this file change should not be shown in this PR.

Could you take a review on your commit?

Best Regards,
Edward

SharonIV0x86: This seems like a mistake, maybe merging the base branch with this can fix it?

src/commands/cmd_tdigest.cc
```diff
@@ -38,7 +38,6 @@ constexpr auto kInfoUnmergedWeight = "Unmerged weight";
 constexpr auto kInfoObservations = "Observations";
 constexpr auto kInfoTotalCompressions = "Total compressions";
 }  // namespace
-
```
LindaSummer: Hi @SharonIV0x86 ,

Could you help keep this blank line for namespace section if you don't mind? 😊

Best Regards,
Edward

LindaSummer:
Hi @SharonIV0x86 

LGTM

Thanks very much for your contribution!😊

Best Regards,
Edward

SharonIV0x86:
> Hi @SharonIV0x86
> 
> LGTM
> 
> Thanks very much for your contribution!😊
> 
> Best Regards, Edward

Done. 
Also kindly take a look at this, as i want to work on [this](https://github.com/apache/kvrocks/issues/2794#issuecomment-2735996546) issue next.

PragmaTwice (approved):

PR #2827, opened 2025-03-13
feat(cmd): add new command KMETADATA

ltagliamonte-dd:
Hello Folks,
To enable SSTable sideloading for Redis types that use versioning, retrieving the metadata.version is necessary when updating an existing key.

To support this, we've introduced a new command: KMETADATA, which returns the metadata for a given key.

Looking forward to your feedback!


git-hulk:
@apache/kvrocks-committers What do you feel about this command proposal?

torwig:
@ltagliamonte-dd Could you please explain how the data returned by this command will be used for "SSTable sideloading"? Just to understand the flow.

ltagliamonte-dd:
> @ltagliamonte-dd Could you please explain how the data returned by this command will be used for "SSTable sideloading"? Just to understand the flow.

Sure thing @torwig, here more details:
For my understanding data types that use versioning (like sets,zset,hash) share the key version with their sub keys.

For example for hash data types (HSET Key subKey1 subKey1val1) there is a rocksdb row for the `Key` and an row for `subKey1`.
The `Key` and the `subKey1` share the `metadata.version`.

When writing to kvrocks there are 2 use cases:
- add a new key
 a new version is created for the `Key` and that version is inherited by `subKey1`
- update an existing key
the version needs to be retrieved from the `Key` and shared with the `subKey1`

This new CMD covers the update of an existing key.

I'm still quite new to the decode/encode part of the project please let me know if my understanding is not correct.

torwig:
@git-hulk Could you please provide a link to the encoding schema? I can't find it :)
@ltagliamonte-dd Actually, for complex data types there is just one, let's say Final-Key that consists of Key (type + version + slot ID, etc.) followed by Subkey. For example, all the elements of HSET `myset` have the common prefix (the prefix has version), subkeys don't.

ltagliamonte-dd:
@torwig thanks, since my use case is just HSET i've been focusing on this function https://github.com/apache/kvrocks/blob/unstable/src/types/redis_hash.cc#L241-L296

going over the code it looks like that when the key exists, metadata.version contains the current version of the Key hash.
This version is used to construct the subkeys for the field-value pairs being set or updated.
Am i interpreting this wrong?

PragmaTwice (commented):
Maybe we can have a better name, `GETXXX` seems not so "Redis".

To avoid collision with future Redis commands, we can prefix with something like `K` or `KVR`, e.g. `KMETADATA`.

src/commands/cmd_key.cc
```diff
@@ -553,6 +553,68 @@ class CommandSort : public Commander {
...
+        break;
+      case kRedisList:
+        metadata = std::make_unique<ListMetadata>();
+        break;
+      default:
+        return {Status::RedisExecErr, "Unimplemented Redis type"};
+    }
+
+    // Get metadata
+    s = redis.GetMetadata(ctx, {type}, nskey, metadata.get());
```
PragmaTwice: Since you don't need other metadata fields, you can just use `GetMetadata(kRedisNone)` and remove the `redis.Type` call & the switch statement.

src/commands/cmd_key.cc
```diff
@@ -553,6 +553,68 @@ class CommandSort : public Commander {
...
 };
 
+class CommandGetMeta : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() != 2) {
+      return {Status::RedisExecErr, errWrongNumOfArguments};
+    }
+    return Status::OK();
+  }
```
PragmaTwice: it can be removed.

ltagliamonte-dd:
> To avoid collision with future Redis commands, we can prefix with something like `K` or `KVR`, e.g. `KMETADATA`.

love KMETADATA!

src/commands/cmd_key.cc
ltagliamonte-dd: for my current use case I'm working with redis hashes.
I added this other cases to make the cmd support future use cases that someone else (or myself) may have to work on.
is it a blocker to leave this is?

ltagliamonte-dd: i'm not following completely this suggestion do you mind giving more details?

src/commands/cmd_key.cc
```diff
@@ -553,6 +553,62 @@ class CommandSort : public Commander {
...
+        break;
+      case kRedisList:
+        metadata = std::make_unique<ListMetadata>();
+        break;
+      default:
+        return {Status::RedisExecErr, "Unimplemented Redis type"};
+    }
+
+    // Get metadata
+    s = redis.GetMetadata(ctx, {type}, nskey, metadata.get());
```
PragmaTwice:
```suggestion
    Metadata metadata;
    s = redis.GetMetadata(ctx, RedisTypes::All(), nskey, &metadata);
```

ltagliamonte-dd:
@torwig @git-hulk @PragmaTwice can i please get a CR on this?

tests/gocase/unit/kmetadata/kmetadata_test.go
```diff
@@ -0,0 +1,276 @@
...
+		case int64:
+			response.ttl = v
+		case int:
+			response.ttl = int64(v)
+		case float64:
+			response.ttl = int64(v)
+		default:
+			return nil, fmt.Errorf("ttl is not a number, got %T", val)
+		}
+	}
```
git-hulk: Can we extract this type's checker using a dedicated function to avoid duplicate codes?

src/commands/cmd_key.cc
```diff
@@ -553,6 +553,27 @@ class CommandSort : public Commander {
...
 
+class CommandKMetadata : public Commander {
+ public:
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    redis::Database redis(srv->storage, conn->GetNamespace());
+    std::string &key = args_[1];
+    std::string nskey = redis.AppendNamespacePrefix(key);
+
+    // Get metadata
+    Metadata metadata(kRedisNone);
```
PragmaTwice:
```suggestion
    Metadata metadata(kRedisNone, false);
```

src/commands/cmd_key.cc
```diff
@@ -553,6 +553,27 @@ class CommandSort : public Commander {
...
 
+class CommandKMetadata : public Commander {
+ public:
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    redis::Database redis(srv->storage, conn->GetNamespace());
+    std::string &key = args_[1];
+    std::string nskey = redis.AppendNamespacePrefix(key);
+
+    // Get metadata
+    Metadata metadata(kRedisNone, false);
```
PragmaTwice:
```suggestion
    Metadata metadata(kRedisNone, false);
```

src/commands/cmd_key.cc
```diff
@@ -553,6 +553,26 @@ class CommandSort : public Commander {
...
+
+    Metadata metadata(kRedisNone, false);
+    auto s = redis.GetMetadata(ctx, RedisTypes::All(), nskey, &metadata);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = conn->Map({{redis::BulkString("type"), redis::BulkString(RedisTypeNames[metadata.Type()])},
+                         {redis::BulkString("size"), redis::Integer(metadata.size)},
+                         {redis::BulkString("ttl"), redis::Integer(metadata.expire)},
+                         {redis::BulkString("flags"), redis::Integer(metadata.flags)},
+                         {redis::BulkString("version"), redis::Integer(metadata.version)}});
```
PragmaTwice: if `metadata.IsSingleKVType()` is true, we can remove `size` and `version` here since it's useless.

ltagliamonte-dd: thanks for the comment, it is true that those fields have no importance for singleKV, but should make easier to unmarshall the cmd response if we keep all the fields, wdyt?

src/commands/cmd_key.cc
```diff
@@ -553,6 +553,26 @@ class CommandSort : public Commander {
...
+    std::string &key = args_[1];
+    std::string nskey = redis.AppendNamespacePrefix(key);
+
+    Metadata metadata(kRedisNone, false);
+    auto s = redis.GetMetadata(ctx, RedisTypes::All(), nskey, &metadata);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = conn->Map({{redis::BulkString("type"), redis::BulkString(RedisTypeNames[metadata.Type()])},
+                         {redis::BulkString("size"), redis::Integer(metadata.size)},
+                         {redis::BulkString("ttl"), redis::Integer(metadata.expire)},
```
PragmaTwice:
```suggestion
                         {redis::BulkString("expire"), redis::Integer(metadata.expire)},
```

git-hulk (dismissed):

src/commands/cmd_key.cc
PragmaTwice: I think we should not give users useless information, even if it can be easier to decode. IMHO you can set a default value and use the default value if the field is missing in your application.

git-hulk:
@ltagliamonte-dd Need to fix this flaky test which was introduced by this PR: https://github.com/apache/kvrocks/actions/runs/13964426962/job/39095603519?pr=2827#step:17:91

ltagliamonte-dd:
> @ltagliamonte-dd Need to fix this flaky test which was introduced by this PR: https://github.com/apache/kvrocks/actions/runs/13964426962/job/39095603519?pr=2827#step:17:91

Fixed via [c05654d](https://github.com/apache/kvrocks/pull/2827/commits/c05654deb450b0b03c007d2611bdab5e92c27673)

src/commands/cmd_key.cc
ltagliamonte-dd: @PragmaTwice For the sideloading use case when IsSingleKVType() you will need version to perform the key update. So the only "useless" field would be size in that case, am I wrong?

ltagliamonte-dd: addressed via 10ced99e1695e9abef90a7deb8e21e1b00b77602

ltagliamonte:
@git-hulk @torwig @PragmaTwice can you please run the CI again it failed for a flaky test.
My PR should be good to go if you folks don't have any other comments.

git-hulk (dismissed):

src/commands/cmd_key.cc
```diff
@@ -553,6 +553,32 @@ class CommandSort : public Commander {
...
+    redis::Database redis(srv->storage, conn->GetNamespace());
+    std::string &key = args_[1];
+    std::string nskey = redis.AppendNamespacePrefix(key);
+
+    Metadata metadata(kRedisNone, false);
+    auto s = redis.GetMetadata(ctx, RedisTypes::All(), nskey, &metadata);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    if (metadata.IsSingleKVType()) {
+      *output = conn->Map({{redis::BulkString("type"), redis::BulkString(RedisTypeNames[metadata.Type()])},
```
PragmaTwice:
```suggestion
      *output = conn->Map({{redis::BulkString("type"), redis::BulkString(metadata.TypeName())},
```

src/commands/cmd_key.cc
```diff
@@ -553,6 +553,32 @@ class CommandSort : public Commander {
...
+    Metadata metadata(kRedisNone, false);
+    auto s = redis.GetMetadata(ctx, RedisTypes::All(), nskey, &metadata);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    if (metadata.IsSingleKVType()) {
+      *output = conn->Map({{redis::BulkString("type"), redis::BulkString(metadata.TypeName())},
+                           {redis::BulkString("expire"), redis::Integer(metadata.expire)},
+                           {redis::BulkString("flags"), redis::Integer(metadata.flags)}});
+    } else {
+      *output = conn->Map({{redis::BulkString("type"), redis::BulkString(RedisTypeNames[metadata.Type()])},
```
PragmaTwice:
```suggestion
      *output = conn->Map({{redis::BulkString("type"), redis::BulkString(metadata.TypeName())},
```

PragmaTwice (approved):

PR #2828, opened 2025-03-13
feat: Add DELPREFIX to delete keys by prefix 

iihimanshuu:
Description of Changes

I introduced updates in two files: commander.cc and cmd_key.cc. The primary changes involve the CommandDelPrefix and RegisterToCommandTable functions. Additionally, there are numerous lines of code changes due to the application of clang-format, which reformatted the codebase for consistency and readability.

solution for issue: https://github.com/apache/kvrocks/issues/2628


iihimanshuu:
when I format document with clang format it shows many line of code changes 
Screenshot 2025-03-14 095137
Screenshot 2025-03-14 095213

src/commands/cmd_key.cc
```diff
@@ -336,25 +336,25 @@ class CommandPersist : public Commander {
   }
 };
 
-class CommandDel : public Commander {
```
PragmaTwice: Some problems:
- `CommandDel` cannot be removed;
- `CommandDelPrefix` should be in this file, instead of another file (if you really build the project with your changes, I think you can find compile errors);
- there should be some test case for this new command

PragmaTwice (changes requested):
Sorry but this is far away from being merged. Also the format CI still failed for this PR, same as your previous PRs (actually you don't need to open so many PRs, you can just open one PR and add new changes into it).

src/commands/cmd_key.cc
```diff
@@ -356,6 +356,30 @@ class CommandDel : public Commander {
...
+        return {Status::RedisParseErr,
+                "wrong number of arguments for 'delprefix' command"};
+      }
+  
+      std::string prefix = args[1];
+      auto db = conn->GetNamespace();
+      auto keys = server->storage->GetKeysWithPrefix(db, prefix);
+  
+      int deleted_count = 0;
+      for (const auto &key : keys) {
```
RiversJin: Perhaps using Rocksdb's deleterange interface would be better?

tests/persistence/delprefix.tcl
PragmaTwice, line 1: Hmmm sorry but we don't use TCL to write test cases, from a long time ago. You can write test cases in golang.

iihimanshuu:
Screenshot 2025-03-21 001242
the build test was successful!

src/commands/cmd_key.cc
```diff
@@ -336,6 +340,51 @@ class CommandPersist : public Commander {
   }
 };
 
+class CommandDelPrefix : public Commander {
+ public:
+  CommandDelPrefix() : Commander() {}
+
+  Status Execute(engine::Context & /*ctx*/, Server *srv, Connection * /*conn*/, std::string *output) override {
```
SharonIV0x86, line 362:
```suggestion
  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
```

The function parameters are commented out.

iihimanshuu: Hi @SharonIV0x86,

Thank you for your review and suggestion. The function parameters `ctx` and `conn` were commented out to suppress the compiler warnings for unused parameters. However, if you think it's better to keep them uncommented for consistency or future use, I'm open to making that change.

Please let me know your thoughts.

iihimanshuu:
@PragmaTwice please review once..

RiversJin:
I just thought of another thing. Currently, kvrocks in cluster mode distributes data across shards based on slots. In the implementation, keys are prefixed with their hash value when stored, but it seems your delprefix implementation might not be accounting for this prefix handling. Have you considered how the hash prefix affects the key matching in the delprefix operation?
Therefore, I believe this feature might be difficult to implement reliably in cluster mode. To avoid inconsistencies, it might be better to explicitly disable the delprefix command when kvrocks is operating in cluster mode.

iihimanshuu:
Hello, @Reviewer

I appreciate your thoughtful comments. You make a good point when you mention that the hash prefix in cluster mode may cause problems with key matching.


### Present Implementation: The hash prefix used in cluster mode is not currently taken into consideration by the delprefix command, which may result in inconsistent results when trying to match and remove keys.

In order to remedy this, I suggest the following:
1. deactivate delprefix in Cluster Mode: When kvrocks is running in cluster mode 2, we can specifically deactivate the delprefix command to avoid irregularities. Future Consideration: We can investigate more reliable prefix deletion methods that appropriately manage the hash prefix in cluster mode. Before making deletions, this may require extra logic to extract and match the hash prefix.

### Next Steps:
I will update the implementation to disable the delprefix command in cluster mode and push the changes for review. Please let me know if you have any further suggestions or if this approach seems reasonable.

Thank you for your valuable input.

frHimanshu:
@PragmaTwice Request for Review

src/commands/cmd_key.cc
```diff
@@ -336,6 +340,56 @@ class CommandPersist : public Commander {
...
+    // Suppress unused parameter warnings
+    (void)ctx;
+    (void)conn;
+
+    if (args_.size() < 2) return {Status::NotOK, "Missing prefix argument"};
+
+    // Placeholder for cluster mode check
+    // if (srv->IsClusterMode()) {
+    //   return {Status::NotOK, "delprefix command is disabled in cluster mode to avoid inconsistencies"};
+    // }
```
PragmaTwice: please don't put commented code block into the repo.

iihimanshuu:
@PragmaTwice request for review

iihimanshuu:
@git-hulk Request to review commits

iihimanshuu:
I would love to hear some suggestions..

PR #2829, opened 2025-03-14
fix(cluster): Resolve forbidden slot range cleanup bug during slot mgration

RiversJin:


tests/gocase/integration/slotmigrate/slotmigrate_test.go
```diff
@@ -1334,6 +1334,21 @@ func TestSlotRangeMigrate(t *testing.T) {
...
 	})
 
+	t.Run("MIGRATE - Migrate back a proper subset range", func(t *testing.T) {
+		migrateSlotRangeAndSetSlot(t, ctx, rdb0, rdb1, id1, "3100-3400")
+		time.Sleep(1 * time.Second)
+		migrateSlotRangeAndSetSlot(t, ctx, rdb1, rdb0, id0, "3200-3300")
+		time.Sleep(1 * time.Second)
+
+		key := "AAA" // CLUSTER KEYSLOT AAA is `3205`, which is in the range of `3200-3500`
+		if rdb0.ClusterKeySlot(ctx, key).Val() != 3205 {
```
torwig: You can use `require.EqualValues` here instead of this `if` check.

RiversJin: thx, ok

git-hulk:
Hi @RiversJin, could you please elaborate a bit on how this happens?

mapleFU (commented):
`SlotMigrator::syncWALByCmd()` would first sync without forbidden the slot, then it would forbidden the slot and fetch the incremental. `SlotMigrator::finishFailedMigration()` would cleanup the `forbidden_slot_`. What raise my curious is that why it didn't solved in the success or cleanup path?

RiversJin:
> `SlotMigrator::syncWALByCmd()` would first sync without forbidden the slot, then it would forbidden the slot and fetch the incremental. `SlotMigrator::finishFailedMigration()` would cleanup the `forbidden_slot_`. What raise my curious is that why it didn't solved in the success or cleanup path? `SlotMigrator::syncWALByCmd()` 

My colleague speculated that this is to allow GetMigrationInfo to print out the migrated slots after migration.

Since after migration, the controller needs to execute setslot to confirm the migration, these forbidden slots could technically be cleared once the migration is confirmed by the controller. However, this approach might be cumbersome that the confirmation could be via setslot or setnodes, and we also need to account for edge cases (e.g., the target shard being deleted?).

So from the perspective of the simplest fix, directly clearing the forbidden slot at the import stage during re-migration would also work.

mapleFU:
> So from the perspective of the simplest fix, directly clearing the forbidden slot at the import stage during re-migration would also work.

The original `forbidden_slot_` is a bit weird, since it's like understanding incremental sync is stopped, and this slot can be Imported, if this is for same slot "point", I think it's a good idea. However for subrange, things would become a bit weird since part of the import can break the whole abandon range?

RiversJin:
> can break the whole abandon range

Yes. 
There‘s an implicit assumption here that only one migration task can exist for a shard at the same time. When a sub-range is being migrated back, it indicates the previous migration-out task has been completed and confirmed by the controller, making this approach valid.

Of course, I noticed there‘s a TODO in the code about supporting concurrent migration tasks. Once this TODO is addressed, we'll need a different solution.

By the way, I think assigning a unique ID to each migration task could help simplify the problem.

mapleFU:
> There‘s an implicit assumption here that only one migration task can exist for a shard at the same time. When a sub-range is being migrated back, it indicates the previous migration-out task has been completed and confirmed by the controller, making this approach valid.

Fair enough, you can add a comment to explain this

RiversJin:
> Fair enough, you can add a comment to explain this

Okay, comments have been added. Due to the code changes, we need to restart the test pipeline.

mapleFU (approved):

git-hulk (approved):

PR #2838, opened 2025-03-20
chore(metadata): wrap RedisTypeNames as a method

PragmaTwice:
We can wrap `RedisTypeNames[i]` inside a method so that it should be more easy to use.

### DIFF
--- a/src/search/search_encoding.h
+++ b/src/search/search_encoding.h
@@ -40,6 +40,8 @@ class IndexMetadata {
   uint8_t flag = 0;  // all reserved
   IndexOnDataType on_data_type;
 
+  const std::string &OnDataTypeString() const { return RedisTypeNames[(size_t)on_data_type]; }
+
   void Encode(std::string *dst) const {
     PutFixed8(dst, flag);
     PutFixed8(dst, uint8_t(on_data_type));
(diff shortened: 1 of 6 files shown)


src/storage/redis_metadata.cc
```diff
@@ -222,6 +222,8 @@ bool Metadata::operator==(const Metadata &that) const {
 
 RedisType Metadata::Type() const { return static_cast<RedisType>(flags & METADATA_TYPE_MASK); }
 
+const std::string &Metadata::TypeString() const { return RedisTypeNames[Type()]; }
```
torwig: `TypeString` => `TypeName`.

src/search/search_encoding.h
```diff
@@ -40,6 +40,8 @@ class IndexMetadata {
   uint8_t flag = 0;  // all reserved
   IndexOnDataType on_data_type;
 
+  const std::string &OnDataTypeString() const { return RedisTypeNames[(size_t)on_data_type]; }
```
torwig: For me, with the `On` prefix the function sounds like a callback :)

PragmaTwice: Haha make sense. But I think we can now leave it as is to avoid massive change.

src/storage/redis_metadata.cc
PragmaTwice: Done : )

torwig (approved):

PR #2839, opened 2025-03-20
feat: add deleterange command and test

baobaomaomeng:
GH-2628
Added DELETERANGE command to delete keys by pattern,use like scan.The command format translates to English as: DELETERANGE cursor [MATCH pattern] [COUNT count] [TYPE type].The default parameters align with the SCAN command.

### DIFF
--- a/src/commands/cmd_key.cc
+++ b/src/commands/cmd_key.cc
@@ -23,6 +23,7 @@
 #include "commander.h"
 #include "commands/ttl_util.h"
 #include "error_constants.h"
+#include "rocksdb/slice.h"
 #include "server/redis_reply.h"
 #include "server/server.h"
 #include "storage/redis_db.h"
@@ -356,6 +357,33 @@ class CommandDel : public Commander {
   }
 };
 
+class CommandDeleteRange : public Commander {
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() != 3) {
+      if (args.size() != 2) {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+      if (args[1].find('*') != args[1].size() - 1) {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+      args_[1] = args[1].substr(0, args[1].size() - 1);
+    }
+    return Status::OK();
+  }
+
+ public:
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    rocksdb::Slice key_begin, key_end;
+    key_begin = args_[1];
+    key_end = args_.size() == 3 ? args_[2] : "";
+    redis::Database redis(srv->storage, conn->GetNamespace());
+    auto s = redis.DeleteRange(ctx, key_begin, key_end);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+    *output = redis::RESP_OK;
+    return Status::OK();
+  }
+};
+
 class CommandRename : public Commander {
  public:
   Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
@@ -595,6 +623,7 @@ REDIS_REGISTER_COMMANDS(Key, MakeCmdAttr<CommandTTL>("ttl", 2, "read-only", 1, 1
                         MakeCmdAttr<CommandPExpireTime>("pexpiretime", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandDel>("del", -2, "write no-dbsize-check", 1, -1, 1),
                         MakeCmdAttr<CommandDel>("unlink", -2, "write no-dbsize-check", 1, -1, 1),
+                        MakeCmdAttr<CommandDeleteRange>("deleterange", -2, "write exclusive", 1, -1, 1),
                         MakeCmdAttr<CommandRename>("rename", 3, "write", 1, 2, 1),
                         MakeCmdAttr<CommandRenameNX>("renamenx", 3, "write", 1, 2, 1),
                         MakeCmdAttr<CommandCopy>("copy", -3, "write", 1, 2, 1),
--- a/src/storage/redis_db.h
+++ b/src/storage/redis_db.h
@@ -111,6 +111,7 @@ class Database {
   [[nodiscard]] rocksdb::Status Expire(engine::Context &ctx, const Slice &user_key, uint64_t timestamp);
   [[nodiscard]] rocksdb::Status Del(engine::Context &ctx, const Slice &user_key);
   [[nodiscard]] rocksdb::Status MDel(engine::Context &ctx, const std::vector<Slice> &keys, uint64_t *deleted_cnt);
+  [[nodiscard]] rocksdb::Status DeleteRange(engine::Context &ctx, const Slice &start, const Slice &end);
   [[nodiscard]] rocksdb::Status Exists(engine::Context &ctx, const std::vector<Slice> &keys, int *ret);
   [[nodiscard]] rocksdb::Status TTL(engine::Context &ctx, const Slice &user_key, int64_t *ttl);
   [[nodiscard]] rocksdb::Status GetExpireTime(engine::Context &ctx, const Slice &user_key, uint64_t *timestamp);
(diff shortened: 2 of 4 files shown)


mapleFU (commented):
Seems that this command is implemented as scan and MDel?

I'm not sure should `deleterange` supports prefix delete ( which could implemented with the help of `rocksdb::DeleteRange` ) or a scan and delete script like this? @git-hulk @PragmaTwice @ShooterIT

git-hulk:
> Seems that this command is implemented as scan and MDel?
> 
> I'm not sure should `deleterange` supports prefix delete ( which could implemented with the help of `rocksdb::DeleteRange` ) or a scan and delete script like this? @git-hulk @PragmaTwice @ShooterIT

From my side, I think the prefix deletion could be a part of this command and we should use `rocksdb::DeleteRange` to avoid the performance issue if possible.

And for the `DeleteRange` command, I have one question: is it good to support the `TYPE` option? which might heavily affect the performance since it needs to scan and retrieve the metadata.

baobaomaomeng:
> > Seems that this command is implemented as scan and MDel?
> > I'm not sure should `deleterange` supports prefix delete ( which could implemented with the help of `rocksdb::DeleteRange` ) or a scan and delete script like this? @git-hulk @PragmaTwice @ShooterIT
> 
> From my side, I think the prefix deletion could be a part of this command and we should use `rocksdb::DeleteRange` to avoid the performance issue if possible.
> 
> And for the `DeleteRange` command, I have one question: is it good to support the `TYPE` option? which might heavily affect the performance since it needs to scan and retrieve the metadata.
may be we need two commands,deletepattern and deleterange,the range delete.Handling special operations for range deletion in regular expressions is overly challenging. Additionally, the 'type' is decoded from iter->value(), and I don't quite understand why it requires scanning metadata when the entire implementation seems to be just a combination of scan + mdel.

baobaomaomeng:
> > Seems that this command is implemented as scan and MDel?
> > I'm not sure should `deleterange` supports prefix delete ( which could implemented with the help of `rocksdb::DeleteRange` ) or a scan and delete script like this? @git-hulk @PragmaTwice @ShooterIT
> 
> From my side, I think the prefix deletion could be a part of this command and we should use `rocksdb::DeleteRange` to avoid the performance issue if possible.
> 
> And for the `DeleteRange` command, I have one question: is it good to support the `TYPE` option? which might heavily affect the performance since it needs to scan and retrieve the metadata.

may be we need two commands,deletepattern and deleterange,the range delete.Handling special operations for range deletion in regular expressions is overly challenging. Additionally, the 'type' is decoded from iter->value(), and I don't quite understand why it requires scanning metadata when the entire implementation seems to be just a combination of scan + mdel.

git-hulk:
> Additionally, the 'type' is decoded from iter->value(), and I don't quite understand why it requires scanning metadata when the entire implementation seems to be just a combination of scan + mdel.

For example, we have 1,000,000 strings and 100 hash keys with random key ranges. And we might need to scan the entire database if we would like to delete keys by the hash type.

And yes, it makes sense to use SCAN + DEL if we could limit the maximum count on each delete. But perhaps it would be better to use `SCAN` + `rocksdb::DeleteRange` if possible.

PragmaTwice:
I don't think SCAN is necessary if we have a good semantics and syntax of this command, e.g. for `DELETERANGE <prefix>`, we don't need to SCAN, instead we can just call `DeleteRange`, right?

git-hulk:
> e.g. for DELETERANGE <prefix>, we don't need to SCAN, instead we can just call DeleteRange, right?

Yes, this would work if we don't support the `COUNT` option.

torwig:
> > e.g. for DELETERANGE , we don't need to SCAN, instead we can just call DeleteRange, right?
> 
> Yes, this would work if we don't support the `COUNT` option.

We usually have `COUNT` to limit the work that the server should do (avoid long-running execution).

baobaomaomeng:
> I don't think SCAN is necessary if we have a good semantics and syntax of this command, e.g. for `DELETERANGE <prefix>`, we don't need to SCAN, instead we can just call `DeleteRange`, right?

i 

> I don't think SCAN is necessary if we have a good semantics and syntax of this command, e.g. for `DELETERANGE <prefix>`, we don't need to SCAN, instead we can just call `DeleteRange`, right?

understand，i will change it for next day.

git-hulk:
> We usually have COUNT to limit the work that the server should do (avoid long-running execution).

@torwig We could simply use `rocksdb::DeleteRange` if we don't need to support the `COUNT`. And it would be fast since it just appends a write batch in rocksdb.

baobaomaomeng:
@git-hulk @PragmaTwice @ShooterIT 

Now, deleterange is a command with a clear semantic for range deletion. This command can be used in two ways:

1.deleterange begin_key end_key, which deletes key-value pairs in the range [begin_key, end_key).
2.deleterange prefix, which deletes key-value pairs with the specified prefix (the prefix must end with * to indicate it is a prefix).

src/commands/cmd_key.cc
```diff
@@ -595,6 +623,7 @@ REDIS_REGISTER_COMMANDS(Key, MakeCmdAttr<CommandTTL>("ttl", 2, "read-only", 1, 1
                         MakeCmdAttr<CommandPExpireTime>("pexpiretime", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandDel>("del", -2, "write no-dbsize-check", 1, -1, 1),
                         MakeCmdAttr<CommandDel>("unlink", -2, "write no-dbsize-check", 1, -1, 1),
+                        MakeCmdAttr<CommandDeleteRange>("deleterange", -2, "write no-dbsize-check", 1, -1, 1),
```
mapleFU: At least, a basic command might need to acquire the key lock, and further more, the deleteRange should be exclusive since it doesn't acquire any keylock when scanning?

baobaomaomeng: > At least, a basic command might need to acquire the key lock, and further more, the deleteRange should be exclusive since it doesn't acquire any keylock when scanning?

yes，rockdb imple deleterange by rangedelete flag，this action only write a flag to memtable

baobaomaomeng: you‘re right

mapleFU (commented):
As for typos, can we check https://github.com/crate-ci/typos/issues/316#issuecomment-2554111826?

src/storage/redis_db.h
```diff
@@ -111,6 +111,7 @@ class Database {
   [[nodiscard]] rocksdb::Status Expire(engine::Context &ctx, const Slice &user_key, uint64_t timestamp);
   [[nodiscard]] rocksdb::Status Del(engine::Context &ctx, const Slice &user_key);
   [[nodiscard]] rocksdb::Status MDel(engine::Context &ctx, const std::vector<Slice> &keys, uint64_t *deleted_cnt);
+  [[nodiscard]] rocksdb::Status DeleteRange(engine::Context &ctx, const Slice &start, const Slice &end);
```
mapleFU, line 114:
```suggestion
  [[nodiscard]] rocksdb::Status DeleteRange(engine::Context &ctx, Slice start, Slice end);
```

src/commands/cmd_key.cc
```diff
@@ -356,6 +357,33 @@ class CommandDel : public Commander {
...
+class CommandDeleteRange : public Commander {
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() != 3) {
+      if (args.size() != 2) {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+      if (args[1].find('*') != args[1].size() - 1) {
+        return {Status::RedisParseErr, errInvalidSyntax};
+      }
+      args_[1] = args[1].substr(0, args[1].size() - 1);
```
mapleFU, line 369: So this is a regex rule?

mapleFU (commented):
What about a "deleteprefix" command?

Current impl looks ok to me, however it supports two kind of cmd:

```
deleterange begin-key end-key
deleterange prefix
```

The (1) deletes `[begin-key, end-key)`, which may need re-delete end-key. The second just delete prefix. I think using same command is ok but maybe it's not a good idea for an dangerous operation like this

jihuayu:
@baobaomaomeng Hi there, I noticed there hasn't been much activity on this PR for a while. Just checking in—are you still working on this?

baobaomaomeng:
no，pelease close it




许世豪
***@***.***

PR #2845, opened 2025-03-21
feat(storage): support for sideloading SSTs

ltagliamonte-dd:
Dear community,

We're working on leveraging RocksDB sideloading to ship updates to our servers instead of loading data through the Redis interface.

Sideloading in RocksDB is more efficient than direct writes because it bypasses the standard write path, reducing write amplification, CPU overhead, and compaction costs. By ingesting pre-built SST files via IngestExternalFile(), it minimizes WAL writes and MemTable pressure, making bulk inserts faster and more efficient.

Since we don't use replication in our setup, we can ship bulk updates directly to each Kvrocks server and load the data efficiently.

We’d love to contribute our sideloading work back to the community and are excited to share our approach. Looking forward to your thoughts!

**Our approach**:
The current approach introduce a new CMD: `SST LOAD path`, which detects the files to load from a folder, and sideloads the data into RocksDB.


src/cli/main.cc
```diff
@@ -52,6 +52,18 @@ extern "C" void SignalHandler(int sig) {
   }
 }
 
+extern "C" void SigUsr1lHandler(int sig) {
```
PragmaTwice: It seems weird that we use a custom signal to trigger the ingest operation. Maybe we can consider to introduce a new command.

git-hulk:
@ltagliamonte-dd @PragmaTwice Do you think it's a good idea to provide a command for this purpose? 

For example:

```
LOAD SST <local dir>
```
So that it would be easy to ingest SSTs online.

ltagliamonte:
@PragmaTwice @git-hulk I'm not married to the use of a signal.
I can definitely add a new server command.
Are you folks ok with the rest of the proposal?

PragmaTwice:
> @ltagliamonte-dd @PragmaTwice Do you think it's a good idea to provide a command for this purpose?
> 
> For example:
> 
> ```
> LOAD SST <local dir>
> ```
> 
> So that it would be easy to ingest SSTs online.

Better to be `SST LOAD`.

git-hulk:
> > @ltagliamonte-dd @PragmaTwice Do you think it's a good idea to provide a command for this purpose?
> > For example:
> > ```
> > LOAD SST <local dir>
> > ```
> > So that it would be easy to ingest SSTs online.
> 
> Better to be `SST LOAD`.

I proposed  `LOAD SST` for the sake of a command would be better to start with its behavior.

PragmaTwice:
But the command name will be `LOAD` if it is. Maybe `LOADSST` or `SSTLOAD` would be better.

`SST LOAD` is similiar to `RDB LOAD`.

git-hulk:
> But the command name will be `LOAD` if it is. Maybe `LOADSST` or `SSTLOAD` would be better.
> 
> `SST LOAD` is similiar to `RDB LOAD`.

Makes sense.

ltagliamonte-dd:
@git-hulk @PragmaTwice re-worked as new CMD, thanks for the feedback

kvrocks.conf
```diff
@@ -1052,5 +1052,10 @@ rocksdb.partition_filters yes
 # Default: 0
 rocksdb.max_compaction_bytes 0
 
+# RocksDB sideloading move_files
+# Move files to the database directory instead of copying them.
+# Default: yes
+rocksdb.sideloading.move_files yes
```
PragmaTwice: It can be in command arguments instead of a config option.

src/storage/storage.cc
```diff
@@ -769,6 +775,65 @@ rocksdb::Status Storage::FlushScripts(engine::Context &ctx, const rocksdb::Write
...
+  if (!default_files.empty()) {
+    status = ingestSST(db_->DefaultColumnFamily(), default_ingest_opts_, default_files);
+    if (!status.ok()) {
+      return status;
+    }
+  }
+  // Process metadata files
+  if (!metadata_files.empty()) {
+    status = ingestSST(GetCFHandle(ColumnFamilyID::Metadata), default_ingest_opts_, metadata_files);
+  }
```
PragmaTwice: I think we have other column families besides these two.

src/storage/storage.cc
```diff
@@ -769,6 +775,65 @@ rocksdb::Status Storage::FlushScripts(engine::Context &ctx, const rocksdb::Write
...
+  std::vector<std::string> metadata_files;
+
+  // Sort files into appropriate vectors based on filename
+  for (const auto &file : sst_files) {
+    if (file.find("metadata") != std::string::npos) {
+      metadata_files.push_back(file);
+    } else {
+      default_files.push_back(file);
+    }
+  }
```
PragmaTwice: Seems too ad-hoc here? Why not just use column family name instead of hard-coded string here?

src/storage/storage.cc
```diff
@@ -769,6 +775,65 @@ rocksdb::Status Storage::FlushScripts(engine::Context &ctx, const rocksdb::Write
   return Write(ctx, options, batch->GetWriteBatch());
 }
 
+rocksdb::Status Storage::IngestSST(const std::string &sst_dir, int* files_loaded) {
```
```suggestion
StatusOr<int> Storage::IngestSST(const std::string &sst_dir) {
```

caipengbo:
I feel that this is not a production-usable feature without considering replication. This scenario is too limiting.

ltagliamonte:
> I feel that this is not a production-usable feature without considering replication. This scenario is too limiting.

Thank you for the feedback! I completely understand your concern. Please know that this initial version of SST load doesn't rule out the possibility of adding replication in the future. We're committed to ensuring that nothing we implement is backwards incompatible.

git-hulk:
> > I feel that this is not a production-usable feature without considering replication. This scenario is too limiting.
> 
> Thank you for the feedback! I completely understand your concern. Please know that this initial version of SST load doesn't rule out the possibility of adding replication in the future. We're committed to ensuring that nothing we implement is backwards incompatible.

Yes, we could allow this command only if it's a standalone instance without any replicas.

src/storage/storage.cc
ltagliamonte-dd: @PragmaTwice I see the following CF defined:
```
enum class ColumnFamilyID : uint32_t {
  PrimarySubkey = 0,
  Metadata,
  SecondarySubkey,
  PubSub,
  Propagate,
  Stream,
  Search,
};
```
from a quick look at this we could hardcode `default` and search in the folder for all the files that matches those CF? honestly i don't think all CF make sense, eager to hear your opinion.

ltagliamonte-dd:
> > > I feel that this is not a production-usable feature without considering replication. This scenario is too limiting.
> > 
> > 
> > Thank you for the feedback! I completely understand your concern. Please know that this initial version of SST load doesn't rule out the possibility of adding replication in the future. We're committed to ensuring that nothing we implement is backwards incompatible.
> 
> Yes, we could allow this command only if it's a standalone instance without any replicas.

addressed via 68ef7f821a916a2b33324e41bc78340700f504eb

kvrocks.conf
ltagliamonte-dd: addessed via c9fd7bf9e16b865482895c6984d08d62ba5d4fc5

src/storage/storage.cc
ltagliamonte-dd: addressed via e7907eb6db91814a55a89e36bf5bb7c0913a6598

ltagliamonte-dd:
@PragmaTwice @git-hulk I believe i've addressed all your comments, the only open one is the handling of other CF.
I've also added unit tests. Looking forward to your feedbacks!

.github/workflows/kvrocks.yaml
```diff
@@ -558,6 +558,8 @@ jobs:
           echo "DEBUG START"
           ldconfig -p | grep libstdc++
           gcc -print-file-name=libstdc++.so
+          export LIBRARY_PATH=$LIBRARY_PATH:/lib64
+          export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/lib64
```
PragmaTwice: Hmmm why?

x.py
```diff
@@ -267,9 +267,14 @@ def download_package(bindir: str) -> None:
     env = os.environ.copy()
     env.update({
         "CGO_CFLAGS": f"-I{rocksdb}",
-        "CGO_LDFLAGS": f"-L{rocksdb_lib} -L{zlib} -L{z4lib} -L{snappy_lib}",
+        "CGO_LDFLAGS": f"-L{rocksdb_lib} -L{zlib} -L{z4lib} -L{snappy_lib} -L/lib64 -L/usr/lib64 -lstdc++",
+        "LD_LIBRARY_PATH": f"/lib64:/usr/lib64:{env.get('LD_LIBRARY_PATH', '')}"
```
PragmaTwice: Could you explain this change? Otherwise we need to revert it.

ltagliamonte: Hello @PragmaTwice please ignore the changes in x.py until I'm done fixing the ci that's failing on my fork.

PragmaTwice:
The function itself looks good to me. Could you make the implementation more generic? I think in Kvrocks we want a good, complete and well-thought-out implementation instead of "oh, I wrote some code and it seems to work well in my case."

ltagliamonte:
> The function itself looks good to me. Could you make the implementation more generic? I think in Kvrocks we want a good, complete and well-thought-out implementation instead of "oh, I wrote some code and it seems to work well in my case."

Thanks I get it I just would like to get more directions on what do you think it's a better/generic design.
As I said I'm not familiar with other CF since my use case is 100% redis hashes.

git-hulk:
@ltagliamonte How could we generate the SST if we would like to use this command? I raise this question because I prefer using the same rocksdb version instead of introducing gorocksdb. So I'm wondering if we should support generating SSTs first by providing a subcommand for Kvrocks.

cc @apache/kvrocks-committers

ltagliamonte:
> @ltagliamonte How could we generate the SST if we would like to use this command? I raise this question because I prefer using the same rocksdb version instead of introducing gorocksdb. So I'm wondering if we should support generating SSTs first by providing a subcommand for Kvrocks.
> 
> cc @apache/kvrocks-committers

Hello @git-hulk in this codebase the use of the gorocksdb library has been introduced just for the unit tests.
I thought about just dropping few pre generated SST files but I didn't think it was the right approach.
I also think that by looking at the unit test code I've added it's going to be extremely simple for any other programming language to generate correct SST files.

git-hulk:
> > @ltagliamonte How could we generate the SST if we would like to use this command? I raise this question because I prefer using the same rocksdb version instead of introducing gorocksdb. So I'm wondering if we should support generating SSTs first by providing a subcommand for Kvrocks.
> > cc @apache/kvrocks-committers
> 
> Hello @git-hulk in this codebase the use of the gorocksdb library has been introduced just for the unit tests. I thought about just dropping few pre generated SST files but I didn't think it was the right approach. I also think that by looking at the unit test code I've added it's going to be extremely simple for any other programming language to generate correct SST files.

@ltagliamonte-dd  Yes, I know the gorocksdb is only for test purposes. I see you put a lot of effort into making the CI pass and generating SSTs, which is also a key step for using this command. So I think if we could provide the tool first and reuse this in our test cases.

ltagliamonte:
> > > @ltagliamonte How could we generate the SST if we would like to use this command? I raise this question because I prefer using the same rocksdb version instead of introducing gorocksdb. So I'm wondering if we should support generating SSTs first by providing a subcommand for Kvrocks.
> > > cc @apache/kvrocks-committers
> > 
> > 
> > Hello @git-hulk in this codebase the use of the gorocksdb library has been introduced just for the unit tests. I thought about just dropping few pre generated SST files but I didn't think it was the right approach. I also think that by looking at the unit test code I've added it's going to be extremely simple for any other programming language to generate correct SST files.
> 
> @ltagliamonte-dd Yes, I know the gorocksdb is only for test purposes. I see you put a lot of effort into making the CI pass and generating SSTs, which is also a key step for using this command. So I think if we could provide the tool first and reuse this in our test cases.

@git-hulk what kind of tool do you envision? we can think of adding some tooling in a later pr, it's a good idea but it shouldn't block this one wdyt?

git-hulk:
> > > > @ltagliamonte How could we generate the SST if we would like to use this command? I raise this question because I prefer using the same rocksdb version instead of introducing gorocksdb. So I'm wondering if we should support generating SSTs first by providing a subcommand for Kvrocks.
> > > > cc @apache/kvrocks-committers
> > > 
> > > 
> > > Hello @git-hulk in this codebase the use of the gorocksdb library has been introduced just for the unit tests. I thought about just dropping few pre generated SST files but I didn't think it was the right approach. I also think that by looking at the unit test code I've added it's going to be extremely simple for any other programming language to generate correct SST files.
> > 
> > 
> > @ltagliamonte-dd Yes, I know the gorocksdb is only for test purposes. I see you put a lot of effort into making the CI pass and generating SSTs, which is also a key step for using this command. So I think if we could provide the tool first and reuse this in our test cases.
> 
> @git-hulk what kind of tool do you envision? we can think of adding some tooling in a later pr, it's a good idea but it shouldn't block this one wdyt?

Yes, it won't block this PR. I intend to avoid putting a lot of time into making the CI correct by first building the SSTs tool.

src/commands/cmd_server.cc
```diff
@@ -1370,6 +1370,51 @@ class CommandPollUpdates : public Commander {
   Format format_ = Format::Raw;
 };
 
+class CommandSST : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() < 3) {
+      return {Status::RedisParseErr, "Invalid number of arguments"};
```
```suggestion
      return {Status::RedisParseErr, errWrongNumOfArguments};
```

src/commands/cmd_server.cc
```diff
@@ -1370,6 +1370,51 @@ class CommandPollUpdates : public Commander {
...
+    }
+    std::string cmd = util::ToLower(args[1]);
```
git-hulk: Could use `Commandparser` here.

src/commands/cmd_server.cc
```diff
@@ -1410,5 +1455,6 @@ REDIS_REGISTER_COMMANDS(Server, MakeCmdAttr<CommandAuth>("auth", 2, "read-only o
                         MakeCmdAttr<CommandReset>("reset", 1, "ok-loading bypass-multi no-script", NO_KEY),
                         MakeCmdAttr<CommandApplyBatch>("applybatch", -2, "write no-multi", NO_KEY),
                         MakeCmdAttr<CommandDump>("dump", 2, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandPollUpdates>("pollupdates", -2, "read-only admin", NO_KEY), )
+                        MakeCmdAttr<CommandPollUpdates>("pollupdates", -2, "read-only admin", NO_KEY),
+                        MakeCmdAttr<CommandSST>("sst", -3, "write admin", 1, 1, 1), )
```
git-hulk: Need to add the `exclusive` flag for this command.

src/storage/storage.cc
git-hulk: @ltagliamonte-dd The SST file writer will record the column family if it's assigned, so I think we could find the column family through the SST table properties when loading.

Please refer to https://github.com/facebook/rocksdb/blob/743a02d6f68811bed8c5f3ce5d7e6d9b6bc18783/table/sst_file_writer.cc#L396

danishgh (approved):

danishgh (approved):

src/storage/storage.cc
PragmaTwice: You can get CF names here: https://github.com/apache/kvrocks/blob/unstable/src/storage/storage.h#L197

ltagliamonte-dd: addressed this via 0ca11358751c7748ca4a0d5228ec4e620bc609aa
please let me know if you have any other feedbacks

ltagliamonte-dd:
@git-hulk @PragmaTwice i've addressed all comments and the CI is working.
looking forward to any other feedback you may have for me to address.
Thanks for the help in seeing this complete!

src/storage/storage.cc
ltagliamonte-dd: @git-hulk I looked at your proposal about inserting the CF name in the sstfile and It can't be done atm.
although the SstFileWriter supports the ColumnFamilyHandle parameter, the only 2 exposed method do not:
https://github.com/facebook/rocksdb/blob/v10.0.1/include/rocksdb/c.h#L2301-L2307
So until this is supported upstream (I can sponsor that effort, and improve this code in a second PR) the only way to identify CF seems to use the file name unfortunately.

git-hulk:
CI is blocked by https://github.com/apache/kvrocks/issues/2868

ltagliamonte:
> CI is blocked by #2868

FYI Ci run successfully on my end https://github.com/ltagliamonte-dd/kvrocks/actions/runs/14274994555

x.py
```diff
@@ -128,7 +128,7 @@ def build(dir: str, jobs: Optional[int], ghproxy: bool, ninja: bool, unittest: b
 
     os.makedirs(dir, exist_ok=True)
 
-    cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo"]
+    cmake_options = ["-DCMAKE_BUILD_TYPE=RelWithDebInfo", "-DCMAKE_POLICY_VERSION_MINIMUM=3.5"]
```
PragmaTwice: can be reverted now after #2870.

x.py
```diff
@@ -223,6 +223,60 @@ def clang_tidy(dir: str, jobs: Optional[int], clang_tidy_path: str, run_clang_ti
     run(run_command, *options, *regexes, verbose=True, cwd=basedir)
 
 
+def is_rocky_linux():
```
PragmaTwice, line 226: hmmm seems in our CI the rocky linux job works well now.

ltagliamonte-dd: looks like it is needed to do CGO https://github.com/ltagliamonte-dd/kvrocks/actions/runs/14361821853/job/40266547259

ltagliamonte-dd: putting back this change fixes the CI for rocky linux: https://github.com/ltagliamonte-dd/kvrocks/actions/runs/14363316453

ltagliamonte-dd: when we put it [back](https://github.com/ltagliamonte-dd/kvrocks/commit/3fcf93884454d7e74e92937e4337de15e73797a9) CI works for rocky linux https://github.com/ltagliamonte-dd/kvrocks/actions/runs/14363316453

x.py
ltagliamonte-dd: with or without the build seems to fail, is the darwin build flacky?
i've revert as asked.

ltagliamonte:
@PragmaTwice @git-hulk can I please get a final review for this pr?

PragmaTwice:
Thank you for your effort. All of these build & CI issues will be fixed soon in the seperate PRs, since it is unrelated to the goal of this PR.

ltagliamonte:
> Thank you for your effort. All of these build & CI issues will be fixed soon in the seperate PRs, since it is unrelated to the goal of this PR.

Thanks @PragmaTwice.
Is the PR mergeable?

PragmaTwice:
> Thanks @PragmaTwice. Is the PR mergeable?

Before that, we need to:
- wait #2876 to be merged to retrigger the CI of this PR;
- revert the changes to `x.py` and `.github/workflows` (if it's really needed, we can open another PR.)

ltagliamonte:
@PragmaTwice the changes to x.py and workflows in this PR are needed because of the CGO requirement to compile grocksdb in the unit tests.

PragmaTwice:
> @PragmaTwice the changes to x.py and workflows in this PR are needed because of the CGO requirement to compile grocksdb in the unit tests.

Ahh I got it. It's due to this PR brings grocksdb into the golang test. I'll check it later.

ltagliamonte:
@PragmaTwice did you had a chance to look at the PR?

ltagliamonte:
@PragmaTwice @git-hulk any updates on merging this?

src/commands/cmd_server.cc
```diff
@@ -1370,6 +1370,49 @@ class CommandPollUpdates : public Commander {
   Format format_ = Format::Raw;
 };
 
+class CommandSST : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() < 3) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
```
git-hulk: Can remove this check since this command has required at least 3 parameters when building attributes, so it will be checked before executing.
```suggestion
```

src/commands/cmd_server.cc
```diff
@@ -1370,6 +1370,49 @@ class CommandPollUpdates : public Commander {
...
+        }
+      } else {
+        return {Status::RedisParseErr, "unknown option: " + parser.TakeStr().GetValue()};
+      }
+    }
+    return Commander::Parse(args);
+  }
+
+  Status Execute([[maybe_unused]] engine::Context &ctx, Server *srv, [[maybe_unused]] Connection *conn,
+                 std::string *output) override {
```
git-hulk, line 1401: We could check the following conditions and not allow the load:

- The instance is in cluster mode
- The instance has any replicas or is in replica mode.

src/storage/storage.cc
```diff
@@ -769,6 +770,101 @@ rocksdb::Status Storage::FlushScripts(engine::Context &ctx, const rocksdb::Write
...
+    for (const auto &cf_name : cf_names) {
+      if (file.find(cf_name) != std::string::npos) {
+        cf_files[cf_name].push_back(file);
+        matched = true;
+        break;
+      }
+    }
+    // If no match found, assume it belongs to default CF
+    if (!matched) {
+      cf_files[rocksdb::kDefaultColumnFamilyName].push_back(file);
```
git-hulk: It would be better to return an error if the column family is mismatched.

.github/workflows/kvrocks.yaml
```diff
@@ -162,11 +163,13 @@ jobs:
             without_jemalloc: -DDISABLE_JEMALLOC=ON
             with_sanitizer: -DENABLE_ASAN=ON
             compiler: gcc
+            ignore_when_asan: -tags="ignore_when_asan"
```
git-hulk, line 166: Where's using this tag?

git-hulk:
@ltagliamonte-dd I left a few comments, and other looks good to me. To see if @PragmaTwice and @mapleFU have time to take another pass.

.github/workflows/kvrocks.yaml
ltagliamonte-dd: this flag is used here https://github.com/ltagliamonte-dd/kvrocks/blob/sideloading/.github/workflows/kvrocks.yaml#L336

src/commands/cmd_server.cc
ltagliamonte-dd: this check is perfomed in https://github.com/ltagliamonte-dd/kvrocks/blob/sideloading/src/storage/storage.cc#L773-L776
since i have already access to _config.

src/storage/storage.cc
ltagliamonte-dd: addressed via 04d0aa3dd4e9ff1d64e9b0337c831aeba29528c4

src/commands/cmd_server.cc
ltagliamonte-dd: @git-hulk how to check `The instance has any replicas or is in replica mode.`?

ltagliamonte-dd: addressed via 66545dada56d37b1090d00e5244723cb6e1ec8c8

src/commands/cmd_server.cc
```diff
@@ -1370,6 +1370,52 @@ class CommandPollUpdates : public Commander {
...
+    if (srv->GetConfig()->cluster_enabled) {
+      return {Status::NotOK, "The SST command is not supported in cluster mode."};
+    }
+    if (srv->IsSlave()) {
```
git-hulk, line 1405: Need to disallow if it has any replicas.

git-hulk (dismissed):

ltagliamonte-dd:
@git-hulk is there a method already there I can use to check if the node
has replicas?

git-hulk:
> @git-hulk is there a method already there I can use to check if the node has replicas?

You could add a method in `Server` to export the replica count, for example：

```C++

size_t GetReplicaCount() {
   slave_threads_mu_.lock();
   auto replica_count = slave_threads_.size();
   slave_threads_mu_.unlock();
   return replica_count;
} 

```

src/commands/cmd_server.cc
ltagliamonte-dd: addressed via ce55b4f027f67ebc3bac455f419fd35f0a74ec76

git-hulk (dismissed):

git-hulk:
@ltagliamonte Thanks for your contribution and patience.

ltagliamonte-dd:
> @ltagliamonte Thanks for your contribution and patience.

thank you @git-hulk.
Looking forward to pull the nightly!!

src/storage/storage.cc
```diff
@@ -769,6 +770,98 @@ rocksdb::Status Storage::FlushScripts(engine::Context &ctx, const rocksdb::Write
   return Write(ctx, options, batch->GetWriteBatch());
 }
 
+StatusOr<int> Storage::IngestSST(const std::string &sst_dir, const rocksdb::IngestExternalFileOptions &ingest_options) {
+  std::vector<std::string> sst_files;
+  DIR *dir = opendir(sst_dir.c_str());
```
caipengbo: Using the RocksDB API is better:

https://github.com/apache/kvrocks/blob/18e6674eecbfe3361891883f7777c1d188e5ba25/src/storage/storage.cc#L1069

ltagliamonte-dd: Hello @caipengbo the GetChildren does basically the same, put all files in a dir in a vector, why it would be better?

caipengbo: 1. RocksDB's API is more portable 
2. The overall code style is more unified

src/storage/storage.cc
```diff
@@ -769,6 +770,98 @@ rocksdb::Status Storage::FlushScripts(engine::Context &ctx, const rocksdb::Write
...
+    }
+  }
+  closedir(dir);
+
+  if (sst_files.empty()) {
+    LOG(WARNING) << "No SST files found in " << sst_dir;
+    return 0;
+  }
+
+  // Create a map to store SST files for each column family
```
caipengbo: I think this comment and some of the comments below are redundant. 

**Good code is comments.**

x.py
```diff
@@ -223,6 +223,60 @@ def clang_tidy(dir: str, jobs: Optional[int], clang_tidy_path: str, run_clang_ti
...
+        output = run_pipe("grep", "-v", "32",stdin=output)
+        output = run_pipe("xargs", "dirname", stdin=output)
+        libstdc_folder = output.read().strip() + "/"
+
+    if libstdc_folder != "":
+        additional_flags = f"-L{libstdc_folder} -lstdc++"
+
+    with_cov = env.get("WITH_COVERAGE","false")
+    if with_cov == "true":
+        additional_flags += "-lgcov"
```
PragmaTwice, line 271: Do we need to retrieve coverage info for go test cases? Maybe we can just ignore them?

src/storage/storage.cc
ltagliamonte-dd: addressed via 66eb43e0d356751f3c7b32ea5e4490eb68108a02

src/storage/storage.cc
ltagliamonte-dd: addressed via 66eb43e0d356751f3c7b32ea5e4490eb68108a02

x.py
ltagliamonte-dd: this is necessary to support the sonarcloud flag in kvrocks.yaml
```sonarcloud: -DCMAKE_CXX_FLAGS=--coverage```
otherwise build for sonarcloud fails

**Review by caipengbo — DISMISSED** [review]

File: src/storage/storage.cc [hunk 21]
```diff
@@ -769,6 +770,90 @@ rocksdb::Status Storage::FlushScripts(engine::Context &ctx, const rocksdb::Write
...
+  }
+
+  std::unordered_map<std::string_view, std::vector<std::string>> cf_files;
+  std::vector<std::string> cf_names;
+  for (const auto &cf : ColumnFamilyConfigs::ListAllColumnFamilies()) {
+    cf_names.emplace_back(cf.Name());
+  }
+  for (const auto &cf_name : cf_names) {
+    cf_files[cf_name] = std::vector<std::string>();
+  }
```
PragmaTwice: Seems this loop is redundant?

git-hulk: @ltagliamonte-dd You can take a look at this comment. And we could move forward to merging the PR if no more comments.

ltagliamonte-dd: addressed via b2e04e802c979355a76cadd322e06890821b9a60

src/storage/storage.cc
```diff
@@ -794,9 +794,7 @@ StatusOr<int> Storage::IngestSST(const std::string &sst_dir, const rocksdb::Inge
   std::vector<std::string> cf_names;
   for (const auto &cf : ColumnFamilyConfigs::ListAllColumnFamilies()) {
     cf_names.emplace_back(cf.Name());
-  }
-  for (const auto &cf_name : cf_names) {
-    cf_files[cf_name] = std::vector<std::string>();
+    cf_files[cf.Name()] = std::vector<std::string>();
```
git-hulk: What @PragmaTwice should be is that we don't need to initialize the empty vector for the map?

PragmaTwice: Yeah we can just remove this line.

src/storage/storage.cc
```diff
@@ -769,6 +770,88 @@ rocksdb::Status Storage::FlushScripts(engine::Context &ctx, const rocksdb::Write
...
+  if (sst_files.empty()) {
+    LOG(WARNING) << "No SST files found in " << sst_dir;
+    return 0;
+  }
+
+  std::unordered_map<std::string_view, std::vector<std::string>> cf_files;
+  std::vector<std::string> cf_names;
+  for (const auto &cf : ColumnFamilyConfigs::ListAllColumnFamilies()) {
+    cf_names.emplace_back(cf.Name());
+    cf_files[cf.Name()] = std::vector<std::string>();
```
git-hulk:
```suggestion
```

git-hulk (approved):

src/storage/storage.cc
ltagliamonte: Thanks!
@PragmaTwice can you please approve?

git-hulk:
@caipengbo @PragmaTwice To see you guys have further comments, or we could merge after the CI becomes green.

caipengbo (approved):
LGTM

Issue #2807, opened 2025-02-26
TDigest: Implement `TDIGEST.CDF` command


LindaSummer:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Refer: https://github.com/apache/kvrocks/issues/2558
We should implement `TDIGEST.CDF` command.

Redis command reference:
https://redis.io/docs/latest/commands/tdigest.cdf/

### Solution

Depends on #2791 #2792.

Follow the implementation of #2791 for this command.

- Add cpp unit tests for it.
- Add go integration tests for it.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

LindaSummer:
Hi Team,

I'm working on this issue now. 😊
I plan to add more unit tests and integration tests for it.

Best Regards,
Edward

PR #2849, opened 2025-03-23
feat(tdigest): Add support for QUANTILE command

### LINKED ISSUE #2794: TDigest: Implement QUANTILE command for TDigest Algorithm

SharonIV0x86:
Closes #2794 

This PR adds support for ``TDIGEST.QUANTILE`` command.


danishgh (approved):

SharonIV0x86:
@LindaSummer I have added the basic structure for the ``QUANTILE`` command but as you mentioned in the issue, there seems to be a problem with the locking mechanism or something else.


```
E20250330 23:50:42.491079 130596879718080 redis_tdigest.cc:524] metadata has 24 merged nodes, but got 4
```

I tried modifying the quantile function as mentioned in the issue and the previous PRs but the issue persists. Any help is appreciated.

LindaSummer:
Hi @SharonIV0x86 ,

Thanks very much for effort!😊

I'll try to test it in local in recent days and will sync if there is any new updates.

Best Regards,
Edward

SharonIV0x86:
> Hi @SharonIV0x86 ,
> 
> Thanks very much for effort!😊
> 
> I'll try to test it in local in recent days and will sync if there is any new updates.
> 
> Best Regards, Edward

Works, till then ill try to fiddle around. Cheers.

LindaSummer:
Hi @SharonIV0x86 ,

Sorry for delay of updating. 😊

In recent I have some personal affairs which occupied my schedule so maybe a little delay in response.

I'm now working on this ticket and find that there may be some bugs inside it. I will double confirm it and create a PR to fix it if bug exists.

Best Regards,
Edward

SharonIV0x86:
> Hi @SharonIV0x86 ,
> 
> Sorry for delay of updating. 😊
> 
> In recent I have some personal affairs which occupied my schedule so maybe a little delay in response.
> 
> I'm now working on this ticket and find that there may be some bugs inside it. I will double confirm it and create a PR to fix it if bug exists.
> 
> Best Regards, Edward

No issues, meanwhile i tried looking into it, no matter what i did there seems to be a mismatch between the actual information and the tdigest metadata now one thing i can deduce is that is not in the Quantile function but this seems to be the issue with ``mergeCurrentBuffer`` function or some other centroids function being called from the Quantile function.

Or this is related to locks as you suggested. 

There is no hurry from my side in fixing this issue but i am also planning to implement the [TDIGEST.CDF](https://github.com/apache/kvrocks/issues/2807) and [TDIGEST.RANK](https://redis.io/docs/latest/commands/tdigest.rank/) in upcoming weeks and they most likely will depend in this issue.

LindaSummer:
Hi @SharonIV0x86 ,

I have tested the command line successfully with removing the lock and announcing the command "write" for quick validation after fixing #2878 .

I will try to find a way to improve the performance with less critical section.

Best Regards,
Edward

SharonIV0x86:
> Hi @SharonIV0x86 ,
> 
> I have tested the command line successfully with removing the lock and announcing the command "write" for quick validation after fixing #2878 .
> 
> I will try to find a way to improve the performance with less critical section.
> 
> Best Regards, Edward

Okay, thanks for looking into it. So, how should i proceed then? should i wait for #2878 to be merged?

LindaSummer:
> > Hi @SharonIV0x86 ,
> > I have tested the command line successfully with removing the lock and announcing the command "write" for quick validation after fixing #2878 .
> > I will try to find a way to improve the performance with less critical section.
> > Best Regards, Edward
> 
> Okay, thanks for looking into it. So, how should i proceed then? should i wait for #2878 to be merged?

Hi @SharonIV0x86 ,

Maybe we could wait for it to be merged before this PR.
It will affect our go integration test cases.

If you don't mind, I will try to solve the lock issue since I find that it may not be so easy to just add a lock key in current connection lock management.
After this PR solved, we could follow it for other commands like `CDF`.

Best Regards,
Edward

SharonIV0x86:
> Hi @SharonIV0x86 ,
> 
> Maybe we could wait for it to be merged before this PR. It will affect our go integration test cases.
> 
> If you don't mind, I will try to solve the lock issue since I find that it may not be so easy to just add a lock key in current connection lock management. After this PR solved, we could follow it for other commands like `CDF`.
> 
> Best Regards, Edward

Absolutely no issues whatsoever, there is no hurry i am happy to wait 😄 
Let me know if there’s anything I can help with in the meantime.

SharonIV0x86:
@LindaSummer Hi, i have tested the quantile command after the merging of the #2878 and it works as expected.

However, there is a bit of difference in quantile values when compared to the original implementation of [ redis ``tdigest.quantile``](https://redis.io/docs/latest/commands/tdigest.quantile/) but i believe kvrocks quantile is more accurate as it uses linear interpolation so it should be good.

I have also added a go test case so if any changes are required pls lmk. 👍🏼

src/commands/cmd_tdigest.cc
```diff
@@ -242,11 +242,47 @@ class CommandTDigestMax : public CommandTDigestMinMax {
...
+        return {Status::RedisExecErr, errKeyNotFound};
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
+    std::vector<std::string> quantile_strings;
+    quantile_strings.reserve(result.quantiles.size());
+    for (const auto &q : result.quantiles) {
+      quantile_strings.push_back(std::to_string(q));
+    }
+    *output = redis::MultiBulkString(RESP::v2, quantile_strings);
```
PragmaTwice:
```suggestion
    *output = conn->MultiBulkString(quantile_strings);
```

src/commands/cmd_tdigest.cc
```diff
@@ -242,11 +242,47 @@ class CommandTDigestMax : public CommandTDigestMinMax {
...
+    for (size_t i = 2; i < args.size(); i++) {
+      auto value = ParseFloat(args[i]);
+      if (!value) {
+        return {Status::RedisParseErr, errValueIsNotFloat};
+      }
+      values_.push_back(*value);
+    }
+    return Status::OK();
+  }
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) {
```
PragmaTwice: please add `override` to them, as reported in CI.

src/commands/cmd_tdigest.cc
```diff
@@ -242,11 +242,47 @@ class CommandTDigestMax : public CommandTDigestMinMax {
  public:
   CommandTDigestMax() : CommandTDigestMinMax(false) {}
 };
+class CommandTDigestQuantile : public Commander {
+  Status Parse(const std::vector<std::string> &args) {
```
PragmaTwice: `override` also here. You can check CI logs to know this.
image

SharonIV0x86: mb, ill fix it.

LindaSummer:
Hi @SharonIV0x86 ,

I'm very sorry for delay in recent. 

My baby was born in recent, so I need to support my family these days.

Maybe I will be delay in response. I will go through this problem in this week as soon as I could.

Best Regards,
Edward

SharonIV0x86:
> Hi @SharonIV0x86 ,
> 
> I'm very sorry for delay in recent.
> 
> My baby was born in recent, so I need to support my family these days.
> 
> Maybe I will be delay in response. I will go through this problem in this week as soon as I could.
> 
> Best Regards, Edward

Absolutely no issues man take your time, there is no hurry with this and congrats btw👏🏼. 
The PRs and issues can always wait you should give time to your family right now.

src/commands/cmd_tdigest.cc
```diff
@@ -242,11 +242,47 @@ class CommandTDigestMax : public CommandTDigestMinMax {
...
+ private:
+  std::string key_name_;
+  std::vector<double> values_;
+};
 REDIS_REGISTER_COMMANDS(TDigest, MakeCmdAttr<CommandTDigestCreate>("tdigest.create", -2, "write", 1, 1, 1),
                         MakeCmdAttr<CommandTDigestInfo>("tdigest.info", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandTDigestAdd>("tdigest.add", -3, "write", 1, 1, 1),
                         MakeCmdAttr<CommandTDigestMax>("tdigest.max", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandTDigestMin>("tdigest.min", 2, "read-only", 1, 1, 1),
+                        MakeCmdAttr<CommandTDigestQuantile>("tdigest.quantile", -2, "read-only", 1, 1, 1),
```
LindaSummer:
```suggestion
                        MakeCmdAttr<CommandTDigestQuantile>("tdigest.quantile", -3, "read-only", 1, 1, 1),
```

tests/gocase/unit/type/tdigest/tdigest_test.go
```diff
@@ -310,4 +311,43 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
...
+
+		// Quantile on a non existent key
+		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.QUANTILE", keyPrefix+"iDoNotExist").Err(), errMsgKeyNotExist)
+
+		// Creating a key
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.CREATE", keyPrefix+"01", "compression", "100").Err())
+
+		key := keyPrefix + "01"
+		//Adding some data to tdigest 1 2 2 3 3 3 4 4 4 4 5 5 5 5 5
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.ADD", key, "1", "2", "2", "3", "3", "3", "4", "4", "4", "4", "5", "5", "5", "5", "5").Err())
```
LindaSummer: Maybe we could add some unordered inputs with some negative numbers as test cases. 😊

src/commands/cmd_tdigest.cc
```diff
@@ -242,11 +242,47 @@ class CommandTDigestMax : public CommandTDigestMinMax {
...
+        return {Status::RedisParseErr, errValueIsNotFloat};
+      }
+      values_.push_back(*value);
+    }
+    return Status::OK();
+  }
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    TDigest tdigest(srv->storage, conn->GetNamespace());
+    TDigestQuantitleResult result;
+    auto s = tdigest.Quantile(ctx, key_name_, values_, &result);
```
LindaSummer: We should handle the case for an empty tdigest in `TDigest::Quantile`.
Cpp unit test cases should also be adapted after we handle it in `TDigest::Quantile`.

Here is my test case in redis stack.
```
127.0.0.1:6389> flushall
OK
127.0.0.1:6389> tdigest.create hello
OK
127.0.0.1:6389> tdigest.quantile hello 0.1
1) "nan"
127.0.0.1:6389>
```

Here is the output of current implementation.
```
127.0.0.1:6666> flushall
OK
127.0.0.1:6666> tdigest.create hello
OK
127.0.0.1:6666> tdigest.quantile hello 0.1
(error) ERR Invalid argument: invalid quantile or empty tdigest
```

LindaSummer (commented):
Hi @SharonIV0x86 ,

Sorry for delay review. 😊

For the lock of quantile merging, I have checked the lock in `redis_connection.cc` and `Database::AppendNamespacePrefix`.

They lock the same key for same tdigest user key.

So, it is safe to lock the [`ns_key`](https://github.com/apache/kvrocks/blob/009a1bd80bdde88c091a69aca1539be0f58d2658/src/types/redis_tdigest.cc#L190-L193).

https://github.com/apache/kvrocks/blob/009a1bd80bdde88c091a69aca1539be0f58d2658/src/server/redis_connection.cc#L521

https://github.com/apache/kvrocks/blob/009a1bd80bdde88c091a69aca1539be0f58d2658/src/storage/redis_db.cc#L520-L522

Best Regards,
Edward

src/commands/cmd_tdigest.cc
SharonIV0x86: Sorry for delayed response i was busy with an event. About this yes, i have added this check locally but i will just push it in sometime.

tests/gocase/unit/type/tdigest/tdigest_test.go
SharonIV0x86: Sure will do that.

SharonIV0x86:
> Hi @SharonIV0x86 ,
> 
> Sorry for delay review. 😊
> 
> For the lock of quantile merging, I have checked the lock in `redis_connection.cc` and `Database::AppendNamespacePrefix`.
> 
> They lock the same key for same tdigest user key.
> 
> So, it is safe to lock the [`ns_key`](https://github.com/apache/kvrocks/blob/009a1bd80bdde88c091a69aca1539be0f58d2658/src/types/redis_tdigest.cc#L190-L193).
> 
> https://github.com/apache/kvrocks/blob/009a1bd80bdde88c091a69aca1539be0f58d2658/src/server/redis_connection.cc#L521
> 
> https://github.com/apache/kvrocks/blob/009a1bd80bdde88c091a69aca1539be0f58d2658/src/storage/redis_db.cc#L520-L522
> 
> Best Regards, Edward

Understood. So the current logic is fine and we can most probably refactor it into a function later on #2907

SharonIV0x86:
> Hi @SharonIV0x86 ,
> 
> Sorry for delay review. 😊
> 
> For the lock of quantile merging, I have checked the lock in `redis_connection.cc` and `Database::AppendNamespacePrefix`.
> 
> They lock the same key for same tdigest user key.
> 
> So, it is safe to lock the [`ns_key`](https://github.com/apache/kvrocks/blob/009a1bd80bdde88c091a69aca1539be0f58d2658/src/types/redis_tdigest.cc#L190-L193).
> 
> https://github.com/apache/kvrocks/blob/009a1bd80bdde88c091a69aca1539be0f58d2658/src/server/redis_connection.cc#L521
> 
> https://github.com/apache/kvrocks/blob/009a1bd80bdde88c091a69aca1539be0f58d2658/src/storage/redis_db.cc#L520-L522
> 
> Best Regards, Edward

Understood. So the current logic is fine and we can most probably refactor it into a function later on #2907

src/commands/cmd_tdigest.cc
SharonIV0x86: I can add another error constant for this in ``error_constants.h`` file if that works? Because we can use the same one for CDF, Rank and etc commands as well in future.

LindaSummer: > I can add another error constant for this in `error_constants.h` file if that works? Because we can use the same one for CDF, Rank and etc commands as well in future.

Hi @SharonIV0x86 ,

For my side, I think the "nan" is not an error and defining a literal string inside tdigest scope may be better. 😊

Best Regards,
Edward

src/commands/cmd_tdigest.cc
```diff
@@ -242,11 +242,50 @@ class CommandTDigestMax : public CommandTDigestMinMax {
...
+    if (!s.ok()) {
+      if (s.IsNotFound()) {
+        return {Status::RedisExecErr, errKeyNotFound};
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
+    if (values_.empty()) {
+      return {Status::RedisExecErr, "invalid quantile or empty tdigest"};
```
LindaSummer: Hi @SharonIV0x86 ,

Sorry for my unclear description for this behavior. 😄

Currently, redis stack returns nan as response for none centroids.
We'd better follow this behavior.

Adding the logic to tdigest ranther than command may be better to keep the command module clean.

After modifying the tdigest, the cpp unit test should also be updated.

Best Regards,
Edward

tests/gocase/unit/type/tdigest/tdigest_test.go
```diff
@@ -310,4 +311,66 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
 		infoAfterEmptyReset := toTdigestInfo(t, rsp.Val())
 		require.EqualValues(t, 100, infoAfterEmptyReset.Compression)
 	})
+	t.Run("tdigest.quantile with different arguments", func(t *testing.T) {
```
LindaSummer, line 314: Hi @SharonIV0x86 ,

We could also add an unordered sequence and an empty tdigest as cases. 😊

Best Regards,
Edward

LindaSummer (commented):
Hi @SharonIV0x86 ,

Thanks very much for your effort. 😊
Left some comments.

Best Regards,
Edward

PragmaTwice:
@LindaSummer Would you like to have a look?

LindaSummer:
> @LindaSummer Would you like to have a look?

Hi @PragmaTwice and @SharonIV0x86 ,

Of course!😊

I will go through this PR later today.

Best Regards,
Edward

src/commands/cmd_tdigest.cc
```diff
@@ -265,13 +265,16 @@ class CommandTDigestQuantile : public Commander {
...
-    if (values_.empty()) {
-      return {Status::RedisExecErr, "invalid quantile or empty tdigest"};
-    }
     std::vector<std::string> quantile_strings;
     quantile_strings.reserve(result.quantiles.size());
-    for (const auto &q : result.quantiles) {
-      quantile_strings.push_back(std::to_string(q));
+    if (!result.has_centroids) {
+      for (size_t i = 0; i < values_.size(); ++i) {
+        quantile_strings.emplace_back("nan");
```
LindaSummer: Hi @SharonIV0x86 ,

Maybe we could make "nan" a constexpr literal string in this file. 😊
It would also be used in other commands.

Best Regards,
Edward

src/types/redis_tdigest.cc
```diff
@@ -225,6 +225,10 @@ rocksdb::Status TDigest::Quantile(engine::Context& ctx, const Slice& digest_name
   if (auto status = dumpCentroids(ctx, ns_key, metadata, &centroids); !status.ok()) {
     return status;
   }
+  if (centroids.empty()) {
```
LindaSummer: Hi @SharonIV0x86 ,

At the first time we load the metadata, we should know this tdigest has no data inside it.
This could be checked before the centroids retrieving.

And it would be better to add cpp unit test for this behavior.

Best Regards,
Edward

LindaSummer (commented):
Hi @SharonIV0x86 ,

Left some comments.

Best Regards,
Edward

src/types/redis_tdigest.cc
SharonIV0x86: So just to be clear, after we load the metadata [here](https://github.com/apache/kvrocks/blob/6719b607141fd2b07f5d5a967971fc1d6de0f87a/src/types/redis_tdigest.cc#L191C19-L191C27) i add a check to do something like this.

```cpp
if(metadata.merged_nodes == 0 && metadata.unmerged_nodes == 0){
  //tdigest does not contain any data. so set the flag
}
```
Correct me if im wrong.

> And it would be better to add cpp unit test for this behavior.

I'll try working on this as well.

src/commands/cmd_tdigest.cc
SharonIV0x86: Good idea, will do this.

src/types/redis_tdigest.cc
LindaSummer: > So just to be clear, after we load the metadata [here](https://github.com/apache/kvrocks/blob/6719b607141fd2b07f5d5a967971fc1d6de0f87a/src/types/redis_tdigest.cc#L191C19-L191C27) i add a check to do something like this.
> 
> 
> 
> ```cpp
> 
> if(metadata.merged_nodes == 0 && metadata.unmerged_nodes == 0){
> 
>   //tdigest does not contain any data. so set the flag
> 
> }
> 
> ```
> 
> Correct me if im wrong.
> 
> 
> 
> > And it would be better to add cpp unit test for this behavior.
> 
> 
> 
> I'll try working on this as well.

Hi @SharonIV0x86 ,

We could directly use `total_observations` to check the tdigest's elements. 😊

Best Regards,
Edward

SharonIV0x86:
@LindaSummer I have added the cpp test case, i believe it needs some modifications because in the cpp test case its currently not possible to directly compare the values returned by the quantile function with the "nan" string as the ``result.quantile`` vector is returned empty if the tdigest is empty. The behaviors differ in the command implementation.

The closest i could do was to add a assert for the value of ``result.has_centroids`` and it should be true and if its found true then it can be assumed that in the ``Execute`` function of the command implementation will construct the "nan" strings vector and will return it as a response.

SharonIV0x86:
> @LindaSummer I have added the cpp test case, i believe it needs some modifications because in the cpp test case its currently not possible to directly compare the values returned by the quantile function with the "nan" string as the `result.quantiles` vector is returned empty if the tdigest is empty. The behaviors differ in the command implementation.
 
The closest i could do was to add a assert for the value of ``result.has_centroids`` and it should be true and if its found true then it can be assumed that in the ``Execute`` function of the command implementation will construct the "nan" strings vector and will return it as a response.

LindaSummer:
> @LindaSummer I have added the cpp test case, i believe it needs some modifications because in the cpp test case its currently not possible to directly compare the values returned by the quantile function with the "nan" string as the ``result.quantile`` vector is returned empty if the tdigest is empty. The behaviors differ in the command implementation.
> 
> 
> 
> The closest i could do was to add a assert for the value of ``result.has_centroids`` and it should be true and if its found true then it can be assumed that in the ``Execute`` function of the command implementation will construct the "nan" strings vector and will return it as a response. 

Hi @SharonIV0x86 ,

You could directly test the flag you have added. 😊
If you want a more clear semantic of the result, maybe `std::optional<std::vector<double>` is a choice.

And the nan string test could be in integration test.

Best Regards,
Edward

LindaSummer (approved):
Hi @SharonIV0x86 ,

Thanks very much for your contribution and huge effort! 😄
I have made some small changes on current code.

LGTM.

Best Regards,
Edward

PragmaTwice (approved):

SharonIV0x86:
> Hi @SharonIV0x86 ,
> 
> Thanks very much for your contribution and huge effort! 😄 I have made some small changes on current code.
> 
> LGTM.
> 
> Best Regards, Edward

Thanks for the changes you made, the code is much better now with ranges and views. 
I have opened a PR for the quantile command in kvrocks-website https://github.com/apache/kvrocks-website/pull/300

Issue #2841, opened 2025-03-21
Proposal for Handling Orphaned Slaves in Cluster Mode


RiversJin:
### Search before asking

- [ ] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Currently, there is a minor issue in Kvrocks cluster mode: When a slave node is deleted via the controller, the controller only notifies this change to nodes within the cluster. As a result, the deleted slave remains unaware of its removal and continues replication workflows, which is counterintuitive. This proposal aims to resolve this issue.

#### Key Considerations:
1. **Unreliable Controller-to-Node Notification**:  
   Relying on the controller to synchronize deletion events to the target node is unreliable. When a node is deleted, it often indicates an unhealthy state (e.g., network instability). We cannot guarantee successful delivery of deletion notifications (e.g., messages may arrive but fail to return acknowledgments). Thus, unlimited retries on the controller side are not a robust design.

2. **Protocol Flexibility**:  
   While Kvrocks' master-slave replication protocol is inspired by Redis, it is not identical. Introducing additional replication steps distinct from Redis is acceptable.

3. **Topology Synchronization Priority**:  
   The controller does not prioritize updating topology information for the current shard’s master. Consequently, masters must account for potentially stale topology data when validating slave replication requests and avoid outright rejection.

### Solution

1. **Slave Metadata Propagation**:  
   During `replconf` execution, slaves will send their **node ID** and **current config version** to the master. The master will store this metadata.

2. **Version-Based Validation**:  
   After processing `replconf` or `cluster setnodes` commands, the master will scan connections meeting the following criteria and return a specific error to trigger replication termination on the slave side:  
   - **Condition**: The master’s topology version is newer than the slave’s (`master_version > slave_version`), **and** the slave’s node ID no longer exists in the master’s topology.

3. **Cleanup Workflow**:  
   Slaves receiving the specific error will terminate replication and clean up their state.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

git-hulk:
@RiversJin Thanks for your proposal.

> During replconf execution, slaves will send their node ID and current config version to the master. The master will store this metadata.

Do you mean the master will keep this metadata inside the memory instead of persisting on disk?

RiversJin:
> Do you mean the master will keep this metadata inside the memory instead of persisting on disk?

Storing the id and version in memory is sufficient. Specifically, placing this information in the Connection object would be appropriate.

[...]

PR #2850, opened 2025-03-23
feat(cluster): disconnect replication connections to orphan slaves

RiversJin:
it solves https://github.com/apache/kvrocks/issues/2841


src/commands/cmd_replication.cc
```diff
@@ -166,41 +177,66 @@ class CommandReplConf : public Commander {
...
+      peer_version_ = *parse_result;
+      return Status::OK();
+    }
+
+    return {Status::RedisParseErr, errUnknownOption};
   }
 
   Status Execute([[maybe_unused]] engine::Context &ctx, [[maybe_unused]] Server *srv, Connection *conn,
                  std::string *output) override {
-    if (port_ != 0) {
```
RiversJin, line 191: Based on the `replconf` command definition, it's impossible to reach this code path without `port_` being set, so I've removed the validation check for `port_`.

PragmaTwice:
Thank you for your contribution! Could you add some test cases for it?

RiversJin:
> Thank you for your contribution! Could you add some test cases for it?

Of course, but save the boring tasks for weekdays :P.

RiversJin:
> Thank you for your contribution! Could you add some test cases for it?

@PragmaTwice  After a closer look, this PR doesn't introduce any new changes other than disconnecting data replication for nodes removed from the cluster. I added a small snippet of assertion code in cluster_test.go for this scenario. Do you think there are any other areas that might need more testing?

src/cluster/cluster.cc
```diff
@@ -23,24 +23,33 @@
...
 #include "parse_util.h"
 #include "replication.h"
 #include "server/server.h"
 #include "string_util.h"
 #include "time_util.h"
 
-ClusterNode::ClusterNode(std::string id, std::string host, int port, int role, std::string master_id,
-                         const std::bitset<kClusterSlots> &slots)
-    : id(std::move(id)), host(std::move(host)), port(port), role(role), master_id(std::move(master_id)), slots(slots) {}
+ClusterNode::ClusterNode(std::string &&id, std::string &&host, int port, int role, std::string &&master_id,
```
mapleFU: why these are changed?

src/cluster/cluster.cc
```diff
@@ -68,8 +77,8 @@ Status Cluster::SetNodeId(const std::string &node_id) {
 
   myid_ = node_id;
   // Already has cluster topology
-  if (version_ >= 0 && nodes_.find(node_id) != nodes_.end()) {
-    myself_ = nodes_[myid_];
+  if (version_ >= 0 && nodes_->count(node_id) > 0) {
+    myself_ = nodes_->at(node_id);
```
mapleFU: if the performance is important, why not use `find()` and iterator based operations? Which could make it just search once

src/cluster/cluster.cc
```diff
@@ -96,11 +105,13 @@ Status Cluster::SetSlotRanges(const std::vector<SlotRange> &slot_ranges, const s
...
 
   // Get the node which we want to assign slots into it
-  std::shared_ptr<ClusterNode> to_assign_node = nodes_[node_id];
-  if (to_assign_node == nullptr) {
+  auto it = nodes_->find(node_id);
+  if (it == nodes_->end()) {
     return {Status::NotOK, "No this node in the cluster"};
   }
 
+  auto to_assign_node = it->second;
```
mapleFU:
```suggestion
  std::shared_ptr<ClusterNode> to_assign_node;
  if (auto it = nodes_.find(); it != nodes_.end()) {
     to_assign_node = ..
  } else {
    return {Status::NotOK, "No this node in the cluster"};
  }
```

src/cluster/cluster.cc
```diff
@@ -197,8 +214,11 @@ Status Cluster::SetClusterNodes(const std::string &nodes_str, int64_t version, b
   }
 
   myself_ = nullptr;
-  if (!myid_.empty() && nodes_.find(myid_) != nodes_.end()) {
-    myself_ = nodes_[myid_];
+  if (!myid_.empty()) {
+    auto it = nodes_->find(myid_);
+    if (it != nodes_->end()) {
```
mapleFU: `if(auto it = ;`

src/cluster/cluster.cc
```diff
@@ -252,11 +272,14 @@ Status Cluster::SetMasterSlaveRepl() {
       srv_->slot_migrator->SetStopMigrationFlag(false);
       LOG(INFO) << "Change server role to master, restart migration task";
     }
+    if (!is_slave) {
+      srv_->CleanupOrphanSlaves(version_, *nodes_);
+    }
```
mapleFU: So this is what this patch actually does?

src/server/redis_connection.h
```diff
@@ -217,4 +229,44 @@ class Connection : public EvbufCallbackBase<Connection> {
...
+  uint32_t GetPort() const { return port_; }
+
+  std::string_view GetStringView() const { return str_; }
+  std::string_view GetPeerID() const { return peer_id_; }
+  std::string_view GetAddr() const { return addr_; }
+  int64_t GetPeerVersion() const { return peer_version_; }
+
+ private:
+  std::string_view ip_;
+  std::string_view addr_;
```
mapleFU: This looks really unsafe..Who would owns memory for them?

src/server/server.cc
```diff
@@ -354,6 +355,27 @@ void Server::DisconnectSlaves() {
...
 }
 
+void Server::CleanupOrphanSlaves(int64_t version, const ClusterNodes &nodes) {
+  std::lock_guard<std::mutex> lg(slave_threads_mu_);
+
+  for (auto &slave_thread : slave_threads_) {
+    const auto peer_info = slave_thread->GetConn()->GetPeerInfo();
+    auto peer_version = peer_info->GetPeerVersion();
+    if (peer_version < 0 || peer_version > version) {
+      // The peer version is greater than the current version,
```
mapleFU, line 365: what does `peer_version < 0` means?

src/server/server.cc
```diff
@@ -354,6 +355,27 @@ void Server::DisconnectSlaves() {
   }
 }
 
+void Server::CleanupOrphanSlaves(int64_t version, const ClusterNodes &nodes) {
+  std::lock_guard<std::mutex> lg(slave_threads_mu_);
+
+  for (auto &slave_thread : slave_threads_) {
+    const auto peer_info = slave_thread->GetConn()->GetPeerInfo();
```
mapleFU: do we need check `GetPeerInfo != nullptr`?

src/server/redis_connection.h
```diff
@@ -134,12 +139,18 @@ class Connection : public EvbufCallbackBase<Connection> {
...
+
+  void SetPeerInfo(std::unique_ptr<PeerInfo> &&peer_info) { peer_info_ = std::move(peer_info); }
+
+  const PeerInfo *GetPeerInfo() {
+    if (peer_info_) {
+      return peer_info_.get();
+    }
+
+    SetPeerInfo(std::make_unique<PeerInfo>(ip_, port_, "", -1));
+    return peer_info_.get();
```
mapleFU: Emmm when would this happens?

src/cluster/cluster.cc
PragmaTwice: Some notes:
1. for `std::bitset`, in most impl it is stored on the stack memory instead heap so move ctor has no difference from the copy ctor. usually you don't need to move it. 
2. for `ClassA(std::string x) : x(x) {}`, you can pass either lval or rval to it. for rval, it involves two move ctors. And for `ClassA(std::string &&x) : x(x) {}`, you can only pass rval to it, and also the performance difference is quite little or none.

So these changes are not so useful.

src/cluster/cluster.cc
```diff
@@ -977,3 +991,13 @@ Status Cluster::Reset() {
...
+
+// Note that if current version is lower than the given version,
+// it can't be determined whether the node is in the cluster, so just regard it as in the cluster.
+bool Cluster::IsInCluster(const std::string &node_id, int64_t version) const {
+  if (version < 0 || this->version_ < version) {
+    return true;
+  }
+
+  return nodes_->count(node_id) > 0;
+}
```
git-hulk, line 994:
```suggestion
}

```

src/cluster/cluster_defs.h
```diff
@@ -41,6 +41,7 @@ inline constexpr const char *errNoMasterNode = "The node isn't a master";
 inline constexpr const char *errClusterNoInitialized = "The cluster is not initialized";
 inline constexpr const char *errInvalidClusterNodeInfo = "Invalid cluster nodes info";
 inline constexpr const char *errInvalidImportState = "Invalid import state";
+inline constexpr const char *errYouAreFired = "You are fired";
```
git-hulk: I think it's not a right wording. Could be something like `disconnected due to the topology change` or others.

git-hulk:
@RiversJin, I found that this PR introduces many code styles and fixes, except for the disconnecting feature. Would you mind separating them into a few dedicated PRs that would be easy to review and clarify the context?

RiversJin:
> I found that this PR introduces many code styles and fixes, except for the disconnecting feature. Would you mind separating them into a few dedicated PRs that would be easy to review and clarify the context?

@git-hulk  Sure! The unrelated tweaks were legacy from our internal codebase. I've removed the obvious ones, but the iterator-based index optimization in Cluster::nodes_ (avoiding double lookups) is interleaved in some places. It's technically out of scope, but actually improves performance. but need some time to remove those...

PragmaTwice:
Yeah these changes in `parse_utils.h` looks good to me, so maybe you can open a separate PR for it.

PR #2863, opened 2025-04-01
feat: Add support of the command Migrate

### LINKED ISSUE #1949: Add support of the command MIGRATE
Refer to https://redis.io/commands/migrate/

clouder-zhai:
Close #1949 

Command Desc: 
1. add support for the basic dump command, which supports an atomic operation of dump and restore
2. the command only support the basic migrate command: MIGRATE host port key destination-db timeout

Redis doc: https://redis.io/docs/latest/commands/migrate/

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -31,10 +31,12 @@
 #include "common/time_util.h"
 #include "config/config.h"
 #include "error_constants.h"
+#include "redis_protocol.h"
 #include "server/redis_connection.h"
 #include "server/redis_reply.h"
 #include "server/server.h"
 #include "stats/disk_stats.h"
+#include "status.h"
 #include "storage/rdb/rdb.h"
 
 namespace redis {
@@ -1257,8 +1259,15 @@ class CommandDump : public Commander {
       return Status::OK();
     }
 
+    auto s = DumpKey(ctx, srv, conn, key);
+    if (!s.IsOK()) return s;
+    *output = redis::BulkString(s.GetValue());
+    return Status::OK();
+  }
+  static StatusOr<std::string> DumpKey(engine::Context &ctx, Server *srv, Connection *conn, const std::string &key) {
     RedisType type = kRedisNone;
-    db_status = redis.Type(ctx, key, &type);
+    redis::Database redis(srv->storage, conn->GetNamespace());
+    rocksdb::Status db_status = redis.Type(ctx, key, &type);
     if (!db_status.ok()) return {Status::RedisExecErr, db_status.ToString()};
 
     std::string result;
@@ -1267,8 +1276,7 @@ class CommandDump : public Commander {
     auto s = rdb.Dump(key, type);
     if (!s.IsOK()) return s;
     CHECK(dynamic_cast<RdbStringStream *>(rdb.GetStream().get()) != nullptr);
-    *output = redis::BulkString(static_cast<RdbStringStream *>(rdb.GetStream().get())->GetInput());
-    return Status::OK();
+    return static_cast<RdbStringStream *>(rdb.GetStream().get())->GetInput();
   }
 };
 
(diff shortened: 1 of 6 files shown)


src/commands/cmd_server.cc
```diff
@@ -1257,8 +1258,17 @@ class CommandDump : public Commander {
...
     }
 
+    std::string dump_result;
+    auto s = DumpKey(ctx, srv, conn, key, dump_result);
+    if (!s.IsOK()) return s;
+    *output = redis::BulkString(dump_result);
+    return Status::OK();
+  }
+  static Status DumpKey(engine::Context &ctx, Server *srv, Connection *conn, std::string &key,
+                        std::string &dump_result) {
```
```suggestion
  static StatusOr<std::string> DumpKey(engine::Context &ctx, Server *srv, Connection *conn, std::string &key) {
```

src/commands/cmd_server.cc
```diff
@@ -1370,6 +1380,86 @@ class CommandPollUpdates : public Commander {
   Format format_ = Format::Raw;
 };
 
+class CommandMigrate : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() != 6) {
+      return {Status::RedisExecErr, errWrongNumOfArguments};
+    }
```
```suggestion
```

src/commands/cmd_server.cc
```diff
@@ -1410,5 +1500,6 @@ REDIS_REGISTER_COMMANDS(Server, MakeCmdAttr<CommandAuth>("auth", 2, "read-only o
                         MakeCmdAttr<CommandReset>("reset", 1, "ok-loading bypass-multi no-script", NO_KEY),
                         MakeCmdAttr<CommandApplyBatch>("applybatch", -2, "write no-multi", NO_KEY),
                         MakeCmdAttr<CommandDump>("dump", 2, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandPollUpdates>("pollupdates", -2, "read-only admin", NO_KEY), )
+                        MakeCmdAttr<CommandPollUpdates>("pollupdates", -2, "read-only admin", NO_KEY),
+                        MakeCmdAttr<CommandMigrate>("migrate", -6, "write", 1, 1, 1))
```
```suggestion
                        MakeCmdAttr<CommandMigrate>("migrate", 6, "write", 1, 1, 1))
```

PragmaTwice: Also the key position spec is wrong (1,1,1).

src/commands/cmd_server.cc
clouder-zhai: when I run ./x.py format, it automatically becomes two lines

PragmaTwice: I mean we can use `StatusOr` here instead of `Status`.

src/commands/cmd_server.cc
```diff
@@ -1410,5 +1555,6 @@ REDIS_REGISTER_COMMANDS(Server, MakeCmdAttr<CommandAuth>("auth", 2, "read-only o
                         MakeCmdAttr<CommandReset>("reset", 1, "ok-loading bypass-multi no-script", NO_KEY),
                         MakeCmdAttr<CommandApplyBatch>("applybatch", -2, "write no-multi", NO_KEY),
                         MakeCmdAttr<CommandDump>("dump", 2, "read-only", 1, 1, 1),
-                        MakeCmdAttr<CommandPollUpdates>("pollupdates", -2, "read-only admin", NO_KEY), )
+                        MakeCmdAttr<CommandPollUpdates>("pollupdates", -2, "read-only admin", NO_KEY),
+                        MakeCmdAttr<CommandMigrate>("migrate", -6, "write", 1, -1, 1))
```
PragmaTwice, line 1559: `1, -1, 1` means that all arguments are keys, which is obviously not correct here.

You can refer to here for how to generate dynamic key ranges: https://github.com/apache/kvrocks/blob/unstable/src/commands/cmd_zset.cc#L1552

PR #2873, opened 2025-04-09
fix(protocol): inline mode should allow the quoted string

### LINKED ISSUE #2871: Kvrocks doesn't respect single quotes in --pipe mode
### Search before asking

### Version

Try to add a value with a space in between using redis-cli. Everything works correctly as long as you do it manually, using your keyboard:
`LPUSH kvrocks_bug 'lorem ipsum'
`
To check if everything is fine:
`LRANGE kvrocks_bug 0 -1`
`1) "lorem ipsum"`

But if you put the same exact LPUSH command into a redis file and then try to execute it using a --pipe parameter:
`> redis-cli --pipe </tmp/kvrocks.redis`

Then it doesn't seem to respect single quotes around 'lorem ipsum':
`> LRANGE kvrocks_bug 0 -1`
`1) "ipsum'"`
`2) "'lorem"`

If you do the same exact steps in redis then it works correctly. If you don't use --pipe parameter in redis-cli then it works correctly as well. It's just a problem when you pipe the file directly to the server.

### Minimal reproduce step

Create a simple file called kvrocks.redis and put a single line into it:
`LPUSH kvrocks_bug 'lorem ipsum'`

Then import the file using redis-cli with --pipe parameter:
`redis-cli --pipe </tmp/kvrocks.redis`

Make sure to provide all the authentication parameters.
Then go into kvrocks (you can use redis-cli in interactive mode) and execute:
`LRANGE kvrocks_bug 0 -1`

I expected to see just one value:
`1) "lorem ipsum"`

### What did you see instead?

There are two values:
`1) "ipsum'"
2) "'lorem"`

### Anything Else?

_No response_

git-hulk:
This closes #2871.

### DIFF
--- a/src/common/string_util.cc
+++ b/src/common/string_util.cc
@@ -276,6 +276,71 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
   return {prefix, ""};
 }
 
+StatusOr<std::vector<std::string>> SplitArguments(std::string_view in) {
+  std::vector<std::string> arguments;
+  std::string current_string;
+
+  enum { NORMAL, QUOTED, SINGLE_QUOTED, ESCAPE } state = NORMAL;
+
+  bool is_quoted = false;
+  for (const char c : in) {
+    switch (state) {
+      case NORMAL:
+        if (std::isspace(c)) {
+          if (!current_string.empty()) {
+            arguments.emplace_back(std::move(current_string));
+            current_string.clear();
+          }
+          // skip spaces
+        } else if (c == '"' || c == '\'') {
+          state = c == '"' ? QUOTED : SINGLE_QUOTED;
+          is_quoted = c == '"';
+        } else {
+          current_string.push_back(c);
+        }
+        break;
+      case QUOTED:
+      case SINGLE_QUOTED:
+        if (c == '\\') {
+          state = ESCAPE;
+        } else if ((c == '"' && state == QUOTED) || (c == '\'' && state == SINGLE_QUOTED)) {
+          state = NORMAL;
+        } else {
+          current_string.push_back(c);
+        }
+        break;
+      case ESCAPE:
+        if (c == '"' || c == '\'' || c == '\\') {
+          current_string.push_back(c);
+        } else if (c == 't') {
+          current_string.push_back('\t');
+        } else if (c == 'r') {
+          current_string.push_back('\r');
+        } else if (c == 'n') {
+          current_string.push_back('\n');
+        } else if (c == 'v') {
+          current_string.push_back('\v');
+        } else if (c == 'f') {
+          current_string.push_back('\f');
+        } else if (c == 'b') {
+          current_string.push_back('\b');
+        }
+        state = is_quoted ? QUOTED : SINGLE_QUOTED;
+        break;
+    }
+  }
+  if (state == QUOTED || state == SINGLE_QUOTED) {
+    return {Status::NotOK, "unclosed quote string"};
+  }
+  if (state == ESCAPE) {
+    return {Status::NotOK, "unexpected trailing escape character"};
+  }
+  if (!current_string.empty()) {
+    arguments.emplace_back(std::move(current_string));
+  }
+  return arguments;
+}
+
 std::vector<std::string> RegexMatch(const std::string &str, const std::string &regex) {
   std::regex base_regex(regex);
   std::smatch pieces_match;
--- a/src/common/string_util.cc
+++ b/src/common/string_util.cc
@@ -276,14 +276,26 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
   return {prefix, ""};
 }
 
+static int HexDigitToInt(const char c) {
+  if (c >= '0' && c <= '9') {
+    return c - '0';
+  } else if (c >= 'a' && c <= 'f') {
+    return c - 'a' + 10;
+  } else if (c >= 'A' && c <= 'F') {
+    return c - 'A' + 10;
+  }
+  return 0;
+}
+
 StatusOr<std::vector<std::string>> SplitArguments(std::string_view in) {
   std::vector<std::string> arguments;
   std::string current_string;
 
   enum State { NORMAL, DOUBLE_QUOTED, SINGLE_QUOTED, ESCAPE } state = NORMAL;
 
-  State state_before_escape = NORMAL;
-  for (const char c : in) {
+  bool done = false;
+  for (size_t i = 0; i < in.size() && !done; i++) {
+    const auto c = in[i];
     switch (state) {
       case NORMAL:
         if (std::isspace(c)) {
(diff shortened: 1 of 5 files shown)


tests/cppunit/string_util_test.cc
```diff
@@ -268,3 +268,41 @@ TEST(StringUtil, RegexMatchExtractSSTFile) {
...
+      // normal cases
+      {"a b c", {"a", "b", "c"}},
+      {"a\tb\nc\fd", {"a", "b", "c", "d"}},
+
+      //  quote cases
+      {"hello \"a b\" c", {"hello", "a b", "c"}},
+      {"'a b' c", {"a b", "c"}},
+
+      {"\"a\\\"b\" c", {"a\"b", "c"}},
+      {"'a\\' b' c", {"a' b", "c"}},
```
PragmaTwice: Could you add a test case like `command "a b" " c d "`, with multiple quoted string?

git-hulk: Sure, will have it later.

PragmaTwice: Also we can use unescaped raw string `R"(...)"` in C++.

git-hulk: @PragmaTwice Thanks for your tips. Will use the raw string to make it more clear.

src/common/string_util.cc
```diff
@@ -276,6 +276,71 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
...
+      case NORMAL:
+        if (std::isspace(c)) {
+          if (!current_string.empty()) {
+            arguments.emplace_back(std::move(current_string));
+            current_string.clear();
+          }
+          // skip spaces
+        } else if (c == '"' || c == '\'') {
+          state = c == '"' ? QUOTED : SINGLE_QUOTED;
+          is_quoted = c == '"';
```
PragmaTwice: It seems `state` itself is enough to determine whether it is quoted? Is `is_quoted` necessary?

src/common/string_util.cc
```diff
@@ -276,6 +276,71 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
   return {prefix, ""};
 }
 
+StatusOr<std::vector<std::string>> SplitArguments(std::string_view in) {
+  std::vector<std::string> arguments;
+  std::string current_string;
+
+  enum { NORMAL, QUOTED, SINGLE_QUOTED, ESCAPE } state = NORMAL;
```
PragmaTwice:
```suggestion
  enum { NORMAL, DOUBLE_QUOTED, SINGLE_QUOTED, ESCAPE } state = NORMAL;
```

src/common/string_util.cc
PragmaTwice: Ahh I got it. Maybe we can use `state_before_escape` to store the previous before ESCAPE and revert to it after ESCAPE.

src/common/string_util.cc
```diff
@@ -276,6 +276,71 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
...
+          state = c == '"' ? QUOTED : SINGLE_QUOTED;
+          is_quoted = c == '"';
+        } else {
+          current_string.push_back(c);
+        }
+        break;
+      case QUOTED:
+      case SINGLE_QUOTED:
+        if (c == '\\') {
+          state = ESCAPE;
```
PragmaTwice, line 305:
```suggestion
        if (c == '\\') {
          state_before_escape = state;
          state = ESCAPE;
```

mapleFU:
Where can we find the original redis code? What would `a"b` string like in this scenerio?

git-hulk:
> Where can we find the original redis code? What would `a"b` string like in this scenerio?

Could refer to: https://github.com/redis/redis/blob/96a0cfdea27fa28f93b8071ec1b3c6ef8b315298/src/sds.c#L990

src/common/string_util.cc
```diff
@@ -276,6 +276,71 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
   return {prefix, ""};
 }
 
+StatusOr<std::vector<std::string>> SplitArguments(std::string_view in) {
```
mapleFU, line 290: Can we add a comment that currently `\xHH` (hex) is not supportted?

src/common/string_util.cc
```diff
@@ -276,6 +276,71 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
...
+          current_string.push_back('\r');
+        } else if (c == 'n') {
+          current_string.push_back('\n');
+        } else if (c == 'v') {
+          current_string.push_back('\v');
+        } else if (c == 'f') {
+          current_string.push_back('\f');
+        } else if (c == 'b') {
+          current_string.push_back('\b');
+        }
```
mapleFU: The redis code is:

```c
switch(*p) {
                        case 'n': c = '\n'; break;
                        case 'r': c = '\r'; break;
                        case 't': c = '\t'; break;
                        case 'b': c = '\b'; break;
                        case 'a': c = '\a'; break;
                        default: c = *p; break;
                        }
```

Should we `push_back(c)` after checks above?

src/common/string_util.cc
```diff
@@ -276,6 +276,71 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
...
+  std::vector<std::string> arguments;
+  std::string current_string;
+
+  enum State { NORMAL, DOUBLE_QUOTED, SINGLE_QUOTED, ESCAPE } state = NORMAL;
+
+  State state_before_escape = NORMAL;
+  for (const char c : in) {
+    switch (state) {
+      case NORMAL:
+        if (std::isspace(c)) {
```
mapleFU, line 301:
```
switch(*p) {
                    case ' ':
                    case '\n':
                    case '\r':
                    case '\t':
                    case '\0':
```

should all of these being support?

src/common/string_util.cc
```diff
@@ -276,6 +276,71 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
...
+          current_string.push_back(c);
+        }
+        break;
+      case DOUBLE_QUOTED:
+      case SINGLE_QUOTED:
+        if (c == '\\') {
+          state_before_escape = state;
+          state = ESCAPE;
+        } else if ((c == '"' && state == DOUBLE_QUOTED) || (c == '\'' && state == SINGLE_QUOTED)) {
+          state = NORMAL;
```
mapleFU, line 337: should we checks that:

```
 /* closing quote must be followed by a space or
                         * nothing at all. */
```

src/common/string_util.cc
```diff
@@ -276,6 +276,71 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
...
+        } else {
+          current_string.push_back(c);
+        }
+        break;
+      case ESCAPE:
+        if (c == '"' || c == '\'' || c == '\\') {
+          current_string.push_back(c);
+        } else if (c == 't') {
```
mapleFU: I found that redis doesn't check these in single quote mode but I don't understand why 😅

mapleFU (commented):
I'm thinking that can we skip this function when string not contains any of `'` and `"`?

src/common/string_util.cc
PragmaTwice: Yeah, different behavior in `a "b"c d` with redis here.

src/common/string_util.cc
PragmaTwice: Maybe we should support this to let user put binary data into values.

src/common/string_util.cc
git-hulk: what's the meaning of this comment, Redis also emplaces the escaped char: https://github.com/redis/redis/blob/b33a405bf16cfa147f1736a6a3cd64c2fd2b3f7d/src/sds.c#L1030

PragmaTwice: I think the meaning is that, we need to consider cases like `\q`, where `q` is not a valid escape char.

PragmaTwice: Seems here we just ignore it.

mapleFU: I mean the final line:
```
 default: c = *p; break;
```

In our "ESCAPE" mode, this is absent?

git-hulk: Oh, I see. Good point. It's now ignored, and we should continue to behave the same way with Redis. I will fix them when I have time later.

src/common/string_util.cc
```diff
@@ -276,6 +276,107 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
...
+          state = NORMAL;
+        } else {
+          current_string.push_back(c);
+        }
+        break;
+      case DOUBLE_QUOTED:
+        if (c == '\\') {
+          state = ESCAPE;
+        } else if (c == '"') {
+          //
```
mapleFU: Remove this line?

src/common/string_util.cc
```diff
@@ -276,6 +276,107 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
...
 }
 
+static int hexDigitToInt(const char c) {
+  if (c >= '0' && c <= '9') {
+    return c - '0';
+  } else if (c >= 'a' && c <= 'f') {
+    return c - 'a' + 10;
+  } else if (c >= 'A' && c <= 'F') {
+    return c - 'A' + 10;
+  }
```
mapleFU, line 286: UNREACHABLE?

src/common/string_util.cc
```diff
@@ -276,6 +276,107 @@ std::pair<std::string, std::string> SplitGlob(std::string_view glob) {
...
+        } else if (c == '\'') {
+          state = SINGLE_QUOTED;
+        } else {
+          current_string.push_back(c);
+        }
+        break;
+      case SINGLE_QUOTED:
+        if (c == '\\' && (i + 1) < in.size() && in[i + 1] == '\'') {
+          current_string.push_back('\'');
+          i++;
```
mapleFU, line 319: ++i;

src/common/string_util.cc
git-hulk:
```suggestion
```

src/common/string_util.cc
git-hulk: Return 0 here to keep consistent with Redis on this.

src/common/string_util.cc
git-hulk: It should have the same effect here?

src/common/string_util.cc
git-hulk: return 0 to keep consistent with Redis behavior.

mapleFU (approved):

PR #2878, opened 2025-04-13
fix(tdigest): fix centroid encoding overwrite for same mean

LindaSummer:
## Issue

TDigest only cares about weight when merging, so it is possible to have centroids with same mean.
This will lead the centroids overwriting during serialization.

## Proposed Changes

Add a seq id to make centroids mean encoding keep same order as the original centroids list without confliction.

## Related Issue
#2849


mapleFU:
Sorry for delaying, I'm quite busy this week, will take a round tonight

src/types/redis_tdigest.cc
```diff
@@ -332,10 +332,13 @@ std::string TDigest::internalBufferKey(const std::string& ns_key, const TDigestM
...
 
 std::string TDigest::internalKeyFromCentroid(const std::string& ns_key, const TDigestMetadata& metadata,
-                                             const Centroid& centroid) const {
+                                             const Centroid& centroid, uint32_t seq) const {
   std::string sub_key;
   PutFixed8(&sub_key, static_cast<uint8_t>(SegmentType::kCentroids));
   PutDouble(&sub_key, centroid.mean);  // It uses EncodeDoubleToUInt64 and keeps original order of double
+  // The tdigest centroids only cares about the weight rather than the mean, so different centroids may have same mean,
+  // we should keep them with same original order, this seq id could be discarded in decode
+  PutFixed32(&sub_key, seq);
```
mapleFU, line 341: Is Fixed16 enough for this?

mapleFU (approved):

src/types/redis_tdigest.cc
LindaSummer: Hi @mapleFU ,

Thanks very much for your kind review!😊

I thought about the size when adding this seq-id.
The compression is uint32 now, so the limitation of centroids is uint32 maximum.

So I decided to use uint32 for safety.

Best Regards,
Edward

PragmaTwice (approved):

PR #2879, opened 2025-04-13
fix: typo in compaction checker

geonove:
I think there could be a small typo in the log message


PragmaTwice (approved):

compile_commands.json
```diff
@@ -0,0 +1 @@
+build/compile_commands.json
```
PragmaTwice: Ahh, we need to remove this file : )

PragmaTwice (changes requested):
See above.

compile_commands.json
geonove: Will do

geonove: Done!

PragmaTwice (approved):

PR #2881, opened 2025-04-14
refactor: replace glog with spdlog

### LINKED ISSUE #2267: Investigate the possiblity and method to replace GLOG with spdlog
### Search before asking

### Motivation

spdlog (https://github.com/gabime/spdlog) is a modern logging library based on fmt library (or `std::format`).

It's better to use spdlog instead of glog because we extensively utilize the fmt library.

Note: we need to specify `SPDLOG_FMT_EXTERNAL` since we need to keep there's only one fmt library in our deps.

### Solution

_No response_

PragmaTwice:
It closes #2267.

- A new header `common/logging.h` to provide log functions and also simulation interfaces of glog so that we don't need to replace all logging call `LOG(..)` in this PR;
- clang minimum requirement bumped form 8 to 9;
- support multiple destination in config option `log-dir`.


NOTICE
```diff
@@ -71,6 +71,7 @@ The text of each license is also included in licenses/LICENSE-[project].txt
 * hat-trie(https://github.com/Tessil/hat-trie)
 * pegtl(https://github.com/taocpp/PEGTL, NOTE: changed to Boost Software License Version 1.0 in main branch)
 * cpptrace(https://github.com/jeremy-rifkin/cpptrace)
+* spdlog(https://github.com/gabime/spdlog)
```
git-hulk, line 73: We should remove the `glog` from the NOTICE?

src/config/config.cc
```diff
@@ -227,7 +221,7 @@ Config::Config() {
       {"migrate-batch-rate-limit-mb", false, new IntField(&migrate_batch_rate_limit_mb, 16, 0, INT_MAX)},
       {"unixsocket", true, new StringField(&unixsocket, "")},
       {"unixsocketperm", true, new OctalField(&unixsocketperm, 0777, 1, INT_MAX)},
-      {"log-retention-days", false, new IntField(&log_retention_days, -1, -1, INT_MAX)},
+      {"log-retention-days", true, new IntField(&log_retention_days, -1, -1, INT_MAX)},
```
git-hulk, line 224: So, `spdlog` cannot change the retention days after initialization?

PragmaTwice: Currently yes.

NOTICE
PragmaTwice: Done.

git-hulk (approved):

PR #2883, opened 2025-04-15
chore(deps): bump rocksdb to v10.1.3

aleksraiden:
Bump rocksdb to v10.1.3 - https://github.com/facebook/rocksdb/releases/tag/v10.1.3

**Key changes**
- Fix a bug where resurrected full_history_ts_low from a previous session that enables UDT is used by this session that disables UDT
- Fix a bug where tail size of remote compaction output is not persisted in primary db's manifest
- Fixed stats for Tiered Storage with preclude_last_level feature
- format_version < 2 in BlockBasedTableOptions is no longer supported for writing new files. Support for reading such files is deprecated and might be removed in the future. CompressedSecondaryCacheOptions::compress_format_version == 1 is also deprecated
- Implemented API DB::GetPropertiesOfTablesByLevel that retrieves table properties for files in each LSM tree level
- Add a new field num_l0_files in CompactionJobInfo about the number of L0 files in the CF right before and after the compaction
- Added a new DBOptions.calculate_sst_write_lifetime_hint_set setting that allows to customize which compaction styles SST write lifetime hint calculation is allowed on. Today RocksDB supports only two modes kCompactionStyleLevel and kCompactionStyleUniversal


PragmaTwice:
Closed since #2912 is merged.

PR #2903, opened 2025-04-26
feat(logs): enable saving the slow logs into the file

greatsharp:
closed old pull request due to instead glog with spdlog


greatsharp:
Snipaste_2025-04-26_11-13-02

kvrocks.conf
```diff
@@ -498,6 +498,16 @@ slowlog-log-slower-than 100000
...
 
+# If you need to record all slow logs to logfile additionally, set to yes.
+# no means only record them in memory.
+# Default: no
+slowlog-save-to-logfile no
+
+# Record slow logs to logfiles with this level
+# Possible values: info, warning
+# Default: info
+slowlog-log-level info
```
PragmaTwice: TBH the log level seems not so useful.

src/stats/log_collector.cc
```diff
@@ -89,6 +101,35 @@ void LogCollector<T>::PushEntry(std::unique_ptr<T> &&entry) {
...
+      }
+      cmd.pop_back();
+    }
+    auto log = fmt::format("[slowlog] id: {}, timestamp: {}, duration: {}, cmd: {}, ip: {}, port: {}, client_name: {}",
+                           entry->id, entry->time, entry->duration, cmd, entry->ip, entry->port, entry->client_name);
+    if (log_level_ == spdlog::level::info) {
+      LOG(INFO) << log;
+    } else {
+      LOG(WARNING) << log;
+    }
```
PragmaTwice:
```suggestion
    info("[slowlog] id: {}, timestamp: {}, duration: {}, cmd: {}, ip: {}, port: {}, client_name: {}",
         entry->id, entry->time, entry->duration, cmd, entry->ip, entry->port, entry->client_name);
```

kvrocks.conf
```diff
@@ -498,6 +498,16 @@ slowlog-log-slower-than 100000
 # You can reclaim memory used by the slow log with SLOWLOG RESET.
 slowlog-max-len 128
 
+# If you need to record all slow logs to logfile additionally, set to yes.
+# no means only record them in memory.
+# Default: no
+slowlog-save-to-logfile no
```
```suggestion
slowlog-dump-to-logfile no
```

src/stats/log_collector.cc
```diff
@@ -89,6 +101,35 @@ void LogCollector<T>::PushEntry(std::unique_ptr<T> &&entry) {
...
 }
 
+template <>
+void LogCollector<SlowEntry>::PushEntry(std::unique_ptr<SlowEntry> &&entry) {
+  std::lock_guard<std::mutex> guard(mu_);
+  entry->id = ++id_;
+  entry->time = util::GetTimeStamp();
+  if (max_entries_ > 0 && !entries_.empty() && entries_.size() >= static_cast<size_t>(max_entries_)) {
+    entries_.pop_back();
+  }
```
PragmaTwice: Could you make these code reusable instead of just copying them?

greatsharp: I am thinking about how to dump both perf log and slowlog to log file, but can distinguish of them.

kvrocks.conf
greatsharp: > TBH the log level seems not so useful.
default log-level info will print so many logs, user many change it to warning in production env.
if user set defualt log-level config to warning, then slowlog couldn't be saved unless change slowlog-log-level to warning.

PragmaTwice: Hmm then I think we can just follow the global log-level, instead of creating a new log-level config option.

PragmaTwice: Or, at least we can merge these two option to one, e.g. `slowlog-dump-logfile-level off/info/warning`.

src/stats/log_collector.cc
greatsharp: I used "Template Specialization" at this place, i have no idea to avoid moving  entry parameter, and avoid to lock the mutex several times, so just copying this part to keep the code clean and clearly.

src/stats/log_collector.cc
```diff
@@ -89,6 +95,35 @@ void LogCollector<T>::PushEntry(std::unique_ptr<T> &&entry) {
...
+
+  if (dump_to_logfile_level_ == spdlog::level::info || dump_to_logfile_level_ == spdlog::level::warn) {
```
PragmaTwice:
```suggestion
  if (dump_to_logfile_level_ != spdlog::level::off) {
```

src/stats/log_collector.cc
git-hulk: I think we could remove the duplicate code like this commit: https://github.com/git-hulk/kvrocks/commit/dd0bdecebedf3280b463aa72f92f8f7f42955e89

greatsharp: > I think we could remove the duplicate code like this commit: [git-hulk@dd0bdec](https://github.com/git-hulk/kvrocks/commit/dd0bdecebedf3280b463aa72f92f8f7f42955e89)

interesting, check again

git-hulk (approved):

PragmaTwice (approved):

Issue #2889, opened 2025-04-20
Tracking Issues: replace logging calls from GLOG style to spdlog


PragmaTwice:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

In #2881, we have switched the logging library from GLOG to spdlog, but we didn't replace all logging calls (instead we have a mocking API of GLOG, which is translated to spdlog calls).

So there are lots of trivial tasks, to replace calls like `LOG(..) << ... << ..` to `info(..)`/`warn(..)`/`error(..)`. And here is a tracking issue for that.

- [ ] #2888 
- [ ] #2890
- [ ] #2894
- [ ] #2905 
- [ ] #2909
- [ ] #2936
- [ ] #2928
- [ ] #2892
- [ ] #2891

If there are so many calls in one source directory, we can also split them to more small tasks : )

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

XueSongTap:
Can I work on this ?

PragmaTwice:
> Can I work on this ?

Sure. Feel free to pick any of these tasks.

yezhizi:
It looks like `search/*` might not have any `LOG` command.

![Image](https://github.com/user-attachments/assets/a65caeaa-110a-46bb-a2ab-2d173cc8bb67)

yezhizi:
I noticed that some of the GLOG calls currently in use haven’t been translated to their spdlog equivalents yet. Do they need to be replaced? @PragmaTwice 
https://github.com/apache/kvrocks/blob/5f8e94bd39ec8f6552fc751fb2ae9325e185062b/src/common/logging.h#L117-L127

PragmaTwice:
> I noticed that some of the GLOG calls currently in use haven’t been translated to their spdlog equivalents yet. Do they need to be replaced? [@PragmaTwice](https://github.com/PragmaTwice)

GLOG has been removed in kvrocks. Currently `LOG`, `DLOG`, `CHECK`.. are all mocking APIs implemented with spdlog calls.

For `DLOG`, I think we can just replace it with `debug()` level logging calls, since from my side logging for only debug-mode build is not so useful (cc @mapleFU @git-hulk). (I actually dislike the idea of `DLOG` itself).

For `CHECK` and `DCHECK`, we can keep it as is for now.

PR #2909, opened 2025-04-29
chore(log): replace logging calls in storage/*

SharonIV0x86:
Part of #2889

### DIFF
--- a/src/storage/compaction_checker.cc
+++ b/src/storage/compaction_checker.cc
@@ -131,7 +131,7 @@ void CompactionChecker::PickCompactionFilesForCf(const engine::ColumnFamilyConfi
     }
   }
   if (best_delete_ratio > 0.1 && !best_start_key.empty() && !best_stop_key.empty()) {
-    info("[compaction checker] Going to compact the key in file: {}, delete ration: {}", best_filename,
+    info("[compaction checker] Going to compact the key in file: {}, delete ratio: {}", best_filename,
          best_delete_ratio);
     auto s = storage_->Compact(cf, &best_start_key, &best_stop_key);
     if (!s.ok()) {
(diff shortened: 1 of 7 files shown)


src/storage/event_listener.cc
```diff
@@ -163,22 +166,23 @@ void EventListener::OnBackgroundError(rocksdb::BackgroundErrorReason reason, roc
...
   }
 
-  LOG(ERROR) << fmt::format("[event_listener/background_error] reason: {}, bg_error: {}", reason_str, error_str);
+  error("[event_listener/background_error] reason: {}, bg_error: {}", reason_str, error_str);
 }
 
-void EventListener::OnTableFileDeleted(const rocksdb::TableFileDeletionInfo &info) {
-  LOG(INFO) << "[event_listener/table_file_deleted] db: " << info.db_name << ", sst file: " << info.file_path
-            << ", status: " << info.status.ToString();
+void EventListener::OnTableFileDeleted(const rocksdb::TableFileDeletionInfo &info_) {
```
SharonIV0x86: @PragmaTwice Hi, i had to rename the function parameter to ``info_`` for some functions as it was conflicting with the spdlog's ``info``.

SharonIV0x86:
@PragmaTwice 
The build and test for Darwin (Clang) is failing because there are no test files in the ``gocase/util`` folder. Since this folder just has utility code, maybe we could add a simple dummy test file to stop the CI from failing. Do you think that this would be a good idea?

src/storage/compaction_checker.cc
```diff
@@ -133,11 +131,11 @@ void CompactionChecker::PickCompactionFilesForCf(const engine::ColumnFamilyConfi
     }
   }
   if (best_delete_ratio > 0.1 && !best_start_key.empty() && !best_stop_key.empty()) {
-    LOG(INFO) << "[compaction checker] Going to compact the key in file: " << best_filename
-              << ", delete ratio: " << best_delete_ratio;
+    info("[compaction checker] Going to compact the key in file: {}, delete ration: {}", best_filename,
```
PragmaTwice:
```suggestion
    info("[compaction checker] Going to compact the key in file: {}, delete ratio: {}", best_filename,
```

PragmaTwice (approved):

PR #2937, opened 2025-05-07
fix(zset): wrong RESP reply in ZRANDMEMBER and ZMSCORE command

weim0000:
image
image

I want to add some test case for commandzmscore and commandzrandmember, but I cloud not find any folder or file to test commandXXX.

after fix:

image
image

### DIFF
--- a/src/commands/cmd_zset.cc
+++ b/src/commands/cmd_zset.cc
@@ -1109,9 +1109,9 @@ class CommandZMScore : public Commander {
       for (const auto &member : members) {
         auto iter = mscores.find(member.ToString());
         if (iter == mscores.end()) {
-          values.emplace_back(conn->NilString());
+          values.emplace_back("");
         } else {
-          values.emplace_back(conn->Double(iter->second));
+          values.emplace_back(util::Float2String(iter->second));
         }
       }
     }
@@ -1451,7 +1451,7 @@ class CommandZRandMember : public Commander {
     }
 
     if (no_parameters_)
-      *output = s.IsNotFound() ? conn->NilString() : redis::BulkString(result_entries[0]);
+      *output = s.IsNotFound() ? conn->NilString() : result_entries[0];
     else
       *output = Array(result_entries);
     return Status::OK();
--- a/src/commands/cmd_zset.cc
+++ b/src/commands/cmd_zset.cc
@@ -1109,13 +1109,13 @@ class CommandZMScore : public Commander {
       for (const auto &member : members) {
         auto iter = mscores.find(member.ToString());
         if (iter == mscores.end()) {
-          values.emplace_back("");
+          values.emplace_back(conn->NilString());
         } else {
-          values.emplace_back(util::Float2String(iter->second));
+          values.emplace_back(conn->Double(iter->second));
         }
       }
     }
-    *output = conn->MultiBulkString(values);
+    *output = Array(values);
     return Status::OK();
   }
 };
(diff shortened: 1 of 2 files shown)


src/commands/cmd_zset.cc
```diff
@@ -1109,9 +1109,9 @@ class CommandZMScore : public Commander {
       for (const auto &member : members) {
         auto iter = mscores.find(member.ToString());
         if (iter == mscores.end()) {
-          values.emplace_back(conn->NilString());
+          values.emplace_back("");
         } else {
-          values.emplace_back(conn->Double(iter->second));
+          values.emplace_back(util::Float2String(iter->second));
```
PragmaTwice: We can replace `conn->MultiBulkString(values);` in 1118th line to `Array(values)` instead of changing here.

weim0000: I think it is better than Array(values), because conn->MultiBulkString(values) converge protocol mean；and values only expression values；  If use Array(values), then values include some protocol meaning，it is incomplete

git-hulk: That's correct to use `redis::Array` instead of `MultiBulkString` because it needs to return the double type in RESP3.

git-hulk:
> I want to add some test case for commandzmscore and commandzrandmember, but I cloud not find any folder or file to test commandXXX.

@weim0000 You could add test cases in [zset_test.go](https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/type/zset/zset_test.go)

src/commands/cmd_zset.cc
weim0000: OK，get it，I will update

weim0000:
> > I want to add some test case for commandzmscore and commandzrandmember, but I cloud not find any folder or file to test commandXXX.
> 
> @weim0000 You could add test cases in [zset_test.go](https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/type/zset/zset_test.go)

sorry, [zset_test.go] test redis_zset.cc, but these bugs is in cmd_zset.cc,  there is no xx_test.go for cmd_xx.cc

PragmaTwice:
> sorry, [zset_test.go] test redis_zset.cc, but these bugs is in cmd_zset.cc, there is no xx_test.go for cmd_xx.cc

In golang test cases, we start a kvrocks instance and connect it via go-redis client library. So that these RESP replies can be tested well. No need to add a test file for cmd_xx.cc.

tests/gocase/unit/type/zset/zset_test.go
```diff
@@ -1308,6 +1324,10 @@ func basicTests(t *testing.T, rdb *redis.Client, ctx context.Context, enabledRES
 		}
 		createZset(rdb, ctx, "zset", z)
 
+		// ZRANDMEMBER zset
+		str := rdb.Do(ctx, "ZRANDMEMBER", "zset").Val()
```
git-hulk, line 1328: Can use `rdb.ZRandMember`.

PragmaTwice: Seems not a major issue, maybe we can merge it first.

PragmaTwice (approved):

tests/gocase/unit/type/zset/zset_test.go
weim0000: Sorry, it can't use rdb.ZRandMember, because this function have a `count` param,
ZRandMember(ctx context.Context, key string, count int);
and  this bug must not input `count` param;

weim0000: image

git-hulk: I see, thank you!

PR #2943, opened 2025-05-07
feat(cmds): Add support for CLIENT REPLY subcommand and related tests

DCjanus:
This PR implements the `CLIENT REPLY` subcommand, allowing clients to control server reply behavior with `ON`, `OFF`, and `SKIP` modes. 

This change is part of [issue #2422](https://github.com/apache/kvrocks/issues/2422).

### Note

I noticed that there was a previous attempt to address this in [PR #2495](https://github.com/apache/kvrocks/pull/2495), but it seems to have been inactive for a while. I apologize for opening a new PR, but I hope this contribution can help move the feature forward.


src/commands/cmd_server.cc
```diff
@@ -497,15 +508,26 @@ class CommandClient : public Commander {
...
 
-    return {Status::RedisInvalidCmd, "Syntax error, try CLIENT LIST|INFO|KILL ip:port|GETNAME|SETNAME"};
+    return {Status::RedisInvalidCmd, "Syntax error, try CLIENT LIST|INFO|KILL ip:port|GETNAME|SETNAME|REPLY"};
   }
 
  private:
   std::string addr_;
   std::string conn_name_;
   std::string subcommand_;
+  std::string reply_mode_arg_;
```
PragmaTwice: maybe we can parse it in `Parse()` and store a `ReplyMode reply_mode_` here?

src/commands/cmd_server.cc
```diff
@@ -497,15 +508,26 @@ class CommandClient : public Commander {
...
+    } else if (subcommand_ == "reply") {
+      if (reply_mode_arg_ == "on") {
+        conn->SetReplyMode(Connection::ReplyMode::ON);
+      } else if (reply_mode_arg_ == "off") {
+        conn->SetReplyMode(Connection::ReplyMode::OFF);
+      } else if (reply_mode_arg_ == "skip") {
+        conn->SetReplyMode(Connection::ReplyMode::SKIP_ONCE_PENDING);
+      }
+      *output = redis::RESP_OK;
+      return Status::OK();
```
PragmaTwice: Currently we have 4 states in the `ReplyMode` enum, and the `SKIP_ONCE_PENDING` seems to ignore the reply of `CLIENT REPLY` itself. But if we just don't reply in `CLIENT REPLY OFF|SKIP`, maybe we can remove the 3rd state. e.g.
```c++
conn->SetReplyMode(reply_mode_); // reply_mode_ can be ON, OFF or SKIP (3 states).

if (reply_mode_ != Connection::ReplyMode::SKIP) {
  *output = redis::RESP_OK;
}
return Status::OK();
```

WDYT?

DCjanus: Thank you for your suggestion.

> Currently we have 4 states in the `ReplyMode` enum, and the `SKIP_ONCE_PENDING` seems to ignore the reply of `CLIENT REPLY` itself. But if we just don't reply in `CLIENT REPLY OFF|SKIP`, maybe we can remove the 3rd state. e.g.
> 
> ```c++
> conn->SetReplyMode(reply_mode_); // reply_mode_ can be ON, OFF or SKIP (3 states).
> 
> if (reply_mode_ != Connection::ReplyMode::SKIP) {
>   *output = redis::RESP_OK;
> }
> return Status::OK();
> ```
> 
> WDYT?

Thank you for your suggestion.

The reason I didn’t write it this way at first is that I wasn’t very familiar with the codebase. I didn’t realize that leaving `output` empty (not setting a reply) is actually a supported and safe behavior in this scenario.

I was initially concerned that not setting `output` might cause unexpected issues. Later, I found that the `Reply` function is only called when `output` is not empty, so leaving it empty is safe and works as intended here.

Changes has been made at [this commit](https://github.com/apache/kvrocks/pull/2943/commits/01f38e5c10af4ad9bd9e0a372b81390d42d69eee)

src/commands/cmd_server.cc
DCjanus: > maybe we can parse it in `Parse()` and store a `ReplyMode reply_mode_` here?

Changes has been made in [this commit](https://github.com/apache/kvrocks/pull/2943/commits/ca8f0afae0e6e365b06bc8ec30cda5dbb7e292ab)

PragmaTwice (approved):

PragmaTwice:

PR #2985, opened 2025-05-21
fix(list): dst-key's version is 0 in rpoplpush cmd for non-existent dst-key

weim0000:
fix(list): dst-key's version is 0 in rpoplpush cmd for non-existent dst-key

image


git-hulk:
@weim0000 Good catch!

PragmaTwice:
@weim0000 For adding golang test case, you can use command `KMETADATA` to retrieve the version in metadata.

weim0000:
> @weim0000 For adding golang test case, you can use command `KMETADATA` to retrieve the version in metadata.

done

PragmaTwice (approved):

PR #2989, opened 2025-05-22
chore(config): mark the configuration `rocksdb.row_cache_size` as deprecated

### LINKED ISSUE #2987: Unable to migrate slots if rocksdb.row_cache_size is enabled
### Search before asking

### Version

Tried in two environments, and was able to see the migration fail on both

## Environment 1
Amazon Linux release 2023.7.20250512
kvrocks version 2.12.1

## Environment 2 (on my local mac)
Docker with: FROM apache/kvrocks:latest
 - org.opencontainers.image.created: 2025-03-11T02:12:09.877Z
 - org.opencontainers.image.revision: b8db0e75553ccf6e02cde66c382d86a7d7ec5570 

### Minimal reproduce step

If we launch with the `working` configuration, and then follow the common steps, migration will work. 
However, if we follow the `not working` configuration, and then follow the common steps, migration will not work

### Working
Launch 2 kvrocks nodes with `--cluster-enabled yes`

### Not Working
Launch 2 kvrocks nodes with `--cluster-enabled yes --rocksdb.row_cache_size 1000`

### Common Steps: 
With my setup, I have two nodes on port 6000, and 6003

1. Start up a kvrocks controller
2. Create a namespace 
```
./_build/kvctl create namespace test-ns
```
3. Create a cluster of just 1 node with the first node (i've added kvrocks0 to my /etc/hosts to point to 127.0.0.1)
```
./_build/kvctl create cluster test-cluster0 --nodes kvrocks0:6000 -n test-ns
```
4. Create a new shard with the 2nd node
```
./_build/kvctl create shard -n test-ns -c test-cluster0 --nodes kvrocks3:6003
```
5. migrate slots
```
./_build/kvctl migrate slot 10922-16383 --target 1 -n test-ns -c test-cluster0
```

```
controller/cluster.go:360","msg":"Migrate the slot successfully","namespace":"test-ns","cluster":"test-cluster0","slot":"10922-16383"}
```

### With --rocksdb.row_cache_size 1000:
On the kvrocks-controller I'll see: 
```
"caller":"controller/cluster.go:349","msg":"Failed to migrate the slot","namespace":"test-ns","cluster":"test-cluster0","slot":"10922-16383"}
```

And on the kvrocks node I see: 
```
E20250521 22:39:02.587806 270160076237728 slot_migrate.cc:195] [migrate] Failed to start migrating slot(s) 10922-16383. Error: failed to set import status on destination node: failed to check the response from the destination node: got invalid response of length
96: -ERR clear keys of slot(s) error: Not implemented: DeleteRange is not compatible with row cache.                                                                                                                                                                  
E20250521 22:39:02.588477 270160076237728 slot_migrate.cc:245] [migrate] Failed to finish a failed migration of slot(s) 10922-16383. Error: failed to set import status on destination node: failed to check the response from the destination node: got invalid      
response of length 63: -ERR mismatch slot, importing slot(s): -1, but got: 10922-16383                                                                                                                                                                                
```

To me, it looks like this is the reason: `Not implemented: DeleteRange is not compatible with row cache.`

### Anything Else?

I think if migrating slots is just not supported with row cache enabled, maybe the readme for the config should mention that. 

git-hulk:
This closes #2987.

In most scenarios, the row cache does not significantly improve performance.
It now conflicts with the DeleteRange API, which is a critical feature for
fast deletion. So, we decided to remove the row cache configuration to avoid this situation.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -819,7 +819,7 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
     auto s = field->Set(input.second);
     if (!s.IsOK()) return s.Prefixed(fmt::format("failed to set value of field '{}'", field_key));
   } else if (deprecated_fields_.find(field_key) != deprecated_fields_.end()) {
-    std::cout << fmt::format("The configuration '{}' at line {} is deprecated.", field_key, line_number) << std::endl;
+    std::cout << fmt::format("WARNING: '{}' at line {} is deprecated and does not take effect.", field_key, line_number) << std::endl;
   } else {
     std::cout << fmt::format("WARNING: '{}' at line {} is not a valid configuration key.", field_key, line_number)
               << std::endl;
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -819,7 +819,8 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
     auto s = field->Set(input.second);
     if (!s.IsOK()) return s.Prefixed(fmt::format("failed to set value of field '{}'", field_key));
   } else if (deprecated_fields_.find(field_key) != deprecated_fields_.end()) {
-    std::cout << fmt::format("WARNING: '{}' at line {} is deprecated and does not take effect.", field_key, line_number) << std::endl;
+    std::cout << fmt::format("WARNING: '{}' at line {} is deprecated and does not take effect.", field_key, line_number)
+              << std::endl;
   } else {
     std::cout << fmt::format("WARNING: '{}' at line {} is not a valid configuration key.", field_key, line_number)
               << std::endl;
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -131,6 +131,8 @@ Status SetRocksdbCompression(Server *srv, const rocksdb::CompressionType compres
 };
 
 Config::Config() {
+  deprecated_fields_ = {"rocksdb.row_cache_size"};
+
   struct FieldWrapper {
     std::string name;
     bool readonly;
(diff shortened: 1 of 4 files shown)


src/config/config.cc
```diff
@@ -817,6 +818,8 @@ Status Config::parseConfigFromPair(const std::pair<std::string, std::string> &in
     field->line_number = line_number;
     auto s = field->Set(input.second);
     if (!s.IsOK()) return s.Prefixed(fmt::format("failed to set value of field '{}'", field_key));
+  } else if (deprecated_fields_.find(field_key) != deprecated_fields_.end()) {
+    std::cout << fmt::format("The configuration '{}' at line {} is deprecated.", field_key, line_number) << std::endl;
```
git-hulk: Users will be confused if we directly remove the configuration and print a warning message with an invalid key. So we add a dedicated set to log deprecated fields here.

```suggestion
    std::cout << fmt::format("WARNING: '{}' at line {} is deprecated and does not take effect.", field_key, line_number) << std::endl;
```

aleksraiden (approved):

PragmaTwice (approved):

torwig (approved):

ltagliamonte-dd:
Hello Folks,
I just updated my kvrocks to latest unstable and i noticed the warm message:
```
WARNING: 'rocksdb.row_cache_size' at line -1 is deprecated and does not take effect.
```

Our workload is heavy reads and we do use row_cache_size, it helps immensely if your workload is read-heavy and i bet most of the real usage application for kvrocks/redis are read heavy.
Can we find a way to keep the config and maybe find another solution?

My understanding is that is not compatible with DeleteRange that is used for replication, maybe can we enable this feature when replication is off? (no cluster mode)?

git-hulk:
@ltagliamonte-dd Did you test if it works well without the row cache? In our previous tests, the row cache didn't bring a significant performance benefit.

ltagliamonte-dd:
> @ltagliamonte-dd Did you test if it works well without the row cache? In our previous tests, the row cache didn't bring a significant performance benefit.

Based on prior tuning, we did we saw important latency improvements when we enabled row_cache. as matter of fact we dedicate 1GB to it on each kvrocks instance.
Since we don't report the row_cache hit data in the INFO output, I'm going to perform a multi day test in our production env, starting from next Monday and I will report back.
@git-hulk If the test clearly show that row_cache helps with performances what would it be the compromise to re-enable it?

PragmaTwice:
Yah if you have a reasonable benchmark data we can revert the change.

PR #2996, opened 2025-05-28
fix(txn): wrong RESP reply for EXEC if error occurred in commit

### LINKED ISSUE #2992: MULTI/EXEC commit failures improperly responded to
### Search before asking

### Version

repros on most unstable trunks

### Minimal reproduce step

Enable 
```
rocksdb.write_options.no_slowdown no
```
in the kvrocks config, and open a server. Send it lots of write load, until a simple `SET` fails with `Low priority write stall`.

Then send
```
MULTI
SET a b
EXEC
```

```
*1
-EXEC Low priority write stall
```

### What did you see instead?

```
*1
+OK
-EXEC Low priority write stall
```

### Anything Else?

Fundamentally, this happens because `Connection::ExecuteCommands` eagerly responds `+OK` to successful mutation commands before the WriteBatch is committed: my current workaround is to defer calling `Connection::Reply` on these replies until after the transaction is successfully committed.

If the commit fails, I arbitrarily change one of the responses to an error. However, this is hacky and doesn't correctly attribute the failures to the commands that attempted a write.

PragmaTwice:
It closes #2992.


PragmaTwice:
cc @nathanlo-hrt To see if it can fix your problem?

nathanlo-hrt:
this works for me, thanks!

git-hulk (approved):

PR #3001, opened 2025-05-30
build(jemalloc): enable the profiling option when compiling the jemalloc

### LINKED ISSUE #3000: Enabling the option `–enable-prof` at compile time

git-hulk:
This closes #3000.

This only enables profiling at the compile stage, but since we set `prof_active` to false, the profiling is not active, so it won't incur any overhead for now. We can enable the profiling by setting the environment MALLOC_CONF like:

```shell
MALLOC_CONF=prof:true,prof_active:true,lg_prof_interval:30,lg_prof_sample:21,prof_prefix:/tmp/jeprof ./kvrocks
```

And it will produce the jemalloc under the `/tmp` directory like the following:

```shell
jeprof.2359.0.i0.heap
jeprof.2359.1.i1.heap  
jeprof.2359.3.i3.heap
```

### DIFF
--- a/cmake/jemalloc.cmake
+++ b/cmake/jemalloc.cmake
@@ -25,7 +25,9 @@ endif()
 
 if (NOT DISABLE_JEMALLOC_PROFILING)
   set(ENABLE_JEMALLOC_PROFILING "--enable-prof")
-  set (JEMALLOC_CONFIG_MALLOC_CONF "prof:true,prof_active:false")
+  # jemalloc profiling is enabled by default, but we can disable it and
+  # enable it at runtime by setting the environment variable.
+  set(JEMALLOC_CONFIG_MALLOC_CONF "prof:true,prof_active:false")
 else()
   set(ENABLE_JEMALLOC_PROFILING "")
 endif()
--- a/cmake/jemalloc.cmake
+++ b/cmake/jemalloc.cmake
@@ -25,9 +25,6 @@ endif()
 
 if (NOT DISABLE_JEMALLOC_PROFILING)
   set(ENABLE_JEMALLOC_PROFILING "--enable-prof")
-  # jemalloc profiling is enabled by default, but we can disable it and
-  # enable it at runtime by setting the environment variable.
-  set(JEMALLOC_CONFIG_MALLOC_CONF "prof:true,prof_active:false")
 else()
   set(ENABLE_JEMALLOC_PROFILING "")
 endif()
--- a/cmake/jemalloc.cmake
+++ b/cmake/jemalloc.cmake
@@ -23,6 +23,13 @@ else()
   set(DISABLE_CACHE_OBLIVIOUS "--disable-cache-oblivious")
 endif()
 
+if (NOT DISABLE_JEMALLOC_PROFILING)
+  set(ENABLE_JEMALLOC_PROFILING "--enable-prof")
+  set (JEMALLOC_CONFIG_MALLOC_CONF "prof:true,prof_active:false")
+else()
+  set(ENABLE_JEMALLOC_PROFILING "")
+endif()
+
 include(cmake/utils.cmake)
 
 FetchContent_DeclareGitHubWithMirror(jemalloc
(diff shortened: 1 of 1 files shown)


cmake/jemalloc.cmake
```diff
@@ -23,6 +23,13 @@ else()
   set(DISABLE_CACHE_OBLIVIOUS "--disable-cache-oblivious")
 endif()
 
+if (NOT DISABLE_JEMALLOC_PROFILING)
```
PragmaTwice, line 26:
```suggestion
if (ENABLE_JEMALLOC_PROFILING)
```

git-hulk: I would like to enable it at compile stage by default.

cmake/jemalloc.cmake
```diff
@@ -23,6 +23,15 @@ else()
...
+  set(ENABLE_JEMALLOC_PROFILING "--enable-prof")
+  # jemalloc profiling is enabled by default, but we can disable it and
+  # enable it at runtime by setting the environment variable.
+  set(JEMALLOC_CONFIG_MALLOC_CONF "prof:true,prof_active:false")
```
PragmaTwice: this is a CMake variable and will not be passed to the build phase of jemalloc.

git-hulk: Oh? I see Jemalloc's configure.ac recognizes this environment, so I would like to test if CMake could pass this. 
https://github.com/jemalloc/jemalloc/blob/edaab8b3ad752a845019985062689551cd6315c1/configure.ac#L1278

PragmaTwice: Seems it can also be configured in runtime via `MALLOC_CONF` env var, so we can ignore it?

git-hulk: This line mainly uses to set prof_active to false by default, so that it won’t collect the profiling unless user sets the MALLOC_CONF to enable it.

git-hulk: I have tested on my side, the heap profiling won't run if we didn't enable `opt.prof=true` via MALLOC_CONF, so we can remove this line.

PragmaTwice:
Is there any performance impact of this flag? e.g. through redis-benchmark.

git-hulk:
> Is there any performance impact of this flag? e.g. through redis-benchmark.

In summary, there's no impact on performance. Many other projects are also enabled when compiling, like ClickHouse and MySQL(Percona)，

The first one is running without enable the `--enable-prof` when compiling.

```
hulk@instance-20241021-041032:~$ redis-benchmark -c -p 6666 -n 1000000 -r 1000000 -t set -d 128
====== 6666 -n 1000000 -r 1000000 -t set -d 128 ======
  100000 requests completed in 4.52 seconds
  0 parallel clients
  3 bytes payload
  keep alive: 1
  host configuration "save": 900 1 300 10 60 10000
  host configuration "appendonly": no
  multi-thread: no

99.62% <= 0.1 milliseconds
99.95% <= 0.2 milliseconds
100.00% <= 0.3 milliseconds
100.00% <= 0.7 milliseconds
100.00% <= 1.0 milliseconds
22114.11 requests per second
```

The second one is running with `--enable-prof`, but `prof_active` is false.
```
hulk@instance-20241021-041032:~$ redis-benchmark -c -p 6666 -n 1000000 -r 1000000 -t set -d 128
====== 6666 -n 1000000 -r 1000000 -t set -d 128 ======
  100000 requests completed in 4.52 seconds
  0 parallel clients
  3 bytes payload
  keep alive: 1
  host configuration "save": 900 1 300 10 60 10000
  host configuration "appendonly": no
  multi-thread: no

99.62% <= 0.1 milliseconds
99.95% <= 0.2 milliseconds
99.99% <= 0.3 milliseconds
100.00% <= 0.4 milliseconds
100.00% <= 0.5 milliseconds
100.00% <= 0.6 milliseconds
100.00% <= 1.0 milliseconds
22128.79 requests per second
```

The last one is running with `--enable-prof` and `prof_active` is true.

```
hulk@instance-20241021-041032:~$ redis-benchmark -c -p 6666 -n 1000000 -r 1000000 -t set -d 128
====== 6666 -n 1000000 -r 1000000 -t set -d 128 ======
  100000 requests completed in 4.54 seconds
  0 parallel clients
  3 bytes payload
  keep alive: 1
  host configuration "save": 900 1 300 10 60 10000
  host configuration "appendonly": no
  multi-thread: no

99.51% <= 0.1 milliseconds
99.94% <= 0.2 milliseconds
99.99% <= 0.3 milliseconds
100.00% <= 0.4 milliseconds
100.00% <= 0.4 milliseconds
22011.88 requests per second
```

PragmaTwice (approved):

PR #3007, opened 2025-06-02
feat(command): add a new command to enable/disable/dump the memory profiling

### LINKED ISSUE #3006: Add a new command to allow enabling/disabling/dumping the memory heap

git-hulk:
This closes #3006.

Currently, it's hard to identify where's occupying the memory when it goes high.
To mitigate this issue, we would like to introduce a profiling command to make the
process easier.

To enable this feature, you need to set the `MALLOC_CONF` environment variable before running the Kvrocks server, like the following instruction:

```shell
$ export MALLOC_CONF="prof:true,background_thread:true"
$ ./kvrocks -c kvrocks.conf
```

or 

```shell
MALLOC_CONF="prof:true,background_thread:true" ./kvrocks -c kvrocks.conf
```

You could also enable the profiling feature without activating the profiling behavior to avoid the overhead:

```
MALLOC_CONF="prof:true,prof_active:false,background_thread:true" ./kvrocks -c kvrocks.conf
```

You can then enable it at runtime if necessary:

```shell
## Enable the memory profiling, set prof_active to true.
$ redis-cli> KPROFILE MEMORY ENABLE

## Inspect the profiling status
$ redis-cli> KPROFILE MEMORY status

## Dump the profiling file into /tmp directory, which named with `jeprof_` prefix.
$ redis-cli> KPROFILE MEMORY DUMP /tmp

## Disable the memory profiling, set prof_active to false.
$ redis-cli> KPROFILE MEMORY DISABLE
```


src/commands/cmd_server.cc
```diff
@@ -309,6 +309,44 @@ class CommandDBSize : public Commander {
   }
 };
 
+class CommandProfiling : public Commander {
```
PragmaTwice: Better to be `KPROFILE`?

src/commands/cmd_server.cc
```diff
@@ -309,6 +309,44 @@ class CommandDBSize : public Commander {
...
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    CommandParser parser(args, 1);
+    if (!parser.EatEqICase("memory")) {
+      return {Status::NotOK, "only supports MEMORY subcommand"};
+    }
+    if (parser.EatEqICase("enabled")) {
+      enabled_ = true;
+    } else if (parser.EatEqICase("disabled")) {
+      enabled_ = false;
```
```suggestion
    if (parser.EatEqICase("enable")) {
      enabled_ = true;
    } else if (parser.EatEqICase("disable")) {
      enabled_ = false;
```

action instead of status?

src/server/memory_profiler.cc
```diff
@@ -0,0 +1,85 @@
...
+
+Status checkIfProfilingEnabled() {
+  bool enabled = false;
+  size_t value_size = sizeof(enabled);
+  if (mallctl("opt.prof", &enabled, &enabled_size, nullptr, 0) != 0) {
+    return {Status::NotOK, fmt::format("unable to check if profiling is enabled: {}", strerror(errno))};
+  }
+  if (!enabled) {
+    return {Status::NotOK, "jemalloc profiling isn't enabled, please run Kvrocks with following enviroments: `{}`",
+            "export MALLOC_CONF=\"prof:true,background_thread:true\""};
```
PragmaTwice: Could it be enabled in runtime by mallctl?

git-hulk: Both of them are read-only at runtime, see https://jemalloc.net/jemalloc.3.html

src/commands/cmd_server.cc
git-hulk: OK, it's good to add `K` for all customized commands.

src/commands/cmd_server.cc
git-hulk: It's expected to be a status instead of an action, perhaps we could rename it to `ON|OFF` to make it clearer?

PragmaTwice: nope. we changed the current status via `profiling memory enabled`, and this is called an "action".

PragmaTwice: if we use "enabled" here, this is a status and we should not change it.

note that "enable" means "make it enabled".

git-hulk: Makes sense. I was initially expecting this to be a set status because I think the verb should be the first word(enable memory profiling).

src/commands/cmd_server.cc
```diff
@@ -309,6 +309,54 @@ class CommandDBSize : public Commander {
...
+    }
+    if (parser.EatEqICase("enable")) {
+      enabled_ = true;
+    } else if (parser.EatEqICase("disable")) {
+      enabled_ = false;
+    } else if (parser.EatEqICase("dump")) {
+      if (!parser.Good()) return {Status::NotOK, errWrongNumOfArguments};
+      dump_dir_ = GET_OR_RET(parser.TakeStr());
+    } else {
+      return {Status::NotOK, "MEMORY subcommand must be one of ENABLED, DISABLED, DUMP"};
```
```suggestion
      return {Status::NotOK, "MEMORY subcommand must be one of ENABLE, DISABLE, DUMP"};
```

src/server/memory_profiler.h
```diff
@@ -0,0 +1,33 @@
...
+
+class MemoryProfiler {
+ public:
+  MemoryProfiler() = default;
+  ~MemoryProfiler() = default;
+
+  static std::string AllocatorName();
+  static Status SetProfiling(bool enabled);
+  static Status Dump(std::string_view dir);
+};
```
PragmaTwice:
```suggestion
struct MemoryProfiler {
  static std::string AllocatorName();
  static Status SetProfiling(bool enabled);
  static Status Dump(std::string_view dir);
};
```

src/commands/cmd_server.cc
```diff
@@ -1450,6 +1498,7 @@ REDIS_REGISTER_COMMANDS(Server, MakeCmdAttr<CommandAuth>("auth", 2, "read-only o
                         MakeCmdAttr<CommandFlushAll>("flushall", 1, "write no-dbsize-check exclusive admin", NO_KEY),
                         MakeCmdAttr<CommandDBSize>("dbsize", -1, "read-only", NO_KEY),
                         MakeCmdAttr<CommandSlowlog>("slowlog", -2, "read-only", NO_KEY),
+                        MakeCmdAttr<CommandKProfile>("kprofile", -3, "read-only", NO_KEY),
```
```suggestion
                        MakeCmdAttr<CommandKProfile>("kprofile", -3, "read-only admin", NO_KEY),
```

src/commands/cmd_server.cc
```diff
@@ -309,6 +309,54 @@ class CommandDBSize : public Commander {
...
+    if (parser.EatEqICase("enable")) {
+      enabled_ = true;
+    } else if (parser.EatEqICase("disable")) {
+      enabled_ = false;
+    } else if (parser.EatEqICase("dump")) {
+      if (!parser.Good()) return {Status::NotOK, errWrongNumOfArguments};
+      dump_dir_ = GET_OR_RET(parser.TakeStr());
+    } else {
+      return {Status::NotOK, "MEMORY subcommand must be one of ENABLE, DISABLE, DUMP"};
+    }
```
PragmaTwice: Should we introduce a subcommand to check the current status? e.g. `kprofile memory enabled` -> true/false.

git-hulk: It's nice to have. How about `kprofile memory status` -> `enabled|disabled`?

PragmaTwice: looks good

PragmaTwice: maybe we could distinguish between:
- allocator: libc? jemalloc?
- profiling available?
- enabled?disabled?

git-hulk: The allocator name is now showing at the `info memory` section, and using the profiling command on an unsupported allocator will return an error for now. So I think it should be good to just return `enabled` or `disabled`.

PragmaTwice (approved):

PR #3015, opened 2025-06-04
fix(storage): don't try to commit empty write batches

nathanlo-hrt:
Currently `CommandExec` will try and call `CommitTxn` even if the write batch is empty. Also, RocksDB will reject write batch commits on read-only DBs even if they're empty.

This means that MULTI/EXEC transactions silently fail:
```
MULTI 
+OK
GET a
+QUEUED
EXEC
*1
-ERR Not implemented: Not supported operation in read only mode.
```

With my change, this will succeed on read-only DBs.
```
MULTI
+OK
GET a
+QUEUED
EXEC
*1
$1
b
```
and the usual read-only permission checking will still work properly.
```
MULTI
+OK
SET a b
+QUEUED
EXEC
*1
-ERR Not implemented: Not supported operation in read only mode.
```

### DIFF
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -700,6 +700,11 @@ rocksdb::Status Storage::Write(engine::Context &ctx, const rocksdb::WriteOptions
 
 rocksdb::Status Storage::writeToDB(engine::Context &ctx, const rocksdb::WriteOptions &options,
                                    rocksdb::WriteBatch *updates) {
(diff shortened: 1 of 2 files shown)


git-hulk (dismissed):

git-hulk:
@nathanlo-hrt Would you mind adding a test case for this?

PragmaTwice (approved):

git-hulk (approved):

PR #3018, opened 2025-06-06
fix: add missing CC/CXX compiler info to luajit.cmake

raffertyyu:
fix link error when compiling kvrocks on linux with clang.
Because of missing compiler info, luajit is compiled by default c compiler , which is gcc on most linux platform, leading to a link error.


PragmaTwice (approved):
Looks good to me.

But IMO it should not lead to an error to link a gcc-produced object and a clang-produced object.

PR #3020, opened 2025-06-06
fix: Unset change_level if level_compaction_dynamic_level_bytes is enabled

### LINKED ISSUE #2601: rocksdb::CompactRangeOptions.change_level is not compatible with level_compaction_dynamic_level_bytes
### Search before asking

### Version

6b23c5ac2cf17558d167a3305c8f10ecb944637f

### Minimal reproduce step

Default configuration but with `stats_dump_period_sec` set (so you can see compaction stats) , `redis-cli set a b`, `redis-cli compact`.

Because of how `level_compaction_dynamic_level_bytes` works, there should be only 1 compaction to L6 (the default bottom-most layer)

### What did you see instead?

1 compaction on each layer and 2 compactions to L6

### Anything Else?

I instrumented rocksdb code and this is because `rocksdb::CompactRangeOptions.change_level` is not compatible with `level_compaction_dynamic_level_bytes`.

When `level_compaction_dynamic_level_bytes` is set, `set a b` and `compact` will not compact to L1, it'll jump directly to the last layer (L6). And when `change_level` is set, it then moves this SST file from L6 back to L1.

Now when another periodic compaction comes along, it tries to move this file from L1->L2, L2->L3, ... This is why every layer has 1 compaction and the last layer has 2 compactions.

If you run a manual compaction at this time, since `change_level` is set, it'll move the SST file to L1 again, resulting in the same churn happening.

Obviously this is quite efficiently since these 2 options are trying to undo each other's work..

hotpxl:
See https://github.com/facebook/rocksdb/issues/13671 for more details, but TLDR is when both `change_level` and `level_compaction_dynamic_level_bytes` are set, RocksDB will try to thrash the data between L1 and L6. Under `level_compaction_dynamic_level_bytes`, the intention is to fill up the "base" level first to reduce storage amplification, so we should not set `change_level`

Fixes #2601

### DIFF
--- a/src/storage/compaction_checker.cc
+++ b/src/storage/compaction_checker.cc
@@ -27,7 +27,9 @@
 
 void CompactionChecker::CompactPropagateAndPubSubFiles() {
   rocksdb::CompactRangeOptions compact_opts;
-  compact_opts.change_level = true;
+  // See https://github.com/facebook/rocksdb/issues/13671
+  // change_level doesn't work well with level_compaction_dynamic_level_bytes
+  compact_opts.change_level = !storage_->GetConfig()->rocks_db.level_compaction_dynamic_level_bytes;
   for (const auto &cf :
        {engine::ColumnFamilyConfigs::PubSubColumnFamily(), engine::ColumnFamilyConfigs::PropagateColumnFamily()}) {
     info("[compaction checker] Start to compact the column family: {}", cf.Name());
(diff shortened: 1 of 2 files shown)


git-hulk (approved):

LindaSummer (approved):

ddekob15-boop (commented):
A

PR #3022, opened 2025-06-09
feat(command): Add redis command to flush memtables

### LINKED ISSUE #3021: Add redis command to flush memtable
### Search before asking

### Motivation

There is currently no command that allows the user to flush the memtable through the redis cli, which would be useful if the user wants to guarantee no data loss when WAL is disabled. 

### Solution

Add a command that calls rocksdb::Db::Flush internally

Ryan4253:
Closes #3021 

## Descrption of Changes:
Adds a command to flush rocksdb memtables. Users who want to disable WAL and want to ensure memtables are flushed would find this useful (current alternatives like SIGTERM of shutdown only guarantee that flush happens eventually during destruction, rather than being guaranteed with command execution)

### Unit Test
```
[==========] 429 tests from 69 test suites ran. (20238 ms total)
[  PASSED  ] 428 tests.
```
#### Redis Client
```
<user>:48081> set a b
OK
<user>:48081> flushmemtables wait yes
OK
```
#### Kvrocks
```
[2025-06-09T16:05:08.507529-04:00][I][event_listener.cc:134] [event_listener/flush_begin] column family: metadata, thread_id: 128743, job_id: 4, reason: Manual Flush
[2025-06-09T16:05:08.545747-04:00][I][event_listener.cc:190] [event_listener/table_file_created] column family: metadata, file path: <path>/kvrocks/db/000104.sst, file size: 1404, job_id: 4, reason: flush, status: OK
[2025-06-09T16:05:08.562556-04:00][I][event_listener.cc:145] [event_listener/flush_completed] column family: metadata, thread_id: 128743, job_id: 4, file: <path>/kvrocks/db/000104.sst, reason: Manual Flush, is_write_slowdown: no, is_write_stall: no, largest seqno: 200004, smallest seqno: 200004
[2025-06-09T16:05:08.566222-04:00][I][cmd_server.cc:1473] flushmemtable was triggered by manual with executed success
```

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -1493,6 +1493,40 @@ class CommandSST : public Commander {
   rocksdb::IngestExternalFileOptions ingest_options_;
 };
 
+// command format: flushmemtable [async]
+class CommandFlushMemTable : public Commander {
+ public:
+  Status Parse([[maybe_unused]] const std::vector<std::string> &args) override {
+    if (args.size() > 2) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+
+    flush_options_.wait = true;
+    if (args.size() > 1) {
+      if (util::EqualICase(args[1], "async")) {
+        flush_options_.wait = false;
+      } else {
+        return {Status::RedisParseErr, "parameter must be 'ASYNC'"};
+      }
+    }
+
+    return Status::OK();
+  }
+
+  Status Execute([[maybe_unused]] engine::Context &ctx, Server *srv, [[maybe_unused]] Connection *conn,
+                 std::string *output) override {
+    const auto s = srv->storage->FlushMemTable(nullptr, flush_options_);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = redis::RESP_OK;
+    info("FLUSHMEMTABLE is triggered and executed successfully");
+    return Status::OK();
+  }
+
+ private:
+  rocksdb::FlushOptions flush_options_;
+};
+
 REDIS_REGISTER_COMMANDS(Server, MakeCmdAttr<CommandAuth>("auth", 2, "read-only ok-loading auth", NO_KEY),
                         MakeCmdAttr<CommandPing>("ping", -1, "read-only", NO_KEY),
                         MakeCmdAttr<CommandSelect>("select", 2, "read-only", NO_KEY),
@@ -1535,5 +1569,6 @@ REDIS_REGISTER_COMMANDS(Server, MakeCmdAttr<CommandAuth>("auth", 2, "read-only o
                         MakeCmdAttr<CommandApplyBatch>("applybatch", -2, "write no-multi", NO_KEY),
                         MakeCmdAttr<CommandDump>("dump", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandPollUpdates>("pollupdates", -2, "read-only admin", NO_KEY),
-                        MakeCmdAttr<CommandSST>("sst", -3, "write exclusive admin", 1, 1, 1), )
+                        MakeCmdAttr<CommandSST>("sst", -3, "write exclusive admin", 1, 1, 1),
+                        MakeCmdAttr<CommandFlushMemTable>("flushmemtable", -1, "exclusive write", NO_KEY), )
 }  // namespace redis
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -883,6 +883,11 @@ rocksdb::Status Storage::Compact(rocksdb::ColumnFamilyHandle *cf, const Slice *b
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status Storage::FlushMemTable(rocksdb::ColumnFamilyHandle *cf_handle, const rocksdb::FlushOptions &options) {
+  const auto &cf_handles = cf_handle ? std::vector<rocksdb::ColumnFamilyHandle *>{cf_handle} : cf_handles_;
+  return db_->Flush(options, cf_handles);
+}
+
 uint64_t Storage::GetTotalSize(const std::string &ns) {
   if (ns == kDefaultNamespace) {
     return sst_file_manager_->GetTotalSize();
--- a/src/storage/storage.h
+++ b/src/storage/storage.h
@@ -270,6 +270,8 @@ class Storage {
 
   [[nodiscard]] rocksdb::Status Compact(rocksdb::ColumnFamilyHandle *cf, const rocksdb::Slice *begin,
                                         const rocksdb::Slice *end);
+  [[nodiscard]] rocksdb::Status FlushMemTable(rocksdb::ColumnFamilyHandle *cf_handle,
+                                              const rocksdb::FlushOptions &options);
   [[nodiscard]] StatusOr<int> IngestSST(const std::string &folder,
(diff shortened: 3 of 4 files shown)


src/commands/cmd_server.cc
```diff
@@ -1493,6 +1493,43 @@ class CommandSST : public Commander {
...
+      if (!util::EqualICase(cmd, "wait")) {
+        return {Status::RedisParseErr, "unknown parameter: " + args[1]};
+      }
+
+      const std::string wait_option = GET_OR_RET(parser.TakeStr());
+      if (util::EqualICase(wait_option, "yes")) {
+        flush_options_.wait = true;
+      } else if (util::EqualICase(wait_option, "no")) {
+        flush_options_.wait = false;
+      } else {
```
git-hulk: We're using `sync|async` to determine whether to wait for the response or not. So we could keep them consistent.

git-hulk:
Out of curiosity, how would users use this command? Users should be aware that disabling the Write Ahead Log (WAL) may result in data loss in memtables. Only flushing it in SIGTERM would be helpful? cc @apache/kvrocks-committers 

We could add Go test cases to ensure it works correctly if we reach a consensus to introduce this command.

git-hulk:
@Ryan4253 Thanks for proposing the new command.

Ryan4253:
@git-hulk Thanks for the review!

Disabling the WAL does risk data loss on process crash, but is can also (theoretically) increase write throughput by 2x, and we think this is a worthwhile tradeoff.

This means we must flush all memtables to ensure data integrity on shutdown. Currently, this happens in ```rocksdb::DbImpl```'s destructor which is called unpredictably after we send SIGTERM or redis SHUTDOWN. If the shutdown takes a long time, we can't distinguish between a hung process or a long shutdown, which means we can't guarantee data integrity. 

With a synchronous flush memtable command we can explicitly trigger the flush and wait for it to finish to ensure safe shutdown. Having an async flush memtable command also means we can more eagerly schedule flushes to reduce the window of data loss.

src/commands/cmd_server.cc
```diff
@@ -1535,5 +1567,6 @@ REDIS_REGISTER_COMMANDS(Server, MakeCmdAttr<CommandAuth>("auth", 2, "read-only o
                         MakeCmdAttr<CommandApplyBatch>("applybatch", -2, "write no-multi", NO_KEY),
                         MakeCmdAttr<CommandDump>("dump", 2, "read-only", 1, 1, 1),
                         MakeCmdAttr<CommandPollUpdates>("pollupdates", -2, "read-only admin", NO_KEY),
-                        MakeCmdAttr<CommandSST>("sst", -3, "write exclusive admin", 1, 1, 1), )
+                        MakeCmdAttr<CommandSST>("sst", -3, "write exclusive admin", 1, 1, 1),
+                        MakeCmdAttr<CommandFlushMemTables>("flushmemtables", 2, "blocking write", NO_KEY), )
```
```suggestion
                        MakeCmdAttr<CommandFlushMemTables>("flushmemtables", 2, "exclusive write", NO_KEY), )
```

src/commands/cmd_server.cc
```diff
@@ -1493,6 +1493,38 @@ class CommandSST : public Commander {
   rocksdb::IngestExternalFileOptions ingest_options_;
 };
 
+// command format: flushmemtables [sync|async]
+class CommandFlushMemTables : public Commander {
```
PragmaTwice: I think the name of this command can be `KFLUSH`? cc @git-hulk

src/commands/cmd_server.cc
```diff
@@ -1493,6 +1493,38 @@ class CommandSST : public Commander {
...
+    CommandParser parser(args, 1);
+
+    const std::string param = GET_OR_RET(parser.TakeStr());
+    if (util::EqualICase(param, "sync")) {
+      flush_options_.wait = true;
+    } else if (util::EqualICase(param, "async")) {
+      flush_options_.wait = false;
+    } else {
+      return {Status::RedisParseErr, "Parameter must be 'sync' or 'async', got " + param};
+    }
```
PragmaTwice: I think we can have a default behavior `sync`, so it can be `kflush` or `kflush async`, where `sync` is omitted.

PragmaTwice (commented):
Golang test cases are required for a new command : )

You can add a new test file here: https://github.com/apache/kvrocks/tree/unstable/tests/gocase/unit

src/commands/cmd_server.cc
jihuayu: Doesn't the meaning of `KFLUSH` seem too broad? If we need to FLUSH other things in the future, could it cause confusion?

PragmaTwice: But in Rocksdb API it is just called `Flush` instead of `FlushMemtable`..

PragmaTwice: it can also be called `flushmemtable`, without `s`.

git-hulk: I am more in favor of naming `flushmemtable` since we now have other FLUSH commands, including:

- FLUSHBACKUP
- FLUSHDB
- FLUSHALL

src/commands/cmd_server.cc
```diff
@@ -1493,6 +1493,41 @@ class CommandSST : public Commander {
...
+    if (args.size() == 1) {
+      flush_options_.wait = true;
+      return Status::OK();
+    }
+
+    if (!util::EqualICase(args[1], "async")) {
+      return {Status::RedisParseErr, "Parameter must be 'async', got " + args[1]};
+    }
+
+    flush_options_.wait = false;
```
```suggestion

    flush_options_.wait = true;
    if (args.size() > 1) {
      if (util::EqualICase(args[1], "sync") || util::EqualICase(args[1], "async")) {
        flush_options_.wait = util::EqualICase(args[1], "sync");
      } else {
        return {Status::RedisParseErr, "parameter must be 'sync' or 'async'"};
      }
    }
```

Ryan4253:
Addressed all the changes so far! I don't have go on my local environment and it may be hard for me to get it set up, so I'll need some help with running the unit test I wrote whether locally or directly on CI.

src/commands/cmd_server.cc
```diff
@@ -1493,6 +1493,40 @@ class CommandSST : public Commander {
...
+    return Status::OK();
+  }
+
+  Status Execute([[maybe_unused]] engine::Context &ctx, Server *srv, [[maybe_unused]] Connection *conn,
+                 std::string *output) override {
+    const auto s = srv->storage->FlushMemTable(nullptr, flush_options_);
+    if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
+
+    *output = redis::RESP_OK;
+    info("flushmemtable was triggered by manual with executed success");
```
```suggestion
    info("FLUSHMEMTABLE is triggered and executed successfully");
```

Ryan4253:
test fixed 👍

src/commands/cmd_server.cc
```diff
@@ -1493,6 +1493,40 @@ class CommandSST : public Commander {
...
+ public:
+  Status Parse([[maybe_unused]] const std::vector<std::string> &args) override {
+    if (args.size() > 2) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+
+    flush_options_.wait = true;
+    if (args.size() > 1) {
+      if (util::EqualICase(args[1], "async")) {
+        flush_options_.wait = true;
```
```suggestion
        flush_options_.wait = false;
```

PragmaTwice (approved):

git-hulk (approved):

PR #3027, opened 2025-06-13
feat(storage): allow to enable the blob cache

### LINKED ISSUE #3026: Enable blob cache
### Search before asking

### Motivation

It would be nice to cache blob files too

### Solution

set ```cf_options->blob_cache``` to ```shared_block_cache```

Ryan4253:
Closes #3026 

## Descrption of Changes:
Make the shared block cache cache blob files too.

### Unit Test
```
[==========] 431 tests from 69 test suites ran. (71550 ms total)
[  PASSED  ] 430 tests.
```

#### Enabled
```
rocksdb.blobdb.cache.miss COUNT : 6113
rocksdb.blobdb.cache.hit COUNT : 55008
rocksdb.blobdb.cache.add COUNT : 6113
rocksdb.blobdb.cache.add.failures COUNT : 0
rocksdb.blobdb.cache.bytes.read COUNT : 275535072
rocksdb.blobdb.cache.bytes.write COUNT : 30620017
```

#### Disabled
```
rocksdb.blobdb.cache.miss COUNT : 0
rocksdb.blobdb.cache.hit COUNT : 0
rocksdb.blobdb.cache.add COUNT : 0
rocksdb.blobdb.cache.add.failures COUNT : 0
rocksdb.blobdb.cache.bytes.read COUNT : 0
rocksdb.blobdb.cache.bytes.write COUNT : 0
```

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -320,6 +320,10 @@ max-replication-mb 0
 # Default: 0
 max-io-mb 0
 
+# Whether to cache blob files within the block cache.
+# Default: no
+enable-blob-cache no
+
 # The maximum allowed space (in GB) that should be used by RocksDB.
 # If the total size of the SST files exceeds max_allowed_space, writes to RocksDB will fail.
 # Please see: https://github.com/facebook/rocksdb/wiki/Managing-Disk-Space-Utilization
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -192,6 +192,7 @@ Config::Config() {
       {"log-level", false, new EnumField<spdlog::level::level_enum>(&log_level, log_levels, spdlog::level::info)},
       {"pidfile", true, new StringField(&pidfile, kDefaultPidfile)},
       {"max-io-mb", false, new IntField(&max_io_mb, 0, 0, INT_MAX)},
+      {"enable-blob-cache", true, new YesNoField(&enable_blob_cache, false)},
       {"max-bitmap-to-string-mb", false, new IntField(&max_bitmap_to_string_mb, 16, 0, INT_MAX)},
       {"max-db-size", false, new IntField(&max_db_size, 0, 0, INT_MAX)},
       {"max-replication-mb", false, new IntField(&max_replication_mb, 0, 0, INT_MAX)},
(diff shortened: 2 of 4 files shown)


git-hulk:
Looks good to me, thank you. @mapleFU @PragmaTwice @torwig @aleksraiden Do you think we should add an option to opt out of this?

mapleFU:
The idea general LGTM, but I also wonder, would the performance more depending on block_cache's `high_pri_pool_ratio` and `low_pri_pool_ratio` in `LRUCacheOptions`, since the large blob might easily effect the global block cache?

For Blobs, previous mode would cache the items and fetching the blobs, if the memory is enough for these blobs, not caching the blobs is a bad idea. But when memory is not enough, this mode might utililize SSD's random read

Current mode making blob items into cache, when memory is enough this just enhance the performance, but in other scenerio this would making depending on cache policies 

Generally I think we can add a server side config for this( maybe default enable?) , and allowing config this when users see regression.

Ryan4253:
@git-hulk @mapleFU I agree that introducing this *can* cause regression given limited memory capacity since blob files are less valuable caching targets. I can go ahead and add a config to disable caching blob files if you guys think thats would be more backward compatible. Thanks!

git-hulk:
> @git-hulk @mapleFU I agree that introducing this _can_ cause regression given limited memory capacity since blob files are less valuable caching targets. I can go ahead and add a config to disable caching blob files if you guys think thats would be more backward compatible. Thanks!

Yes, it would be nice to have an option to opt out if users find any performance regression.

Ryan4253:
Option added 👍

kvrocks.conf
```diff
@@ -320,6 +320,10 @@ max-replication-mb 0
 # Default: 0
 max-io-mb 0
 
+# Whether to cache blob files within the block cache.
+# Default: yes
+enable-blob-cache yes
```
torwig: @git-hulk Perhaps we can disable this feature by default, and if users want to activate it and reap the benefits, they can turn it on. WDYT?

Ryan4253: Either works for me, just let me know which 👍

git-hulk: I'm good with disabling by default.

Ryan4253:
@git-hulk @torwig changed default to not enable blob cache

git-hulk:
@Ryan4253 Thanks for your contribution.

git-hulk (approved):

PragmaTwice (approved):

torwig (approved):

PR #3034, opened 2025-06-20
fix(config): Setting rocksdb.level0_slowdown_writes_trigger to 0 disables it

Ryan4253:
## Summary of Changes
1. Setting rocksdb.level0_slowdown_writes_trigger to 0 actually disables slowdown
2. Add comments documenting interaction between rocksdb.level0_slowdown_writes_trigger, rocksdb.level0_stop_writes_trigger, and rocksdb.level0_file_num_compaction_trigger

### Unit Test
```
[==========] 432 tests from 69 test suites ran. (19914 ms total)
[  PASSED  ] 431 tests.
```

### RocksDB Logs
kvrocks.conf:
```
# Soft limit on number of level-0 files. We slow down writes at this point. 
# A value of 0 means that no writing slow down will be triggered by number 
# of files in level-0. If the value is smaller than 
# rocksdb.level0_file_num_compaction_trigger, this will be set to its value
#
# Default: 20
rocksdb.level0_slowdown_writes_trigger 0

# Maximum number of level-0 files. We stop writes at this point. If the value
# is smaller than rocksdb.level0_slowdown_writes_trigger, this will be set to
# its value.
#
# Default: 40
rocksdb.level0_stop_writes_trigger 80

2025/06/25-10:48:53.406965 111964          Options.level0_slowdown_writes_trigger: 80
2025/06/25-10:48:53.406965 111964              Options.level0_stop_writes_trigger: 80
```

```
47255> CONFIG SET rocksdb.level0_stop_writes_trigger 100
OK
2025/06/25-10:49:06.193029 112129 [options/cf_options.cc:1129]            level0_slowdown_writes_trigger: 100
2025/06/25-10:49:06.193031 112129 [options/cf_options.cc:1131]                level0_stop_writes_trigger: 100
```

```
rd1internswe03:47255> CONFIG SET rocksdb.level0_slowdown_writes_trigger 50
OK
2025/06/25-10:49:13.349202 112129 [options/cf_options.cc:1129]            level0_slowdown_writes_trigger: 50
2025/06/25-10:49:13.349203 112129 [options/cf_options.cc:1131]                level0_stop_writes_trigger: 100
```

```
rd1internswe03:47255> CONFIG SET rocksdb.level0_stop_writes_trigger 90
OK
2025/06/25-10:49:19.184543 112129 [options/cf_options.cc:1129]            level0_slowdown_writes_trigger: 50
2025/06/25-10:49:19.184543 112129 [options/cf_options.cc:1131]                level0_stop_writes_trigger: 90
```

```
rd1internswe03:47255> CONFIG SET rocksdb.level0_slowdown_writes_trigger 0
OK
2025/06/25-10:49:34.874148 112129 [options/cf_options.cc:1129]            level0_slowdown_writes_trigger: 90
2025/06/25-10:49:34.874148 112129 [options/cf_options.cc:1131]                level0_stop_writes_trigger: 90
```

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -890,14 +890,18 @@ rocksdb.delayed_write_rate 0
 #  Default: no
 rocksdb.enable_pipelined_write no
 
-# Soft limit on number of level-0 files. We start slowing down writes at this
-#  point. A value <0 means that no writing slow down will be triggered by
-# number of files in level-0.
+# Soft limit on number of level-0 files. We slow down writes at this point. 
+# A value of 0 means that no writing slowdown will be triggered by number 
+# of files in level-0. If this value is smaller than 
+# rocksdb.level0_file_num_compaction_trigger, this will be set to 
+# rocksdb.level0_file_num_compaction_trigger instead.
 #
 # Default: 20
 rocksdb.level0_slowdown_writes_trigger 20
 
-# Maximum number of level-0 files.  We stop writes at this point.
+# Maximum number of level-0 files. We stop writes at this point. If this value
+# is smaller than rocksdb.level0_slowdown_writes_trigger, this will be set to
+# rocksdb.level0_slowdown_writes_trigger instead.
 #
 # Default: 40
 rocksdb.level0_stop_writes_trigger 40
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -281,7 +281,7 @@ Config::Config() {
       {"rocksdb.compaction_readahead_size", false,
        new IntField(&rocks_db.compaction_readahead_size, 2 * MiB, 0, 64 * MiB)},
       {"rocksdb.level0_slowdown_writes_trigger", false,
-       new IntField(&rocks_db.level0_slowdown_writes_trigger, 20, 1, 1024)},
+       new IntField(&rocks_db.level0_slowdown_writes_trigger, 20, 0, 1024)},
       {"rocksdb.level0_stop_writes_trigger", false, new IntField(&rocks_db.level0_stop_writes_trigger, 40, 1, 1024)},
       {"rocksdb.level0_file_num_compaction_trigger", false,
        new IntField(&rocks_db.level0_file_num_compaction_trigger, 4, 1, 1024)},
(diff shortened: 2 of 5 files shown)


kvrocks.conf
```diff
@@ -886,14 +886,17 @@ rocksdb.delayed_write_rate 0
 #  Default: no
 rocksdb.enable_pipelined_write no
 
-# Soft limit on number of level-0 files. We start slowing down writes at this
-#  point. A value <0 means that no writing slow down will be triggered by
-# number of files in level-0.
+# Soft limit on number of level-0 files. We slow down writes at this point. 
+# A value of 0 means that no writing slow down will be triggered by number 
+# of files in level-0. If the value is smaller than 
+# rocksdb.level0_file_num_compaction_trigger, this will be set to its value
```
mapleFU: >  If the value is smaller than 
> # rocksdb.level0_file_num_compaction_trigger, this will be set to its value

is this a bit weird? 🤔 Can you explain it more specifically?

Ryan4253: This should be better:
```
# Maximum number of level-0 files. We stop writes at this point. If this value
# is smaller than rocksdb.level0_slowdown_writes_trigger, this will be set to
# rocksdb.level0_slowdown_writes_trigger instead.
```

kvrocks.conf
```diff
@@ -886,14 +886,18 @@ rocksdb.delayed_write_rate 0
...
 rocksdb.enable_pipelined_write no
 
-# Soft limit on number of level-0 files. We start slowing down writes at this
-#  point. A value <0 means that no writing slow down will be triggered by
-# number of files in level-0.
+# Soft limit on number of level-0 files. We slow down writes at this point. 
+# A value of 0 means that no writing slow down will be triggered by number 
+# of files in level-0. If this value is smaller than 
+# rocksdb.level0_file_num_compaction_trigger, this will be set to 
+# rocksdb.level0_file_num_compaction_trigger instead
```
mapleFU:
```suggestion
# rocksdb.level0_file_num_compaction_trigger instead.
```

mapleFU (commented):
This generic ok to me, but I also wonder:
1. Can this set to < 0 directly to use rocksdb internal syntax? ( maybe: https://github.com/facebook/rocksdb/blob/f2d03736a7ae6dbe226aeffdc3a359b40d69c27e/db/column_family.cc#L328 )
2. When the `level0_file_num_compaction_trigger` is 0 firstly, the `level0_stop_writes_trigger` is used. However, if the `level0_stop_writes_trigger` is enlarged, what would happens here?

kvrocks.conf
Ryan4253: 1. Within RocksDB, when ```level0_slowdown_writes_trigger``` < ```level0_file_num_compaction_trigger```, it is set to ```level0_file_num_compaction_trigger```, which isn't the same as disabling slowdown. We want to set it to ```level_stop_writes_trigger``` instead so we need our own logic.
2. ```level0_file_num_compaction_trigger``` is always > 0 in our code:
```cpp
      {"rocksdb.level0_file_num_compaction_trigger", false,
       new IntField(&rocks_db.level0_file_num_compaction_trigger, 4, 1, 1024)},
```

If ```level0_file_num_compaction_trigger``` > ```level_stop_writes_trigger``` and ```level0_slowdown_writes_trigger``` is set to 0, everything would be set to ```level0_file_num_compaction_trigger``` which is valid.

src/config/config.cc
```diff
@@ -720,6 +720,15 @@ void Config::initFieldCallback() {
...
          return srv->storage->SetOptionForAllColumnFamilies(TrimRocksDbPrefix(k), v);
        }},
+      {"rocksdb.level0_slowdown_writes_trigger",
+       [this]([[maybe_unused]] Server *srv, [[maybe_unused]] const std::string &k,
+              [[maybe_unused]] const std::string &v) -> Status {
+         if (rocks_db.level0_slowdown_writes_trigger == 0) {
+           rocks_db.level0_slowdown_writes_trigger = rocks_db.level0_stop_writes_trigger;
+         }
+
+         return Status::OK();
```
```suggestion
         if (rocks_db.level0_slowdown_writes_trigger == 0) {
           rocks_db.level0_slowdown_writes_trigger = rocks_db.level0_stop_writes_trigger;
         }

         return set_cf_option_cb(srv, k, std::to_string(rocks_db.level0_slowdown_writes_trigger));
```

PragmaTwice: The actual option value changing is missing, correct?

kvrocks.conf
mapleFU: In other words, if:
1. level0_slowdown_writes_trigger == 0, and level_stop_writes_trigger == 40. The `level0_slowdown_writes_trigger` would be 40 in this patch
2. When user enlarge `level_stop_writes_trigger` dynamically, would `level0_slowdown_writes_trigger` stay 40?

src/config/config.cc
Ryan4253: The original code does work since we pass the RocksDB later on open and RocksDB would set the constants accordingly. However, I'll go ahead and switch to explicitly calling SetDBOptions / SetCFOptions since that is what all the other configs do.

kvrocks.conf
Ryan4253: 1. Yes
2. Yes, this is a limitation. 
The main reason I brought up this issue / PR is that the ```kvrocks.conf``` documentation on ```rocksdb.level0_slowdown_writes_trigger``` doesn't match the actual behavior of the code. If you don't think this change is appropriate since it can't be matched at runtime I'm good with changing the docs in ```kvrocks.conf``` as well.

src/config/config.cc
PragmaTwice: Yeah but it will not be changed via `config set` if so.

kvrocks.conf
mapleFU: Emmm I know what you mean here, I just think that we can just set the `level0_slowdown_writes_trigger` same as `level_stop_writes_trigger` in the config explicitly to disable this? 😂

src/config/config.cc
Ryan4253: Oh this made me notice I completely misunderstood how the config callbacks work, will come back with a fix

Ryan4253:
Sorry for the confusion, I misunderstood how the config callbacks worked. This should do what is intended for both initialization and ```CONFIG SET``` and added unit tests too.

src/config/config.h
```diff
@@ -279,6 +279,7 @@ struct Config {
   std::vector<std::string> rename_command_;
   std::string histogram_bucket_boundaries_str_;
   std::set<std::string> deprecated_fields_;
+  bool disable_l0_slowdown_writes_trigger_ = false;
```
PragmaTwice: Is it necessary to introduce a new variable here?

Ryan4253: I added this so that ```CONFIG GET``` would return `level0_stop_writes_trigger``` if the slowdown trigger is disabled. On second thought, we should return 0 instead to signify that it is disabled, so yeah this is not necessary.

Ryan4253:
Simplified implementation

kvrocks.conf
Ryan4253: Oh yeah I agree

PragmaTwice:
PTAL @mapleFU

src/config/config.cc
```diff
@@ -728,6 +728,26 @@ void Config::initFieldCallback() {
...
+       [this, &set_cf_option_cb](Server *srv, const std::string &k, [[maybe_unused]] const std::string &v) -> Status {
+         if (rocks_db.level0_slowdown_writes_trigger == 0) {
+           Status s = set_cf_option_cb(srv, "rocksdb.level0_slowdown_writes_trigger",
+                                       std::to_string(rocks_db.level0_stop_writes_trigger));
+           if (!s.IsOK()) {
+             return s;
+           }
+         }
+
+         return set_cf_option_cb(srv, k, std::to_string(rocks_db.level0_stop_writes_trigger));
```
mapleFU:
```
Status Storage::SetOptionForAllColumnFamilies(const std::string &key, const std::string &value) {
  for (auto &cf_handle : cf_handles_) {
    auto s = db_->SetOptions(cf_handle, {{key, value}});
    if (!s.ok()) return {Status::NotOK, s.ToString()};
  }
  return Status::OK();
}
```

This method create a `std::map<std::string, std::string>`, and update the `cf_handles_` configs. If `level0_slowdown_writes_trigger` is set firstly, I wonder whether it would:
1. being duplicated called?
2. Failed set when `rocks_db.level0_stop_writes_trigger` is too large?

Should we reverse this order? Or set the two elements using std::map together?

Ryan4253: Let's set them together with std::map

Ryan4253: Actually I am not too sure about what you mean by 1. Within the function currently set_cf_option_cb would get called twice but it is called for two different keys (slowdown and stop respectively). For 2 since the two have the same limits, if setting the first one fails it would exit and nothing would be set which I think matches what is intended.

mapleFU: >  Actually I am not too sure about what you mean by 1. 

I just means that rocksdb::DBImpl::SetOptions is called twice and some checking is being done twice

Ryan4253: I see, the code only checks that slowdown <= stop on construction in function ```SanitizeCfOptions```. There are no limits when calling ```rocksdb::DBImpl::SetOptions``` so we are fine since we preserve the invariant.

mapleFU: > Let's set them together with std::map

Yeah, I'll try to approve if this is done

Ryan4253:
Switched to using std::unordered_map! @mapleFU

mapleFU (dismissed):
cc @git-hulk

Ryan4253:
Run clang-format. It seems like adding the two configs pushed clang-format over some limit and now the entire callback map got reformatted.

mapleFU (approved):

PragmaTwice (approved):

PR #3040, opened 2025-06-26
feat(config): allow to configure the rocksdb.sst_file_delete_rate_bytes_per_sec

sryanyuan:
PurgeObsoleteFiles will be called during compaction, flush, and similar operations, and it will be performed in the background threads (avoid-unnecessary-blocking-io is enabled).

Additionally, it may also be triggered when destroying an iterator in a user thread, in the case:

SST files will be deleted by background threads
while archived WAL log files will be unlinked directly in the user thread.
This design might cause high latency if there are many files to be deleted.

rocksdb.sst_file_delete_rate_bytes_per_sec can set the delete rate limit in bytes per second for SST files deletion.

In scenarios involving frequent database iterations (e.g., HGETALL, SCAN) obsolete WAL files may be deleted synchronously, causing latency spikes. Enabling this option activates a controlled slow deletion mechanism, which also resolves WAL deletion latency issues when an iterator is released.

### DIFF
--- a/tests/gocase/unit/config/config_test.go
+++ b/tests/gocase/unit/config/config_test.go
@@ -371,3 +371,28 @@ func TestConfigRocksDBOptions(t *testing.T) {
 		require.EqualValues(t, "1073741824", result[parameter])
 	})
 }
+
+func TestConfigSstFileDeleteRateBytesPerSec(t *testing.T) {
+	t.Parallel()
+	srv := util.StartServer(t, map[string]string{})
+	defer srv.Close()
+
+	cli := srv.NewClient()
+	defer func() { require.NoError(t, cli.Close()) }()
+
+	t.Run("Get and Set rocksdb.sst_file_delete_rate_bytes_per_sec", func(t *testing.T) {
+		ctx := context.Background()
+		parameter := "rocksdb.sst_file_delete_rate_bytes_per_sec"
+		result, err := cli.ConfigGet(ctx, parameter).Result()
+		require.NoError(t, err)
+		require.EqualValues(t, "0", result[parameter])
+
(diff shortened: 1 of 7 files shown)


git-hulk:
@sryanyuan Thanks for your contribution. Could you please add a test case in [config_test.go](https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/config/config_test.go) to ensure this configuration?

sryanyuan:
> @sryanyuan Thanks for your contribution. Could you please add a test case in [config_test.go](https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/config/config_test.go) to ensure this configuration?

Done

tests/gocase/unit/config/config_test.go
```diff
@@ -371,3 +371,28 @@ func TestConfigRocksDBOptions(t *testing.T) {
...
+
+	t.Run("Get and Set rocksdb.sst_file_delete_rate_bytes_per_sec", func(t *testing.T) {
+		ctx := context.Background()
+		parameter := "rocksdb.sst_file_delete_rate_bytes_per_sec"
+		result, err := cli.ConfigGet(ctx, parameter).Result()
+		require.NoError(t, err)
+		require.EqualValues(t, "0", result[parameter])
+
+		util.ErrorRegexp(t, cli.ConfigSet(ctx, parameter, "-1").Err(), ".*out of numeric range")
+		util.ErrorRegexp(t, cli.ConfigSet(ctx, parameter, "18446744073709551616").Err(), ".*Result too large")
```
git-hulk:
```suggestion
		util.ErrorRegexp(t, cli.ConfigSet(ctx, parameter, "18446744073709551616").Err(), ".*out of range")
```

git-hulk (approved):
LGTM, thanks.

PR #3047, opened 2025-07-07
feat(replication): implement WAIT without timeout support

zhixinwen:
This PR implements Redis [WAIT](https://redis.io/docs/latest/commands/wait/) command (pending timeout implementation).

The WAIT command works as follows:
1. Get the current RocksDB latest sequence.
2. Check each replica if they have reached the sequence returned in step 1.  
     1. If yes, directly return.
      2. If not, put the command to blocking.
3. When master sends changes to replicas, wake up the blocked command and see if it meets the number of replicas meet.

Key differences from Redis:
1. Timeout not supported. I tired it initially and [reverted](https://github.com/apache/kvrocks/commit/92bec60b606ccb0aaf8beae70b3d4a838423dc2f), because it has a bug that if master has no new update, the timeout would not be respected. It is more complicated than I expect to do properly. We can follow up on this support later, as I want to ensure the core implementation looks good.
2. Redis tracks latest write per client, but KVRocks latest write is global. This means WAIT in KVRocks may be less efficient and "over wait", but functionality wise it should be OK.
3. Redis replica would ack with the sequence number it has processed. KVRocks is fire and forget style, so when WAIT returns it only guarantees the change has been sent to client, but it does not guarantee client has processed it. This is basically how KVRocks replication works, and I don't think we should touch that for this feature.

### DIFF
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -394,6 +394,6 @@ REDIS_REGISTER_COMMANDS(Replication, MakeCmdAttr<CommandReplConf>("replconf", -3
                         MakeCmdAttr<CommandFetchMeta>("_fetch_meta", 1, "read-only no-multi no-script", NO_KEY),
                         MakeCmdAttr<CommandFetchFile>("_fetch_file", 2, "read-only no-multi no-script", NO_KEY),
                         MakeCmdAttr<CommandDBName>("_db_name", 1, "read-only no-multi", NO_KEY),
-                        MakeCmdAttr<CommandWait>("wait", 2, "read-only no-script blocking", NO_KEY), )
+                        MakeCmdAttr<CommandWait>("wait", 2, "read-only no-multi no-script blocking", NO_KEY), )
 
 }  // namespace redis
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -347,8 +347,8 @@ class CommandDBName : public Commander {
 class CommandWait : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    auto num_replicas_result = ParseInt<int>(args[1], 10);
-    if (!num_replicas_result || *num_replicas_result <= 0) {
+    auto num_replicas_result = ParseInt<int>(args[1], {1, UINT64_MAX}, 10);
+    if (!num_replicas_result) {
       return {Status::RedisParseErr, "numreplicas should be a positive integer"};
     }
 
@@ -370,7 +370,7 @@ class CommandWait : public Commander {
     int reached_replicas = srv->GetReplicasReachedSequence(current_seq);
 
     // If we already have enough replicas, return immediately
-    if (reached_replicas >= num_replicas_) {
+    if (reached_replicas >= static_cast<int>(num_replicas_)) {
       *output = redis::Integer(reached_replicas);
       return Status::OK();
     }
@@ -384,7 +384,7 @@ class CommandWait : public Commander {
   }
 
  private:
-  int num_replicas_ = 0;
+  uint64_t num_replicas_ = 0;
 };
 
 REDIS_REGISTER_COMMANDS(Replication, MakeCmdAttr<CommandReplConf>("replconf", -3, "read-only no-script", NO_KEY),
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -347,8 +347,8 @@ class CommandDBName : public Commander {
 class CommandWait : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    auto num_replicas_result = ParseInt<uint64_t>(args[1], 10);
-    if (!num_replicas_result || *num_replicas_result == 0) {
+    auto num_replicas_result = ParseInt<int64_t>(args[1], 10);
+    if (!num_replicas_result || *num_replicas_result <= 0) {
       return {Status::RedisParseErr, "numreplicas should be a positive integer"};
     }
 
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -347,8 +347,8 @@ class CommandDBName : public Commander {
 class CommandWait : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    auto num_replicas_result = ParseInt<int>(args[1], {1, UINT64_MAX}, 10);
-    if (!num_replicas_result) {
+    auto num_replicas_result = ParseInt<int>(args[1], 10);
+    if (!num_replicas_result || *num_replicas_result <= 0) {
       return {Status::RedisParseErr, "numreplicas should be a positive integer"};
     }
 
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -347,8 +347,8 @@ class CommandDBName : public Commander {
 class CommandWait : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    auto num_replicas_result = ParseInt<int>(args[1], 10);
-    if (!num_replicas_result || *num_replicas_result <= 0) {
+    auto num_replicas_result = ParseInt<uint64_t>(args[1], 10);
+    if (!num_replicas_result) {
       return {Status::RedisParseErr, "numreplicas should be a positive integer"};
     }
 
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -394,6 +394,6 @@ REDIS_REGISTER_COMMANDS(Replication, MakeCmdAttr<CommandReplConf>("replconf", -3
                         MakeCmdAttr<CommandFetchMeta>("_fetch_meta", 1, "read-only no-multi no-script", NO_KEY),
                         MakeCmdAttr<CommandFetchFile>("_fetch_file", 2, "read-only no-multi no-script", NO_KEY),
                         MakeCmdAttr<CommandDBName>("_db_name", 1, "read-only no-multi", NO_KEY),
-                        MakeCmdAttr<CommandWait>("wait", 2, "read-only blocking", NO_KEY), )
+                        MakeCmdAttr<CommandWait>("wait", 2, "read-only no-script blocking", NO_KEY), )
 
 }  // namespace redis
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -347,10 +347,12 @@ class CommandDBName : public Commander {
 class CommandWait : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    auto num_replicas_result = ParseInt<size_t>(args[1], 10);
-    if (!num_replicas_result) {
+    auto num_replicas_result = ParseInt<int>(args[1], 10);
+    if (!num_replicas_result || *num_replicas_result <= 0) {
       return {Status::RedisParseErr, "numreplicas should be a positive integer"};
     }
+
+    num_replicas_ = *num_replicas_result;
 
     return Commander::Parse(args);
   }
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -367,7 +367,7 @@ class CommandWait : public Commander {
 
     // Get current sequence number
     auto current_seq = srv->storage->LatestSeqNumber();
-    
+
     // Check if we already have enough replicas at the current sequence
     int reached_replicas = srv->GetReplicasReachedSequence(current_seq);
 
@@ -379,7 +379,7 @@ class CommandWait : public Commander {
 
     // Block the connection and wait for replicas to catch up
     srv->BlockOnWait(conn, current_seq, num_replicas_);
-    
+
     // The connection will be woken up by WakeupWaitConnections when enough replicas
     // have reached the target sequence
     return {Status::BlockingCmd};
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -347,8 +347,8 @@ class CommandDBName : public Commander {
 class CommandWait : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
-    // WAIT numreplicas timeout
-    if (args.size() != 3) {
+    // WAIT numreplicas
+    if (args.size() != 2) {
       return {Status::RedisParseErr, errWrongNumOfArguments};
     }
 
@@ -359,15 +359,6 @@ class CommandWait : public Commander {
     num_replicas_ = *num_replicas_result;
     if (num_replicas_ < 0) {
       return {Status::RedisParseErr, "numreplicas should be a positive integer"};
-    }
-
-    auto timeout_result = ParseInt<int64_t>(args[2], 10);
-    if (!timeout_result) {
-      return {Status::RedisParseErr, "timeout should be a positive integer"};
-    }
-    timeout_ms_ = *timeout_result;
-    if (timeout_ms_ < 0) {
-      return {Status::RedisParseErr, "timeout should be a positive integer"};
     }
 
     return Commander::Parse(args);
@@ -391,23 +382,16 @@ class CommandWait : public Commander {
       return Status::OK();
     }
 
-    // If timeout is 0, return immediately with current replica count
-    if (timeout_ms_ == 0) {
-      *output = redis::Integer(reached_replicas);
-      return Status::OK();
-    }
-
     // Block the connection and wait for replicas to catch up
-    srv->BlockOnWait(conn, current_seq, num_replicas_, timeout_ms_);
+    srv->BlockOnWait(conn, current_seq, num_replicas_);
     
     // The connection will be woken up by WakeupWaitConnections when enough replicas
-    // have reached the target sequence or when timeout occurs
+    // have reached the target sequence
     return {Status::BlockingCmd};
   }
 
  private:
   int num_replicas_ = 0;
-  int64_t timeout_ms_ = 0;
 };
 
 REDIS_REGISTER_COMMANDS(Replication, MakeCmdAttr<CommandReplConf>("replconf", -3, "read-only no-script", NO_KEY),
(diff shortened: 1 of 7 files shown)


PragmaTwice (commented):

Could you format your code by clang-format-14 via `./x.py format`? See https://kvrocks.apache.org/community/contributing#code-style for details.

src/commands/cmd_replication.cc
```diff
@@ -344,10 +344,61 @@ class CommandDBName : public Commander {
...
 };
 
+class CommandWait : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    // WAIT numreplicas
+    if (args.size() != 2) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+
```
PragmaTwice: Number of arguments will be checked before the `Parse` function, so it can be removed.

src/commands/cmd_replication.cc
```diff
@@ -344,10 +344,61 @@ class CommandDBName : public Commander {
...
+  int num_replicas_ = 0;
+};
+
 REDIS_REGISTER_COMMANDS(Replication, MakeCmdAttr<CommandReplConf>("replconf", -3, "read-only no-script", NO_KEY),
                         MakeCmdAttr<CommandPSync>("psync", -2, "read-only no-multi no-script", NO_KEY),
                         MakeCmdAttr<CommandFetchMeta>("_fetch_meta", 1, "read-only no-multi no-script", NO_KEY),
                         MakeCmdAttr<CommandFetchFile>("_fetch_file", 2, "read-only no-multi no-script", NO_KEY),
-                        MakeCmdAttr<CommandDBName>("_db_name", 1, "read-only no-multi", NO_KEY), )
+                        MakeCmdAttr<CommandDBName>("_db_name", 1, "read-only no-multi", NO_KEY),
+                        MakeCmdAttr<CommandWait>("wait", 2, "read-only blocking", NO_KEY), )
```
PragmaTwice: Maybe we can check if it works well in scripting and MULTI-EXEC? If not we can add some flags here.

src/server/server.cc
```diff
@@ -700,6 +701,76 @@ void Server::OnEntryAddedToStream(const std::string &ns, const std::string &key,
...
+void Server::UnblockOnWait(redis::Connection *conn) {
+  std::lock_guard<std::mutex> guard(wait_contexts_mu_);
+  
+  for (auto it = wait_contexts_.begin(); it != wait_contexts_.end(); ++it) {
+    if (it->conn == conn) {
+      wait_contexts_.erase(it);
+      DecrBlockedClientNum();
+      break;
+    }
+  }
```
PragmaTwice: Maybe it can be replaced by something like `std::find_if`?

tests/gocase/unit/wait/wait_test.go
```diff
@@ -0,0 +1,116 @@
...
+		util.SlaveOf(t, slaveRdb, masterSrv)
+		util.WaitForSync(t, slaveRdb)
+
+		// Now WAIT should complete
+		select {
+		case <-done:
+			// Success - command completed after replica connected
+		case <-time.After(5 * time.Second):
+			t.Fatal("WAIT command did not complete after replica connected")
+		}
```
PragmaTwice, line 122: If the WAIT didn't work, i.e. WAIT is a noop, here we can also get `done` if I understand correctly?

I think we can add a test for such case.

git-hulk:
> Redis tracks latest write per client, but KVRocks latest write is global. This means WAIT in KVRocks may be less efficient and "over wait", but functionality wise it should be OK.

I'm not quite understand this statement. Redis also requires to wait for the number of replica's offset is fulfilled the watermark. And it should not be different with Redis in performance.

> Redis replica would ack with the sequence number it has processed. KVRocks is fire and forget style, so when WAIT returns it only guarantees the change has been sent to client, but it does not guarantee client has processed it. This is basically how KVRocks replication works, and I don't think we should touch that for this feature.

Redis depends on the `replconf` command to update the replication offset. Kvrocks can also enhance this feature by doing this, but it may need more efforts to achieve this. So I feel good to support the WAIT command with current implementation, and improve it then.

src/commands/cmd_replication.cc
zhixinwen: good point, let me add the flags

zhixinwen:
> I'm not quite understand this statement. Redis also requires to wait for the number of replica's offset is fulfilled the watermark. And it should not be different with Redis in performance.
> 
My understanding is in Redis each client knows its last write sequence number, so WAIT would only need to wait for its own write. In KVRocks, the following scenario is possible if I get the code correct:
client1: SET K1 V1 (seq number: 1)
client2: SET K2 V2 (seq number: 2)
client1: WAIT 1 (it would be waiting for seq number 2, but in Redis it would wait for seq number 1)

again I am not expert in either system, so I may be wrong here.

tests/gocase/unit/wait/wait_test.go
zhixinwen: Modified the test to make sure WAIT is blocked until replica is connected

zhixinwen:
When I tried to test it in script I get an error:
```
redis-cli -p 6666 EVAL "redis.call('SET', 'k1', 'v1'); return redis.call('WAIT', 1)" 0
(error) ERR running script (call to f_da109245b6e0086adfc0257445aaf319121fe5b8): 
```

When I tried to use MULTI, I find the whole queue seems to be messed up, for example, I do not expect QUEUED for EXEC:
```
127.0.0.1:6666> MULTI
QUEUED
127.0.0.1:6666(TX)> SET k1 v1
(empty array)
127.0.0.1:6666(TX)> WAIT 3
OK
127.0.0.1:6666(TX)> EXEC
QUEUED 
```

I am marking the cmd not ready for script/multi for now.

git-hulk:
> My understanding is in Redis each client knows its last write sequence number, so WAIT would only need to wait for its own write. In KVRocks, the following scenario is possible if I get the code correct:

oh, I see. Redis did track the write offset for each client. Let me check if possible to do that in Kvrocks as well.

src/commands/cmd_replication.cc
```diff
@@ -344,10 +344,56 @@ class CommandDBName : public Commander {
...
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto num_replicas_result = ParseInt<int>(args[1], 10);
+    if (!num_replicas_result) {
+      return {Status::RedisParseErr, "numreplicas should be a positive integer"};
+    }
+    num_replicas_ = *num_replicas_result;
+    if (num_replicas_ < 0) {
+      return {Status::RedisParseErr, "numreplicas should be a positive integer"};
+    }
```
PragmaTwice:
```suggestion
    auto num_replicas_result = ParseInt<int>(args[1], 10);
    if (!num_replicas_result || *num_replicas_result < 0) {
      return {Status::RedisParseErr, "numreplicas should be a positive integer"};
    }
    num_replicas_ = *num_replicas_result;
```

PragmaTwice: Also, if it is a value that should be non-negative, consider to use `size_t` or `uintN_t`. Then we don't need this check.

tests/gocase/unit/wait/wait_test.go
```diff
@@ -0,0 +1,118 @@
...
+			require.NoError(t, masterRdb.Do(ctx, "WAIT", "1").Err())
+			done <- true
+		}()
+
+		select {
+		case <-done:
+			t.Fatal("WAIT command did not block")
+		default:
+			// Success - command is blocked
+		}
```
PragmaTwice: Maybe we can wait for 1 sec before this `select`?

zhixinwen: Why do we want to wait here?

zhixinwen: You are right, should wait here to avoid no-op WAIT

zhixinwen:
I am eventually looking to implement `WAITOF`, because we want to have min data loss possible.
In the current replication logic in KVRocks, replica does not explicitly track the sequence number it has processed. It just applies whatever sent by primary. Is it possible that there is data loss inbetween the replication connection that would cause issue?

For example, primary sends changes with sequence number 1,2,3 to replica and replica processed them. When primary sends change with sequence number 4 and 5, somehow 4 is lost in network and replica only sees 5. Is this something we would need to prevent this case or it is handled in KVRocks already?

src/common/status.h
```diff
@@ -181,7 +181,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> :
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
-      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
+      : BaseType((typename StringInStatusOr<U>::BaseType&&)(std::move(v))) {}
```
PragmaTwice: Could you revert this change?

PragmaTwice (commented):
Looks good. cc @mapleFU @git-hulk

src/common/status.h
zhixinwen: good catch

src/commands/cmd_replication.cc
zhixinwen: Use size_t does not return error.

git-hulk (dismissed):

src/commands/cmd_replication.cc
```diff
@@ -344,10 +344,54 @@ class CommandDBName : public Commander {
   }
 };
 
+class CommandWait : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto num_replicas_result = ParseInt<int>(args[1], 10);
+    if (!num_replicas_result || *num_replicas_result <= 0) {
+      return {Status::RedisParseErr, "numreplicas should be a positive integer"};
+    }
```
PragmaTwice:
```suggestion
    auto num_replicas_result = ParseInt<uint64_t>(args[1], 10);
    if (!num_replicas_result) {
      return {Status::RedisParseErr, "numreplicas should be a positive integer"};
    }
```

src/server/server.cc
```diff
@@ -700,6 +701,63 @@ void Server::OnEntryAddedToStream(const std::string &ns, const std::string &key,
...
+int Server::GetReplicasReachedSequence(rocksdb::SequenceNumber target_seq) {
+  std::lock_guard<std::mutex> slave_guard(slave_threads_mu_);
+  int reached_replicas = 0;
+  for (const auto &slave : slave_threads_) {
+    if (!slave->IsStopped() && slave->GetCurrentReplSeq() >= target_seq) {
+      reached_replicas++;
+    }
+  }
+  return reached_replicas;
+}
```
```suggestion
size_t Server::GetReplicasReachedSequence(rocksdb::SequenceNumber target_seq) {
  std::lock_guard<std::mutex> slave_guard(slave_threads_mu_);
  size_t reached_replicas = 0;
  for (const auto &slave : slave_threads_) {
    if (!slave->IsStopped() && slave->GetCurrentReplSeq() >= target_seq) {
      reached_replicas++;
    }
  }
  return reached_replicas;
}
```

src/server/server.h
```diff
@@ -229,6 +230,14 @@ class Server {
   void WakeupBlockingConns(const std::string &key, size_t n_conns);
   void OnEntryAddedToStream(const std::string &ns, const std::string &key, const redis::StreamEntryID &entry_id);
 
+  // WAIT command infrastructure
+  void BlockOnWait(redis::Connection *conn, rocksdb::SequenceNumber target_seq, uint64_t num_replicas);
+  void WakeupWaitConnections(rocksdb::SequenceNumber seq);
+  void CleanupWaitConnection(redis::Connection *conn);
+
+  // Helper methods for WAIT command
+  int GetReplicasReachedSequence(rocksdb::SequenceNumber target_seq);
```
```suggestion
  size_t GetReplicasReachedSequence(rocksdb::SequenceNumber target_seq);
```

src/commands/cmd_replication.cc
```diff
@@ -344,10 +344,54 @@ class CommandDBName : public Commander {
...
+    auto current_seq = srv->storage->LatestSeqNumber();
+
+    // Check if we already have enough replicas at the current sequence
+    int reached_replicas = srv->GetReplicasReachedSequence(current_seq);
+
+    // If we already have enough replicas, return immediately
+    if (reached_replicas >= static_cast<int>(num_replicas_)) {
+      *output = redis::Integer(reached_replicas);
+      return Status::OK();
+    }
```
PragmaTwice:
```suggestion
    size_t reached_replicas = srv->GetReplicasReachedSequence(current_seq);

    // If we already have enough replicas, return immediately
    if (reached_replicas >= num_replicas_) {
      *output = redis::Integer(reached_replicas);
      return Status::OK();
    }
```

src/server/server.cc
```diff
@@ -700,6 +701,63 @@ void Server::OnEntryAddedToStream(const std::string &ns, const std::string &key,
...
+  std::lock_guard<std::mutex> guard(wait_contexts_mu_);
+
+  for (auto it = wait_contexts_.begin(); it != wait_contexts_.end();) {
+    // Check if target sequence is reached
+    if (seq >= it->target_seq) {
+      // Count how many replicas have reached the target sequence
+      int reached_replicas = GetReplicasReachedSequence(it->target_seq);
+
+      // If enough replicas have reached the target sequence, wake up the connection
+      if (reached_replicas >= static_cast<int>(it->num_replicas)) {
```
```suggestion
      size_t reached_replicas = GetReplicasReachedSequence(it->target_seq);

      // If enough replicas have reached the target sequence, wake up the connection
      if (reached_replicas >= it->num_replicas) {
```

src/commands/cmd_replication.cc
```diff
@@ -344,10 +344,54 @@ class CommandDBName : public Commander {
   }
 };
 
+class CommandWait : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto num_replicas_result = ParseInt<uint64_t>(args[1], 10);
+    if (!num_replicas_result) {
```
PragmaTwice:
```suggestion
    if (!num_replicas_result || *num_replicas_result == 0) {
```

src/commands/cmd_replication.cc
```diff
@@ -344,10 +344,54 @@ class CommandDBName : public Commander {
   }
 };
 
+class CommandWait : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    auto num_replicas_result = ParseInt<uint64_t>(args[1], 10);
+    if (!num_replicas_result || *num_replicas_result == 0) {
+      return {Status::RedisParseErr, "numreplicas should be a positive integer"};
+    }
```
PragmaTwice:
```suggestion
    auto num_replicas_result = ParseInt<int64_t>(args[1], 10);
    if (!num_replicas_result || *num_replicas_result <= 0) {
      return {Status::RedisParseErr, "numreplicas should be a positive integer"};
    }
```

PragmaTwice (dismissed):

tests/gocase/unit/wait/wait_test.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+func TestWaitCommand(t *testing.T) {
+	// Start master server
+	masterSrv := util.StartServer(t, map[string]string{})
+	defer masterSrv.Close()
+
+	// Start slave server
+	slaveSrv := util.StartServer(t, map[string]string{})
+	defer slaveSrv.Close()
```
PragmaTwice, line 39:
```suggestion
```

tests/gocase/unit/wait/wait_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestWaitCommand(t *testing.T) {
+	// Start master server
+	masterSrv := util.StartServer(t, map[string]string{})
+	defer masterSrv.Close()
+
+	// Start slave server
+	slaveSrv := util.StartServer(t, map[string]string{})
```
PragmaTwice, line 38:
```suggestion
	slaveSrv := util.StartServer(t, map[string]string{})
	defer slaveSrv.close()
```

tests/gocase/unit/wait/wait_test.go
```diff
@@ -0,0 +1,118 @@
...
+	defer slaveSrv.close()
```
PragmaTwice:
```suggestion
	defer slaveSrv.Close()
```

zhixinwen:
The test was failing because `slave_threads_` is cleaned up by `CleanupExitedSlaves` and `checkLivenessIfNeed`, which are not real time. Also `next_repl_seq_.store(curr_seq);` is called without `util::SockSend` being called, so it can be updated with the broken network call.
 
I changed the test so it would wait for the replicas to drop. Once we have ack (which I am working on now), we should not need this.

zhixinwen:
good, only Darwin arm64 fails now. But I cannot reproduce it on my Mac and I do not see a way to rerun the test myself..

I guess the issue is somehow the close was not graceful, and we have to wait for KILL. But we do not give enough time for `require.Eventually` for KILL to kick in. I will use `SLAVEOF NO ONE` instead and increase `require.Eventually` wait time.

zhixinwen:
I tried to wake up WAITOF only after `auto s = util::SockSend(conn_->GetFD(), batches_bulk, conn_->GetBufferEvent());` succeeds, so that it only wake up after the data is sent to replica. But I notice it does not return error after slave is stopped.. it may be related to tcp buffer.

So it is possible that the liveness check would not be able to immediately detect the slave disconnection until after it has filled in the buffer. 5 second only gives enough time for 2 liveness check, and I am increasing the wait time.

zhixinwen:
```
--- FAIL: TestKMetadata (1.04s)
    --- FAIL: TestKMetadata/Test_KMetadata_for_set_type (0.00s)
        kmetadata_test.go:170: 
            	Error Trace:	/home/runner/work/kvrocks/kvrocks/tests/gocase/unit/kmetadata/kmetadata_test.go:170
            	Error:      	Not equal: 
            	            	expected: 4
            	            	actual  : 3
            	Test:       	TestKMetadata/Test_KMetadata_for_set_type
```

unrelated error

git-hulk (approved):

PragmaTwice (approved):

PR #3049, opened 2025-07-10
feat(command): add new command to release block cache

sryanyuan:
Add a command `FLUSHBLOCKCACHE` to explicitly release the block cache when a master node is demoted to a slave, avoiding unnecessary memory retention in the slave role.

### DIFF
--- a/tests/gocase/unit/flushblockcache/flushblockcache_test.go
+++ b/tests/gocase/unit/flushblockcache/flushblockcache_test.go
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package flushblockcache
+
+import (
+	"context"
+	"strconv"
+	"testing"
+
+	"github.com/apache/kvrocks/tests/gocase/util"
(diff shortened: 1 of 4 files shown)


tests/gocase/unit/flushblockcache/flushblockcache_test.go
```diff
@@ -0,0 +1,80 @@
...
+		_, err = rdb.Do(ctx, "FLUSHMEMTABLE").Result()
+		require.NoError(t, err)
+		_, err = rdb.Do(ctx, "GET", "A").Result()
+		require.NoError(t, err)
+		initCacheSize, err := getBlockCacheSize(rdb)
+		require.NoError(t, err)
+		_, err = rdb.Do(ctx, "FLUSHBLOCKCACHE").Result()
+		require.NoError(t, err)
+		cacheSize, err := getBlockCacheSize(rdb)
+		require.Less(t, cacheSize, initCacheSize)
```
PragmaTwice: We can `GET A` after `FLUSHBLOCKCACHE` to ensure that everything works well?

sryanyuan: OK

git-hulk:
@sryanyuan [EraseUnRefEntries](https://github.com/facebook/rocksdb/blob/988357696dc7961789b32092a2e66effb7c2528e/cache/lru_cache.cc#L145) seems only flush evictable items(no reference). I find the [ClearCache](https://github.com/facebook/rocksdb/blob/988357696dc7961789b32092a2e66effb7c2528e/db/db_block_cache_test.cc#L910) implementation in rocksdb test case, but it might have the memory issue if the cache is large enough.

sryanyuan:
> @sryanyuan [EraseUnRefEntries](https://github.com/facebook/rocksdb/blob/988357696dc7961789b32092a2e66effb7c2528e/cache/lru_cache.cc#L145) seems only flush evictable items(no reference). I find the [ClearCache](https://github.com/facebook/rocksdb/blob/988357696dc7961789b32092a2e66effb7c2528e/db/db_block_cache_test.cc#L910) implementation in rocksdb test case, but it might have the memory issue if the cache is large enough.

I will investigate this further.

PragmaTwice (approved):

PragmaTwice:
Maybe we can merge it since the overall function looks good to me.

git-hulk:
@PragmaTwice Yes, the code is in good shape. But I'm not sure if this command is working as @sryanyuan expected since it seems only flush unreference(out-of-date) items from the implementation, maybe we could improve it later.

git-hulk (approved):

Issue #2805, opened 2025-02-26
TDigest: Implement `TDIGEST.MERGE` command


LindaSummer:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Refer: #2558
We should implement `TDIGEST.MERGE` command.

Redis command reference:
https://redis.io/docs/latest/commands/tdigest.merge/

### Solution

Depends on #2791 #2792 #2794 .

Add go integration tests for it.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

LindaSummer:
Hi Team,

Sorry for the pending issues in the past two months.
My baby was born, and my schedule was occupied.
Now I'm working on this command and will create a PR this week if everything goes well. 😊

Best Regards,
Edward

PR #3054, opened 2025-07-14
feat(tdigest): add `TDIGEST.MERGE` command implementation

LindaSummer:
# Issue
#2805

# Proposed Changes
- Add `TDIGEST.MERGE` command implementation
- Add go unit tests


src/commands/cmd_tdigest.cc
```diff
@@ -281,11 +284,103 @@ class CommandTDigestQuantile : public Commander {
...
+      return {Status::RedisParseErr, errWrongKeyword};
+    }
+
+    if (GET_OR_RET(parser.TakeStr()) != kOverrideArg) {
+      return {Status::RedisParseErr, errWrongKeyword};
+    }
+
+    options_.override_flag = true;
+
+    if (parser.Good()) {
```
git-hulk: Could we use a while loop to improve this part?

PragmaTwice: Yeah it can be replaced by a loop like
```cpp
while (parser.Good()) {
  if (parser.EqICase("compression")) {
    // ...
  } else if (parser.EqICase("override")) {
    // ...
  } else {
  }
}
```

PragmaTwice: cc @LindaSummer

LindaSummer: Hi @git-hulk & @PragmaTwice ,

Thanks very much for your suggestions! 😊
I have refactored the parser logic as suggested, and it becomes clearer now.

Best Regards,
Edward

PragmaTwice (approved):

git-hulk (approved):

mapleFU:
( I'm currently focus on other things, I might have time on Friday or later, you can merge this firstly. I would post-review this weekend)

PR #3061, opened 2025-07-18
feat(replication): implement WAIT based on replica acks

zhixinwen:
Add ack mechanism in WAIT and replication logic.

When a replica sync delta change from master, it would send a `replconf ack <sequence number>` to master.`FeedSlaveThread::staticReadCallback` will record the ack and unblock WAIT command. 

I wrote a simple loadtest script to test the setup with one master + slave running locally. The test has 120 concurrent workers, each calling SET with 100 bytes value:
```
Test Configuration:
  Type: set
  Duration: 1m0.000947872s
  Concurrency: 120
  Value Size: 100 bytes

Overall Statistics:
  Total Requests: 4303221
  Successful: 4303101 (100.00%)
  Failed: 120 (0.00%)
  Throughput: 71719.22 requests/second
```

If after each SET, we immediately follow a WAIT:
```
Test Configuration:
  Type: set-wait
  Duration: 1m0.000804367s
  Concurrency: 120
  Value Size: 100 bytes
  Wait Replicas: 1

Overall Statistics:
  Total Requests: 2237526
  Successful: 2237406 (99.99%)
  Failed: 120 (0.01%)
  Throughput: 44506.68 requests/second
```

When I first implement this feature, I tried to be smart and did the following (https://github.com/apache/kvrocks/commit/28baf29c5d8111a3210d23e978ffd2d2bd290171):
1. When there is no client waiting, replica ACK every 1s.
2. When there is client waiting, in `incrementBatchLoopCB`, master would send `_getack`.
3. When replica receives `_getack` , it would ack immediately.
This logic did not work well, I guess the issue is a client can send out too many `_getack` when it is blocked. And replica is busy with processing `_getack` instead of processing data, which makes the replication processing a lot slower:
```
Test Configuration:
  Type: set-wait
  Duration: 2m0.002154771s
  Concurrency: 120
  Value Size: 1024 bytes
  Wait Replicas: 1

Overall Statistics:
  Total Requests: 22996
  Successful: 21916 (95.30%)
  Failed: 1080 (4.70%)
  Throughput: 191.63 requests/second
```

I want to see if this mechanism is good enough for maintainers and I am open to suggestions to make it more efficient.


src/cluster/replication.cc
```diff
@@ -75,6 +75,11 @@ Status FeedSlaveThread::Start() {
 
   if (s) {
     t_ = std::move(*s);
+
+    // Re-enable the bufferevent and set up callbacks after detachment
+    auto bev = conn_->GetBufferEvent();
+    bufferevent_enable(bev, EV_READ);
+    bufferevent_setcb(bev, &FeedSlaveThread::staticReadCallback, nullptr, nullptr, this);
```
zhixinwen, line 82: In `this->loop`, we send data to file descriptor directly in a different thread. Here we can using `bufferevent` to read from the file descriptor in the callback. I think it is ok, because in the callback we only read, and in the loop, we only write. The two buffers should not step on each other. But I am not very familiar with this lib and C..

src/cluster/replication.cc
```diff
@@ -103,6 +108,52 @@ void FeedSlaveThread::checkLivenessIfNeed() {
   }
 }
 
+void FeedSlaveThread::staticReadCallback(bufferevent *bev, void *ctx) {
```
git-hulk, line 111: how about changing it to `replicaReadCallback`

src/cluster/replication.cc
```diff
@@ -563,12 +611,13 @@ ReplicationThread::CBState ReplicationThread::tryPSyncReadCB(bufferevent *bev) {
...
   repl_state_.store(kReplConnected, std::memory_order_relaxed);
   auto input = bufferevent_get_input(bev);
+  bool data_written = false;
   while (true) {
     switch (incr_state_) {
       case Incr_batch_size: {
         // Read bulk length
         UniqueEvbufReadln line(input, EVBUFFER_EOL_CRLF_STRICT);
-        if (!line) return CBState::AGAIN;
+        if (!line) goto AGAIN_LABEL;  // NOLINT
```
git-hulk: Could we avoid the goto statement?

src/cluster/replication.cc
zhixinwen: the name is taken: https://github.com/apache/kvrocks/pull/3061/files/6e91bf3edeef57e780ae2af56eb4b845f2b071c5#diff-19253031b45ceabae16926533fceb06c0c38f0b2c7e866e9a1a578434f3ca5a4R118

src/cluster/replication.cc
zhixinwen: done

src/cluster/replication.cc
```diff
@@ -601,14 +658,15 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
         if (!s.IsOK()) {
           error("[replication] CRITICAL - Failed to write batch to local, {}. batch: 0x{}", s.Msg(),
                 util::StringToHex(batch.Data()));
-          return CBState::RESTART;
+          return {CBState::RESTART, data_written};
```
git-hulk: I think we don't need to add an extra method and return value, just add a function like `sendReplConfAck` after applying the write batch.

zhixinwen: get it, I updated it here: https://github.com/apache/kvrocks/pull/3061/commits/9d44339c4d130923f84716bf3bddb73e6ba3846d.
I don't think we should send ack after applying each write batch though, because if there are several batches in one read callback, it is more efficient to write all of the batches and ack once.

PragmaTwice (commented):
Looks good. Would you like to test if it works well in TLS-enabled replication?

src/cluster/replication.cc
```diff
@@ -603,6 +667,7 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
                 util::StringToHex(batch.Data()));
           return CBState::RESTART;
         }
+        data_written = true;
```
git-hulk, line 670: I'm not sure why do we need to call `sendReplConfAck` in other places? I think it would be good to send the ack after applying the write batch only. 

What I mean is to remove `data_written` and `sendReplConfAck` in other places, and just call `sendReplConfAck` here. Can help to correct me if I missed anything.

zhixinwen:
> Looks good. Would you like to test if it works well in TLS-enabled replication?

I saw the unit test already tested TLS case, https://github.com/apache/kvrocks/actions/runs/16404312734/job/46351253953?pr=3061 and there is TLS test for replica: https://github.com/apache/kvrocks/blob/unstable/tests/gocase/tls/tls_test.go#L142. Is there anything additional I should test?

PragmaTwice:
Yeah but it is skipped via `T.skip` due to some historical reason. Hence I thought that maybe we can do a manual test to ensure it is not broken because of these changes.

zhixinwen:
> Yeah but it is skipped via `T.skip` due to some historical reason. Hence I thought that maybe we can do a manual test to ensure it is not broken because of these changes.

ah sorry, I did not see T.skip. I will report back once I finish a TLS test

src/cluster/replication.cc
zhixinwen: It is possible in one `incrementBatchLoopCB` call, several `rocksdb::WriteBatch` can be constructed from existing data in bufferevent. In my local test env, there can can anything from 3 to 22 write batches in a single call with an average of 15 write batches. 

It is unnecessary and potentially wasteful to send an ack for each of them. Instead, we can just write all the batches to RocksDB (assuming write to RocksDB is fast) and send ack once.

One thing I would potentially do in a following PR is to make this behavior configurable. e.g. user may specify ack on every send / ack on every n send or keep the current behavior. But I would want to benchmark the throughput first. For now, I think it is a reasonable default behavior.

git-hulk: I see. Makes sense.

zhixinwen:
> Yeah but it is skipped via `T.skip` due to some historical reason. Hence I thought that maybe we can do a manual test to ensure it is not broken because of these changes.

Verified manually the TLS replication works.
Master config:
```
# Master configuration with TLS
port 6666
tls-port 6667
tls-cert-file tests/gocase/tls/cert/server.crt
tls-key-file tests/gocase/tls/cert/server.key
tls-ca-cert-file tests/gocase/tls/cert/ca.crt
tls-auth-clients no
tls-protocols "TLSv1.2"
tls-ciphers DEFAULT
dir /tmp/kvrocks-master-data
```

Replica config:
```
# Slave configuration with TLS
port 6668
tls-port 6669
tls-cert-file tests/gocase/tls/cert/server.crt
tls-key-file tests/gocase/tls/cert/server.key
tls-ca-cert-file tests/gocase/tls/cert/ca.crt
tls-auth-clients no
tls-protocols "TLSv1.2"
tls-ciphers DEFAULT

# Enable TLS for replication
tls-replication yes

# Master connection (using TLS port)
slaveof 127.0.0.1 6667

dir /tmp/kvrocks-slave-data
```

port 6667 can only be accessed with tls:
```
coder@zwen-dev:~/kvrocks$ /usr/bin/redis-cli --tls -p 6667 --cacert tests/gocase/tls/cert/ca.crt PING
PONG
coder@zwen-dev:~/kvrocks$ /usr/bin/redis-cli  -p 6667  PING
Error: Connection reset by peer
```

git-hulk (approved):

PragmaTwice (approved):

PragmaTwice (commented):
Generally looks good to me.

The last question from me, I'm wondering if the replication has a noticeable slow-down due to a new blocking IO (`sendString`, take longer for one round in `incrementBatchLoopCB` since we have IO in the end?) and more network traffic?

zhixinwen:
> Generally looks good to me.
> 
> The last question from me, I'm wondering if the replication has a noticeable slow-down due to a new blocking IO (`sendString`, take longer for one round in `incrementBatchLoopCB` since we have IO in the end?) and more network traffic?

TLDR:
Good point, you are right it would have a significant impact (40%) on replication throughput. 
I will follow up with a PR to make the ack behavior configurable, so users can balance between WAIT latency and replication throughput themselves.

Full version:
I measure the replication throughput difference by comparing master and slave sequence difference every 5s when I am sending approximately 310MB/s traffic to master (80k QPS with 4kb payload). I notice the `unstable` branch lag increases by ~84k sequence  every 5s (66MB/S) and this branch lag increases by ~220k sequence every 5s (171 MB/s). Therefore, the current unstable branch has a rough replication throughput of 240MB and this branch has throughput of roughly 140MB. 

This overhead is beyond my expectation, it may take me a some time to figure out why the perf hit is so bad (btw, I dont think `sendString` is blocking because it just adds to the buffer?). Meanwhile, I think it makes sense to make the ack behavior configurable. For our usecase, we want to call WAIT on every write for lossless-data replication, so we want faster ack. I understand most of the users don't need this, so an ack every second or a few dozens of updates is good enough for them.

PragmaTwice:
Thank you for replying. An issue is created to track this.

Several points from mine mind:
1. I think we can make the frequency of `sendReplConfAck` configurable, i.e. maybe we can make it called in every N rounds of `incrementBatchLoopCB` rather than in every `incrementBatchLoopCB`. (Or, we can introduce a timer so that "every N rounds" can be replaced by "every N seconds/minutes" and then `WAIT` should be more expectable.)
2. Maybe we can do a profiling to help us pinpoint the performance bottlenecks.
3. Maybe we can exclude this commit from release 2.13.0 so that we can improve it without considering to avoid breaking changes? And from me it's fine to include b81434f38b470df8a65bf9f987fdcbd7c04cf07a since there's much less performance impact and didn't introduce new kvrocks-specific commands (`replconf ack`). cc @git-hulk @LindaSummer

git-hulk:
> I think we can make the frequency of sendReplConfAck configurable, i.e. maybe we can make it called in every N rounds of incrementBatchLoopCB rather than in every incrementBatchLoopCB. (Or, we can introduce a timer so that "every N rounds" can be replaced by "every N seconds/minutes" and then WAIT should be more expectable.)

Another approach is like the Redis behavior, which sends `REPLCONF ACK` periodically(every second) in replica, and the master can also send `REPLCONF GETACK` to ask replicas reply the offset ASAP after the WAIT command.


> Maybe we can exclude this commit from release 2.13.0 so that we can improve it without considering to avoid breaking changes? And from me it's fine to include https://github.com/apache/kvrocks/commit/b81434f38b470df8a65bf9f987fdcbd7c04cf07a since there's much less performance impact and didn't introduce new kvrocks-specific commands (replconf ack). cc @git-hulk @LindaSummer

+1 to include https://github.com/apache/kvrocks/commit/b81434f38b470df8a65bf9f987fdcbd7c04cf07a. Another potential issue is the WAIT command will be pending forever if any replicas don't support the `REPLCONF ACK`.

LindaSummer:
> Thank you for replying. An issue is created to track this.
> 
> Several points from mine mind:
> 
>     1. I think we can make the frequency of `sendReplConfAck` configurable, i.e. maybe we can make it called in every N rounds of `incrementBatchLoopCB` rather than in every `incrementBatchLoopCB`. (Or, we can introduce a timer so that "every N rounds" can be replaced by "every N seconds/minutes" and then `WAIT` should be more expectable.)
> 
>     2. TODO
> 
>     3. Maybe we can exclude this commit from release 2.13.0 so that we can improve it without considering to avoid breaking changes? And from me it's fine to include [b81434f](https://github.com/apache/kvrocks/commit/b81434f38b470df8a65bf9f987fdcbd7c04cf07a) since there's much less performance impact and didn't introduce new kvrocks-specific commands (`replconf ack`). cc @git-hulk @LindaSummer

Hi @PragmaTwice , @git-hulk ,

Got it! 😊

I'll exclude this commit and include b81434f in release 2.13.0.

Best Regards,
Edward

zhixinwen:
This PR is a MVP to let KVRocks support synch replication without data loss. Want to share some benchmark result.

Setup: One master and one replica running on 64 core machines. Ping latency between the machines is 0.175ms. Each instance has RocksDB and log dir on its own disk (450MB/s write speed). Rest of the config is default. 

Base case: set with 4KB payload with rocksdb.write_options.sync=false on both master and replica
```
Throughput: 83129.45 requests/second (325 MB/s)
Latency Statistics:
  Average: 11.911131ms
  P50: 11.514329ms
  P95: 14.990656ms
  P99: 29.964993ms
  P99.9: 61.902948ms
  Min: 76.214µs
  Max: 626.700766ms
```

set+wait for every call with 4KB payload and rocksdb.write_options.sync=false 
```
Throughput: 40030.77 requests/second (156 MB/s)
Latency Statistics (successful requests only):
  Average: 24.727204ms
  P50: 24.916566ms
  P95: 27.971368ms
  P99: 29.288314ms
  P99.9: 31.054725ms
  Min: 545.906µs
  Max: 79.467267ms
```


set with 4KB payload with rocksdb.write_options.sync=true on both master and replica
```
Throughput: 20732.77 requests/second (80 MB/s)
Latency Statistics:
  Average: 47.752568ms
  P50: 40.683806ms
  P95: 62.193638ms
  P99: 291.432609ms
  P99.9: 502.826896ms
  Min: 161.053µs
  Max: 577.052582ms
```

set+wait with 4KB payload with rocksdb.write_options.sync=true on both master and replica (This setting would guarantee lossless data)
```
Throughput: 17783.96 requests/second (70 MB/s)

Latency Statistics
  Average: 55.734346ms
  P50: 47.073675ms
  P95: 93.116192ms
  P99: 275.485393ms
  P99.9: 448.5747ms
  Min: 1.077095ms
  Max: 548.264188ms
```

---
Update:
In set+wait with 4KB payload with rocksdb.write_options.sync=true on both master and replica (This setting would guarantee lossless data) case, we notice the CPU is becoming bottleneck. We increased worker count to 60 and had 125Mb/s throughput.

zhixinwen:
> +1 to include [b81434f](https://github.com/apache/kvrocks/commit/b81434f38b470df8a65bf9f987fdcbd7c04cf07a). Another potential issue is the WAIT command will be pending forever if any replicas don't support the `REPLCONF ACK`.

@PragmaTwice @git-hulk I think it would be an issue if you release [b81434f](https://github.com/apache/kvrocks/commit/b81434f38b470df8a65bf9f987fdcbd7c04cf07a) first then ack based replication later.

Once [b81434f](https://github.com/apache/kvrocks/commit/b81434f38b470df8a65bf9f987fdcbd7c04cf07a) is released, people can begin to use WAIT command. And when they upgrade to the version with ack, they may see the issue you mentioned during the upgrade time.

If [b81434f](https://github.com/apache/kvrocks/commit/b81434f38b470df8a65bf9f987fdcbd7c04cf07a) is not released, there would be noone able to use WAIT anyway (unless they are using unstable), so the issue above should not exist.

When do you plan to release 2.13.0, I can try to reduce the impact before the release.

PragmaTwice:
> Once [b81434f](https://github.com/apache/kvrocks/commit/b81434f38b470df8a65bf9f987fdcbd7c04cf07a) is released, people can begin to use WAIT command. And when they upgrade to the version with ack, they may see the issue you mentioned during the upgrade time.

Hmm sounds reasonable. We are planning to start the release process of version 2.13.0 in this or the next week. We can postpone it by a week or two depending on the situation, but as a community project and we don’t want to put too much pressure on community members by setting some deadlines : )

We're also keen to hear your thoughts. Do you feel strongly about this feature being included in 2.13.0? I think keeping the replication throughput change within 10% would be ideal, as it shouldn't significantly impact existing users. WDYT?

cc @git-hulk

zhixinwen:
> > Once [b81434f](https://github.com/apache/kvrocks/commit/b81434f38b470df8a65bf9f987fdcbd7c04cf07a) is 
> We're also keen to hear your thoughts. Do you feel strongly about this feature being included in 2.13.0? I think keeping the replication throughput change within 10% would be ideal, as it shouldn't significantly impact existing users. WDYT?
> 
> cc @git-hulk

It is 100% fine for me to not include this feature in 2.13.0. My team will be directly bechmarking against `unstable` branch.

PragmaTwice:
> It is 100% fine for me to not include this feature in 2.13.0. My team will be directly bechmarking against `unstable` branch.

Thank you! Then I think we can exclude both b81434f38b470df8a65bf9f987fdcbd7c04cf07a and 166d0886b0b2a479121c283ef584aeb5b0240523 from 2.13.0 to prevent the emergence of users of `WAIT`, which might impede future changes and optimizations. cc @LindaSummer

zhixinwen:
> Another approach is like the Redis behavior, which sends `REPLCONF ACK` periodically(every second) in replica, and the master can also send `REPLCONF GETACK` to ask replicas reply the offset ASAP after the WAIT command.
> 

This approach did not work well as I mentioned in the PR description. I think in order to make it work, master should send `REPLCONF ACK` once only and master should only be sent after it has sent the change that would unblock a WAIT. It may add other overhead.  Let me know if you have a better way  @git-hulk

PragmaTwice:
Let's move the further discussion to #3070 since this PR is merged.

PR #3069, opened 2025-07-22
build: disable LTO in debug mode

PragmaTwice:
LTO is usually time-consuming, we can disable it in debug mode.


mapleFU (approved):
(Why not keep `ENABLE_IPO` ? If some user disable this when compiling, would user need change the flag here?)

PragmaTwice:
LTO is a more precise name here rather than IPO (it should be a mistake of CMake though).

git-hulk (approved):

PR #3072, opened 2025-07-22
feat(ts): initialize metadata and subkey encoding 

yezhizi:
Proposal: #3044


LindaSummer:
Hi @yezhizi ,

I have a bit of advice that we'd better try to implement something like `TS.INFO` and unit tests to just verify the metadata like the demo in [redis doc](https://redis.io/docs/latest/develop/data-types/timeseries/#create-a-time-series).

This will help us find potential problems before the next implementation.

It's just an optional suggestion. It's also ok to make the unit test in the next stage. 😊

Best Regards,
Edward

src/storage/redis_metadata.cc
```diff
@@ -537,3 +537,25 @@ rocksdb::Status TDigestMetadata::Decode(Slice *input) {
...
+    return s;
+  }
+  GetFixed64(input, &retention_time);
+  GetFixed64(input, &chunk_size);
+  GetFixed8(input, reinterpret_cast<uint8_t *>(&chunk_type));
+  GetFixed8(input, reinterpret_cast<uint8_t *>(&duplicate_policy));
+  GetSizedString(input, &source_key);
+
+  return rocksdb::Status::OK();
+}
```
```suggestion
}

```

Would you like to add a newline in the end of the file?

src/storage/redis_metadata.h
```diff
@@ -364,3 +365,36 @@ class TDigestMetadata : public Metadata {
...
+  explicit TimeSeriesMetadata(uint64_t chunk_size, uint64_t retention_time = 0, bool generate_version = true)
+      : Metadata(kRedisTimeSeries, generate_version),
+        retention_time(retention_time),
+        chunk_size(chunk_size),
+        chunk_type(ChunkType::UNCOMPRESSED),
+        duplicate_policy(TSDuplicatePolicy::BLOCK) {}
+
+  void Encode(std::string *dst) const override;
+  rocksdb::Status Decode(Slice *input) override;
+};
```
PragmaTwice: Same as above.

src/storage/redis_metadata.h
```diff
@@ -364,3 +365,36 @@ class TDigestMetadata : public Metadata {
...
 };
+
+enum class TSDuplicatePolicy : uint8_t {
+  BLOCK = 0,
+  FIRST = 1,
+  LAST = 2,
+  MIN = 3,
+  MAX = 4,
+  SUM = 5,
+};
```
PragmaTwice: Also make this as a nested class like `ChunkType`?

src/storage/redis_metadata.h
```diff
@@ -364,3 +365,36 @@ class TDigestMetadata : public Metadata {
...
+  enum class ChunkType : uint8_t {
+    UNCOMPRESSED = 0,
+    COMPRESSED = 1,
+  };
+
+  uint64_t retention_time;
+  uint64_t chunk_size;
+  ChunkType chunk_type;
+  TSDuplicatePolicy duplicate_policy;
+  Slice source_key;
```
PragmaTwice: `Slice` has no ownership. It's like a `string_view`.

Here I think we can hold an owned string so that we don't need to worry about lifetime.

src/storage/redis_metadata.cc
```diff
@@ -537,3 +537,25 @@ rocksdb::Status TDigestMetadata::Decode(Slice *input) {
...
+  PutFixed8(dst, static_cast<uint8_t>(chunk_type));
+  PutFixed8(dst, static_cast<uint8_t>(duplicate_policy));
+  PutSizedString(dst, source_key);
+}
+
+rocksdb::Status TimeSeriesMetadata::Decode(Slice *input) {
+  if (auto s = Metadata::Decode(input); !s.ok()) {
+    return s;
+  }
+  GetFixed64(input, &retention_time);
```
PragmaTwice, line 560: Should check if the remain length is long enough before the `GetFixed64`.

src/types/redis_timeseries.h
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+#pragma once
+
+#include <cstdint>
+
+#include "storage/redis_db.h"
+#include "storage/redis_metadata.h"
+
+enum class TSubkeyType : uint8_t {
```
PragmaTwice:
```suggestion
enum class TSSubkeyType : uint8_t {
```

src/types/redis_timeseries.h
```diff
@@ -0,0 +1,103 @@
...
+      : aggregator(aggregator),
+        bucket_duration(bucket_duration),
+        alignment(alignment),
+        latest_bucket_idx(latest_bucket_idx) {}
+
+  void Encode(std::string *dst) const;
+  rocksdb::Status Decode(Slice *input);
+};
+
+namespace redis {
```
PragmaTwice: Maybe we can also make `TSDownStreamMeta`, `TSAggregatorType` and `TSSubkeyType` inside the namespace.

src/types/redis_timeseries.h
```diff
@@ -0,0 +1,103 @@
...
+
+  [[nodiscard]] std::string Encode() const;
+
+ private:
+  bool slot_id_encoded_;
+};
+
+class TimeSeries : public SubKeyScanner {
+ public:
+  // TODO:
```
PragmaTwice: Would you like to add a constructor to it?

src/types/redis_timeseries.h
```diff
@@ -0,0 +1,103 @@
...
+ private:
+  rocksdb::ColumnFamilyHandle *timeseries_cf_handle_;
```
PragmaTwice: Maybe we can remove it in this PR since now it is not so useful.

src/types/redis_timeseries.h
```diff
@@ -0,0 +1,103 @@
...
+
+  std::string internalKeyFromChunkID(const std::string &ns_key, const TimeSeriesMetadata &metadata, uint64_t ts) const;
```
PragmaTwice:
```suggestion
  std::string internalKeyFromChunkID(const std::string &ns_key, const TimeSeriesMetadata &metadata, uint64_t id) const;
```

src/types/redis_timeseries.cc
```diff
@@ -0,0 +1,134 @@
...
+  PutFixed64(dst, bucket_duration);
+  PutFixed64(dst, alignment);
+  PutFixed64(dst, latest_bucket_idx);
+  PutFixed8(dst, static_cast<uint8_t>(u64_auxs.size()));
+  for (const auto &aux : u64_auxs) {
+    PutFixed64(dst, aux);
+  }
+  for (const auto &aux : f64_auxs) {
+    PutDouble(dst, aux);
+  }
```
PragmaTwice, line 34: `size of u64_auxs` always equal to `size of f64_auxs`. Fine if it is here.

PragmaTwice, line 37: Should we also add a size for `f64_auxs`? So that we can make `TSDownStreamMeta` extensible, e.g. we can add new fields to this class.

src/types/redis_timeseries.cc
```diff
@@ -0,0 +1,134 @@
...
+  for (uint8_t i = 0; i < u64_auxs_size; i++) {
+    uint64_t aux = 0;
+    GetFixed64(input, &aux);
+    u64_auxs.push_back(std::move(aux));
+  }
+  while (input->size() > 0) {
+    double aux = NAN;
+    GetDouble(input, &aux);
+    f64_auxs.push_back(std::move(aux));
+  }
```
PragmaTwice: Same as above.

src/types/redis_timeseries.h
```diff
@@ -0,0 +1,103 @@
...
+
+class TSRevLabelKey {
+ public:
+  Slice ns;
+  uint16_t slot_id;
+  Slice label_key;
+  Slice label_value;
+  Slice user_key;
+
+  TSRevLabelKey(Slice ns_key, Slice label_key, Slice label_value, bool slot_id_encoded);
```
PragmaTwice:
```suggestion
  TSRevLabelKey(Slice namespace, Slice label_key, Slice label_value, bool slot_id_encoded);
```

`ns_key` usually means `ns | user_key` in Kvrocks, here I think we only care about `ns`.

src/types/redis_timeseries.cc
```diff
@@ -0,0 +1,134 @@
...
+  ns = Slice(ns_key.data(), namespace_size);
+  ns_key.remove_prefix(namespace_size);
+
+  if (slot_id_encoded_) {
+    GetFixed16(&ns_key, &slot_id);
+  }
+  user_key = ns_key;
+}
+
+std::string TSRevLabelKey::Encode() const {
```
PragmaTwice, line 73: Since it is in the new CF, I'm wondering if we can add an enum prefix like `TSSubkeyType` in the subkey CF, so that we can put new keys into this new CF in the future.

src/types/redis_timeseries.h
```diff
@@ -0,0 +1,103 @@
...
+
+  [[nodiscard]] std::string Encode() const;
+
+ private:
+  bool slot_id_encoded_;
```
PragmaTwice: Seems we should not bring `slot_id` into `TSRevLabelKey`, otherwise we can only search keys in one slot id.

WDYT?

PragmaTwice (commented):
The rest looks good. Thank you for your effort : )

src/types/redis_timeseries.h
yezhizi: But it seems we still need to provide the `user_key` as input, because this key is encoded like this:
```
+-----------+--------------+----------------+------------+------------------+--------------+-------------+
|     ns    | cluster slot | label_key_size |  label_key | label_value_size |  label_value |  user_key    | => null
| (1+Xbytes)|   (2byte)    |   (4byte)      |   (Ybyte)  |   (4byte)        |   (Zbyte)    |   (Kbyte)    |
+-----------+--------------+----------------+------------+------------------+--------------+--------------+
```
We need to use labels to index the specific user keys.

src/types/redis_timeseries.h
yezhizi: That makes sense, but  not so sure because I'm not familiar with cluster mode.

src/types/redis_timeseries.h
PragmaTwice: Ahh got your point.

So here `ns_key` is what it means. However I still suggest to split to multiple arguments like `ns` and `user_key`, since in cluster mode slot_id is included in `ns_key`, and as here mentioned https://github.com/apache/kvrocks/pull/3072#discussion_r2224202179 we may need to remove it.

src/types/redis_timeseries.h
PragmaTwice: We can first disable search with labels (e.g. like `ts.mget`) in cluster mode.

src/types/redis_timeseries.cc
yezhizi: Sounds good! So is it okay to encode the keys this way?
```
+----------+-----------+----------------+------------+------------------+--------------+---------+
| KeyType  |     ns    | label_key_size |  label_key | label_value_size |  label_value |  key    | => null
| (1byte)  | (1+Xbytes)|   (4byte)      |   (Ybyte)  |   (4byte)        |   (Zbyte)    | (Kbyte) |
|----------+-----------+----------------+------------+------------------+--------------+---------+
```

src/types/redis_timeseries.h
yezhizi: OK! Got it! (I didn't see the comment above due to message delays.)

src/types/redis_timeseries.cc
yezhizi: BTW, where should I put this key type enum? Can we temporarily put it in `redis_timeseries.h?`

PragmaTwice: Maybe the key type can be put after `ns`.

> Can we temporarily put it in redis_timeseries.h?

Yeah we can put it here.

src/storage/redis_metadata.h
```diff
@@ -364,3 +365,36 @@ class TDigestMetadata : public Metadata {
...
+
+  explicit TimeSeriesMetadata(uint64_t chunk_size, uint64_t retention_time = 0, bool generate_version = true)
```
mapleFU: I don't think we need any default value here, since this should be filled by user?

yezhizi: OK! I will fix it.

PragmaTwice: We should provide a default ctor so that it can be parsed via `Decode`.

src/storage/redis_metadata.h
```diff
@@ -364,3 +365,45 @@ class TDigestMetadata : public Metadata {
...
+  uint64_t retention_time;
+  uint64_t chunk_size;
+  ChunkType chunk_type;
+  TSDuplicatePolicy duplicate_policy;
+  std::string source_key;
+
+  explicit TimeSeriesMetadata(bool generate_version = true)
+      : Metadata(kRedisTimeSeries, generate_version),
+        retention_time(0),
+        chunk_size(1024),
```
```suggestion
        chunk_size(0),
```

src/storage/redis_metadata.h
```diff
@@ -364,3 +365,45 @@ class TDigestMetadata : public Metadata {
...
 };
+
+class TimeSeriesMetadata : public Metadata {
+ public:
+  enum class ChunkType : uint8_t {
+    UNCOMPRESSED = 0,
+    COMPRESSED = 1,
+  };
+
+  enum class TSDuplicatePolicy : uint8_t {
```
```suggestion
  enum class DuplicatePolicy : uint8_t {
```

src/storage/redis_metadata.h
```diff
@@ -364,3 +365,45 @@ class TDigestMetadata : public Metadata {
...
+    LAST = 2,
+    MIN = 3,
+    MAX = 4,
+    SUM = 5,
+  };
+
+  uint64_t retention_time;
+  uint64_t chunk_size;
+  ChunkType chunk_type;
+  TSDuplicatePolicy duplicate_policy;
```
```suggestion
  DuplicatePolicy duplicate_policy;
```

src/storage/redis_metadata.h
```diff
@@ -364,3 +365,45 @@ class TDigestMetadata : public Metadata {
...
+  std::string source_key;
+
+  explicit TimeSeriesMetadata(bool generate_version = true)
+      : Metadata(kRedisTimeSeries, generate_version),
+        retention_time(0),
+        chunk_size(0),
+        chunk_type(ChunkType::UNCOMPRESSED),
+        duplicate_policy(TSDuplicatePolicy::BLOCK) {}
```
```suggestion
        duplicate_policy(DuplicatePolicy::BLOCK) {}
```

src/storage/redis_metadata.h
```diff
@@ -364,3 +365,45 @@ class TDigestMetadata : public Metadata {
...
+  TimeSeriesMetadata(uint64_t retention_time, uint64_t chunk_size, ChunkType chunk_type,
+                     TSDuplicatePolicy duplicate_policy, bool generate_version = true)
```
```suggestion
                     DuplicatePolicy duplicate_policy, bool generate_version = true)
```

PragmaTwice (approved):

Issue #3070, opened 2025-07-22
Improve replication performance after introducing replica ack messages


PragmaTwice:
> The last question from me, I'm wondering if the replication has a noticeable slow-down due to a new blocking IO (`sendString`, take longer for one round in `incrementBatchLoopCB` since we have IO in the end?) and more network traffic?

TLDR:
Good point, you are right it would have a significant impact (40%) on replication throughput. 
I will follow up with a PR to make the ack behavior configurable, so users can balance between WAIT latency and replication throughput themselves.

Full version:
I measure the replication throughput difference by comparing master and slave sequence difference every 5s when I am sending approximately 310MB/s traffic to master (80k QPS with 4kb payload). I notice the `unstable` branch lag increases by ~84k sequence  every 5s (66MB/S) and this branch lag increases by ~220k sequence every 5s (171 MB/s). Therefore, the current unstable branch has a rough replication throughput of 240MB and this branch has throughput of roughly 140MB. 

This overhead is beyond my expectation, it may take me a some time to figure out why the perf hit is so bad (btw, I dont think `sendString` is blocking because it just adds to the buffer?). Meanwhile, I think it makes sense to make the ack behavior configurable. For our usecase, we want to call WAIT on every write for lossless-data replication, so we want faster ack. I understand most of the users don't need this, so an ack every second or a few dozens of updates is good enough for them.

_Originally posted by @zhixinwen in https://github.com/apache/kvrocks/issues/3061#issuecomment-3100948973_

PragmaTwice:
I'm not familiar to the code of replication part, but I can try to propose a solution for this problem.

Introducing two configuration options:
```
replication_ack_duration (in milliseconds/seconds?)
replication_ack_seq_threshold
```

Algorithm:
- In `IncrementBatchLoop`
  - if `current_seq - last_reported_seq > replication_ack_seq_threshold`:
    - ACK `current_seq`
    - update `last_reported_seq`
- In the ack timer (triggered in every `replication_ack_duration`)
  - if `last_reported_seq != current_seq`:
    - ACK `current_seq`
    - update `last_reported_seq`

Current implementation is:
- replication_ack_duration = inf
- replication_ack_seq_threshold = 0

Remaining problems:
- How to tune these two values? (performance v.s. `WAIT` latency)
- Should we combine it with active fetching (`GETACK`)?
- What is the current behavior? e.g. how many `ACK` is sent in 1s?

cc @zhixinwen @git-hulk

[...]

PR #3075, opened 2025-07-24
feat(replication): implement _getack for quick WAIT response

zhixinwen:
This should solve: https://github.com/apache/kvrocks/issues/3070

Key changes:
- Use mutimap to save wait_context, so given a seq number, it is faster to check which waits would be unblocked.
- Use shared lock to reduce lock contention.
- When master send change to replica, it would check if any wait can be unblocked. If yes, it would send a `replconf getack` right after sending out the change.
- `last_replconf_getack_seq_` is kept by master, so it would send unnecessary replconf getack` for the same wait_context
- When there is no wait, replica send ack every 1s.

Test shows this mechanism has similar replication speed compared to the code before https://github.com/apache/kvrocks/pull/3061 was committed. It also shows better performance than https://github.com/apache/kvrocks/pull/3061 when every write request is followed by a `WAIT`.

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -150,7 +150,6 @@ void FeedSlaveThread::readCallback(bufferevent *bev, [[maybe_unused]] void *ctx)
   commands->clear();
 
   if (max_seq != 0) {
-    info("[replication] max_seq: {}", max_seq);
     ack_seq_.store(max_seq);
 
     // Wake up any WAIT connections that might be waiting for this sequence
@@ -219,8 +218,6 @@ void FeedSlaveThread::loop() {
     // Check if this change would unblock any WAIT command
     auto largest_unblockable_seq = srv_->LargestTargetSeqToWakeup(batch.sequence);
     if (largest_unblockable_seq > last_replconf_getack_seq_) {
-      info("[replication] largest_unblockable_seq: {}, last_replconf_getack_seq_: {}", largest_unblockable_seq,
-           last_replconf_getack_seq_);
       // Send replconf getack to the slave to get acknowledgment
       auto s = util::SockSend(conn_->GetFD(), redis::BulkString("replconf getack"), conn_->GetBufferEvent());
       if (!s.IsOK()) {
@@ -634,7 +631,6 @@ void ReplicationThread::sendReplConfAck(bufferevent *bev, bool force) {
 
   // If force is true, always send ack. Otherwise, check if it has been 1s from last ack
   if (force || (now - last_ack_time_secs_) >= 1) {
-    info("[replication] send replconf ack, seq: {}", storage_->LatestSeqNumber());
     SendString(bev, redis::ArrayOfBulkStrings({"replconf", "ack", std::to_string(storage_->LatestSeqNumber())}));
     last_ack_time_secs_ = now;
   }
@@ -658,7 +654,6 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
         }
         incr_bulk_len_ = line.length > 0 ? std::strtoull(line.get() + 1, nullptr, 10) : 0;
         if (incr_bulk_len_ == 0) {
-          error("[replication] Invalid increment data size");
           return CBState::RESTART;
         }
         incr_state_ = Incr_batch_data;
@@ -676,7 +671,6 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
         const char *bulk_data =
             reinterpret_cast<const char *>(evbuffer_pullup(input, static_cast<ssize_t>(incr_bulk_len_ + 2)));
         std::string bulk_string = std::string(bulk_data, incr_bulk_len_);
-        info("[replication] bulk_string: {}", bulk_string);
         evbuffer_drain(input, incr_bulk_len_ + 2);
         incr_state_ = Incr_batch_size;
 
@@ -690,7 +684,6 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
         }
 
         if (bulk_string == "replconf getack") {
-          info("[replication] replconf getack command received, force ack");
           // master would send the replconf getack command to the master to get acknowledgment
           // don't write replconf getack to db here.
           force_ack = true;
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -218,9 +218,15 @@ void FeedSlaveThread::loop() {
     // Check if this change would unblock any WAIT command
     auto largest_unblockable_seq = srv_->LargestTargetSeqToWakeup(batch.sequence);
     if (largest_unblockable_seq > last_replconf_getack_seq_) {
-      last_replconf_getack_seq_ = largest_unblockable_seq;
       // Send replconf getack to the slave to get acknowledgment
-      SendString(conn_->GetBufferEvent(), redis::BulkString("replconf getack"));
+      auto s = util::SockSend(conn_->GetFD(), redis::BulkString("replconf getack"), conn_->GetBufferEvent());
+      if (!s.IsOK()) {
+        error("Write error while sending replconf getack to slave: {}. batches: 0x{}", s.Msg());
+        Stop();
+        return;
+      } else {
+        last_replconf_getack_seq_ = largest_unblockable_seq;
+      }
     }
 
     while (!IsStopped() && !srv_->storage->WALHasNewData(curr_seq)) {
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -154,6 +154,16 @@ void FeedSlaveThread::readCallback(bufferevent *bev, [[maybe_unused]] void *ctx)
   }
 }
 
+bool FeedSlaveThread::shouldSendGetAck(rocksdb::SequenceNumber seq) {
+  rocksdb::SequenceNumber largest_unblockable_seq = srv_->LargestTargetSeqToWakeup(seq);
+  if (largest_unblockable_seq > last_getack_seq_) {
+    last_getack_seq_ = largest_unblockable_seq;
+    return true;
+  }
+
+  return false;
+}
+
 void FeedSlaveThread::loop() {
   // is_first_repl_batch was used to fix that replication may be stuck in a dead loop
   // when some seqs might be lost in the middle of the WAL log, so forced to replicate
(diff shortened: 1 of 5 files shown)


PragmaTwice:
I'll give a review on the next week.

zhixinwen:
https://github.com/zhixinwen/kvrocks/pull/2 is stacked on this PR.

src/server/server.h
```diff
@@ -380,7 +381,7 @@ class Server {
   std::atomic<uint64_t> total_clients_{0};
 
   // slave
-  std::mutex slave_threads_mu_;
+  std::shared_mutex slave_threads_mu_;
```
PragmaTwice: Would you like to split changes related to `slave_threads_mu_` to a new PR (e.g. with a title like `optimize slave threads via rwlock`)?

It seems not so related to the `WAIT` command?

PragmaTwice: These changes are quite straightforward, but many lines are involved so reviewers may be overwhelmed and may not able to concentrate on WAIT changes 🤔

src/cluster/replication.cc
```diff
@@ -203,6 +206,21 @@ void FeedSlaveThread::loop() {
         Stop();
         return;
       }
+
+      // Check if this change would unblock any WAIT command
+      auto largest_unblockable_seq = srv_->LargestTargetSeqToWakeup(batch.sequence);
+      if (largest_unblockable_seq > last_replconf_getack_seq_) {
+        // Send replconf getack to the slave to get acknowledgment
+        auto s = util::SockSend(conn_->GetFD(), redis::BulkString("replconf getack"), conn_->GetBufferEvent());
```
PragmaTwice: Seems here we use a relatively casual form `redis::BulkString("replconf getack")`. I'm wondering if it can hurt the extensibility and make some confusion to people, e.g. people may think that there is a subcommand `replconf getack`, but actually here it is just a "placeholder" rather than a real "command".

PragmaTwice: I'm having concern since once we decide we pass `replconf getack` in this way, we can hardly change it since some users may partially upgrade the version of kvrocks in a cluster so compatibility matters.

zhixinwen: I don't think there is a good way to pass actual command to the replication instance in the current architecture.
I can change `replconf getack` to `_getack` so it is less confusing.

In terms of compatibility, I don't see a good way either. If we do need to make breaking change, I feel we would need to add a handshake step to determine the protocol version.  I think the current replication logic has the same issue, you have some extra flexibility to pass in some more flags, but it is hard to change the flow.

Let me know if you have better ideas.

src/server/server.h
zhixinwen: reverted https://github.com/apache/kvrocks/pull/3075/commits/d52e799cc800cf4467e814b93df711f067b7939f

src/cluster/replication.cc
PragmaTwice: Yeah I know that it seems no best way to do this.

If you dont mind, I think it is better to replace it with `ArrayOfBulkStrings`? So that it looks consistent with other places.

PragmaTwice: Ahh seems there is already a "ping". So maybe a bulk string "_getack" looks good. No need to change to "ArrayOfBulkStrings".

zhixinwen: sure

src/cluster/replication.cc
```diff
@@ -203,6 +206,21 @@ void FeedSlaveThread::loop() {
...
         return;
       }
+
+      // Check if this change would unblock any WAIT command
+      auto largest_unblockable_seq = srv_->LargestTargetSeqToWakeup(batch.sequence);
+      if (largest_unblockable_seq > last_replconf_getack_seq_) {
+        // Send _getack to the slave to get acknowledgment
+        auto s = util::SockSend(conn_->GetFD(), redis::BulkString("_getack"), conn_->GetBufferEvent());
+        if (!s.IsOK()) {
+          error("Write error while sending _getack to slave: {}. batches: 0x{}", s.Msg());
```
PragmaTwice: `batches: 0x{}`: the format argument is missing in the `error(..)`.

src/cluster/replication.cc
```diff
@@ -608,28 +626,34 @@ ReplicationThread::CBState ReplicationThread::tryPSyncReadCB(bufferevent *bev) {
...
         if (!line) {
           if (data_written) {
-            sendReplConfAck(bev);
+            sendReplConfAck(bev, force_ack);
           }
           return CBState::AGAIN;
         }
         incr_bulk_len_ = line.length > 0 ? std::strtoull(line.get() + 1, nullptr, 10) : 0;
         if (incr_bulk_len_ == 0) {
-          error("[replication] Invalid increment data size");
```
PragmaTwice: Why the error is removed?

PragmaTwice (commented):
Generally looks good except the comment above.

I'm not familiar with the replication part so call for another review. cc @git-hulk.

src/cluster/replication.cc
PragmaTwice: Recheck the code. It's due to we use `continue` in `bulk_string == "_getack"` case, so now `incr_bulk_len_ ` may be zero. Right?

src/cluster/replication.cc
```diff
@@ -654,11 +678,18 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
...
-            sendReplConfAck(bev);
+            sendReplConfAck(bev, force_ack);
           }
           return CBState::AGAIN;
         }
 
+        if (bulk_string == "_getack") {
+          // master would send the _getack command to the master to get acknowledgment
+          // don't write _getack to db here.
+          force_ack = true;
```
PragmaTwice, line 695: Should we just `sendReplConfAck(.., true)` and `return` here instead of `continue`?

src/cluster/replication.cc
zhixinwen: I added a bunch of debug logs when I worked on this feature.
When I removed the logs I accidentally deleted this as well. Sorry for the confusion.

src/cluster/replication.cc
zhixinwen: I don't think so, this would lower replication throughput. Even though it tries to ack ASAP, but lowering throughput can actually make the end to end latency longer.

For example,
T1: SET + WAIT is called
T2: WAL entry and _getack sent to replication
T3: Replication ack
Although the current impl increases the time between T2 & T3, the time diff should be small. But if the throughput is lower, we can see a large gap between T1 and T2.

I am actually seeing 500ms in T1 and T2 in my test env for some cases, I want to follow up with https://github.com/zhixinwen/kvrocks/pull/2 and even more aggressive batching  to increase throughput and make the behavior configurable. Users who heavily use WAIT would need to tune the settings for their workload.

PragmaTwice: Good to me.

PragmaTwice (dismissed):
LGTM. cc @git-hulk Could you give another review when available?

git-hulk:
> Generally looks good except the comment above.
> 
> I'm not familiar with the replication part so call for another review. cc @git-hulk.

okay, will have a look tomorrow.

src/cluster/replication.cc
```diff
@@ -203,6 +206,21 @@ void FeedSlaveThread::loop() {
...
+      auto largest_unblockable_seq = srv_->LargestTargetSeqToWakeup(batch.sequence);
+      if (largest_unblockable_seq > last_replconf_getack_seq_) {
+        // Send _getack to the slave to get acknowledgment
+        auto s = util::SockSend(conn_->GetFD(), redis::BulkString("_getack"), conn_->GetBufferEvent());
+        if (!s.IsOK()) {
+          error("Write error while sending _getack to slave: {}", s.Msg());
+          Stop();
+          return;
+        } else {
+          last_replconf_getack_seq_ = largest_unblockable_seq;
```
git-hulk: Should we set `last_replconf_getack_seq_` to `largest_unblockable_seq` when the value of `largest_unblockable_seq` is zero?

By the way, do you think if it's good to merge the `_getack` command into `batches_bulk`? so that we can reduce one send call like the following:

```C++
bool ReplicationThread::shouldSendGetACK(SequenceNumber seq) {
  auto largest_unblockable_seq = srv_->LargestTargetSeqToWakeup(batch.sequence);
  if (largest_unblockable_seq > last_replconf_getack_seq_) {
     return true;
  }
  last_replconf_getack_seq_ = largest_unblockable_seq;
  return false
}

...

    if (is_first_repl_batch || batches_bulk.size() >= kMaxDelayBytes || updates_in_batches >= kMaxDelayUpdates ||
        srv_->storage->LatestSeqNumber() - batch.sequence <= kMaxDelayUpdates) {
      if (shouldSendGetACK(batch.sequence)) {
        batches_bulk.emplace_back(redis::BulkString("_getack"))
      }
      // Send entire bulk which contain multiple batches
      auto s = util::SockSend(conn_->GetFD(), batches_bulk, conn_->GetBufferEvent());
      if (!s.IsOK()) {
        error("Write error while sending batch to slave: {}. batches: 0x{}", s.Msg(), util::StringToHex(batches_bulk));
        Stop();
        return;
      }

```

git-hulk (commented):
Overall is good to me, one comment inline.

src/cluster/replication.cc
zhixinwen:
```
Should we set last_replconf_getack_seq_ to largest_unblockable_seq when the value of largest_unblockable_seq is zero?
```
I am not sure if I get you.

`last_replconf_getack_seq_` is used to prevent unnecessary `_getack` sent. If we set it back to zero, it would cause unnecessary sends.

The check in `largest_unblockable_seq > last_replconf_getack_seq_` ensures we would not set `last_replconf_getack_seq_` back to zero when `largest_unblockable_seq` is zero.

zhixinwen: I renamed `last_replconf_getack_seq_` to `last_getack_seq` and make it a local var.

zhixinwen: good call to reduce `util::SockSend` call, let me update it.

But in `shouldSendGetAck`, we should only update `last_getack_seq_` if `largest_unblockable_seq > last_getack_seq_`

git-hulk (dismissed):

src/cluster/replication.cc
git-hulk: > The check in largest_unblockable_seq > last_replconf_getack_seq_ ensures we would not set last_replconf_getack_seq_ back to zero when largest_unblockable_seq is zero.

Yes, exactly. Thanks for your response and update.

PragmaTwice (dismissed):

PragmaTwice:
https://github.com/apache/kvrocks/actions/runs/16682305082/job/47243428856?pr=3075
```
--- FAIL: TestSlaveLostMaster (7.94s)
    client.go:98: forwarding tcp stream stopped
    server.go:125: 
        	Error Trace:	/__w/kvrocks/kvrocks/tests/gocase/util/server.go:125
        	            				/__w/kvrocks/kvrocks/tests/gocase/util/server.go:147
        	            				/__w/kvrocks/kvrocks/tests/gocase/util/server.go:116
        	            				/__w/kvrocks/kvrocks/tests/gocase/integration/replication/replication_test.go:591
        	            				/__w/kvrocks/kvrocks/tests/gocase/integration/replication/replication_test.go:624
        	            	signal: aborted (core dumped)
        	Test:       	TestSlaveLostMaster
FAIL
```

Hmmm aborted? Not sure if related to this PR. Just for record.

For regular test workflow in CI, we have a step to find all crashes and dump the stacktrace, but it's in containered workflow (debian), so no stacktrace available.

I'll retrigger a full re-run, and merge if no error happen.

git-hulk (approved):

PR #3077, opened 2025-07-26
fix(migration): fix migration oom and 'Resource temporarily unavailable' error

greatsharp:
Why does OOM happened when doing migration?
- If slot only has string or other no-subkey data, then batch_sender will put raw-key data continuously, it makes rep_ string in WriteBatch class append string continuously, then OOM at last.

Why should we sleep for a while when function 'SockReadLine' ?
 - If we send a big data by APPLYBATCH command to destination node, the destination node could not process this command rapidly, it will report 'Resource temporarily unavailable' error if we read response immediately.
 - So we should sleep for a while then retry reading from fd.

### DIFF
--- a/src/cluster/batch_sender.cc
+++ b/src/cluster/batch_sender.cc
@@ -100,7 +100,7 @@ Status BatchSender::sendApplyBatchCmd(int fd, const rocksdb::WriteBatch &write_b
 
   GET_OR_RET(util::SockSend(fd, redis::ArrayOfBulkStrings({"APPLYBATCH", write_batch.Data()})));
 
-  std::string line = GET_OR_RET(util::SockReadLine(fd));
+  std::string line = GET_OR_RET(util::SockReadLineWithRetry(fd, 10, 500));
 
   if (line.compare(0, 1, "-") == 0) {
     return {Status::NotOK, line};
--- a/src/common/io_util.cc
+++ b/src/common/io_util.cc
@@ -310,8 +310,21 @@ Status SockSetBlocking(int fd, int blocking) {
 
 StatusOr<std::string> SockReadLine(int fd) {
   UniqueEvbuf evbuf;
-  if (evbuffer_read(evbuf.get(), fd, -1) <= 0) {
-    return Status::FromErrno("read response err");
+  while (true) {
+    int ret = evbuffer_read(evbuf.get(), fd, -1);
+    if (ret > 0) {
+      break;
+    } else if (ret == 0) {
+      return Status::FromErrno("read response err: connection closed");
+    } else {
+      if (errno == EAGAIN || errno == EWOULDBLOCK) {
+        // Resource temporarily unavailable, sleep for a while
+        std::this_thread::sleep_for(std::chrono::milliseconds(200));
+        continue;
+      } else {
+        return Status::FromErrno("read response err");
+      }
+    }
   }
 
   UniqueEvbufReadln line(evbuf.get(), EVBUFFER_EOL_CRLF_STRICT);
--- a/src/common/io_util.cc
+++ b/src/common/io_util.cc
@@ -335,7 +335,7 @@ StatusOr<std::string> SockReadLineWithRetry(int fd, int retry_times, int retry_i
     }
 
     if (ret == 0) {
-      return Status::FromErrno("read response err: connection closed");
+      return {Status::NotOK, "read response err: connection closed"};
     }
 
     if (errno == EAGAIN || errno == EWOULDBLOCK) {
--- a/src/common/io_util.cc
+++ b/src/common/io_util.cc
@@ -314,17 +314,19 @@ StatusOr<std::string> SockReadLine(int fd) {
     int ret = evbuffer_read(evbuf.get(), fd, -1);
     if (ret > 0) {
       break;
-    } else if (ret == 0) {
-      return Status::FromErrno("read response err: connection closed");
-    } else {
-      if (errno == EAGAIN || errno == EWOULDBLOCK) {
-        // Resource temporarily unavailable, sleep for a while
-        std::this_thread::sleep_for(std::chrono::milliseconds(200));
-        continue;
-      } else {
-        return Status::FromErrno("read response err");
-      }
     }
+
+    if (ret == 0) {
+      return Status::FromErrno("read response err: connection closed");
+    }
+
+    if (errno == EAGAIN || errno == EWOULDBLOCK) {
+      // Resource temporarily unavailable, sleep for a while
+      std::this_thread::sleep_for(std::chrono::milliseconds(200));
+      continue;
+    }
+
+    return Status::FromErrno("read response err");
   }
 
   UniqueEvbufReadln line(evbuf.get(), EVBUFFER_EOL_CRLF_STRICT);
--- a/src/common/io_util.cc
+++ b/src/common/io_util.cc
@@ -310,7 +310,25 @@ Status SockSetBlocking(int fd, int blocking) {
 
 StatusOr<std::string> SockReadLine(int fd) {
   UniqueEvbuf evbuf;
-  while (true) {
+  if (evbuffer_read(evbuf.get(), fd, -1) <= 0) {
+    return Status::FromErrno("read response err");
+  }
+
+  UniqueEvbufReadln line(evbuf.get(), EVBUFFER_EOL_CRLF_STRICT);
+  if (!line) {
+    return Status::FromErrno("read response err(empty)");
+  }
+
+  return std::string(line.get(), line.length);
+}
+
+StatusOr<std::string> SockReadLineWithRetry(int fd, int retry_times, int retry_interval_ms) {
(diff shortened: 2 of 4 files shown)


src/common/io_util.cc
```diff
@@ -310,8 +310,21 @@ Status SockSetBlocking(int fd, int blocking) {
 
 StatusOr<std::string> SockReadLine(int fd) {
   UniqueEvbuf evbuf;
-  if (evbuffer_read(evbuf.get(), fd, -1) <= 0) {
-    return Status::FromErrno("read response err");
+  while (true) {
+    int ret = evbuffer_read(evbuf.get(), fd, -1);
+    if (ret > 0) {
+      break;
+    } else if (ret == 0) {
```
torwig: Here we can make code more "flat":
```
if (ret > 0) {
  break;
}
 
if (ret == 0) { 
  return Status::FromErrno("read response err: connection closed");
} 

if (errno == EAGAIN || errno == EWOULDBLOCK) {
    // Resource temporarily unavailable, sleep for a while
    std::this_thread::sleep_for(std::chrono::milliseconds(200));
    continue;
}

return Status::FromErrno("read response err");
```

src/common/io_util.cc
```diff
@@ -310,7 +310,22 @@ Status SockSetBlocking(int fd, int blocking) {
 
 StatusOr<std::string> SockReadLine(int fd) {
   UniqueEvbuf evbuf;
-  if (evbuffer_read(evbuf.get(), fd, -1) <= 0) {
+  while (true) {
```
PragmaTwice: Should we retry on a higher level rather than `SockReadLine`? cc @git-hulk

git-hulk: Agree with this. Perhaps you could also provide another sock read function with limited retry count.

git-hulk:
@greatsharp Nice shoot.

src/common/io_util.cc
greatsharp: How about add a new function like below?

```
StatusOr<std::string> SockReadLineWithRetry(int fd, int retry_times, int retry_interval_ms) {
  if (retry_times <= 0 || retry_interval_ms <= 0) {
    return SockReadLine(fd);
  }
 
  UniqueEvbuf evbuf;
  while (retry_times-- > 0) {
    int ret = evbuffer_read(evbuf.get(), fd, -1);
    if (ret > 0) {
      break;
    }

    if (ret == 0) {
      return Status::FromErrno("read response err: connection closed");
    }

    if (errno == EAGAIN || errno == EWOULDBLOCK) {
      // Resource temporarily unavailable, sleep for a while
      std::this_thread::sleep_for(std::chrono::milliseconds(retry_interval_ms));
      continue;
    }

    return Status::FromErrno("read response err");
  }
  
  UniqueEvbufReadln line(evbuf.get(), EVBUFFER_EOL_CRLF_STRICT);
  if (!line) {
    return Status::FromErrno("read response err(empty)");
  }

  return std::string(line.get(), line.length);
}
```

greatsharp: Add a new function can tell callers there is a 'Resource temporarily unavailable' error, callers can pass retry times and interval, how do you see? @git-hulk @torwig @PragmaTwice

git-hulk: Sounds great to me.

git-hulk (dismissed):

src/common/io_util.cc
```diff
@@ -322,6 +322,39 @@ StatusOr<std::string> SockReadLine(int fd) {
...
+
+  UniqueEvbuf evbuf;
+  while (retry_times-- > 0) {
+    int ret = evbuffer_read(evbuf.get(), fd, -1);
+    if (ret > 0) {
+      break;
+    }
+
+    if (ret == 0) {
+      return Status::FromErrno("read response err: connection closed");
```
PragmaTwice:
```suggestion
      return {Status::NotOK, "read response err: connection closed"};
```

src/cluster/batch_sender.cc
```diff
@@ -100,7 +100,7 @@ Status BatchSender::sendApplyBatchCmd(int fd, const rocksdb::WriteBatch &write_b
 
   GET_OR_RET(util::SockSend(fd, redis::ArrayOfBulkStrings({"APPLYBATCH", write_batch.Data()})));
 
-  std::string line = GET_OR_RET(util::SockReadLine(fd));
+  std::string line = GET_OR_RET(util::SockReadLineWithRetry(fd, 10, 500));
```
PragmaTwice, line 103: In non-blocking IO, it's quite normal to receive a EAGAIN or EWOULDBLOCK to indicate that currently there's no data to retrieve from the stream.

In such case, if we just make it a busy retry loop (or with a sleep), we are actually converting it to a blocking socket.

And IMHO it seems not a good method.

greatsharp: After reviewing the raw-key-value migration process, the source node needs to send WriteBatch data sequentially, which is more suitable for blocking I/O

PragmaTwice:
Maybe you can just keep the "fix migration oom" part in this PR, so we can merge it first.

And for the nonblocking io part we can take some time to see if we have good idea.

src/cluster/batch_sender.cc
zhixinwen: I am confused here. I think the fd here is made by `util:SockConnect`. From the code, it does looks like `SockConnect` returns  a blocking fd. Am I missing anything?

greatsharp: I am also confused here, util::SocketConnect returns a blocking socket fd, but how could a blocking socket report EAGAIN or EWOULDBLOCK error? 

Add below test logic after  dst_fd_.Reset(*result) in SlotMigrator::startMigration() method, 
`
int flags = fcntl(*result, F_GETFL, 0);
if ((flags & O_NONBLOCK) == 0) {
   LOG(ERROR) << "block socket";
} else {
   LOG(ERROR) << "non_block socket";
}

`


the log print out "block socket"

greatsharp: @git-hulk @PragmaTwice @zhixinwen 
I still used util::SockReadLineWithRetry(fd, 10, 500) to resolve this 'Resource temporarily unavailable' error, it works well in big data cluster migration, but how could a blocking socket report EAGAIN or EWOULDBLOCK error?

Could you please give me some guide?

zhixinwen: My case is a bit different, I get this error in `checkMultipleResponses` and I think that may be due to the 1s timeout setting.

What is the full error message you see? I think it can come from several places.. Also I am not sure if evbuffer should be used on a blocking connection, that lib is intended for non-blocking fd.

greatsharp: > My case is a bit different, I get this error in `checkMultipleResponses` and I think that may be due to the 1s timeout setting.
> 
> What is the full error message you see? I think it can come from several places.. Also I am not sure if evbuffer should be used on a blocking connection, that lib is intended for non-blocking fd.

please use raw-key-value migrate type

zhixinwen: I was using `raw-key-value` and I got:
```
[2025-09-19T19:02:16.393369+00:00][E][slot_migrate.cc:195] [migrate] Failed to start migrating slot(s) 100. Error: failed to set import status on destination node: failed to check the response from the destination node: failed to read response: Resource temporarily unavailable
```

zhixinwen: I was having compaction stop due to migration and my theory is the blocked rocksdb call blocks the event loop and paused all command processing. In fact, even PING was frozen on the target node during the time.

zhixinwen: Added some log for debugging:
```
  GET_OR_RET(util::SockSend(fd, redis::ArrayOfBulkStrings({"APPLYBATCH", write_batch.Data()})));
  // INSERT_YOUR_CODE
  // Log the SO_RCVTIMEO (receive timeout) for fd
  struct timeval tv;
  socklen_t tv_len = sizeof(tv);
  if (getsockopt(fd, SOL_SOCKET, SO_RCVTIMEO, &tv, &tv_len) == 0) {
    LOG(INFO) << "[migrate] fd " << fd << " SO_RCVTIMEO: " << tv.tv_sec << "s " << tv.tv_usec << "us";
  } else {
    LOG(WARNING) << "[migrate] Failed to get SO_RCVTIMEO for fd " << fd << ": " << strerror(errno);
  }

  std::string line = GET_OR_RET(util::SockReadLine(fd));
```

and I get `[2025-09-25T18:08:58.993539+00:00][I][batch_sender.cc:107] [migrate] fd 5488 SO_RCVTIMEO: 1s 0us`.

The 1s timeout is set because it was set in`checkMultipleResponses` which later affects `APPLYBATCH`. The `Resource temporarily unavailable` error is due to the timeout.

I think fixing compaction is the best way to go, but we should think about how to retry failure in general.

greatsharp: I checked the socket SO_RCVTIMEO with your code, it is set to 1s after invoked checkMultipleResponses method.

PR #3087, opened 2025-07-31
feat(replication): make replication delay configurable

zhixinwen:
The batching behavior should be configurable because it would depend on user's workload ect.

Here is a test for the throughput impact on my test setup.
Test payload: 4kb, command: SET + WAIT

replication-delay-bytes | replication-delay-updates | QPS | MB/s
-- | -- | -- | --
16384 | 16 | 38545.80 | 150 (default setup)
65536 | 16 | 43583.46 | 170
131072 | 32 | 44367 | 173
262144 | 64 | 44887.44 | 175
524288 | 128 | 38890.10 | 151

We can get 17% throughput increase with optimal tuning.

### DIFF
--- a/src/cluster/replication.h
+++ b/src/cluster/replication.h
@@ -32,6 +32,7 @@
 #include <utility>
 #include <vector>
 
+#include "config/config.h"
 #include "event_util.h"
 #include "io_util.h"
 #include "rocksdb/write_batch.h"
@@ -64,8 +65,7 @@ using FetchFileCallback = std::function<void(const std::string &, uint32_t)>;
 
 class FeedSlaveThread {
  public:
-  explicit FeedSlaveThread(Server *srv, redis::Connection *conn, rocksdb::SequenceNumber next_repl_seq)
-      : srv_(srv), conn_(conn), next_repl_seq_(next_repl_seq), req_(srv) {}
+  explicit FeedSlaveThread(Server *srv, redis::Connection *conn, rocksdb::SequenceNumber next_repl_seq);
   ~FeedSlaveThread() = default;
 
   Status Start();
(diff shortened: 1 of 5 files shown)


PragmaTwice (dismissed):
LGTM.

src/cluster/replication.h
```diff
@@ -64,8 +65,13 @@ using FetchFileCallback = std::function<void(const std::string &, uint32_t)>;
...
  public:
-  explicit FeedSlaveThread(Server *srv, redis::Connection *conn, rocksdb::SequenceNumber next_repl_seq)
-      : srv_(srv), conn_(conn), next_repl_seq_(next_repl_seq), req_(srv) {}
+  FeedSlaveThread(Server *srv, redis::Connection *conn, rocksdb::SequenceNumber next_repl_seq)
+      : srv_(srv),
+        conn_(conn),
+        next_repl_seq_(next_repl_seq),
+        req_(srv),
+        max_delay_bytes_(srv->GetConfig()->max_replication_delay_bytes),
+        max_delay_updates_(srv->GetConfig()->max_replication_delay_updates) {}
```
PragmaTwice: Need to move it to `replication.cc` to avoid incomplete type.

src/cluster/replication.h
```diff
@@ -36,6 +36,7 @@
 #include "io_util.h"
 #include "rocksdb/write_batch.h"
 #include "server/redis_connection.h"
+#include "server/server.h"
```
PragmaTwice: Remove this.

src/cluster/replication.h
```diff
@@ -87,8 +93,9 @@ class FeedSlaveThread {
   redis::Request req_;
   std::atomic<rocksdb::SequenceNumber> ack_seq_ = 0;
 
-  static const size_t kMaxDelayUpdates = 16;
-  static const size_t kMaxDelayBytes = 16 * 1024;
+  // Configurable delay limits
+  size_t max_delay_bytes_ = 16 * 1024;
+  size_t max_delay_updates_ = 16;
```
PragmaTwice:
```suggestion
  size_t max_delay_bytes_;
  size_t max_delay_updates_;
```

Default value seems not so useful here.

src/cluster/replication.h
zhixinwen: could you explain what is incomplete type here?
Sorry not familiar with C++

zhixinwen: ok, I see you are referring to the lint error. I will google it, thx.

git-hulk (approved):

PragmaTwice (approved):

PR #3090, opened 2025-08-03
feat(replication): optimize wait and replication locks

zhixinwen:
Change `wait_contexts_mu_` and `slave_threads_mu_` to `shared_mutex`.
The main motivation is:
1. WAIT command uses `GetReplicasReachedSequence` to find out the number of replicas it has reached. If `slave_threads_mu_` uses `unique_lock`, all WAIT cmd needs to be serialized. Also metrics needs to grab `slave_threads_mu_` and it would also slow dow WAIT.
2. `wait_contexts_mu_` is grabbed by `LargestTargetSeqToWakeup` and it is being called frequently when WAIT is used heavily. `LargestTargetSeqToWakeup` is read only so the change can reduce lock contension.

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -336,13 +336,13 @@ Status Server::AddSlave(redis::Connection *conn, rocksdb::SequenceNumber next_re
     return s;
   }
 
-  std::lock_guard<std::mutex> lg(slave_threads_mu_);
+  std::unique_lock<std::shared_mutex> lg(slave_threads_mu_);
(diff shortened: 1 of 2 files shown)


PragmaTwice (approved):
LGTM. In the future maybe we can utilize some concurrent data structures in TBB.

e.g. https://uxlfoundation.github.io/oneTBB/main/tbb_userguide/concurrent_vector_ug.html

PR #3092, opened 2025-08-04
feat(replication): replication support group sync

zhixinwen:
If we want no data loss, we have to turn on `rocksdb.write_options.sync`.

This would create a big bottleneck for replication throughput because of its single threaded nature. Master would be affected less, because RocksDB would batch several concurrent write into one WriteBatch.

However, with ack based replication impelemented, the replica does not necessarily have to sync for every write. It just needs to sync before it acks.

This PR adds an option `replication-group-sync`. If both `replication-group-sync` and `rocksdb.write_options.sync` are on, replication would call sync once before ack. This would still guarantee all the acked changes would have no data loss, but the replication throughput would be much higher. We saw 10+% throughput increase with this option on.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -193,7 +193,7 @@ replication-connect-timeout-ms 3100
 # future 'clusterx setnodes' commands because the replication thread is blocked on recv.
 replication-recv-timeout-ms 3200
 
-# Ignoed when rocksdb.write_options.sync is no.
+# Ignored when rocksdb.write_options.sync is no.
 # When rocksdb.write_options.sync is yes, the replica will:
 # 1) Pull the latest changes from master
 # 2) Write the changes to replica's local storage. Each write would be called with rocksdb.write_options.sync = true. And the write would be synced to disk.
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -193,6 +193,21 @@ replication-connect-timeout-ms 3100
 # future 'clusterx setnodes' commands because the replication thread is blocked on recv.
 replication-recv-timeout-ms 3200
 
+# Ignoed when rocksdb.write_options.sync is no.
+# When rocksdb.write_options.sync is yes, the replica will:
+# 1) Pull the latest changes from master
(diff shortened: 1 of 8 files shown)


zhixinwen:
The idea is to reduce to number of fsync calls needed.
Another thing I tried is to use `WriteBatchHandler` to combine all the `WriteBatch` within one call of `incrementBatchLoopCB` and write the merged batch once. However, I do not see much improvement with setup. I cannot understand this behavior because in my rocksdb local test, this method should be even more efficient than what is implemented in this PR. Not sure if the `WriteBatchHandler` overhead is too big.

This is the PR: https://github.com/zhixinwen/kvrocks/pull/5, in case anyone is interested:

PragmaTwice (approved):

zhixinwen:
`Resource deadlock avoid` seems to occur when the thread owner joins itself, I do not see when this would be possible.

zhixinwen:
@PragmaTwice I don't think the seg fault is related to this PR, can we merge it first

zhixinwen:
Root cause the issue: https://github.com/apache/kvrocks/issues/3112
This is a systematic issue in KVRocks

git-hulk (approved):

PragmaTwice:
I'll merge it after the CI passing.

PragmaTwice:
```
--- FAIL: TestSlaveLostMaster (11.45s)
    client.go:98: forwarding tcp stream stopped
    server.go:125: 
        	Error Trace:	/__w/kvrocks/kvrocks/tests/gocase/util/server.go:125
        	            				/__w/kvrocks/kvrocks/tests/gocase/util/server.go:147
        	            				/__w/kvrocks/kvrocks/tests/gocase/util/server.go:116
        	            				/__w/kvrocks/kvrocks/tests/gocase/integration/replication/replication_test.go:591
        	            				/__w/kvrocks/kvrocks/tests/gocase/integration/replication/replication_test.go:624
        	            	signal: aborted (core dumped)
        	Test:       	TestSlaveLostMaster
FAIL
```

Just for recording. Re-run.

PR #3094, opened 2025-08-04
fix(migration): send migration batch while full to avoid oom

greatsharp:


greatsharp:
submit a issue #3095 to track the "Resource temporarily unavailable" error

PragmaTwice (approved):
LGTM cc @git-hulk

git-hulk (dismissed):

src/cluster/slot_migrate.cc
```diff
@@ -1296,6 +1296,9 @@ Status SlotMigrator::sendSnapshotByRawKV() {
 
     auto subkey_iter = iter.GetSubKeyIterator();
     if (!subkey_iter) {
+      if (batch_sender.IsFull()) {
```
git-hulk, line 1299: We should reset the batch sender after sending?

git-hulk (approved):

src/cluster/slot_migrate.cc
git-hulk: The sender will clear the sent batch.

PR #3098, opened 2025-08-04
feat(replication): set low watermark when reading replication data

zhixinwen:
When master sends large amount of data, the replication read callback can prematurely wakeup only to find it still does not have enough data. Set a watermark would reduce unnecessary wakeup

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -671,6 +671,8 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
           if (data_written) {
             sendReplConfAck(bev, force_ack);
           }
+          // set a watermark so the callback won't be called again until the data is enough
+          bufferevent_setwatermark(bev, EV_READ, incr_bulk_len_ + 2, 0);
           return CBState::AGAIN;
         }
 


src/cluster/replication.cc
```diff
@@ -671,6 +671,8 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
           if (data_written) {
             sendReplConfAck(bev, force_ack);
           }
+          // set a watermark so the callback won't be called again until the data is enough
+          bufferevent_setwatermark(bev, EV_READ, incr_bulk_len_ + 2, 0);
```
PragmaTwice, line 675: Should the highmark be larger than the lowmark?

zhixinwen: We should not set high watermark, because if there is more data than high watermark, no more data would be read into the buffer from the network. We should read as much data as possible for optimal throughput.

Leaving high watermark as zero would disable it.
https://github.com/libevent/libevent/blob/master/bufferevent_sock.c#L174

git-hulk (approved):

PragmaTwice (approved):

Issue #3097, opened 2025-08-04
segfault in FeedSlaveThread


zhixinwen:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

We were running on test with `2.12.1` + https://github.com/apache/kvrocks/pull/3047 + https://github.com/apache/kvrocks/pull/3061

```
E20250803 09:11:53.714233 139623546201664 redis_connection.cc:110] [connection] Going to remove the client: 127.0.0.1:54549, while encounter error: Connection reset by peer, SSL Error: error:00000000:lib(0)::reason(0)
E20250803 10:30:12.468891 139623504238144 replication.cc:180] Fatal error encountered, WAL iterator is discrete, some seq might be lost, sequence 465907582 expected, but got 465925769
E20250803 12:20:12.239719 139623243707968 replication.cc:180] Fatal error encountered, WAL iterator is discrete, some seq might be lost, sequence 496763511 expected, but got 496771469
E20250803 13:20:12.841936 139623243707968 replication.cc:180] Fatal error encountered, WAL iterator is discrete, some seq might be lost, sequence 514094241 expected, but got 514102200
E20250803 14:20:12.635324 139623243707968 replication.cc:180] Fatal error encountered, WAL iterator is discrete, some seq might be lost, sequence 531497261 expected, but got 531506354
E20250803 15:00:12.795350 139623504238144 replication.cc:180] Fatal error encountered, WAL iterator is discrete, some seq might be lost, sequence 542936338 expected, but got 542953399
E20250803 15:00:48.842797 139623678342720 redis_connection.cc:110] [connection] Going to remove the client: 10.14.24.143:19425, while encounter error: Connection reset by peer, SSL Error: error:00000000:lib(0)::reason(0)
E20250803 15:00:48.846916 139623726577216 redis_connection.cc:110] [connection] Going to remove the client: 10.14.24.143:19441, while encounter error: Connection reset by peer, SSL Error: error:00000000:lib(0)::reason(0)
E20250803 15:00:48.848751 139623678342720 redis_connection.cc:110] [connection] Going to remove the client: 10.14.24.143:19497, while encounter error: Connection reset by peer, SSL Error: error:00000000:lib(0)::reason(0)
[...]

PR #3104, opened 2025-08-06
feat(replication): clean up all wait_contexts in CleanupWaitConnection

zhixinwen:
This should mitigate https://github.com/apache/kvrocks/issues/3097#issuecomment-3161427225, see https://github.com/apache/kvrocks/issues/3097#issuecomment-3161427225 for root cause. I have tested locally and the seg fault is fixed.

For a proper fix, I will send a followup PR with:
1. Disable connection read callback when it is blocked by WAIT
2. Add a test to ensure while a connection is blocked by WAIT, it will not process additional commands.

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -744,16 +744,17 @@ void Server::CleanupWaitConnection(redis::Connection *conn) {
     if (it->second.conn == conn) {
       it = wait_contexts_.erase(it);
       erased_count++;
+      // Technically only one client is unblocked, but we call IncrBlockedClientNum for each added wait context,
+      // so we need to call DecrBlockedClientNum for each erased wait context.
+      // Multiple wait contexts on the same connection should not happen, but we should be defensive.
+      DecrBlockedClientNum();
     } else {
       ++it;
     }
   }
 
   if (erased_count > 0) {
-    DecrBlockedClientNum();
-    if (erased_count > 1) {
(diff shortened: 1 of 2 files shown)


git-hulk (dismissed):

src/server/server.cc
```diff
@@ -737,11 +737,23 @@ void Server::WakeupWaitConnections(rocksdb::SequenceNumber seq) {
...
+    if (it->second.conn == conn) {
+      it = wait_contexts_.erase(it);
+      erased_count++;
+    } else {
+      ++it;
+    }
+  }
+
+  if (erased_count > 0) {
     DecrBlockedClientNum();
```
PragmaTwice: `DecrBlockedClientNum` should be called more than once?

src/server/server.cc
```diff
@@ -737,11 +737,23 @@ void Server::WakeupWaitConnections(rocksdb::SequenceNumber seq) {
...
+    if (erased_count > 1) {
+      warn("[server] {} wait contexts found for connection with fd {}, expect 1", erased_count, conn->GetFD());
+    }
```
PragmaTwice: Hmm I want to know under what circumstances multiple wait contexts will appear on the same connection..

PragmaTwice: Ahh got it https://github.com/apache/kvrocks/issues/3097#issuecomment-3161427225.

src/server/server.cc
zhixinwen: good catch

git-hulk (approved):

PR #3107, opened 2025-08-10
feat(ts): Add support for data writing and `TS.CREATE`, `TS.ADD/MADD` commands

yezhizi:
Now, we can do something like this:
```
127.0.0.1:6666> ts.create test2 CHUNK_SIZE 3 RETENTION 10
OK
127.0.0.1:6666> ts.madd test2 1 10 test2 3 10 test2 2 20 test2 3 20 test2 4 20 test2 13 20 test2 1 20 test2 14 20
1) (integer) 1
2) (integer) 3
3) (integer) 2
4) (error) ERR Error at upsert, update is not supported when DUPLICATE_POLICY is set to BLOCK mode
5) (integer) 4
6) (integer) 13
7) (error) ERR Timestamp is older than retention
8) (integer) 14
```

### DIFF
--- a/src/commands/cmd_timeseries.cc
+++ b/src/commands/cmd_timeseries.cc
@@ -58,10 +58,10 @@ namespace redis {
 class KeywordCommandBase : public Commander {
  public:
   KeywordCommandBase(size_t skip_num, size_t tail_skip_num) : skip_num_(skip_num), tail_skip_num_(tail_skip_num) {}
-  virtual ~KeywordCommandBase() = default;
 
   Status Parse(const std::vector<std::string> &args) override {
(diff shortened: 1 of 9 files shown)


PragmaTwice:
Hmm for out-of-order TS writing, IMO it should not be too different with ordered writing in the current design. Is it right? Or is there something I'm missing?

yezhizi:
> Hmm for out-of-order TS writing, IMO it should not be too different with ordered writing in the current design. Is it right? Or is there something I'm missing?

Yes, there is no difference between them. The PR didn’t support the `split` logic at first, so there were issues with adding out-of-order data. But now it’s supported.

PragmaTwice:
Thank you for your explanation. I'll take a review in this weekend.

PragmaTwice (approved):

PR #3115, opened 2025-08-15
fix(function): reset lua states from all workers in FUNCTION DELETE

PragmaTwice:
This PR contains two changes:
1. in FUNCTION DELETE, we make sure that all lua states are reset instead of just cleaning in one thread.
2. make sure the script is read-only in FUNCTION LOAD.


git-hulk (approved):
LGTM

PR #3124, opened 2025-08-18
fix(search): wrong upper bound in reverse scan for infinity

zhenghaoz:
If `scan->range.r` is infinity, `IntervalSet::PrevNum` leads wrong result.


PragmaTwice (approved):

PragmaTwice:
Great catch! Thank you.

PR #3125, opened 2025-08-18
build: add a hint to the build command for MacOS

xuzifu666:
Sometime it would error out when build in MacOS with command: ./py.x build, such as:
```
stdlib.h file not found.
```

we can use this way to build success, so I add it to the document for MacOS users:
```
./x.py build -DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ $ git clone https://github.com/apache/kvrocks.git
 $ cd kvrocks
 $ ./x.py build 
 # `./x.py build -h` to check more options
-#If you encounter "stdlib.h file not found" error in macOS, 
+#If you encounter "stdlib.h file not found" error in MacOS, 
 #you can add '-DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk' to try again.
 ```
 
--- a/README.md
+++ b/README.md
@@ -82,8 +82,8 @@ $ git clone https://github.com/apache/kvrocks.git
 $ cd kvrocks
 $ ./x.py build 
 # `./x.py build -h` to check more options
-#If you encounter "stdlib.h file not found" error in MacOS, 
-#you can add '-DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk' to try again.
+# If you encounter "stdlib.h file not found" error in MacOS, 
+# you can add '-DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk' to try again.
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,7 +80,10 @@ It is as simple as:
 ```shell
 $ git clone https://github.com/apache/kvrocks.git
 $ cd kvrocks
-$ ./x.py build # `./x.py build -h` to check more options
+$ ./x.py build 
+# `./x.py build -h` to check more options
+#If you encounter "stdlib.h file not found" error in MacOS, 
+#you can add '-DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk' to try again.
```
git-hulk:
```suggestion
# If you encounter "stdlib.h file not found" error in MacOS, 
# you can add '-DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk' to try again.
```

xuzifu666: Done

git-hulk (approved):

PragmaTwice (changes requested):
Did some docs checking..

https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html
> In order to pass an explicit macOS SDK via the compiler's -isysroot flag, users may configure their build tree with -DCMAKE_OSX_SYSROOT=macosx, or export SDKROOT="$(xcrun --sdk macosx --show-sdk-path)" in their environment.


Maybe we can add it to our faqs (https://kvrocks.apache.org/docs/faq) instead of here. Also not sure if the hardcoded path is good enough (`SDKROOT` is enough? not a mac user though..).

xuzifu666:
> Did some docs checking..
> 
> https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html
> 
> > In order to pass an explicit macOS SDK via the compiler's -isysroot flag, users may configure their build tree with -DCMAKE_OSX_SYSROOT=macosx, or export SDKROOT="$(xcrun --sdk macosx --show-sdk-path)" in their environment.
> 
> Maybe we can add it to our faqs (https://kvrocks.apache.org/docs/faq) instead of here. Also not sure if the hardcoded path is good enough (`SDKROOT` is enough? not a mac user though..).

Thanks for suggestion, I agree with moving it to faq, and I would verify whether the way by SDKROOT is OK. @PragmaTwice

xuzifu666:
I had tried ```export SDKROOT="$(xcrun --sdk macosx --show-sdk-path)```, it also throws stdlib.h not found, so I file a new pr for it. https://github.com/apache/kvrocks-website/pull/317  @git-hulk @PragmaTwice

PragmaTwice:
Closed via apache/kvrocks-website#317.

PR #3130, opened 2025-08-19
feat(tdigest): add the support of TDIGEST.REVRANK command

### LINKED ISSUE #3063: TDigest: Implement `TDIGEST.REVRANK` command
### Search before asking

### Motivation

We should implement `TDIGEST.REVRANK` command.

Redis command reference:
https://redis.io/docs/latest/commands/tdigest.revrank/

### Solution

Depends on #2791 #2792.

Follow the implementation of #2794 for this command.

donghao526:
# ISSUE
It closes #3063.

## Proposed Changes
Add TDIGEST.REVRANK command implementation
Add cpp unit tests


PragmaTwice:
Thank you for your contribution. Could you add some golang test cases for it?

Refer to https://github.com/apache/kvrocks/blob/unstable/tests/gocase/unit/type/tdigest/tdigest_test.go.

donghao526:
@PragmaTwice ok，I will add some golang test

src/types/redis_tdigest.cc
```diff
@@ -186,6 +186,67 @@ rocksdb::Status TDigest::Add(engine::Context& ctx, const Slice& digest_name, con
...
+        return status;
+      }
+
+      if (auto status = storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch()); !status.ok()) {
+        return status;
+      }
+
+      ctx.RefreshLatestSnapshot();
+    }
+  }
```
LindaSummer, line 239: Hi @donghao526 ,

The merge action could be refactored into a function to reduce duplication of the same logic.

Best Regards,
Edward

src/types/redis_tdigest.cc
```diff
@@ -186,6 +186,67 @@ rocksdb::Status TDigest::Add(engine::Context& ctx, const Slice& digest_name, con
...
+  result->clear();
+  result->reserve(inputs.size());
+
+  for (auto value : inputs) {
+    auto status_or_rank = TDigestRevRank(dump_centroids, value);
+    if (!status_or_rank) {
+      return rocksdb::Status::InvalidArgument(status_or_rank.Msg());
+    }
+    result->push_back(*status_or_rank);
+  }
```
LindaSummer: Hi @donghao526 ,

We could sort the inputs and get the ranks with just one scan of the centroids since it's sorted.

Best Regards,
Edward

src/types/tdigest.h
```diff
@@ -150,3 +150,22 @@ inline StatusOr<double> TDigestQuantile(TD&& td, double q) {
   diff /= (lc.weight / 2 + rc.weight / 2);
   return Lerp(lc.mean, rc.mean, diff);
 }
+
+template <typename TD>
+inline StatusOr<int> TDigestRevRank(TD&& td, double value) {
```
LindaSummer: Hi @donghao526 ,

We need to use a stable way to compare between doubles.

It will be tough to assume that the two double numbers are equal to or greater than.

After solving this, we should add some test cases for this corner case.

Best Regards,
Edward

LindaSummer: Hi @donghao526 ,

Since the other code snippets use this way now. You could leave it with the current logic.

I will try to create a new PR to solve the unstable comparison problem in this file.

Best Regards,
Edward

LindaSummer (commented):
Hi @donghao526 ,

Left some comments.

Best Regards,
Edward

src/types/redis_tdigest.cc
donghao526: Hi @LindaSummer 
I encountered a problem when I was testing. After the nodes merged, are there two adjacent centroids can be with the same mean?

I Test with
```
TDIGEST.CREATE s COMPRESSION 1000

TDIGEST.ADD s 10 10 10 10 20 20
```

I found the centroids after merged are:
(1) mean: 10 weight: 1
(2) mean: 10 weight: 1
(3) mean: 10 weight: 1
(4) mean: 10 weight: 1
(5) mean: 20 weight: 1
(6) mean: 20 weight: 1

Is this as expected or a bug?

LindaSummer: Hi @donghao526 ,

It is expected, and you could refer to #2878 for more details.
So we need a stable way for both serialization and deserialization.

The trigger for the merge is the weight, not the mean. So we could treat the mean only as a label of one centroid. The whole logic is driven by weight.

Best Regards,
Edward

donghao526: got it

src/types/tdigest.h
donghao526: OK, after your new PR, I can help to fix here.

donghao526:
@LindaSummer  I have modified the code according to your suggestion, please review.
1. The merge action has been **refactored into the **mergeNodes** function**
2. I have **sorted the inputs in **TDigestRevRank****, now we could get the ranks with just one scan of the centroids

Best Regards

src/commands/cmd_tdigest.cc
```diff
@@ -176,6 +176,49 @@ class CommandTDigestAdd : public Commander {
...
+    std::vector<int> result;
+    result.reserve(inputs_.size());
+    if (const auto s = tdigest.RevRank(ctx, key_name_, inputs_, result); !s.ok()) {
+      if (s.IsNotFound()) {
+        return {Status::RedisExecErr, errKeyNotFound};
+      }
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    if (result.data()) {
```
LindaSummer:
```suggestion
    if (!result.empty()) {
```

src/types/redis_tdigest.cc
```diff
@@ -186,8 +188,37 @@ rocksdb::Status TDigest::Add(engine::Context& ctx, const Slice& digest_name, con
...
+  metadata->Encode(&metadata_bytes);
+  if (auto status = batch->Put(metadata_cf_handle_, ns_key, metadata_bytes); !status.ok()) {
+    return status;
+  }
+
+  if (auto status = storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch()); !status.ok()) {
+    return status;
+  }
+
+  ctx.RefreshLatestSnapshot();
```
LindaSummer, line 216: ~~We'd better refresh the metadata after the snapshot refresh. Another thread may update it.~~
A lock guards this merge action, so the metadata should remain consistent after refreshing the snapshot.

src/types/tdigest.h
```diff
@@ -150,3 +151,74 @@ inline StatusOr<double> TDigestQuantile(TD&& td, double q) {
...
+        current_mean_cumulative_weight += centroid.weight / 2;
+        cumulative_weight += centroid.weight;
+      }
+
+      // assign the reverse rank for the inputs[indices[i - 1]]
+      result[indices[i - 1]] = static_cast<int>(current_mean_cumulative_weight);
+      i--;
+
+      // handle the prev inputs which has the same value
+      while ((i > 0) && (inputs[indices[i]] == inputs[indices[i - 1]])) {
```
LindaSummer: Maybe we could use a `map<double, std::vector<size_t>>` to construct a sorted input and group redundant input with rank into a group, and it may simplify our logic.

src/types/tdigest.h
```diff
@@ -150,3 +151,74 @@ inline StatusOr<double> TDigestQuantile(TD&& td, double q) {
...
+    iter->Prev();
+  }
+
+  // handle inputs less than minimum
+  while (i > 0) {
+    result[indices[i - 1]] = static_cast<int>(td.TotalWeight());
+    i--;
+  }
+  return Status::OK();
+}
```
LindaSummer: Add a newline to the file.

src/types/tdigest.h
```diff
@@ -150,3 +151,74 @@ inline StatusOr<double> TDigestQuantile(TD&& td, double q) {
   diff /= (lc.weight / 2 + rc.weight / 2);
   return Lerp(lc.mean, rc.mean, diff);
 }
+
+template <typename TD>
+inline Status TDigestRank(TD&& td, const std::vector<double>& inputs, std::vector<int>& result) {
```
LindaSummer: This function seems specialized for `RevRank` rather than `Rank`.

Could we implement the `Rank`, then wrap the `Iterator` to a reverse version with the same logic to construct `RevRank`?

LindaSummer (commented):
Hi @donghao526 ,

Thanks very much for your great effort! 😊

Left some comments.

Best Regards,
Edward

LindaSummer:
Hi @donghao526 ,

Sorry to interrrupt you.
Do you still work on this? I want to continue work on this draft PR if you don't mind. 😊

Best Regards,
Edward

donghao526:
> Hi @donghao526 ,
> 
> Sorry to interrrupt you. Do you still work on this? I want to continue work on this draft PR if you don't mind. 😊
> 
> Best Regards, Edward

Hi @LindaSummer ,

Sorry, I was very busy in the past two months. I'm still work on this, and I will submit new patch within the a week.

LindaSummer:
> > Hi @donghao526 ,
> > Sorry to interrrupt you. Do you still work on this? I want to continue work on this draft PR if you don't mind. 😊
> > Best Regards, Edward
> 
> Hi @LindaSummer ,
> 
> Sorry, I was very busy in the past two months. I'm still work on this, and I will submit new patch within the a week.

Hi @donghao526 ,

Thanks very much for your effort and contribution! ❤

You could do this anytime when you have time.

Your personal affairs are always in first priority and we don't need to do it in a rush. 😊

Best Regards,
Edward

src/types/redis_tdigest.cc
```diff
@@ -67,18 +67,18 @@ class DummyCentroids {
...
     // because we must guarantee the iterator to be inside the valid range before iteration.
     bool Prev() {
-      if (Valid() && iter_ != centroids_.cbegin()) {
+      if (Valid()) {
         std::advance(iter_, -1);
       }
       return Valid();
     }
-    bool Valid() const { return iter_ != centroids_.cend(); }
+    bool Valid() const { return iter_ < centroids_.cend() && iter_ >= centroids_.cbegin(); }
```
LindaSummer: It seems that `iter_` should always greater or equal to the `cbegin()`.

src/commands/cmd_tdigest.cc
```diff
@@ -176,6 +176,49 @@ class CommandTDigestAdd : public Commander {
...
+
+    if (!result.empty()) {
+      std::vector<std::string> rev_ranks;
+      rev_ranks.reserve(result.size());
+      for (const auto v : result) {
+        rev_ranks.push_back(redis::Integer(v));
+      }
+      *output = redis::Array(rev_ranks);
+    } else {
+      *output = redis::BulkString("nan");
```
LindaSummer: refer to [tdigest.revrank](https://redis.io/docs/latest/commands/tdigest.revrank/).

> All values are -2 if the sketch is empty.

src/types/tdigest.h
```diff
@@ -150,3 +152,70 @@ inline StatusOr<double> TDigestQuantile(TD&& td, double q) {
...
+  for (auto index : indices) {
+    result[index] = static_cast<int>(cumulative_weight);
+  }
+}
+
+template <typename TD>
+inline Status TDigestRevRank(TD&& td, const std::vector<double>& inputs, std::vector<int>& result) {
+  std::map<double, std::vector<size_t>> value_to_indices;
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    value_to_indices[inputs[i]].push_back(i);
```
LindaSummer: it is hard to compare two double number for a map. we need a stable way for the compare operator.

LindaSummer (commented):
Hi @donghao526 ,

Left with some comments.

Please also go thorugh the comments left by github-copilot.

Best Regards,
Edward

src/commands/cmd_tdigest.cc
donghao526: This logic has implemented in the tdigest.RevRank. So if tdigest.RevRank returns Status::OK(), the result can not be empty, I removed the the conditional judgment here.

src/types/redis_tdigest.cc
donghao526: I found when traversing the centroids_ in reverse , if the iter_ == cbegin(), traversal cannot be stopped by using Valid(). But there is indeed a risk of iterator out of bounds here. 

Instead, I decided to use IsBegin() to explicit break the loop.

src/types/tdigest.h
donghao526: Ok, Use DoubleComparator to ensure the two double values are equal

src/commands/cmd_tdigest.cc
```diff
@@ -176,6 +176,45 @@ class CommandTDigestAdd : public Commander {
   std::vector<double> values_;
 };
 
+class CommandTDigestRevRank : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    key_name_ = args[1];
+    inputs_.reserve(args.size() - 2);
+    for (size_t i = 2; i < args.size(); i++) {
```
LindaSummer: Would it be better to unique the input string here and then parse to double?
After unique, we don't need to use the below `map` to group them and it will keep the result stable basing on the command's inoput.

src/types/tdigest.h
```diff
@@ -150,3 +152,93 @@ inline StatusOr<double> TDigestQuantile(TD&& td, double q) {
   diff /= (lc.weight / 2 + rc.weight / 2);
   return Lerp(lc.mean, rc.mean, diff);
 }
+
+inline void AssignRankForEqualInputs(const std::vector<size_t>& indices, double cumulative_weight,
```
LindaSummer: Would it be better that we only allow the `Status TDigestRevRank(TD&& td, const std::vector<double>& inputs, std::vector<int>& result)` in this file only allows unique inputs?

src/types/tdigest.h
```diff
@@ -150,3 +152,93 @@ inline StatusOr<double> TDigestQuantile(TD&& td, double q) {
...
+
+template <typename TD>
+inline Status TDigestRevRank(TD&& td, const std::vector<double>& inputs, std::vector<int>& result) {
+  std::map<double, std::vector<size_t>, DoubleComparator> value_to_indices;
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    value_to_indices[inputs[i]].push_back(i);
+  }
+
+  double cumulative_weight = 0;
+  result.resize(inputs.size());
```
LindaSummer:
```suggestion
  result.clear();
  result.resize(inputs.size(), -2);
```

Maybe it would be better to us `-2` to init the result and add a guard to validate the result.

LindaSummer (commented):
Hi @donghao526 ,

General LGTM.
Left some comments.

Best Regards,
Edward

src/commands/cmd_tdigest.cc
donghao526: Yes, unique the input before calling TDigestRevRank is better

src/commands/cmd_tdigest.cc
```diff
@@ -180,39 +180,47 @@ class CommandTDigestRevRank : public Commander {
...
+    unique_inputs_.reserve(unique_inputs_set.size());
+    size_t i = 0;
+    for (const auto &input : unique_inputs_set) {
+      auto value = ParseFloat(input);
       if (!value) {
         return {Status::RedisParseErr, errValueIsNotFloat};
       }
-      inputs_.push_back(*value);
+      unique_inputs_.push_back(*value);
+      unique_inputs_order_[input] = i++;
```
LindaSummer:
```suggestion
      unique_inputs_order_[input] = i;
      ++i;
```

Making the increment a newline would make the code better for reading.

src/types/tdigest.h
```diff
@@ -150,3 +152,92 @@ inline StatusOr<double> TDigestQuantile(TD&& td, double q) {
...
+  }
+
+  // handle inputs less than minimum
+  while (it != value_to_indices.rend()) {
+    result[it->second] = static_cast<int>(td.TotalWeight());
+    ++it;
+  }
+
+  for (auto r : result) {
+    if (r < -2) {
```
LindaSummer:
```suggestion
    if (r <= -2) {
```

When the centroids are not empty, no revrank should be `-2`.

LindaSummer (commented):
Hi @donghao526 ,

Generally LGTM.
Left two comments.

Thanks for your effort!❤️

Best Regards,
Edward

src/types/tdigest.h
```diff
@@ -150,3 +152,92 @@ inline StatusOr<double> TDigestQuantile(TD&& td, double q) {
...
+struct DoubleComparator {
+  bool operator()(const double& a, const double& b) const { return DoubleCompare(a, b) == -1; }
+};
+
+template <typename TD>
+inline Status TDigestRevRank(TD&& td, const std::vector<double>& inputs, std::vector<int>& result) {
+  std::map<double, size_t, DoubleComparator> value_to_indices;
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    value_to_indices[inputs[i]] = i;
+  }
```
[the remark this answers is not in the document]
donghao526: The inputs have been deduplicated before calling this function, and there are no duplicate input values

src/commands/cmd_tdigest.cc
```diff
@@ -176,6 +176,53 @@ class CommandTDigestAdd : public Commander {
...
+    unique_inputs_.reserve(unique_inputs_set.size());
+    size_t i = 0;
+    for (const auto &input : unique_inputs_set) {
+      auto value = ParseFloat(input);
+      if (!value) {
+        return {Status::RedisParseErr, errValueIsNotFloat};
+      }
+      unique_inputs_.push_back(*value);
+      unique_inputs_order_[input] = i++;
+    }
```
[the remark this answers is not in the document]
donghao526: The origin_inputs_ vector stores the correct order of the origin inputs. So the order of the results is deterministic.

src/commands/cmd_tdigest.cc
```diff
@@ -176,6 +176,54 @@ class CommandTDigestAdd : public Commander {
   std::vector<double> values_;
 };
 
+class CommandTDigestRevRank : public Commander {
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    key_name_ = args[1];
+
+    std::unordered_set<std::string> unique_inputs_set(args.begin() + 2, args.end());
```
LindaSummer:
```suggestion
    std::set<std::string> unique_inputs_set(args.begin() + 2, args.end());
```

We don't need an unordered one here, a normal `std::set` should be ok.

LindaSummer (dismissed):
LGTM, left one nitpit comment.

src/commands/cmd_tdigest.cc
donghao526: @LindaSummer Yeah, std::set is ok. 

This is my first PR for a large project, and I've learned a lot. Thank you😊.

PragmaTwice (approved):

LindaSummer (approved):

Issue #3132, opened 2025-08-20
Support DIFF, DIFF1, ANDOR, and ONE for command BITOP


PragmaTwice:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Refer to https://redis.io/docs/latest/commands/bitop/

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

xuzifu666:
Can you assign this task to me? thanks @PragmaTwice

PragmaTwice:
> Can you assign this task to me? thanks [@PragmaTwice](https://github.com/PragmaTwice)

Sure! Assigned.

ShivamDixit20:
assign this issue to me if still open

PR #3134, opened 2025-08-21
feat(bitop): Support DIFF, DIFF1, ANDOR, and ONE for command BITOP

xuzifu666:
issue: https://github.com/apache/kvrocks/issues/3132


PragmaTwice:
Could you format the code via `./x.py format` and `clang-format-14`?

See https://kvrocks.apache.org/community/contributing for more information.

git-hulk:
@xuzifu666 The CI is now using the version clang-format 14, which behaves differently from clang-format > 14 in some ways.

xuzifu666:
> @xuzifu666 The CI is now using the version clang-format 14, which behaves differently from clang-format > 14 in some ways.

Thansk for reminder and help, I had changed the clang-format version and code format seems OK now. @git-hulk @PragmaTwice

tests/gocase/unit/type/bitmap/bitmap_test.go
```diff
@@ -271,6 +271,17 @@ func TestBitmap(t *testing.T) {
 		require.EqualValues(t, []string{"\x55\xff\x00\xaa"}, GetBitmap(t, rdb, ctx, "s"))
 	})
 
+	t.Run("BITOP DIFF|DIFF1|ANDOR|ONE don't change the string with single input key", func(t *testing.T) {
+		Set2SetBit(t, rdb, ctx, "a", []byte("\x01\x02\xff"))
+		Set2SetBit(t, rdb, ctx, "b", []byte("\x01\x02\xff"))
+		Set2SetBit(t, rdb, ctx, "c", []byte("\x01\x02\xff"))
+		require.NoError(t, rdb.BitOpDiff(ctx, "res1", "a", "b", "c").Err())
+		require.NoError(t, rdb.BitOpDiff1(ctx, "res2", "a", "b", "c").Err())
```
git-hulk, line 279: For the `diff1`, the result bits should be set in one or more of `b` or `c`, but not  in `a`. So the result should be `\x00\x00\x00`? because a, b, c are the same.

xuzifu666: Yes, it should be 0 according to the definition. I transplanted it completely according to the implementation of Redis. I will check the original logic again.

src/types/redis_bitmap.cc
```diff
@@ -554,7 +554,9 @@ rocksdb::Status Bitmap::BitOp(engine::Context &ctx, BitOpFlags op_flag, const st
           for (uint64_t i = 0; i < frag_numkeys; i++) {
             lp[i] = reinterpret_cast<const uint64_t *>(fragments[i].data());
           }
-          memcpy(frag_res.get(), fragments[0].data(), frag_minlen);
+          if (op_flag != kBitOpDiff && op_flag != kBitOpDiff1 && op_flag != kBitOpAndOr) {
```
xuzifu666, line 557: I implemented this based on the Redis solution: https://github.com/redis/redis/pull/13898
However, the results are different. I've been looking for the cause but haven't found it yet. I'm not sure if I've overlooked something.

tests/gocase/unit/type/bitmap/bitmap_test.go
xuzifu666: Had fixed the issue.

xuzifu666:
A new PR follow it https://github.com/apache/kvrocks/pull/3381 and close it.

PR #3138, opened 2025-08-22
feat(Dockerfile): use a numeric UID instead of a username

### LINKED ISSUE #3135: Docker image should have numeric user specification
### Search before asking

### Motivation

I'm trying to ensure that all Docker images used in my project are compatible with the [`runAsNonRoot` option](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#securitycontext-v1-core) in Kubernetes security contexts. While the Kvrocks image is already configured to use a non-root user, Kubernetes doesn't support a string username specification in conjunction with this option (it shows a "container has runAsNonRoot and image has non-numeric user" error).

### Solution

It would be useful if the Docker image was configured with a numeric UID instead (e.g. `USER 999`).

SpecLad:
Also, use hardcoded IDs when creating the user and group, to ensure they remain stable.

Closes #3135.


aleksraiden (approved):

aleksraiden:
Good for me, compatibility with Kuber are important . Thanks, @SpecLad

PragmaTwice (approved):

PR #3139, opened 2025-08-23
feat(scripting): support strict key-accessing mode for lua scripting

PragmaTwice:
Accessing undeclared keys in lua scripting may lead to unexpected behavior in the current design of kvrocks (also in redis, refer to https://redis.io/docs/latest/commands/eval/), so in this PR we add a new option `lua-strict-key-accessing` to prevent users to access undeclared keys in lua, e.g.

```
EVAL "return redis.call('set', 'a', 1)" // ERROR!

EVAL "return redis.call('set', KEYS[1], 1)" 1 a // ok
```

This check is performed in both lua scripting and lua functions.


git-hulk (approved):
LGTM

PR #3145, opened 2025-08-27
perf(storage): eliminate unnecessary `rocksdb::DB::ListColumnFamilies()`

jonahgao:
Historically, it was used for doing some checks.
https://github.com/apache/kvrocks/blob/2bbfe5aa9531f6e76bfd10a2cc450b9bfa0f15d9/src/storage.cc#L175-L182

But these checks no longer exist today, so this operation should be unnecessary.

`ListColumnFamilies()` needs to iterate through the MANIFEST, and might be costly for a large one.
For example, 80MB MANIFEST, took 1.6 seconds.


git-hulk (approved):

git-hulk:
@jonahgao nice catch.

PragmaTwice (approved):

PR #3146, opened 2025-08-27
fix(scan): pattern-based SCAN iterations may skip remaining keys

sryanyuan:
Pattern-based SCAN iterations may skip remaining keys in a hash slot, causing incomplete key retrieval.

**Bug Reproduction:**
- 10 keys in the same hash slot: ["119483", "166988", "210695", "223656", "48063", "59022", "65976", "74937", "88379", "99338"]
- Initial SCAN with pattern `2*`:
  - Returns cursor C1 and **empty keyset** (no keys match `2*`)
  - Records "119483" as last scanned key
- Subsequent SCAN with cursor C1 and same pattern:
  1. RocksDB iterator seeks to "119483"
  2. Calls `Next()` → gets "166988" (next key in slot)
  3. "166988" ∉ `2*` pattern → no key returned
  4. **Error**: Scan incorrectly increments slot index
  5. **Result**: Remaining 8 keys in slot are skipped

**Bug Fix Implementation:**
When scanning with a previous scan cursor and match pattern:
1. If the last scanned key is lexicographically before the pattern's start range:
   → Use the pattern's minimum matching key as the seek key
   → Instead of using the last scanned key
   
**Example:**
- Pattern: `2*` → Minimum matching key = `"2"` (hex: \x32)
- Last scanned key: `"119483"` (hex: \x31\x31...)
- Since `"119483"` < `"2"` lexically:
  ✓ **Correct:** Seek to `"2"` 
  ✗ **Buggy:** Seek to `"119483"`

### DIFF
--- a/src/storage/redis_db.cc
+++ b/src/storage/redis_db.cc
@@ -339,9 +339,13 @@ rocksdb::Status Database::Scan(engine::Context &ctx, const std::string &cursor, 
   }
 
   if (!cursor.empty()) {
-    iter->Seek(ns_cursor);
-    if (iter->Valid()) {
-      iter->Next();
+    if (storage_->IsSlotIdEncoded() && !ns_prefix.empty() && ns_prefix.compare(ns_cursor) > 0) {
+      iter->Seek(ns_prefix);
+    } else {
+      iter->Seek(ns_cursor);
+      if (iter->Valid()) {
+        iter->Next();
+      }
     }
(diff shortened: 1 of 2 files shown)


git-hulk (approved):

git-hulk:
Related issue: https://github.com/apache/kvrocks/issues/3023.

danishgh (commented):
0

PR #3155, opened 2025-09-02
feat(script): make EVAL & FCALL exclusive if lua-strict-key-accessing is disabled

PragmaTwice:
Accessing undeclared keys may lead to unexpected behavior in both Kvrocks and Redis.

Specifically in Kvrocks, undeclared keys will not be protected by key locks, which may face atomic issues of rocksdb IO.

If `lua-strict-key-accessing` is enabled, we can garantee that there is no write on undeclared keys, so EVAL & FCALL in multiple threads should be safe.

But for disabled `lua-strict-key-accessing`, it is quite dangerous if users use undeclared keys in scripting. So we add `exclusive` flag to EVAL, EVALSHA and FCALL commands in such case.

NOTE that *_RO commands are not affected.

### DIFF
--- a/src/commands/cmd_function.cc
+++ b/src/commands/cmd_function.cc
@@ -120,10 +120,17 @@ uint64_t GenerateFunctionFlags(uint64_t flags, const std::vector<std::string> &a
   return flags;
 }
 
-REDIS_REGISTER_COMMANDS(Function,
-                        MakeCmdAttr<CommandFunction>("function", -2, "exclusive no-script", NO_KEY,
-                                                     GenerateFunctionFlags),
-                        MakeCmdAttr<CommandFCall<>>("fcall", -3, "write no-script", GetScriptEvalKeyRange),
-                        MakeCmdAttr<CommandFCall<true>>("fcall_ro", -3, "read-only no-script", GetScriptEvalKeyRange));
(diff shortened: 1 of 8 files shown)


danishgh (approved):

src/commands/cmd_function.cc
```diff
@@ -120,10 +120,17 @@ uint64_t GenerateFunctionFlags(uint64_t flags, const std::vector<std::string> &a
...
+uint64_t GenerateFCallFlags(uint64_t flags, const std::vector<std::string> &, const Config &config) {
+  if (!config.lua_strict_key_accessing) {
+    return flags | kCmdExclusive;
+  }
+
+  return flags;
+}
+
+REDIS_REGISTER_COMMANDS(
+    Function, MakeCmdAttr<CommandFunction>("function", -2, "exclusive no-script", NO_KEY, GenerateFunctionFlags),
```
git-hulk, line 132: We need to remove the `exclusive` flag from the command.

danishgh (approved):

src/commands/cmd_function.cc
PragmaTwice: FUNCTION command is not affected by this PR.

git-hulk: Ohh... I misread the line.

git-hulk (approved):

danishgh (approved):

Issue #3048, opened 2025-07-10
[OSPP 2025] Tracking issues: Support time series data structure and commands


yezhizi:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Part of [OSPP 2025](https://summer-ospp.ac.cn/org/prodetail/259430064?lang=zh&list=pro).
This task is to implement the time series data structure and its commands on Kvrocks. 

We have broken down the task into multiple subtasks and will track them in this issue.

- [x] Initialize metadata.  (#3072)

- [x] Write Operations(`TS.CREATE/TS.ADD/TS.MADD`): Support uncompressed chunk storage.  (#3091 #3107 )

- [x] Query Operations(`TS.INFO`/`TS.GET`/`TS.RANGE`): Implement time-range scans with chunk iteration. (#3133 #3140 #3142 ) 

- [x] Multi-Series Queries(`TS.MGET`/`TS.MRANGE`): Support label-based indexing and filter expressions. (#3164 #3167 )

- [x] Downsampling Rules(`TS.CREATERULE`): Implement downstream_key metadata and aggregation logic. (#3151 )

- [x] Remove/recycle related label index in `Index` CF when a TS key was deleted/expired. (#3175 )

- [x] Retention: Implement retention policy with compaction filters. (#3177 )

- [x] Deletion & Other Write Commands(`TS.DEL`, `TS.INCRBY/DECRBY`): Support deleting a range of time series data. (#3171 #3174 )

- [ ] (Optional) Compressed Chunk Support: Integrate Gorilla compression for timestamps/values.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PragmaTwice:
Proposal can be found here: https://github.com/apache/kvrocks/discussions/3044

LindaSummer:
Hi @yezhizi ,

Thanks for your effort! 😊
When these tasks are ready for the community to participate, we'd better list the dependencies of each sub-task.
It will help us to pick up the available tasks first and work in the expected order in your plan.

Best Regards,
Edward

yezhizi:
Hi! @LindaSummer 

[...]

PR #3164, opened 2025-09-07
feat(ts): Add label-based indexing support and `TS.MGET` command

yezhizi:
Part of #3048 
Add a new CF for label-based indexing. The key format is as following:
```
+-----------+--------------+----------------+------------+------------------+--------------+---------+
|     ns    |  index_type  | label_key_size |  label_key | label_value_size |  label_value |  key    | => null
| (1+Xbytes)|    (1byte)   |    (4byte)     |   (Ybyte)  |   (4byte)        |   (Zbyte)    | (Kbyte) |
+-----------+--------------+----------------+------------+------------------+--------------+---------+
```
| enum | index_type |
| ---- | ---------- |
| 0    | `TS_LABEL`   |

Now we can do:
```
127.0.0.1:6666> TS.CREATE temp:TLV LABELS type temp location TLV
127.0.0.1:6666> TS.CREATE temp:JLM LABELS type temp location JLM
127.0.0.1:6666> TS.MADD temp:TLV 1000 30 temp:TLV 1010 35 temp:TLV 1020 9999 temp:TLV 1030 40
127.0.0.1:6666> TS.MADD temp:JLM 1005 30 temp:JLM 1015 35 temp:JLM 1025 9999 temp:JLM 1035 40
127.0.0.1:6666> TS.MGET WITHLABELS FILTER type=temp
1) 1) "temp:JLM"
   2) 1) 1) "location"
         2) "JLM"
      2) 1) "type"
         2) "temp"
   3) 1) 1) (integer) 1035
         2) (double) 40
2) 1) "temp:TLV"
   2) 1) 1) "location"
         2) "TLV"
      2) 1) "type"
         2) "temp"
   3) 1) 1) (integer) 1030
         2) (double) 40
127.0.0.1:6666> TS.MGET SELECTED_LABELS location FILTER type=temp
1) 1) "temp:JLM"
   2) 1) 1) "location"
         2) "JLM"
   3) 1) 1) (integer) 1035
         2) (double) 40
2) 1) "temp:TLV"
   2) 1) 1) "location"
         2) "TLV"
   3) 1) 1) (integer) 1030
         2) (double) 40
```


danishgh (approved):

src/types/redis_timeseries.h
```diff
@@ -173,6 +174,25 @@ struct TSRangeOption {
   BucketTimestampType bucket_timestamp_type = BucketTimestampType::Start;
 };
 
+struct TSMGetOption {
+  struct FilterOption {
```
PragmaTwice, line 178: It seems the filter is used by both MGET and MRANGE, so is TSMGetOption::FilterOption a good name for it?

src/types/redis_timeseries.h
```diff
@@ -173,6 +174,25 @@ struct TSRangeOption {
...
+    std::unordered_map<std::string, std::set<std::string>> labels_equals;
+    std::unordered_map<std::string, std::set<std::string>> labels_not_equals;
+    std::set<std::string> labels_exists;
+    std::set<std::string> labels_not_exists;
```
PragmaTwice: Could labels_equals {"label", empty set} represent labels_not_exists "label"? So that we can use just two fields for the filter.

src/types/redis_timeseries.h
yezhizi: My consideration is that `TSMRangeOption` can inherit from `TSMGetOption` to use `FilterOption`.  Is that OK?
```cpp
struct TSMRangeOption : TSMGetOption, TSRangeOption {
    // some other fields
};
```

src/types/redis_timeseries.h
yezhizi: Good idea!

src/types/redis_timeseries.h
PragmaTwice: Looks fine to me : )

src/commands/cmd_timeseries.cc
```diff
@@ -784,12 +793,255 @@ class CommandTSGet : public CommandTSAggregatorBase {
   std::string user_key_;
 };
 
+class CommandTSMGetBase : public CommandTSAggregatorBase {
+ public:
+  using FilterOption = TSMGetOption::FilterOption;
+
+  class TSMRangeFilterParser {
```
PragmaTwice: Should we move it to redis_timeseries.h so that we can add some unit tests for it?

yezhizi: OK

PragmaTwice (dismissed):

yezhizi:
@PragmaTwice  Could you tell me how to fix this test error? I don't really understand why it failed.
```
/Users/runner/work/kvrocks/kvrocks/tests/cppunit/compact_test.cc:196: Failure
Value of: sg.IsNotFound()
  Actual: false
Expected: true

[  FAILED  ] 1 test, listed below:
[  FAILED  ] Compact.SearchFilter
```

src/storage/storage.cc
```diff
@@ -355,6 +355,13 @@ Status Storage::Open(DBOpenMode mode) {
   search_opts.disable_auto_compactions = config_->rocks_db.disable_auto_compactions;
   SetBlobDB(&search_opts);
 
+  rocksdb::BlockBasedTableOptions index_table_opts = InitTableOptions();
+  rocksdb::ColumnFamilyOptions index_opts(options);
+  search_opts.table_factory.reset(rocksdb::NewBlockBasedTableFactory(index_table_opts));
+  search_opts.compaction_filter_factory = std::make_shared<IndexFilterFactory>(this);
+  search_opts.disable_auto_compactions = config_->rocks_db.disable_auto_compactions;
```
PragmaTwice: `index_opts` instead of `search_opts` here : )

src/commands/cmd_timeseries.cc
```diff
@@ -126,13 +126,26 @@ std::string FormatCreateRuleResAsRedisReply(TSCreateRuleResult res) {
...
+
 }  // namespace
 
 namespace redis {
 
 class KeywordCommandBase : public Commander {
  public:
-  KeywordCommandBase(size_t skip_num, size_t tail_skip_num) : skip_num_(skip_num), tail_skip_num_(tail_skip_num) {}
+  KeywordCommandBase(size_t skip_num, size_t tail_skip_num) : skip_num_(skip_num), tail_skip_num_(tail_skip_num) {
+    handlers_.reserve(32);  // Avoid Realloc space, or keywords_ may be invalid.
```
PragmaTwice: Maybe you can try deque instead of vector.

deque::push_back will not invalidate element references.

PragmaTwice:
@yezhizi Not necessary in this PR, but we need to consider:
- if a TS key is deleted, how to also delete related label index?
- if a TS key is expired, how to handle it in the compaction filter of the new CF?

Maybe when we are accessing the label index, we should check if the TS key is currently available (not expired or deleted), and in compaction we can recycle label index with expired/deleted TS keys.

PragmaTwice (approved):

yezhizi:
Okay, I’ll try to improve on this.

src/commands/cmd_timeseries.cc
```diff
@@ -784,12 +784,93 @@ class CommandTSGet : public CommandTSAggregatorBase {
...
+class CommandTSMGet : public CommandTSMGetBase {
+ public:
+  CommandTSMGet() : CommandTSMGetBase(0, 0) { registerDefaultHandlers(); }
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() < 3) {
+      return {Status::RedisParseErr, "wrong number of arguments for 'ts.mget' command"};
+    }
+    return CommandTSMGetBase::Parse(args);
+  }
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
```
PragmaTwice, line 829: I think we should exclude the cluster mode here? Maybe just return an error to indicate that it is not supported. cc @yezhizi

yezhizi: Okay, I will submit a new PR to address it later.

PR #3168, opened 2025-09-10
fix(list): check the status of rocksdb iterator

jonahgao:
In case of an I/O error when iterating, abort the operation; otherwise, it may return incomplete results or partially update a list.


jonahgao:
The new code coverage seems difficult to improve unless a fault injection mechanism is introduced🤔.

PragmaTwice (approved):

git-hulk (approved):

PR #3171, opened 2025-09-11
feat(ts): Add `TS.INCRBY` and `TS.DECRBY` command 

yezhizi:
Part of #3048 

In the current implementation, `TS.INCRBY` defaults to using `0` as the timestamp when `TIMESTAMP` is not specified, which differs from Redis' native behavior (using the server's timestamp). 
This feature could be implemented in the future, along with the `TS.ADD key * value` syntax to support automatically using the server's timestamp.

```
127.0.0.1:6666>  TS.INCRBY a 232 TIMESTAMP 1657811829000
(integer) 1657811829000
127.0.0.1:6666> TS.INCRBY a 157 TIMESTAMP 1657811829000
(integer) 1657811829000
127.0.0.1:6666> TS.INCRBY a 432 TIMESTAMP 1657811829000
(integer) 1657811829000
127.0.0.1:6666> ts.get a
1) 1) (integer) 1657811829000
   2) (double) 821
127.0.0.1:6666> ts.incrby a 100 TIMESTAMP 50
(error) ERR timestamp must be equal to or higher than the maximum existing timestamp
127.0.0.1:6666> 
```

### DIFF
--- a/src/commands/cmd_timeseries.cc
+++ b/src/commands/cmd_timeseries.cc
@@ -982,9 +982,6 @@ class CommandTSIncrBy : public CommandTSCreateBase {
  public:
   CommandTSIncrBy() { registerDefaultHandlers(); }
   Status Parse(const std::vector<std::string> &args) override {
-    if (args.size() < 3) {
-      return {Status::RedisParseErr, "wrong number of arguments for 'ts.incrby' command"};
-    }
     CommandParser parser(args, 2);
     auto value_parse = parser.TakeFloat<double>();
     if (!value_parse.IsOK()) {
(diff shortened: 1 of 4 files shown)


src/commands/cmd_timeseries.cc
```diff
@@ -978,6 +978,67 @@ class CommandTSMRange : public CommandTSRangeBase, public CommandTSMGetBase {
   TSMRangeOption option_;
 };
 
+class CommandTSIncrBy : public CommandTSCreateBase {
+ public:
+  CommandTSIncrBy() { registerDefaultHandlers(); }
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() < 3) {
+      return {Status::RedisParseErr, "wrong number of arguments for 'ts.incrby' command"};
+    }
```
PragmaTwice: can be removed. it will be checked in other place.

yezhizi:
Also add `TS.DECRBY` command : )

PragmaTwice (approved):

PR #3175, opened 2025-09-15
feat(ts): Implement `IndexFilter` to remove label indexes of deleted/expired TS keys

yezhizi:


src/storage/compact_filter.cc
```diff
@@ -216,9 +217,35 @@ bool SearchFilter::Filter([[maybe_unused]] int level, const Slice &key, [[maybe_
...
-bool IndexFilter::Filter([[maybe_unused]] int level, [[maybe_unused]] const Slice &key,
-                         [[maybe_unused]] const Slice &value, [[maybe_unused]] std::string *new_value,
-                         [[maybe_unused]] bool *modified) const {
+bool IndexFilter::Filter([[maybe_unused]] int level, const Slice &key, [[maybe_unused]] const Slice &value,
+                         [[maybe_unused]] std::string *new_value, [[maybe_unused]] bool *modified) const {
+  auto db = stor_->GetDB();
+
+  // There only `TS_LABEL` index type for now, so we just handle the `TS_LABEL` index here.
+  // If add more index types in the future, need to distinguish them.
+  auto rev_key = redis::TSRevLabelKey(key);
```
PragmaTwice: I think currently we also need to check if it is `TS_LABEL`, and ignore non-`TS_LABEL` keys (as an early return).

PragmaTwice (approved):

PR #3177, opened 2025-09-16
feat(ts): Support retention in SubKeyFilter for expired TS chunk deletion

yezhizi:
Part of #3048


yezhizi:
Since TSMetadata doesn't keep the latest sample timestamp, compacting a TS chunk key requires a SeekForPrev call each time to get it, which could slow down compactions. 
Maybe we can add a `lastTimestamp` field in TSMetadata to maintain an approximate latest timestamp, updating it whenever we insert a new chunk (same as the current metadata update timing). WDYT? cc @PragmaTwice

PragmaTwice:
> Because TSMetadata doesn't keep the latest sample timestamp, compacting a TS chunk key requires a SeekForPrev call each time to get it, which could slow down compactions.

Would you mind to point out the location of this `SeekForPrev` (and the call stack)?

src/storage/compact_filter.cc
```diff
@@ -132,6 +133,20 @@ bool SubKeyFilter::Filter([[maybe_unused]] int level, const Slice &key, const Sl
...
+    Slice input(cached_metadata_);
+    auto s = ts_metadata.Decode(&input);
+    if (!s.ok()) {
+      error("[compact_filter/subkey] Failed to decode timeseries metadata, namespace: {}, key: {}, err: {}",
+            ikey.GetNamespace(), ikey.GetKey(), s.ToString());
+      return false;
+    }
+    auto timeseries = std::make_unique<redis::TimeSeries>(stor_, ikey.GetNamespace().ToString());
+    engine::Context ctx(stor_);
+    return timeseries->IsChunkExpired(ctx, ikey.GetKey(), ts_metadata, value);
```
yezhizi: In `SubKeyFilter::Filter()`, we should check whether the TS chunk key has exceeded the retention time.  (The code snippet does not yet include a check for whether it is a TS Chunk. Please ignore this for now.)

src/types/redis_timeseries.cc
```diff
@@ -2109,4 +2109,19 @@ rocksdb::Status TimeSeries::Del(engine::Context &ctx, const Slice &user_key, uin
...
+  auto last_ts = chunk->GetLastTimestamp();
+  std::vector<TSSample> get_samples;
+  auto s = getCommon(ctx, ns_key, metadata, true, &get_samples);
+  CHECK(s.ok());
+  if (get_samples.empty()) return true;
+  uint64_t latest_ts = get_samples.back().ts;
+  uint64_t retention_bound =
+      (metadata.retention_time > 0 && metadata.retention_time < latest_ts) ? latest_ts - metadata.retention_time : 0;
+  return last_ts < retention_bound;
+}
```
yezhizi: And here we call `getCommon` to get the latest timestamp to determine whether the chunk has exceeded the retention time.

`getCommon`  will call `SeekForPrev` : https://github.com/apache/kvrocks/blob/4025a6d23698b6bd5383511f259d10988fe7dc0e/src/types/redis_timeseries.cc#L1294-L1309

yezhizi: cc @PragmaTwice

PragmaTwice:
Sorry for the late reply.

I checked the related code and understood the situation. Retention is defined to be the "maximum age for samples compared to the highest reported timestamp" so we need to get the highest ts first. I'm fine to add a new field to metadata (updated while a new chunk inserted).

PragmaTwice:
We should modify the docs for this change? After this is merged.

yezhizi:
> We should modify the docs for this change? After this is merged.

Sure

PragmaTwice (approved):

PR #3182, opened 2025-09-17
fix(ts): Disable `TS.MGET/MRANGE` in cluster mode

yezhizi:
Refer to https://github.com/apache/kvrocks/pull/3164#discussion_r2354171469


PragmaTwice (dismissed):
Looks good to me. Should we also disable users to add label for a ts key in cluster mode?

yezhizi:
> Looks good to me. Should we also disable users to add label for a ts key in cluster mode?

I think it's reasonable to disable users from adding labels, because if labels cannot serve as a secondary index, they are likely meaningless

PragmaTwice (approved):

PR #3193, opened 2025-09-19
fix(command): MONITOR command should redact sensitive tokens

git-hulk:


PragmaTwice:
Is this behavior consistent with Redis?

git-hulk:
> Is this behavior consistent with Redis?

@PragmaTwice The redeacted behavior is the same as Redis. However, Redis will also skip commands with the `admin` flag; it's not necessary for Kvrocks, as Kvrocks has some extra admin commands, such as profiling commands.

~~Another difference is that Redis flagged the function call and eval* as the skip monitor commands, and I haven't investigated the reason for now.~~

torwig (approved):

PR #3194, opened 2025-09-19
perf: optimize command  EXISTS using MultiGet

### LINKED ISSUE #3169: Performance Question: Could EXISTS command benefit from MultiGet optimization?
### Search before asking

### Summary
Hi! 👋 While exploring the codebase, I noticed the `EXISTS` command uses sequential `Get` operations, while other multi-key commands like `MGET` and `MDEL` use `MultiGet`. I'm wondering if this is intentional or if there might be a performance optimization opportunity here.

### Current Implementation
Currently, `Database::existsInternal()` uses a sequential approach:

```cpp
rocksdb::Status Database::existsInternal(engine::Context &ctx, const std::vector<std::string> &keys, int *ret) {
  *ret = 0;
  rocksdb::Status s;
  std::string value;
  for (const auto &key : keys) {  // Sequential Get operations
    s = storage_->Get(ctx, ctx.GetReadOptions(), metadata_cf_handle_, key, &value);
    if (!s.ok() && !s.IsNotFound()) return s;
    if (s.ok()) {
      s = metadata.Decode(value);
      if (!s.ok()) return s;
      if (!metadata.Expired()) *ret += 1;
    }
  }
  return rocksdb::Status::OK();
}
```

### What I observed
Other multi-key commands use `MultiGet`:
- `MDEL` and `MGET` both use `storage_->MultiGet()` for batch operations
- `EXISTS` uses a loop with individual `Get` calls

### Questions
I'm curious about:
1. Is there a specific reason `EXISTS` uses sequential `Get` instead of `MultiGet`?
2. Would a `MultiGet`-based performance optimization be welcome?

### Potential benefits
A `MultiGet`-based implementation could potentially provide:
- Better performance through batch operations (fewer RocksDB calls)
- Consistency with other multi-key commands (`MGET`, `MDEL`, etc.)

### Code location
`src/storage/redis_db.cc` - `Database::existsInternal()` (lines 624-639)

### Solution

_No response_

cyningsun:
Close #3169


src/storage/redis_db.cc
```diff
@@ -208,12 +208,39 @@ rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &k
...
+  std::vector<rocksdb::PinnableSlice> pin_values(slice_keys.size());
+  storage_->MultiGet(ctx, ctx.DefaultMultiGetOptions(), metadata_cf_handle_, slice_keys.size(), slice_keys.data(),
+                     pin_values.data(), statuses.data());
+
+  for (size_t i = 0; i < slice_keys.size(); i++) {
+    if (!statuses[i].ok() && !statuses[i].IsNotFound()) return statuses[i];
+    if (statuses[i].ok()) {
+      Metadata metadata(kRedisNone, false);
+      // Explicit construct a rocksdb::Slice to avoid the implicit conversion from
+      // PinnableSlice to Slice.
```
mapleFU: Why? I don't understand this

mapleFU (dismissed):
General lgtm

src/storage/redis_db.cc
cyningsun: @mapleFU  thanks for your comment!

**If you're referring to the comment (A):** 
I agree it's overly verbose. It can indeed be cleaned up.

**If you're referring to the explicit construction (B):**
The explicit construction is mainly for safety - since `Decode()` calls `remove_prefix()` internally, it would modify the original `pin_values[i]` state if passed directly. It's better to create a temporary `Slice` copy, leaving original data intact, which matches the pattern used elsewhere in the codebase. 

**What's your take on this?** I'd be curious to hear your perspective

mapleFU: The command is ok to me but I think "it would modify the original pin_values[i] state if passed directly" it's more specifically rather than "implicit conversion"?

cyningsun:
```
  [[nodiscard]] virtual rocksdb::Status Decode(Slice *input);
  [[nodiscard]] rocksdb::Status Decode(Slice input);
```

Sorry for my previous incorrect explanation. Actually, due to call the second function of `Metadata` and parameter is passed `by value`, an implicit conversion occurs, which modifies the copied value rather than pin_values itself. Therefore, the comment should be accurate.  The explicit construction has no obvious different with implicit construction, that's what I mean `overly verbose`

cyningsun: Comment is moved from caller to function declaration

src/storage/redis_db.cc
```diff
@@ -208,12 +206,37 @@ rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &k
...
+  std::vector<rocksdb::Status> statuses(slice_keys.size());
+  std::vector<rocksdb::PinnableSlice> pin_values(slice_keys.size());
+  storage_->MultiGet(ctx, ctx.DefaultMultiGetOptions(), metadata_cf_handle_, slice_keys.size(), slice_keys.data(),
+                     pin_values.data(), statuses.data());
+
+  for (size_t i = 0; i < slice_keys.size(); i++) {
+    if (!statuses[i].ok() && !statuses[i].IsNotFound()) return statuses[i];
+    if (statuses[i].ok()) {
+      Metadata metadata(kRedisNone, false);
+      auto s = metadata.Decode(rocksdb::Slice(pin_values[i].data(), pin_values[i].size()));
```
mapleFU: However, PinnableSlice wouldn't be changed if pass by value?

cyningsun: yes, that's correct

src/storage/redis_db.cc
```diff
@@ -208,12 +206,37 @@ rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &k
...
+  std::vector<rocksdb::Status> statuses(slice_keys.size());
+  std::vector<rocksdb::PinnableSlice> pin_values(slice_keys.size());
+  storage_->MultiGet(ctx, ctx.DefaultMultiGetOptions(), metadata_cf_handle_, slice_keys.size(), slice_keys.data(),
+                     pin_values.data(), statuses.data());
+
+  for (size_t i = 0; i < slice_keys.size(); i++) {
+    if (!statuses[i].ok() && !statuses[i].IsNotFound()) return statuses[i];
+    if (statuses[i].ok()) {
+      Metadata metadata(kRedisNone, false);
+      auto s = metadata.Decode(&rocksdb::Slice(pin_values[i].data(), pin_values[i].size()));
```
mapleFU: Hmmm why not just

```
auto s = metadata.Decode(pin_values[i]);
```

cyningsun:
```
 // Explicit construct a rocksdb::Slice to avoid the implicit conversion from
// PinnableSlice to Slice.
```
In fact, it can. The initial implementation is referred to `MDel`'s implementation. That's why there is a annotation. Upon our discussion and analysis, it is indeed not very reasonable. Let me make some modifications

src/storage/redis_db.h
```diff
@@ -155,7 +155,6 @@ class Database {
 
  private:
   // Already internal keys
-  [[nodiscard]] rocksdb::Status existsInternal(engine::Context &ctx, const std::vector<std::string> &keys, int *ret);
```
mapleFU, line 158: I'm so sorry to comment this again and again. But why not:

```
[[nodiscard]] rocksdb::Status existsInternal(engine::Context &ctx, const nonstd::span<Slice> &keys, int *ret) {
  // From benchmark, ...
  if (keys.size() == 1) {
     // Get..
  } else {
    // MultiGet
  }
}
```

git-hulk (approved):

src/storage/redis_db.cc
cyningsun: @mapleFU Thanks so much for commenting—please don't be sorry! I really welcome the discussion, and I'm glad we're digging into the details.

You're absolutely right, and in fact, I completely agree with your thinking. The implementation you're proposing is a more conservative and robust approach. If I were working on this within a company codebase, I would likely advocate for the same solution you've outlined, for a few key reasons:

*   In real-world usage, the number of keys for `EXISTS` maybe smaller than for `MGET` or `MDEL`.
*   With a smaller number of keys and a higher in-memory hit rate, the potential performance difference between a single `Get` and a `MultiGet` becomes more significant.

The main reason I went with the current implementation are:

*   In the benchmark testing(#3169), the performance regression for smaller key lists was measured to be quite small—within 5%. (`the memory to disk ratio is indeed not very high`)
*   `MGET` and `MDEL` also shares the same problem for handling small lists of keys
*   Using the same `MultiGet` path kept the implementation simpler and more elegant.

However, you've made a very valid point that deserves more thorough testing. **Before we merge this, let me run some more benchmarks.** I'll specifically focus on scenarios with a higher cache hit rate and very small key lists to better quantify the impact. Once I have those results, we can make a more data-driven decision on whether to keep the current unified implementation or adopt the optimized approach you suggested for `existsInternal`.

I really appreciate you pushing for the best solution! Let's sync up again after the next round of tests.

src/storage/redis_db.h
cyningsun: @mapleFU Thanks so much for commenting—please don't be sorry! I really welcome the discussion, and I'm glad we're digging into the details.

You're absolutely right, and in fact, I completely agree with your thinking. The implementation you're proposing is a more conservative and robust approach. If I were working on this within a company codebase, I would likely advocate for the same solution you've outlined, for a few key reasons:

*   In real-world usage, the number of keys for `EXISTS` maybe smaller than for `MGET` or `MDEL`.
*   With a smaller number of keys and a higher in-memory hit rate, the potential performance difference between a single `Get` and a `MultiGet` becomes more significant.

The main reason I went with the current implementation are:

*   In the benchmark testing(#3169), the performance regression for smaller key lists was measured to be quite small—within 5%. (`the memory to disk ratio is indeed not very high`)
*   `MGET` and `MDEL` also shares the same problem for handling small lists of keys
*   Using the same `MultiGet` path kept the implementation simpler and more elegant.

However, you've made a very valid point that deserves more thorough testing. **Before we merge this, let me run some more benchmarks.** I'll specifically focus on scenarios with a higher cache hit rate and very small key lists to better quantify the impact. Once I have those results, we can make a more data-driven decision on whether to keep the current unified implementation or adopt the optimized approach you suggested for `existsInternal`.

I really appreciate you pushing for the best solution! Let's sync up again after the next round of tests.

src/storage/redis_db.cc
```diff
@@ -208,12 +208,37 @@ rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &k
...
+  slice_keys.reserve(keys.size());
+
   for (const auto &key : keys) {
     ns_keys.emplace_back(AppendNamespacePrefix(key));
+    slice_keys.emplace_back(ns_keys.back());
+  }
+
+  std::vector<rocksdb::Status> statuses(slice_keys.size());
+  std::vector<rocksdb::PinnableSlice> pin_values(slice_keys.size());
+  storage_->MultiGet(ctx, ctx.DefaultMultiGetOptions(), metadata_cf_handle_, slice_keys.size(), slice_keys.data(),
```
PragmaTwice, line 229: Following your benchmark here https://github.com/apache/kvrocks/issues/3169#issuecomment-3294659021, should we use `Get` instead if `keys.size()` is 1?

cyningsun: Hi @PragmaTwice, we actually just discussed this exact point in another thread. The short answer is that I agree it's a valid optimization. I'll running additional tests for high cache-hit scenarios before deciding whether to implement the conditional Get for a single key.

You can see the full discussion here: https://github.com/apache/kvrocks/pull/3194#discussion_r2365617179

PR #3195, opened 2025-09-19
refactor(storage): optimize MDel with MultiGet options

cyningsun:
1. Replace `ctx.GetReadOptions()` with `ctx.DefaultMultiGetOptions()` in the `Database::MDel` method to use read options specifically optimized for MultiGet operations.


git-hulk (dismissed):

src/storage/redis_db.cc
```diff
@@ -183,17 +183,15 @@ rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &k
...
 
   for (size_t i = 0; i < slice_keys.size(); i++) {
     if (!statuses[i].ok() && !statuses[i].IsNotFound()) return statuses[i];
     if (statuses[i].IsNotFound()) continue;
 
     Metadata metadata(kRedisNone, false);
-    // Explicit construct a rocksdb::Slice to avoid the implicit conversion from
-    // PinnableSlice to Slice.
-    auto s = metadata.Decode(rocksdb::Slice(pin_values[i].data(), pin_values[i].size()));
+    auto s = metadata.Decode(pin_values[i]);
```
PragmaTwice: What's the meaning?

> Pass PinnableSlice directly instead of make copy before passed by value when calling Metadata::Decode

I don't think this is a copy. IMHO it's just a slice/view. cc @mapleFU

PragmaTwice (dismissed):
See the comment.

src/storage/redis_db.cc
PragmaTwice: If this is just a style change, please change the PR title and description.

cyningsun: hi @PragmaTwice  Yes, the expression was not clear enough. It should be stated as avoiding the redundant creation of rocksdb::Slice objects.

`rocksdb::Slice()` creates a rocksdb::Slice object, and then passing it by value to `[[nodiscard]] rocksdb::Status Decode(Slice input);` will create another new rocksdb::Slice object. If a PinnableSlice is passed directly, only one rocksdb::Slice object is created when it is passed by value to `[[nodiscard]] rocksdb::Status Decode(Slice input);`. Compared to the former, the latter avoids creating one additional rocksdb::Slice object. Is this understanding correct?

PragmaTwice: 1. `Slice` is just a trivial 16-byte object and it is very cheap to copy and construct on stack.
2. Due to [Copy Elision](https://en.cppreference.com/w/cpp/language/copy_elision.html) in C++, there is no additional object.

So if you want this change, please change the title to style changes instead of a perf optimization.

cyningsun: @PragmaTwice Thanks for the technical insight. I've updated the PR title and description

mapleFU: > Pass PinnableSlice directly instead of make copy before passed by value when calling Metadata::Decode

@PragmaTwice I think this might because afraid of `Slice*` which changes the value internal.

cyningsun: @PragmaTwice @mapleFU, I've reverted this change due to static code analysis warnings about object slicing when passing `PinnableSlice` directly to functions expecting `Slice` parameters.

mapleFU (approved):

git-hulk (approved):

PR #3219, opened 2025-10-07
feat(replication): add replication-no-slowdown config

zhixinwen:
We need to set `rocksdb.write_options.no_slowdown` to yes in production because if write stall happens, it would block the event loop and KVRocks would not be able to serve any other requests including `INFO`. This means when write stall happens, we would see the cluster frozen and failover happens unnecessarily.

However, a side effect of `rocksdb.write_options.no_slowdown` is it would cause replication failure when write stall happens. The replica would continuously tries to reconnect and restart the process, adding extra load on master.

The solution is to add a separate config to fine control the behavior.

### DIFF
--- a/src/common/status.h
+++ b/src/common/status.h
@@ -181,7 +181,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> : 
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
-      : BaseType((typename StringInStatusOr<U>::BaseType&&)(std::move(v))) {}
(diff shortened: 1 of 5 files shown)


src/common/status.h
```diff
@@ -181,7 +181,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> :
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
-      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
+      : BaseType((typename StringInStatusOr<U>::BaseType&&)(std::move(v))) {}
```
git-hulk: Need to revert this line to pass the format check.

```suggestion
      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
```

PragmaTwice (approved):

PR #3220, opened 2025-10-07
fix(replication): send ack on ping even if no data written

zhixinwen:
In the old code, we only ack when there is data written and `sendReplConfAck` would only send ack if not ack was sent in the last 1 sec. This works fine for continuous traffic. However, if write stops and the last write did not trigger an ack due to 1s rate limit, no more ack would be sent.

If user runs `INFO replication` on master, the user would see a replication lag that never get catch up. In this PR we send ack back whenever replication sees ping, because ping is only sent when master has no data, this should not add extra load.

### DIFF
--- a/src/common/status.h
+++ b/src/common/status.h
@@ -181,7 +181,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> : 
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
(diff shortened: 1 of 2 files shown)


dushenglin (approved):

src/common/status.h
```diff
@@ -181,7 +181,7 @@ struct StringInStatusOr<T, std::enable_if_t<sizeof(T) < sizeof(std::string)>> :
   StringInStatusOr(StringInStatusOr<U>&& v) : BaseType(new std::string(*std::move(v))) {}  // NOLINT
   template <typename U, typename std::enable_if_t<!StringInStatusOr<U>::inplace, int> = 0>
   StringInStatusOr(StringInStatusOr<U>&& v)  // NOLINT
-      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
+      : BaseType((typename StringInStatusOr<U>::BaseType&&)(std::move(v))) {}
```
```suggestion
      : BaseType((typename StringInStatusOr<U>::BaseType &&)(std::move(v))) {}
```

PragmaTwice (approved):

PR #3232, opened 2025-10-23
chore(deps): Bump RocksDB version to v10.7.5

aleksraiden:
Bump RocksDB version to v10.7.5 (see: https://github.com/facebook/rocksdb/releases/tag/v10.7.5)

Please merge after PR #3229 

**Key changes**

- The MultiScan API contract is updated. After a multi scan range got prepared with Prepare API call, the following seeks must seek the start of each prepared scan range in order. In addition, when limit is set, upper bound must be set to the same value of limit before each seek
- Fixed a performance regression in LZ4 compression that started in version 10.6.0
- Add the fail_if_no_udi_on_open flag in BlockBasedTableOption to control whether a missing user defined index block in a SST is a hard error or not
- A new flag memtable_veirfy_per_key_checksum_on_seek is added to AdvancedColumnFamilyOptions. When it is enabled, it will validate key checksum along the binary search path on skiplist based memtable during seek operation
- Introduce option MultiScanArgs::use_async_io to enable asynchronous I/O during MultiScan
- Add new option MultiScanArgs::max_prefetch_size that limits the memory usage of per file pinning of prefetched blocks
- Improved sst_dump
- HyperClockCache with no estimated_entry_charge is now production-ready and is the preferred block cache implementation vs. LRUCache. Please consider updating your code to minimize the risk of hitting performance bottlenecks or anomalies from LRUCache. See cache.h for more detail.
- **RocksDB now requires a C++20 compatible compiler (GCC >= 11, Clang >= 10, Visual Studio >= 2019)**, including for any code using RocksDB headers
- The default provided block cache implementation is now HyperClockCache instead of LRUCache, when block_cache is nullptr (default) and no_block_cache==false (default)
- Add a new experimental PerKeyPointLockManager to improve efficiency under high lock contention
- Majorly improved CPU efficiency and scalability of parallel compression (CompressionOptions::parallel_threads > 1), though this efficiency improvement makes parallel compression
- Add a new table property "rocksdb.key.smallest.seqno" which records the smallest sequence number of all keys in file
- Bug fixes


PragmaTwice:
Same as https://github.com/apache/kvrocks/pull/3229#issuecomment-3435511822.

PragmaTwice:
Note that from this release, C++20 is required. cc @git-hulk @mapleFU

Hence unless c++20 enabled for Kvrocks, we cannot upgrade to this release (or future releases).

git-hulk:
> Note that from this release, C++20 is required. cc @git-hulk @mapleFU
> 
> Hence unless c++20 enabled for Kvrocks, we cannot upgrade to this release (or future releases).

Could we update this rocksdb version after our next release(might be 2.14.0)? because it's about time to release our next version, and we can avoid requiring a compiler upgrade for it.

aleksraiden:
> > Note that from this release, C++20 is required. cc @git-hulk @mapleFU
> > Hence unless c++20 enabled for Kvrocks, we cannot upgrade to this release (or future releases).
> 
> Could we update this rocksdb version after our next release(might be 2.14.0)? because it's about time to release our next version, and we can avoid requiring a compiler upgrade for it.

I think yes, its a good idea

aleksraiden:
As I think, now we can merge this PR into upstream?

git-hulk:
> As I think, now we can merge this PR into upstream?

Yes, I think it's good to go.

khanbilal732 (approved):

khanbilal732 (approved):

PragmaTwice (approved):

torwig (approved):

git-hulk (approved):

PragmaTwice:
There is a stable failure in Alpine 3 CI. Need some investigation.

https://github.com/apache/kvrocks/actions/runs/20409431635/job/58647113948?pr=3232

khanbilal732 (approved):

Issue #322, opened 2021-07-05
any benchmark against redis?


hiqsociety:
any benchmark against redis?

git-hulk:
It makes no sense to compare the Kvrocks with Redis directly since the storage and thread model was totally different. We have a simple performance benchmark on README: https://github.com/KvrocksLabs/kvrocks#performance, from the throughput side, the Kvrocks would be better than Redis in many commands.

hiqsociety:
how much slower do u think kvrocks will be compared with redis?

git-hulk:
It depends on the disk type, the latency of most commands on NVMe should be as good as Redis, SATA disks would much worse but we didn’t have any performance on it.

hengsok:
@hiqsociety I ran a simple test using redis-benchmark tool, I can say that kvrocks perform very well as below using DigitalOcean. Eventhough it persists data to disk, it can be almost as fast as Redis/KeyDB. I reckon if we give it more cpu cores and ram, it would perform even better. Very good job to kvrocks team @git-hulk @ShooterIT. We are awaiting their cluster mode feature and hope they can make their cluster mode to work like Redis whereby it can automatically handle failover and high availability!

Single Node - General Purpose / 16 GB / 4 vCPUs/ NVMe instance
kvrocks

PING_INLINE: 88339.23 requests per second
PING_BULK: 92678.41 requests per second
SET: 58513.75 requests per second
GET: 96525.09 requests per second
INCR: 50968.40 requests per second
LPUSH: 46728.97 requests per second
RPUSH: 46970.41 requests per second
LPOP: 43308.79 requests per second
RPOP: 42973.79 requests per second
SADD: 53390.28 requests per second
HSET: 39261.88 requests per second
SPOP: 63211.12 requests per second
LPUSH (needed to benchmark LRANGE): 45187.53 requests per second
LRANGE_100 (first 100 elements): 44742.73 requests per second
LRANGE_300 (first 300 elements): 21542.44 requests per second
LRANGE_500 (first 450 elements): 15434.48 requests per second
LRANGE_600 (first 600 elements): 12137.39 requests per second
MSET (10 keys): 16835.02 requests per second

I didn't get to save the results for above machine spec for KeyDB which is a forked of Redis but run on multithreads.


4C (Shared), 8GB, NVMe, 10 worker threads
kvrocks

PING_INLINE: 59630.29 requests per second
PING_BULK: 63131.31 requests per second
SET: 34352.46 requests per second
[...]

PR #3237, opened 2025-10-26
ci: update the version of clang-tidy to 18

PragmaTwice:
This is a follow-up PR of #322.


git-hulk:
@PragmaTwice It requires to reformat the code after upgrading to 18.

PragmaTwice:
> @PragmaTwice It requires to reformat the code after upgrading to 18.

seems that there's some difference between my local clang-format-18 and clang-format-18 in CI.

git-hulk:
> > @PragmaTwice It requires to reformat the code after upgrading to 18.
> 
> seems that there's some difference between my local clang-format-18 and clang-format-18 in CI.

That's strange. I also got the same format as your secondary format:

```
-  explicit MoveIterator(Iter iter) : Iter(iter){};
+  explicit MoveIterator(Iter iter) : Iter(iter) {};
```

But the CI behaves the same as the old version(14). One difference is that CI is using `18.1.3` and my local is `18.1.8`.

git-hulk (approved):
LGTM

PR #3240, opened 2025-10-30
perf(cluster): use multithreads to optimize sendSnapshotByRawKV

hll1213181368:
Use multithreads to improve slotRange migration speed of sendSnapshotByRawKV.

### DIFF
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -176,7 +176,7 @@ class CommandTDigestAdd : public Commander {
   std::vector<double> values_;
 };
 
-template <bool reverse>
+template <bool Reverse>
 class TDigestRankCommand : public Commander {
  public:
   Status Parse(const std::vector<std::string> &args) override {
@@ -202,7 +202,16 @@ class TDigestRankCommand : public Commander {
     TDigest tdigest(srv->storage, conn->GetNamespace());
     std::vector<int> result;
     result.reserve(origin_inputs_.size());
-    if (const auto s = tdigest.Rank(ctx, key_name_, unique_inputs_, reverse, result); !s.ok()) {
+
+    if (const auto s =
+            [&]() {
+              if constexpr (Reverse) {
+                return tdigest.RevRank(ctx, key_name_, unique_inputs_, result);
+              } else {
+                return tdigest.Rank(ctx, key_name_, unique_inputs_, result);
+              }
+            }();
+        !s.ok()) {
       if (s.IsNotFound()) {
         return {Status::RedisExecErr, errKeyNotFound};
       }
--- a/src/types/redis_tdigest.cc
+++ b/src/types/redis_tdigest.cc
@@ -47,114 +47,81 @@
 
 namespace redis {
 
+namespace {
+template <bool Reverse, typename Container>
+inline decltype(auto) GetCbeginIter(const Container& centroids) {
+  if constexpr (Reverse) {
+    return centroids.crbegin();
+  } else {
+    return centroids.cbegin();
+  }
+}
+
+template <bool Reverse, typename Container>
+inline decltype(auto) GetCendIter(const Container& centroids) {
+  if constexpr (Reverse) {
+    return centroids.crend();
+  } else {
+    return centroids.cend();
+  }
+}
+}  // namespace
+
 // TODO: It should be replaced by a iteration of the rocksdb iterator
+template <bool Reverse>
 class DummyCentroids {
  public:
-  class BaseIterator {
-   public:
-    virtual ~BaseIterator() = default;
-    virtual bool Next() = 0;
-    virtual bool Prev() = 0;
-    virtual bool Valid() const = 0;
-    virtual std::unique_ptr<BaseIterator> Clone() const = 0;
-    virtual StatusOr<Centroid> GetCentroid() const = 0;
-  };
-
   DummyCentroids(const TDigestMetadata& meta_data, const std::vector<Centroid>& centroids)
       : meta_data_(meta_data), centroids_(centroids) {}
-  class Iterator : public BaseIterator {
+  class Iterator {
    public:
-    Iterator(std::vector<Centroid>::const_iterator&& iter, const std::vector<Centroid>& centroids)
-        : iter_(iter), centroids_(centroids) {}
-    std::unique_ptr<BaseIterator> Clone() const override {
-      if (iter_ != centroids_.cend()) {
-        return std::make_unique<Iterator>(std::next(centroids_.cbegin(), std::distance(centroids_.cbegin(), iter_)),
-                                          centroids_);
+    using IterType = std::conditional_t<Reverse, std::vector<Centroid>::const_reverse_iterator,
+                                        std::vector<Centroid>::const_iterator>;
+    Iterator(IterType iter, const std::vector<Centroid>& centroids) : iter_(iter), centroids_(centroids) {}
+    std::unique_ptr<Iterator> Clone() const {
+      if (iter_ != GetCendIter<Reverse>(centroids_)) {
+        return std::make_unique<Iterator>(
+            std::next(GetCbeginIter<Reverse>(centroids_), std::distance(GetCbeginIter<Reverse>(centroids_), iter_)),
+            centroids_);
       }
-      return std::make_unique<Iterator>(centroids_.cend(), centroids_);
+      return std::make_unique<Iterator>(GetCendIter<Reverse>(centroids_), centroids_);
     }
-    bool Next() override {
+    bool Next() {
       if (Valid()) {
         std::advance(iter_, 1);
       }
-      return iter_ != centroids_.cend();
+      return iter_ != GetCendIter<Reverse>(centroids_);
     }
 
     // The Prev function can only be called for item is not cend,
     // because we must guarantee the iterator to be inside the valid range before iteration.
-    bool Prev() override {
-      if (Valid() && iter_ != centroids_.cbegin()) {
+    bool Prev() {
+      if (Valid() && iter_ != GetCendIter<Reverse>(centroids_)) {
         std::advance(iter_, -1);
       }
       return Valid();
     }
-    bool Valid() const override { return iter_ != centroids_.cend(); }
-    StatusOr<Centroid> GetCentroid() const override {
-      if (iter_ == centroids_.cend()) {
+    bool Valid() const { return iter_ != GetCendIter<Reverse>(centroids_); }
+    StatusOr<Centroid> GetCentroid() const {
+      if (iter_ == GetCendIter<Reverse>(centroids_)) {
         return {::Status::NotOK, "invalid iterator during decoding tdigest centroid"};
       }
       return *iter_;
     }
 
    private:
-    std::vector<Centroid>::const_iterator iter_;
+    IterType iter_;
     const std::vector<Centroid>& centroids_;
   };
 
-  class ReverseIterator final : public BaseIterator {
-   public:
-    ReverseIterator(std::vector<Centroid>::const_reverse_iterator&& iter, const std::vector<Centroid>& centroids)
-        : iter_(iter), centroids_(centroids) {}
-    std::unique_ptr<BaseIterator> Clone() const override {
-      if (iter_ != centroids_.crend()) {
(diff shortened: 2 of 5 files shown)


git-hulk:
@hll1213181368 Do you test how much it will be improved if you increase the number of threads?

hll1213181368:
@git-hulk @PragmaTwice 

Test Case: one kvrocks cluster has 4 shard, every shard has 1TB data.  
Set the migrate parameters : migrate-type is raw-key-value, migrate-batch-rate-limit-mb is 256MB/s，migrate-batch-size-kb is 20MB.

Test result:
1）The migration slotRange time is basically sentSnapshotByRawKV.
2）The time consumption of sentSnapshotByRawKV can be mainly divided into four steps: 
    1. Iterate query of rocksdb data; 
    2. Rate limitor; 
    3. SockSend：Sending data to the target node（APPLYBATCH）; 
    4. SockReadLine：Determine whether the data migrate successfully to the target node was .
3）migrate-batch-rate-limit-mb：use to ratelimitor to get token, when the  migrate-batch-rate-limit-mb is 256MB/s, the ratelitor time consumption is almost negligible. And It cannot improve the processing speed of SockSend and SockReadLine.  =》The migrate parameters are already the optimal solution.
4）try to use multithreads to optimize sendSnapshotByRawKV.
5）Test 10 slotRange. The following is a comparison and analysis of the time consumption before and after using multithreading.

image

src/cluster/slot_migrate.cc
```diff
@@ -1260,8 +1263,51 @@ Status SlotMigrator::sendSnapshotByRawKV() {
   auto slot_range = slot_range_.load();
   info("[migrate] Migrating snapshot of slot(s) {} by raw key value", slot_range.String());
 
-  auto prefix = ComposeSlotKeyPrefix(namespace_, slot_range.start);
-  auto upper_bound = ComposeSlotKeyUpperBound(namespace_, slot_range.end);
+  int total_slots = slot_range.end - slot_range.start + 1;
+  unsigned int max_parallelism = std::thread::hardware_concurrency();
+  unsigned int available_slots = static_cast<unsigned int>(total_slots);
```
git-hulk: I think this line is unnecessary; you can just cast max_parallelism to `int`.

src/cluster/slot_migrate.cc
```diff
@@ -1260,8 +1263,51 @@ Status SlotMigrator::sendSnapshotByRawKV() {
...
+  int parallelism = static_cast<int>(std::min(max_parallelism, available_slots));
+  int base = total_slots / parallelism;
+  int rem = total_slots % parallelism;
```
git-hulk: Those two should be `slots_per_thread` and `remain_slots?`

src/cluster/slot_migrate.cc
```diff
@@ -1260,8 +1263,51 @@ Status SlotMigrator::sendSnapshotByRawKV() {
...
+
+  std::vector<std::future<Status>> results;
+  int cur_start = slot_range.start;
+  for (int i = 0; i < parallelism; i++) {
+    int count = base + (i < rem ? 1 : 0);
+    int cur_end = cur_start + count - 1;
+    if (count <= 0) break;
```
git-hulk: When the count would be 0?

src/cluster/slot_migrate.cc
```diff
@@ -1260,8 +1263,51 @@ Status SlotMigrator::sendSnapshotByRawKV() {
...
+
+    results.emplace_back(std::async(std::launch::async, [=]() -> Status {
+      int fd = createConnectToDstNode();
+      if (fd < 0) {
+        return Status(Status::NotOK, fmt::format("thread[{}] createConnectToDstNode failed", i));
```
git-hulk:
```suggestion
        return Status(Status::NotOK, fmt::format("failed to connect the destination node in thread[{}]", i));
```

src/cluster/slot_migrate.cc
```diff
@@ -1260,8 +1263,51 @@ Status SlotMigrator::sendSnapshotByRawKV() {
...
+      close(fd);
+      return s;
+    }));
+
+    cur_start = cur_end + 1;
+  }
+
+  // Wait til finish
+  for (auto &result : results) {
+    auto s = result.get();
```
git-hulk: What if one of the threads fails? Others are still migrating?

src/cluster/slot_migrate.cc
```diff
@@ -1325,16 +1370,44 @@ Status SlotMigrator::sendSnapshotByRawKV() {
...
-  info(
-      "[migrate] Succeed to migrate snapshot range, slot(s): {}, elapsed: {} ms, sent: {} bytes, rate: {:.2f} kb/s, "
-      "batches: {}, entries: {}",
-      slot_range.String(), elapsed, batch_sender.GetSentBytes(), batch_sender.GetRate(start_ts),
-      batch_sender.GetSentBatchesNum(), batch_sender.GetEntriesNum());
-
   return Status::OK();
 }
 
+int SlotMigrator::createConnectToDstNode() {
```
git-hulk: You can just `SockConnect` instead of creating another one.

src/cluster/slot_migrate.cc
```diff
@@ -1272,12 +1318,11 @@ Status SlotMigrator::sendSnapshotByRawKV() {
   auto no_txn_ctx = engine::Context::NoTransactionContext(storage_);
   engine::DBIterator iter(no_txn_ctx, read_options);
 
-  BatchSender batch_sender(*dst_fd_, migrate_batch_size_bytes_, migrate_batch_bytes_per_sec_);
+  BatchSender batch_sender(fd, migrate_batch_size_bytes_, migrate_batch_bytes_per_sec_ / parallelism);
```
git-hulk: `migrate_batch_bytes_per_sec_ / parallelism`, this might affect the migration performance when the keys exist in a few slots. The better way is to share a global rate limiter between threads.

src/cluster/slot_migrate.cc
```diff
@@ -1325,16 +1370,44 @@ Status SlotMigrator::sendSnapshotByRawKV() {
...
+  int one = 1;
+  setsockopt(fd, IPPROTO_TCP, TCP_NODELAY, &one, sizeof(one));
+
+  if (::connect(fd, (struct sockaddr *)&addr, sizeof(addr)) != 0) {
+    error("connect to {}:{} failed: {}", dst_ip_, dst_port_, strerror(errno));
+    close(fd);
+    return -1;
+  }
+
+  // add fd authOnDstNode
```
PragmaTwice: if you want to add a comment, make sure that it is readable by other people.

src/cluster/slot_migrate.cc
```diff
@@ -1260,8 +1263,51 @@ Status SlotMigrator::sendSnapshotByRawKV() {
   auto slot_range = slot_range_.load();
   info("[migrate] Migrating snapshot of slot(s) {} by raw key value", slot_range.String());
 
-  auto prefix = ComposeSlotKeyPrefix(namespace_, slot_range.start);
-  auto upper_bound = ComposeSlotKeyUpperBound(namespace_, slot_range.end);
+  int total_slots = slot_range.end - slot_range.start + 1;
+  unsigned int max_parallelism = std::thread::hardware_concurrency();
```
PragmaTwice: better to have an option instead of just querying nproc.

PragmaTwice (commented):
Since this is a performance improvement, I think it would be great if a benchmark (before and after this patch) is attached.

src/cluster/slot_migrate.cc
hll1213181368: yes, count can not be 0

src/cluster/slot_migrate.cc
hll1213181368: if one of the threads fails, Others stop.

src/cluster/slot_migrate.cc
hll1213181368: OK

src/cluster/slot_migrate.cc
git-hulk: @hll1213181368 Seems this comment was resolved? Did I miss anything?

git-hulk:
@hll1213181368 There are still two comments to be resolved, and you won't be able to mark them as resolved without any answer or actions.

For https://github.com/apache/kvrocks/pull/3240#discussion_r2484096543, we need to handle the partial failure case.

For https://github.com/apache/kvrocks/pull/3240#discussion_r2484237567, you should support a configuration, or explain what do you think.

hll1213181368:
> @hll1213181368 There are still two comments to be resolved, and you won't be able to mark them as resolved without any answer or actions.
> 
> For [#3240 (comment)](https://github.com/apache/kvrocks/pull/3240#discussion_r2484096543), we need to handle the partial failure case.
> 
> For [#3240 (comment)](https://github.com/apache/kvrocks/pull/3240#discussion_r2484237567), you should support a configuration, or explain what do you think.

@git-hulk @PragmaTwice I have response this two comment. Do you have good idea?

git-hulk:
@hll1213181368 I didn't see that. Please check if your reply comments are in pending status.

hll1213181368:
if one of the threads fails, Others thread still migrate. But the total result is NotOK. This slotRange migrate failed. we need to analysis the problem or retry to migrate this slotRange.

@git-hulk Do you think so? Or do you have any good idea?

hll1213181368:
@PragmaTwice how to set threads for this migrate task. I don't think more better idea.
how do you think this idea, The config of kvrocks to add parameters. 

for example: migrate-fixed-thread、migrate-min-threads、migrate-max-threads.

```
int calculate_parallelism(const SlotRange& slot_range, const ThreadConfig& config) {
    int total_slots = slot_range.end - slot_range.start + 1;
    
    // prefer to use migrate_fixed_threads
    if (config.fixed_threads > 0) {
        return std::min(config.migrate_fixed_threads, total_slots);
    }
    
    unsigned int hardware_threads = std::thread::hardware_concurrency();
    int detected_threads = hardware_threads > 0 ? static_cast<int>(hardware_threads) : 1;
    
    // configuration constraints between migrate_min_threads and migrate_max_threads
    int parallelism = detected_threads;
    if (config.max_threads > 0) {
        parallelism = std::min(parallelism, config.migrate_max_threads);
    }
    parallelism = std::max(parallelism, config.migrate_min_threads);
    
    return std::min(parallelism, total_slots);
}
```


@git-hulk @PragmaTwice  Later I will commit another pr. parallel to improve AsyncScanDBSize task. Similar to this deal principle.

git-hulk:
> > @hll1213181368 I didn't see that. Please check if your reply comments are in pending status.
> 
> @PragmaTwice how to set threads for this migrate task. I don't think more better idea. image
> 
> @git-hulk @PragmaTwice Later I will commit another pr. parallel to improve AsyncScanDBSize task. Similar to this deal principle.

Those comments are pending and not visible to others. Please release them via the comment button.

hll1213181368:
@git-hulk  Above I have reply comments. Better to solve one thread fails is to try migrate this failed slotRange again.

git-hulk:
> if one of the threads fails, Others thread still migrate. But the total result is NotOK. This slotRange migrate failed. we need to analysis the problem or retry to migrate this slotRange.
> 
> @git-hulk Do you think so? Or do you have any good idea?

I'm good with that allows the partial failure while doing the parallel migration. But the current implementation doesn't respect this behavior. We should collect and wait for all migration threads to finish, then check whether each is successful, partially successful, or a failure.

hll1213181368:
> > if one of the threads fails, Others thread still migrate. But the total result is NotOK. This slotRange migrate failed. we need to analysis the problem or retry to migrate this slotRange.
> > @git-hulk Do you think so? Or do you have any good idea?
> 
> I'm good with that allows the partial failure while doing the parallel migration. But the current implementation doesn't respect this behavior. We should collect and wait for all migration threads to finish, then check whether each is successful, partially successful, or a failure.

@git-hulk The kvrocks-controller sends a range slots migration request, and the kvrocks node returns either success or failure result for the entire range. For a failed range, the kvrocks-controller is called again to retry the entire range. Successful slots within the range will be returned as successfully migrated, while failed slots will be migrated again. I think it's unnecessary for the kvrocks node to return which slots in the range succeeded and which failed.

PR #3243, opened 2025-10-31
fix(replication): fix stuck replication due to wrong watermark

zhixinwen:
In order to reduce unnecessary callbacks, we set read watermark based on data size:
https://github.com/apache/kvrocks/blob/2b7e69a5acd9e1490b96aafc84746d1baef6579e/src/cluster/replication.cc#L700

However, this watermark is not reset after reading the data, which could result in stuck replication. For example:
1. A batch of size 16kb is received by replica and watermark is set to 16kb
2. A PING was received after the batch by the replica and the callback returns without resetting the watermark. Replica would continue to process data only after master sends at least 16kb data.
3. Master only sent 1kb data, and that data is not processed by replica until master sends enough PING to replica to trigger the 16kb watermark. This causes the replication to get stuck.

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -680,6 +680,8 @@ ReplicationThread::CBState ReplicationThread::incrementBatchLoopCB(bufferevent *
           if (data_written) {
             sendReplConfAck(bev, force_ack);
           }
+          // We should reset the watermark to 0 to read the next RESP parts after reading a batch.
+          bufferevent_setwatermark(bev, EV_READ, 0, 0);          
           return CBState::AGAIN;
         }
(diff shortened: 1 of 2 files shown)


git-hulk:
@zhixinwen Thanks for your fix. Have you ever tested how much the performance will be affected if we don't set the low watermark?

zhixinwen:
> @zhixinwen Thanks for your fix. Have you ever tested how much the performance will be affected if we don't set the low watermark?

I don't remeber the exact details TBH, especially after I have done several other optimizations.
My gut feeling is it may not matter a lot since replica is IO bound not CPU bound.

git-hulk (approved):

PragmaTwice (approved):

PR #3248, opened 2025-11-05
fix(replication): fix _getack send check

zhixinwen:
The original check was wrong. It should use the end of a write batch instead of the beginning sequence.

### DIFF
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -214,7 +214,8 @@ void FeedSlaveThread::loop() {
     //    kMaxDelayUpdates than latest sequence.
     if (is_first_repl_batch || batches_bulk.size() >= max_delay_bytes_ || updates_in_batches >= max_delay_updates_ ||
         srv_->storage->LatestSeqNumber() - batch.sequence <= max_delay_updates_) {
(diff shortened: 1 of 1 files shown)


PragmaTwice:
It will be cherry-picked to 2.14.0 (#3247) after merging.

git-hulk (dismissed):

src/cluster/replication.cc
```diff
@@ -214,7 +214,8 @@ void FeedSlaveThread::loop() {
     //    kMaxDelayUpdates than latest sequence.
     if (is_first_repl_batch || batches_bulk.size() >= max_delay_bytes_ || updates_in_batches >= max_delay_updates_ ||
         srv_->storage->LatestSeqNumber() - batch.sequence <= max_delay_updates_) {
-      if (shouldSendGetAck(batch.sequence)) {
+      // get the last sequence number of the batch, because WAIT uses the last sequence number to wake up the connection.
```
git-hulk:
```suggestion
      // get the last sequence number of the batch, because WAIT uses
      // the last sequence number to wake up the connection.
```

git-hulk (approved):

PragmaTwice (approved):

PR #3249, opened 2025-11-06
feat(tdigest): add the support of TDIGEST.RANK command

### LINKED ISSUE #3062: TDigest: Implement `TDIGEST.RANK` command
### Search before asking

### Motivation

We should implement `TDIGEST.RANK` command.

Redis command reference:
https://redis.io/docs/latest/commands/tdigest.rank/

### Solution

Depends on #2791 #2792.

Follow the implementation of #2794 for this command.

donghao526:
# ISSUE
It closes #3062 

## Proposed Changes
1. Add support of TDIGEST.RANK command
2. Refactor the REVRANK and RANK to same logic with two different iterators for a cleaner implementation
3. Add related cpp and golang unit tests


LindaSummer:
Hi @donghao526 ,

Thanks very much for your great work!

I will go through this PR this weekend. 😊

Best Regards,
Edward

src/commands/cmd_tdigest.cc
```diff
@@ -176,8 +176,9 @@ class CommandTDigestAdd : public Commander {
   std::vector<double> values_;
 };
 
-class CommandTDigestRevRank : public Commander {
+class TDigestRankCommand : public Commander {
```
LindaSummer, line 180: Maybe we could use template to dispatch the commands like below.
The `reverse_` could be a compile time constraint.
https://github.com/apache/kvrocks/blob/77b01b6fed23d2763500756d74d9e6e31deb932d/src/commands/cmd_script.cc#L29-L57

src/types/redis_tdigest.cc
```diff
@@ -93,11 +101,59 @@ class DummyCentroids {
     const std::vector<Centroid>& centroids_;
   };
 
-  std::unique_ptr<Iterator> Begin() { return std::make_unique<Iterator>(centroids_.cbegin(), centroids_); }
-  std::unique_ptr<Iterator> End() {
+  class ReverseIterator final : public BaseIterator {
+   public:
+    ReverseIterator(std::vector<Centroid>::const_reverse_iterator&& iter, const std::vector<Centroid>& centroids)
```
LindaSummer, line 106: Current implementation is ok to me.
If we want a cleaner implementation, maybe we could use crtp to constraint the `std::vector<Centroid>::const_reverse_iterator` as an compile time interface with `cbegin()` and `cend()`.
This is optional.

src/types/tdigest.h
```diff
@@ -171,67 +172,85 @@ struct DoubleComparator {
...
 
   result.clear();
   result.resize(inputs.size(), -2);
-  auto it = value_to_indices.rbegin();
 
-  // handle inputs larger than maximum
-  while (it != value_to_indices.rend() && it->first > td.Max()) {
-    result[it->second] = -1;
-    ++it;
+  using ForwardIter = typename decltype(value_to_index)::iterator;
```
LindaSummer: Maybe we could make here a `constexpr if` here or a template dispatching. The `reverse` value could be evaluated in compile time.

tests/gocase/unit/type/tdigest/tdigest_test.go
```diff
@@ -587,4 +587,73 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
 			require.EqualValues(t, rank, expected[i])
 		}
 	})
+
```
LindaSummer, line 590: Maybe we could add a case with more different or same and unordered inputs to have a comprehensive integration test.

LindaSummer (commented):
Hi @donghao526 ,

Thanks very much for your effort! 👍

Left some commets.

Sorry that my server crashed yesterday😵, and delayed the review for one day.

Best Regards,
Edward

src/types/redis_tdigest.cc
donghao526: > Current implementation is ok to me. If we want a cleaner implementation, maybe we could use crtp to constraint the `std::vector<Centroid>::const_reverse_iterator` as an compile time interface with `cbegin()` and `cend()`. This is optional.

@LindaSummer Since I am not very familiar with CRTP, modifying this code will take some time. If possible, I can submit the changes in a new PR later.

donghao526:
Hi @LindaSummer 

Thank you for taking the time to review my code😊. I have made changes.
1. Use a template to dispatch the CommandTDigestRevRank and CommandTDigestRank commands.
2. Use a constexpr to enable the reverse value to be evaluated at compile time within TDigestRankImpl.
3. Add a case with more different or same and unordered inputs to have a comprehensive integration test.

Best Regards,
Hao

PragmaTwice (approved):
Looks good. cc @LindaSummer

src/types/redis_tdigest.cc
LindaSummer: Hi @donghao526 ,
Of course. This is an optional suggestion. 😊

src/types/redis_tdigest.cc
```diff
@@ -244,7 +300,7 @@ rocksdb::Status TDigest::RevRank(engine::Context& ctx, const Slice& digest_name,
   }
 
   auto dump_centroids = DummyCentroids(metadata, centroids);
-  auto status = TDigestRevRank(dump_centroids, inputs, result);
+  auto status = TDigestRank(dump_centroids, inputs, reverse, result);
```
LindaSummer, line 303: We could make all the flags for `reverse` a compile-time argument.
It's an optional suggestion.
We could do it in a future refactor.😊

LindaSummer (commented):
HI @donghao526 ,

LGTM🎉

This iterator pattern would help all similar commands in TDIGEST.

Best Regards,
Edward

LindaSummer (approved):

donghao526:
> HI @donghao526 ,
> 
> LGTM🎉
> 
> This iterator pattern would help all similar commands in TDIGEST.
> Thanks very much for your effort! ❤
> 
> Best Regards,
> Edward

Thanks for your review.🫡

Best regards,
Hao

src/types/redis_tdigest.cc
donghao526: Ok☺️

PR #3262, opened 2025-11-15
feat(ts): add the support of TWA aggregator to Range and MRange

### LINKED ISSUE #3217: TimeSeries: Support `twa` aggregator for range queries
### Search before asking

### Motivation

When using `TS.RANGE/TS.MRANGE` for range queries, it supports specifying `twa` (time-weighted average) as the aggregator

Redis reference:
https://redis.io/docs/latest/commands/ts.range/
[Twa details](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redistimeseries/redistimeseries-1.8-release-notes/#whats-new-in-18)

### Solution

Refer to others aggregator like `sum`, `avg`
Refer to [this code snippets](https://github.com/apache/kvrocks/blob/8f04af34e804587a8f5bfe206937bb20e21d187b/src/types/redis_timeseries.cc#L770)

var-nan:
It closes #3217 .
`rangeCommon()` in redis_timeseries.cc will add two samples, `prev_sample` and `next_sample` to the `res` vector when the aggregator is twa. These two samples are used to compute the area of the polygons that are at the front of the first bucket and at the end of last bucket.

`prev_sample` is the biggest sample in the data with timestamp less than or equal to first sample of filtered range and `next_sample` is the smallest sample in the data with timestamp greater than or equal to the last sample of filtered range.

TODO: test TWA with FILTER_BY_TS/FILTER_BY_VALUE option.


var-nan:
[279c9bd](https://github.com/apache/kvrocks/pull/3262/commits/279c9bdd457782ffa0b12d637d9b59398719abcd) will correctly calculate TWA when samples are filtered with `FILTER_BY_TS/FITLER_BY_VALUE`.

PragmaTwice:
```
/home/runner/work/kvrocks/kvrocks/src/types/redis_timeseries.cc:91:70: error: variable 'prev_available' is not initialized [cppcoreguidelines-init-variables,-warnings-as-errors]
Suppressed 8485 warnings (8464 in non-user code, 21 NOLINT).
   91 |   bool is_twa_aggregator = aggregator.type == TSAggregatorType::TWA, prev_available, next_available;
2 warnings treated as errors
      |                                                                      ^             
      |                                                                                     = false
/home/runner/work/kvrocks/kvrocks/src/types/redis_timeseries.cc:91:86: error: variable 'next_available' is not initialized [cppcoreguidelines-init-variables,-warnings-as-errors]
   91 |   bool is_twa_aggregator = aggregator.type == TSAggregatorType::TWA, prev_available, next_available;
      |                                                                                      ^             
      |                                                                                                     = false
```

The CI failed due to a clang-tidy report in the changes. Could you fix it to pass the CI?

var-nan:
Some of these errors didn't show up when I ran `./x.py check tidy`  locally; I guess it's due to a version change.

yezhizi:
Sorry for the wait! Been a bit busy these days. I'll review this later today. : )

src/types/redis_timeseries.cc
```diff
@@ -72,10 +72,33 @@ struct Reducer {
                                           [](const TSSample &a, const TSSample &b) { return a.v < b.v; });
     return max->v - min->v;
   }
+  static inline double Twa(nonstd::span<const TSSample> samples) {
```
yezhizi: The name `twa` might not be appropriate here since it calculates an area, not a “average”. Maybe `Area` is more intuitive.

src/types/redis_timeseries.cc
```diff
@@ -111,6 +134,7 @@ std::vector<TSSample> AggregateSamplesByRangeOption(std::vector<TSSample> sample
     if (option.is_return_empty && spans[i].empty()) {
       switch (aggregator.type) {
         case TSAggregatorType::SUM:
+        case TSAggregatorType::TWA:
         case TSAggregatorType::COUNT:
           sample.v = 0;
           break;
```
yezhizi: Here the result is not 0 when specify the `EMPTY` flag. It should be:
> 	Average value over the bucket's timeframe based on linear interpolation of the last sample before the bucket's start and the first sample after the bucket's end. NaN when no such samples.

Refer to [[EMPTY] (since RedisTimeSeries v1.8)](https://redis.io/docs/latest/commands/ts.range/)

src/types/redis_timeseries.cc
```diff
@@ -1103,6 +1206,13 @@ rocksdb::Status TimeSeries::rangeCommon(engine::Context &ctx, const Slice &ns_ke
     }
   }
 
+  if (is_twa_aggregator) {
+    // prev_sample might not get initialized, if first element is in first bucket.
+    prev_sample = prev_sample.ts == TSSample::MAX_TIMESTAMP ? temp_results.front() : prev_sample;
+    temp_results.push_back(prev_sample);
+    temp_results.push_back(next_sample);
+  }
```
yezhizi: `next_sample` might not get initialized also here.

src/types/redis_timeseries.cc
```diff
@@ -126,6 +150,71 @@ std::vector<TSSample> AggregateSamplesByRangeOption(std::vector<TSSample> sample
...
+          r = next_available ? bucket_right : spans[i].back().ts;
+        } else {
+          size_t x = non_empty_left_bucket(i), y = non_empty_right_bucket(i);
+          area += spans[i].front().ts != bucket_left ? front_area(bucket_left, spans[x].back(), spans[i].front()) : 0;
+          area += spans[i].back().ts != bucket_right ? end_area(bucket_right, spans[i].back(), spans[y].front()) : 0;
+        }
+        sample.v += area;
+        sample.v /= std::max(static_cast<double>(r - l), 1.0);
+      }
     } else {
```
yezhizi: This part is a bit complex to follow. Maybe we can split it into two steps:
1. Get the `prev` and `next` for each bucket
2. Calculate the TWA for each bucket.

For step 1, we can do it before the loop starts, and focus on calculating the TWA within the for loop.
Like this:
```
// Some container to store the prev and next of each bucket...
// Get the prev and next for each bucket...
// ...

for (size_t i = 0; i < spans.size(); i++) {
...

// Calculate the TWA for spans[i]

...
}
```
This might be a helpful way to organize this code logic better. : )

var-nan:
Thanks @yezhizi . I was about to push the clang-tidy fixes.

src/types/redis_timeseries.cc
```diff
@@ -97,7 +101,114 @@ std::vector<TSSample> AggregateSamplesByRangeOption(std::vector<TSSample> sample
...
+  /// Computes area of polygon from start of the current bucket to the first sample of the current span.
+  /// Total Area = Area of bottom rectangle + Area of above triangle.
+  auto front_area = [](uint64_t bucket_left, const TSSample &prev, const TSSample &curr) {
+    auto x = static_cast<double>(bucket_left - prev.ts);  // Distance from
+    auto y = static_cast<double>(curr.ts - prev.ts);
+    auto z = curr.v - prev.v;
+    auto triangle_area = (z * (y - (x * x) / y)) / 2;
+    auto rect_area = static_cast<double>(y - x) * prev.v;
+    return triangle_area + rect_area;
+  };
```
yezhizi: Here, we can use `Reducer::Area` to calculate. We can using linear interpolation to compute the value at the boundary, `boundary_v`. And then call `Reducer::Area({bucket_left, boundary_v}, curr)` to obtain the area.

src/types/redis_timeseries.cc
```diff
@@ -97,7 +101,114 @@ std::vector<TSSample> AggregateSamplesByRangeOption(std::vector<TSSample> sample
...
+  /// Computes area of polygon from the last sample of the current span to the end of current bucket.
+  /// Total Area = Area of bottom rectangle + Area of above triangle.
+  auto end_area = [](uint64_t bucket_right, const TSSample &curr, const TSSample &next) {
+    auto x = static_cast<double>(bucket_right - curr.ts);
+    auto y = static_cast<double>(next.ts - curr.ts);
+    auto z = next.v - curr.v;
+    auto rect_area = x * curr.v;
+    auto triangle_area = (x * x * z) / (2 * y);
+    return triangle_area + rect_area;
+  };
```
yezhizi: Same as `front_area`

src/types/redis_timeseries.cc
```diff
@@ -97,7 +101,114 @@ std::vector<TSSample> AggregateSamplesByRangeOption(std::vector<TSSample> sample
...
+    }
+    // Process last bucket.
+    TSSample sample;
+    sample.ts = bucket_left;
+    if (bucket_left == option.end_ts) {  // Calculate last sample.
+      double y_diff = next_sample.v - prev_sample.v;
+      auto x_diff = static_cast<double>(next_sample.ts - prev_sample.ts);
+      auto x_prime_diff = static_cast<double>(option.end_ts - prev_sample.ts);
+      double y_prime_diff = (x_prime_diff * y_diff) / x_diff;
+      sample.v = y_prime_diff + prev_sample.v;
```
yezhizi: It seems this is also using linear interpolation to calculate the value at an intermediate point? Maybe we should extract this logic into a separate function so that it can be easily called elsewhere, such as in the `front_area` above.

src/types/redis_timeseries.cc
```diff
@@ -97,7 +101,114 @@ std::vector<TSSample> AggregateSamplesByRangeOption(std::vector<TSSample> sample
...
+  if (is_twa_aggregator && option.is_return_empty && samples.empty()) {
+    const bool early_return = prev_sample.ts == TSSample::MAX_TIMESTAMP || next_sample.ts == TSSample::MAX_TIMESTAMP ||
+                              prev_sample.ts == next_sample.ts;  // When filter entire range lies left or right to data.
+    if (early_return) {
+      res = std::move(samples);
+      return res;
+    }
+    // Both prev and next should be available. Total range should be in between the prev and next samples.
+    assert(prev_sample.ts <= option.start_ts && option.end_ts <= next_sample.ts);
+
```
yezhizi: Remove this

src/types/redis_timeseries.cc
```diff
@@ -97,7 +101,114 @@ std::vector<TSSample> AggregateSamplesByRangeOption(std::vector<TSSample> sample
...
+  std::vector<std::pair<TSSample, TSSample>> neighbors;
+  neighbors.reserve(spans.size());
+  for (size_t i = 0; i < spans.size(); i++) {
+    TSSample prev = (i != 0) ? spans[non_empty_left_bucket_idx(i)].back() : prev_sample;
+    TSSample next = (i != (spans.size() - 1)) ? spans[non_empty_right_bucket_idx(i)].front() : next_sample;
+    neighbors.emplace_back(prev, next);
+    assert(spans[i].empty() ||
+           (neighbors[i].first.ts <= spans[i].front().ts && spans[i].back().ts <= neighbors[i].second.ts));
+  }  // Should follow: neighbors[i].first <= span[i].front() <= span[i].back() <= neighbors[i].second;
```
yezhizi: Why is there an `assert` here? Under what case would this assertion fail?

var-nan: I used that for debugging purposes. I'll remove that.

src/types/redis_timeseries.cc
```diff
@@ -97,7 +101,96 @@ std::vector<TSSample> AggregateSamplesByRangeOption(std::vector<TSSample> sample
...
+  if (is_twa_aggregator) {
+    const bool discard_boundaries = !option.filter_by_ts.empty() || option.filter_by_value.has_value();
+    next_sample = samples.back();
+    samples.pop_back();
+    prev_sample = samples.back();
+    samples.pop_back();
+    // When FILTER_BY_TS/FILTER_BY_VALUE is enabled, discard out-of-boundary samples.
+    prev_available = discard_boundaries ? false : !samples.empty() && (samples.front().ts != prev_sample.ts);
+    next_available = discard_boundaries ? false : !samples.empty() && (samples.back().ts != next_sample.ts);
+  }
```
yezhizi, line 133: I think we can pass `next_sample`, `next_available`, etc. as function parameters instead of calculating them inside the `AggregateSamplesByRangeOption` function. For example, we can add a struct:
```
struct TWABounds {
    std::optional<TSSample> prev_sample;
    std::optional<TSSample> next_sample;
};
```
And modify the function interface to:
```
AggregateSamplesByRangeOption(std::vector<TSSample> samples, const TSRangeOption &option, const TWABounds&)
```

src/types/redis_timeseries.cc
```diff
@@ -97,7 +101,96 @@ std::vector<TSSample> AggregateSamplesByRangeOption(std::vector<TSSample> sample
...
+  };
+
+  std::vector<std::pair<TSSample, TSSample>> neighbors;
+  neighbors.reserve(spans.size());
+  for (size_t i = 0; i < spans.size(); i++) {
+    TSSample prev = (i != 0) ? spans[non_empty_left_bucket_idx(i)].back() : prev_sample;
+    TSSample next = (i != (spans.size() - 1)) ? spans[non_empty_right_bucket_idx(i)].front() : next_sample;
+    neighbors.emplace_back(prev, next);
+  }
+
```
yezhizi, line 197: The nested while loops inside the for loop result in O(N^2) time complexity, which can be optimized to O(N). We could:
1. Iterate from 0 to N to resolve all prev neighbors by maintaining a "last seen non-empty" variable.
2. Iterate from N to 0 to resolve all next neighbors similarly.

src/types/redis_timeseries.cc
var-nan: Hi @yezhizi the `prev_sample` and `next_sample` are properly initialized as neighbors to the filtered range, as long as the entire dataset is not empty. In other cases, those two are initialized with MAX_TIMESTAMP.

I declared the struct as

``` 
struct TWABounds{
    TSSample prev_sample;
    TSSample next_sample;
}
```


Is there any other need for them to be declared as `optional`

yezhizi: Using `MAX_TIMESTAMP` is fine. I just think `std::optional` has clearer semantics to explicitly indicate the presence or absence of a boundary sample. : )

jihuayu:
@var-nan Hi there, I noticed there hasn't been much activity on this PR for a while. Just checking in—are you still working on this?

PR #3268, opened 2025-11-20
refactor(tdigest): convert the flags for reverse in TDigest::Rank to a compile-time argument

donghao526:
The reverse flag in the current `TDigest::Rank` implementation is a runtime value. Therefore, it should be converted to a compile-time argument.


src/types/tdigest.h
```diff
@@ -172,7 +172,7 @@ struct DoubleComparator {
 };
 
 template <typename TD, bool Reverse>
```
PragmaTwice:
```suggestion
template <bool Reverse, typename TD>
```

src/types/redis_tdigest.h
```diff
@@ -132,4 +253,39 @@ class TDigest : public SubKeyScanner {
...
+    }
+  }
+
+  std::vector<Centroid> centroids;
+  if (auto status = dumpCentroids(ctx, ns_key, metadata, &centroids); !status.ok()) {
+    return status;
+  }
+
+  auto dump_centroids = DummyCentroids(metadata, centroids);
+  auto status = TDigestRank<DummyCentroids&, reverse>(dump_centroids, inputs, result);
```
PragmaTwice:
```suggestion
  auto status = TDigestRank<reverse>(dump_centroids, inputs, result);
```

src/types/redis_tdigest.h
```diff
@@ -132,4 +253,39 @@ class TDigest : public SubKeyScanner {
                                              Centroid* centroid) const;
 };
 
+template <bool reverse>
```
PragmaTwice:
```suggestion
template <bool Reverse>
```

src/types/redis_tdigest.h
```diff
@@ -33,6 +33,126 @@
...
+      }
+      return *iter_;
+    }
+
+   private:
+    std::vector<Centroid>::const_iterator iter_;
+    const std::vector<Centroid>& centroids_;
+  };
+
+  class ReverseIterator final : public BaseIterator {
```
PragmaTwice: Could we have something like `template <bool Reverse> class Iterator` instead of virtual functions?

LindaSummer:
Hi @donghao526 ,

I will go through this PR in three days.

Best Regards,
Edward

src/types/redis_tdigest.h
donghao526: Of course, I have refactored the code.

donghao526:
> Hi @donghao526 ,
> 
> Thanks very much for your effort! ❤ I will go through this PR in three days.
> 
> Best Regards, Edward

Hi @LindaSummer 

Thanks a lot. You can review the code at your convenience.😊

src/types/redis_tdigest.h
```diff
@@ -33,6 +33,115 @@
...
+      : meta_data_(meta_data), centroids_(centroids) {}
+  class Iterator {
+   public:
+    using IterType = std::conditional_t<Reverse, std::vector<Centroid>::const_reverse_iterator,
+                                        std::vector<Centroid>::const_iterator>;
+
+    Iterator(IterType iter, const std::vector<Centroid>& centroids) : iter_(iter), centroids_(centroids) {}
+    std::unique_ptr<Iterator> Clone() const {
+      if constexpr (Reverse) {
+        if (iter_ != centroids_.crend()) {
```
LindaSummer: Maybe we could add a private method like below.
```cpp
 class Ietrator {
...

private:

template <typename Container>
decltype(auto) get_cbegin_iter(Container centroids) const {
  if constexpr (Reverse) {
    return centroids.rcbegin();
  } else {
    return centroids.cbegin() ;
  }
}

template <typename Container>
decltype(auto) get_cend_iter(Container centroids) const {
  if constexpr (Reverse) {
    return centroids.rcend();
  } else {
    return centroids.cend() ;
  }
}
```

Then we could unify the `centroids.cebgin()` and `centroids.cend()`.
This could help us reduce the `constexpr if` in every iterator getter.

src/types/redis_tdigest.h
```diff
@@ -132,4 +242,39 @@ class TDigest : public SubKeyScanner {
...
+  }
+
+  std::vector<Centroid> centroids;
+  if (auto status = dumpCentroids(ctx, ns_key, metadata, &centroids); !status.ok()) {
+    return status;
+  }
+
+  auto dump_centroids = DummyCentroids<Reverse>(metadata, centroids);
+  auto status = TDigestRank<Reverse>(dump_centroids, inputs, result);
+  if (!status) {
```
LindaSummer: We could follow sonarqube and use the init-statement.

```suggestion
  if (auto status = TDigestRank<Reverse>(dump_centroids, inputs, result); !status) {
```

src/types/tdigest.h
```diff
@@ -266,13 +266,4 @@ inline Status TDigestRankImpl(TD&& td, const std::vector<double>& inputs, std::v
...
-
-template <typename TD>
-inline Status TDigestRank(TD&& td, const std::vector<double>& inputs, bool reverse, std::vector<int>& result) {
-  if (reverse) {
-    return TDigestRankImpl<TD, true>(std::forward<TD>(td), inputs, result);
-  } else {
-    return TDigestRankImpl<TD, false>(std::forward<TD>(td), inputs, result);
-  }
-}
+}
```
LindaSummer: Add an endline for this file.

LindaSummer (commented):
Hi @donghao526 ,

Thanks very much for your contribution! ❤

Generally LGTM.

Left some comments.

Best Regards,
Edward

src/types/redis_tdigest.h
donghao526: Good suggestion, I will refactor it.
Thanks!😊

src/types/redis_tdigest.h
```diff
@@ -132,4 +228,38 @@ class TDigest : public SubKeyScanner {
                                              Centroid* centroid) const;
 };
 
+template <bool Reverse>
+rocksdb::Status TDigest::Rank(engine::Context& ctx, const Slice& digest_name, const std::vector<double>& inputs,
```
PragmaTwice: I think maybe we can just expose two non-template function TDigest::Rank and TDigest::RevRank so that we can put the definition of this function into the .cc file instead of header?

PragmaTwice: Not necesary in this PR.

PragmaTwice (commented):
Generally looks good. I'll leave it to @LindaSummer.

src/types/redis_tdigest.h
```diff
@@ -33,6 +33,101 @@
...
+      if constexpr (Reverse) {
+        return centroids.crend();
+      } else {
+        return centroids.cend();
+      }
+    }
+  };
+
+  std::unique_ptr<Iterator> Begin() const {
+    if constexpr (Reverse) {
```
LindaSummer:
```suggestion
    return std::make_unique<Iterator>(getCbeginIter(centroids_), centroids_);
```

Maybe here could also use the util.

src/types/redis_tdigest.h
```diff
@@ -33,6 +33,101 @@
...
+      return std::make_unique<Iterator>(centroids_.crbegin(), centroids_);
+    } else {
+      return std::make_unique<Iterator>(centroids_.cbegin(), centroids_);
+    }
+  }
+  std::unique_ptr<Iterator> End() const {
+    if (centroids_.empty()) {
+      if constexpr (Reverse) {
```
LindaSummer: Same as above.

LindaSummer (dismissed):
Hi @donghao526 ,

Thanks very much for your contribution! ❤

LGTM. Left some optional nitpick comments.

Best Regards,
Edward

src/types/redis_tdigest.h
donghao526: As the Begin and End functions are outside the Iterator class, I made getCbeginIter and getCendIter static. Is this a good solution?

LindaSummer: I think the getbegin and getend are both stateless, static is ok. 
maybe they could be a free inline function protected by a `detail` private namespace if you prefer this way.

src/types/redis_tdigest.h
```diff
@@ -33,6 +33,93 @@
 #include "tdigest.h"
 
 namespace redis {
+
+namespace detail {
```
LindaSummer: To reduce conflict with other symbols in `redis` namespace, it would be better guarded by a namespace like `tdigest`.

After we refactor this in cc file, this could just be in an anonymous namespace.

LindaSummer (commented):
Hi @donghao526 ,

Thanks very much for your contribution!❤

LGTM, left one comment for namespace design.

Best Regards,
Edward

src/types/redis_tdigest.h
donghao526: Hi @LindaSummer 

1. I have exposed two non-template functions, TDigest::Rank and TDigest::RevRank, and placed their definitions in the .cc file.
2. I have placed the GetCBeginIter and GetCEndIter functions inside an anonymous namespace.

Thanks a lot!

LindaSummer (approved):
Hi @donghao526 ,

Thanks very much for your contribution! ❤

LGTM

Best Regards,
Edward

donghao526:
> Hi @donghao526 ,
> 
> Thanks very much for your contribution! ❤
> 
> LGTM
> 
> Best Regards,
> Edward

Thanks for your review😊

LindaSummer:
Hi @donghao526 ,

The CI has some issue and I'm working on #3290 .

This PR will be merged after this issue solved.

Best Regards,
Edward

PragmaTwice (approved):

PR #3271, opened 2025-11-25
fix(info): the slave lag in INFO command might be overflowed

greatsharp:
we found the master_repl_offset is smaller than slave offset during exec "info replication" command,  then the lag value is overflowed

eg:
#Replication
role:master
connected_slaves:2
slave0:xxxxxx
slave1:ip=xxxxx,port=xxxxx,offset=190907912,lag=18446744073709551607
master_repl_offset:190907903

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -1214,21 +1214,25 @@ Server::InfoEntries Server::GetReplicationInfo() {
     entries.emplace_back("slave_priority", config_->slave_priority);
   }
 
-  int idx = 0;
-  rocksdb::SequenceNumber latest_seq = storage->LatestSeqNumber();
-
+  std::vector<std::tuple<std::string, uint32_t, uint64_t>> slave_repl_seq;
   {
     std::shared_lock<std::shared_mutex> guard(slave_threads_mu_);
     entries.emplace_back("connected_slaves", slave_threads_.size());
     for (const auto &slave : slave_threads_) {
       if (slave->IsStopped()) continue;
 
-      entries.emplace_back(
-          "slave" + std::to_string(idx),
-          fmt::format("ip={},port={},offset={},lag={}", slave->GetConn()->GetAnnounceIP(),
-                      slave->GetConn()->GetAnnouncePort(), slave->GetAckSeq(), latest_seq - slave->GetAckSeq()));
-      ++idx;
+      slave_repl_seq.emplace_back(slave->GetConn()->GetAnnounceIP(), slave->GetConn()->GetAnnouncePort(),
+                                  slave->GetAckSeq());
     }
+  }
+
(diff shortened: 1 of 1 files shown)


src/server/server.cc
```diff
@@ -1214,23 +1214,27 @@ Server::InfoEntries Server::GetReplicationInfo() {
...
-  rocksdb::SequenceNumber latest_seq = storage->LatestSeqNumber();
-
+  std::vector<std::tuple<std::string, uint32_t, uint64_t>> slave_repl_seq;
   {
     std::shared_lock<std::shared_mutex> guard(slave_threads_mu_);
     entries.emplace_back("connected_slaves", slave_threads_.size());
     for (const auto &slave : slave_threads_) {
       if (slave->IsStopped()) continue;
 
-      entries.emplace_back(
```
git-hulk, line 1226: @greatsharp Could we just simply normalize the lag to `0` if the latest sequence number < slave reply offset?

git-hulk (dismissed):

git-hulk:
@greatsharp Thanks for your contribution!

PragmaTwice (changes requested):
I think `latest_seq = storage->LatestSeqNumber()` should be put after the guard. So that the lag between master seq and replica seq can be more small.

src/server/server.cc
```diff
@@ -1215,24 +1215,25 @@ Server::InfoEntries Server::GetReplicationInfo() {
...
+    rocksdb::SequenceNumber latest_seq = storage->LatestSeqNumber();
     for (const auto &slave : slave_threads_) {
       if (slave->IsStopped()) continue;
 
+      auto slave_ack_seq = slave->GetAckSeq();
       entries.emplace_back(
           "slave" + std::to_string(idx),
           fmt::format("ip={},port={},offset={},lag={}", slave->GetConn()->GetAnnounceIP(),
-                      slave->GetConn()->GetAnnouncePort(), slave->GetAckSeq(), latest_seq - slave->GetAckSeq()));
+                      slave->GetConn()->GetAnnouncePort(), slave_ack_seq >= latest_seq ? latest_seq : slave_ack_seq,
```
PragmaTwice, line 1230:
```suggestion
                      slave->GetConn()->GetAnnouncePort(), slave_ack_seq,
```
slave_ack_seq should be enough here?

greatsharp: if slave_ack_seq > latest_seq, then offset > master_repl_offset, but the lag is 0, it looks like mismatch

PragmaTwice: Sounds reasonable.

PragmaTwice (approved):

git-hulk (approved):

PR #3277, opened 2025-11-30
fix(stream): fix XPENDING bug and add exclusive range support

i18nsite:
[English](#en) / [中文](#cn)

---

<a id="en"></a>
### Summary
This PR fixes a critical bug in the `XPENDING` command and implements the missing exclusive range support feature introduced in Redis 6.2.0.

### Issues Fixed
Fixes #3276

#### 1. Bug Fix: Empty Results When start_id == end_id
**Problem**: `XPENDING` incorrectly returned an empty list when querying a single entry (start_id equals end_id), even though the pending entry existed.

**Root Causes**:
- **Parsing Error**: In `cmd_stream.cc`, `CommandXPending::Parse` incorrectly used `start_id` to parse the `end_id` parameter
- **Exclusive Upper Bound**: In `redis_stream.cc`, `GetPendingEntries` used an exclusive upper bound for iteration, excluding entries at `end_id`

**Fixes Applied**:
- Corrected the parsing logic to use `end_id` parameter correctly
- Modified iteration bounds to be inclusive of `end_id`  
- Added optimization for single-entry lookup (direct Get instead of iteration)
- Optimized loop comparison using `rocksdb::Slice` to avoid string allocation

#### 2. Feature: Exclusive Range Support
**Implementation**: Added support for exclusive ranges using `(` prefix as specified in Redis 6.2.0 documentation.

**Changes**:
- Added `exclude_start` and `exclude_end` flags to `StreamPendingOptions` struct
- Updated `CommandXPending::Parse` to detect and handle `(` prefix
- Modified `GetPendingEntries` iteration logic to respect exclusion flags

**Usage Examples**:
```
XPENDING mystream mygroup (1-0 +      # Exclude start
XPENDING mystream mygroup - (100-0    # Exclude end  
XPENDING mystream mygroup (1-0 (100-0 # Exclude both
```

### Testing
Added comprehensive test suite in `tests/gocase/unit/type/stream/xpending_test.go` covering:
- Bug reproduction case (start_id == end_id)
- Summary form
- Extended form with range and count
- Consumer filtering
- Idle time filtering
- Exclusive ranges (all combinations)

All tests pass successfully.

### Files Changed
- `src/commands/cmd_stream.cc`: Fixed parsing and added exclusive range handling
- `src/types/redis_stream.cc`: Fixed iteration bounds and optimized single-entry lookup
- `src/types/redis_stream_base.h`: Added exclusion flags to StreamPendingOptions
[...]

PR #3279, opened 2025-11-30
fix(stream): the wrong end id in XPENDING command

i18nsite:
fix xpending end_id

### DIFF
--- a/src/commands/cmd_stream.cc
+++ b/src/commands/cmd_stream.cc
@@ -868,7 +868,7 @@ class CommandXPending : public Commander {
       }
 
       if (end_id != "+") {
-        auto s = ParseStreamEntryID(start_id, &options_.end_id);
+        auto s = ParseStreamEntryID(end_id, &options_.end_id);
         if (!s.IsOK()) {
           return s;
         }


src/commands/cmd_stream.cc
```diff
@@ -868,7 +868,7 @@ class CommandXPending : public Commander {
       }
 
       if (end_id != "+") {
-        auto s = ParseStreamEntryID(start_id, &options_.end_id);
+        auto s = ParseStreamEntryID(end_id, &options_.end_id);
```
git-hulk, line 871: @i18nsite, could you please add a Go test case to prevent this from happening in the future?

i18nsite: https://github.com/apache/kvrocks/pull/3277/files#diff-09c52b90be671f76754318d6b53aaaa01daf53200c7fc4e6eb921e453366c4fdR66 这个pull request中包含了这个的测试，我这里就不想单独搞了

PragmaTwice (changes requested):
Several issues that blocks it to be merged:
- CI are failing due to this change.
- No test case.

git-hulk:
@PragmaTwice CI failure is not related to this PR, but the test case is required.

i18nsite:
算了，稍等，我让ai拆分下

PR #3280, opened 2025-12-01
fix(ci): the sonar c/cpp action has been deprecated

git-hulk:


.github/workflows/sonar.yaml
```diff
@@ -33,8 +33,10 @@ jobs:
           repository: ${{ github.event.workflow_run.head_repository.full_name }}
           ref: ${{ github.event.workflow_run.head_sha }}
           fetch-depth: 0
-      - name: Install sonar-scanner and build-wrapper
-        uses: SonarSource/sonarcloud-github-c-cpp@v3
+      - name: 'Install Build Wrapper'
+        uses: SonarSource/sonarqube-scan-action/install-build-wrapper@v6.0.0
+      - name: 'Install SonarQube Scanner'
+        uses: SonarSource/sonarqube-scan-action@v6
```
PragmaTwice: The fix is not correct. We should follow the README of https://github.com/SonarSource/sonarqube-github-c-cpp.

But IMHO now we can just replace it with sonarqube-github-c-cpp.

.github/workflows/sonar.yaml
```diff
@@ -33,8 +33,10 @@ jobs:
           repository: ${{ github.event.workflow_run.head_repository.full_name }}
           ref: ${{ github.event.workflow_run.head_sha }}
           fetch-depth: 0
-      - name: Install sonar-scanner and build-wrapper
-        uses: SonarSource/sonarcloud-github-c-cpp@v3
+      - name: 'Install Build Wrapper'
+        uses: SonarSource/sonarqube-scan-action/install-build-wrapper@v6.0.0
+      - name: 'Install SonarQube Scanner'
+        uses: SonarSource/sonarqube-scan-action@v6.0.0
```
PragmaTwice: This is to perform the scan instead of install something.

PragmaTwice:
https://github.com/apache/kvrocks/blob/unstable/.github/workflows/sonar.yaml

This action should also be fixed. IMHO now the simplest way is to use sonarqube-github-c-cpp.

git-hulk:
> https://github.com/apache/kvrocks/blob/unstable/.github/workflows/sonar.yaml
> 
> This action should also be fixed. IMHO now the simplest way is to use sonarqube-github-c-cpp.

This requires adding this archived action to Apache's allowed list. Let me try the `SonarSource/sonarqube-scan-action`.

PragmaTwice:
> https://github.com/apache/kvrocks/blob/unstable/.github/workflows/sonar.yaml

Ahh sorry, wrong file. https://github.com/apache/kvrocks/blob/unstable/.github/workflows/kvrocks.yaml this should also be fixed.

git-hulk:
> > https://github.com/apache/kvrocks/blob/unstable/.github/workflows/sonar.yaml
> 
> Ahh sorry, wrong file. https://github.com/apache/kvrocks/blob/unstable/.github/workflows/kvrocks.yaml this should also be fixed.

Oh, I see. I did miss this file, which also has the Sonar build wrapper.

PragmaTwice (approved):

PR #3294, opened 2025-12-13
feat: add redis-databases option to support SELECT command

### LINKED ISSUE #3291: Proposal: Support `SELECT <dbid>` as an alias for `AUTH <dbid>` when namespace mapping is configured
### Search before asking

### Motivation

Hi Kvrocks Team,

Kvrocks already supports authenticating namespaces using numeric dbid when users configure mappings like:

```
namespace.ns0 0
namespace.ns1 1
...
namespace.ns15 15
```

This allows:

```
AUTH 3
```

to switch the active namespace.

💡 Feature Request

I propose that Kvrocks should support:

```
SELECT <dbid>
```

as a direct alias of:

```
AUTH <dbid>
```

when dbid → namespace mapping exists.

Why?

Unsupported command `SELECT` in Kvrocks, All db are some.

But since Kvrocks already has a dbid → namespace mapping mechanism for AUTH, it is natural to reuse it for SELECT.

Expected behavior

If mapping exists:

```
namespace.ns3 3
```

Then:

```
SELECT 3
```

should internally behave exactly like:

```
AUTH 3
```

If the dbid is not configured → return an authentication error.

📘 Related Work

I built a small proxy that implements this translation:

SELECT dbid → AUTH dbid

Proxy code:
https://github.com/hanxi/kvrocks-proxy/blob/main/main.go

But ideally this compatibility behavior should be built into Kvrocks directly.

Related issue: https://github.com/apache/kvrocks/issues/492

🙏 Request for Comments

Before preparing a PR, I’d like feedback from the community:

1. Is mapping SELECT <dbid> to AUTH <dbid> acceptable?

2. Should this behavior be always enabled or require a config flag?

3. Any concerns about side effects when supporting Redis clients?

Thanks!

### Solution

_No response_

hanxi:
Closes #3291

Implements redis-select-compatible configuration option that enables Redis SELECT command support by mapping database indices to namespaces. When enabled, SELECT 0-15 commands are translated to namespace switches, allowing legacy Redis clients to work seamlessly with Kvrocks.

Key Implementation:
- redis-select-compatible: Enable/disable SELECT command support (default: no)
- redis-select-db-prefix: Namespace prefix for database mapping (default: __db)
- redis-select-db-count: Number of logical databases (default: 16, range: 1-10000)
- SELECT <dbid> internally calls SetNamespace({prefix}{dbid}) to switch context
- Validates namespace conflicts on startup to prevent data corruption

Changes:
- Add three new readonly config options in kvrocks.conf
- Implement SELECT command handler with namespace-based switching
- Add startup validation to check namespace conflicts
- Add comprehensive configuration documentation

This provides native SELECT command support without external proxies, maintaining compatibility with Redis clients while leveraging Kvrocks' namespace architecture.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -388,20 +388,12 @@ max-bitmap-to-string-mb 16
 # Default: yes
 redis-cursor-compatible yes
 
-# Number of databases for Redis SELECT compatibility mode.
-# When set to a value greater than 0, enables Redis SELECT command compatibility.
-# When set to 0 (default), SELECT command always returns OK but does not switch databases.
+# Number of Redis-compatible databases.
+# When set to 0 (default), SELECT command acts like a no-op i.e. Redis-compatible databases are disabled.
+# When set to a positive value, SELECT command is Redis-compatible and the namespace feature is disabled.
 #
-# With a value of 16, you can use SELECT 0 through SELECT 15.
-# Each database index is mapped to a namespace name with "db" prefix.
-# DB 0 uses the default namespace, DB 1 uses "db1", DB 2 uses "db2", and so on.
-#
-# IMPORTANT NOTES:
-# 1. This is a READONLY configuration and can only be set at startup
-# 2. Setting this to 0 disables SELECT compatibility mode (default behavior)
-# 3. Setting this to a value > 0 enables SELECT compatibility mode
-# 4. Valid range is 0 to INT_MAX
-# 5. Cannot be used together with namespace feature (namespace tokens)
+# For example, if you set redis-databases to 16, then SELECT 0 through SELECT 15 are available.
+# Note that SELECT 0 always maps to the default namespace, regardless of whether this feature is enabled.
 #
 # Default: 0
 redis-databases 0
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -393,8 +393,8 @@ redis-cursor-compatible yes
 # When set to 0 (default), SELECT command always returns OK but does not switch databases.
 #
 # With a value of 16, you can use SELECT 0 through SELECT 15.
-# Each database index is mapped to a namespace name using the index number directly.
-# DB 0 uses the default namespace, DB 1 uses "1", DB 2 uses "2", and so on.
+# Each database index is mapped to a namespace name with "db" prefix.
+# DB 0 uses the default namespace, DB 1 uses "db1", DB 2 uses "db2", and so on.
 #
 # IMPORTANT NOTES:
 # 1. This is a READONLY configuration and can only be set at startup
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -393,7 +393,7 @@ redis-cursor-compatible yes
 # When set to a positive value, SELECT command is Redis-compatible and the namespace feature is disabled.
 #
 # For example, if you set redis-databases to 16, then SELECT 0 through SELECT 15 are available.
-# Note that SELECT 0 always maps to the default namespace, regardless of whether this feature is enabled.
+# Note that SELECT 0 maps to the default namespace even when this feature is enabled.
 #
 # Default: 0
 redis-databases 0
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -393,14 +393,14 @@ redis-cursor-compatible yes
 # When set to 0 (default), SELECT command always returns OK but does not switch databases.
 #
 # With a value of 16, you can use SELECT 0 through SELECT 15.
-# Each database index is mapped to a namespace name with format "db001", "db002", etc.
-# DB 0 uses the default namespace, DB 1 uses "db001", DB 2 uses "db002", and so on.
+# Each database index is mapped to a namespace name using the index number directly.
+# DB 0 uses the default namespace, DB 1 uses "1", DB 2 uses "2", and so on.
 #
 # IMPORTANT NOTES:
 # 1. This is a READONLY configuration and can only be set at startup
 # 2. Setting this to 0 disables SELECT compatibility mode (default behavior)
 # 3. Setting this to a value > 0 enables SELECT compatibility mode
-# 4. Valid range is 0 to 10000
+# 4. Valid range is 0 to INT_MAX
 # 5. Cannot be used together with namespace feature (namespace tokens)
 #
 # Default: 0
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -388,6 +388,24 @@ max-bitmap-to-string-mb 16
 # Default: yes
 redis-cursor-compatible yes
 
+# Number of databases for Redis SELECT compatibility mode.
+# When set to a value greater than 0, enables Redis SELECT command compatibility.
+# When set to 0 (default), SELECT command always returns OK but does not switch databases.
+#
+# With a value of 16, you can use SELECT 0 through SELECT 15.
+# Each database index is mapped to a namespace name with format "db001", "db002", etc.
+# DB 0 uses the default namespace, DB 1 uses "db001", DB 2 uses "db002", and so on.
+#
+# IMPORTANT NOTES:
+# 1. This is a READONLY configuration and can only be set at startup
+# 2. Setting this to 0 disables SELECT compatibility mode (default behavior)
+# 3. Setting this to a value > 0 enables SELECT compatibility mode
+# 4. Valid range is 0 to 10000
+# 5. Cannot be used together with namespace feature (namespace tokens)
+#
+# Default: 0
+redis-databases 0
+
 # Whether to enable the RESP3 protocol.
 #
 # Default: yes
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -226,8 +226,8 @@ class CommandSelect : public Commander {
     // DB 0 uses default namespace
     std::string ns = kDefaultNamespace;
     if (db_index > 0) {
-      // Use database index as namespace with format "db001", "db002", etc.
-      ns = fmt::format("db{:03d}", db_index);
+      // Use database index as namespace directly
+      ns = std::to_string(db_index);
     }
     conn->SetNamespace(ns);
     *output = redis::RESP_OK;
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -71,6 +71,9 @@ class CommandNamespace : public Commander {
     if (config->repl_namespace_enabled && config->IsSlave() && sub_command != "get") {
       return {Status::RedisExecErr, "namespace is read-only for slave"};
     }
+    if (config->redis_databases > 0) {
+      return {Status::RedisExecErr, "namespace command is not allowed when redis-databases > 0"};
+    }
     if (args_.size() == 3 && sub_command == "get") {
       if (args_[2] == "*") {
         std::vector<std::string> namespaces;
@@ -201,8 +204,32 @@ class CommandPing : public Commander {
 
 class CommandSelect : public Commander {
  public:
-  Status Execute([[maybe_unused]] engine::Context &ctx, [[maybe_unused]] Server *srv, [[maybe_unused]] Connection *conn,
-                 std::string *output) override {
+  Status Execute([[maybe_unused]] engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    Config *config = srv->GetConfig();
+    // If redis-databases is 0, just return OK (default behavior)
+    if (config->redis_databases == 0) {
+      *output = redis::RESP_OK;
+      return Status::OK();
+    }
+
+    // Parse database index
+    auto parse_result = ParseInt<int>(args_[1], 10);
+    if (!parse_result) {
+      return {Status::RedisParseErr, "Invalid DB number"};
+    }
+    int db_index = *parse_result;
+    // Validate database index range
+    if (db_index < 0 || db_index >= config->redis_databases) {
+      return {Status::RedisParseErr, "DB number is out of range"};
+    }
+
+    // DB 0 uses default namespace
+    std::string ns = kDefaultNamespace;
+    if (db_index > 0) {
+      // Use database index as namespace with format "db001", "db002", etc.
+      ns = fmt::format("db{:03d}", db_index);
+    }
+    conn->SetNamespace(ns);
     *output = redis::RESP_OK;
     return Status::OK();
   }
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -226,8 +226,8 @@ class CommandSelect : public Commander {
     // DB 0 uses default namespace
     std::string ns = kDefaultNamespace;
     if (db_index > 0) {
-      // Use database index as namespace directly
-      ns = std::to_string(db_index);
+      // Use database index as namespace with "db" prefix
+      ns = "db" + std::to_string(db_index);
     }
     conn->SetNamespace(ns);
     *output = redis::RESP_OK;
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -226,8 +226,7 @@ class CommandSelect : public Commander {
     // DB 0 uses default namespace
     std::string ns = kDefaultNamespace;
     if (db_index > 0) {
-      // Use database index as namespace with "db" prefix
-      ns = "db" + std::to_string(db_index);
+      ns = std::string(kDatabaseNamespacePrefix) + std::to_string(db_index);
     }
     conn->SetNamespace(ns);
     *output = redis::RESP_OK;
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -233,7 +233,7 @@ Config::Config() {
       {"log-retention-days", true, new IntField(&log_retention_days, -1, -1, INT_MAX)},
       {"persist-cluster-nodes-enabled", false, new YesNoField(&persist_cluster_nodes_enabled, true)},
       {"redis-cursor-compatible", false, new YesNoField(&redis_cursor_compatible, true)},
-      {"redis-databases", true, new IntField(&redis_databases, 0, 0, 1000)},
+      {"redis-databases", true, new IntField(&redis_databases, 0, 0, INT_MAX)},
       {"resp3-enabled", false, new YesNoField(&resp3_enabled, true)},
       {"repl-namespace-enabled", false, new YesNoField(&repl_namespace_enabled, false)},
       {"proto-max-bulk-len", false,
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -233,6 +233,7 @@ Config::Config() {
       {"log-retention-days", true, new IntField(&log_retention_days, -1, -1, INT_MAX)},
       {"persist-cluster-nodes-enabled", false, new YesNoField(&persist_cluster_nodes_enabled, true)},
       {"redis-cursor-compatible", false, new YesNoField(&redis_cursor_compatible, true)},
+      {"redis-databases", true, new IntField(&redis_databases, 0, 0, 1000)},
       {"resp3-enabled", false, new YesNoField(&resp3_enabled, true)},
       {"repl-namespace-enabled", false, new YesNoField(&repl_namespace_enabled, false)},
       {"proto-max-bulk-len", false,
@@ -889,6 +890,12 @@ Status Config::finish() {
   }
   if ((cluster_enabled) && !load_tokens.empty()) {
     return {Status::NotOK, "enabled cluster mode wasn't allowed while the namespace exists"};
+  }
+  if ((redis_databases > 0) && !load_tokens.empty()) {
+    return {Status::NotOK, "redis-databases > 0 is not allowed while any non-default namespace exists"};
+  }
+  if ((redis_databases > 0) && (cluster_enabled)) {
+    return {Status::NotOK, "cluster mode and redis-databases cannot be enabled at the same time"};
   }
   if (unixsocket.empty() && binds.size() == 0) {
     binds.emplace_back(kDefaultBindAddress);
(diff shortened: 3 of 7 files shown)


git-hulk:
@hanxi Thanks for your contribution. Another approach is to use the DB ID to replace the namespace when `redis-select-compatible` is enabled.

To clarify, we would use the DB ID as the namespace and don't need the corresponding token for those DBs (namespace). And we need to take care of the following scenarios:

- Cannot enable `redis-select-compatible` if there are any other namespaces except for the default namespace.
- Cannot create a namespace or include a namespace token in the configuration file.

This way would be more transparent and easier to maintain by avoiding mixing the namespace and the Redis DB behavior.

To see if @PragmaTwice @torwig @mapleFU @caipengbo have any comments on this?

hanxi:
Your suggestion is good. Not using DB ID and namespace at the same time can indeed reduce conflicts, and it will also make the code more concise, so there is no need for the `redis-select-db-prefix` configuration item. Should we keep the `redis-select-db-count` configuration item?

git-hulk:
> Your suggestion is good. Not using DB ID and namespace at the same time can indeed reduce conflicts, and it will also make the code more concise, so there is no need for the `redis-select-db-prefix` configuration item. Should we keep the `redis-select-db-count` configuration item?

Yes, `redis-select-db-prefix` is not needed. And for the `redis-select-db-count`, perhaps `redis-databases` would be better to keep consistent with Redis, which is named `databases`.

src/server/server.cc
```diff
@@ -158,6 +158,18 @@ Status Server::Start() {
   if (!s.IsOK()) {
     return s;
   }
+
+  // Validate redis-select-compatible mode configuration
+  if (config_->redis_select_compatible) {
```
git-hulk: You can check this after loading the configuration file: https://github.com/apache/kvrocks/blob/unstable/src/config/config.cc#L886C16-L886C22

src/commands/cmd_server.cc
```diff
@@ -201,8 +201,29 @@ class CommandPing : public Commander {
...
+      return {Status::RedisParseErr, "Invalid DB index"};
+    }
+    int db_index = *parse_result;
+    // Validate database index range
+    if (db_index < 0 || db_index >= config->redis_databases) {
+      return {Status::RedisParseErr, "Invalid DB index"};
+    }
+
+    // Use database index as namespace
+    std::string ns = std::to_string(db_index);
```
git-hulk: Should we use the default namespace if the DB is 0?

git-hulk:
@hanxi Except for comments inline. There are two issues to be addressed:

1. Using the namespace command should return an error if the Redis DB compatible feature is enabled.
2. Add Go test cases for this feature. For HOW-TO page, can refer to: https://kvrocks.apache.org/community/contributing#tests

hanxi:
> @hanxi Except for comments inline. There are two issues to be addressed:
> 
> 1. Using the namespace command should return an error if the Redis DB compatible feature is enabled.
> 2. Add Go test cases for this feature. For HOW-TO page, can refer to: https://kvrocks.apache.org/community/contributing#tests

Thanks for the feedback! I'll address both issues:

1. Add error handling for namespace command when redis-select-compatible is enabled
2. Add Go test cases for this feature

I'll need some time to familiarize myself with the testing framework and will submit the updates soon.

src/server/server.cc
```diff
@@ -158,6 +158,7 @@ Status Server::Start() {
   if (!s.IsOK()) {
     return s;
   }
+
```
git-hulk:
```suggestion
```

PragmaTwice:
Maybe we can have just one config option `redis-databases`, and use value `0` to disable it, and then `redis-select-compatible` can be removed.

Lots of edge cases need to be considered, like:
- behavior when cluster is enabled
- behavior when `repl-namespace-enabled yes`
- behavior when non-default namespace is defined
- handling of NAMESPACE subcommands
- ...

PragmaTwice:
And for the "namespace name" if this feature is enabled, I think it can be `db<NUMBER>`? Not sure if it benefits if we use fixed length numbers like db001, db002..

hanxi:
That's a good suggestion to use only `redis-databases` .

PragmaTwice:
Ahh, also we need to ensure that, if this feature is enabled, then no namespace can be created. To avoid chaos.

hanxi:
> And for the "namespace name" if this feature is enabled, I think it can be `db<NUMBER>`? Not sure if it benefits if we use fixed length numbers like db001, db002..

The current implementation uses __namespace for db0 and numbers 1-15 for db1-15.


> Ahh, also we need to ensure that, if this feature is enabled, then no namespace can be created. To avoid chaos.

The current implementation is that once select is enabled, namespace cannot be used.

PragmaTwice:
> The current implementation is that once select is enabled, namespace cannot be used.

Hmm what will happen if users execute `NAMESPACE ADD ..` in redis clients? Also, when `repl-namespace-enabled yes`, namespaces will be stored in rocksdb instead of kvrocks.conf. We should consider that.

src/commands/cmd_server.cc
```diff
@@ -71,6 +71,9 @@ class CommandNamespace : public Commander {
     if (config->repl_namespace_enabled && config->IsSlave() && sub_command != "get") {
       return {Status::RedisExecErr, "namespace is read-only for slave"};
     }
+    if (config->redis_select_compatible) {
+      return {Status::RedisExecErr, "namespace command is not allowed when redis-select-compatible mode is enabled"};
```
hanxi: @PragmaTwice `NAMESPACE ADD .. ` will failed.

PragmaTwice: Ahh got it. thank you.

PragmaTwice:
And maybe we can ensure that `CLIENT INFO` can return the correct db number.

src/config/config.cc
```diff
@@ -892,8 +891,8 @@ Status Config::finish() {
   if ((cluster_enabled) && !load_tokens.empty()) {
     return {Status::NotOK, "enabled cluster mode wasn't allowed while the namespace exists"};
   }
-  if ((redis_select_compatible) && !load_tokens.empty()) {
-    return {Status::NotOK, "redis-select-compatible mode wasn't allowed while the namespace exists"};
+  if ((redis_databases > 0) && !load_tokens.empty()) {
```
git-hulk, line 894: @hanxi You can also check if `cluster_enabled` is enabled here. If enabled, redis database feture should be also disasllowed.

tests/gocase/unit/select/select_test.go
```diff
@@ -0,0 +1,282 @@
+/*
```
PragmaTwice, line 1: Could you also test some commands like AUTH, HELLO in golang tests?

hanxi: Sure, I'll add the missing test cases for commands like AUTH and HELLO in the Go tests.

src/server/redis_connection.cc
```diff
@@ -68,9 +68,31 @@ Connection::~Connection() {
...
+  if (srv_->GetConfig()->redis_databases > 0) {
+    // Parse db number from namespace (format: "db0001", "db0002", etc.)
+    int db_num = 0;
+    if (ns_ != kDefaultNamespace && ns_.size() >= 3 && ns_.substr(0, 2) == "db") {
+      try {
+        db_num = std::stoi(ns_.substr(2));
+      } catch (...) {
+        // If parsing fails, default to 0
+        db_num = 0;
+      }
```
git-hulk:
```suggestion
      db_num = ParseInt<int>(ns_.substr(2), 10).ValueOr(0);
```

git-hulk:
Looks good to me, one nit suggestion inline.

src/server/redis_connection.cc
```diff
@@ -68,9 +68,31 @@ Connection::~Connection() {
...
+    }
+
+    result = fmt::format("id={} addr={} fd={} name={} age={} idle={} flags={} db={} qbuf={} obuf={} cmd={}\n", id_,
+                         addr_, bufferevent_getfd(bev_), name_, GetAge(), GetIdleTime(), GetFlags(), db_num,
+                         evbuffer_get_length(Input()), evbuffer_get_length(Output()), last_cmd_);
+  } else {
+    result = fmt::format("id={} addr={} fd={} name={} age={} idle={} flags={} namespace={} qbuf={} obuf={} cmd={}\n",
+                         id_, addr_, bufferevent_getfd(bev_), name_, GetAge(), GetIdleTime(), GetFlags(), ns_,
+                         evbuffer_get_length(Input()), evbuffer_get_length(Output()), last_cmd_);
```
PragmaTwice: These code seems highly repeated. Could you move them out of if..else?

src/commands/cmd_server.cc
```diff
@@ -201,8 +204,32 @@ class CommandPing : public Commander {
...
+
+    // Parse database index
+    auto parse_result = ParseInt<int>(args_[1], 10);
+    if (!parse_result) {
+      return {Status::RedisParseErr, "Invalid DB index"};
+    }
+    int db_index = *parse_result;
+    // Validate database index range
+    if (db_index < 0 || db_index >= config->redis_databases) {
+      return {Status::RedisParseErr, "Invalid DB index"};
```
```suggestion
      return {Status::RedisParseErr, "DB number is out of range"};
```

src/commands/cmd_server.cc
```diff
@@ -201,8 +204,32 @@ class CommandPing : public Commander {
...
+    // If redis-databases is 0, just return OK (default behavior)
+    if (config->redis_databases == 0) {
+      *output = redis::RESP_OK;
+      return Status::OK();
+    }
+
+    // Parse database index
+    auto parse_result = ParseInt<int>(args_[1], 10);
+    if (!parse_result) {
+      return {Status::RedisParseErr, "Invalid DB index"};
```
```suggestion
      return {Status::RedisParseErr, "Invalid DB number"};
```

src/config/config.cc
```diff
@@ -233,6 +233,7 @@ Config::Config() {
       {"log-retention-days", true, new IntField(&log_retention_days, -1, -1, INT_MAX)},
       {"persist-cluster-nodes-enabled", false, new YesNoField(&persist_cluster_nodes_enabled, true)},
       {"redis-cursor-compatible", false, new YesNoField(&redis_cursor_compatible, true)},
+      {"redis-databases", true, new IntField(&redis_databases, 0, 0, 10000)},
```
PragmaTwice: Is 10000 suitable? not sure what's the limit in redis.

src/config/config.cc
```diff
@@ -890,6 +891,12 @@ Status Config::finish() {
   if ((cluster_enabled) && !load_tokens.empty()) {
     return {Status::NotOK, "enabled cluster mode wasn't allowed while the namespace exists"};
   }
+  if ((redis_databases > 0) && !load_tokens.empty()) {
+    return {Status::NotOK, "redis-databases > 0 wasn't allowed while the namespace exists"};
```
```suggestion
    return {Status::NotOK, "redis-databases > 0 is not allowed while any non-default namespace exists"};
```

src/config/config.cc
```diff
@@ -890,6 +891,12 @@ Status Config::finish() {
...
+  }
+  if ((redis_databases > 0) && (cluster_enabled)) {
+    return {Status::NotOK, "enabled cluster mode and redis-databases > 0 wasn't allowed at the same time"};
```
```suggestion
    return {Status::NotOK, "cluster mode and redis-databases cannot be enabled at the same time"};
```

src/config/config.cc
hanxi: Redis has no limit—change it to 1000?

src/server/redis_connection.cc
hanxi: Got it, I'll refactor it.

git-hulk:
@hanxi Could you please fix this format issue: https://github.com/apache/kvrocks/actions/runs/20254211884/job/58152672869#step:6:1

hanxi:
> @hanxi Could you please fix this format issue: https://github.com/apache/kvrocks/actions/runs/20254211884/job/58152672869#step:6:1

Ok

src/config/config.cc
PragmaTwice: > change it to 1000?

Why?

> Redis has no limit

Are you sure? It seems to be `2^31 - 1`.

PragmaTwice: cc @git-hulk

src/server/redis_connection.cc
```diff
@@ -68,9 +68,30 @@ Connection::~Connection() {
 }
 
 std::string Connection::ToString() {
-  return fmt::format("id={} addr={} fd={} name={} age={} idle={} flags={} namespace={} qbuf={} obuf={} cmd={}\n", id_,
-                     addr_, bufferevent_getfd(bev_), name_, GetAge(), GetIdleTime(), GetFlags(), ns_,
-                     evbuffer_get_length(Input()), evbuffer_get_length(Output()), last_cmd_);
+  std::string result;
```
PragmaTwice:
```suggestion
```

src/server/redis_connection.cc
```diff
@@ -68,9 +68,30 @@ Connection::~Connection() {
...
+      db_num = ParseInt<int>(ns_.substr(2), 10).ValueOr(0);
+    }
+    db_or_ns_field = "db";
+    db_or_ns_value = std::to_string(db_num);
+  } else {
+    db_or_ns_field = "namespace";
+    db_or_ns_value = ns_;
+  }
+
+  result = fmt::format("id={} addr={} fd={} name={} age={} idle={} flags={} {}={} qbuf={} obuf={} cmd={}\n", id_, addr_,
```
PragmaTwice:
```suggestion
  std::string result = fmt::format("id={} addr={} fd={} name={} age={} idle={} flags={} {}={} qbuf={} obuf={} cmd={}\n", id_, addr_,
```

src/config/config.cc
hanxi: Some cloud service providers support 256, for example, Tair's limit is 256. Using 1000 would just occupy three digits, such as aligning from db001 to db999.

git-hulk: Yes, it's INT_MAX(2^31 - 1) in Redis open source version. Perhaps we should follow this value.

hanxi: If we use INT_MAX (2^31 - 1) as the maximum value, I think it would be better not to have the 'db' prefix.

git-hulk: It's good to me. @PragmaTwice WDYT?

git-hulk (dismissed):

src/config/config.cc
PragmaTwice: > If we use INT_MAX (2^31 - 1) as the maximum value, I think it would be better not to have the 'db' prefix.

why? i prefer keep the prefix so it is easy to debug and identify.

hanxi: After using INT_MAX, namespace names can no longer use zero-padded formats like db001. Should we change them to db1 instead? Do you have any other suggestions?

PragmaTwice: > Should we change them to db1 instead?

yup. regardless of there's a prefix or not. it's alphabetical order in rocksdb and we cannot iterate from db1 to dbN in integer order. cc @git-hulk

hanxi:
Should we define a constant to represent the database prefix?

```c
constexpr const char *kDatabaseNamespacePrefix = "db";
```

git-hulk:
> Should we define a constant to represent the database prefix?
> 
> ```c
> constexpr const char *kDatabaseNamespacePrefix = "db";
> ```


Yes, it would be better since many places are using this.

git-hulk (dismissed):

src/server/redis_connection.cc
```diff
@@ -68,9 +68,28 @@ Connection::~Connection() {
...
+  // When redis-databases > 0 (SELECT compatibility mode), show db field instead of namespace
+  std::string db_or_ns_field;
+  std::string db_or_ns_value;
+
+  if (srv_->GetConfig()->redis_databases > 0) {
+    // Parse db number from namespace (format: kDatabaseNamespacePrefix + number, e.g., "db1", "db2", etc.)
+    int db_num = 0;
+    const size_t prefix_len = strlen(kDatabaseNamespacePrefix);
+    if (ns_ != kDefaultNamespace && ns_.size() >= prefix_len &&
+        ns_.compare(0, prefix_len, kDatabaseNamespacePrefix) == 0) {
```
PragmaTwice:
```suggestion
    if (util::StartsWith(ns_, kDatabaseNamespacePrefix)) {
```

PragmaTwice (commented):
Maybe we can make sure that in this mode AUTH will reject any password rather than the admin password. (e.g. when namespaces are stored in rocksdb we cannot detect if there's no namespace in the startup.)

kvrocks.conf
```diff
@@ -388,6 +388,24 @@ max-bitmap-to-string-mb 16
...
+# With a value of 16, you can use SELECT 0 through SELECT 15.
+# Each database index is mapped to a namespace name with "db" prefix.
+# DB 0 uses the default namespace, DB 1 uses "db1", DB 2 uses "db2", and so on.
+#
+# IMPORTANT NOTES:
+# 1. This is a READONLY configuration and can only be set at startup
+# 2. Setting this to 0 disables SELECT compatibility mode (default behavior)
+# 3. Setting this to a value > 0 enables SELECT compatibility mode
+# 4. Valid range is 0 to INT_MAX
+# 5. Cannot be used together with namespace feature (namespace tokens)
```
```suggestion
# Number of Redis-compatible databases.
# When set to 0 (default), SELECT command acts like a no-op i.e. Redis-compatible databases are disabled.
# When set to a positive value, SELECT command is Redis-compatible and the namespace feature is disabled.
#
# For example, if you set redis-databases to 16, then SELECT 0 through SELECT 15 are available.
# Note that SELECT 0 always maps to the default namespace, regardless of whether this feature is enabled.
```

LLM usually dumps lots of useless words..

hanxi:
> Maybe we can make sure that in this mode AUTH will reject any password rather than the admin password. (e.g. when namespaces are stored in rocksdb we cannot detect if there's no namespace in the startup.)

Is it like this?
```c
class CommandAuth : public Commander {
 public:
  Status Execute([[maybe_unused]] engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
    auto &user_password = args_[1];
    std::string ns;
    AuthResult result = srv->AuthenticateUser(user_password, &ns);
    switch (result) {
      case AuthResult::NO_REQUIRE_PASS:
        return {Status::RedisExecErr, "Client sent AUTH, but no password is set"};
      case AuthResult::INVALID_PASSWORD:
        return {Status::RedisExecErr, "Invalid password"};
      case AuthResult::IS_USER: {
        Config *config = srv->GetConfig();
        if (config->redis_databases > 0) {
          return {Status::RedisExecErr, "namespace AUTH command is not allowed when redis-databases > 0"};
        }
        conn->BecomeUser();
        break;
      }
      case AuthResult::IS_ADMIN:
        conn->BecomeAdmin();
        break;
    }
    conn->SetNamespace(ns);
    *output = redis::RESP_OK;
    return Status::OK();
  }
};
```

PragmaTwice:
Yup. Also for CommandHello.

git-hulk:
Server::Start will load the tokens before serving; it can also check at https://github.com/apache/kvrocks/blob/65db00a86bc9031f4d974eb0b6f1893bd126c524/src/server/server.cc#L157

PragmaTwice:
> Server::Start will load the tokens before serving; it can also check at
> 
> https://github.com/apache/kvrocks/blob/65db00a86bc9031f4d974eb0b6f1893bd126c524/src/server/server.cc#L157

should we clear namespaces at the start if redis-databases are enabled? not sure.. maybe just check it for AUTH and HELLO?

git-hulk:
> > Server::Start will load the tokens before serving; it can also check at
> > https://github.com/apache/kvrocks/blob/65db00a86bc9031f4d974eb0b6f1893bd126c524/src/server/server.cc#L157
> 
> should we clear namespaces at the start if redis-databases are enabled? not sure.. maybe just check it for AUTH and HELLO?

Oh, I mean that we can disallow users to start the server if there are any namespaces in rocksdb while redis.database is enabled.

PragmaTwice:
Looks good. We can modify this function instead: https://github.com/apache/kvrocks/blob/unstable/src/server/namespace.cc#L71

Then we don't need to add checks for AUTH and HELLO.

cc @hanxi

src/server/namespace.cc
```diff
@@ -77,6 +77,10 @@ Status Namespace::LoadAndRewrite() {
   auto s = loadFromDB(&db_tokens);
   if (!s.IsOK()) return s;
 
+  if (!db_tokens.empty() && config->redis_databases > 0) {
+    return {Status::NotOK, "cannot enable redis_databases when namespaces exist in db"};
```
PragmaTwice:
```suggestion
    return {Status::NotOK, "cannot enable redis-databases when namespaces exist in db"};
```

kvrocks.conf
```diff
@@ -388,6 +388,16 @@ max-bitmap-to-string-mb 16
 # Default: yes
 redis-cursor-compatible yes
 
+# Number of Redis-compatible databases.
+# When set to 0 (default), SELECT command acts like a no-op i.e. Redis-compatible databases are disabled.
+# When set to a positive value, SELECT command is Redis-compatible and the namespace feature is disabled.
+#
+# For example, if you set redis-databases to 16, then SELECT 0 through SELECT 15 are available.
+# Note that SELECT 0 always maps to the default namespace, regardless of whether this feature is enabled.
```
```suggestion
# Note that SELECT 0 maps to the default namespace even when this feature is enabled.
```

PragmaTwice (approved):

git-hulk (approved):

khanbilal732 (approved):

git-hulk:
@hanxi Thanks for your contribution and prompt shifting for each suggestion.

hanxi:
Thanks for the review and the quick merge! Happy to contribute to the project.

PR #3297, opened 2025-12-17
perf(cluster): use multithreads to optimize dbsize scan

hll1213181368:


jihuayu:
Hi @hll1213181368 Are you still working on this PR?

Issue #3309, opened 2025-12-25
Add support of the DIGEST command


PragmaTwice:
Refer to https://redis.io/docs/latest/commands/digest/.

chakkk309:
Hi, may i pick up this issue?

PR #3317, opened 2025-12-30
feat(delex): Implement DELEX Command

### LINKED ISSUE #3310: Add support of the DELEX command
https://redis.io/docs/latest/commands/delex/.

Depends on #3309.

Valay17:
Fixes #3310

Implement DELEX Command

Digest function commented out


src/commands/cmd_string.cc
```diff
@@ -108,6 +108,56 @@ class CommandGetEx : public Commander {
...
+        opt_ = '2';
+        hash_or_val_ = GET_OR_RET(parser.TakeStr());
+      } else if (parser.EatEqICase("ifeq")) {
+        opt_ = '3';
+        hash_or_val_ = GET_OR_RET(parser.TakeStr());
+      } else if (parser.EatEqICase("ifne")) {
+        opt_ = '4';
+        hash_or_val_ = GET_OR_RET(parser.TakeStr());
+      } else {
+        return {Status::RedisParseErr, errInvalidSyntax};
```
PragmaTwice, line 129: Please use an `enum class` instead of meaningless chars like `1`, `2`, `3`, `4`.

src/types/redis_string.cc
```diff
@@ -181,6 +182,48 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, std::optional<char> &opt_,
+                              std::optional<std::string> &hash_or_val_, std::optional<bool> &res_) {
```
```suggestion
rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, const DelExOption &option, bool &deleted) {
```

src/types/redis_string.cc
```diff
@@ -181,6 +182,48 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
...
+  }
+  switch (opt_.value()) {
+    case '1':
+      if (hash_or_val_.value() == "12345") {  // StringDigest(value)
+        res_ = true;
+      }
+      break;
+    case '2':
+      if (hash_or_val_.value() != "12345") {  // StringDigest(value)
+        res_ = true;
```
PragmaTwice: You can just add the StringDigest function, otherwise this PR cannot be merged.

tests/cppunit/types/string_test.cc
```diff
@@ -367,4 +528,4 @@ TEST_F(RedisStringTest, LCS) {
                     },
                     4},
                    std::get<StringLCSIdxResult>(rst));
-}
+}
```
PragmaTwice:
```suggestion
}

```

PragmaTwice (commented):
Hey thank you for your contribution, could you please add some golang test cases?

src/commands/cmd_string.cc
Valay17: Understood. I’ll replace these with an enum class.

src/types/redis_string.cc
Valay17: Understood. I’ll add the StringDigest function before the final PR.

Valay17:
Thank you @PragmaTwice, I am working on them

src/types/redis_string.cc
```diff
@@ -181,6 +182,48 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
...
 }
 
+rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, DelExOption &option,
+                              std::optional<std::string> &hash_or_value, bool &deleted) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+  std::string value;
+  rocksdb::Status s = getValue(ctx, ns_key, &value);
+  if (!s.ok() || s.IsNotFound()) {
+    return s;
+  }
```
git-hulk: `IsNotFound` is included by `!s.ok()`

```suggestion
  if (!s.ok()) return s;
```

git-hulk:
@Valay17 I think we could add a digest util function in this PR, so that other PRs won't block it. cc @PragmaTwice

Valay17:
Sure, will add it as well and also update the places where I have commented it, also a question regarding the placement of the header files xxh3.h and xxhash.h, which directory would be appropriate to place it in?

src/types/redis_string.h
```diff
@@ -34,6 +34,8 @@ struct StringPair {
   Slice value;
 };
 
+enum class DelExOption { NONE, IFDEQ, IFDNE, IFEQ, IFNE };
```
PragmaTwice: Maybe like this? We can also put the value inside it.

```suggestion
struct DelExOption {
  enum Type { NONE, IFDEQ, IFDNE, IFEQ, IFNE };
  Type type;
  std::string value;
};
```

PragmaTwice:
> Sure, will add it as well and also update the places where I have commented it, also a question regarding the placement of the header files xxh3.h and xxhash.h, which directory would be appropriate to place it in?

You don't need to place the header file, these will be automatically fetched and put into your build dir: https://github.com/apache/kvrocks/blob/unstable/cmake/xxhash.cmake.

You can just `#include "xxh3.h"`.

Valay17:
Perfect! Will add it accordingly and also change enum to struct

Valay17:
Hi, could you please tell me what is the reason for my go test cases failure?

src/commands/cmd_string.cc
```diff
@@ -108,6 +108,51 @@ class CommandGetEx : public Commander {
...
+ public:
+  Status Parse(const std::vector<std::string> &args) override {
+    if (args.size() > 4) {
+      return {Status::RedisParseErr, errWrongNumOfArguments};
+    }
+
+    CommandParser parser(args, 2);
+    while (parser.Good()) {
+      if (parser.EatEqICase("ifdeq")) {
+        option = {DelExOption::Type::IFDEQ, GET_OR_RET(parser.TakeStr())};
```
PragmaTwice:
```suggestion
        option = {DelExOption::IFDEQ, GET_OR_RET(parser.TakeStr())};
```

src/types/redis_string.cc
```diff
@@ -181,6 +182,47 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
...
+      }
+      break;
+    case DelExOption::Type::IFNE:
+      if (option.value != value) {
+        deleted = true;
+      }
+      break;
+    default:
+      return rocksdb::Status::InvalidArgument();
+      break;
```
```suggestion
      return rocksdb::Status::InvalidArgument();
```

src/types/redis_string.cc
```diff
@@ -181,6 +182,47 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
...
 
+rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, DelExOption &option, bool &deleted) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+  std::string value;
+  rocksdb::Status s = getValue(ctx, ns_key, &value);
+  if (!s.ok()) return s;
+
+  if (option.type == DelExOption::Type::NONE && option.value == "") {
+    return storage_->Delete(ctx, storage_->DefaultWriteOptions(), metadata_cf_handle_, ns_key);
+  }
```
```suggestion
  if (option.type == DelExOption::Type::NONE) {
    deleted = true;
    return storage_->Delete(ctx, storage_->DefaultWriteOptions(), metadata_cf_handle_, ns_key);
  }
```

src/types/redis_string.cc
```diff
@@ -181,6 +182,47 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, DelExOption &option, bool &deleted) {
```
```suggestion
rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, const DelExOption &option, bool &deleted) {
```

src/commands/cmd_string.cc
```diff
@@ -108,6 +108,51 @@ class CommandGetEx : public Commander {
...
+
+  Status Execute(engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    redis::String string_db(srv->storage, conn->GetNamespace());
+    auto s = string_db.DelEX(ctx, args_[1], option, deleted);
+
+    if (!s.ok() && !s.IsNotFound()) {
+      return {Status::RedisExecErr, s.ToString()};
+    }
+
+    if (s.IsNotFound() || (option.type == DelExOption::Type::NONE && !deleted)) {
```
PragmaTwice:
```suggestion
    if (s.IsNotFound() || !deleted) {
```

src/common/string_util.h
```diff
@@ -80,7 +81,8 @@ std::string StringJoin(const T &con, F &&f, std::string_view sep = ", ") {
 
 template <typename T>
 std::string StringJoin(const T &con, std::string_view sep = ", ") {
-  return StringJoin(con, [](const auto &v) -> decltype(auto) { return v; }, sep);
+  return StringJoin(
+      con, [](const auto &v) -> decltype(auto) { return v; }, sep);
```
PragmaTwice: please revert this.

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -277,6 +277,79 @@ func testString(t *testing.T, configs util.KvrocksServerConfigs) {
...
+		value := "test-strings-value69"
+		require.NoError(t, rdb.Set(ctx, key, value, 0).Err())
+		require.Equal(t, value, rdb.Get(ctx, key).Val())
+		
+		// More than 4 args
+		require.Error(t, rdb.Do(ctx, "DelEX", "test-string-key69", "random", "random", "random").Err())
+		//Incorrect option
+		require.Error(t, rdb.Do(ctx, "DELEX", "test-string-key69", "random", "random").Err())
+		//True cases for all options
+		//digest := rdb.Do(ctx, "DIGEST", value).Val().(string)
```
PragmaTwice: why it's commented?

src/types/redis_string.h
```diff
@@ -34,6 +34,12 @@ struct StringPair {
   Slice value;
 };
 
+struct DelExOption {
+  enum class Type { NONE, IFDEQ, IFDNE, IFEQ, IFNE };
+  Type type;
+  std::string value;
+};
```
PragmaTwice, line 44:
```suggestion
struct DelExOption {
  enum class Type { NONE, IFDEQ, IFDNE, IFEQ, IFNE };
  Type type;
  std::string value;
  
  DelExOption() : type(None) {}
  DelExOption(Type type, std::string value) : type(type), value(std::move(value)) {}
};
```

src/commands/cmd_string.cc
```diff
@@ -108,6 +108,51 @@ class CommandGetEx : public Commander {
...
+      *output = redis::Integer(0);
+    } else {
+      *output = redis::Integer(1);
+    }
+    return Status::OK();
+  }
+
+ private:
+  DelExOption option = {DelExOption::Type::NONE, ""};
```
PragmaTwice:
```suggestion
  DelExOption option;
```

PragmaTwice:
> Hi, could you please tell me what is the reason for my go test cases failure?

Please check the CI log: https://github.com/apache/kvrocks/actions/runs/20613605397/job/59205809804

Please note that you should build and test your changes on your local before pushing them. You don't need to wait for the upstream CI.

Refer to https://kvrocks.apache.org/community/contributing.

src/commands/cmd_string.cc
Valay17: I am using enum class instead of enum, please let me know if I should switch from enum class to enum

tests/gocase/unit/type/strings/strings_test.go
Valay17: It depends on the DIGEST command which is yet to be implemented in the issue #3309 once that is done, just need to uncomment

src/commands/cmd_string.cc
PragmaTwice: You can just use enum.

Valay17:
Hi, I have done all the changes and confirmed all testcases are running on my end, please check and let me know if there's any other changes I should be making, thanks!

src/types/redis_string.cc
```diff
@@ -182,6 +183,48 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
...
 }
 
+rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, const DelExOption &option,
+                              bool &deleted) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+  std::string value;
+  rocksdb::Status s = getValue(ctx, ns_key, &value);
+  if (!s.ok()) return s;
+
+  if (option.type == DelExOption::NONE && option.value == "") {
```
```suggestion
  if (option.type == DelExOption::NONE) {
```

src/types/redis_string.cc
```diff
@@ -182,6 +183,48 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, const DelExOption &option,
+                              bool &deleted) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
+  std::string value;
+  rocksdb::Status s = getValue(ctx, ns_key, &value);
+  if (!s.ok()) return s;
```
PragmaTwice: If DelExOption is NONE, we don't need to get this value.

src/types/redis_string.cc
```diff
@@ -182,6 +183,48 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, const DelExOption &option,
+                              bool &deleted) {
+  std::string ns_key = AppendNamespacePrefix(user_key);
```
PragmaTwice, line 190:
```suggestion
                              bool &deleted) {
  deleted = false;
  std::string ns_key = AppendNamespacePrefix(user_key);
```

src/common/string_util.h
PragmaTwice: check it again.

src/types/redis_string.cc
```diff
@@ -182,6 +183,49 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
...
+  deleted = false;
+  std::string ns_key = AppendNamespacePrefix(user_key);
+  std::string value;
+  rocksdb::Status s = getValue(ctx, ns_key, &value);
+  if (!s.ok()) return s;
+
+  if (option.type == DelExOption::NONE) {
+    deleted = true;
+    return storage_->Delete(ctx, storage_->DefaultWriteOptions(), metadata_cf_handle_, ns_key);
+  }
```
```suggestion
  if (option.type == DelExOption::NONE) {
    deleted = true;
    return storage_->Delete(ctx, storage_->DefaultWriteOptions(), metadata_cf_handle_, ns_key);
  }

  std::string value;
  rocksdb::Status s = getValue(ctx, ns_key, &value);
  if (!s.ok()) return s;
```

PragmaTwice:
All CI is still failing.

Please ensure c++ and golang tests pass before you request for review.

Valay17:
Confirmed 3 times that both cpp and go tests are running locally, please review, thanks

Valay17:
For clang-tidy, it is telling me this, could guide me on how to go ahead with it:
 if (s.IsNotFound() || !deleted) {
      |                            ^~~~~~~~
      |                            deleted_
  DelExOption option;
      |               ^~~~~~
      |               option_
 should I rename all variables with underscores at end? or keep them as is?
 
 also the GO test is failing for TestClusterReset, for this one:
 [Build and test (Darwin Clang arm64 with OpenSSL, macos-14, auto, -DENABLE_OPENSSL=ON)](https://github.com/apache/kvrocks/actions/runs/20650464268/job/59294920722?pr=3317#logs)
 
 all others are passing

src/commands/cmd_string.cc
```diff
@@ -108,6 +108,51 @@ class CommandGetEx : public Commander {
...
+      *output = redis::Integer(0);
+    } else {
+      *output = redis::Integer(1);
+    }
+    return Status::OK();
+  }
+
+ private:
+  DelExOption option;
+  bool deleted = false;
```
git-hulk: 1. `option` should be renamed to `option_`.
2. deleted can be a function variable instead of the class's variable.

src/types/redis_string.cc
```diff
@@ -182,6 +183,52 @@ rocksdb::Status String::GetEx(engine::Context &ctx, const std::string &user_key,
   return rocksdb::Status::OK();
 }
 
+rocksdb::Status String::DelEX(engine::Context &ctx, const std::string &user_key, const DelExOption &option,
+                              bool &deleted) {
+  deleted = false;
+  std::string ns_key = AppendNamespacePrefix(user_key);
```
git-hulk: This can prevent returning `deleted=true` along with an error while it fails to delete the key from db.
```suggestion
  deleted = false;

  std::string val;
  std::string ns_key = AppendNamespacePrefix(user_key);
  rocksdb::Status s = getValue(ctx, ns_key, &val);
  if (!s.ok()) return s;

  bool matched = false;
  switch (option.type) {
    case DelExOption::NONE:
      matched = true;
      break;
    case DelExOption::IFDEQ:
      matched = option.value == util::StringDigest(val);
    case DelExOption::IFDNE:
      matched = option.value != util::StringDigest(val);
      break;
    case DelExOption::IFEQ:
      matched = option.value == val;
      break;
    case DelExOption::IFNE:
      matched = option.value != val;
      break;
    default:
      return rocksdb::Status::InvalidArgument();
  }
  if (matched) {
    s = storage_->Delete(ctx, storage_->DefaultWriteOptions(), metadata_cf_handle_, ns_key);
    deleted = s.ok();
    return s;
  }
```

src/commands/cmd_string.cc
Valay17: Thank you for the clarification, I have done the changes accordingly

src/types/redis_string.cc
Valay17: Agreed. At the moment, the Execute function in cmd_string is already written with this behavior in mind. Please let me know if you’d prefer that I keep it as it is, or if you’d like me to update it based on your suggestion.

My main reason for keeping it this way is that the Exists command checks for the key a bit more efficiently. Since the “no arguments” case is really just about verifying whether the key exists (and then deleting it), this approach felt slightly faster than calling getValue.

That said, I wasn’t completely sure whether the Delete function already handles this internally, as I couldn’t fully confirm it by tracing the parent function calls. I’m happy to adjust this if there’s a clearer or more preferred approach. Thank You!

git-hulk: @Valay17 Thanks for getting back to me so quickly. 

> My main reason for keeping it this way is that the Exists command checks for the key a bit more efficiently. 

`KeyExist` in Kvrocks also uses the `rocksdb::Get` to fetch the key value. So maybe we can make this code simple if possible.

> That said, I wasn’t completely sure whether the Delete function already handles this internally, as I couldn’t fully confirm it by tracing the parent function calls. 

The Delete function only append a delete entry on its WAL, so it won't check if a key has been existed or not.

git-hulk (approved):

PragmaTwice (approved):

PR #3318, opened 2025-12-31
feat(string): Add to Support MSETEX Command

### LINKED ISSUE #3307: Add support of the MSETEX command
### Search before asking

### Motivation

Refer to https://redis.io/docs/latest/commands/msetex/.

### Solution

_No response_

ggyuchive:
## Issue
closes #3307 

## Description
In redis, it supports MSETEX which supports MSET with TTL options from v8.4.0.
Syntax of `MSETEX` is below.
```
MSETEX numkeys key value [key value ...] [NX | XX] [EX seconds |
  PX milliseconds | EXAT unix-time-seconds |
  PXAT unix-time-milliseconds | KEEPTTL]
```
This change adds support for the `MSETEX` command in kvrocks, making it compatible with Redis.

- Reference: https://redis.io/docs/latest/commands/msetex/

### DIFF
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -427,8 +427,7 @@ rocksdb::Status String::MSet(engine::Context &ctx, const std::vector<StringPair>
       if (s.ok() && old_expire != 0) {
         metadata.expire = old_expire;
       }
-    }
-    else {
+    } else {
       metadata.expire = args.expire;
     }
     metadata.Encode(&bytes);
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -198,7 +198,7 @@ rocksdb::Status String::GetDel(engine::Context &ctx, const std::string &user_key
 
 rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, const std::string &value) {
   std::vector<StringPair> pairs{StringPair{user_key, value}};
-  return MSet(ctx, pairs, /*expire=*/0);
+  return MSet(ctx, pairs);
 }
 
 rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, const std::string &value,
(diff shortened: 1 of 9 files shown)


src/types/redis_string.cc
```diff
@@ -393,46 +393,71 @@ rocksdb::Status String::IncrByFloat(engine::Context &ctx, const std::string &use
   return updateRawValue(ctx, ns_key, raw_value);
 }
 
-rocksdb::Status String::MSet(engine::Context &ctx, const std::vector<StringPair> &pairs, uint64_t expire_ms) {
+rocksdb::Status String::MSet(engine::Context &ctx, const std::vector<StringPair> &pairs, StringMSetArgs args,
+                             bool *flag) {
+  if (flag != nullptr) {
```
PragmaTwice, line 399: We can assume here that flag cannot be nullptr, and remove this check.

PragmaTwice: Just a nit though.

src/types/redis_string.cc
```diff
@@ -393,46 +393,71 @@ rocksdb::Status String::IncrByFloat(engine::Context &ctx, const std::string &use
...
+  rocksdb::Status s = batch->PutLogData(log_data.Encode());
   if (!s.ok()) return s;
+
   for (const auto &pair : pairs) {
     std::string bytes;
     Metadata metadata(kRedisString, false);
-    metadata.expire = expire_ms;
+    if (args.keep_ttl) {
+      uint64_t old_expire = 0;
+      rocksdb::Status s = GetExpireTime(ctx, pair.key, &old_expire);
```
PragmaTwice, line 428: Optimization: there are duplicated IO accesses in `Exists` and `GetExpireTime`, maybe we can unfiy them and only read a key once.

PragmaTwice (dismissed):

src/types/redis_string.cc
ggyuchive: Actually this `MSet` can pass nullptr in
```
rocksdb::Status String::MSet(engine::Context &ctx, const std::vector<StringPair> &pairs) {
  return MSet(ctx, pairs, {/*expire=*/0, StringSetType::NONE, /*keep_ttl=*/false}, nullptr);
}
```
It does not need flag, so I added to check it.

PragmaTwice (approved):

src/types/redis_string.cc
PragmaTwice: Good to me.

src/types/redis_string.cc
ggyuchive: I understood, it's good optimization point. I will fix it with new PR.

PR #3322, opened 2026-01-02
perf(string): reduce read access in MSETEX

ggyuchive:
When run `MSETEX` command with `NX|XX` and `KEEPTTL`, it access 2*N times for read.
- Reduce read access from 2*N to N.
- Change `Exists` ret type from `int` to `uint32_t`.

### DIFF
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -434,38 +434,43 @@ rocksdb::Status String::MSet(engine::Context &ctx, const std::vector<StringPair>
                              bool *flag = nullptr) {
   if (flag) *flag = false;
 
-  auto batch = storage_->GetWriteBatchBase();
-  WriteBatchLogData log_data(kRedisString);
-  auto s = batch->PutLogData(log_data.Encode());
-  if (!s.ok()) return s;
-
-  for (const auto &pair : pairs) {
-    Metadata old_metadata(kRedisNone, false);
-    if (args.type != StringSetType::NONE || args.keep_ttl) {
-      std::string ns_key = AppendNamespacePrefix(pair.key);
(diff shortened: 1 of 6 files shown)


src/types/redis_string.cc
```diff
@@ -395,43 +395,38 @@ rocksdb::Status String::IncrByFloat(engine::Context &ctx, const std::string &use
...
-      return rocksdb::Status::OK();
-    }
-  }
+                             bool *flag = nullptr) {
+  if (flag) *flag = false;
 
   auto batch = storage_->GetWriteBatchBase();
   WriteBatchLogData log_data(kRedisString);
-  rocksdb::Status s = batch->PutLogData(log_data.Encode());
+  auto s = batch->PutLogData(log_data.Encode());
```
PragmaTwice: I think we should get and modify the batch after all checks (e.g. NX/XX checks below). Even we don't write this batch into storage, in transaction mode we'll use this batch to commit. cc @git-hulk

ggyuchive: @PragmaTwice , you're right. This change has an issue in tx mode.
```
127.0.0.1:6666> msetex 1 key1 value1
(integer) 1
127.0.0.1:6666> get key2
(nil)
127.0.0.1:6666> multi
OK
127.0.0.1:6666(TX)> msetex 2 key1 value100 key2 value200 xx
QUEUED
127.0.0.1:6666(TX)> exec
1) (integer) 0
127.0.0.1:6666> get key1
"value100"
127.0.0.1:6666>
```
key1 must be value1 after commit. I will fix it.

PragmaTwice (approved):

git-hulk (approved):

PR #3327, opened 2026-01-06
perf(hash): use MultiGet to reduce RocksDB calls in HMSET

sryanyuan:
HMSET previously compared existing values by issuing a separate RocksDB Get
for each field. When the number of fields is large, this results in multiple
Get calls and extra overhead.

This change replaces multiple Get calls with a single MultiGet call, batching
the reads and improving performance for HMSET with many fields.

### DIFF
--- a/src/types/redis_hash.cc
+++ b/src/types/redis_hash.cc
@@ -258,9 +258,8 @@ rocksdb::Status Hash::MSet(engine::Context &ctx, const Slice &user_key, const st
   if (!s.ok()) return s;
   std::unordered_set<std::string_view> field_set;
 
-  rocksdb::ReadOptions read_options = ctx.DefaultMultiGetOptions();
   std::vector<rocksdb::Slice> keys;
-  std::vector<std::string> sub_keys;
+  std::vector<std::string> keys_encoded;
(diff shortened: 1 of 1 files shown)


PragmaTwice:
Would you like to provide benchmark data about this change?

sryanyuan:
> Would you like to provide benchmark data about this change?

In our production environment, benchmarks on real traffic demonstrate that batching reads with MultiGet in HMSET improves performance by approximately 30%.

Without MultiGet: setting 27,000 fields took ~101 ms
With MultiGet: setting 29,500 fields took ~76 ms
This optimisation reduces the overhead of multiple RocksDB Get calls and increases throughput for large hash updates.

PragmaTwice (approved):

PragmaTwice:
I should note that this structure is wrong in transaction mode:
```c++
for (...) {
  if (..) return ..;

  batch->Put(...)
}

batch->Commit()
```

because if early returns happen, in transaction mode we will still commit this batch in EXEC command.

But this is not introduced in this PR, so I'm fine to merge. It can be fixed later.

git-hulk (approved):

PR #3340, opened 2026-01-13
feat(replication): add sequence padding to avoid full sync on misaligned WriteBatch sequence

sryanyuan:
In Kvrocks replication, a replica's requested incremental sequence must align with the start sequence of a WriteBatch in the master's WAL. For example, if a WriteBatch starts at sequence 10 and contains 3 records, valid starting points are 10 or 13. Requests starting at 11 or 12 are considered invalid, causing incremental sync to fail and triggering a costly full sync.

A common cause of misaligned sequences occurs during master-slave failover: if the old master still has WAL entries that were not yet replicated to the old slave when the role switch happens, and the new master continues to accept writes, then re-establishing replication from this new master may, in certain cases, result in the replica requesting a sequence that falls inside an existing WriteBatch. This mismatch forces a full sync, even if only a few records are missing.

This change introduces an optional configuration `replication-enable-sequence-padding` (default: no). When enabled, the master will send dummy WriteBatch entries to pad the replication stream, advancing the replica's sequence to the next valid position. This allows incremental sync to continue while skipping the missing records, avoiding full sync when only a small number of logs are lost.

Trade-off: skipped records are not applied on the replica, potentially causing minor data inconsistency. This configuration is mainly intended for cache-like use cases, where the application can tolerate partial data loss or temporary inconsistency in favor of maintaining replication continuity and avoiding expensive full syncs.

Additionally, a new INFO metric `sync_partial_padding` is added to track the number of partial sync operations that succeeded due to sequence padding. This complements existing metrics:
- `sync_partial_ok`: successful partial syncs without padding
- `sync_partial_err`: failed partial syncs
- `sync_partial_padding`: successful partial syncs that relied on padding

This metric helps operators monitor how often padding is used to avoid costly full syncs, and assess potential data inconsistency risk.

Changes:
- Added `replication-enable-sequence-padding` config and documentation.
- Implemented padding logic in replication send path.
- Added INFO metric `sync_partial_padding` to record padding-based partial syncs.
- Added unit tests `TestReplicationSequencePadding` to verify behavior.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -231,6 +231,24 @@ replication-delay-bytes 16384
 # Default: 16 updates
 replication-delay-updates 16
 
+# Sequence padding allows the master to insert dummy WriteBatch records into
+# the replication stream when the replica's requested sequence number does not
+# align with the start sequence of any WriteBatch in the master's WAL.
+#
+# For example, if a WriteBatch starts at sequence 10 and contains 3 records,
+# valid starting sequences for partial sync are 10 or 13. A request starting
+# at 11 or 12 would normally be considered invalid and trigger a full sync.
+# When enabled, the master will send padding batches to advance the replica's
+# sequence to the next valid position, allowing incremental replication to
+# continue while skipping the missing records.
+#
+# This can avoid expensive full synchronization when only a small number of
+# log entries are missing. The trade-off is that the skipped records will not
+# be applied on the replica, potentially causing minor data inconsistency.
+#
+# Default: no
+replication-enable-sequence-padding no
+
 # TCP listen() backlog.
 #
 # In high requests-per-second environments you need an high backlog in order
--- a/src/cluster/replication.cc
+++ b/src/cluster/replication.cc
@@ -57,10 +57,12 @@
 #include <openssl/ssl.h>
 #endif
 
-FeedSlaveThread::FeedSlaveThread(Server *srv, redis::Connection *conn, rocksdb::SequenceNumber next_repl_seq)
+FeedSlaveThread::FeedSlaveThread(Server *srv, redis::Connection *conn, rocksdb::SequenceNumber next_repl_seq,
+                                 uint32_t padded_seq_count /*= 0*/)
     : srv_(srv),
       conn_(conn),
       next_repl_seq_(next_repl_seq),
+      padded_seq_count_(padded_seq_count),
       req_(srv),
       max_delay_bytes_(srv->GetConfig()->max_replication_delay_bytes),
       max_delay_updates_(srv->GetConfig()->max_replication_delay_updates) {}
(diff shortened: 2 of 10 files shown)


git-hulk:
@sryanyuan Thanks for your PR. I know it's a good point for avoiding unnecessary full syncs in some scenarios, but my concern is that it might cause long-term inconsistency between the master and the replica, even in the metadata and its fields.

For instance, the master has a not-yet-replicated command `HSET hash f0 v0`, and we use the padding sequence to replace it. Aside from the inconsistency between the master and replica, it also has inconsistent HASH metadata and fields if we continue appending new fields to this hash key.

cc @PragmaTwice @torwig @caipengbo

sryanyuan:
> @sryanyuan Thanks for your PR. I know it's a good point for avoiding unnecessary full syncs in some scenarios, but my concern is that it might cause long-term inconsistency between the master and the replica, even in the metadata and its fields.
> 
> For instance, the master has a not-yet-replicated command `HSET hash f0 v0`, and we use the padding sequence to replace it. Aside from the inconsistency between the master and replica, it also has inconsistent HASH metadata and fields if we continue appending new fields to this hash key.
> 
> cc @PragmaTwice @torwig @caipengbo

Thanks for pointing this out — you’re right, skipping WriteBatch entries via sequence padding can lead to long-term inconsistencies not only in the actual data but also in metadata like hash length 🥲

sryanyuan:
> @sryanyuan Thanks for your PR. I know it's a good point for avoiding unnecessary full syncs in some scenarios, but my concern is that it might cause long-term inconsistency between the master and the replica, even in the metadata and its fields.
> 
> For instance, the master has a not-yet-replicated command `HSET hash f0 v0`, and we use the padding sequence to replace it. Aside from the inconsistency between the master and replica, it also has inconsistent HASH metadata and fields if we continue appending new fields to this hash key.
> 
> cc @PragmaTwice @torwig @caipengbo

This PR indeed carries the risk we discussed earlier — skipping certain WriteBatches via replication-enable-sequence-padding may cause long-term inconsistency between metadata and actual data. However, it’s important to note that similar risks can also occur even without enabling padding, particularly in master–slave failover scenarios where the new replica continues replication from its existing sequence.

For example:

Suppose the new replica has executed HSET key field value, resulting in a hash length of 1, with sequence range [1,2] not replicated to the new master.
After failover, the new master accepts write requests, inevitably overwriting these sequences, which leads to inconsistency.
When the new replica starts replication from sequence 3 and the master subsequently performs another HSET on the same hash, the WriteBatch metadata in the master reflects a hash length of 1 (based on its own state).
Once replicated, this metadata overwrites the replica’s metadata, even though the replica actually has two fields in the hash — causing mismatch between metadata and actual data.
This demonstrates that enabling padding increases the risk because it explicitly skips missing sequences, but even without padding, log continuation after failover can introduce inconsistencies.

Mitigation and Operational Considerations

By default, replication-enable-sequence-padding should remain disabled.
It may be enabled in scenarios where rapid failover recovery is critical and partial data loss is acceptable — serving as a fallback to prevent full sync and reduce downtime.
Operators should plan for post-failover repair, such as:
Adding a new replica for a full sync to restore consistency
Replaying operation logs (oplog)
Implementing periodic checks to detect mismatches between metadata and actual data (e.g., hlen vs actual field count)
We could also consider adding a built-in detection mechanism in the future to automatically flag keys where metadata and actual data do not match, so that corrective actions can be triggered.


----------------
Another configuration option, use-rsid-psync, can be used to address the above inconsistency risks more directly

PR #3342, opened 2026-01-16
fix(sideloading): avoid the expected data deletion during SST side loading

ltagliamonte-dd:
Hello folks,
we think we may experiencing data deletion during expensive (thousand of sst files) sst side loading operations.
what we think it is happening is:
- T1 - load data CF
- T2 - auto compactions runs which deletes the data as keys on the metadata CF are not yet loaded
- T3 - load metadata CF

for this reason we would love to switch the sst ingestion logic to use IngestExternalFiles that should take care of sideloading all files from different CF atomically.

### DIFF
--- a/src/storage/storage.cc
+++ b/src/storage/storage.cc
@@ -802,6 +802,7 @@ StatusOr<int> Storage::IngestSST(const std::string &sst_dir, const rocksdb::Inge
     return 0;
   }
 
+  // Group files by column family
   std::unordered_map<ColumnFamilyID, std::vector<std::string>> cf_files;
   for (const auto &file : sst_files) {
     bool matched = false;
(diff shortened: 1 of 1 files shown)


ltagliamonte-dd:
@aleksraiden would you mind run the CI for this so i can test a nightly build?

ltagliamonte:
@git-hulk thanks for the CI run, the failures look like flaky isn't it?
Where can I find the nightly build for latest Sha?

git-hulk:
@ltagliamonte-dd Yes, this test failure is not related to this PR.

> Where can I find the nightly build for latest SHA?

The Docker nightly build can only get after merging the PR. But you can also build it manually and push it to your own Docker repository if you're in a hurry to have a test.

ltagliamonte:
Thank you @git-hulk , do you mind reviewing and maybe ping someone else so we can get this merged?
It should be pretty straightforward we aren't changing the behavior and the test pass.
Thanks!

src/storage/storage.cc
```diff
@@ -802,12 +802,27 @@ StatusOr<int> Storage::IngestSST(const std::string &sst_dir, const rocksdb::Inge
     return 0;
   }
 
-  std::unordered_map<ColumnFamilyID, std::vector<std::string>> cf_files;
+  // Group files by column family and build ingestion arguments atomically
```
git-hulk: @ltagliamonte-dd I have one nit suggestion. It would be more concise and easier to read by refactoring the for-loop to the following:

1. Group SST files by the column family id;
2. Build the ingest args with grouped SST files.

```C++
  const auto &column_families = ColumnFamilyConfigs::ListAllColumnFamilies();
  std::unordered_map<ColumnFamilyID, std::vector<std::string>> column_family_files;
  for (const auto &file : sst_files) {
    auto iter = std::find_if(column_families.begin(), column_families.end(),
                             [&file](const auto &cf) { return file.find(cf.Name()) != std::string::npos; });
    if (iter == column_families.end()) {
      return {Status::NotOK, fmt::format("SST file '{}' does not match any known column family name", file)};
    }
    column_family_files[iter->Id()].push_back(file);
  }

  std::vector<rocksdb::IngestExternalFileArg> ingest_args;
  ingest_args.reserve(column_family_files.size());
  for (auto &[cf_id, files] : column_family_files) {
    rocksdb::IngestExternalFileArg arg;
    arg.column_family = GetCFHandle(cf_id);
    arg.external_files = std::move(files);
    arg.options = ingest_options;
    ingest_args.push_back(std::move(arg));
  }
```

ltagliamonte-dd: thanks fixed via a0949b35bf3affdd303cd331f969931b58c97156

ltagliamonte-dd: Cc @git-hulk

git-hulk (approved):

PR #3343, opened 2026-01-18
chore(conf): change the default cache to hyper clock cache

git-hulk:
RocksDB has made the default cache implementation to HyperClockCache since the version [v10.7.5](https://github.com/facebook/rocksdb/releases/tag/v10.7.5), so we could change it as our default cache accordingly.

According to [Hyping the hyper clock cache in RocksDB](https://smalldatum.blogspot.com/2022/10/hyping-hyper-clock-cache-in-rocksdb.html), using the hyper clock cache can provide better performance than LRU. But I didn't reproduce the significant performance difference between them via `redis-benchmark` on my local.

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -755,8 +755,8 @@ rocksdb.block_cache_size 4096
 # that offers much improved CPU efficiency vs. LRU cache under high parallel
 # load or high contention.
 #
-# default lru
-rocksdb.block_cache_type lru
+# Default: hcc
+rocksdb.block_cache_type hcc
 
 # Number of open files that can be used by the DB.  You may need to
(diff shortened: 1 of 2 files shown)


PragmaTwice (approved):

PragmaTwice:
> But I didn't reproduce the significant performance difference between them via redis-benchmark on my local.

Could we test by setting a smaller cache size?

git-hulk:
> Could we test by setting a smaller cache size?

Yes, I have set it to 16MiB only and am testing with a value of 128 bytes, but the result is slightly better than LRU. I guess the reason is that my VM's CPU is too small (2 cores).

PR #3357, opened 2026-01-30
fix(replication): prevent WAL exhaustion from slow consumers

### LINKED ISSUE #3356: Replication feed thread blocks indefinitely on slow consumer, causing WAL exhaustion and forced fullsync
### Search before asking

### Version

- Master: Kvrocks 2.10.1
- Replica: Kvrocks 2.14.0
- OS: Linux 6.12.53-69.119.amzn2023.aarch64 (Amazon Linux 2023)

### Minimal reproduce step

1. Set up a master-replica configuration with:
     - High write rate (~7K ops/sec)
     - Small WAL retention: `rocksdb.max_total_wal_size 1024` (1GB)

  2. Introduce network congestion or slowness on the replica side that causes it to consume replication data slower than the
  master produces it

  3. The TCP send buffer on the master fills up, and the replication feed thread blocks on write() indefinitely

  4. Wait for WAL rotation to prune old WAL files while the feed thread is still blocked

  5. When the connection eventually drops, or the thread unblocks, observe:
     - Master logs: "Fatal error encountered, WAL iterator is discrete, some seq might be lost"
     - Replica attempts psync, fails with "sequence out of range"
     - Full resync is triggered

  The issue is that step 3 can last indefinitely (we observed 44 hours) with no timeout, errors, or warnings logged.

 1. The master should detect when a replica falls too far behind and proactively disconnect it before WAL is exhausted

  2. Socket sends to replicas should have a timeout to prevent indefinite blocking

  3. Warning logs when replication lag grows significantly

  4. When disconnected early (while the sequence is still in WAL), the replica should be able to psync successfully on reconnect
  instead of requiring a full resync

### What did you see instead?

The replication feed thread blocked for 44 hours with no logs or errors:

  I20260127 22:16:21.006304  2857 replication.cc:115] WAL was rotated, would reopen again
                            [... 44 hours of silence ...]
  I20260129 18:36:55.603111  2857 replication.cc:115] WAL was rotated, would reopen again
  E20260129 18:36:55.646749  2857 replication.cc:126] Fatal error encountered, WAL iterator is discrete, some seq might be
  lost, sequence 480156205527 expected, but got 481055967952
  W20260129 18:36:55.646785  2857 replication.cc:84] Slave thread was terminated

  The replica then failed to psync ("sequence out of range") and required a full resync.

  Root cause: In FeedSlaveThread::loop(), the call to util::SockSend() (line 225) blocks indefinitely when the TCP buffer is
  full. The underlying WriteImpl() has no timeout mechanism. During this blocked period, the master continues writing, and WAL
  files are pruned, leaving the replica's sequence no longer available.

### Anything Else?

Here is the graph for the situation we suffered

Image

I've drafted a possible solution, using Claude code, given that I'm not an expert or dev for C++, that could address this issue with three components:

  1. **Socket send timeout**: New `SockSendWithTimeout()` function using poll() with configurable timeout (default 30s)

  2. **Replication lag detection**: Check lag at start of each loop iteration, disconnect if it exceeds the configurable threshold
  (default 100M sequences)

  3. **Exponential backoff on reconnection**: Prevents rapid reconnect loops for persistently slow replicas (1s, 2s, 4s... up
  to 60s)

  New configuration options:
  - `max-replication-lag`: Max sequence lag before disconnecting slow consumer
  - `replication-send-timeout-ms`: Socket send timeout in milliseconds

  I'm happy to submit the idea PR with the potential fix idea. The changes touch:
  - src/config/config.h, config.cc (new config options)
  - src/common/io_util.h, io_util.cc (SockSendWithTimeout)
  - src/cluster/replication.h, replication.cc (lag detection, timeout usage, backoff)

  Workaround for affected users: Increase `rocksdb.max_total_wal_size` significantly (e.g., 16GB) to extend WAL retention and
  reduce the likelihood of exhaustion

ethervoid:
The replication feed thread could block indefinitely when sending data to a slow replica. If the replica wasn't consuming data fast enough, the TCP send buffer would fill and the feed thread would block on write() with no timeout. During this time, WAL files would rotate and be pruned, leaving the replica's sequence unavailable when the thread eventually unblocked or the connection dropped.

This commit adds three mechanisms to address the issue:

1. Socket send timeout: New SockSendWithTimeout() function that uses poll() to wait for socket writability with a configurable timeout (default 30 seconds). This prevents indefinite blocking.

2. Replication lag detection: At the start of each loop iteration, check if the replica has fallen too far behind (configurable via max-replication-lag, default 100M sequences). If exceeded, disconnect the slow consumer before WAL is exhausted, allowing psync on reconnect.

3. Exponential backoff on reconnection: When a replica is disconnected, it now waits with exponential backoff (1s, 2s, 4s... up to 60s) before reconnecting. This prevents rapid reconnection loops for persistently slow replicas. The backoff resets on successful psync or fullsync.

New configuration options:
- max-replication-lag: Maximum sequence lag before disconnecting (default: 100M)
- replication-send-timeout-ms: Socket send timeout in ms (default: 30000)

Fixes https://github.com/apache/kvrocks/issues/3356

### DIFF
--- a/src/common/io_util.cc
+++ b/src/common/io_util.cc
@@ -29,7 +29,10 @@
 #include <poll.h>
 #include <sys/types.h>
 
+#include <chrono>
+
 #include "fmt/ostream.h"
+#include "scope_exit.h"
 #include "server/tls_util.h"
 
 #ifdef __linux__
@@ -466,6 +469,100 @@ Status SockSend(int fd, const std::string &data, [[maybe_unused]] bufferevent *b
 #else
   return SockSend(fd, data);
 #endif
+}
+
+Status SockSendWithTimeout(int fd, const std::string &data, int timeout_ms) {
+  // Fall back to blocking send if timeout is non-positive
+  if (timeout_ms <= 0) {
+    return SockSend(fd, data);
+  }
+
+  ssize_t n = 0;
+  auto start = std::chrono::steady_clock::now();
+
+  while (n < static_cast<ssize_t>(data.size())) {
+    // Check if we've exceeded the timeout
+    auto elapsed =
+        std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start).count();
+    if (elapsed >= timeout_ms) {
+      return {Status::NotOK, fmt::format("send timeout after {} ms, sent {} of {} bytes", elapsed, n, data.size())};
+    }
+
+    // Calculate remaining timeout
+    int remaining_ms = timeout_ms - static_cast<int>(elapsed);
+
+    // Wait for socket to be writable with timeout
+    int ready = AeWait(fd, AE_WRITABLE, remaining_ms);
+    if (ready == 0) {
+      return {Status::NotOK, fmt::format("send timeout waiting for socket, sent {} of {} bytes", n, data.size())};
+    }
+    if (ready < 0) {
+      return Status::FromErrno("poll error while sending");
+    }
+
+    ssize_t nwritten = write(fd, data.data() + n, data.size() - n);
+    if (nwritten == -1) {
+      if (errno == EAGAIN || errno == EWOULDBLOCK) {
+        // Socket buffer is full, continue waiting
+        continue;
+      }
+      return Status::FromErrno();
+    }
+    n += nwritten;
+  }
+  return Status::OK();
+}
+
+Status SockSendWithTimeout(int fd, const std::string &data, [[maybe_unused]] bufferevent *bev, int timeout_ms) {
+  // Fall back to blocking send if timeout is non-positive
+  if (timeout_ms <= 0) {
+    return SockSend(fd, data, bev);
+  }
+
+#ifdef ENABLE_OPENSSL
+  auto ssl = bufferevent_openssl_get_ssl(bev);
+  if (ssl) {
+    // Save original flags and set socket to non-blocking for timeout support
+    int orig_flags = fcntl(fd, F_GETFL);
+    if (orig_flags == -1) return Status::FromErrno("fcntl(F_GETFL)");
+
+    auto s = SockSetBlocking(fd, 0);
+    if (!s.IsOK()) return s;
+
+    // Restore original flags on scope exit
+    auto restore_flags = MakeScopeExit([fd, orig_flags] { fcntl(fd, F_SETFL, orig_flags); });
+
+    ssize_t n = 0;
+    auto start = std::chrono::steady_clock::now();
+
+    while (n < static_cast<ssize_t>(data.size())) {
+      auto elapsed =
+          std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - start).count();
+      if (elapsed >= timeout_ms) {
+        return {Status::NotOK,
+                fmt::format("SSL send timeout after {} ms, sent {} of {} bytes", elapsed, n, data.size())};
+      }
+
+      int remaining_ms = timeout_ms - static_cast<int>(elapsed);
+      int ready = AeWait(fd, AE_WRITABLE, remaining_ms);
+      if (ready <= 0) {
+        return {Status::NotOK, fmt::format("SSL send timeout waiting for socket, sent {} of {} bytes", n, data.size())};
+      }
+
+      int nwritten = SSL_write(ssl, data.data() + n, static_cast<int>(data.size() - n));
+      if (nwritten <= 0) {
+        int err = SSL_get_error(ssl, nwritten);
+        if (err == SSL_ERROR_WANT_WRITE || err == SSL_ERROR_WANT_READ) {
+          continue;
+        }
+        return {Status::NotOK, fmt::format("SSL_write error: {}", err)};
+      }
+      n += nwritten;
+    }
+    return Status::OK();
+  }
+#endif
+  return SockSendWithTimeout(fd, data, timeout_ms);
 }
 
 StatusOr<int> SockConnect(const std::string &host, uint32_t port, [[maybe_unused]] ssl_st *ssl, int conn_timeout,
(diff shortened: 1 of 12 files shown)


src/common/io_util.cc
```diff
@@ -468,6 +471,93 @@ Status SockSend(int fd, const std::string &data, [[maybe_unused]] bufferevent *b
 #endif
 }
 
+Status SockSendWithTimeout(int fd, const std::string &data, int timeout_ms) {
+  ssize_t n = 0;
+  auto start = std::chrono::steady_clock::now();
+
```
git-hulk, line 482: Should we check if the value of timeout_ms is > 0?

src/common/io_util.cc
```diff
@@ -468,6 +471,93 @@ Status SockSend(int fd, const std::string &data, [[maybe_unused]] bufferevent *b
...
+  while (n < static_cast<ssize_t>(data.size())) {
```
git-hulk, line 483: We could reuse `WriteImpl` to simplify this implementation.

src/config/config.cc
```diff
@@ -206,6 +206,8 @@ Config::Config() {
       {"replication-no-slowdown", false, new YesNoField(&replication_no_slowdown, true)},
       {"replication-delay-bytes", false, new IntField(&max_replication_delay_bytes, 16 * 1024, 1, INT_MAX)},
       {"replication-delay-updates", false, new IntField(&max_replication_delay_updates, 16, 1, INT_MAX)},
+      {"max-replication-lag", false, new Int64Field(&max_replication_lag, 100000000, 1, INT64_MAX)},
```
git-hulk: Can we allow setting `max_replication_lag` to 0 for disabling this feature, and make it the default value for now?

ethervoid: Good suggestion - I've updated the code to skip the lag check when max_replication_lag == 0, so users can disable this feature if needed. I've set it to 0 by default so this feature is disabled by default

src/common/io_util.cc
ethervoid: I considered this but I don't think it's a great fit here. WriteImpl is a simple blocking loop that treats any -1 from write() as fatal. The timeout version needs fundamentally different behavior:
- Tracking elapsed time with steady_clock
- Polling for writability with poll()
- Handling EAGAIN/EWOULDBLOCK gracefully (continue waiting vs error)
- 
so reusing it would either lose behavior or require a bigger refactor. I’d rather prefer keep the explicit loop for now. What do you think?

src/common/io_util.cc
ethervoid: Good catch. I've added a guard in both SockSendWithTimeout overloads to fall back to the blocking SockSend when timeout_ms <= 0. This handles it defensively even though the config currently enforces >= 1000.

src/common/io_util.cc
git-hulk: @ethervoid Thanks for your reply and investigation. I forgot the difference. That makes sense to me.

src/config/config.cc
git-hulk: @ethervoid Thank you! We also need to update `100000000` => 0 and `1` => `0` at this line.

ethervoid: Done :)

git-hulk:
@PragmaTwice @torwig @caipengbo This PR would be helpful for users to identify what's going wrong when the replication is broken.

PragmaTwice (commented):
The code generally looks good to me.

PragmaTwice:
@sryanyuan Would you like to give a review on this PR since you were working on #3340?

sryanyuan:
> Would you like to give a review on this PR since you were working on #3340?

I’ll take a look and share any feedback I can

sryanyuan:
I’ve gone through the changes and here are some observations and suggestions:

* Master not logging send errors

    Currently, the master does not log any send errors until WAL sequence continuation fails.

    This seems to be caused by slow network transmission — each send operation takes a long time to complete, which accumulates delay over time. Eventually, WAL entries are cleaned up before the slave can catch up.

* Replication lag detection config

    The newly added configuration for replication lag detection to proactively disconnect a slave might help in some cases, but it may not fully solve the slow transmission problem.

* Potential issue on the slave side

    One major risk I see is that on the slave side, a half-open connection can remain for a long time before triggering a timeout and reconnect, which eventually leads to continuation failure.

    Adding a read timeout on the slave side could help mitigate this scenario.

* Master-side send timeout & efficiency improvements

    Adding a send timeout on the master side could also help in disconnecting half-open slave connections earlier.

    However, to truly improve transmission efficiency and avoid this situation, techniques such as compressing WAL logs before sending might be worth considering.

Overall, the changes go in the right direction for detecting and handling lag earlier, but I think addressing connection timeout handling (both master and slave) and optimizing WAL transmission could make the solution more robust.

git-hulk:
This PR also looks good to me. @ethervoid We need to fix this lint error in CI.

ethervoid:
@git-hulk All the tests working

@sryanyuan Thank you for the thorough report :bow: . I'll comment on your feedback

> Master not logging send errors
> 
> Currently, the master does not log any send errors until WAL sequence continuation fails.
> 
> This seems to be caused by slow network transmission — each send operation takes a long time to complete, which accumulates delay over time. Eventually, WAL entries are cleaned up before the slave can catch up.

I think this is addressed by the changes in this PR. In the SockSendWithTimeout method, we're logging an error is logged when the send fails, including by timeouts 

> 
> Replication lag detection config
> 
> The newly added configuration for replication lag detection to proactively disconnect a slave might help in some cases, but it may not fully solve the slow transmission problem.

Yeah, agree. This change fixes the symptom but doesn't fixes the root cause that could be outside the application control.

> 
> Potential issue on the slave side
> 
> One major risk I see is that on the slave side, a half-open connection can remain for a long time before triggering a timeout and reconnect, which eventually leads to continuation failure.
> 

I can work on another PR to include slave-side timeouts too

> Adding a read timeout on the slave side could help mitigate this scenario.
> 
> Master-side send timeout & efficiency improvements
> 
> Adding a send timeout on the master side could also help in disconnecting half-open slave connections earlier.
> 
> However, to truly improve transmission efficiency and avoid this situation, techniques such as compressing WAL logs before sending might be worth considering.

Agree with it. That would be a good feature to send the WAL batches compressed

git-hulk (approved):

PR #3372, opened 2026-02-15
refactor: use stderr for CLI error output

sanjana2505006:
I noticed that the CLI was printing error messages (like invalid arguments or config load failures) to `stdout`, which isn't ideal for scripting or standard error handling. This PR redirects those critical errors to `stderr`.

I also enhanced the [PrintUsage](cci:1://file:///Users/sanjana./kvrocks/src/cli/main.cc:62:0-73:1) function to accept an output stream, so `-h`/`--help` still goes to `stdout` (as expected), but incorrect usage warnings go to `stderr`. The usage message itself has been slightly polished for better readability.

Changes:
- PrintUsage now takes an optional `std::ostream&` argument (defaults to `cout`).
- Invalid command-line arguments now print usage to `cerr` and return exit code 1.
- Config loading and logger initialization errors now print to `cerr`.
- Added `<iostream>` include for correctness.


sanjana2505006:
@git-hulk Whenever you get a chance, please review and let me know if any changes are needed

src/cli/main.cc
```diff
@@ -59,15 +60,17 @@ struct NewOpt {
...
-  std::cout << program << " implements the Redis protocol based on rocksdb" << std::endl
-            << "Usage:" << std::endl
-            << std::left << new_opt << "-c, --config <filename>" << "set config file to <filename>, or `-` for stdin"
-            << std::endl
-            << new_opt << "-v, --version" << "print version information" << std::endl
-            << new_opt << "-h, --help" << "print this help message" << std::endl
-            << new_opt << "--<config-key> <config-value>"
-            << "overwrite specific config option <config-key> to <config-value>" << std::endl;
+static void PrintUsage(const char *program, std::ostream &os = std::cout) {
+  os << program << " implements the Redis protocol based on RocksDB" << std::endl
```
PragmaTwice, line 64: maybe we can just change it to `std::cerr`?

sanjana2505006: I actually went with the `std::ostream` parameter because I wanted to stick to the standard way most CLI tools handle output. Usually, when someone asks for help explicitly (like with `--help`), they expect it on `stdout` so they can pipe it to `grep` or `less`. But if they just mess up a command, the usage warning should definitely go to `stderr`.

I also noticed that `--version` uses stdout right now, so this keeps things consistent.  
Let me know if that makes sense! I'm happy to simplify it to just `std::cerr` if you'd prefer less boilerplate, but I figured this separation was worth the extra bit of code.

sanjana2505006:
@PragmaTwice whenever you get a chance, please let me know your thoughts on this. If you still think we should just use `std::cerr` for simplicity, I'm happy to make that change!

sanjana2505006:
@PragmaTwice I've addressed the linting failures and fixed a small regression in the `InitSpdlog` function logic. The code now respects the 120-character line limit as per the project guidelines. Please review whenever you have a chance!

sanjana2505006:
Whenever you have a moment, could you please take another look and let me know if everything looks good?
If there's anything else you'd like me to change or improve, I'm happy to update it.

jihuayu:
@sanjana2505006 Thanks for the careful catch. LGTM overall, but there are some formatting issues in the code.
Could you run `./x.py check format --clang-format-path clang-format-18` to make sure the CI passes?

sanjana2505006:
@jihuayu I ran `./x.py check format --clang-format-path /opt/homebrew/opt/llvm@18/bin/clang-format` and everything is fine here. I have attached the screenshot of the terminal:
Screenshot 2026-03-11 at 3 18 40 PM

jihuayu:
Same issue on my local machine as yours. I'm investigating it - could you tell me your `clang-format-18 --version` output?

sanjana2505006:
My clang-format version is 'Homebrew clang-format version 18.1.8'

sanjana2505006:
I am on macOS using Homebrew, so it was installed as just clang-format. The output of `/opt/homebrew/opt/llvm@18/bin/clang-format --version` is: `Homebrew clang-format version 18.1.8`

jihuayu:
The clang-format version used in CI is 18.1.3; it might be this issue. I will fix it first.

jihuayu (approved):
LGTM, if no one else has any objections.

sanjana2505006:
Ah got it, thank you @jihuayu  for looking into it and fixing the clang-format version. 😊

git-hulk (approved):

PR #3379, opened 2026-02-28
feat(info): add thread CPU time measurement for worker thread monitoring

sryanyuan:
## Summary
This PR introduces worker thread CPU time monitoring to identify bottlenecks and fixes precision issues in process CPU time calculations. 

## Problem Statement
1. **Lack of granular thread monitoring**: The current `INFO` command provides only overall process CPU time, which includes background threads. This makes it difficult to identify bottlenecks in worker threads specifically, as background thread activity can mask true worker thread utilization.

2. **Precision loss in CPU time calculation**: The existing CPU time calculation truncates microsecond values due to integer division (`tv_usec / 1000000`), causing millisecond data loss. This results in inaccurate reporting, especially for short operations where sub-second precision matters.

### 1. Worker Thread CPU Time Monitoring
- Added new `worker_cpu_time` metric to track CPU usage per worker thread
- Implemented per-thread CPU time collection using platform-specific APIs
- Formatted output with microsecond precision (6 decimal places) to match system precision
- Sample output: `worker_cpu_time:[0.123456,0.456789,1.234567]`

### 2. CPU Time Calculation Fix
- Fixed precision loss in `used_cpu_user` and `used_cpu_sys` calculations
- Replaced integer division with floating point division:  
  **Before**: `tv_usec / 1000000` → truncates to integer  
  **After**: `static_cast<double>(tv_usec) / 1e6` → preserves microseconds

## Benefits
- **Pinpoint worker thread bottlenecks**: Isolate worker thread CPU usage from background threads
- **Accurate performance analysis**: Microsecond-precision timing enables precise performance profiling
- **Better load diagnosis**: Distinguish between actual worker saturation and background activity

### DIFF
--- a/src/common/thread_util.cc
+++ b/src/common/thread_util.cc
@@ -41,7 +41,7 @@ void ThreadSetName(const char *name) {
 #ifdef __APPLE__
 double ThreadGetCPUTime(std::thread::native_handle_type thread_id) {
   if (!thread_id) {
-    return -1.0;
+    return 0.0;
   }
 
   mach_port_t mach_thread = pthread_mach_thread_np(thread_id);
@@ -50,7 +50,7 @@ double ThreadGetCPUTime(std::thread::native_handle_type thread_id) {
   mach_msg_type_number_t count = THREAD_BASIC_INFO_COUNT;
 
   if (thread_info(mach_thread, THREAD_BASIC_INFO, (thread_info_t)&info, &count) != KERN_SUCCESS) {
-    return -1.0;
+    return 0.0;
   }
 
   return (static_cast<double>(info.user_time.seconds) + static_cast<double>(info.user_time.microseconds) / 1e6) +
@@ -59,17 +59,17 @@ double ThreadGetCPUTime(std::thread::native_handle_type thread_id) {
 #else
 double ThreadGetCPUTime(std::thread::native_handle_type thread_id) {
   if (!thread_id) {
-    return -1.0;
+    return 0.0;
   }
 
   clockid_t clock_id;
   if (pthread_getcpuclockid(thread_id, &clock_id) != 0) {
-    return -1.0;
+    return 0.0;
   }
 
   timespec ts;
   if (clock_gettime(clock_id, &ts) != 0) {
-    return -1.0;
+    return 0.0;
   }
   return static_cast<double>(ts.tv_sec) + static_cast<double>(ts.tv_nsec) / 1e9;
 }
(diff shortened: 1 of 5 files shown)


src/common/thread_util.cc
```diff
@@ -33,6 +38,43 @@ void ThreadSetName(const char *name) {
...
+    return -1.0;
+  }
+
+  return (static_cast<double>(info.user_time.seconds) + static_cast<double>(info.user_time.microseconds) / 1e6) +
+         (static_cast<double>(info.system_time.seconds) + static_cast<double>(info.system_time.microseconds) / 1e6);
+}
+#else
+double ThreadGetCPUTime(std::thread::native_handle_type thread_id) {
+  if (!thread_id) {
+    return -1.0;
```
PragmaTwice: why -1.0 instead of 0?

src/server/server.cc
```diff
@@ -1402,10 +1402,18 @@ Server::InfoEntries Server::GetCpuInfo() {  // NOLINT(readability-convert-member
...
   entries.emplace_back("used_cpu_user", static_cast<float>(self_ru.ru_utime.tv_sec) +
-                                            static_cast<float>(self_ru.ru_utime.tv_usec / 1000000));
+                                            static_cast<float>(self_ru.ru_utime.tv_usec) / 1e6);
+
+  std::vector<double> thread_cpu_times(worker_threads_.size());
+  for (std::size_t i{0}; i < worker_threads_.size(); ++i) {
+    thread_cpu_times[i] = util::ThreadGetCPUTime(worker_threads_[i]->GetNativeHandle());
+  }
+  entries.emplace_back(
+      "worker_cpu_time",
```
PragmaTwice, line 1415:
```suggestion
      fmt::format("worker_cpu_time#{}", worker_threads_[i]->GetId()),
```

something like that?

PragmaTwice: ahh sorry misunderstood this.

src/common/thread_util.cc
sryanyuan: Thanks for the suggestion. Agreed - originally used -1.0 to explicitly flag errors, but given:
* Monitoring context makes 0.0 values statistically rare
* Any 0.0 result can serve as valid error indicator
* Simpler to treat 0 as "no measurable CPU time"

Will change to return 0.0 for:
* Invalid thread_id
* System call failures

This maintains functionality while simplifying error handling. Good catch!

PragmaTwice (approved):

Issue #3385, opened 2026-03-06
Notify nodes of slave failure via Controller to sync topology


Paragrf:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

I've observed that when the Controller detects a down slave node, it does not push this status update to the rest of the cluster. Consequently, clients are unable to receive topology updates through the server and must rely on their own health checking and traffic shedding mechanisms. This is problematic for simpler clients that lack sophisticated detection logic.

### Solution

Would it be more effective for KVRocks to maintain consistency with Redis behavior in this scenario? I suggest we introduce a new command or modify existing interfaces to allow the Controller to push "node down" information to the surviving KVRocks nodes. By doing so, clients can retrieve the updated cluster state directly from the remaining active nodes, thereby accelerating topology updates and minimizing the impact of node failures.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Paragrf:
@git-hulk @PragmaTwice What do you think of this idea? I'm open to any feedback or better alternatives.

git-hulk:
@Paragrf Could we extend the `CLUSETRX SETNODES` to allow changing the replica status?

Paragrf:
> [@Paragrf](https://github.com/Paragrf) Could we extend the `CLUSETRX SETNODES` to allow changing the replica status?

Alright, I'll work on implementing this next week.

jihuayu:
1. Redis is supposed to persist the fail status into nodes.conf, but it doesn't appear to be doing so right now. Do we need do it?
2. Are there specific fields in CLUSTER INFO (such as, if you set master,fail we need change cluster_state,  cluster_slots_ok, cluster_slots_fail) that should also be updated to reflect this?
@Paragrf @git-hulk

Paragrf:
> * Redis is supposed to persist the fail status into nodes.conf, but it doesn't appear to be doing so right now. Do we need do it?
> * Are there specific fields in CLUSTER INFO (such as, if you set master,fail we need change cluster_state,  cluster_slots_ok, cluster_slots_fail) that should also be updated to reflect this?

@jihuayu Thank you for your suggestions.

[...]

PR #3386, opened 2026-03-09
feat(cluster): propagate slave failure state via CLUSTER NODES fail flag

Paragrf:
### Summary
Based on https://github.com/apache/kvrocks/issues/3385

### Background
To maintain parity with Redis Cluster's behavior during slave node outages, this update enables Kvrocks to refresh its topology accordingly. This ensures clients stay synchronized with the server state and helps mitigate the effects of node failures.

### Implementation
Support the "slave,fail" role token in CLUSTERX SETNODES so that controllers can mark a downed slave in the cluster topology. Surviving nodes reflect the failed state as a "fail" flag in CLUSTER NODES output, allowing cluster-aware clients to update their routing tables without manual intervention.

### DIFF
--- a/src/cluster/cluster.cc
+++ b/src/cluster/cluster.cc
@@ -573,6 +573,7 @@ std::string Cluster::genNodesDescription() {
 
     // Flags
     if (node->id == myid_) node_str.append("myself,");
+    if (node->failed) node_str.append("fail,");
     if (node->role == kClusterMaster) {
       node_str.append("master - ");
     } else {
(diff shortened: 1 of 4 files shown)


src/cluster/cluster.cc
```diff
@@ -573,6 +573,7 @@ std::string Cluster::genNodesDescription() {
 
     // Flags
     if (node->id == myid_) node_str.append("myself,");
+    if (node->failed) node_str.append("fail,");
```
git-hulk: According to the Redis implementation, the `fail` flag should be after the role flag(master|slave). See https://github.com/redis/redis/blob/unstable/src/cluster_legacy.c#L5340C26-L5340C45

Paragrf: My apologies, it was an oversight on my part. I found that Lettuce actually supports this format, which is why the issue went unnoticed. No worries though, I've already fixed it.

Paragrf:
@git-hulk Everything looks solid now that the concerns are resolved. Could we proceed with merging this?

git-hulk (dismissed):

src/cluster/cluster.cc
```diff
@@ -843,7 +855,9 @@ Status Cluster::parseClusterNodes(const std::string &nodes_str, ClusterNodes *no
     }
 
     // Create master node
-    (*nodes)[id] = std::make_shared<ClusterNode>(id, host, port, role, master_id, slots);
+    auto master_node = std::make_shared<ClusterNode>(id, host, port, role, master_id, slots);
+    master_node->failed = node_failed;
+    (*nodes)[id] = master_node;
```
sryanyuan: Might emplace be more efficient here? nodes->emplace(id, std::move(node)); avoids the extra default construction step.

sryanyuan (dismissed):

src/cluster/cluster.cc
Paragrf: DONE

sryanyuan (approved):

git-hulk (approved):

thanujakat-oss (approved):

PR #3387, opened 2026-03-09
feat(command): Add subcommand abstraction 

jihuayu:
I have introduced the [subcommand framework](https://github.com/apache/kvrocks/discussions/3382) while making every effort to avoid modifying the existing command registration and execution logic. I plan to wait until the subcommand functionality is stable and fully migrated before performing a unified refactoring.

Currently, I have modified the namespace command as a reference example.

## Some important 

1. New concept Subcommand Family: Organizes a root command and its multiple subcommands into a single command group. A command family consists of three parts: the root command name, the subcommand parser, and the subcommand table.

2. New concept ResolvedCommand (Subcommand Resolution Result): Instead of simply identifying the "first token input by the user," this explicitly distinguishes between the "root command" and the "actual command to be executed." For example, NAMESPACE ADD would be resolved as root = namespace, with the final command being namespace|add.

3. New concept Canonical Name: Introduces a unified internal naming convention root|sub for subcommands (e.g., namespace|add). This name is used for internal capabilities such as COMMAND INFO, profiling, and statistics, removing the dependency on the raw input text.

4. Registration-time Conflict Prevention: Adds startup-time validation for scenarios such as duplicate command family registration, duplicate subcommand registration, or parent-child command mismatches to prevent silent overrides.

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -890,13 +890,13 @@ class CommandCommand : public Commander {
       } else if (sub_command == "info") {
         CommandTable::GetCommandsInfo(output, std::vector<std::string>(args_.begin() + 2, args_.end()));
       } else if (sub_command == "getkeys") {
-        auto cmd_iter = CommandTable::GetOriginal()->find(util::ToLower(args_[2]));
-        if (cmd_iter == CommandTable::GetOriginal()->end()) {
+        std::vector<std::string> cmd_tokens(args_.begin() + 2, args_.end());
+        auto resolved = CommandTable::Resolve(cmd_tokens);
+        if (!resolved) {
           return {Status::RedisUnknownCmd, "Invalid command specified"};
         }
 
-        auto key_indexes = GET_OR_RET(CommandTable::GetKeysFromCommand(
-            cmd_iter->second, std::vector<std::string>(args_.begin() + 2, args_.end())));
+        auto key_indexes = GET_OR_RET(CommandTable::GetKeysFromCommand(resolved->attributes, cmd_tokens));
 
         if (key_indexes.size() == 0) {
           return {Status::RedisExecErr, "Invalid arguments specified for command"};
@@ -905,7 +905,7 @@ class CommandCommand : public Commander {
         std::vector<std::string> keys;
         keys.reserve(key_indexes.size());
         for (const auto &key_index : key_indexes) {
-          keys.emplace_back(args_[key_index + 2]);
+          keys.emplace_back(cmd_tokens[key_index]);
         }
         *output = conn->MultiBulkString(keys);
       } else {
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -40,6 +40,17 @@
 
 namespace redis {
 
+namespace {
+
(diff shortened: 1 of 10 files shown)


src/commands/subcommand_registry.h
```diff
@@ -0,0 +1,62 @@
...
+namespace redis {
+
+enum class CommandCategory : uint8_t;
+struct CommandAttributes;
+
+using SubcommandResolver = std::function<std::optional<std::string>(const std::vector<std::string> &)>;
+using SubcommandMap = std::map<std::string, const CommandAttributes *, std::less<>>;
+
+std::optional<std::string> ResolveSubcommandByArgIndex(const std::vector<std::string> &args, size_t index);
+SubcommandResolver MakeArgIndexSubcommandResolver(size_t index);
```
PragmaTwice: These seem overly abstract to me.

src/commands/cmd_server.cc
```diff
@@ -1671,4 +1711,13 @@ REDIS_REGISTER_COMMANDS(
...
     MakeCmdAttr<CommandFlushMemTable>("flushmemtable", -1, "exclusive write", NO_KEY),
     MakeCmdAttr<CommandFlushBlockCache>("flushblockcache", 1, "exclusive write", NO_KEY), )
+
+REDIS_REGISTER_SUBCOMMANDS(
+    Server, "namespace", MakeArgIndexSubcommandResolver(1), DefaultSubcommandFallback(),
+    MakeSubCmdAttr<CommandNamespaceGet>("namespace", "get", 3, "read-only admin skip-monitor", NO_KEY),
+    MakeSubCmdAttr<CommandNamespaceSet>("namespace", "set", 4, "read-only admin skip-monitor", NO_KEY),
+    MakeSubCmdAttr<CommandNamespaceAdd>("namespace", "add", 4, "read-only admin skip-monitor", NO_KEY),
+    MakeSubCmdAttr<CommandNamespaceDel>("namespace", "del", 3, "read-only admin skip-monitor", NO_KEY),
+    MakeSubCmdAttr<CommandNamespaceCurrent>("namespace", "current", 2, "read-only skip-monitor", NO_KEY))
```
PragmaTwice: I think we should just use `REDIS_REGISTER_COMMANDS` instead a new `REDIS_REGISTER_SUBCOMMANDS`.

PragmaTwice (commented):
TBH I’m not really eager to review a huge chunk of LLM-generated code, but I took a look today anyway. I think some parts are handled in an overly complicated way, and I can’t really see the use cases that justify that level of complexity.

jihuayu:
@PragmaTwice Thank you very much for your feedback; I consider it crucial.

As you noted, the code was generated with LLM assistance, but every design choice was the result of my own deliberation and refinement. I reviewed every line myself. I did have some reservations, feeling that the code was becoming a bit too complex and difficult to review, but since I felt the changes were justified, I wanted to submit them to see what the community thought.

Your feedback made me realize that LLMs naturally tend to over-expand abstractions in an attempt to build something "all-encompassing." I've learned that I should minimize the scope of abstraction as much as possible to make the code more review-friendly. The use of an LLM isn't the mistake—excessive complexity is.

Thank you again for your review. I will be making adjustments accordingly.

PS: I believe that "too complex and difficult to review" is, in itself, excellent feedback.

Issue #3380, opened 2026-03-04
Potential Split-Brain Risk During Network Partitions in Kvrocks Cluster


redwood9:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

# Problem Statement
The current kvrocks and controller have a split-brain risk under network partition scenarios. Due to the lack of a lease mechanism, the old kvrocks master may continue accepting writes indefinitely.

# Design Principles
The following principles are derived from the analysis, balancing implementation cost against benefit.

1. kvrocks master gates write operations on a lease; no lease means writes are rejected (configurable). The controller probe must carry: current master ID, cluster version, and lease duration. When kvrocks receives a probe, and the master ID and version match, it extends the lease.
2. Shorten the controller session TTL to detect network partition events more quickly.
3. Allow manual designation of the active partition; the controller provides a one-command switch to move the kvrocks master to the specified partition, for use when a partition failure is already known.
4. kvrocks master validates its master status via an API call on each write (configurable); the controller exposes an API that returns who the current master is.

# Rationale
## The Core Problem: Old Master Keeps Writing During Network Partition
Consider the following sequence of events:
1. T=0: Network partition. Controller loses visibility of M. M is unaware and continues accepting writes.
2. T=3~12s: Probe fails 4 consecutive times. Only the failure counter increments; no action is taken.
3. T=15s: 5th failure. Controller elects S as the new master, writes to etcd, pushes SETNODES to S. S becomes master.
4. M never receives SETNODES. It still believes itself to be master and continues accepting writes. Two masters coexist; data diverges.
5. Network recovers. Controller detects M has a stale version and pushes the new topology. M is demoted to slave and syncs from S. All writes made during the partition are lost.

The faster the master detects the anomaly and stops writing, the sooner data divergence is prevented. The cost tradeoffs are:
1. Per-write: master checks with the controller on every write. Most effective, but adds network round-trip cost, increases etcd concurrency, and raises write latency.
2. Per-write: master checks a local lease cache; rejects writes if the cache is expired. The controller probe refreshes the cache. Better cost/benefit balance, but a brief split-brain window exists if a partition occurs during a valid cache period and a new master is assigned.
3. Options 1 and 2 are complementary. Use option 1 for high-consistency requirements, option 2 otherwise. This can be controlled by configuration.

## Impact of Single-Node Controller Probe
[...]

PR #3397, opened 2026-03-20
feat(cluster): Introduce lease mechanism for Kvrocks cluster to mitigate the brain-split issue

redwood9:
Based on the ideas I proposed in issue #3380
https://github.com/apache/kvrocks/issues/3380

this is the first step of the solution: 
introducing a lease mechanism for the Kvrocks master node.
This mechanism ensures that the maximum split-brain window is bounded by the lease TTL. It also comes with flexible configuration options:

1. No config / disabled — Behaves exactly like the current system. If the lease feature is not enabled, there is zero impact on existing behavior.
2. Reject writes — When the lease expires, the node rejects all write operations to prevent any data conflicts / divergence.
3. Log only — On lease expiration, the node only logs an error (no write rejection). This makes it easy to catch via monitoring / alerting systems without disrupting traffic.

To minimize invasion into the existing codebase structure, the implementation follows these strategies:

1. Controller probing — Add a new HEARTBEAT command (non-conflicting with existing standard Redis commands). The response format remains unchanged to keep controller integration simple and backward-compatible.
2. Unified lease check — Add the lease validation logic inside Storage::writeToDB(), so all write paths are guarded in one central place.

I will also submit the corresponding HEARTBEAT implementation in the controller at the same time / in parallel.


src/commands/cmd_cluster.cc
```diff
@@ -243,8 +244,25 @@ class CommandClusterX : public Commander {
       return CommandTable::ParseSlotRanges(args.back(), slot_ranges_);
     }
 
+    // CLUSTERX HEARTBEAT <master_node_id> <lease_ms> <election_version>
+    if (subcommand_ == "heartbeat") {
+      if (args.size() != 5) return {Status::RedisParseErr, errWrongNumOfArguments};
+      master_node_id_ = args[2];
+
+      auto parse_lease_ms = ParseInt<uint64_t>(args[3], 10);
```
git-hulk, line 252: Should have a proper range for the lease timeout?

src/storage/storage.cc
```diff
@@ -719,6 +719,23 @@ rocksdb::Status Storage::Write(engine::Context &ctx, const rocksdb::WriteOptions
 
 rocksdb::Status Storage::writeToDB(engine::Context &ctx, const rocksdb::WriteOptions &options,
                                    rocksdb::WriteBatch *updates) {
+  // Master lease check: applied here so it covers both Storage::Write() and CommitTxn().
+  // Only active when master_lease_mode != disabled. Read mode once to avoid TOCTOU.
+  auto lease_mode = config_->master_lease_mode;
```
git-hulk, line 724: The lease state should be maintained inside the server instead of the storage. And you can refuse the write operation like the slave's read-only mode. And the lease mode should only take effect if the cluster mode was enabled.

src/config/config.cc
```diff
@@ -93,6 +93,12 @@ const std::vector<ConfigEnum<BlockCacheType>> cache_types{[] {
 const std::vector<ConfigEnum<MigrationType>> migration_types{{"redis-command", MigrationType::kRedisCommand},
                                                              {"raw-key-value", MigrationType::kRawKeyValue}};
 
+const std::vector<ConfigEnum<MasterLeaseMode>> master_lease_modes{
+    {"disabled", MasterLeaseMode::kDisabled},
+    {"log-only", MasterLeaseMode::kLogOnly},
```
git-hulk, line 98: `log-only` could be `dry-run`?

git-hulk:
@redwood9 My another concern is that the Kvrocks will forcefully depends on the availability of the controller once the lease mode was enabled. And it cannot fully resolve the network partition issue, the network partition duration is totally depended on the lease timeout.

jihuayu:
@git-hulk I agree with your point. I think it’s better to make this an optional feature, especially since AP and CP are mutually exclusive. Users can then choose based on their specific requirements.

jihuayu (commented):
For me, the main issue is that write-protection should be handled at the command level rather than the storage layer.

git-hulk:
> @git-hulk I agree with your point. I think it’s better to make this an optional feature, especially since AP and CP are mutually exclusive. Users can then choose based on their specific requirements.

Kvrocks is now supporting the `WAIT` command to achieve the majority writes behavior.

jihuayu:
> Kvrocks is now supporting the `WAIT` command to achieve the majority writes behavior.Kvrocks 现在支持`等待`命令，以实现大多数写入行为。


Thanks for the reminder. I now think that the core value of this PR is introducing a lease mechanism to the server, which helps define the upper bound of the write window for a stale master. 
However, depended on a single controller poses a significant risk; I believe we should complete all feature only after supporting multiple controllers.

Or perhaps we should wait until the full design is finalized before considering this PR?

redwood9:
> @redwood9 My another concern is that the Kvrocks will forcefully depends on the availability of the controller once the lease mode was enabled. And it cannot fully resolve the network partition issue, the network partition duration is totally depended on the lease timeout.

@git-hulk @jihuayu 
This is a really good question and definitely worth discussing in depth. Here's my take:

1. The core of the lease mechanism is to allow the master to safely stop accepting writes during network partitions. Different use cases may call for different solutions:
   1. In Redis Sentinel mode, `min-replicas-to-write` is the primary mechanism. However, if there's only one replica and the master/replica are spread across two regions, it's easy to trigger failures even when `min-replicas-to-write==1`.

   2. For scenarios that demand extremely strong consistency and have low concurrency, we need to verify the master's identity on every write. This is exactly what step 4 in the proposal aims to solve.

   3. For cases that can tolerate a short partition but not a prolonged one, leases are a great fit and can be flexibly configured based on actual needs. This is the main scenario this PR is targeting.

   4. For situations where partitions are acceptable but we want alerts when they occur so humans can handle them, this PR also covers it. In this mode, the system behavior remains identical to the current implementation — we only add an alert when a lease anomaly is detected.

   5. For the scenario mentioned in #1.3, once the lease expires we can fall back to the approach in #1.2 (verifying master status on every write). This fallback can be implemented as part of the #1.2 work.

2. The controller is already a strongly consistent, multi-region system with high availability.

3. Overall, since users deploy Kvrocks in a wide variety of scenarios, I believe the best approach is to provide rich configuration options so they can make the right trade-offs for their specific use case.

4. Maybe we can dive deeper into the details. Once we settle on a concrete solution, I’d be happy to push a PR implementing it.

jihuayu:
@redwood9 I agree with your. I look forward to seeing your overall design.

Issue #1665, opened 2023-08-10
Make default value of config option in kvrocks.conf and config.cc identical


PragmaTwice:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Currently it is not same for some option value.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

git-hulk:
Resolved by https://github.com/apache/kvrocks/pull/1664

PragmaTwice:
> Resolved by #1664

I don't think so.

iihimanshuu:
this issue needs to be closed i guess

PR #3398, opened 2026-03-22
chore(config): remove the deprecated configuration `rocksdb.max_background_{compactions,flushes}`  and set default to -1

yatriks:
This PR will:
- update the default values of `rocksdb.max_background_compactions` and `rocksdb.max_background_flushes` from `2` to `-1`
- unset them from `kvrocks.conf` and update comments to reflect RocksDB's decision logic for max background threads

NOTE: This doesn't make it so that `rocksdb.max_background_compactions` is unsettable, like `rocksdb.max_background_flushes` currently is. This may be desirable for people who do currently use the old overrides, but afaik it would lock people in to `max_background_flushes = 2` - but that's already the case as far as I can tell. 

I ran each of:
```
./x.py format (unrelated changes)
./x.py build --unittest
./x.py test cpp 
```
Tests all pass and formatting changes are unrelated.

This is effectively an addition to https://github.com/apache/kvrocks/issues/1665
See https://github.com/apache/kvrocks/discussions/3395 for more background.


yatriks:
Also I cannot recall if setting READONLY on the option makes it read-only for everything or just CLI...that would inform whether we update max_background_compactions to readonly. I can verify later.

kvrocks.conf
```diff
@@ -820,23 +820,21 @@ rocksdb.min_write_buffer_number_to_merge 1
...
-rocksdb.max_background_compactions -1
+# rocksdb.max_background_compactions -1
 
 # DEPRECATED: it is automatically decided based on the value of rocksdb.max_background_jobs
 # Maximum number of concurrent background memtable flush jobs, submitted by
 # default to the HIGH priority thread pool. If the HIGH priority thread pool
 # is configured to have zero threads, flush jobs will share the LOW priority
 # thread pool with compaction jobs.
-rocksdb.max_background_flushes -1
+# rocksdb.max_background_flushes -1
```
git-hulk: I think we can remove the entire section since we don't expect any new users to use this configuration.

git-hulk (dismissed):

yatriks:
@git-hulk Sorry will need another review.
 also set `max_background_compactions` to READONLY (which I checked doesn't preclude setting it from kvrocks.conf) for consistency. Feel free to revert this if it's problematic for a reason I did not foresee.

src/config/config.cc
```diff
@@ -262,8 +262,8 @@ Config::Config() {
       {"rocksdb.min_write_buffer_number_to_merge", false,
        new IntField(&rocks_db.min_write_buffer_number_to_merge, 1, 1, 256)},
       {"rocksdb.target_file_size_base", false, new IntField(&rocks_db.target_file_size_base, 128, 1, 1024)},
-      {"rocksdb.max_background_compactions", false, new IntField(&rocks_db.max_background_compactions, 2, -1, 32)},
-      {"rocksdb.max_background_flushes", true, new IntField(&rocks_db.max_background_flushes, 2, -1, 32)},
+      {"rocksdb.max_background_compactions", true, new IntField(&rocks_db.max_background_compactions, -1, -1, 32)},
```
git-hulk: @yatriks I think we don't need to change this to avoid breaking the old behavior.

```suggestion
      {"rocksdb.max_background_compactions", false, new IntField(&rocks_db.max_background_compactions, -1, -1, 32)},
```

yatriks: 👍 done, thanks!

git-hulk (approved):

PR #3405, opened 2026-03-29
feat(cms): add Count-Min Sketch data structure

Tangruilin:
Implement CMS (Count-Min Sketch) probabilistic data structure with the following commands:
- CMS.INITBYDIM key width depth
- CMS.INITBYPROB key error probability
- CMS.INCRBY key item increment [...]
- CMS.QUERY key item [...]
- CMS.MERGE dest numkeys src... [WEIGHTS ...]
- CMS.INFO key

Storage design:
- Metadata stored in Metadata CF
- Count Matrix stored in PrimarySubkey CF (per-bucket storage)
- Uses MurmurHash64 for layer-specific hashing

Related: #2425

### DIFF
--- a/src/types/redis_cms.cc
+++ b/src/types/redis_cms.cc
@@ -1,0 +1,542 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#include "redis_cms.h"
+
+#include <algorithm>
+#include <cmath>
+
+#include "vendor/murmurhash2.h"
+
+namespace redis {
+
+// ============================================================================
+// Private helper methods
+// ============================================================================
+
+rocksdb::Status CMS::getCMSMetadata(engine::Context &ctx, const Slice &ns_key, CMSMetadata *metadata) {
+  return Database::GetMetadata(ctx, {kRedisCMS}, ns_key, metadata);
+}
+
+std::string CMS::getBucketKey(const Slice &ns_key, uint64_t version, uint32_t bucket_id) {
+  std::string sub_key;
+  PutFixed32(&sub_key, bucket_id);
+  return InternalKey(ns_key, sub_key, version, storage_->IsSlotIdEncoded()).Encode();
+}
+
+uint64_t CMS::hashItem(const Slice &item, uint32_t layer) {
+  // Use MurmurHash64 with layer as seed
+  // This ensures different hash functions for each layer
+  return HllMurMurHash64A(item.data(), static_cast<int>(item.size()), layer);
+}
+
+uint32_t CMS::getCol(const Slice &item, uint32_t layer, uint32_t width) {
+  uint64_t hash = hashItem(item, layer);
+  return static_cast<uint32_t>(hash % width);
+}
+
+std::pair<uint32_t, uint32_t> CMS::calcDimFromProb(double error_rate, double probability) {
+  // Formula from RedisBloom implementation:
+  //   width = ceil(2 / error_rate)
+  //   depth = ceil(log10(probability) / log10(0.5))
+  //
+  // Note: RedisBloom uses simplified approximations of the CMS paper formulas.
+  // This differs from the theoretical formulas (width = e/error, depth = ln(1/probability)).
+  //
+  // Per Redis documentation (https://redis.io/docs/latest/commands/cms.initbyprob/):
+  //   - error: Estimate size of error, as a percent of total counted items
+  //   - probability: The desired probability for inflated count (i.e., the probability
+  //                   that the estimate exceeds the true count by more than error_rate).
+  auto width = static_cast<uint32_t>(std::ceil(2.0 / error_rate));
+  auto depth = static_cast<uint32_t>(std::ceil(std::log10(probability) / std::log10(0.5)));
+
+  // Clamp to allowed range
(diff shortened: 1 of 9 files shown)


Tangruilin:
I noticed that operations like ZINCRBY (ZSet) and HINCRBY (Hash) don't use any locks, even though they are read-modify-write operations. From my testing with concurrent unit tests, I
 haven't observed any race conditions.

 Questions about Kvrocks threading model:

 1. Is Kvrocks single-threaded when executing commands?
 2. How does Kvrocks ensure thread safety for read-modify-write operations like HINCRBY and ZINCRBY without explicit locking?
 3. Do we need locks for CMS operations (INCRBY, MERGE) similar to how TDigest uses LockGuard?

 Current locking status across types:

 | Type | INCRBY has lock | Notes |
 |------|-----------------|------ -|
 | Hash | ❌ | HINCRBY - read-modify-write |
 | ZSet | ❌ | ZINCRBY - read-modify-write + dual index update |
 | TDigest | ✅ | Uses LockGuard |
 | CMS (this PR) | ❌ | Currently no lock |

 I've included concurrent tests for CMS in tests/gocase/unit/type/cms/cms_test.go to verify correctness. Would appreciate clarification on whether locks are needed for CMS operations.

Tangruilin:
AI was used in this patch for specific code implementation and approach analysis. The main design was evaluated by humans, discussion refer:https://github.com/apache/kvrocks/discussions/3404

tests/cppunit/types/cms_test.cc
```diff
@@ -0,0 +1,255 @@
...
+  void SetUp() override {
+    TestBase::SetUp();
+    [[maybe_unused]] auto s = cms_->Del(*ctx_, "cms");
+    for (int x = 1; x <= 3; x++) {
+      s = cms_->Del(*ctx_, "cms" + std::to_string(x));
+    }
+  }
+
+  void TearDown() override {
+    TestBase::SetUp();
```
Tangruilin: error. should be TearDown

src/types/redis_cms.cc
```diff
@@ -0,0 +1,447 @@
...
+    return rocksdb::Status::InvalidArgument("probability must be between 0 and 1");
+  }
+
+  auto [width, depth] = calcDimFromProb(error_rate, probability);
+  return InitByDim(ctx, key, width, depth);
+}
+
+/// CMS.INCRBY - Increment counters for given items
+/// Usage: CMS.INCRBY key item increment [item increment ...]
+rocksdb::Status CMS::IncrBy(engine::Context &ctx, const Slice &key,
```
Tangruilin, line 199: I assessed that a lock is needed here. However, I ran concurrent tests and also reviewed Kvrocks' HSET, HASH and other modules. I found that their write operations don't use locks. Is this by design? Is concurrency handled at some other level?

any suggestion？ @PragmaTwice @git-hulk

src/types/redis_cms.cc
```diff
@@ -0,0 +1,447 @@
...
+uint32_t CMS::getCol(const Slice &item, uint32_t layer, uint32_t width) {
+  uint64_t hash = hashItem(item, layer);
+  return static_cast<uint32_t>(hash % width);
+}
+
+std::pair<uint32_t, uint32_t> CMS::calcDimFromProb(double error_rate, double probability) {
+  // Formula from Count-Min Sketch paper:
+  // width = ceil(e / error_rate)
+  // depth = ceil(ln(1 - probability) / ln(1 - 1/e))
+  auto width = static_cast<uint32_t>(std::ceil(2.718281828 / error_rate));
```
Tangruilin: use `std::numbers::e`

src/types/redis_cms.cc
```diff
@@ -0,0 +1,447 @@
...
+  // This ensures different hash functions for each layer
+  return HllMurMurHash64A(item.data(), static_cast<int>(item.size()), layer);
+}
+
+uint32_t CMS::getCol(const Slice &item, uint32_t layer, uint32_t width) {
+  uint64_t hash = hashItem(item, layer);
+  return static_cast<uint32_t>(hash % width);
+}
+
+std::pair<uint32_t, uint32_t> CMS::calcDimFromProb(double error_rate, double probability) {
```
Tangruilin, line 55: `probability` use success probability here.
should be fail_probability

refer to: https://redis.io/docs/latest/commands/cms.initbyprob/

src/types/redis_cms.cc
Tangruilin: RedisBloom uses ceil(2 / error) for the width calculation. I believe this is because C doesn't have a built-in constant for e (Euler's number). In our C++ implementation, using e would
 be more mathematically accurate according to the original CMS paper.

 Should we follow the paper's formula with e, or should we fully match Redis's implementation for compatibility?

@PragmaTwice

src/types/redis_cms.cc
git-hulk: @Tangruilin Thanks for your PR. I will take a look at your proposal first, you could read [Guidelines for AI-assisted Contributions](https://kvrocks.apache.org/community/contributing#guidelines-for-ai-assisted-contributions).

Tangruilin: I have alread read it before do this.

and I have a proposal already: https://github.com/apache/kvrocks/discussions/3404

Ths for look

Tangruilin: Hi @git-hulk,

   Thank you for taking the time to review this PR.

   For this patch, you could focus on reviewing the proposal-related parts first. I'll address any feedback you have on the design and implementation approach.

   Regarding the concurrency concern I mentioned earlier — I noticed that commands like `HSET`, `HINCRBY`, and `ZINCRBY` in Kvrocks don't use explicit locks, even though they involve
 read-modify-write operations. Is this by design? Is thread safety handled at a different level (e.g., command execution model)?

jihuayu:
Hi @Tangruilin 
This PR is a bit too massive. Could you implement only the minimum viable features in the first PR?
Please make sure the clang-tidy and format checks pass.

Kvrocks' write commands have key-level locks at the framework level.

Tangruilin:
> Hi @Tangruilin This PR is a bit too massive. Could you implement only the minimum viable features in the first PR? Please make sure the clang-tidy and format checks pass.
> 
> Kvrocks' write commands have key-level locks at the framework level.

Thanks for the review! I'll split this PR. The first PR will focus on the basic CMS and INIT commands only.

I've marked this PR as draft for now. Once the split is complete, I'll change it back to ready for review.

In the meantime, you could review the proposal first: https://github.com/apache/kvrocks/discussions/3404

Tangruilin:
This PR has grown too large. Let me sync with the committers on the overall design first, then split it into smaller, reviewable PRs.

Issue #3410, opened 2026-03-30
The TDIGEST.MERGE command in Kvrocks behaves differently from Redis when the destination key already exists.


Tangruilin:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

unstable

### Minimal reproduce step

 ```bash
#!/bin/bash
# TDIGEST.MERGE bug reproduction script
# Usage: ./reproduce_tdigest_merge.sh [redis-cli path] [port]

CLI="${1:-redis-cli}"
PORT="${2:-6379}"
PREFIX="$CLI -p $PORT"

echo "============================================"
echo "TDIGEST.MERGE Bug Reproduction"
echo "============================================"

echo -e "\n--- Step 1: Clean up ---"
$PREFIX DEL td_src1 td_src2 td_dest

echo -e "\n--- Step 2: Create source sketches ---"
$PREFIX TDIGEST.CREATE td_src1 COMPRESSION 100
$PREFIX TDIGEST.CREATE td_src2 COMPRESSION 100
$PREFIX TDIGEST.ADD td_src1 1 2 3
$PREFIX TDIGEST.ADD td_src2 4 5 6

echo -e "\n--- Step 3: Create destination sketch ---"
$PREFIX TDIGEST.CREATE td_dest COMPRESSION 100

echo -e "\n--- Step 4: Merge into existing destination ---"
echo "Command: TDIGEST.MERGE td_dest 2 td_src1 td_src2"
$PREFIX TDIGEST.MERGE td_dest 2 td_src1 td_src2

echo -e "\n--- Step 5: Check destination ---"
$PREFIX TDIGEST.INFO td_dest

echo -e "\n--- Step 6: Clean up ---"
$PREFIX DEL td_src1 td_src2 td_dest

echo -e "\n============================================"
echo "Test completed"
echo "============================================"
 ```

### What did you expect to see?

**Redis result**

<img width="866" height="760" alt="Image" src="https://github.com/user-attachments/assets/caeba60c-e7a5-46ef-ba00-e118137ecb15" />

### What did you see instead?

**kvrocks result**

<img width="724" height="738" alt="Image" src="https://github.com/user-attachments/assets/1c65cfca-4948-4c54-81ac-47878b64d91d" />

### Anything Else?

no

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Tangruilin:
[...]

PR #3412, opened 2026-03-31
fix(tdigest): merge into existing dest without OVERRIDE

Tangruilin:
WIP: fix TDIGEST.MERGE to match Redis behavior

Issue: #3410

When destination key already exists without OVERRIDE flag:
- Redis: merge dest + sources together
- Kvrocks (before): return error

Changes:
1. Remove error return when dest exists without OVERRIDE
2. Read dest centroids and add to merge list
3. Add check to prevent duplicate merge if dest in sources
4. Update tests to verify merge and override behavior


Tangruilin:
@jihuayu Hi, this PR is ready for review. Thanks!

Tangruilin:
@jihuayu  "I'm sorry the CI didn't pass.

 However, it seems the CI format configuration is not using the default configuration. I ran ./x.py check format locally and it passed, but the CI command is ./x.py check format
 --clang-format-path clang-format-18.

 This can be confusing. Below is a screenshot of my local run. Perhaps we should keep the CI consistent with the default configuration and use configuration files instead of command-line
 arguments."
 
image

Tangruilin:
@jihuayu 
"The lint issue was introduced by MR [https://github.com/apache/kvrocks/pull/3419](https://github.com/apache/kvrocks/pull/3419) It seems this MR was merged without passing the format check.

 Should kvrocks consider enforcing a rule that blocks merging when CI doesn't pass, to avoid similar CI pollution issues?"

Tangruilin:
> @jihuayu "The lint issue was introduced by MR #3419 It seems this MR was merged without passing the format check.
> 
> Should kvrocks consider enforcing a rule that blocks merging when CI doesn't pass, to avoid similar CI pollution issues?"

image

tests/gocase/unit/type/tdigest/tdigest_test.go
```diff
@@ -463,16 +463,44 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
...
 
-		// merge the source into the destination without override
-		require.ErrorContains(t, rdb.Do(ctx, "TDIGEST.MERGE", destKey, 2, sourceKey1, sourceKey2).Err(), errMsgKeyAlreadyExists)
+		// merge the source into the existing destination without override
+		// should merge dest + sources together (not return error)
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.MERGE", destKey, 2, sourceKey1, sourceKey2).Err())
 
-		// merge the source into the destination with override
+		// verify the merged result contains data from both dest and sources
+		rsp := rdb.Do(ctx, "TDIGEST.INFO", destKey)
```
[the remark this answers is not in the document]
Tangruilin: done

jihuayu:
@Tangruilin Yes, you are right!

src/types/redis_tdigest.cc
```diff
@@ -473,21 +472,8 @@ rocksdb::Status TDigest::Merge(engine::Context& ctx, const Slice& dest_digest,
...
-      std::string metadata_bytes;
-      metadata.Encode(&metadata_bytes);
-      if (auto status = batch->Put(metadata_cf_handle_, source_ns_key, metadata_bytes); !status.ok()) {
-        return status;
-      }
-    } else if (metadata.merged_nodes > 0) {
-      if (auto status = dumpCentroids(ctx, source_ns_key, metadata, &source_centroids); !status.ok()) {
-        return status;
-      }
+    if (auto status = getCentroidsForMerge(ctx, source_ns_key, batch, &metadata, &source_centroids); !status.ok()) {
```
LindaSummer, line 475: This is great to reduce redundancy of this action.👍

src/types/tdigest.cc
```diff
@@ -362,7 +362,7 @@ class TDigest {
   TDigestImpl impl_;
 };
 
-TDigest::TDigest(uint64_t delta) : impl_(TDigestImpl(delta)) { Reset({}); }
+TDigest::TDigest(uint64_t delta) : impl_(TDigestImpl(delta)) { Reset(); }
```
LindaSummer, line 365: Maybe we could do refactor in a new separate PR.

LindaSummer (commented):
Hi @Tangruilin ,

Generally LGTM with some comments. 😊

Could we add some cpp unit tests for this change?

src/types/tdigest.cc
Tangruilin: this isn't a refactor but rather a bugfix. When calling Reset({}), both min and max are initialized to 0. This causes a problem: if the source's min is greater than 0, the actual minimum value during merge would be incorrectly updated to 0.

I discovered this issue while adding a related test case. If this fix isn't included in this PR, my test won't pass.

Tangruilin:
> Hi @Tangruilin ,
> 
> Generally LGTM with some comments. 😊
> 
> Could we add some cpp unit tests for this change?

Sure. I will add cpp unit before merge

jihuayu:
@Tangruilin  It looks like the code incorrectly overrides the COMPRESSION parameter in the following scenario. Could you add more test cases for this?
```txt
TDIGEST.CREATE src COMPRESSION 200
TDIGEST.ADD src 1

TDIGEST.CREATE dest COMPRESSION 100
TDIGEST.ADD dest 2

TDIGEST.INFO dest
TDIGEST.MERGE dest 1 src
TDIGEST.INFO dest
```
The Redis results are as follows:
```txt
127.0.0.1:6379> TDIGEST.INFO dest
 1) Compression
 2) (integer) 100
```

src/types/tdigest.cc
LindaSummer: Nice catch!
Could we create a new PR after this one and make it reset with `nan` or `+/-inf` to avoid this problem?

Tangruilin:
> @Tangruilin It looks like the code incorrectly overrides the COMPRESSION parameter in the following scenario. Could you add more test cases for this?
> 
> ```
> TDIGEST.CREATE src COMPRESSION 200
> TDIGEST.ADD src 1
> 
> TDIGEST.CREATE dest COMPRESSION 100
> TDIGEST.ADD dest 2
> 
> TDIGEST.INFO dest
> TDIGEST.MERGE dest 1 src
> TDIGEST.INFO dest
> ```
> 
> The Redis results are as follows:
> 
> ```
> 127.0.0.1:6379> TDIGEST.INFO dest
>  1) Compression
>  2) (integer) 100
> ```

Done

src/types/tdigest.cc
Tangruilin: Actually, kvrocks already has a struct for initialization:
 ```cpp
  struct CentroidsWithDelta {
    std::vector<Centroid> centroids;
    uint64_t delta;
    double min;
    double max;
    double total_weight;
  };
```
  Unfortunately, the min and max initialization in this struct is also incorrect. I'll submit a follow-up PR to fix the initialization values of this struct and disable implicit deduction for this function to prevent future misuse.

**Comment by Tangruilin** [comment]
@jihuayu ji
The CI seems to fail because of  env

**Review by jihuayu — DISMISSED** [review]
It looks OK. 
However, the logic in this function is a bit complex, so I’ll need some help from other reviewers.

**Review by LindaSummer — DISMISSED** [review]
Hi @Tangruilin ,

LGTM 🚀

With the compression parameter improvement, its behavior should match the Redis' now.

File: src/types/redis_tdigest.h [hunk 4]
```diff
@@ -124,6 +124,23 @@ class TDigest : public SubKeyScanner {
...
+   * @brief Get centroids for merge operation.
+   *
+   * If the tdigest has unmerged buffer, merge it first and update metadata to batch.
+   * Otherwise, just dump existing centroids.
+   * @param ctx The context of the operation.
+   * @param ns_key The namespace key of the t-digest.
+   * @param batch The write batch to store metadata updates.
+   * @param metadata The metadata of the t-digest (may be updated if buffer is merged).
+   * @param centroids The output vector to store the centroids.
+   * @return rocksdb::Status
```
PragmaTwice:
```suggestion
```

Tangruilin: Ths

jihuayu (approved):

PR #3413, opened 2026-04-01
feat(commands): introduce subcommand as a primary abstraction

jihuayu:
This PR introduces subcommand as a first-class abstraction.

Key changes include:
- Supported standardized registration and parsing processes for subcommands.
- Implemented metadata management for subcommand structures.
- Refactored namespace as a reference implementation for this new abstraction.

### AI Disclosure
- The architectural design of this PR was completed independently. Some code segments were developed with AI.
- All test cases in this PR were either autonomously generated by AI or generated based on natural language descriptions provided by me.


src/commands/cmd_server.cc
```diff
@@ -1670,5 +1731,11 @@ REDIS_REGISTER_COMMANDS(
     MakeCmdAttr<CommandPollUpdates>("pollupdates", -2, "read-only admin", NO_KEY),
     MakeCmdAttr<CommandSST>("sst", -3, "write exclusive admin", 1, 1, 1),
     MakeCmdAttr<CommandFlushMemTable>("flushmemtable", -1, "exclusive write", NO_KEY),
-    MakeCmdAttr<CommandFlushBlockCache>("flushblockcache", 1, "exclusive write", NO_KEY), )
+    MakeCmdAttr<CommandFlushBlockCache>("flushblockcache", 1, "exclusive write", NO_KEY),
+    MakeSubCmdAttr<CommandNamespaceGet>("namespace", "get", 3, "read-only admin skip-monitor", NO_KEY),
+    MakeSubCmdAttr<CommandNamespaceSet>("namespace", "set", 4, "read-only admin skip-monitor", NO_KEY),
+    MakeSubCmdAttr<CommandNamespaceAdd>("namespace", "add", 4, "read-only admin skip-monitor", NO_KEY),
+    MakeSubCmdAttr<CommandNamespaceDel>("namespace", "del", 3, "read-only admin skip-monitor", NO_KEY),
```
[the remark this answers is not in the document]
jihuayu: @PragmaTwice I have the same question. However, the existing namespace command is read-only. Do we need to modify it?

src/commands/commander.cc
```diff
@@ -83,6 +149,32 @@ void CommandTable::GetCommandsInfo(std::string *info, const std::vector<std::str
   }
 }
 
+StatusOr<ResolvedCommand> CommandTable::Resolve(const std::vector<std::string> &cmd_tokens) {
+  if (cmd_tokens.empty()) {
+    return {Status::RedisUnknownCmd};
+  }
+
```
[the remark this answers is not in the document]
jihuayu: To avoid making this PR too large, I’ll handle this in a follow-up PR.

jihuayu:
Hi @git-hulk. I noticed that all Redis CLI-related test cases are failing on the Ubuntu platform, even though they pass just fine on my local Ubuntu machine. Any ideas on what might be causing this?

git-hulk:
> Hi @git-hulk. I noticed that all Redis CLI-related test cases are failing on the Ubuntu platform, even though they pass just fine on my local Ubuntu machine. Any ideas on what might be causing this?

It seems like a flaky test in Ubuntu 22.04, but I also cannot tell without diving deep.

src/commands/commander.cc
```diff
@@ -20,22 +20,88 @@
...
+    std::abort();
+  }
+
+  return {normalized_parent, normalized_sub};
+}
+
+const CommandAttributes *CommandTable::registerCommand(CommandAttributes attr, CommandCategory category) {
+  if (original_commands.contains(attr.name) || commands.contains(attr.name)) {
+    std::cout << fmt::format("Duplicate command registration for '{}'", attr.name) << std::endl;
+    std::abort();
```
Tangruilin, line 52: "Are std::cout and std::abort used for debugging?"

src/commands/cmd_server.cc
```diff
@@ -65,63 +76,113 @@ class CommandAuth : public Commander {
...
+
+class CommandNamespaceCurrent : public Commander {
+ public:
+  Status Execute([[maybe_unused]] engine::Context &ctx, Server *srv, Connection *conn, std::string *output) override {
+    if (IsNamespaceReadOnlyOnSlave(srv)) {
+      return {Status::RedisExecErr, "namespace is read-only for slave"};
+    }
+    if (IsNamespaceCommandDisabled(srv)) {
+      return {Status::RedisExecErr, "namespace command is not allowed when redis-databases > 0"};
+    }
```
Tangruilin, line 180: "Both IsNamespaceReadOnlyOnSlave and IsNamespaceCommandDisabled checks seem to be needed by all subcommands. Could we consider abstracting them into a common function?"

src/commands/commander.cc
jihuayu: This serves as a reminder for developers to avoid this error, as it can lead to unpredictable behavior. 
In standard execution flow, this code should never be reached.

src/commands/cmd_server.cc
jihuayu: Namespace GET not need IsNamespaceReadOnlyOnSlave

src/commands/commander.cc
Tangruilin: "Typically, for such cases, we should use FATAL logging and assert_debug for safeguards, rather than std::cout and std::abort. Here are two main reasons:

 1. For a server service, we don't print std::cout to the terminal, so using std::cout vs. a logging library doesn't make a fundamental difference. Even if we want terminal output, I
 think error messages should go to std::cerr rather than std::cout. (PS: I believe the best practice is to log at FATAL level.)
 2. Although this is code that should never be reached, at the server level, nothing is worse than a crash. If this path is somehow triggered in production (meaning kvrocks has a bug),
 the application shouldn't crash. A better approach would be to use assert_debug, which only crashes in debug mode. In release mode, the impact should be minimized.

 What do you think?"

Tangruilin: "However, it looks like this code path will only be reached during initialization. Both approaches seem acceptable."

src/commands/commander.cc
```diff
@@ -20,22 +20,88 @@
...
+
+std::pair<std::string, std::string> CommandTable::parseSubcommandName(const std::string &name) {
+  auto delimiter = name.find('|');
+  if (delimiter == std::string::npos || delimiter == 0 || delimiter + 1 >= name.size()) {
+    std::cout << fmt::format("Encountered invalid subcommand name '{}'", name) << std::endl;
+    std::abort();
+  }
+
+  auto normalized_parent = util::ToLower(name.substr(0, delimiter));
+  auto normalized_sub = util::ToLower(name.substr(delimiter + 1));
```
Tangruilin, line 40: The check here looks redundant. If `delimiter + 1 >= name.size()` is true, then `name.substr(delimiter + 1) `must be empty.

src/commands/commander.cc
```diff
@@ -190,17 +190,17 @@ StatusOr<std::vector<int>> CommandTable::GetKeysFromCommand(const CommandAttribu
...
 
-  bool has_key_args = true;
+  Status status;
   std::vector<int> key_indexes;
 
   attributes->ForEachKeyRange(
-      [&key_indexes, &cmd_tokens](const std::vector<std::string> &, CommandKeyRange key_range) {
-        key_range.ForEachKeyIndex([&key_indexes](int i) { key_indexes.push_back(i); }, cmd_tokens.size());
+      [&](const std::vector<std::string> &, CommandKeyRange key_range) {
+        key_range.ForEachKeyIndex([&](int i) { key_indexes.push_back(i); }, cmd_tokens.size());
```
Tangruilin, line 198: Why change the lambda capture back here?

jihuayu: This is not a necessary change. Keep modifications to a minimum to make the review process easier.

PR #3419, opened 2026-04-03
fix(conf): cap proto-max-bulk-len to 4 GiB instead of UINT64_MAX 

songqing:
- Cap proto_max_bulk_len to 4G so that bulk_len_+2 used for
      CRLF accounting in RESP parsing can never wrap around, see` if (evbuffer_get_length(input) < bulk_len_ + 2) return Status::OK();` in [code](https://github.com/apache/kvrocks/blob/da5e46307047b3e89753f0c9c02996430e758a5f/src/server/redis_request.cc#L123)
    4 GiB is chosen because:
    - RocksDB recommends keeping individual values under 3 GiB for
      performance and memory reasons.
    - It eliminates both the uint64_t overflow and the ssize_t cast hazard.
    - It is still far above any realistic bulk string size in practice.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -239,7 +239,7 @@ Config::Config() {
       {"resp3-enabled", false, new YesNoField(&resp3_enabled, true)},
       {"repl-namespace-enabled", false, new YesNoField(&repl_namespace_enabled, false)},
       {"proto-max-bulk-len", false,
-       new IntWithUnitField<uint64_t>(&proto_max_bulk_len, std::to_string(512 * MiB), 1 * MiB, UINT64_MAX)},
+       new IntWithUnitField<uint64_t>(&proto_max_bulk_len, std::to_string(512 * MiB), 1 * MiB, UINT64_MAX - 2)},
       {"json-max-nesting-depth", false, new IntField(&json_max_nesting_depth, 1024, 0, INT_MAX)},
       {"json-storage-format", false,
        new EnumField<JsonStorageFormat>(&json_storage_format, json_storage_formats, JsonStorageFormat::JSON)},
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -239,7 +239,7 @@ Config::Config() {
       {"resp3-enabled", false, new YesNoField(&resp3_enabled, true)},
       {"repl-namespace-enabled", false, new YesNoField(&repl_namespace_enabled, false)},
       {"proto-max-bulk-len", false,
-       new IntWithUnitField<uint64_t>(&proto_max_bulk_len, std::to_string(512 * MiB), 1 * MiB, UINT64_MAX - 2)},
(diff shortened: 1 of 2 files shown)


src/server/server.cc
```diff
@@ -1751,10 +1751,11 @@ StatusOr<std::vector<rocksdb::BatchResult>> Server::PollUpdates(uint64_t next_se
                                                                 bool is_strict) const {
   std::vector<rocksdb::BatchResult> batches;
   auto latest_sequence = storage->LatestSeqNumber();
-  if (next_sequence == latest_sequence + 1) {
-    // return empty result if there is no new updates
-    return batches;
-  } else if (next_sequence > latest_sequence + 1) {
+  if (next_sequence > latest_sequence) {
```
git-hulk: What's the point of changing this?

songqing: If latest_sequence == UINT64_MAX, then latest_sequence + 1 wraps to 0. This makes any next_sequence == 0 look like "no new updates" and any next_sequence > 0 look like "out of range", which is incorrect logic. While RocksDB sequence numbers are unlikely to reach 2^64-1 in practice, the logic is fragile.

src/config/config.cc
```diff
@@ -239,7 +239,7 @@ Config::Config() {
       {"resp3-enabled", false, new YesNoField(&resp3_enabled, true)},
       {"repl-namespace-enabled", false, new YesNoField(&repl_namespace_enabled, false)},
       {"proto-max-bulk-len", false,
-       new IntWithUnitField<uint64_t>(&proto_max_bulk_len, std::to_string(512 * MiB), 1 * MiB, UINT64_MAX)},
+       new IntWithUnitField<uint64_t>(&proto_max_bulk_len, std::to_string(512 * MiB), 1 * MiB, UINT64_MAX - 2)},
```
git-hulk: This configuration controls the maximum bulk string length, which excludes the protocol separator. But I think we can refine this by having a proper upper limit.

RocksDB recommends using a value under 3 GiB; see [1]. Redis uses `LONG_MAX` as its upper limit, see [2]. So I suggest using the value like 3 or 4 GiB as Kvrocks upper limit. cc @PragmaTwice 

[1] https://github.com/facebook/rocksdb/issues/513 
[2] https://github.com/redis/redis/blob/a0bad9a0486fa586a2389cb38953e1137de5e8bd/src/config.c#L3272

src/server/server.cc
git-hulk: @songqing, I think the whole server won't work at all when the latest_sequence reaches UINT64_MAX.

songqing: ok, reverted

src/config/config.cc
songqing: ok, change to 4GB now, 4GB is indeed better

git-hulk (approved):

PR #3420, opened 2026-04-03
fix(stream): correct subkey prefix encoding in DestroyGroup

songqing:
DestroyGroup was constructing scan prefixes using only
    PutFixed64(group_name.size()) + group_name, which does not match the
    actual internal key format used by [internalKeyFromGroupName](https://github.com/apache/kvrocks/blob/da5e46307047b3e89753f0c9c02996430e758a5f/src/types/redis_stream.cc#L171),
    internalKeyFromConsumerName, and internalPelKeyFromGroupAndEntryId.

The real keys are prefixed with PutFixed64(UINT64_MAX) + PutFixed8(type)
    before the group name length and name. The wrong prefix meant the scan
    range could miss all group/consumer/PEL keys or match unrelated stream
    entry keys, causing data leaks or corruption on XGROUP DESTROY.

Fix by iterating over the three subkey types (GroupMetadata,
    ConsumerMetadata, PelEntry) and building the correct prefix for each,
    consistent with the internal key encoding helpers.

Besides, add two test cases verifying that DestroyGroup correctly cleans up all
    group-related subkeys (group metadata, consumer metadata, PEL entries):

- DestroyGroupCleansUpConsumersAndPelEntries: verifies full cleanup
      including that stream entries are preserved and the group can be
      re-created cleanly.
- DestroyGroupDoesNotAffectOtherGroups: verifies that destroying one
      group does not affect another group's consumers or PEL entries.

### DIFF
--- a/src/types/redis_stream.cc
+++ b/src/types/redis_stream.cc
@@ -783,7 +783,9 @@ rocksdb::Status Stream::DestroyGroup(engine::Context &ctx, const Slice &stream_n
   }
 
   *destroyed = true;
-  metadata.group_number -= 1;
+  if (metadata.group_number > 0) {
+    metadata.group_number -= 1;
+  }
   std::string metadata_bytes;
   metadata.Encode(&metadata_bytes);
   s = batch->Put(metadata_cf_handle_, ns_key, metadata_bytes);
--- a/src/types/redis_stream.cc
+++ b/src/types/redis_stream.cc
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "db_util.h"
+#include "string_util.h"
 #include "time_util.h"
 
 namespace redis {
@@ -751,6 +752,14 @@ rocksdb::Status Stream::DestroyGroup(engine::Context &ctx, const Slice &stream_n
     return rocksdb::Status::InvalidArgument(errXGroupSubcommandRequiresKeyExist);
   }
 
+  std::string group_key = internalKeyFromGroupName(ns_key, metadata, group_name);
+  std::string val;
+  s = storage_->Get(ctx, ctx.GetReadOptions(), stream_cf_handle_, group_key, &val);
+  if (s.IsNotFound()) {
+    return rocksdb::Status::OK();
+  }
+  if (!s.ok()) return s;
+
   auto batch = storage_->GetWriteBatchBase();
   WriteBatchLogData log_data(kRedisStream);
   s = batch->PutLogData(log_data.Encode());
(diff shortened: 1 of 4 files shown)


src/types/redis_stream.cc
```diff
@@ -756,28 +756,41 @@ rocksdb::Status Stream::DestroyGroup(engine::Context &ctx, const Slice &stream_n
...
-  std::string prefix_key = InternalKey(ns_key, sub_key_prefix, metadata.version, storage_->IsSlotIdEncoded()).Encode();
+  for (auto type : {StreamSubkeyType::StreamConsumerGroupMetadata, StreamSubkeyType::StreamConsumerMetadata,
+                    StreamSubkeyType::StreamPelEntry}) {
+    std::string sub_key_prefix;
+    PutFixed64(&sub_key_prefix, UINT64_MAX);
+    PutFixed8(&sub_key_prefix, static_cast<uint8_t>(type));
+    PutFixed64(&sub_key_prefix, group_name.size());
+    sub_key_prefix += group_name;
+
+    std::string next_version_prefix_key =
```
jihuayu: Isn't this upper bound too large?

jihuayu (commented):
I realized that we don't need to track delete_cnt. We can just use DeleteRange directly without iterating.
Please take a look at cmd_stream.cc L575. We only care whether `delete_cnt` is zero or not, rather than its specific value.

~~Can we avoid global scans for consumer metadata and PEL?~~

~~For instance, we could perform a prefix scan for each. By reusing the encoding prefixes defined in redis_stream.cc (line 221) and redis_stream.cc (line 266) respectively.~~

songqing:
> I realized that we don't need to track delete_cnt. We can just use DeleteRange directly without iterating. Please take a look at cmd_stream.cc L575. We only care whether `delete_cnt` is zero or not, rather than its specific value.
> 
> ~Can we avoid global scans for consumer metadata and PEL?~
> 
> ~For instance, we could perform a prefix scan for each. By reusing the encoding prefixes defined in redis_stream.cc (line 221) and redis_stream.cc (line 266) respectively.~

Good idea, fixed with DeleteRange, the code is much simple now

src/types/redis_stream.cc
songqing: Yes, the upper bound in the original implementation is too large, but I have a filter in the iterator loop to make sure only delete the needed data.
And now, I have changed to use the same version's end_key instead of the next version with filter. thanks for your suggestion.

src/types/redis_stream.cc
```diff
@@ -751,42 +752,42 @@ rocksdb::Status Stream::DestroyGroup(engine::Context &ctx, const Slice &stream_n
...
   }
 
-  if (*delete_cnt != 0) {
-    metadata.group_number -= 1;
-    std::string metadata_bytes;
-    metadata.Encode(&metadata_bytes);
-    s = batch->Put(metadata_cf_handle_, ns_key, metadata_bytes);
-    if (!s.ok()) return s;
-  }
+  *delete_cnt = 1;
```
jihuayu: So what is the significance of this parameter now?

songqing: Yes, this is a problem, according to [Redis XGROUP DESTROY](https://redis.io/docs/latest/commands/xgroup-destroy/), `Integer reply: the number of destroyed consumer groups, either 0 or 1.`, change to bool is better, fixed.

tests/cppunit/types/stream_test.cc
```diff
@@ -2518,10 +2518,130 @@ TEST_F(RedisStreamTest, StreamConsumerGroupCreateAndDestroy) {
...
+
+  // Consume entries in both groups to create PEL entries
+  redis::StreamRangeOptions range_options;
+  range_options.start = redis::StreamEntryID::Minimum();
+  range_options.end = redis::StreamEntryID::Maximum();
+  range_options.count = 10;
+  range_options.with_count = true;
+  range_options.exclude_start = true;
+  std::vector<redis::StreamEntry> entries;
+  std::string c1 = "c1", c2 = "c2";
```
LindaSummer: Could we split this into two declaration?

tests/cppunit/types/stream_test.cc
```diff
@@ -2518,10 +2518,130 @@ TEST_F(RedisStreamTest, StreamConsumerGroupCreateAndDestroy) {
...
   auto s = stream_->CreateGroup(*ctx_, stream_name, create_options, group_name);
   EXPECT_TRUE(s.ok());
-  uint64_t delete_cnt = 0;
-  s = stream_->DestroyGroup(*ctx_, stream_name, group_name, &delete_cnt);
-  EXPECT_TRUE(delete_cnt != 0);
-  delete_cnt = 0;
-  s = stream_->DestroyGroup(*ctx_, stream_name, group_name, &delete_cnt);
-  EXPECT_TRUE(delete_cnt == 0);
+  bool destroyed = false;
+  s = stream_->DestroyGroup(*ctx_, stream_name, group_name, &destroyed);
```
LindaSummer, line 2528: Could we add `GetGroupInfo` before and after destroy?

LindaSummer (commented):
Hi @songqing ,

Left some comment on test cases.

Generally LGTM.

Thanks very much for your effort!😊

tests/cppunit/types/stream_test.cc
songqing: done

tests/cppunit/types/stream_test.cc
songqing: done

jihuayu (approved):

jihuayu:
Sorry, I looked at the wrong PR. I didn't mean to close it.

LindaSummer (approved):

jihuayu:
@songqing Thank you for your hard work.

PR #3421, opened 2026-04-03
fix(zset): initialize *inter_cnt before early return in InterCard

songqing:
When any input key is empty (mscores.empty()), InterCard returned OK
    without ever assigning *inter_cnt, leaving callers with an
    uninitialized value. Redis ZINTERCARD should return 0 in this case.

Fix by initializing *inter_cnt = 0 at the function entry.

Although the count at the caller layer is initialized to 0, which seems to be fortuitously safe, other potential callers (such as internal calls, test code, and future refactoring) may not necessarily initialize it beforehand.


jihuayu (approved):
Speaking strictly about this PR, I have no objections. Adding defensive code can certainly improve the robustness of the code.

However, I do have one question: does this specific coding style align with our established conventions?

songqing:
> Speaking strictly about this PR, I have no objections. Adding defensive code can certainly improve the robustness of the code.
> 
> However, I do have one question: does this specific coding style align with our established conventions?

Thanks for your comment, I push this PR only want to fix the potential bug as I mentioned in the description, as coding style, it's hard to follow for everyone.

jihuayu:
Looks like there’s an issue with the CI. I’ll look into it.

PragmaTwice (approved):

Issue #3407, opened 2026-03-30
kvrocks lacks proper flow control in network and request layers, potentially causing severe long-tail latency


Tangruilin:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

 I found two issues in kvrocks:

 Issue 1: The network layer has no bandwidth limit for individual requests (rate_limit_group_ is never initialized), which may cause network bandwidth contention between Client A and
 Client B.
```bash
$grep -r rate_limit_group_
src/server/worker.cc:  if (rate_limit_group_) {
src/server/worker.cc:    bufferevent_rate_limit_group_free(rate_limit_group_);
src/server/worker.cc:  if (rate_limit_group_cfg_) {
src/server/worker.cc:    ev_token_bucket_cfg_free(rate_limit_group_cfg_);
src/server/worker.cc:  if (rate_limit_group_) {
src/server/worker.cc:    bufferevent_add_to_rate_limit_group(bev, rate_limit_group_);
src/server/worker.cc:  if (rate_limit_group_) {
src/server/worker.cc:    bufferevent_add_to_rate_limit_group(bev, rate_limit_group_);
src/server/worker.cc:  if (rate_limit_group_) {
src/server/worker.cc:  if (rate_limit_group_) {
src/server/worker.cc:    if (rate_limit_group_ != nullptr) {
src/server/worker.h:  struct bufferevent_rate_limit_group *rate_limit_group_ = nullptr;
src/server/worker.h:  struct ev_token_bucket_cfg *rate_limit_group_cfg_ = nullptr;
```
rate_limit_group_ always be nullptr.

Issue 2: The request scheduling in kvrocks is command-based, without any differentiation based on request type or data size.
example:
```bash
   // redis_connection.cc - ExecuteCommands

   while (!to_process_cmds->empty()) {
       CommandTokens cmd_tokens = std::move(to_process_cmds->front ());
       to_process_cmds->pop_front();

       // execute
       s = ExecuteCommand(ctx, cmd_name, cmd_tokens, current_cmd.get(), &reply);

       // response
       if (!reply.empty()) Reply(reply);
   }
```
In this case, when CPU-intensive requests arrive and preempt CPU resources, small requests are treated unfairly.
This leads to severe long-tail latency for requests.

### Solution

For Issue 1: I believe adding a default configuration for rate_limit_group_ and supporting customization in the config file would be sufficient to solve this problem.

[...]

PR #3424, opened 2026-04-06
feat(worker): initialize rate_limit_group_ for network bandwidth limi…

Tangruilin:
Fixes part 1 of #3407.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -191,6 +191,7 @@ Config::Config() {
       {"log-level", false, new EnumField<spdlog::level::level_enum>(&log_level, log_levels, spdlog::level::info)},
       {"pidfile", true, new StringField(&pidfile, kDefaultPidfile)},
       {"max-io-mb", false, new IntField(&max_io_mb, 0, 0, INT_MAX)},
+      {"net-rate-limit-mb", true, new IntField(&net_rate_limit_mb, 0, 0, INT_MAX)},
       {"enable-blob-cache", true, new YesNoField(&enable_blob_cache, false)},
       {"max-bitmap-to-string-mb", false, new IntField(&max_bitmap_to_string_mb, 16, 0, INT_MAX)},
       {"max-db-size", false, new IntField(&max_db_size, 0, 0, INT_MAX)},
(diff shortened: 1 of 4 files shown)


jihuayu:
I’m not quite sure if this change makes sense for us. @git-hulk @PragmaTwice 

I feel like this feature is testable; if so, we’re going to need some test cases. @Tangruilin

git-hulk:
> I’m not quite sure if this change makes sense for us. @git-hulk @PragmaTwice
> 
> I feel like this feature is testable; if so, we’re going to need some test cases. @Tangruilin

I feel it's strange. I'm not quite understanding why we need to control the bandwidth since it might heavily affect the user performance.

Tangruilin:
> I feel it's strange. I'm not quite understanding why we need to control the bandwidth since it might heavily affect the user performance.

This parameter already existed in kvrocks; it just hadn't been added to the configuration file until now.

The purpose of this parameter in libevent is to prevent multiple clients from competing for network traffic. It simply provides users with an option, while the default behavior remains consistent with kvrocks' original behavior.

refer to: http://www.wangafu.net/~nickm/libevent-book/Ref6a_advanced_bufferevents.html#_rate_limiting

Tangruilin:
> > I’m not quite sure if this change makes sense for us. @git-hulk @PragmaTwice
> > I feel like this feature is testable; if so, we’re going to need some test cases. @Tangruilin
> 
> I feel it's strange. I'm not quite understanding why we need to control the bandwidth since it might heavily affect the user performance.

This parameter is used for tuning when the bottleneck is the network interface card of the machine where kvrocks is running, rather than the client side. When kvrocks' network bandwidth is saturated, multiple clients may compete for bandwidth with each other. In this scenario, if users want to prevent high-traffic clients from competing with low-traffic clients, they can achieve this by setting this flag.

Tangruilin:
It looks like Redis doesn't support traffic control for clients at the server layer either. If kvrocks doesn't need this either, let's close this PR

jihuayu:
Let's close this PR. It feels out of place for a third-party middleware to be controlling network traffic. If this is a requirement, I believe it should be handled at the gateway level, or we should have the Ops team scale up the resources.

Issue #3409, opened 2026-03-30
Add support for dictionary compression settings


pleaseshutup:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

It seems the default settings are max_dict_bytes = 0 and zstd_max_train_bytes = 0 
If we can enable these settings in kvrocks.conf we could get a pretty big zstd benefit for storing a large amount of fairly normalized json strings

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

nagisa-kunhah:
hello, can assign it to me？

git-hulk:
@nagisa-kunhah Assigned, thank you.

PR #3425, opened 2026-04-07
feat: Add support for dictionary compression settings

nagisa-kunhah:
issue: #3409


nagisa-kunhah:
Hi @pleaseshutup , sorry to bother you. This PR is ready for review now.
Could you please take a look when convenient? Thanks!

pleaseshutup:
@nagisa-kunhah I reviewed your PR and I think this will work great. 
I did not pull your code to test on a new build, however, I did the same changes to src/storage/storage.cc that your pr does with the config and verified that changing those settings does in fact implement the dictionary compression

My data set was 3 billion json documents of various sizes but common schema compression went from 1.4TB on disk to 1.1TB on disk going from lz4 to zstd and from 1.1TB on disk to 0.8TB on disk when enabling dictionary compression after forcing compaction on the entire data set.

tests/cppunit/config_test.cc
```diff
@@ -149,9 +151,12 @@ TEST(Config, GetRenameCommand) {
...
   std::ofstream output_file(path, std::ios::out);
-  output_file << "rename-command KEYS KEYS_NEW" << "\n";
-  output_file << "rename-command GET GET_NEW" << "\n";
-  output_file << "rename-command SET SET_NEW" << "\n";
+  output_file << "rename-command KEYS KEYS_NEW"
+              << "\n";
+  output_file << "rename-command GET GET_NEW"
+              << "\n";
+  output_file << "rename-command SET SET_NEW"
+              << "\n";
```
jihuayu: Why change it?

tests/cppunit/config_test.cc
```diff
@@ -171,9 +176,12 @@ TEST(Config, Rewrite) {
...
```
jihuayu: Why change it?

src/config/config.cc
```diff
@@ -252,8 +252,10 @@ Config::Config() {
       /* rocksdb options */
       {"rocksdb.compression", false,
        new EnumField<rocksdb::CompressionType>(&rocks_db.compression, compression_types,
-                                               rocksdb::CompressionType::kNoCompression)},
```
jihuayu, line 255: You've changed the default value. Is this change necessary? What will be the impact on the users?

tests/cppunit/config_test.cc
```diff
@@ -189,6 +197,32 @@ TEST(Config, Rewrite) {
   unlink(path);
 }
 
+TEST(Config, LoadRocksDBDictionaryCompressionOptions) {
```
jihuayu, line 194: The test cases lack a closed-loop logic. Currently, they don't verify whether the configuration has actually taken effect or if it's being correctly passed down to RocksDB.

jihuayu (commented):
Great job. Only some small problems.

src/config/config.cc
nagisa-kunhah: ok, restore it

tests/cppunit/config_test.cc
nagisa-kunhah: ok, restore it

tests/cppunit/config_test.cc
nagisa-kunhah: restore it

tests/cppunit/config_test.cc
nagisa-kunhah: Got it. I added another test, RocksDBDictionaryCompressionOptions, to cover this case and verify the configuration is correctly propagated to RocksDB.

src/config/config.cc
```diff
@@ -254,6 +254,8 @@ Config::Config() {
        new EnumField<rocksdb::CompressionType>(&rocks_db.compression, compression_types,
                                                rocksdb::CompressionType::kNoCompression)},
       {"rocksdb.compression_level", true, new IntField(&rocks_db.compression_level, 32767, INT_MIN, INT_MAX)},
+      {"rocksdb.max_dict_bytes", true, new UInt32Field(&rocks_db.max_dict_bytes, 0, 0, UINT32_MAX)},
+      {"rocksdb.zstd_max_train_bytes", true, new UInt32Field(&rocks_db.zstd_max_train_bytes, 0, 0, UINT32_MAX)},
```
PragmaTwice: maybe we can rename them to `rocksdb.compression_max_dict_bytes` and `rocksdb.compression_zstd_max_train_bytes`?

nagisa-kunhah: ok, fix it

jihuayu (approved):

PragmaTwice (approved):

PR #3428, opened 2026-04-08
fix(set): propagate storage errors in Set instead of treating as NotFound

songqing:
Previously, when `storage_->Get()` returned a non-NotFound error (e.g., I/O error or corruption) during SADD, the code fell through to the `batch->Put` path, incorrectly treating the error as "member not found" and adding the member. This could silently inflate `metadata.size` and insert duplicate members on storage failures.

Add an explicit `!s.IsNotFound()` check to propagate real errors, which is consistent with how `Set::Remove` already handles the same pattern.


jihuayu (commented):
Great! 
It seems SREM has the same issue. Could you fix it as well and add the corresponding test cases?

songqing:
> Great! It seems SREM has the same issue. Could you fix it as well and add the corresponding test cases?
done

PragmaTwice (approved):

jihuayu (approved):

PR #3441, opened 2026-04-10
fix(hyperloglog): use namespace-prefixed key in PFMERGE GetMetadata

songqing:
Merge called GetMetadata with the raw user key instead of the namespace-prefixed key, causing the destination's existing metadata to never be found and potentially orphaning old sub-keys.

### DIFF
--- a/src/types/redis_hyperloglog.cc
+++ b/src/types/redis_hyperloglog.cc
@@ -240,7 +240,7 @@ rocksdb::Status HyperLogLog::Merge(engine::Context &ctx, const Slice &dest_user_
   std::vector<std::string> registers;
   HyperLogLogMetadata metadata;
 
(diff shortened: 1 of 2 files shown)


jihuayu (dismissed):

jihuayu:
It looks like this won't leave behind any stale data that needs manual cleanup. We can just update the code directly.

songqing:
> It looks like this won't leave behind any stale data that needs manual cleanup. We can just update the code directly.

I just add a new test to verify the fix, without the fix, the new test will fail and add comment on the test

jihuayu (approved):

jihuayu:
Since there haven't been any objections to this PR for a while, I'm going to merge it.

PragmaTwice (approved):

PR #3442, opened 2026-04-10
fix(stream): allow group and consumer names starting with digits

songqing:
Redis does not restrict group or consumer names from starting with digits. The restriction was kvrocks-specific and broke compatibility.


jihuayu:
Hi @Yangsx-1. I noticed that you authored the original code. Was there a specific rationale behind this implementation?

jihuayu (approved):
LGTM. There's no point in limiting this to non-numbers.

jihuayu:
Since there haven't been any objections to this PR for a while, I'm going to merge it.

LindaSummer (approved):

Issue #3267, opened 2025-11-20
sentinel monitor kvrocks


dbabao:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

My Kvrocks instance (kvrockstest-001) uses a non-cluster mode master-slave architecture.  kvrocks version 2.14.0。sentinel version 5.0.14。  

During testing with Sentinel supervision, I discovered that after executing the monitor add command on all Sentinel nodes, the Sentinel info display showed abnormalities - it indicated only 1 Sentinel instance and failed to recognize the slave nodes of this Kvrocks instance.   The ports between Sentinel nodes and between Sentinel and Kvrocks nodes are mutually accessible.

The issue can be resolved by either executing sentinel reset kvrockstest-001 or restarting the Sentinel service。

Commands to add supervision:

```
sentinel monitor kvrockstest-001 192.168.56.103 6666 2
sentinel set kvrockstest-001 down-after-milliseconds 20000
sentinel set kvrockstest-001 failover-timeout 5000
sentinel set kvrockstest-001 parallel-syncs 1
sentinel set kvrockstest-001 auth-pass 123456
```

**sentinel info信息：**
127.0.0.1:27379> info Sentinel
# Sentinel
sentinel_masters:1
sentinel_tilt:0
sentinel_running_scripts:0
sentinel_scripts_queue_length:0
sentinel_simulate_failure_flags:0
master0:name=kvrockstest-001,status=sdown,address=192.168.56.103:6666,slaves=0,sentinels=1

**sentinel service log msg：**
36204:X 20 Nov 2025 17:35:57.673 # +monitor master kvrockstest-001 192.168.56.103 6666 quorum 2
36204:X 20 Nov 2025 17:35:57.680 # +set master kvrockstest-001 192.168.56.103 6666 down-after-milliseconds 20000
36204:X 20 Nov 2025 17:35:57.689 # +set master kvrockstest-001 192.168.56.103 6666 failover-timeout 5000
36204:X 20 Nov 2025 17:35:57.704 # +set master kvrockstest-001 192.168.56.103 6666 parallel-syncs 1
36204:X 20 Nov 2025 17:35:58.506 # +set master kvrockstest-001 192.168.56.103 6666 auth-pass 123456
36204:X 20 Nov 2025 17:36:17.688 # +sdown master kvrockstest-001 192.168.56.103 6666
[...]

PR #3443, opened 2026-04-10
fix(server): preserve existing connections after requirepass changes

aviralgarg05:
**Description**

This fixes a Redis Sentinel compatibility issue reported in #3267.

When Kvrocks accepted a connection while `requirepass` was still empty, it did not immediately mark that connection as authenticated. Instead, it deferred assigning the default namespace/admin context until the connection sent its first command.

That differs from Redis behavior.

In practice, this created a race with older Sentinel versions. Sentinel could establish a connection to the master before `auth-pass` was configured, but only send its first probe afterward. Redis continues to allow that already-open connection to operate, while Kvrocks responded with `NOAUTH`. As a result, Sentinel could temporarily treat the master as down and fail to discover replicas until a reset or restart forced a fresh connection cycle.

This change fixes the behavior at the source:

- Passwordless connections are now initialized as authenticated when they are accepted.
- Existing connections opened before `requirepass` is later set continue to behave like Redis.
- New connections created after `requirepass` is set still require authentication as expected.

**Files changed**

- [src/server/redis_connection.cc](/Users/aviralgarg/Everything/kvrocks/src/server/redis_connection.cc)
- [tests/gocase/unit/auth/auth_test.go](/Users/aviralgarg/Everything/kvrocks/tests/gocase/unit/auth/auth_test.go)

**Why this is safe**

The change is limited to the connection initialization path and only affects connections created while the server is running without a password configured.

It does not relax authentication for:
- servers that already start with `requirepass` set
- connections created after `requirepass` is enabled

So the fix restores Redis-compatible semantics without changing the normal protected path.

**Test coverage**

Added a regression test that verifies:

1. a connection opened before `CONFIG SET requirepass ...` remains usable
2. a new connection opened after `requirepass` is set receives `NOAUTH`

**Validation**

Ran:

- `./x.py build -j 8`
- `./x.py format`
- `./x.py check format`
- `go test -timeout=1800s -run 'TestNoAuth|TestAuth' ./unit/auth -binPath=/Users/aviralgarg/Everything/kvrocks/build/kvrocks -cliPath=redis-cli -workspace=/Users/aviralgarg/Everything/kvrocks/tests/gocase/workspace`

The auth test suite was run 3 times and passed each time.

I also manually verified the repro condition:
- before the fix, an idle connection created before `requirepass` was enabled received `-NOAUTH`
- after the fix, the same connection correctly receives `+PONG`

**Notes**

Two repo-wide checks are currently blocked by the local environment rather than this change:

- `./x.py check tidy`
  - after wiring `run-clang-tidy` from local LLVM, it reports existing unrelated tidy failures outside this patch
- `./x.py check golangci-lint`
  - local `golangci-lint` is `2.8.0`, while the repo requires `2.9.0+`

### DIFF
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -52,10 +52,6 @@ Connection::Connection(bufferevent *bev, Worker *owner)
   int64_t now = util::GetTimeStamp();
   create_time_ = now;
   last_interaction_ = now;
-  if (srv_->GetConfig()->requirepass.empty()) {
-    BecomeAdmin();
-    SetNamespace(kDefaultNamespace);
-  }
 }
 
 Connection::~Connection() {
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -52,6 +52,10 @@ Connection::Connection(bufferevent *bev, Worker *owner)
   int64_t now = util::GetTimeStamp();
   create_time_ = now;
   last_interaction_ = now;
+  if (srv_->GetConfig()->requirepass.empty()) {
+    BecomeAdmin();
+    SetNamespace(kDefaultNamespace);
+  }
 }
 
 Connection::~Connection() {
--- a/tests/gocase/integration/replication/replication_test.go
+++ b/tests/gocase/integration/replication/replication_test.go
@@ -329,7 +329,7 @@ func TestReplicationWithLimitSpeed(t *testing.T) {
 		require.Eventually(t, func() bool {
 			return slave.LogFileMatches(t, ".*skip count: 1.*")
 		}, 50*time.Second, 1000*time.Millisecond)
-		util.WaitForSync(t, slaveClient)
+		util.WaitForOffsetSync(t, masterClient, slaveClient, 50*time.Second)
 		require.Equal(t, "b", slaveClient.Get(ctx, "a").Val())
 	})
 }
--- a/src/server/redis_connection.h
+++ b/src/server/redis_connection.h
@@ -154,6 +154,7 @@ class Connection : public EvbufCallbackBase<Connection> {
   bool IsAdmin() const { return is_admin_; }
   void BecomeAdmin() { is_admin_ = true; }
   void BecomeUser() { is_admin_ = false; }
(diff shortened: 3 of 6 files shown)


src/server/redis_connection.cc
```diff
@@ -52,6 +52,10 @@ Connection::Connection(bufferevent *bev, Worker *owner)
   int64_t now = util::GetTimeStamp();
   create_time_ = now;
   last_interaction_ = now;
+  if (srv_->GetConfig()->requirepass.empty()) {
```
git-hulk: We cannot make any changes without studying the root cause and understanding the codebase.

git-hulk (changes requested):

jihuayu:
It looks like this crash is 100% reproducible. Could you take a look? @aviralgarg05

tests/gocase/integration/replication/replication_test.go
```diff
@@ -329,7 +329,7 @@ func TestReplicationWithLimitSpeed(t *testing.T) {
 		require.Eventually(t, func() bool {
 			return slave.LogFileMatches(t, ".*skip count: 1.*")
 		}, 50*time.Second, 1000*time.Millisecond)
-		util.WaitForSync(t, slaveClient)
+		util.WaitForOffsetSync(t, masterClient, slaveClient, 50*time.Second)
```
jihuayu, line 332: Why we need to change it?

jihuayu (commented):
LGTM. I do have one small question, though.

tests/gocase/integration/replication/replication_test.go
aviralgarg05: `WaitForSync` only checks `master_link_status == up`, which is weaker than what this test needs.

In this case we restart the master after a partial SST transfer, so the replica link can come back before the resumed transfer has fully caught up. That was making the test flaky on slower matrix variants, especially `-DENABLE_LUAJIT=OFF`.

`WaitForOffsetSync` waits for the master and replica offsets to match, which is the actual condition we care about before checking the replicated key.

PR #3448, opened 2026-04-13
feat(ci): add caching for CMake FetchContent in CI workflow

jihuayu:
Due to the current instability of GitHub services, dependency download failures occur frequently during CI runs. I add FetchContent caching to prevent these network issues from failing the CI.

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -32,6 +32,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.number || github.run_id }}
   cancel-in-progress: true
 
+env:
+  KVROCKS_DEPS_CACHE_DIR: build-deps
+
 jobs:
   precondition:
     name: Precondition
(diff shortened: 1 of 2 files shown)


PragmaTwice (changes requested):
You can use `DEPS_FETCH_DIR` instead of caching these `_deps` directories (which may bring some chaos).

jihuayu:
@PragmaTwice Thanks for the reminder. I’ve made the changes. 
Do you think it’s better to cache each job individually (now), or do a full cache at the beginning of the action?

PragmaTwice:
> env.CMAKE_JEMALLOC_VARIANT, env.CMAKE_LUA_VARIANT,

I don't think we need this. They are useless and introduce complexity.

> Do you think it’s better to cache each job individually (now), or do a full cache at the beginning of the action?

I think we'd better to choose the simplest way.

jihuayu:
> env.CMAKE_JEMALLOC_VARIANT, env.CMAKE_LUA_VARIANT,
> I don't think we need this. They are useless and introduce complexity.

This variable is just for readability; otherwise, we’d be stuck facing a very long cache key.
```${{ format('cmake-fetch-archives-v1-{0}-{1}-{2}-{3}', matrix.os, startsWith(matrix.os, 'macos') && 'without-jemalloc' || matrix.without_jemalloc && 'without-jemalloc' || 'with-jemalloc' ,  matrix.without_luajit && 'without-luajit' || 'with-luajit' , hashFiles('CMakeLists.txt', 'cmake/**/*')) }}```


As I understand it, the Lua and jemalloc options must be cached separately because they pull different source codes. Is that correct?

**Comment by PragmaTwice** [comment]
> As I understand it, the Lua and jemalloc options must be cached separately because they pull different source codes. Is that correct?

Nope. The dep dir can be shared. The filename is different for different deps.

File: .github/workflows/kvrocks.yaml [hunk 1]
```diff
@@ -65,13 +68,40 @@ jobs:
...
+        with:
+          path: ${{ env.KVROCKS_DEPS_CACHE_DIR }}
+          key: kvrocks-full-deps-${{ hashFiles('x.py', 'CMakeLists.txt', 'cmake/**/*.cmake') }}
+          enableCrossOsArchive: true
+      - name: Warm Kvrocks dependency cache
+        if: ${{ steps.cache-kvrocks-deps.outputs.cache-hit != 'true' }}
+        run: |
+          ./x.py build .github/ci-cache-default --skip-build --dep-dir ${{ env.KVROCKS_DEPS_CACHE_DIR }}
+          ./x.py build .github/ci-cache-lua --skip-build --dep-dir ${{ env.KVROCKS_DEPS_CACHE_DIR }} -D ENABLE_LUAJIT=OFF
+          rm -rf .github/ci-cache-default .github/ci-cache-lua
```
PragmaTwice: why we need this job?

PragmaTwice: we don't need a job to "warm" the cache.. you can try `--dep-dir` on your local to know how it works.

jihuayu: GitHub Actions caches are immutable. Once a cache is created for a specific key, it cannot be updated or overwritten.

So, we need to create all files intended for caching in a single step.

PragmaTwice: ahh got it.

please put it in `check-and-lint`. also you can use `./x.py fetch-deps`.

jihuayu: All is right. Thank you!

PragmaTwice (approved):

Issue #3447, opened 2026-04-12
TDIGEST.MERGE Parse error


Tangruilin:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Version

unstable

### Minimal reproduce step

command `TDIGEST.MERGE dest 1 src COMPRESSION 100` has  ouput `ERR wrong keyword`

### What did you expect to see?

merge success

### What did you see instead?

ERR wrong keyword

<img width="631" height="73" alt="Image" src="https://github.com/user-attachments/assets/6c6d65af-6c33-4bdc-880b-cca30fe00cec" />

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Tangruilin:
I found this issue while working on PR #3412  
To keep that PR focused, I plan to address this in a separate PR.

Tangruilin:
redis output
```bash
127.0.0.1:6380> TDIGEST.CREATE src
(error) ERR T-Digest: key already exists
127.0.0.1:6380> TDIGEST.ADD src 1 2 3 4 5
OK
127.0.0.1:6380> TDIGEST.CREATE dest
OK
127.0.0.1:6380> TDIGEST.MERGE dest 1 src COMPRESSION 100
OK
127.0.0.1:6380> 
```

LindaSummer:
Hi @Tangruilin ,
Thanks very much for your effort!

PR #3449, opened 2026-04-13
fix(tdigest): correct TDIGEST.MERGE parser for COMPRESSION parameter.

Tangruilin:
…(#3447)

The parser incorrectly used two independent `if` statements to check COMPRESSION and OVERRIDE parameters. When COMPRESSION was successfully parsed, the code continued to the second `if` which would fail since the parser had reached the end, causing "ERR wrong keyword" error.

Fixed by changing the second `if` to `else if` to form a proper if-else-if-else chain.

### DIFF
--- a/src/commands/cmd_tdigest.cc
+++ b/src/commands/cmd_tdigest.cc
@@ -466,9 +466,7 @@ class CommandTDigestMerge : public Commander {
         } else {
           options_.compression = *compression;
         }
-      }
-
-      if (parser.EatEqICase(kOverrideArg)) {
+      } else if (parser.EatEqICase(kOverrideArg)) {
         if (options_.override_flag) {  // override already set
           return {Status::RedisParseErr, errWrongNumOfArguments};
         }
--- a/tests/gocase/unit/type/tdigest/tdigest_test.go
+++ b/tests/gocase/unit/type/tdigest/tdigest_test.go
@@ -525,6 +525,53 @@ func tdigestTests(t *testing.T, configs util.KvrocksServerConfigs) {
 		validation(newDestKey2)
 	})
 
+	// https://github.com/apache/kvrocks/issues/3447
+	t.Run("tdigest.merge with COMPRESSION parameter (issue #3447)", func(t *testing.T) {
+		keyPrefix := "tdigest_merge_compression_"
+
+		srcKey := keyPrefix + "src"
+		destKey := keyPrefix + "dest"
+
+		require.NoError(t, rdb.Do(ctx, "TDIGEST.CREATE", srcKey, "compression", "100").Err())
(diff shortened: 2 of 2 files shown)


Tangruilin:
fix before:
image

after:
image

.gitignore
```diff
@@ -52,3 +52,6 @@
 /build/
 /cmake-build-*/
 /build-*/
+
+# go cache
+/pkg
```
jihuayu: Why change it?

Tangruilin: The /pkg directory is where Go caches dependencies when running tests. If not ignored, you'd have to be careful every time you submit a PR to avoid accidentally committing it.

src/commands/cmd_tdigest.cc
```diff
@@ -466,9 +466,7 @@ class CommandTDigestMerge : public Commander {
         } else {
           options_.compression = *compression;
         }
-      }
-
-      if (parser.EatEqICase(kOverrideArg)) {
+      } else if (parser.EatEqICase(kOverrideArg)) {
```
LindaSummer, line 469: Refer to [Redis official doc tdiget.merge](https://redis.io/docs/latest/commands/tdigest.merge/), the `override` is optional after `compression`.
> TDIGEST.MERGE destination-key numkeys source-key [source-key ...] [COMPRESSION compression] [OVERRIDE]

This change would make  `tdigest.merge dest_key 1 source_key compression 101 override` discard the `override` option and lead to unexpected behavior.

.gitignore
LindaSummer: We should keep SRP (Single Responsibility Principle) for each patch.
It is very important when we try to publish release and cherry-pick patches to specific version.
If we need this gitignore change, please create a new PR for it.

LindaSummer (commented):
Hi @Tangruilin ,

It seems this change involved some unexpected behavior.
It would be better to add related test cases for this scenario.

.gitignore
Tangruilin: okey. i will submit it in another patch

src/commands/cmd_tdigest.cc
Tangruilin: Actually, this won't cause any issue. The implementation in kvrocks differs from redisBloom here - kvrocks ensures that the trailing override won't be ignored through `while(parser.Good())`. I
  had already considered this scenario before making the code changes.

  PS: I've added corresponding test cases for this.

jihuayu (approved):

src/commands/cmd_tdigest.cc
LindaSummer: Sorry for my misunderstanding of this part. You are right, and the refactor should work well. 👍

LindaSummer (approved):
Hi @Tangruilin ,

LGTM🚀

Thanks very much for your great work!

Issue #3215, opened 2025-10-05
TimeSeries: Support `IGNORE` option when creating timesereies


yezhizi:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

When creating a time series, the `[IGNORE ignoreMaxTimedestampDiff ignoreMaxValDiff]` options can be specified. During samples writing through `TS.ADD/TS.MADD`, samples meeting certain condition will be ignored.

Redis reference:
https://redis.io/docs/latest/commands/ts.create/

### Solution

- Add two fields(ignoreMaxTimedestampDiff ignoreMaxValDiff) to timeseries metadata 
- When adding samples, determine whether the "ignore" condition is met.
- Add go integrated test

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

marcomit:
I can work on this!

PR #3450, opened 2026-04-14
feat(ts): Support IGNORE option when creating TimeSeries

lauk20:
Resolves issue #3215.

AI Usage: AI was used to write the initial draft and tests. Manually reviewed the code logic to verify and fix issues.

### DIFF
--- a/src/storage/redis_metadata.cc
+++ b/src/storage/redis_metadata.cc
@@ -631,7 +631,7 @@ rocksdb::Status TimeSeriesMetadata::Decode(Slice *input) {
   if (auto s = Metadata::Decode(input); !s.ok()) {
     return s;
   }
-  if (input->size() < sizeof(uint64_t) * 3 + sizeof(uint8_t) * 2 + sizeof(uint32_t) + sizeof(double)) {
+  if (input->size() < sizeof(uint64_t) * 3 + sizeof(uint8_t) * 2 + sizeof(uint32_t)) {
     return rocksdb::Status::InvalidArgument(kErrMetadataTooShort);
   }
 
@@ -643,8 +643,12 @@ rocksdb::Status TimeSeriesMetadata::Decode(Slice *input) {
   GetSizedString(input, &source_key_slice);
   source_key = source_key_slice.ToString();
   GetFixed64(input, &last_timestamp);
-  GetFixed64(input, &ignore_max_time_diff);
-  GetDouble(input, &ignore_max_val_diff);
+  ignore_max_time_diff = 0;
+  ignore_max_val_diff = 0.0;
+  if (input->size() >= sizeof(uint64_t) + sizeof(double)) {
+    GetFixed64(input, &ignore_max_time_diff);
+    GetDouble(input, &ignore_max_val_diff);
+  }
 
   return rocksdb::Status::OK();
 }
--- a/src/storage/redis_metadata.cc
+++ b/src/storage/redis_metadata.cc
@@ -548,13 +548,15 @@ void TimeSeriesMetadata::Encode(std::string *dst) const {
   PutFixed8(dst, static_cast<uint8_t>(duplicate_policy));
   PutSizedString(dst, source_key);
   PutFixed64(dst, last_timestamp);
+  PutFixed64(dst, ignore_max_time_diff);
+  PutDouble(dst, ignore_max_val_diff);
 }
 
 rocksdb::Status TimeSeriesMetadata::Decode(Slice *input) {
   if (auto s = Metadata::Decode(input); !s.ok()) {
     return s;
   }
-  if (input->size() < sizeof(uint64_t) * 2 + sizeof(uint8_t) * 2 + sizeof(uint32_t)) {
+  if (input->size() < sizeof(uint64_t) * 3 + sizeof(uint8_t) * 2 + sizeof(uint32_t) + sizeof(double)) {
     return rocksdb::Status::InvalidArgument(kErrMetadataTooShort);
   }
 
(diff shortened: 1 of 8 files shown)


src/storage/redis_metadata.cc
```diff
@@ -566,6 +568,8 @@ rocksdb::Status TimeSeriesMetadata::Decode(Slice *input) {
   GetSizedString(input, &source_key_slice);
   source_key = source_key_slice.ToString();
   GetFixed64(input, &last_timestamp);
+  GetFixed64(input, &ignore_max_time_diff);
```
jihuayu, line 646: This will cause compatibility issues with old data, right?

jihuayu (changes requested):
It looks like this PR will cause backward incompatibility with old data.
We should probably draft a proposal for the storage format changes.

tests/gocase/unit/type/timeseries/timeseries_test.go
```diff
@@ -202,6 +202,24 @@ func testTimeSeries(t *testing.T, configs util.KvrocksServerConfigs) {
 		require.ErrorContains(t, rdb.Do(ctx, "ts.add", key, "1000", "13.4").Err(), "update is not supported when DUPLICATE_POLICY is set to BLOCK mode")
 	})
 
+	t.Run("TS.ADD Ignore Option", func(t *testing.T) {
+		ignoreKey := "test_add_ignore_key"
+		require.NoError(t, rdb.Del(ctx, ignoreKey).Err())
+		require.NoError(t, rdb.Do(ctx, "ts.create", ignoreKey, "duplicate_policy", "last", "ignore", "5", "2").Err())
+
+		require.Equal(t, int64(1000), rdb.Do(ctx, "ts.add", ignoreKey, "1000", "10").Val())
+		require.Equal(t, int64(1003), rdb.Do(ctx, "ts.add", ignoreKey, "1003", "11").Val())
```
jihuayu: It will return `1000` in Redis Stack

jihuayu (changes requested):

tests/gocase/unit/type/timeseries/timeseries_test.go
lauk20: Thanks for catching this. I have updated the implementation to fix this issue.

src/storage/redis_metadata.cc
lauk20: Yes, I believe it will cause compatibility issues with old data.

One other option we have is we can add backward compatibility like so:
```cpp
ignore_max_time_diff = 0;
ignore_max_val_diff = 0.0;
if (input->size() >= sizeof(uint64_t) + sizeof(double)) {
  GetFixed64(input, &ignore_max_time_diff);
  GetDouble(input, &ignore_max_val_diff);
}
```

This will only Decode the values if they exist in the metadata.

How do you think we should proceed?

jihuayu:
Hi @lauk20. If you want to move forward with this PR, please fix the format and lint CI errors. Thanks

lauk20:
> Hi @lauk20. If you want to move forward with this PR, please fix the format and lint CI errors. Thanks

@jihuayu Thanks for the update. I have committed a fix for the lint.

src/storage/redis_metadata.cc
jihuayu: @lauk20  Backward compatibility is definitely a must, but we should consider if forward compatibility is also needed. Does your current proposal support it?

@git-hulk Do we need to support both backward and forward compatibility? Is there a possibility that users might roll back their versions?

lauk20: @jihuayu I have committed the backward compatibility fixes. 

I do not believe that forward compatibility is supported. It may be difficult given that if new metadata is written on a newer version and the user rolls back to an older version, the metadata for `ignore_max_time_diff` and `ignore_max_val_diff` will still already be written on disk in RocksDB. I am unsure if there is any parts of the codebase that depends on the metadata format to be formatted strictly based on the existing metadata format.

src/types/redis_timeseries.h
```diff
@@ -257,8 +259,7 @@ enum class TSAlterMode : uint8_t {
   RETENTION = 1,
   CHUNK_SIZE = 1 << 1,
   DUPLICATE_POLICY = 1 << 2,
-  IGNORE = 1 << 3,
-  LABELS = 1 << 4,
+  LABELS = 1 << 3,
```
jihuayu: We change the `LABELS ` value. Why we need change it?

You can use comments to indicate that this item is not being used.

tests/gocase/unit/type/timeseries/timeseries_test.go
```diff
@@ -201,6 +201,24 @@ func testTimeSeries(t *testing.T, configs util.KvrocksServerConfigs) {
 		require.ErrorContains(t, rdb.Do(ctx, "ts.add", key, "1000", "13.4").Err(), "update is not supported when DUPLICATE_POLICY is set to BLOCK mode")
 	})
 
+	t.Run("TS.ADD Ignore Option", func(t *testing.T) {
```
jihuayu, line 204: The test cases are missing a significant number of negative scenarios where parameters fail to take effect. Could you please add them?

jihuayu (commented):
From what I can see, this change is both backward and forward compatible. Great job on the refactoring.

This is a fairly large change, so it took me some time to go through it all. Those are my thoughts for now. Thanks! @lauk20

tests/gocase/unit/type/timeseries/timeseries_test.go
lauk20: Thanks for the review. I have added more test scenarios for the IGNORE option. @jihuayu

PR #3456, opened 2026-04-17
fix(geo): delete store_key instead of user_key in SearchStore

songqing:
When GEOSEARCHSTORE yields zero results, the code incorrectly deleted the source key (user_key) instead of the destination key (store_key).


git-hulk (approved):

git-hulk:
@songqing Good catch!

jihuayu (approved):

PR #3459, opened 2026-04-22
fix(geo): validate positive COUNT arguments

jihuayu:
The original code overlooked the requirement for COUNT to be a positive integer, resulting in an inconsistency with the Redis protocol. 

I have modified the implementation to refactor the special case where count == 0 is treated as if no count were provided, ensuring code cleanliness and logical transparency.

Assisted-by: Codex

### DIFF
--- a/src/commands/cmd_geo.cc
+++ b/src/commands/cmd_geo.cc
@@ -89,16 +89,9 @@ class CommandGeoBase : public Commander {
   }
 
   static Status ParseCount(std::string_view raw_count, size_t *count) {
-    auto signed_count = ParseInt<int64_t>(raw_count, 10);
-    if (!signed_count) {
-      return {Status::RedisParseErr, errValueNotInteger};
-    }
+     auto signed_count = GET_OR_RET(ParseInt<int64_t>(raw_count, {0, INT64_MAX}, 10));
 
-    if (*signed_count <= 0) {
-      return {Status::RedisParseErr, "COUNT must be > 0"};
-    }
-
-    *count = static_cast<size_t>(*signed_count);
+    *count = static_cast<size_t>(signed_count);
     return Status::OK();
   }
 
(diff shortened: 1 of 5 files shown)


src/commands/cmd_geo.cc
```diff
@@ -88,13 +88,32 @@ class CommandGeoBase : public Commander {
...
+    if (signed_count) {
+      if (*signed_count <= 0) {
+        return {Status::RedisParseErr, "COUNT must be > 0"};
+      }
+
+      *count = static_cast<size_t>(*signed_count);
+      return Status::OK();
+    }
+
+    auto unsigned_count = ParseInt<size_t>(raw_count, 10);
```
PragmaTwice: didn't get why you parse it twice.

PragmaTwice (commented):
could you put why you make this change instead of what you change in PR description?

src/commands/cmd_geo.cc
jihuayu: Ohhh. I got what your means. I will change it latter.

~~Since we need to check if it's a positive number, size_t is always non-negative.~~

~~Perhaps we could use a helper function to determine if it's a positive number? It might make the logic clearer.~~

jihuayu:
> could you put why you make this change instead of what you change in PR description?



> could you put why you make this change instead of what you change in PR description?

Done. Thanks!

src/commands/cmd_geo.cc
jihuayu: Done

src/commands/cmd_geo.cc
```diff
@@ -88,13 +88,26 @@ class CommandGeoBase : public Commander {
...
+    auto signed_count = ParseInt<int64_t>(raw_count, 10);
+    if (!signed_count) {
+      return {Status::RedisParseErr, errValueNotInteger};
+    }
+
+    if (*signed_count <= 0) {
+      return {Status::RedisParseErr, "COUNT must be > 0"};
+    }
+
+    *count = static_cast<size_t>(*signed_count);
```
git-hulk:
```suggestion
     auto signed_count = GET_OR_RET(ParseInt<int64_t>(raw_count, {0, INT64_MAX}, 10));

    *count = static_cast<size_t>(signed_count);
```

jihuayu: Thanks!

PR #3462, opened 2026-04-25
chore: add /pkg to .gitignore

Tangruilin:
Assisted-by: Claude Code:glm-5

brefore:
image

after:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@
 /build/
 /cmake-build-*/
 /build-*/
+/pkg


jihuayu:
I'm curious why I've never run into this issue. Could you please let me know if you are using a specific IDE or a special configuration?

Tangruilin:
> I'm curious why I've never run into this issue. Could you please let me know if you are using a specific IDE or a special configuration?

 I'm using VS Code as my IDE, and I run tests with ./x.py test go. However, it seems that after I delete the /pkg directory once, it doesn't get added back when I run the command again.

Tangruilin:
> I'm curious why I've never run into this issue. Could you please let me know if you are using a specific IDE or a special configuration?

 It seems this is caused by a deprecated configuration in VS Code's Go extension: go.inferGopath. When it's set to true, it uses GOROOT as the cache location. It looks like it's a
 configuration issue on my end.

PR #3463, opened 2026-04-25
fix(tdigest): fix incorrect min/max default initialization

Tangruilin:
CentroidsWithDelta struct was default-initializing min and max to 0, which caused incorrect behavior when merging empty tdigests. Also delete Reset({}) overload to prevent accidental misuse.

Assisted-by: Claude Code:glm-5

### DIFF
--- a/src/types/tdigest.cc
+++ b/src/types/tdigest.cc
@@ -28,6 +28,7 @@ refer to https://github.com/apache/arrow/blob/27bbd593625122a4a25d9471c8aaf5df54
 #include <fmt/format.h>
 
 #include <algorithm>
+#include <initializer_list>
 #include <iterator>
 #include <queue>
 
@@ -355,6 +356,8 @@ class TDigest {
   void Merge(const std::vector<TDigest>& others);
   void Add(const std::vector<double>& items);
   void Reset(const CentroidsWithDelta& centroid_list);
+  // Callers must explicitly specify the initial state to reset to; implicit initialization like Reset({}) is forbidden.
+  void Reset(std::initializer_list<CentroidsWithDelta>) = delete;
   void Reset();
   CentroidsWithDelta DumpCentroids() const;
 
--- a/src/types/tdigest.h
+++ b/src/types/tdigest.h
@@ -22,6 +22,7 @@
 
 #include <fmt/format.h>
 
+#include <limits>
 #include <map>
 #include <numeric>
 #include <variant>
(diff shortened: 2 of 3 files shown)


src/types/tdigest.cc
```diff
@@ -355,6 +355,7 @@ class TDigest {
   void Merge(const std::vector<TDigest>& others);
   void Add(const std::vector<double>& items);
   void Reset(const CentroidsWithDelta& centroid_list);
+  void Reset(std::initializer_list<CentroidsWithDelta>) = delete;
   void Reset();
   CentroidsWithDelta DumpCentroids() const;
```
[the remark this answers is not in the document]
jihuayu: I agree with Copilot. Could you please add the header file?

src/types/tdigest.cc
```diff
@@ -355,6 +355,7 @@ class TDigest {
   void Merge(const std::vector<TDigest>& others);
   void Add(const std::vector<double>& items);
   void Reset(const CentroidsWithDelta& centroid_list);
+  void Reset(std::initializer_list<CentroidsWithDelta>) = delete;
```
jihuayu, line 360: Seeing this delete on its own is a bit confusing. Could you please add a comment explaining why it's being deleted?

src/types/tdigest.h
```diff
@@ -47,10 +48,10 @@ struct Centroid {
 
 struct CentroidsWithDelta {
```
jihuayu, line 49: Please add a regression test for merging only empty TDIGEST sources, then adding samples and verifying TDIGEST.MIN/MAX. This covers the externally observable behavior fixed by this PR.

jihuayu (commented):
General LGTM. Thanks!

src/types/tdigest.cc
Tangruilin: Done

src/types/tdigest.h
Tangruilin: Done
before:

Image

after:

Image

tests/cppunit/types/tdigest_test.cc
```diff
@@ -948,3 +948,63 @@ TEST_F(RedisTDigestTest, MergeWithUserSpecifiedCompression) {
   // Verify total observations: dest(1) + src(1) = 2
   EXPECT_EQ(metadata.total_observations, 2);
 }
+
+TEST_F(RedisTDigestTest, MergeEmptySourcesThenAddSamples) {
```
jihuayu: I don’t think we need a C++ unit test here. Could you change it to a Go integration test instead? For command-level behavior like this, we generally prefer Go integration tests.

Tangruilin: Done

jihuayu (approved):

LindaSummer (approved):
Hi @Tangruilin ,

LGTM 🚀
Thanks very much for your contribution!

PR #3466, opened 2026-04-28
fix(scripting): reject negative FCALL key count

songqing:
FCALL and FCALL_RO accepted numkeys=-1 because the validation only rejected values
smaller than -1. That value was then used to construct key/arg vector ranges from
the command arguments, which can produce an invalid iterator range and crash or
mis-handle the request.

Reject all negative numkeys values before splitting keys and args, matching the
command error message and preventing malformed FCALL input from reaching the Lua
call path.


jihuayu:
Hi @songqing! If you used AI tools to assist with your programming, please specify the tools and models you used. This will help us better assess the reliability of the code. Thanks.

jihuayu (approved):
Good catch! Thank you!

songqing:
> Hi @songqing! If you used AI tools to assist with your programming, please specify the tools and models you used. This will help us better assess the reliability of the code. Thanks.

I use codex now😄

git-hulk (approved):

Issue #3469, opened 2026-04-30
BITFIELD does not support Redis positional offset '#N' syntax


nkroker:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.


### Motivation

## Summary                                                                                                                                                         
                                                                                                                                                                     
  `BITFIELD` / `BITFIELD_RO` commands do not support Redis positional offset syntax `#N`.                                                                            
                                                                                                                                                                     
  ## Steps to Reproduce                                                                                                                                              
                                                                                                                                                                 
  BITFIELD mykey OVERFLOW SAT INCRBY u16 #0 1
     
  **Redis:** returns `(integer) 1`                                                                                                                                   
  **kvrocks:** returns error `bit offset is not an integer or out of range`                                                                                          
                                                                                                                                                                     
  ## Expected Behavior                                                                                                                                               
                                                                                                                                                                     
  `#N` expands to `N * bit_width`. Matches Redis behavior per [BITFIELD docs](https://redis.io/commands/bitfield/).                                                  
                                                                                                                                                                     
  | Encoding | `#0` | `#1` | `#2` |                                                                                                                                  
  |----------|------|------|------|                                                                                                                                  
  | `u8`     | 0    | 8    | 16   |                                                                                                                                  
  | `u16`    | 0    | 16   | 32   |                                                                                                                                  
  | `i32`    | 0    | 32   | 64   |                                                                                                                                  
                                                                                                                                                                     
  ## Actual Behavior                                                                                                                                                 
                                                                                                                                                                     
  Parser calls `ParseInt<uint32_t>` on raw offset string. `"#0"` fails — `#` prefix unhandled.                                                                       
                                                                                                                                                                     
  ## Environment                                                                                                                                                     
                                                                                                                                                                 
  - kvrocks version: `2.15.0`
  - Redis version tested against: `7.x`                                                                                                                              
                                                                                                                                                                     
  ## Additional Context                                                                                                                                              
                                                                                                                                                                     
  Affects `BITFIELD` GET, SET, INCRBY subcommands and `BITFIELD_RO`.

### Solution

## Solution                                                                                                                                                        
                                                                                                                                                                     
  In `CommandBitfield::Parse()` (`src/commands/cmd_bit.cc`), after encoding is parsed (bit width known), detect `#` prefix before calling `GetBitOffsetFromArgument`:
                                                                                                                                                                     
[...]

PR #3470, opened 2026-04-30
fix(bitfield): support Redis positional '#N' offset syntax

nkroker:
Linked issue #3469 

## Problem                                                                                                                                                         
                                                                                                                                                                     
  `BITFIELD` commands with positional offset `#N` syntax failed in kvrocks. Redis supports `#N` as shorthand for `N * bit_width`, so `INCRBY u16 #0 1` means bit     
  offset `0 * 16 = 0`, `#1` means `16`, etc.                                                                                                                         
                                                                                                                                                                     
  Example that worked in Redis but not kvrocks:                                                                                                                  
  BITFIELD mykey OVERFLOW SAT INCRBY u16 #0 1                                                                                                                        
                                                                                                                                                                     
  ## Root Cause                                                                                                                                                      
                                                                                                                                                                     
  `GetBitOffsetFromArgument` called `ParseInt<uint32_t>` directly on offset string. `"#0"` fails integer parsing — `#` prefix was never handled.                     
                                                                                                                                                                     
  ## Fix                                                                                                                                                             
                                                                                                                                                                 
  In `CommandBitfield::Parse()`, after encoding is parsed (bit width known), detect `#` prefix and expand: `offset = N * encoding.Bits()`. Plain integer offsets
  unchanged.
                                                                                                                                                                     
  ## Test                                                                                                                                                            
                                                                                                                                                                     
  Added integration tests in `tests/gocase/unit/type/bitmap/bitmap_test.go` covering:                                                                                
  - `SET`/`GET` with `#N` across multiple positions                                                                                                              
  - `INCRBY` with `#N`                                                                                                                                               
  - `OVERFLOW SAT INCRBY` with `#N` (original failing case)                                                                                                          
  - `BITFIELD_RO GET` with `#N`

### DIFF
--- a/src/commands/cmd_bit.cc
+++ b/src/commands/cmd_bit.cc
@@ -302,8 +302,14 @@ class CommandBitfield : public Commander {
       }
       cmd.encoding = encoding.GetValue();
 
-      // parse offset
-      if (!GetBitOffsetFromArgument(group[2], &cmd.offset).IsOK()) {
+      // parse offset — support Redis '#N' positional syntax: #N means N * bit_width
+      if (!group[2].empty() && group[2][0] == '#') {
+        auto pos_parse = ParseInt<uint32_t>(group[2].substr(1), 10);
+        if (!pos_parse) {
+          return {Status::RedisParseErr, "bit offset is not an integer or out of range"};
+        }
+        cmd.offset = *pos_parse * cmd.encoding.Bits();
+      } else if (!GetBitOffsetFromArgument(group[2], &cmd.offset).IsOK()) {
         return {Status::RedisParseErr, "bit offset is not an integer or out of range"};
       }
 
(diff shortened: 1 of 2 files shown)


src/VERSION.txt
```diff
@@ -1 +1 @@
-unstable
\ No newline at end of file
+2.15.0
```
PragmaTwice: Please revert this.

nkroker: Done @PragmaTwice

jihuayu (commented):
Good catch. There is just one small issue: `BITFIELD bf SET u16 '#268435456' 1` returns an out-of-range error in Redis, but in this PR it is calculated as offset 0.

Could you also add extra test cases for invalid and boundary `#N` offsets, such as `#`, `#abc`, `#-1`, and a multiplication overflow case?

jihuayu (approved):

git-hulk (approved):

PR #3452, opened 2026-04-15
feat(string): support IFEQ/IFNE/IFDEQ/IFDNE conditions for SET command

kirito632:
- Add conditional SET options: IFEQ, IFNE, IFDEQ, IFDNE
- Propagate WRONGTYPE error when key type mismatch for conditional operations
- Handle NX option with wrong type correctly (treat as key exists)
- Add test case for SET NX GET on wrong type

### AI-Assisted Contribution Disclosure
This contribution complies with the [ASF AI-assisted contribution guidelines](https://kvrocks.apache.org/community/contributing#guidelines-for-ai-assisted-contributions).
*   **AI Tool Usage :** AI was used as an assistant to suggest code patterns (e.g., `util::EqualICase`), scaffold Go test structures, and help locate the root cause of the `NX+GET` behavior issue.
*   **Human Implementation & Validation:** I independently implemented the core conditional logic (`IFEQ/IFNE/IFDEQ/IFDNE`), added the 16-char digest validation, and manually fixed the `NX+GET` short-circuiting bug to strictly match Redis 8.x semantics. I have thoroughly reviewed, understood, and locally tested every line of code in this PR.

This commit only includes SET command changes, DelEX changes will be in a separate PR as requested.

PR #3475, opened 2026-05-02
feat(string): support IFEQ/IFNE/IFDEQ/IFDNE in SET command

kirito632:
## What
Reintroduce conditional options IFEQ / IFNE / IFDEQ / IFDNE for the SET command.

## Context
This is a clean, reworked version of the previously closed #3452.  
It addresses all previous feedback, reverts the accidental deletion of `CommandDelEX`, and is fully rebased onto the latest `unstable` branch.  

All CI checks have been successfully run and verified on my personal fork before submission.

## Behavior
- IFEQ / IFNE: case-sensitive value comparison  
- IFDEQ / IFDNE: case-insensitive digest comparison  
- Return `WRONGTYPE` for non-string keys  
- Mutually exclusive with NX/XX  
- Fix: `SET key value NX GET` now correctly returns `WRONGTYPE` for non-string keys (matching Redis 8.x behavior)

## Testing
- C++ unit test: Kept only the IFDEQ empty-string boundary coverage  
- Go integration tests: Syntax, behavior, edge cases (uppercase & malformed digests), and regression coverage  

## AI-assisted Contribution Disclosure
AI was used for code pattern suggestions, test scaffolding, and debugging assistance.  
Core logic, bug fixes, validation, and final implementation were written and verified manually.

@jihuayu  PTAL, thanks!

### DIFF
--- a/src/types/redis_string.cc
+++ b/src/types/redis_string.cc
@@ -239,7 +239,7 @@ rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, c
 }
 
 rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, const std::string &value,
-                            StringSetArgs args, std::optional<std::string> &ret) {
+                            const StringSetArgs &args, std::optional<std::string> &ret) {
   uint64_t expire = 0;
   std::string ns_key = AppendNamespacePrefix(user_key);
 
@@ -249,6 +249,24 @@ rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, c
     uint64_t old_expire = 0;
     auto s = getValueAndExpire(ctx, ns_key, &old_value, &old_expire);
     if (!s.ok() && !s.IsNotFound() && !s.IsInvalidArgument()) return s;
+    // If the existing key is not a string type, enforce expected behaviors:
+    if (s.IsInvalidArgument()) {
+      // For conditional comparisons (IFEQ/IFNE/IFDEQ/IFDNE), reading the old value is required,
+      // so return the underlying WRONGTYPE (InvalidArgument) error.
+      if (args.type == StringSetType::IFEQ || args.type == StringSetType::IFNE || args.type == StringSetType::IFDEQ ||
+          args.type == StringSetType::IFDNE) {
+        return s;
+      }
+      // For NX option, treat a wrong type as "key exists" so the condition is not met.
+      if (args.type == StringSetType::NX) {
+        // If GET is also specified, we need to return the WRONGTYPE error
+        // because GET requires reading the old value.
+        if (args.get) return s;
+        ret = std::nullopt;
+        return rocksdb::Status::OK();
+      }
+      // For other options, continue (e.g., XX may still proceed since key exists).
+    }
     // GET option
     if (args.get) {
       if (s.IsInvalidArgument()) {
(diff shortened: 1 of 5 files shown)


tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -1252,3 +1252,397 @@ func testString(t *testing.T, configs util.KvrocksServerConfigs) {
 		require.Equal(t, value, rdb.Get(ctx, key).Val())
 	})
 }
+
+func TestSetConditional(t *testing.T) {
```
jihuayu: The new SET conditions fail to inherit the txn-context-enabled matrix. Existing string tests (lines 40, 49) are executed in both modes.

tests/gocase/unit/type/strings/strings_test.go
```diff
@@ -1252,3 +1252,397 @@ func testString(t *testing.T, configs util.KvrocksServerConfigs) {
...
+		for _, tc := range testCases {
+			t.Run(tc.name, func(t *testing.T) {
+				err := rdb.Do(ctx, "SET", "ifd-bad-digest", "world", tc.option, tc.digest).Err()
+				require.ErrorContains(t, err, "must be exactly 16 hexadecimal characters")
+				require.Equal(t, "hello", rdb.Get(ctx, "ifd-bad-digest").Val())
+			})
+		}
+	})
+
+	// ── Property tests (using testing/quick via subtests) ───────────────────
```
jihuayu: These aren't property tests; where did you use testing/quick? 

Please ensure you've actually read through your code and understand it yourself.

jihuayu (commented):
Thanks for your contribution!

The SET command is critical, as 99% of our users rely on it. 
We must ensure that its test cases are both comprehensive and accurate.

tests/gocase/unit/type/strings/strings_test.go
kirito632: Apologies for the earlier issues in test scaffolding and comments — I’m fixing them now.

While moving tests into `testString`, I noticed that the `configs` parameter is currently ignored:

```go
srv := util.StartServer(t, map[string]string{})
```
So tests are not actually running with txn-context enabled.

Would you prefer me to fix this in this PR, or keep this PR focused and address it separately?

Thanks for your patience!

kirito632: Before pushing here, I ran the full CI matrix on my personal fork and noticed two new environment-level failures. I did some triage and confirmed they are upstream environment/toolchain issues, completely unrelated to the SET command changes:

ArchLinux Container (jemalloc build failure): The rolling release updated to GCC 16.1.1, which removed the internal std::__throw_bad_alloc() symbol. This breaks jemalloc 5.3.1 during compilation. Jemalloc upstream has fixed this in master, but since 5.4.0 isn't out, it fails the build here.

Ubuntu ARM Clang (Server Startup Failure): The TestReplicationGroupSyncConfig fails exclusively on the ARM64 Clang runner with a connection refused error. The Kvrocks server process seems to exit unexpectedly during startup without generating an "Ooops!" crash log. Since the exact same code passes perfectly on ARM GCC and all x86 runners, this appears to be an ARM64 Clang-specific compiler or environment compatibility issue.

Please let me know your thoughts on how to proceed!

aleksraiden: We known and fix will be included into this PR #3478  or separated. Thanks about your investigation

aleksraiden: Build in ArchLinux fixed in this PR #3480

jihuayu: @kirito632 All right. Fixed

kirito632: I have updated the PR to address the two issues you pointed out (matrix inheritance and comments).
I also passed the actual configs parameter into util.StartServer.
I have pre-verified the full CI matrix on my personal fork and it is strictly green.

jihuayu (dismissed):

src/types/redis_string.cc
```diff
@@ -249,6 +249,26 @@ rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, c
...
+        return s;
+      }
+      // For NX option, treat a wrong type as "key exists" so the condition is not met.
+      if (args.type == StringSetType::NX) {
+        // If GET is also specified, we need to return the WRONGTYPE error
+        // because GET requires reading the old value.
+        if (args.get) {
+          return s;
+        }
+        if (!args.get) ret = std::nullopt;
```
git-hulk: Remove the unnecessary condition check: `if (!args.get)`

```suggestion
        if (args.get) return s;
        ret = std::nullopt;
```

src/types/redis_string.cc
```diff
@@ -271,6 +291,38 @@ rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, c
...
+      // condition met only when key exists AND value matches
+      bool matched = s.ok() && (old_value == args.cmp_value);
+      if (!matched) {
+        if (!args.get) ret = std::nullopt;
+        return rocksdb::Status::OK();
+      }
+      if (!args.get) ret = "";
+    } else if (args.type == StringSetType::IFNE) {
+      // condition not met when key exists AND value matches; key-not-found counts as met
+      bool not_matched = s.ok() && (old_value == args.cmp_value);
```
git-hulk: the meaning of this result should be `matched`? instead of `not_matched`?

src/types/redis_string.cc
```diff
@@ -271,6 +291,38 @@ rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, c
...
+      // condition met only when key exists AND digest matches (case-insensitive)
+      bool matched = s.ok() && util::EqualICase(util::StringDigest(old_value), args.cmp_value);
+      if (!matched) {
+        if (!args.get) ret = std::nullopt;
+        return rocksdb::Status::OK();
+      }
+      if (!args.get) ret = "";
+    } else if (args.type == StringSetType::IFDNE) {
+      // condition not met when key exists AND digest matches (case-insensitive); key-not-found counts as met
+      bool not_matched = s.ok() && util::EqualICase(util::StringDigest(old_value), args.cmp_value);
```
git-hulk: The same issue with the above comment

git-hulk (approved):

jihuayu (approved):

PR #3478, opened 2026-05-02
chore(build): change base docker image to debian:trixie-slim

aleksraiden:
Change base docker image to debian:trixie-slim - current used bookworm are too old. So, I propose to use stable tagged Debian image - now it's a codename trixie.

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -561,9 +561,6 @@ jobs:
 
           # FIXME: Workaround for gcc 15 issue for rocksdb https://github.com/apache/kvrocks/issues/2927
           echo "CXXFLAGS=$CXXFLAGS -include cstdint" >> $GITHUB_ENV
-          
-          # FIXME: Workaround for jemalloc 5.3.1 + GCC 16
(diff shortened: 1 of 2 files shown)


PragmaTwice:
Maybe we can pin to `trixie-slim`? `stable` can be changed without our notice.

aleksraiden:
> Maybe we can pin to `trixie-slim`? `stable` can be changed without our notice.

Hmm, at a first, I think about stable for all (all next releases). But now you are right, better would be a fixed name for each release. Thanks, I fix it.

jihuayu (dismissed):

aleksraiden:
Also, we need to disable jemalloc (use system) at Arch  (due to error with std::__throw_bad_alloc, its patch at this pr https://github.com/jemalloc/jemalloc/pull/2900)

Maybe make this change a separate PR?

aleksraiden:
Build in ArchLinux fixed in this PR https://github.com/apache/kvrocks/pull/3480

PragmaTwice (approved):

PragmaTwice:
Better to add a trixie job here:
https://github.com/apache/kvrocks/blob/5dcf4f6b31ebafd2da57f6b24212c496a3f6871a/.github/workflows/kvrocks.yaml#L514

Currently only Debian 12.

aleksraiden:
> build-and-test-in-container

I doing this at new PR, thanks! Good idea

PR #3480, opened 2026-05-03
chore(build): jemalloc disable flag for ArchLinux build

aleksraiden:
We need to disable jemalloc (and use system) at Arch (due to error with std::__throw_bad_alloc, its patch at this PR https://github.com/jemalloc/jemalloc/pull/2900)

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -525,6 +525,7 @@ jobs:
           - name: ArchLinux
             image: archlinux:base
             compiler: gcc
+            disable_jemalloc: -DDISABLE_JEMALLOC=ON
           - name: Rocky Linux 8
             image: rockylinux:8
             compiler: gcc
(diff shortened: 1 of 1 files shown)


PragmaTwice (dismissed):

.github/workflows/kvrocks.yaml
```diff
@@ -525,6 +525,7 @@ jobs:
           - name: ArchLinux
             image: archlinux:base
             compiler: gcc
+            disable_jemalloc: -DDISABLE_JEMALLOC=ON
```
PragmaTwice, line 529:
```suggestion
            # FIXME: enable this when https://github.com/jemalloc/jemalloc/pull/2900 is released
            disable_jemalloc: -DDISABLE_JEMALLOC=ON
```

PragmaTwice (approved):

.github/workflows/kvrocks.yaml
aleksraiden: Thanks, I missed about notice

torwig (approved):

Issue #3351, opened 2026-01-26
CuckooFilter: Implement CF.ADD command for CuckooFilter


LiuQhahah:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Implement CF.ADD command

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #3481, opened 2026-05-04
feat(cf): add cf.add command

nagisa-kunhah:
task id: #3351

### DIFF
--- a/src/cluster/slot_migrate.cc
+++ b/src/cluster/slot_migrate.cc
@@ -718,11 +718,6 @@ StatusOr<KeyMigrationResult> SlotMigrator::migrateOneKey(const rocksdb::Slice &k
       }
       break;
     }
-    case kRedisCuckooFilter: {
-      return {Status::NotOK,
-              "unsupported migration type: MBbloomCF command migration is not supported; "
-              "use raw key-value migration"};
-    }
     case kRedisHyperLogLog: {
       // HyperLogLog migration by cmd is not supported,
       // since it's hard to restore the same key structure for HyperLogLog
--- a/src/cluster/slot_migrate.cc
+++ b/src/cluster/slot_migrate.cc
@@ -718,6 +718,11 @@ StatusOr<KeyMigrationResult> SlotMigrator::migrateOneKey(const rocksdb::Slice &k
       }
       break;
     }
+    case kRedisCuckooFilter: {
+      return {Status::NotOK,
+              "unsupported migration type: MBbloomCF command migration is not supported; "
+              "use raw key-value migration"};
+    }
     case kRedisHyperLogLog: {
       // HyperLogLog migration by cmd is not supported,
       // since it's hard to restore the same key structure for HyperLogLog
--- a/src/storage/batch_extractor.cc
+++ b/src/storage/batch_extractor.cc
@@ -60,9 +60,6 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
     Metadata metadata(kRedisNone);
     auto s = metadata.Decode(value);
     if (!s.ok()) return s;
-    if (metadata.Type() == kRedisCuckooFilter) {
-      return rocksdb::Status::NotSupported("MBbloomCF command migration is not supported; use raw key-value migration");
-    }
 
     if (metadata.Type() == kRedisString) {
       command_args = {"SET", user_key, value.ToString().substr(Metadata::GetOffsetAfterExpire(value[0]))};
@@ -264,9 +261,6 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
         }
         break;
       }
-      case kRedisCuckooFilter:
-        return rocksdb::Status::NotSupported(
-            "MBbloomCF command migration is not supported; use raw key-value migration");
       // TODO: to implement the case of kRedisBloomFilter
       default:
         break;
--- a/src/storage/batch_extractor.cc
+++ b/src/storage/batch_extractor.cc
@@ -60,6 +60,9 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
     Metadata metadata(kRedisNone);
     auto s = metadata.Decode(value);
     if (!s.ok()) return s;
+    if (metadata.Type() == kRedisCuckooFilter) {
+      return rocksdb::Status::NotSupported("MBbloomCF command migration is not supported; use raw key-value migration");
+    }
 
     if (metadata.Type() == kRedisString) {
       command_args = {"SET", user_key, value.ToString().substr(Metadata::GetOffsetAfterExpire(value[0]))};
@@ -261,7 +264,10 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
         }
         break;
       }
-        // TODO: to implement the case of kRedisBloomFilter
+      case kRedisCuckooFilter:
+        return rocksdb::Status::NotSupported(
+            "MBbloomCF command migration is not supported; use raw key-value migration");
+      // TODO: to implement the case of kRedisBloomFilter
       default:
         break;
     }
--- a/src/storage/batch_extractor.cc
+++ b/src/storage/batch_extractor.cc
@@ -261,7 +261,7 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
         }
         break;
       }
-      // TODO: to implement the case of kRedisBloomFilter
+        // TODO: to implement the case of kRedisBloomFilter
       default:
         break;
     }
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -347,8 +347,6 @@ class CuckooChainMetadata : public Metadata {
   uint16_t expansion;
 
   /// The capacity of the first filter.
-  /// Later sub-filter capacities are derived as base_capacity * expansion^n after restart,
-  /// so we need to persist the base value in metadata.
   uint64_t base_capacity;
 
   /// Number of fingerprints per bucket
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -59,8 +59,8 @@ enum RedisType : uint8_t {
 };
 
 inline constexpr const std::array<std::string_view, kRedisTypeMax> RedisTypeNames = {
-    "none",      "string", "hash",      "list",      "set",         "zset",      "bitmap",
-    "sortedint", "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "cuckoofilter"};
+    "none",   "string",    "hash",      "list",        "set",       "zset",       "bitmap",      "sortedint",
+    "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "cuckoofilter"};
 
 struct RedisTypes {
   RedisTypes(std::initializer_list<RedisType> list) {
@@ -358,8 +358,12 @@ class CuckooChainMetadata : public Metadata {
 
   explicit CuckooChainMetadata(bool generate_version = true)
       : Metadata(kRedisCuckooFilter, generate_version),
-        n_filters(0), expansion(0), base_capacity(0),
-        bucket_size(0), max_iterations(0), num_deleted_items(0) {}
+        n_filters(0),
+        expansion(0),
+        base_capacity(0),
+        bucket_size(0),
+        max_iterations(0),
+        num_deleted_items(0) {}
 
   void Encode(std::string *dst) const override;
   using Metadata::Decode;
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -59,7 +59,7 @@ enum RedisType : uint8_t {
 };
 
 inline constexpr const std::array<std::string_view, kRedisTypeMax> RedisTypeNames = {
-    "none",   "string",    "hash",      "list",        "set",       "zset",       "bitmap",      "sortedint",
+    "none",   "string",    "hash",      "list",        "set",       "zset",       "bitmap",   "sortedint",
     "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "MBbloomCF"};
 
 struct RedisTypes {
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -60,7 +60,7 @@ enum RedisType : uint8_t {
 
 inline constexpr const std::array<std::string_view, kRedisTypeMax> RedisTypeNames = {
     "none",   "string",    "hash",      "list",        "set",       "zset",       "bitmap",      "sortedint",
-    "stream", "MBbloomCF", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "cuckoofilter"};
+    "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "MBbloomCF"};
 
 struct RedisTypes {
   RedisTypes(std::initializer_list<RedisType> list) {
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -54,12 +54,13 @@ enum RedisType : uint8_t {
   kRedisHyperLogLog = 11,
   kRedisTDigest = 12,
   kRedisTimeSeries = 13,
+  kRedisCuckooFilter = 14,
   kRedisTypeMax
 };
 
 inline constexpr const std::array<std::string_view, kRedisTypeMax> RedisTypeNames = {
     "none",      "string", "hash",      "list",      "set",         "zset",      "bitmap",
-    "sortedint", "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries"};
+    "sortedint", "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "cuckoofilter"};
 
 struct RedisTypes {
   RedisTypes(std::initializer_list<RedisType> list) {
@@ -332,6 +333,40 @@ class BloomChainMetadata : public Metadata {
   uint32_t GetCapacity() const;
 
   bool IsScaling() const { return expansion != 0; };
+};
+
+class CuckooChainMetadata : public Metadata {
+ public:
+  /// The number of sub-filters in the chain
+  uint16_t n_filters;
+
+  /// Expansion factor for new filters
+  /// When a filter is full, a new one is created with capacity = base_capacity * expansion^n
+  uint16_t expansion;
+
+  /// The capacity of the first filter
+  uint64_t base_capacity;
+
+  /// Number of fingerprints per bucket
+  uint8_t bucket_size;
+
+  /// Maximum number of cuckoo kicks before considering filter full
+  uint16_t max_iterations;
+
+  /// Track number of deleted items for maintenance
+  uint64_t num_deleted_items;
+
+  explicit CuckooChainMetadata(bool generate_version = true)
+      : Metadata(kRedisCuckooFilter, generate_version),
+        n_filters(0), expansion(0), base_capacity(0),
+        bucket_size(0), max_iterations(0), num_deleted_items(0) {}
+
+  void Encode(std::string *dst) const override;
+  using Metadata::Decode;
+  rocksdb::Status Decode(Slice *input) override;
+
+  uint64_t GetTotalCapacity() const;
+  bool IsScaling() const { return expansion > 0; }
 };
 
 enum class JsonStorageFormat : uint8_t {
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -60,7 +60,7 @@ enum RedisType : uint8_t {
 
 inline constexpr const std::array<std::string_view, kRedisTypeMax> RedisTypeNames = {
     "none",   "string",    "hash",      "list",        "set",       "zset",       "bitmap",      "sortedint",
-    "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "cuckoofilter"};
+    "stream", "MBbloomCF", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "cuckoofilter"};
 
 struct RedisTypes {
   RedisTypes(std::initializer_list<RedisType> list) {
@@ -335,7 +335,7 @@ class BloomChainMetadata : public Metadata {
   bool IsScaling() const { return expansion != 0; };
 };
 
-constexpr uint32_t kCuckooFilterDefaultPageSize = 2048;
+constexpr uint32_t kCuckooFilterDefaultPageSize = 2048;  // bytes
 
 class CuckooChainMetadata : public Metadata {
  public:
@@ -346,7 +346,9 @@ class CuckooChainMetadata : public Metadata {
   /// When a filter is full, a new one is created with capacity = base_capacity * expansion^n
   uint16_t expansion;
 
-  /// The capacity of the first filter
+  /// The capacity of the first filter.
+  /// Later sub-filter capacities are derived as base_capacity * expansion^n after restart,
+  /// so we need to persist the base value in metadata.
   uint64_t base_capacity;
 
   /// Number of fingerprints per bucket
@@ -358,7 +360,7 @@ class CuckooChainMetadata : public Metadata {
   /// Track number of deleted items for maintenance
   uint64_t num_deleted_items;
 
-  /// Target maximum payload size for each persisted Cuckoo Filter page
+  /// Target maximum payload size for each persisted Cuckoo Filter page, in bytes
   uint32_t page_size;
 
   explicit CuckooChainMetadata(bool generate_version = true)
--- a/src/storage/redis_metadata.h
+++ b/src/storage/redis_metadata.h
@@ -335,6 +335,8 @@ class BloomChainMetadata : public Metadata {
   bool IsScaling() const { return expansion != 0; };
 };
 
+constexpr uint32_t kCuckooFilterDefaultPageSize = 2048;
+
 class CuckooChainMetadata : public Metadata {
  public:
   /// The number of sub-filters in the chain
@@ -356,6 +358,9 @@ class CuckooChainMetadata : public Metadata {
   /// Track number of deleted items for maintenance
   uint64_t num_deleted_items;
 
+  /// Target maximum payload size for each persisted Cuckoo Filter page
+  uint32_t page_size;
+
   explicit CuckooChainMetadata(bool generate_version = true)
       : Metadata(kRedisCuckooFilter, generate_version),
         n_filters(0),
@@ -363,7 +368,8 @@ class CuckooChainMetadata : public Metadata {
         base_capacity(0),
         bucket_size(0),
         max_iterations(0),
-        num_deleted_items(0) {}
+        num_deleted_items(0),
+        page_size(kCuckooFilterDefaultPageSize) {}
 
   void Encode(std::string *dst) const override;
   using Metadata::Decode;
--- a/src/types/cuckoo_filter.h
+++ b/src/types/cuckoo_filter.h
@@ -1,0 +1,61 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace redis {
+
+// Cuckoo filter implementation from the paper:
+// "Cuckoo Filter: Practically Better Than Bloom" by Fan et al.
+// This is a bucket-based storage implementation where each bucket is stored
+// as an independent key-value pair in RocksDB
+class CuckooFilter {
+ public:
+  // Calculate the optimal number of buckets for the filter
+  static uint32_t OptimalNumBuckets(uint64_t capacity, uint8_t bucket_size) {
+    // A load factor of 95.5% is chosen for the cuckoo filter
+    uint32_t num_buckets = static_cast<uint32_t>(capacity / bucket_size / 0.955);
+    // Round up to next power of 2 for better hash distribution
+    if (num_buckets == 0) num_buckets = 1;
+    uint32_t power = 1;
+    while (power < num_buckets) power <<= 1;
+    return power;
+  }
+
+  // Generate fingerprint from hash (8-bit fingerprint, non-zero)
+  static uint8_t GenerateFingerprint(uint64_t hash) {
+    uint8_t fp = hash & 0xFF;
+    return fp == 0 ? 1 : fp;  // Ensure non-zero fingerprint
+  }
+
+  // Calculate alternate bucket index using XOR
+  static uint32_t GetAltBucketIndex(uint32_t bucket_idx, uint8_t fingerprint, uint32_t num_buckets) {
+    // Use a simple hash of the fingerprint for the XOR operation
+    uint32_t fp_hash = fingerprint * 0x5bd1e995;  // MurmurHash2 constant
+    return (bucket_idx ^ fp_hash) % num_buckets;
+  }
+};
+
+}  // namespace redis
--- a/src/types/cuckoo_filter.h
+++ b/src/types/cuckoo_filter.h
@@ -34,8 +34,8 @@ namespace redis {
 
 // Cuckoo filter implementation from the paper:
 // "Cuckoo Filter: Practically Better Than Bloom" by Fan et al.
-// This is a bucket-based storage implementation where each bucket is stored
-// as an independent key-value pair in RocksDB
+// Buckets are grouped into page values in RocksDB. The Cuckoo algorithm still
+// works with logical bucket indexes, while the storage layer maps buckets to pages.
 //
 // Hash calculation follows RedisBloom's design:
 // - fp = hash % 255 + 1 (fingerprint, non-zero, range: 1-255)
@@ -85,13 +85,10 @@ class CuckooFilter {
     return hash ^ (static_cast<uint64_t>(fingerprint) * 0x5bd1e995);
   }
 
-  // Legacy function for backward compatibility with tests
-  // Converts bucket index to hash, applies GetAltHash, then converts back to bucket index
+  // Calculate an alternate bucket from a bucket index and fingerprint.
   static uint32_t GetAltBucketIndex(uint32_t bucket_idx, uint8_t fingerprint, uint32_t num_buckets) {
-    // Treat bucket_idx as a hash value for the calculation
     uint64_t hash = bucket_idx;
     uint64_t alt_hash = GetAltHash(fingerprint, hash);
-    // Convert back to bucket index
     return static_cast<uint32_t>(alt_hash % num_buckets);
   }
 
--- a/src/types/cuckoo_filter.h
+++ b/src/types/cuckoo_filter.h
@@ -44,7 +44,7 @@ class CuckooFilter {
   // Calculate the optimal number of buckets for the filter
   static uint32_t OptimalNumBuckets(uint64_t capacity, uint8_t bucket_size) {
     // A load factor of 95.5% is chosen for the cuckoo filter
-    uint32_t num_buckets = static_cast<uint32_t>(capacity / bucket_size / 0.955);
+    auto num_buckets = static_cast<uint32_t>(static_cast<long double>(capacity) / bucket_size / 0.955L);
     // Round up to next power of 2 for better hash distribution
     if (num_buckets == 0) num_buckets = 1;
     uint32_t power = 1;
--- a/src/types/cuckoo_filter.h
+++ b/src/types/cuckoo_filter.h
@@ -21,6 +21,7 @@
 #pragma once
 
 #include <cstdint>
+#include <limits>
 #include <string>
 #include <utility>
 #include <vector>
@@ -41,6 +42,12 @@ namespace redis {
 // - bucket_index = hash % num_buckets (only apply modulo when indexing)
 class CuckooFilter {
  public:
+  static bool IsCapacitySupported(uint64_t capacity, uint8_t bucket_size) {
+    if (bucket_size == 0) return false;
+    auto num_buckets = static_cast<long double>(capacity) / bucket_size / 0.955L;
+    return num_buckets <= static_cast<long double>(std::numeric_limits<uint32_t>::max() / 2 + 1ULL);
+  }
+
   // Calculate the optimal number of buckets for the filter
   static uint32_t OptimalNumBuckets(uint64_t capacity, uint8_t bucket_size) {
     // A load factor of 95.5% is chosen for the cuckoo filter
--- a/src/types/cuckoo_filter.h
+++ b/src/types/cuckoo_filter.h
@@ -47,7 +47,7 @@ constexpr uint64_t kCuckooFilterAltHashMultiplier = 0x5bd1e995ULL;
 // - h1 = hash (primary hash)
 // - h2 = h1 ^ (fp * 0x5bd1e995) (alternate hash via XOR)
 // - bucket_index = hash % num_buckets (only apply modulo when indexing)
-class CuckooFilter {
+class CuckooFilterHelper {
  public:
   static bool IsCapacitySupported(uint64_t capacity, uint8_t bucket_size) {
     uint32_t num_buckets = 0;
--- a/src/types/cuckoo_filter.h
+++ b/src/types/cuckoo_filter.h
@@ -102,6 +102,30 @@ class CuckooFilterHelper {
 
   // Convenience overload for std::string
   static uint64_t Hash(const std::string &item) { return Hash(item.data(), item.size()); }
+
+  // Calculate the capacity of a sub-filter at a given index in the chain.
+  // Returns false if overflow would occur.
+  static bool CalculateFilterCapacity(uint64_t base_capacity, uint16_t expansion, uint16_t filter_index,
+                                      uint64_t *filter_capacity) {
+    uint64_t capacity = base_capacity;
+    for (uint16_t i = 0; i < filter_index; ++i) {
+      if (expansion != 0 && capacity > std::numeric_limits<uint64_t>::max() / expansion) return false;
+      capacity *= expansion;
+    }
+    *filter_capacity = capacity;
+    return true;
+  }
+
+  // Calculate the number of buckets for a sub-filter at a given index.
+  static rocksdb::Status GetFilterNumBuckets(uint64_t base_capacity, uint16_t expansion, uint8_t bucket_size,
+                                             uint16_t filter_index, uint32_t *num_buckets) {
+    uint64_t filter_capacity = 0;
+    if (!CalculateFilterCapacity(base_capacity, expansion, filter_index, &filter_capacity) ||
+        !IsCapacitySupported(filter_capacity, bucket_size)) {
+      return rocksdb::Status::Corruption("invalid metadata: filter capacity is too large");
+    }
+    return CalculateRequiredBuckets(filter_capacity, bucket_size, num_buckets);
+  }
 };
 
 }  // namespace redis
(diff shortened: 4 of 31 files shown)


jihuayu:
Hi @nagisa-kunhah. Thank you for your contribution.
In this PR, we have introduced a new data structure. This is a significant update that includes additions to the storage architecture. Could you describe the reasoning behind this design? We would like to understand your design philosophy and key decision points.

Additionally, please follow our AI policy: [https://kvrocks.apache.org/community/contributing#guidelines-for-ai-assisted-contributions](https://github.com/apache/kvrocks/pull/3481). Please let us know which AI tools and models you used, as this will help us better review the code.

nagisa-kunhah:
@jihuayu Hi, thank you for the review. I have summarized the design and implementation details below to explain the layering, responsibilities, and key storage decisions.
# 1 Design Overview

The design follows Kvrocks' existing layered architecture. Conceptually, the change can be divided into four layers:

1. Command layer
2. Type layer
3. Metadata/storage encoding layer
4. RocksDB persistence layer

## 1.1 Command layer

The command layer introduces RedisBloom-compatible Cuckoo Filter commands, such as `CF.RESERVE`, `CF.ADD`, `CF.EXISTS`, and `CF.MEXISTS`. Following the existing architecture, each command is implemented as a `Commander` subclass and registered through the existing command registration mechanism.

## 1.2 Type layer

The type layer introduces the new `CuckooChain` abstraction to represent each logical Cuckoo Filter.

The reason for using a chain abstraction is that Cuckoo Filters are not easy to resize in place: the bucket positions depend on the current bucket count, so resizing a filter would require rebuilding existing data. Instead, the design appends new sub-filters when expansion is needed. This model is inspired by RedisBloom's scalable Cuckoo Filter design.

`CuckooChain` implements the high-level operations for a logical Cuckoo Filter, including `RESERVE`, `ADD`, `EXISTS`, `MEXISTS`, and related key-level behavior.

To keep the type layer separated from the algorithm details, `CuckooFilter` is introduced as an internal helper. It provides Cuckoo Filter-specific calculations used by `CuckooChain`, while `CuckooChain` remains responsible for the Redis/Kvrocks-facing behavior.

## 1.3 Metadata/storage encoding layer

At the metadata/storage encoding layer, the implementation adds `CuckooChainMetadata` to store the state and parameters for each logical Cuckoo Filter. This metadata extends the existing Kvrocks metadata model, so common key-level fields such as type, size, expiration, and version are still handled consistently with other data types.

The Cuckoo-specific fields describe the filter chain, including the number of sub-filters, base capacity, bucket size, expansion factor, and insertion iteration limit. This keeps the logical filter state compact while leaving the actual bucket contents in subkeys.

## 1.4 RocksDB persistence layer

At the persistence layer, the implementation reuses Kvrocks' existing metadata/subkey model. The logical key metadata is stored as metadata, and individual buckets are stored as internal subkeys. This follows the same general pattern used by other complex Redis data structures in Kvrocks.

The metadata entry and bucket entries are written through RocksDB write batches when they need to be updated together. This keeps the logical filter state and the modified bucket data consistent without introducing a separate persistence path for Cuckoo Filter.

# 2 Logical Structure

A logical Cuckoo Filter is associated with one user key. Internally, it is represented as a chain of sub-filters rather than a single resizable filter.

The relationship between the main concepts is:

cuckoo_filter_logical_structure

- The logical filter is the user-visible Cuckoo Filter associated with a Redis key.
- A sub-filter is one filter segment in the chain. Expansion appends new sub-filters to the chain.
- Each sub-filter contains `num_buckets` buckets, and `num_buckets` is rounded to a power of two.
- A bucket belongs to one sub-filter and contains a fixed number of slots.
- A slot stores one fingerprint. A zero value means the slot is empty.
- A fingerprint is a compact representation derived from the item hash.

# 3 Implementation Details

## 3.1 Data Layout

### 3.1.1 Metadata

Each logical Cuckoo Filter key has one `CuckooChainMetadata` entry. The metadata describes the filter-level state and configuration, while the actual bucket data is stored separately.

The metadata is stored in the `metadata` column family. The RocksDB key is the namespace-prefixed logical Redis key, represented as `ns_key` in the implementation.

Conceptually:

```text
metadata CF:
  <ns_key> -> CuckooChainMetadata
```

The metadata contains the following fields:

| Field Name | Data Type | Size (Bytes) | Description |
|---|---|---:|---|
| `size` | `uint64_t` | 8 | The total number of items recorded in the entire filter chain. This field is inherited from the base `Metadata`. |
| `expire` | `uint64_t` | 8 | The expiration timestamp of the logical key. This field is inherited from the base `Metadata`. |
| `version` | `uint64_t` | 8 | The metadata version used to separate current subkeys from stale subkeys. This field is inherited from the base `Metadata`. |
| `n_filters` | `uint16_t` | 2 | The number of sub-filters in the chain. |
| `expansion` | `uint16_t` | 2 | The growth factor used when a new sub-filter is appended. |
| `base_capacity` | `uint64_t` | 8 | The capacity of the first sub-filter. The capacity of later sub-filters is derived from this value and `expansion`. |
| `bucket_size` | `uint8_t` | 1 | The number of fingerprint slots each bucket can hold. |
| `max_iterations` | `uint16_t` | 2 | The maximum number of relocation attempts during insertion. |
| `num_deleted_items` | `uint64_t` | 8 | The number of deleted items recorded by the filter. |

This metadata belongs to the logical Cuckoo Filter as a whole. It is not stored per bucket or per sub-filter.

### 3.1.2 Bucket Storage

The bucket data is stored as internal subkeys under the logical Cuckoo Filter key. Each bucket is identified by both a `filter_index` and a `bucket_index`.

The bucket subkey is constructed from:

- `filter_index`: identifies which sub-filter in the chain the bucket belongs to.
- `bucket_index`: identifies the bucket inside that sub-filter.

The bucket value is a fixed-size byte array whose length is `bucket_size`. Each byte stores one fingerprint. A value of `0` represents an empty slot, while valid fingerprints are stored as non-zero values.

Conceptually, the bucket layout is:

```text
PrimarySubkey CF:
  <bucket_key> -> bucket_data

bucket_key = InternalKey(<ns_key>, <encoded filter_index and bucket_index>, version)
```

Here, `filter_index` and `bucket_index` are encoded into the bucket subkey as binary fields, not as a numeric sum.

This layout keeps the logical filter metadata separate from the bucket contents, while still reusing Kvrocks' existing internal subkey model.

## 3.2 Hashing Model

### 3.2.1 Item Hash

Each item is first converted into a 64-bit hash using `HllMurMurHash64A` with seed `0`. This is the actual function name used in the codebase. It follows Redis' MurmurHash64A-style hash implementation and keeps the hashing model close to RedisBloom's Cuckoo Filter design.

The item hash is the base value used to derive both the fingerprint and the candidate bucket positions. Since these values determine where the item is stored and looked up, the hash function is part of the persistent data layout and should remain stable once data has been written.

### 3.2.2 Fingerprint

The fingerprint is generated from the item hash as `hash % 255 + 1`, producing an 8-bit non-zero value in the range `1..255`; `0` is reserved as the empty-slot marker.

### 3.2.3 Candidate Buckets

For each sub-filter, an item has two candidate buckets. The first bucket is derived directly from the item hash:

```text
bucket1 = hash % num_buckets
```

The second bucket is derived from both the hash and the fingerprint:

```text
delta = fingerprint * 0x5bd1e995
bucket2 = (hash ^ delta) % num_buckets
```

The constant `0x5bd1e995` follows RedisBloom's Cuckoo Filter implementation and is used as the mixing constant for deriving the alternate bucket.

During kick-out insertion, the original item hash of an evicted fingerprint is no longer available. Instead, the implementation uses the current bucket index and the fingerprint to compute the alternate bucket:

```text
alternate_bucket = (current_bucket ^ delta) % num_buckets
```

This is valid because `num_buckets` is always rounded to a power of two. When `num_buckets = 2^k`, modulo is equivalent to keeping the lower `k` bits, so:

```text
(hash ^ delta) % num_buckets == ((hash % num_buckets) ^ delta) % num_buckets
```

This lets the kick-out path move a fingerprint between its two candidate buckets using only the current bucket index and the fingerprint.

## 3.3 Current Write Path

### 3.3.1 CF.RESERVE

`CF.RESERVE` creates the logical Cuckoo Filter key and initializes its metadata. It validates the requested capacity and configuration parameters, checks that the key does not already exist, and then creates a `CuckooChainMetadata` entry with the initial filter configuration.

The initial metadata records the base capacity, bucket size, maximum insertion iterations, expansion factor, and initializes `n_filters` to `1`. The initial number of buckets is derived from the requested capacity and bucket size. The calculation uses a target load factor of `0.955`, which reserves extra slots instead of assuming that all slots can be filled successfully. The result is then rounded to a power of two.

The implementation does not preallocate all buckets during reserve. Buckets are created lazily when they are first written. This keeps `CF.RESERVE` lightweight and avoids writing empty bucket data for sparse filters.

### 3.3.2 CF.ADD

`CF.ADD` inserts an item into an existing logical Cuckoo Filter. It first loads and decodes the `CuckooChainMetadata`, then computes the item hash and fingerprint.

For each sub-filter in the chain, the implementation derives the two candidate buckets for the item. It reads these buckets, treats missing buckets as empty buckets, and tries to place the fingerprint into any available slot in either bucket.

For insertion, sub-filters are checked from the first one to the latest one, in `filter_index` order from `0` to `n_filters - 1`. This prioritizes reusing available slots in earlier sub-filters before placing data into newer sub-filters, which keeps the chain more compact and avoids expanding the effective write target too aggressively.

**This is different from RedisBloom, which checks sub-filters from the latest one back to the first one.**

If a free slot is found, the updated bucket data and the updated metadata are written in the same write batch. This keeps the bucket content and the logical filter state updated atomically.

If no free slot is available in the candidate buckets, the implementation falls back to kick-out insertion on the latest sub-filter. The kick-out path relocates existing fingerprints between their candidate buckets and writes all modified buckets together when the insertion succeeds.

### 3.3.3 Expansion

Expansion is triggered when insertion cannot find a free slot and kick-out insertion also fails. Instead of resizing an existing sub-filter in place, the implementation appends a new sub-filter to the chain.

The new sub-filter is represented by increasing `n_filters` in `CuckooChainMetadata`. Its capacity is derived from `base_capacity`, `expansion`, and the new `filter_index`.

The existing buckets are not rebuilt or moved during expansion. This avoids rewriting existing filter data. After expansion, the insertion is retried against the newly added sub-filter.

jihuayu:
Thanks for the proposal! These images are beautiful.

Hi @git-hulk @torwig @PragmaTwice @aleksraiden @LiuQhahah.
In this PR, we've added a new data structure. This part https://github.com/apache/kvrocks/pull/3481#issuecomment-4390189366 requires a review—could you all take a look?

jihuayu:
Hi @nagisa-kunhah. Regarding the current proposal, I have the following suggestions:

### Paging the Buckets
I believe that assigning one key per bucket will lead to a massive number of small keys, which is fatal to the system's performance. The overhead of RocksDB internal keys, memtables, index/filters, and compaction will be significantly greater than one or a few bytes. 

I suggest we introduce a **Page** abstraction for buckets, where one page contains multiple buckets. For example, a 1KB page could contain 256 buckets.

I have performed a rough estimation as follows:

#### Full Page Utilization
| Page Size | Buckets/Page | Actual Page Size | Full Page Utilization | Worst-case Amplification (1 Bucket) | Worst-case Amplification vs. Bucket-per-Key | Buckets Needed to Break Even |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 1KB | 256 | 1101B | 93.01% | 275x | 13.6x | 14 |
| 2KB | 512 | 2125B | 96.38% | 531x | 26.2x | 27 |
| 4KB | 1024 | 4173B | 98.15% | 1043x | 51.5x | 52 |
| 8KB | 2048 | 8269B | 99.07% | 2067x | 102.1x | 103 |

**Note on "Buckets Needed to Break Even":** This represents the minimum number of buckets that must be used within a page for the "paged" approach to become more space-efficient than the "bucket-per-key" approach. This occurs at approximately 5% occupancy.

#### Scenario for Default Capacity = 1024
| Page Size | Buckets/Page | Pages before 1st Expansion | New Sub-filter Pages before 2nd Expansion | Total Pages before 2nd Expansion |
| :--- | :--- | :--- | :--- | :--- |
| 1KB | 256 | 2 | 4 | 6 |
| 2KB | 512 | 1 | 2 | 3 |
| 4KB | 1024 | 1 | 1 | 2 |
| 8KB | 2048 | 1 | 1 | 2 |

Based on these findings, I recommend a default page size of **2KB** or **4KB**.

### Using MultiGet / Batch Read for Candidate Buckets/Pages
Operations like `CF.EXISTS`, `MEXISTS`, and `ADD` typically access two candidate buckets in each sub-filter. If we use bucket-level keys, this results in many random reads. Even with page-level keys, we should aggregate requests and use `MultiGet` to reduce the number of round-trips to RocksDB.

### Insertion Order: Prioritize Latest -> Old
RedisBloom queries from the newest sub-filter to the oldest. In Kvrocks, if we proceed from old to new, every write operation will first hit the older, fuller filters. This is likely to increase read amplification and the probability of "kick-outs." I suggest maintaining consistency with RedisBloom's "latest -> old" approach.

nagisa-kunhah:
@jihuayu Thanks for the suggestions. My understanding is that the Page abstraction is a persistent storage-layout unit, not an application-level cache.

I plan to replace the current bucket key layout:

```text
InternalKey(ns_key, version, filter_index + bucket_index) -> bucket data
```
with a page-based layout:
```text
InternalKey(ns_key, version, filter_index + page_index) -> multiple consecutive buckets
```
The bucket mapping would be:
```text
buckets_per_page = page_size / bucket_size
page_index = bucket_index / buckets_per_page
```
Each sub-filter would own its own set of pages, so pages are not shared across sub-filters. Also, I would treat `page_size` as the upper bound for a page value, not a fixed physical size. Small sub-filters and the last page of a sub-filter may store less than `page_size`


For the first version, I would like to use a fixed internal page size, likely `2KB`, and derive `buckets_per_page` when needed. We can consider making it configurable later, but then `page_size` should probably be stored in metadata because it affects the on-disk layout.

Could you confirm whether this matches your expectation, especially:
1. pages are scoped within each sub-filter and are not shared across sub-filters.
2. starting with a fixed internal page size is acceptable for the first version?

For the suggestions about `MultiGet` and insertion order, I agree with both points and will update the implementation accordingly.

jihuayu:
@nagisa-kunhah Your understanding is spot on. There’s no need to rush into code changes just yet, as others might still chime in with feedback. We can wait until everyone is on the same page before you start refactoring.


> page_size should probably be stored in metadata

I agree with you. Storing the page size makes sense; it will definitely make future extensions much easier to handle.

nagisa-kunhah:
@jihuayu  Excuse me, I’ve added the paging implementation and now included CuckooPageSet. The overall architecture is illustrated in the attached diagram. The diff is a bit large — would you recommend splitting it into smaller PRs for easier review?
图片

jihuayu:
Hi @nagisa-kunhah 
Breaking this into multiple commits makes it hard to create a self-contained logic flow, and it’s difficult to verify the correctness of individual PRs. This approach is OK. 

The review might take longer since the PR is quite large.

By the way, your image is very clear, and I really like it.

src/storage/redis_metadata.h
```diff
@@ -334,6 +335,50 @@ class BloomChainMetadata : public Metadata {
   bool IsScaling() const { return expansion != 0; };
 };
 
+constexpr uint32_t kCuckooFilterDefaultPageSize = 2048;
```
jihuayu: We need a comment here to indicate the units.

src/storage/redis_metadata.h
```diff
@@ -54,12 +54,13 @@ enum RedisType : uint8_t {
...
   kRedisTimeSeries = 13,
+  kRedisCuckooFilter = 14,
   kRedisTypeMax
 };
 
 inline constexpr const std::array<std::string_view, kRedisTypeMax> RedisTypeNames = {
-    "none",      "string", "hash",      "list",      "set",         "zset",      "bitmap",
-    "sortedint", "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries"};
+    "none",   "string",    "hash",      "list",        "set",       "zset",       "bitmap",      "sortedint",
+    "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "cuckoofilter"};
```
jihuayu: The type is `MBbloomCF` in redis source.

src/storage/redis_metadata.h
```diff
@@ -334,6 +335,50 @@ class BloomChainMetadata : public Metadata {
...
+ public:
+  /// The number of sub-filters in the chain
+  uint16_t n_filters;
+
+  /// Expansion factor for new filters
+  /// When a filter is full, a new one is created with capacity = base_capacity * expansion^n
+  uint16_t expansion;
+
+  /// The capacity of the first filter
+  uint64_t base_capacity;
```
jihuayu, line 350: Why we need store it at each Cuckoo Filter? Could you explain your decision?

src/types/cuckoo_filter.h
```diff
@@ -0,0 +1,103 @@
...
+    return OptimalNumBuckets(capacity, bucket_size, &num_buckets).ok();
+  }
+
+  // Calculate the optimal number of buckets for the filter.
+  static rocksdb::Status OptimalNumBuckets(uint64_t capacity, uint8_t bucket_size, uint32_t* num_buckets) {
+    if (bucket_size == 0) {
+      return rocksdb::Status::InvalidArgument("bucket_size must be larger than 0");
+    }
+
+    constexpr long double kLoadFactor = 0.955L;
```
jihuayu: We'd better extract it and keep all the constants in one place.

src/types/cuckoo_filter.h
```diff
@@ -0,0 +1,103 @@
...
+    if (required_buckets == 0) required_buckets = 1;
+
+    // Round up to next power of 2 for better hash distribution.
+    uint32_t power = 1;
+    while (power < required_buckets) power <<= 1;
+    *num_buckets = power;
+    return rocksdb::Status::OK();
+  }
+
+  // Generate fingerprint from hash (8-bit fingerprint, non-zero, range: 1-255)
```
jihuayu: This line is a redundant comment; the code is already clear.

src/types/cuckoo_filter.h
```diff
@@ -0,0 +1,103 @@
...
+// - bucket_index = hash % num_buckets (only apply modulo when indexing)
+class CuckooFilter {
+ public:
+  static bool IsCapacitySupported(uint64_t capacity, uint8_t bucket_size) {
+    uint32_t num_buckets = 0;
+    return OptimalNumBuckets(capacity, bucket_size, &num_buckets).ok();
+  }
+
+  // Calculate the optimal number of buckets for the filter.
+  static rocksdb::Status OptimalNumBuckets(uint64_t capacity, uint8_t bucket_size, uint32_t* num_buckets) {
```
jihuayu: This function name is no good. May be `CalculateRequiredBuckets` is better.

src/storage/redis_metadata.h
```diff
@@ -334,6 +335,50 @@ class BloomChainMetadata : public Metadata {
...
+  /// Number of fingerprints per bucket
+  uint8_t bucket_size;
+
+  /// Maximum number of cuckoo kicks before considering filter full
+  uint16_t max_iterations;
+
+  /// Track number of deleted items for maintenance
+  uint64_t num_deleted_items;
+
+  /// Target maximum payload size for each persisted Cuckoo Filter page
```
jihuayu: We need number units.

src/types/cuckoo_filter.h
```diff
@@ -0,0 +1,103 @@
...
+// "Cuckoo Filter: Practically Better Than Bloom" by Fan et al.
+// Buckets are grouped into page values in RocksDB. The Cuckoo algorithm still
+// works with logical bucket indexes, while the storage layer maps buckets to pages.
+//
+// Hash calculation follows RedisBloom's design:
+// - fp = hash % 255 + 1 (fingerprint, non-zero, range: 1-255)
+// - h1 = hash (primary hash)
+// - h2 = h1 ^ (fp * 0x5bd1e995) (alternate hash via XOR)
+// - bucket_index = hash % num_buckets (only apply modulo when indexing)
+class CuckooFilter {
```
jihuayu: This class looks like a helper, but the current naming could easily mislead us into thinking it's the Cuckoo Filter object itself.

src/types/cuckoo_filter_page.h
```diff
@@ -0,0 +1,91 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "storage/redis_metadata.h"
+#include "storage/storage.h"
+
+namespace redis {
+
+class CuckooPageSet {
```
jihuayu: The set suffix looks a bit misleading. It's just a single page, not a collection of pages, so it could easily cause confusion.

And I feel like the API boundary for this class is a bit blurry. The biggest issue is that it mixes 'storage access' and 'insertion policy' within the same class's public surface. Callers might get confused about whether this class is responsible for 'page storage access' or 'Cuckoo bucket insertion policy'.

src/types/redis_cuckoo_chain.h
```diff
@@ -0,0 +1,60 @@
...
+namespace redis {
+
+// Default values for Cuckoo Filter
+const uint32_t kCFDefaultCapacity = 1024;
+const uint8_t kCFDefaultBucketSize = 4;  // 4 fingerprints per bucket
+const uint16_t kCFDefaultMaxIterations = 500;
+const uint16_t kCFDefaultExpansion = 2;
+const uint16_t kCFMaxExpansion = 32768;
+
+class CuckooChain : public Database {
```
jihuayu, line 38: It seems we are missing a layer of abstraction, SubFilter.
It feels a bit too low-level for CuckooChain to interact with pages directly.

```text
CuckooChain
  -> SubFilter 0
    -> Page 0
    -> Page 1
    -> Page 2
  -> SubFilter 1
    -> Page 0
    -> Page 1
  -> SubFilter 2
    -> Page 0
    -> ...
```

src/types/redis_cuckoo_chain.h
```diff
@@ -0,0 +1,60 @@
...
+ public:
+  CuckooChain(engine::Storage *storage, const std::string &ns) : Database(storage, ns) {}
+
+  // CF.RESERVE command - creates a new cuckoo filter with specified capacity
+  rocksdb::Status Reserve(engine::Context &ctx, const Slice &user_key, uint64_t capacity, uint8_t bucket_size,
+                          uint16_t max_iterations, uint16_t expansion, uint32_t page_size);
+
+  // CF.ADD command - adds an item to the cuckoo filter.
```
jihuayu: CuckooChain should be a storage abstraction (rather than an instruction execution abstraction), so its Reserve and Add functions need to be designed around storage. The commands will have a thin orchestration layer to call CuckooChain.

So, there's no need to point out the corresponding command names in the comments.

src/types/redis_cuckoo_chain.cc
```diff
@@ -0,0 +1,350 @@
...
+
+  // Calculate the number of buckets needed for this filter
+  uint32_t num_buckets = 0;
+  s = CuckooFilter::OptimalNumBuckets(capacity, bucket_size, &num_buckets);
+  if (!s.ok()) return s;
+
+  INFO(
+      "Creating cuckoo filter with capacity={}, bucket_size={}, num_buckets={}, max_iterations={}, expansion={}, "
+      "page_size={}",
+      capacity, bucket_size, num_buckets, max_iterations, static_cast<int>(expansion), page_size);
```
jihuayu: It seems we don't need the info

src/types/redis_cuckoo_chain.cc
```diff
@@ -0,0 +1,350 @@
...
+
+  // Create a write batch for atomic operation
+  auto batch = storage_->GetWriteBatchBase();
+  WriteBatchLogData log_data(kRedisCuckooFilter, std::vector<std::string>{"CF.RESERVE", user_key.ToString()});
```
jihuayu: use `reserve` not "CF.RESERVE"

src/types/redis_cuckoo_chain.cc
```diff
@@ -0,0 +1,350 @@
...
+    CuckooPageSet pages(storage_, ctx, ns_key, metadata, storage_->IsSlotIdEncoded());
+    bool inserted = false;
+    s = pages.TryInsertInCandidateBuckets(current_filter_idx, num_buckets, bucket1_idx, bucket2_idx, fingerprint,
+                                          &inserted);
+    if (!s.ok()) return s;
+
+    if (inserted) {
+      // Successfully inserted, write to storage atomically
+      auto batch = storage_->GetWriteBatchBase();
+      WriteBatchLogData log_data(kRedisCuckooFilter, std::vector<std::string>{"CF.ADD", user_key.ToString()});
```
jihuayu: use `add` not "CF.ADD"

jihuayu (dismissed):
Hi @nagisa-kunhah. Sorry for the delay! I took a look at this PR, and overall, it looks good. 
The only thing is that I feel CuckooChain is doing a bit too much. Thanks!


Sorry, this was meant to be a comment, not an approval. Clicked the wrong button.

src/storage/redis_metadata.h
nagisa-kunhah: `base_capacity` represents the capacity of the first sub-filter. Later sub-filter capacities are derived from it using `base_capacity * expansion^i`.
Since both `base_capacity` and `expansion` are user-provided during `CF.RESERVE` and may differ across filters, we need to store `base_capacity` in each filter’s metadata.

src/types/redis_cuckoo_chain.h
nagisa-kunhah: ok, I added a thin `CuckooSubFilter` abstraction between `CuckooChain` and pages. `CuckooChain` now works with sub-filters, and page-level operations are handled inside `CuckooSubFilter`.

src/storage/redis_metadata.h
```diff
@@ -54,12 +54,13 @@ enum RedisType : uint8_t {
...
   kRedisTimeSeries = 13,
+  kRedisCuckooFilter = 14,
   kRedisTypeMax
 };
 
 inline constexpr const std::array<std::string_view, kRedisTypeMax> RedisTypeNames = {
-    "none",      "string", "hash",      "list",      "set",         "zset",      "bitmap",
-    "sortedint", "stream", "MBbloom--", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries"};
+    "none",   "string",    "hash",      "list",        "set",       "zset",       "bitmap",      "sortedint",
+    "stream", "MBbloomCF", "ReJSON-RL", "hyperloglog", "TDIS-TYPE", "timeseries", "cuckoofilter"};
```
PragmaTwice: why do you change `MBbloom--` to `MBbloomCF`?

nagisa-kunhah: Sorry, restore it. I just want to change “cuckoofilter” to “MBbloomCF”.

src/types/cuckoo_filter_page.h
nagisa-kunhah: I addressed this by narrowing the abstraction instead of keeping the old `CuckooPageSet` naming. 

There are two changes here:

1. Renamed `CuckooPageSet` to `CuckooPageCache`, since this class is really a per-operation page cache for cuckoofilter pages, not a set-like container.

2. Moved the candidate-bucket insertion policy out of the page cache boundary. The page cache now only owns page/bucket access concerns, such as loading/prefetching pages, reading/writing bucket slots, and writing dirty pages back. The sub-filter layer owns the insertion policy, including the order of trying the primary and secondary buckets.

jihuayu: Great! If next round of review is possible, please @ me.

nagisa-kunhah:
@jihuayu 
Hi, I'm ready for the next round of review. Here's a summary of the changes since last time:
- Decoupled CuckooPageCache from CuckooChainMetadata — it now takes scalar parameters (bucket_size, page_size) instead of holding a reference to the metadata struct.
- Refactored CuckooSubFilter — removed the metadata_ member; construction takes only the fields it needs.
- Moved CalculateFilterCapacity and GetFilterNumBuckets to CuckooFilterHelper — these are now static methods, making them independently testable and decoupled from the chain implementation.
- Expanded Go integration tests from 3 to 16 test cases covering edge cases (zero capacity, large capacity, expansion, full filter behavior, etc.).

src/types/redis_cuckoo_chain.cc
```diff
@@ -0,0 +1,269 @@
...
+  s = batch->Put(metadata_cf_handle_, ns_key, metadata_bytes);
+  if (!s.ok()) return s;
+
+  // Pages are created lazily on first write. Reserve only persists metadata so sparse filters don't preallocate page
+  // values that may never be used.
+
+  return storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+}
+
+rocksdb::Status CuckooChain::Add(engine::Context &ctx, const Slice &user_key, const Slice &item, bool *added) {
```
jihuayu, line 127: I feel this function is too long. Could we encapsulate the standard Cuckoo Filter steps into private helper functions? For example: `tryCuckooInsert`, `tryCuckooKickOut`, and `expandCuckooChain`.

src/types/redis_cuckoo_chain.cc
```diff
@@ -0,0 +1,269 @@
...
+    *added = true;
+    return rocksdb::Status::OK();
+  }
+
+  // Kick-out failed, try to expand if allowed
+  if (metadata.expansion > 0) {
+    if (metadata.n_filters >= UINT16_MAX) return rocksdb::Status::Aborted("maximum number of filters reached");
+
+    metadata.n_filters++;
+    INFO("add expanded to {} filters", metadata.n_filters);
```
jihuayu: We don't need INFO

src/types/redis_cuckoo_chain.cc
```diff
@@ -0,0 +1,269 @@
...
+    s = CuckooFilterHelper::GetFilterNumBuckets(metadata.base_capacity, metadata.expansion, metadata.bucket_size,
+                                                new_filter_idx, &new_num_buckets);
+    if (s.IsCorruption()) {
+      return rocksdb::Status::Aborted("maximum filter capacity reached");
+    }
+    if (!s.ok()) return s;
+
+    CuckooSubFilter new_filter(storage_, ctx, ns_key, storage_->IsSlotIdEncoded(), metadata.version,
+                               metadata.bucket_size, metadata.page_size, new_filter_idx, new_num_buckets);
+    s = new_filter.TryInsertPrimaryBucket(hash, fingerprint, &inserted);
```
jihuayu: Although the new sub-filter’s TryInsertPrimaryBucket can guarantee correctness, it feels semantically confusing. I suggest adding the necessary comments or using a normal insert instead.

src/types/cuckoo_filter_sub_filter.cc
```diff
@@ -0,0 +1,100 @@
...
+
+rocksdb::Status CuckooSubFilter::KickOutInsert(uint64_t hash, uint8_t fingerprint, uint16_t max_iterations,
+                                               bool *inserted) {
+  *inserted = false;
+
+  uint32_t current_bucket_idx = getPrimaryBucketIndex(hash);
+  uint8_t current_fp = fingerprint;
+  uint32_t victim_slot = 0;
+
+  for (uint16_t iteration = 0; iteration < max_iterations; ++iteration) {
```
jihuayu, line 56: This code pollutes dirty pages when the kick operation fails, which is inconsistent with the function's semantics. We can use the way RedisBloom do: if the kick fails, it swaps the previously swapped slots back in reverse order.

src/types/redis_cuckoo_chain.cc
```diff
@@ -0,0 +1,269 @@
...
+
+    *added = true;
+    return rocksdb::Status::OK();
+  }
+
+  // Kick-out failed, try to expand if allowed
+  if (metadata.expansion > 0) {
+    if (metadata.n_filters >= UINT16_MAX) return rocksdb::Status::Aborted("maximum number of filters reached");
+
+    metadata.n_filters++;
```
jihuayu: Could you move it to after the code that successfully creates the sub-filter? That would make the code clearer.

src/types/cuckoo_filter_page.cc
```diff
@@ -0,0 +1,231 @@
...
+    pages_.emplace(locations[i].page_key, std::move(page_entry));
+  }
+  return rocksdb::Status::OK();
+}
+
+rocksdb::Status CuckooPageCache::normalizePage(const rocksdb::Status &status, uint32_t expected_size, PageEntry *page) {
+  if (!status.ok() && !status.IsNotFound()) return status;
+  if (status.IsNotFound()) page->data.clear();
+  if (page->data.size() > expected_size) return rocksdb::Status::Corruption("invalid cuckoo filter page size");
+  if (page->data.size() < expected_size) page->data.resize(expected_size, 0);
```
jihuayu: I think we need to distinguish whether this page has already been created. If the page already exists, this should be treated as page corruption rather than automatically filling it with zeros.

src/types/cuckoo_filter_sub_filter.cc
```diff
@@ -0,0 +1,100 @@
...
+  if (!s.ok() || *inserted || bucket1_idx == bucket2_idx) return s;
+
+  return pages_.TryInsertInBucket(filter_index_, num_buckets_, bucket2_idx, fingerprint, inserted);
+}
+
+rocksdb::Status CuckooSubFilter::TryInsertPrimaryBucket(uint64_t hash, uint8_t fingerprint, bool *inserted) {
+  return pages_.TryInsertInBucket(filter_index_, num_buckets_, getPrimaryBucketIndex(hash), fingerprint, inserted);
+}
+
+rocksdb::Status CuckooSubFilter::KickOutInsert(uint64_t hash, uint8_t fingerprint, uint16_t max_iterations,
```
jihuayu: I think `TryKickOutInsert` is better. It may insert failed.

src/types/cuckoo_filter_page.cc
```diff
@@ -0,0 +1,231 @@
...
+  return std::max<uint32_t>(1, page_size / bucket_size);
+}
+
+uint32_t GetPageIndex(uint32_t bucket_index, uint32_t buckets_per_page) { return bucket_index / buckets_per_page; }
+
+uint32_t GetBucketOffset(uint32_t bucket_index, uint32_t buckets_per_page, uint8_t bucket_size) {
+  return (bucket_index % buckets_per_page) * bucket_size;
+}
+
+uint32_t GetPageValueSize(uint32_t page_index, uint32_t num_buckets, uint32_t buckets_per_page, uint8_t bucket_size) {
```
jihuayu: What does `PageValueSize` mean? I'm not quite sure what this name means.
Can you change a better name?

jihuayu (commented):
Hi @nagisa-kunhah. Thank you for bringing such a great implementation.

There are only a few minor issues left. Thank you.

src/types/cuckoo_filter_sub_filter.cc
nagisa-kunhah: In the current implementation, normal insert, kick-out insert, and expansion insert use separate `CuckooSubFilter` / `CuckooPageCache` instances.

If `TryKickOutInsert` fails, the dirty pages only stay in that temporary cache. We only call `last_filter.WriteToBatch()` when `inserted == true`, so failed kick-out changes are discarded and won't be written back to RocksDB.

My understanding is that RedisBloom needs to swap back because it mutates the in-memory filter directly. For Kvrocks, discarding the failed cache seems simpler and cheaper.

Do you think this approach is acceptable, or do you prefer rollback for clearer function semantics?

src/types/cuckoo_filter_page.cc
nagisa-kunhah: OK, I changed it.

src/types/cuckoo_filter_page.cc
nagisa-kunhah: Sure, I renamed it to `GetExpectedPageSize`.

This helper calculates the expected data size of a cuckoo filter page, based on how many buckets belong to that page. The last page may contain fewer buckets than a full page, so its expected size can be smaller.

src/types/cuckoo_filter_sub_filter.cc
jihuayu: What you said makes sense. But I feel this is more about how the lifecycle of dirty pages should be managed. 
I would prefer not to have the caller be responsible for managing dirty pages. If dirty pages must be managed by the caller, we need to explicitly document that in the comments.

However, this is indeed a trade-off, since the rollback operation would introduce additional time complexity. You can freely choose the solution you prefer, either adding a comment or performing the rollback.

nagisa-kunhah: I think the dirty page lifecycle can be managed by `CuckooSubFilter`. Cuckoo filter mutations are scoped to a single sub-filter, so the sub-filter can decide whether the staged page changes should be kept or discarded.

I added `CuckooPageCache::Discard()` so `CuckooSubFilter` can explicitly discard failed staged changes. I chose this instead of reverse swaps because swapping back would add extra overhead while the changes are still only in the page cache.

However, when the operation is successful, these dirty pages may not be committed by the sub filter, since the `engine::Storage` is managed by `CuckooChain`— and this storage is needed when writing data back to RocksDB. I don't know if this is acceptable or if we can refine it better.

jihuayu: Your design is excellent. I think it looks great.

I do have one small suggestion, though: the name **Discard** feels a bit too generic. Would it be possible to rename it to **DiscardCachedPages**?

nagisa-kunhah: ok, I change it to **DiscardCachedPages**.

src/types/redis_cuckoo_chain.cc
nagisa-kunhah: Your suggestion makes sense. I’ve refactored the function into three private helpers: `tryCuckooInsert`, `tryCuckooKickOut`, and `expandAndInsertCuckooChain`. Thanks for the tip!

jihuayu:
Hi @nagisa-kunhah. If you're ready for another round of review, please @ me directly. Thanks!

nagisa-kunhah:
> Hi @nagisa-kunhah. If you're ready for another round of review, please @ me directly. Thanks!

Ok, I'm reviewing the code by myself these days. I'll notify you when I'm ready.

nagisa-kunhah:
Hi, I'm ready for the next round of review. Here's a summary of the functional changes since last time:

  - Aligned the default Cuckoo Filter parameters with RedisBloom compatibility:
    - default BUCKETSIZE is now 2
    - default MAXITERATIONS is now 20
    - default EXPANSION is now 1
  - Normalized CF.RESERVE EXPANSION to the next power of two before storing it in metadata, matching RedisBloom behavior.
  - Preserved EXPANSION 0 as the non-scaling mode.
  - Added disk-size support for Cuckoo Filter keys so size-related stats no longer treat MBbloomCF keys as missing.

I also found that Cuckoo Filter support is still missing in the command-based migration path. Since Cuckoo Filter pages store fingerprints instead of the original items, command-based migration cannot reconstruct the original CF.ADD commands directly. Would it be OK to add migration support in a separate follow-up PR?
cc @jihuayu @PragmaTwice

jihuayu:
Thank you for your careful and detailed work.

>  Would it be OK to add migration support in a separate follow-up PR?

Yes! It would be best to put it in a new PR.


PS: I asked Copilot to review the code. It is not 100% correct, so feel free to ignore its comments if you disagree.

src/types/redis_cuckoo_chain.h
```diff
@@ -30,9 +30,9 @@ class CuckooSubFilter;
 
 // Default values for a newly created cuckoo filter.
 const uint32_t kCFDefaultCapacity = 1024;
-const uint8_t kCFDefaultBucketSize = 4;  // 4 fingerprints per bucket
-const uint16_t kCFDefaultMaxIterations = 500;
-const uint16_t kCFDefaultExpansion = 2;
+const uint8_t kCFDefaultBucketSize = 2;
+const uint16_t kCFDefaultMaxIterations = 20;
```
jihuayu, line 34: Is this change intended to align with Redis?

jihuayu (commented):
LGTM. Once the test failures are fixed, it should be good to go. Thank you for your continued contributions over such a long time.

jihuayu:
I’m not very sure about some parts of the code. Could you help me check the logic behind the changes in these files and see if there are any issues? cc @git-hulk @PragmaTwice 


### File list

src/cluster/slot_migrate.cc
src/storage/batch_extractor.cc
src/stats/disk_stats.h

src/types/redis_cuckoo_chain.h
nagisa-kunhah: Yes, this change is intended to align with RedisBloom. I referenced the default config in its module.conf, source: https://github.com/RedisBloom/RedisBloom/blob/5dad71021dbfc969d9384f170ed4956a83e2113d/module.conf#L64

jihuayu:
Sorry for the delay, this PR will need a few more days. Although it already has an LGTM, introducing a new data structure is a sensitive change, so I’d like to wait a few more days to see if anyone else has review comments.

src/storage/batch_extractor.cc
```diff
@@ -261,7 +264,10 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
         }
         break;
       }
-        // TODO: to implement the case of kRedisBloomFilter
+      case kRedisCuckooFilter:
+        return rocksdb::Status::NotSupported(
+            "MBbloomCF command migration is not supported; use raw key-value migration");
+      // TODO: to implement the case of kRedisBloomFilter
```
jihuayu: Remove it. 
https://github.com/nagisa-kunhah/kvrocks/blob/a6d50b66fa84ada9f19ea7a5bd52cd1f447183c4/src/storage/batch_extractor.cc#L270

src/storage/batch_extractor.cc
```diff
@@ -60,6 +60,9 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
     Metadata metadata(kRedisNone);
     auto s = metadata.Decode(value);
     if (!s.ok()) return s;
+    if (metadata.Type() == kRedisCuckooFilter) {
+      return rocksdb::Status::NotSupported("MBbloomCF command migration is not supported; use raw key-value migration");
+    }
```
jihuayu:
```suggestion

```

src/cluster/slot_migrate.cc
```diff
@@ -718,6 +718,11 @@ StatusOr<KeyMigrationResult> SlotMigrator::migrateOneKey(const rocksdb::Slice &k
       }
       break;
     }
+    case kRedisCuckooFilter: {
+      return {Status::NotOK,
+              "unsupported migration type: MBbloomCF command migration is not supported; "
+              "use raw key-value migration"};
+    }
```
jihuayu:
```suggestion

```

Please remove it. We just do as BloomFilter

tests/gocase/integration/replication/replication_test.go
```diff
@@ -344,10 +357,10 @@ func TestReplicationWithLimitSpeed(t *testing.T) {
 	util.Populate(t, slaveClient, "", 1026, 1)
 
 	t.Run("resume broken transfer based files", func(t *testing.T) {
-		// Try to transfer some files, because max-replication-mb 1,
-		// so maybe more than 5 files are transferred for sleep 5s.
 		util.SlaveOf(t, slaveClient, master)
-		time.Sleep(5 * time.Second)
+		require.Eventually(t, func() bool {
+			return slave.LogFileMatches(t, `.*\[fetch\] Fetched .*`)
+		}, 50*time.Second, time.Second)
```
jihuayu: This code seems unrelated to this PR; please revert to it.

tests/gocase/integration/replication/replication_test.go
```diff
@@ -359,7 +372,9 @@ func TestReplicationWithLimitSpeed(t *testing.T) {
 		require.Eventually(t, func() bool {
 			return slave.LogFileMatches(t, ".*skip count: 1.*")
 		}, 50*time.Second, 1000*time.Millisecond)
-		util.WaitForSync(t, slaveClient)
+		require.Eventually(t, func() bool {
+			return util.FindInfoEntry(slaveClient, "master_link_status") == "up"
+		}, 50*time.Second, 100*time.Millisecond)
```
jihuayu: This code seems unrelated to this PR; please revert to it.

jihuayu (commented):
It seems no one else had the time to look into this, so I did some research on it today.

@nagisa-kunhah I saw that you modified a lot of code related to slot migrate and extractor, which is great. However, I'm concerned it might affect other data structures. I suggest we revert those changes for now and align with the design of the Bloom filter.

Once this final issue is fixed, I will merge the code. 
Thanks for all your hard work this month. Victory is just around the corner!

nagisa-kunhah:
Hi @jihuayu, I have reverted the changes related to slot migration and the extractor, and aligned the behavior with RedisBloom.

I think we can continue supporting the remaining Cuckoo Filter commands in follow-up work as follows.

Thanks again for your long and careful review!

jihuayu (approved):

jihuayu:
Hi @nagisa-kunhah Thanks for your contribution. The PR has been merged.

Next, there are a few things we still need to do:

1. We need to update our documentation to record this data structure: https://kvrocks.apache.org/community/data-structure-on-rocksdb
2. We need to create tracking issues to manage the remaining work related to this command.

Could you help with these?

nagisa-kunhah:
> Hi @nagisa-kunhah Thanks for your contribution. The PR has been merged.
> 
> Next, there are a few things we still need to do:
> 
>     1. We need to update our documentation to record this data structure: https://kvrocks.apache.org/community/data-structure-on-rocksdb
> 
>     2. We need to create tracking issues to manage the remaining work related to this command.
> 
> 
> Could you help with these?

@jihuayu Yes, I'd like to do them. Regarding the documentation, do we just need to submit a PR to the kvrocks-website repo?

jihuayu:
>  do we just need to submit a PR to the kvrocks-website repo?

@nagisa-kunhah Yes! We need to update the documentation when we release, but we can prepare the PR first.

PR #3484, opened 2026-05-10
fix(replication): require admin for transfer commands

git-hulk:
Mark PSYNC, _FETCH_META, and _FETCH_FILE as admin-only commands so namespace users cannot invoke replication transfer internals.

Move the regression coverage into the integration replication suite, where the command behavior is exercised against a running server.

Assisted-by: Codex/GPT 5.5 xhigh

### DIFF
--- a/src/commands/cmd_replication.cc
+++ b/src/commands/cmd_replication.cc
@@ -483,9 +483,9 @@ class CommandWait : public Commander,
 };
 
 REDIS_REGISTER_COMMANDS(Replication, MakeCmdAttr<CommandReplConf>("replconf", -3, "read-only no-script", NO_KEY),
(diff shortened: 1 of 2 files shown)


git-hulk:
This might introduce the breaking change for the syncing tools, need to highlight this in our release notes.

jihuayu (approved):

jihuayu:
Do we need to follow a process of deprecation before deactivation?

git-hulk:
> Do we need to follow a process of deprecation before deactivation?

We don't have a process for those potential breaking changes for now, except for highlighting them in our release notes.

PragmaTwice (approved):

PragmaTwice:
I think we should treat them as "internal commands" and users are not expected to use them directly.

git-hulk:
> I think we should treat them as "internal commands" and users are not expected to use them directly.

Yes, exactly.

PR #3491, opened 2026-05-17
feat(info): add server_name:kvrocks flag

greatsharp:
If we want to manage kvrocks cluster and valkey(redis) cluster in one kvrocks-controller, we may need to distinguish them.
I found valkey has a flag 'server_name: valkey', so I would like to add a server_name flag in kvrocks too.

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -1219,6 +1219,7 @@ Server::InfoEntries Server::GetServerInfo() {
 
   Server::InfoEntries entries;
   entries.emplace_back("version", VERSION);  // deprecated
+  entries.emplace_back("server_name", SERVER_NAME);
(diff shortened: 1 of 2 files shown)


jihuayu:
Hi @greatsharp Thanks for your PR.
At present, compatibility with Valkey is not our goal. If you want to check if it's Kvrocks, you can just look for the kvrocks_version field. 

And, I think server_name is usually for the node name, not the type (such as: kafka).

cc @PragmaTwice @git-hulk What's your take on this?

PragmaTwice:
> need to distinguish them

It should be enough to check if `kvrocks_version` exists.

greatsharp:
kvrocks_versoin is ok.
valkey has both server_name: valkey and valkey_versoin fileld,
redis only has redis_versoin

PR #3492, opened 2026-05-17
fix(scripting): bump LuaJIT to reject bytecode loading

git-hulk:
Upgrade LuaJIT to a build that disables loading binary bytecode chunks, mitigating a DoS where crafted bytecode could crash the server. Add a regression test that confirms `loadstring` on malformed bytecode is refused with "attempt to load chunk with wrong mode" and the server keeps serving requests.

Co-authored-by: Claude Opus 4.7 (1M context)


jihuayu (dismissed):

jihuayu:
I noticed that we haven't updated this in a while. Kvrocks currently references a dependency from a April 2023 commit. In this particular version, we have the luajit.h file, but LuaJIT has since changed its build method, I don't see this file in repo now.

We need `cd src` and run `make luajit.h`. To make luajit.h

https://github.com/RocksLabs/LuaJIT/commit/c0a8e68325ec261a77bde1c8eabad398168ffe74

git-hulk:
> We need `cd src` and run `make luajit.h`. To make luajit.h

Yes, let me dive deep while I get time.

jihuayu:
There’s also a small issue: it looks like our current CMake script copies the headers before the build step, so the generated luajit.h will not be copied either.


https://github.com/apache/kvrocks/blob/unstable/cmake/luajit.cmake#L67
The file command only copies files during the configure phase, so it may miss the generated luajit.h that we actually need.

jihuayu (approved):

Issue #3231, opened 2025-10-23
Support for Namespaces and Numbered Databases in Kvrocks Cluster Mode


i18nsite:
[English](#en) / [中文](#cn)

<a id="en"></a>
# Support for Namespaces and Numbered Databases in Kvrocks Cluster Mode

## Feature Request

Valkey 9.0 has introduced full support for Numbered Databases in cluster mode. Previously, cluster deployments could only use a single database (db 0), which limited system flexibility and scalability. With this new feature, developers can now create multiple logical databases within the same cluster, avoiding key name conflicts and easily isolating workloads for different applications or tenants. This is particularly useful for multi-tenant architectures and sharding scenarios.

We suggest that Kvrocks should also add this support to enhance its capabilities in cluster environments.

https://www.linuxmi.com/valkey-9-0.html

## About Valkey

Valkey is a drop-in replacement for Redis, continuing the open-source tradition under the BSD license. Major cloud providers have joined to support Valkey as an alternative to Redis after Redis changed its license to non-open-source licenses (RSALv2 and SSPLv1) in March 2024.

Valkey currently has 6 core maintainers from Alibaba, Huawei, Tencent, Amazon, Google, and Ericsson. Arch Linux has officially deprecated Redis in favor of Valkey.

Given these developments, it would be beneficial for Kvrocks to consider aligning with Valkey's enhancements to remain competitive and compatible with the evolving ecosystem.

<a id="cn"></a>
# Kvrocks集群模式支持命名空间和编号数据库

## 功能请求

Valkey 9.0 在集群模式中新增了对编号数据库（Numbered Databases）的完整支持。过去，集群部署只能使用单一数据库（db 0），限制了系统的灵活性和可扩展性。现在，开发者可以在同一集群中创建多个逻辑数据库，避免键名冲突，并能轻松隔离不同应用或租户的负载——这对多租户架构和分片应用场景尤其有用。

建议Kvrocks也添加以上支持，以增强其在集群环境中的功能。

https://www.linuxmi.com/valkey-9-0.html

## 关于Valkey

Valkey是Redis的直接替代品，它在BSD许可证下继续开源传统。在Redis于2024年3月将其许可证更改为非开源许可证（RSALv2和SSPLv1）后，各大云厂商已加入支持Valkey作为替代方案。

Valkey目前有6位核心维护者，分别来自阿里、华为、腾讯、亚马逊、谷歌和爱立信。Arch Linux官方已将废弃Redis，转而支持Valkey。

[...]

PR #3500, opened 2026-05-24
feat(namespace): support namespaces and numbered databases in cluster mode

Abhinav2656:
**Resolves:**  #3231

**What this PR does / why we need it:**
This PR removes the restrictions that previously prevented the use of multiple namespaces and numbered databases when cluster mode is enabled. 

Specifically, it:
* Removes the `kErrClusterModeEnabled` rejection logic in `src/server/namespace.cc` to allow namespace creation and modification under cluster mode.
* Removes the configuration initialization guards in `src/config/config.cc` that forced a failure when both `cluster_enabled` and namespace tokens/databases were present.
* Ensures namespace configurations are correctly written to the persistent store regardless of the cluster state by removing the early return block.

**Brief test explanation:**
* Re-compiled successfully with native C++ tests passing.
* Executed and passed the Go integration tests specifically for `unit/namespace` (`go test ./unit/namespace ...`) under the `tests/gocase` suite.

Please let me know if any architectural adjustments are required. I am fully open to your review and stand ready to iterate on this logic as needed.

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -910,15 +910,6 @@ Status Config::finish() {
   if (requirepass.empty() && !load_tokens.empty()) {
     return {Status::NotOK, "requirepass empty wasn't allowed while the namespace exists"};
   }
-  if ((cluster_enabled) && !load_tokens.empty()) {
-    return {Status::NotOK, "enabled cluster mode wasn't allowed while the namespace exists"};
-  }
-  if ((redis_databases > 0) && !load_tokens.empty()) {
-    return {Status::NotOK, "redis-databases > 0 is not allowed while any non-default namespace exists"};
-  }
-  if ((redis_databases > 0) && (cluster_enabled)) {
-    return {Status::NotOK, "cluster mode and redis-databases cannot be enabled at the same time"};
-  }
   if (unixsocket.empty() && binds.size() == 0) {
     binds.emplace_back(kDefaultBindAddress);
   }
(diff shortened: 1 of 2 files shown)


PragmaTwice (changes requested):
Thanks for the contribution, but simply dropping these checks doesn't actually make namespaces work in cluster mode.

Abhinav2656:
Understood. I see now that removing the restrictions only exposes the lack of cluster-aware namespace routing and replication underneath. To ensure I build this correctly, could you point me toward the specific areas in the cluster topology or replication logic that need to be adapted to fully support namespace propagation?

jihuayu:
Hi @Abhinav2656. Thank you for your PR.

This is a fairly large project. Slot migration/cleanup still only handles `__namespace` by default, and we also need to consider whether slot mappings should be isolated across tenants. If you're interested in working on this, you can start by putting together a proposal.

Abhinav2656:
Hey @jihuayu, thanks for reviewing the PR and for the opportunity! I’m really excited to tackle a project of this scale. You raise a crucial point regarding the default namespace limitations and the need for proper tenant isolation across slot mappings.

To ensure we get the architecture right, I’m going to put together a detailed proposal/blueprint outlining how we handle these boundaries during migration. I'll share it here for your feedback first, and once we're completely aligned on the design, I'll move to the implementation.

jihuayu:
> Hey @jihuayu, thanks for reviewing the PR and for the opportunity! I’m really excited to tackle a project of this scale. You raise a crucial point regarding the default namespace limitations and the need for proper tenant isolation across slot mappings.
> 
> To ensure we get the architecture right, I’m going to put together a detailed proposal/blueprint outlining how we handle these boundaries during migration. I'll share it here for your feedback first, and once we're completely aligned on the design, I'll move to the implementation.

You can submit proposals in [discussions](https://github.com/apache/kvrocks/discussions). We will talk in it.

Abhinav2656:
I have drafted the full architectural proposal in the Discussions here: https://github.com/apache/kvrocks/discussions/3508

I am closing this PR for now so we can focus on reaching a consensus on the topology design first. I will open a fresh PR with the correct implementation once the blueprint is approved.

PR #3504, opened 2026-05-27
feat(stream): support XACKDEL command

kirito632:
## What

This PR introduces the XACKDEL stream command with configurable
PEL cleanup behaviors.

It also fixes an existing issue in XINFO CONSUMERS where consumer metadata
from different groups could leak into the same scan result.

## XINFO CONSUMERS Fix

Root cause:

[`GetConsumerInfo()`](src/types/redis_stream.cc) sets `iterate_upper_bound` by incrementing only the
stream version while keeping the sub-key prefix unchanged.

Since `InternalKey` encoding places the version field before the sub-key,
consumer records belonging to other groups can still fall within the
same `[version, version+1)` iteration range.

Fix:

Extract and validate the group name from each internal key during iteration,
filtering out unrelated consumers.

## XACKDEL

`XACKDEL key group [KEEPREF|DELREF|ACKED] IDS numids id [id ...]`

Deletes stream entries with configurable PEL cleanup behavior.

Supported deletion strategies:

**KEEPREF** (default):
Delete the stream entry only, without touching existing PEL references.
**DELREF**:
Delete the stream entry and remove it from all consumer groups' PELs.
**ACKED**:
Delete the stream entry only if it has been acknowledged by all consumer groups.

Per-ID return codes:

- `1`: entry deleted
- `2`: entry retained (e.g. ACKED condition not satisfied)
- `-1`: entry not found or duplicate ID within the same request

## Implementation Notes

- Pending-number decrements are aggregated in memory and flushed through
a single WriteBatch to avoid repeated metadata updates during DELREF operations.
- Duplicate IDs are deduplicated to preserve idempotency and prevent
stream metadata corruption.
- The originating command name is propagated into
[`WriteBatchExtractor`](src/storage/batch_extractor.cc) so replicas replay the exact
same deletion semantics.
- Internal stream metadata and PEL delete events are filtered out in
[`WriteBatchExtractor`](src/storage/batch_extractor.cc) to avoid generating invalid
replicated XDEL commands.

## Testing

Added Go integration tests covering:

- multi-group DELREF cleanup
- ACKED blocking semantics
- duplicate-ID idempotency
- invalid syntax handling
- XINFO CONSUMERS group isolation

## AI-Assisted Contribution Disclosure

This contribution complies with the ASF AI-assisted contribution guidelines.

AI tools were used to assist with English phrasing and some test scaffolding.
The core C++ implementation, debugging, correctness analysis, and final verification
were independently completed by me.

### DIFF
--- a/src/cluster/batch_sender.cc
+++ b/src/cluster/batch_sender.cc
@@ -60,7 +60,7 @@ Status BatchSender::PutLogData(const rocksdb::Slice &blob) {
   if (!s.ok()) {
     return {Status::NotOK, fmt::format("failed to put log data to migration batch, {}", s.ToString())};
   }
-  pending_logdata_only_ = true;
+  pending_logdata_only_ = pending_logdata_only_ || pending_entries_ == 0;
   pending_entries_++;
   entries_num_++;
   return Status::OK();
--- a/src/cluster/batch_sender.cc
+++ b/src/cluster/batch_sender.cc
@@ -38,6 +38,7 @@ Status BatchSender::Put(rocksdb::ColumnFamilyHandle *cf, const rocksdb::Slice &k
     return {Status::NotOK, fmt::format("failed to put key value to migration batch, {}", s.ToString())};
   }
 
+  pending_logdata_only_ = false;
   pending_entries_++;
   entries_num_++;
   return Status::OK();
@@ -48,6 +49,7 @@ Status BatchSender::Delete(rocksdb::ColumnFamilyHandle *cf, const rocksdb::Slice
   if (!s.ok()) {
     return {Status::NotOK, fmt::format("failed to delete key from migration batch, {}", s.ToString())};
   }
+  pending_logdata_only_ = false;
   pending_entries_++;
   entries_num_++;
   return Status::OK();
@@ -58,6 +60,7 @@ Status BatchSender::PutLogData(const rocksdb::Slice &blob) {
   if (!s.ok()) {
     return {Status::NotOK, fmt::format("failed to put log data to migration batch, {}", s.ToString())};
   }
+  pending_logdata_only_ = true;
   pending_entries_++;
   entries_num_++;
   return Status::OK();
@@ -89,6 +92,7 @@ Status BatchSender::Send() {
   sent_bytes_ += write_batch_.GetDataSize();
   sent_batches_num_++;
   pending_entries_ = 0;
+  pending_logdata_only_ = false;
   write_batch_.Clear();
   return Status::OK();
 }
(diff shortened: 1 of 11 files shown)


jihuayu:
Hi @kirito632. If the core C++ implementation part was also completed by AI, please state that directly as well — no need to worry about it. 
As long as you can take responsibility for the final code, we welcome AI-generated code. 
We can also exchange some related techniques and workflows.

kirito632:
Thanks for the clarification.
AI tools were indeed involved not only in wording and test scaffolding, but also in discussing implementation approaches and generating parts of the initial C++ code during development.
That said, I independently reviewed, debugged, modified, and verified the final implementation locally, including the stream deletion semantics, replication handling and metadata updates. I take full responsibility for the final code and behavior in this PR.
I'm happy to exchange related workflows and techniques as well.

kirito632:
Note on ACKED semantics:

  The Redis documentation states that ACKED only deletes entries that were "read and
  acknowledged by all consumer groups", but it does not explicitly describe how
  `last_delivered_id` interacts with PEL state when determining whether an entry can be
  deleted.

  I verified Redis 8.2 behavior with several scenarios:

  * A consumer group created with `$` does not block deletion of entries that predate its
  delivery cursor, even if it never acknowledged those entries.
  * A consumer group whose `last_delivered_id` is still behind the entry does block deletion,
  even when no PEL entry exists.
  * A consumer group that has already delivered the entry but still has a PEL entry for it
  also blocks deletion.
  * Once the PEL entry is acknowledged and removed, the same group no longer blocks deletion,
  even if its `last_delivered_id` is far beyond the entry.

  These results suggest that ACKED is not determined solely by PEL membership. Redis 8.2
  appears to treat the following conditions as independently preventing deletion:

  * The group has not yet reached the entry (`entry_id > last_delivered_id`).
  * The group still has a pending reference, meaning a PEL entry exists for the entry.

  The additional `last_delivered_id` check is intended to preserve this Redis 8.2 behavior by
  distinguishing two no-PEL cases: a group that has not reached the entry yet must still
  block deletion, while a group that was created past the entry, for example using `$`,
  should not block deletion.

jihuayu:
Good job. Verifying behavior against a real Redis instance is a method we commonly use. If there are still any doubts, you can also ask an AI tool to read the Redis source code—then the answer will become immediately clear.

jihuayu:
When your code is ready for review, please @ me and let me know. Thanks.

kirito632:
I noticed the same UINT64_MAX issue also exists in some shared stream ID parsers (ParseStreamEntryID(), ParseRangeEnd(), ParseNextStreamEntryIDStrategy()).

I intentionally limited this fix to the new XACKDEL parser because those shared helpers are used by several existing stream commands and changing them would significantly broaden the scope of this PR.

I'm happy to address the shared parsers in a separate follow-up PR if needed.

PR #3507, opened 2026-05-30
perf(test): parallelize bitmap subtests and optimize Set2SetBit with …

Tangruilin:
…pipeline

Part of #2642

- Add t.Parallel() to all 34 subtests for concurrent execution
- Use util.KeyPrefix(t) for unique key names per subtest
- Replace per-bit SETBIT calls with pipelined SETBIT in Set2SetBit (reduces ~8000 RTTs to 1 RTT for 1000-byte vectors)
- Replace defer with t.Cleanup() so resources live until subtests finish
- Move KeyPrefix helper to dedicated keys.go
- Fix duplicate test name "BITPOS BIT not found check check"
- Test time: 52s → 11s (~4.7x speedup)

Assisted-by: Claude Code:DeepSeek-v4-pro


jihuayu:
Greate! Thanks for the PR! I'll take a look at it in a few days.

tests/gocase/unit/type/bitmap/bitmap_test.go
```diff
@@ -122,261 +125,313 @@ func SimulateBitOp(op BITOP, values ...[]byte) string {
...
+		Set2SetBit(t, rdb, ctx, b2, []byte("\xac\x81\x32\x5d\xfe"))
+		Set2SetBit(t, rdb, ctx, b3, []byte("\xff\xff\xff\xff"))
+		require.EqualValues(t, []string{"\x00", "\x00\x00\x00\x00\x00", "\xac\x81\x32\x5d\xfe", "\xff\xff\xff\xff"}, GetBitmap(t, rdb, ctx, b0, b1, b2, b3))
 	})
 
 	t.Run("GET bitmap with out of max size", func(t *testing.T) {
+		t.Parallel()
+		p := util.KeyPrefix(t)
+		k := p + "b0"
 		require.NoError(t, rdb.Do(ctx, "config", "set", "max-bitmap-to-string-mb", 1).Err())
```
jihuayu, line 151: This test modifies the global config. For the sake of rigor, I think we should not run it in parallel.

jihuayu (commented):
Hi @Tangruilin I've been a bit busy lately, sorry for the late response. 
Overall, this was done really well. Thank you for your contribution in helping save our time.

tests/gocase/util/server.go
```diff
@@ -292,7 +292,6 @@ func StartServerWithCLIOptions(
 		},
 	}
 }
-
```
git-hulk, line 295: Revert this unrelated change

tests/gocase/unit/type/bitmap/bitmap_test.go
Tangruilin: I will fix it tonight

Tangruilin: done

jihuayu (approved):
Thank you!

Tangruilin:
> Thank you!

It looks like the coverage check didn't pass on my PR. The changes are entirely in Go test files, so I'm wondering — could we exclude test code from the coverage analysis?

jihuayu:
> > Thank you!
> 
> It looks like the coverage check didn't pass on my PR. The changes are entirely in Go test files, so I'm wondering — could we exclude test code from the coverage analysis?

Because this project won’t block the merge, you don’t need to worry about it. 

PS: I think there may still be some issues with the current coverage calculation.

git-hulk (approved):

PR #3511, opened 2026-06-03
ci: isolate kvrocks2redis coverage data

jihuayu:
The previous `go test` code coverage data could be overwritten by the coverage data from `kvrocks2redis`, which caused the overall coverage metric to be lower than expected.

This fixes the SonarCloud coverage job so the kvrocks2redis test no longer corrupts gcov data generated by the main C++ and Go tests.

Previously, the coverage job ran the normal test suite and then ran the kvrocks2redis test from the same coverage build. Both `kvrocks` and `kvrocks2redis` link `kvrocks_objs`, so they wrote to the same `.gcda` files. This produced `libgcov profiling error: overwriting an existing profile data with a different timestamp` and could make SonarCloud report some covered files as uncovered.

The fix keeps the coverage data isolated:

- collect the main C++/Go coverage into `coverage-main.json`
- run the kvrocks2redis source server from a separate coverage build directory
- run the kvrocks2redis tool itself from a non-coverage build directory to avoid writing conflicting `kvrocks_objs` profiles
- stop daemonized test processes before collecting coverage
- merge both tracefiles into one SonarQube XML report

This does not add new command coverage by itself. It makes sure coverage produced by existing tests is preserved and merged correctly instead of being overwritten by the kvrocks2redis stage.

Assisted-by: Codex/GPT5.5 xhigh

### DIFF
--- a/.github/workflows/kvrocks.yaml
+++ b/.github/workflows/kvrocks.yaml
@@ -369,6 +369,14 @@ jobs:
           cp -r build _build
           build-wrapper-linux-x86-64 --out-dir ${{ env.SONARCLOUD_OUTPUT_DIR }} ./x.py build -j$NPROC --unittest --compiler ${{ matrix.compiler }} ${{ matrix.sonarcloud }} --dep-dir ${{ env.KVROCKS_DEPS_CACHE_DIR }}
 
+      - name: Build kvrocks2redis test targets (SonarCloud)
+        if: ${{ matrix.sonarcloud }}
+        run: |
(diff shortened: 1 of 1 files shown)


jihuayu:
The reason for this PR is that I noticed the coverage report in https://github.com/apache/kvrocks/pull/3502 looked strange. The coverage for `src/commands/cmd_stream.cc` should not have been 0.

PragmaTwice (changes requested):
This is not acceptable to me that the CI time increases by about 35min (1h 24m 59s in total).

See https://github.com/apache/kvrocks/actions/runs/26887545442/job/79334883335?pr=3511.

jihuayu:
Ummm, I overlooked that. This is indeed unacceptable. I decided to parallelize the build steps for kvrocks2redis and Kvrocks.

Do you have a better approach?

PragmaTwice:
You can just disable the kvrocks2redis test in such matrix config.. Please do not increase CI time. It's already too slow..

Issue #3515, opened 2026-06-11
Flaky CI failures in proxy-based replication tests


jihuayu:
Several recent `kvrocks.yaml` CI runs failed in replication tests that simulate broken or slow replication links.

The most frequent failure is `TestSlaveLostMaster`:
```
--- FAIL: TestSlaveLostMaster
    client.go:147: forward tcp stream failed, err: read tcp 127.0.0.1:...: read: connection reset by peer
FAIL github.com/apache/kvrocks/tests/gocase/integration/replication
```

Examples:
- https://github.com/apache/kvrocks/actions/runs/27181757902
- https://github.com/apache/kvrocks/actions/runs/27136241761
- https://github.com/apache/kvrocks/actions/runs/27135463274
- https://github.com/apache/kvrocks/actions/runs/26640257024
- https://github.com/apache/kvrocks/actions/runs/26447014065

This does not seem related to the PR contents. It has appeared across unrelated stream/cf/hash/dependency changes and across different CI matrix entries, especially macOS arm64, but not only there.

The failure appears to come from `util.SimpleTCPProxy` calling `t.Fatalf` from the proxy forwarding goroutine when the TCP connection is reset. In `TestSlaveLostMaster`, the test intentionally breaks the proxy connection to simulate a lost master, so a connection reset during that phase may be expected and should probably not fail the test directly.

Proxy-based tests worth auditing:
- `TestSlaveLostMaster`
- `TestSlowConsumerBug`
- `TestSlowConsumerBlocksIndefinitely`

### Execute the necessary steps
- Investigate the reasons for failure
- Confirm the purpose of the test cases
- Propose solutions (communication is very important)
- Complete the code implementation

nhancdt2602:
@jihuayu I have a hypothesis about this flaky test. If it seems valid, I'd be happy work on a fix.

### Hypothesis

I believe `TestSlaveLostMaster` occasionally fails with a connection reset error because the connection is terminated with an RST packet instead of a FIN packet. As a result, the proxy treats it as an unexpected error and calls t.Fatalf, causing the test to fail.

```
n, err := src.Read(buf)
if err != nil {
    if err == io.EOF {
        break // clean close
    }
    log.Fatal(err) // or logger.Fatal / panic on unexpected error
}
```

`TestSlaveLostMaster` intentionally breaks the replication connection. It first calls `cancelProxy()`, then issues SETNODES to point the replica to a nonexistent master.

[...]

PR #3531, opened 2026-06-18
fix(test): extend the replication wait time to prevent the flaky test

git-hulk:
Currently, WaitForSync only allows 5s for master_link_status to reach "up",
which is not enough under loaded CI runners (e.g. SonarCloud coverage builds);

Reproduced locally by running 16 background CPU burners and looping the test:
```
for i in $(seq 1 6); do
   go test -count=1 -run TestReplicationWithLimitSpeed$ \
      ./integration/replication/... \
      -binPath=.../build/kvrocks -workspace=/tmp/kvrocks-test-ws
done
```

Before the fix: 1/6 runs failed with the same trace as the CI flake
("Condition never satisfied" at client.go:50, called from replication_test.go:362).
After the fix: 8/8 runs pass under the same CPU load (run times 34-51s vs. an
unloaded baseline of ~28s, confirming load was active).

Related flaky run: https://github.com/apache/kvrocks/actions/runs/27660686644/job/81804342125

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/gocase/integration/replication/replication_test.go
+++ b/tests/gocase/integration/replication/replication_test.go
@@ -359,7 +359,9 @@ func TestReplicationWithLimitSpeed(t *testing.T) {
 		require.Eventually(t, func() bool {
 			return slave.LogFileMatches(t, ".*skip count: 1.*")
 		}, 50*time.Second, 1000*time.Millisecond)
-		util.WaitForSync(t, slaveClient)
+		require.Eventually(t, func() bool {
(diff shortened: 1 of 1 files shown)


PragmaTwice (approved):

jihuayu:
This PR is related to the issue https://github.com/apache/kvrocks/issues/3515.

jihuayu (approved):

PR #3536, opened 2026-06-22
fix(server): scope CLIENT LIST/INFO/KILL to caller's namespace

git-hulk:
Non-admin (tenant) connections could previously enumerate and terminate connections belonging to other namespaces — including the admin namespace and replication links — via CLIENT LIST and CLIENT KILL. Filter both per-worker iteration and the slave-thread enumeration by the caller's namespace, allowing only admin (default-namespace) callers to see or kill connections outside their own namespace.

Assistant By Claude Opus 4.7

### DIFF
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -1805,16 +1805,19 @@ void Server::SlowlogPushEntryIfNeeded(const std::vector<std::string> *args, uint
   slow_log_.PushEntry(std::move(entry));
 }
 
-std::string Server::GetClientsStr() {
+std::string Server::GetClientsStr(const redis::Connection *conn) {
   std::string clients;
   for (const auto &t : worker_threads_) {
-    clients.append(t->GetWorker()->GetClientsStr());
+    clients.append(t->GetWorker()->GetClientsStr(conn));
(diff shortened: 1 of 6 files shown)


src/server/server.cc
```diff
@@ -1805,16 +1805,19 @@ void Server::SlowlogPushEntryIfNeeded(const std::vector<std::string> *args, uint
...
+    clients.append(t->GetWorker()->GetClientsStr(conn));
   }
 
-  std::shared_lock<std::shared_mutex> guard(slave_threads_mu_);
-
-  for (const auto &st : slave_threads_) {
-    clients.append(st->GetConn()->ToString());
+  // Slave (replication) connections live outside any tenant namespace, so
+  // only admin (default-namespace) callers may enumerate them.
+  if (conn == nullptr || conn->IsAdmin()) {
```
PragmaTwice: Does these `nullptr` check useful? I feel it's too "defensive programming".

git-hulk: It's good to remove the nullptr check.

PragmaTwice (approved):

Issue #2915, opened 2025-05-01
Receive notification for SET and DEL events on replicas (minimal implementation of notify-keyspace-events)


clickfreak:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Applications connected to the KVRocks read-only replica use their own cache to reduce response latency and minimize direct requests to KVRocks. However, the application must react to key changes in order to provide up-to-date responses.

We don’t use TTLs for keys and don’t expect key expiration events — this is also difficult to implement due to the nature of the KVRocks underlying storage (rocksdb). As minimum we need to react on SET and DEL commands, apps itself can control the subscription using PSUBSCRIBE and PUNSUBSCRIBE commands, it will allow to replace Redis/Valkey with minimal effort.

KVRocks is excellent for distributing data across the fleet without wasting RAM, and it’s a very strong candidate to replace our current Redis storage. Keyspace events are the last missing piece preventing us from fully switching from Redis/Valkey to KVRocks.

### Solution

From what I understand, KVRocks uses a WAL internally to replicate changes. This means that when the WAL is applied on replicas, it should be possible to detect key changes and emit events to applications with subscriptions by pattern (PSUBSCRIBE/PUNSUBSCRIBE).

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PragmaTwice:
Yeah we didn't support keyspace notification now (https://redis.io/docs/latest/develop/use/keyspace-notifications/).

Maybe it could be an OSPP project. cc @git-hulk

Aetherance:
Hi, I'd like to take this issue.

I plan to implement the scope described here: emit notifications for SET and DEL events on replicas, integrated with the existing PSUBSCRIBE/PUNSUBSCRIBE behavior. I'll check the replication apply path and add tests for the externally observable behavior.

If this sounds good, I'll start working on a PR.

jihuayu:
Hi @Aetherance. Thank you for your attention. Kvrocks currently does not support Redis keyspace notifications. Your solution looks reasonable. However, it would be better if you first submit an overall proposal in the [GitHub Discussions page](https://github.com/apache/kvrocks/discussions). After we reach a clear agreement, we can proceed with the implementation, which can help avoid rework later. Thank you.

Aetherance:
[...]

PR #3541, opened 2026-07-01
feat(keyspace): introduce notification support for the primary

Aetherance:
Implement initial keyspace notifications for set and del events, compatible with Redis notify-keyspace-events.

This PR adds the initial notification emitters and configuration support:
 - Support `K`, `E`, `g`, `$`, and `A` flags, where `A` currently expands to the implemented event classes `g$`.
 - Emit `set` notifications only when the shared `Set` path actually applies a write, including conditional `SET` variants.
 - Emit `del` notifications only for keys that are actually deleted through the shared delete path.
 - Deduplicate repeated keys in a single delete operation, so the same key is deleted and notified once.
  - Queue notifications inside `MULTI/EXEC` and publish them after a successful commit.
  - Map notification DB names correctly:
    - default namespace uses DB `0`
    - `redis-databases` namespaces map back to Redis DB indexes

This PR only adds the initial set and del event paths on primary nodes, keeping the initial patch small and reviewable. Commands sharing the same underlying mutation APIs may also produce these events. Replica-side notifications and additional event types, if supported, can be added in follow-up PRs.

  Unsupported notification classes are rejected for now until their emitters are implemented.

  Ref Proposal: #3533
  Tracking Issue: #2915
  
  Assisted by Codex/GPT-5.5.


Aetherance:
I added optional output parameters to `String::Set` and `Database::MDel` so the command layer can emit notifications based on what actually changed.

For `SET`, this avoids duplicating the conditional `SET` decision logic in the command layer. For `DEL`, this avoids an extra `EXISTS` pass before deletion, which would add unnecessary storage reads and overhead. It also lets `MDel` return the keys it actually deleted.

The alternative would be to keep these storage APIs unchanged and infer the notification decisions in the command layer. That would keep the APIs smaller, but the trade-off is that `SET` would duplicate conditional `SET`/`GET` logic outside `String::Set`, and `DEL` would need extra reads before `MDel`. Over time, this could also drift from the actual write behavior.

I chose the optional-output approach for this PR, but I’m open to maintainers’ preference here.

jihuayu (commented):
Looks great. I like the code you wrote.

Redis keyspace notifications treat `$` as the string command class. Besides `SET`, commands such as `SETEX`/`PSETEX`, `GETSET`, `MSET`, `APPEND`, `INCR*`, and `SETRANGE` also emit string events (`set`, `append`, `incrby`, `setrange`, etc.; `SETEX`/`PSETEX` also emit `expire`). This PR currently only emits notifications for literal `SET`. Is the remaining string command coverage intentionally deferred to a follow-up PR?

Aetherance:
> Looks great. I like the code you wrote.看起来很棒。我喜欢你写的代码。
> 
> Redis keyspace notifications treat `$` as the string command class. Besides `SET`, commands such as `SETEX`/`PSETEX`, `GETSET`, `MSET`, `APPEND`, `INCR*`, and `SETRANGE` also emit string events (`set`, `append`, `incrby`, `setrange`, etc.; `SETEX`/`PSETEX` also emit `expire`). This PR currently only emits notifications for literal `SET`. Is the remaining string command coverage intentionally deferred to a follow-up PR?Redis 键空间通知将 `$` 视为字符串命令类型。除了 `SET` 之外， `SETEX` 、 `PSETEX` 、 `GETSET` 、 `MSET` 、 `APPEND` 、 `INCR*` 和 `SETRANGE` 等命令也会触发字符串相关的事件（如 `set` 、 `append` 、 `incrby` 、 `setrange` 等； `SETEX` / `PSETEX` 也会触发 `expire` 相关事件）。目前的这个 PR 仅针对 `SET` 这种命令类型发出通知。至于其他字符串命令的处理，是否打算在后续的 PR 中再处理呢？

@jihuayu 

Yes, that is intentionally deferred to follow-up PRs.

This PR focuses on the minimal `SET` and `DEL` notification support first. Adding the rest of the string command coverage in the same PR would make the change much larger and harder to review, so I intentionally kept the scope limited here.

I’m happy to cover the remaining string commands such as `SETEX`/`PSETEX`, `GETSET`, `MSET`, `APPEND`, `INCR*`, and `SETRANGE` in follow-up PRs once the basic notification path is accepted.

jihuayu (dismissed):
Thank you! I’ll wait for others to reply over the next two days.

src/types/redis_string.cc
```diff
@@ -345,7 +347,9 @@ rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, c
   metadata.expire = expire;
   metadata.Encode(&new_raw_value);
   new_raw_value.append(value);
-  return updateRawValue(ctx, ns_key, new_raw_value);
+  auto s = updateRawValue(ctx, ns_key, new_raw_value);
+  if (s.ok() && applied != nullptr) *applied = true;
+  return s;
```
PragmaTwice: Why can't we just use the returned status to check? I can't see the value of `applied`.

src/storage/redis_db.cc
```diff
@@ -158,8 +159,10 @@ rocksdb::Status Database::Del(engine::Context &ctx, const Slice &user_key) {
   return storage_->Delete(ctx, storage_->DefaultWriteOptions(), metadata_cf_handle_, ns_key);
 }
 
-rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &keys, uint64_t *deleted_cnt) {
+rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &keys, uint64_t *deleted_cnt,
+                               std::vector<std::string> *deleted_user_keys) {
```
PragmaTwice: I think it's better to use `Slice` or `string_view` instead.

git-hulk:
@jihuayu I will take a look at this implementation since it might heavily impact the performance.

PragmaTwice (changes requested):
I don’t think this keyspace notification design is scalable.

Adding more code for each new event is easy, especially with AI agents. But that is not the real problem. The real problem is whether the design is maintainable.

This PR does not solve that. It adds duplicated code for each event and requires changes in many command implementations. I don’t see a clear shared abstraction here.

I don’t think we should merge this in the current form. We should first design a better general mechanism.

git-hulk:
> We should first design a better general mechanism.

Yes, I fully agree with this concern. We have an entry point to execute commands, so it should be possible to catch the changed events/data in one place instead of doing this in each command.

src/types/redis_string.cc
Aetherance: `Status` only tells whether the command errored, not whether it mutated the key. For example, src/types/redis_string.cc:L286-L289 returns OK for SET ... NX when the key already exists:

```cpp
  if (args.type == StringSetType::NX && s.ok()) {
    // if NX option given, the key already exist: return nil
    if (!args.get) ret = std::nullopt;
    return rocksdb::Status::OK();
  }
 ```

 This is a valid no-op per SET key value NX semantics, while a real successful write also returns OK. So the caller still needs applied to distinguish no-op success from write success.

Comment (Aetherance) [reply src/types/redis_string.cc]: the relevant code
https://github.com/apache/kvrocks/blob/970c2876034b9cea2d3364b5cde345ccf57e17fe/src/types/redis_string.cc#L284-L288

Comment (PragmaTwice) [reply src/types/redis_string.cc]: ahh looks reasonable.

**Comment by Aetherance** [comment]
@PragmaTwice @git-hulk Thanks for the careful review.

When I started implementing keyspace notifications, I also considered using a single common trigger point. However, while working on SET/DEL, I found that the generic command execution entry does not have enough semantic information to generate correct notifications.

For example, a conditional SET may return OK without actually writing the key, such as SET ... NX when the key already exists. DEL also needs to know the exact keys that were actually removed; cases like DEL k k k must not emit duplicate
notifications. These details are only known after the concrete command/type logic evaluates the command semantics, and cannot be derived reliably from the command name, arguments, Status, or reply at the generic execution entry.

 I agree the current implementation is not scalable enough. I’ll revisit the design and try to centralize the shared parts better.  I also think it may not be realistic to add more keyspace notification coverage without touching individual command/type implementations at all, because some event decisions still need to be made where the Redis semantics are fully known.

**Comment by git-hulk** [comment]
@Aetherance Thanks for your effort and reply.

> For example, a conditional SET may return OK without actually writing the key, such as SET ... NX when the key already exists. DEL also needs to know the exact keys that were actually removed; cases like DEL k k k must not emit duplicate
notifications.

After iterating on the current implementation and proposal, I understood your intention. And yes, it's a bit tricky to depend on rocksdb WAL to achieve this feature.

For the implementation, we can check config flags, channel names, Pub/Sub publishing in one place like `srv->NotifyKeyspaceEvent`. So we can just add one line for each place where we need to emit the event.

**Comment by Aetherance** [comment]
Hi @PragmaTwice @git-hulk, I have updated the design to introduce a common keyspace notification collection/publishing mechanism around the command execution entry point.

The common event representation is:

```cpp
struct KeyspaceEvent {
  // Event class flag such as g/$, not K/E channel selectors.
  int type_flag;
  std::string event;
  std::string ns;
  std::string key;
};
```

A reusable collector is attached to each command execution:

```cpp
// Collects semantic keyspace events for one command; Connection owns publish
// timing.
class KeyspaceEventCollector {
 public:
  // Sets namespace/config for the next command and drops any previous events.
  void Begin(std::string ns, int notify_flags);
  bool IsEnabled(int type_flag) const;
  void Add(int type_flag, std::string_view event, std::string_view key);
  // Moves out events collected during Execute.
  std::vector<KeyspaceEvent> Take();

 private:
  int notify_flags_ = 0;
  std::string ns_;
  std::vector<KeyspaceEvent> events_;
};
```

The command execution entry point now owns the shared notification lifecycle: begin collection, execute the command, take the collected events, and publish or queue them only if the command succeeds. The rest of the non-notification execution flow remains unchanged:

```cpp
Status Connection::ExecuteCommand(..., Commander *current_cmd, std::string *reply) {
  // existing non-notification logic ...

  current_cmd->BeginKeyspaceEventCollection(
      GetNamespace(), srv_->GetConfig()->notify_keyspace_events);
  auto s = current_cmd->Execute(ctx, srv_, this, reply);
  auto events = current_cmd->TakeKeyspaceEvents();

  if (s.IsOK()) {
    QueueOrPublishKeyspaceEvents(std::move(events));
  }

  // existing non-notification logic ...
  return s;
}
```

With this structure, notification support for a command follows the same pattern: the command/type layer reports factual semantic events into the collector, and the common execution path handles filtering, command success, transaction queueing, and publishing.

For example:

```cpp
if (set_applied) {
  keyspace_event_collector_.Add(kNotifyString, "set", args_[1]);
}
```

There is still some command-specific work needed when adding notification support for more commands, because the exact event semantics cannot always be inferred from the command name or arguments alone. For example, `DEL a a a` should emit only one `del` event, so `MDel` returns the actual deleted keys. Similarly, other commands may need to expose minimal outcome information such as whether a conditional write was applied or which keys were really changed.

So the common mechanism is now centralized and reusable, while each command only provides the minimal factual outcome needed to build the correct event list.

Does this direction look reasonable to you? If there is still something you would like to see changed, please let me know and I will update it.

Aetherance:
It looks like the full CI has not run on the latest commit. The previous CI run failed during startup because it was blocked by the repository’s action policy. Could a maintainer please help re-run or enable the CI when convenient?
@PragmaTwice @jihuayu  @git-hulk

Aetherance:
I found that this PR has significant room for performance optimization. I’ll convert it to a draft for now and request a review once I’ve completed the optimizations. Thanks for all reviewers!

src/common/keyspace_events.h
```diff
@@ -0,0 +1,68 @@
...
+
+#pragma once
+
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "status.h"
+
+// Flags for notify-keyspace-events, separate from RedisType.
```
Aetherance: To reduce the review burden and keep the scope of the changes manageable, this PR only adds support for the SET and DEL commands. Currently, only the $ and g event classes are supported. Support for other commands will be added in follow-up PRs.

src/common/keyspace_events.cc
```diff
@@ -0,0 +1,96 @@
...
+
+#include "keyspace_events.h"
+
+#include <cstring>
+#include <utility>
+
+#include "config/config.h"
+#include "fmt/format.h"
+
+bool ShouldNotifyKeyspaceEvent(int notify_flags, int type_flag) {
```
Aetherance: A notification is generated here only when both conditions are met: the event type is enabled, and at least one of K or E is enabled.

src/server/redis_connection.cc
```diff
@@ -466,7 +467,17 @@ Status Connection::ExecuteCommand(engine::Context &ctx, const std::string &cmd_n
 
   auto start = std::chrono::high_resolution_clock::now();
   bool is_profiling = IsProfilingEnabled(cmd_name);
+
+  keyspace_event_notify_flags_ = srv_->GetConfig()->notify_keyspace_events;
```
Aetherance: The keyspace_event_collector is created only when keyspace notifications are enabled, and notifications are emitted only when the collector exists. Therefore, when keyspace notifications are disabled—or when a command’s event type is not enabled—the command skips all notification-related work except for a few lightweight condition checks. As a result, the performance impact in these cases should be negligible.

Aetherance: By the way, support for additional commands can be added by simply calling conn->AddKeyspaceEvent. Some minor changes may also be needed to accurately capture the full semantics of the event, but no major changes to the notification mechanism should be necessary.

src/server/redis_connection.cc
```diff
@@ -709,10 +734,40 @@ void Connection::ExecuteCommands(std::deque<CommandTokens> *to_process_cmds) {
...
 void Connection::ResetMultiExec() {
   in_exec_ = false;
   multi_error_ = false;
   multi_cmds_.clear();
+  // Drop events from failed or aborted transactions.
+  pending_keyspace_events_.clear();
+  // Retain capacity for typical transactions, but request releasing unusually large buffers.
+  constexpr std::size_t kMaxRetainedKeyspaceEvents = 1024;
+  if (pending_keyspace_events_.capacity() > kMaxRetainedKeyspaceEvents) {
+    pending_keyspace_events_.shrink_to_fit();
```
Aetherance, line 756: clear() does not release the vector’s capacity, so a large transaction may cause a long-lived connection to retain excessive memory and potentially lead to OOM. To mitigate this, I call shrink_to_fit() when the capacity exceeds 1,024.

For small transactions, we can simply call `clear()` and reuse the allocated memory.

src/server/server.cc
```diff
@@ -478,6 +479,17 @@ int Server::PublishMessage(const std::string &channel, const std::string &msg) {
   return cnt;
 }
 
+void Server::NotifyKeyspaceEvent(int flags, const std::string &event, const std::string &ns, const std::string &key) {
+  const std::string db = MapNamespaceToKeyspaceDB(ns, GetConfig()->redis_databases);
```
Aetherance: Kvrocks has both ns and db.  `MapNamespaceToKeyspaceDB` provides a unified representation of the two when naming Pub/Sub channels.

src/types/redis_string.cc
```diff
@@ -239,7 +239,9 @@ rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, c
 }
 
 rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, const std::string &value,
-                            const StringSetArgs &args, std::optional<std::string> &ret) {
+                            const StringSetArgs &args, std::optional<std::string> &ret, bool *applied) {
```
Aetherance: Here, `Status::OK` does not necessarily mean that a conditional `SET` was applied, so an additional `applied` flag is needed. So does it in `DEL`

src/commands/cmd_key.cc
```diff
@@ -372,9 +374,17 @@ class CommandDel : public Commander {
     uint64_t cnt = 0;
     redis::Database redis(srv->storage, conn->GetNamespace());
 
-    auto s = redis.MDel(ctx, keys, &cnt);
+    const bool notify_del = GetAttributes()->name == "del" && conn->IsKeyspaceEventEnabled(kNotifyGeneric);
```
Aetherance: UNLINK is intentionally excluded here. Whether to support it will be decided in a follow-up PR or discussion.

Aetherance (commented):
To make the review easier, I added inline comments in several places where the intent may not be immediately clear from the code itself. I hope this makes the implementation easier to understand.

Thanks for the review! 🙏

Aetherance:
I ran multiple rounds of local performance benchmarks using `redis-benchmark`, covering several scenarios. I also included Redis as a reference to verify that this PR does not introduce a performance regression.

For example, the SET benchmark was run using the following command:

```bash
redis-benchmark -h 127.0.0.1 -p 6670 --threads 4 -c 64 -P 16 \
    -n 1000000 -r 1000000000 --seed 2002 --csv \
    SET bench:set:__rand_int__ 0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
```

The benchmarks covered the Kvrocks `unstable` branch, the current PR branch, and Redis with notifications both disabled and enabled, allowing comparison with the keyspace notification implementation in this PR.

### Kvrocks

| Version / configuration | Test | RPS | Avg latency (ms) | Min (ms) | P50 (ms) | P95 (ms) | P99 (ms) | Max (ms) |
|---|---|---:|---:|---:|---:|---:|---:|---:|
| unstable | DEL | 399,042.28 | 2.441 | 0.312 | 2.287 | 3.911 | 4.239 | 13.383 |
| unstable | SET | 498,753.09 | 1.798 | 0.248 | 1.831 | 2.287 | 2.479 | 9.895 |
| unstable | PING | 1,992,032.00 | 0.275 | 0.048 | 0.239 | 0.455 | 0.471 | 2.671 |
| PR, notifications disabled | DEL | 398,724.09 | 2.348 | 0.480 | 2.351 | 2.727 | 3.015 | 11.791 |
| PR, notifications disabled | SET | 570,450.62 | 1.757 | 0.096 | 1.839 | 2.367 | 2.551 | 11.679 |
| PR, notifications disabled | PING | 1,992,032.00 | 0.249 | 0.024 | 0.255 | 0.343 | 0.383 | 2.087 |
| PR, `KE$g` | DEL | 332,225.91 | 2.914 | 0.864 | 2.735 | 3.983 | 5.199 | 12.815 |
| PR, `KE$g` | SET | 399,201.59 | 2.327 | 0.352 | 2.423 | 3.351 | 4.151 | 13.863 |

### Redis reference

| Configuration | Test | RPS | Avg latency (ms) | Min (ms) | P50 (ms) | P95 (ms) | P99 (ms) | Max (ms) |
|---|---|---:|---:|---:|---:|---:|---:|---:|
| Notifications disabled | DEL | 1,329,787.25 | 0.636 | 0.120 | 0.639 | 0.911 | 1.063 | 3.271 |
| Notifications disabled | SET | 996,016.00 | 0.801 | 0.120 | 0.823 | 1.111 | 1.615 | 3.423 |
| `KE$g` | DEL | 664,451.81 | 1.411 | 0.168 | 1.439 | 1.871 | 2.247 | 7.111 |
| `KE$g` | SET | 569,800.56 | 1.601 | 0.192 | 1.615 | 2.175 | 2.655 | 4.311 |

I ran multiple rounds of these benchmarks and observed no significant variation between runs.

The results show no meaningful performance difference between the current PR and the original `unstable` branch when notifications are disabled or when unrelated commands are executed. This indicates that the PR has no measurable performance impact in these scenarios.

With `KE$g` enabled and an active subscriber connected, SET and DEL performance decreased. In this benchmark, the notification overhead in Kvrocks, measured as throughput loss, was approximately 16.68% for DEL and 30.02% for SET. Under the same conditions, the corresponding overhead in Redis was approximately 50.03% for DEL and 42.79% for SET.

Because Redis and Kvrocks use different storage engines, their absolute RPS values are not directly comparable. Instead, the relevant comparison is the relative performance change before and after notifications are enabled. The official Redis documentation also notes that keyspace notifications are disabled by default partly because the feature consumes additional CPU resources. [Redis Keyspace Notify](https://redis.io/docs/latest/develop/pubsub/keyspace-notifications/)

To further identify the source of the performance overhead, I profiled the PR branch using Linux `perf` and generated pprof-style differential call graphs from the samples. The graphs compare two scenarios with `KE$g` enabled: one without subscribers and one with an active subscriber. Red nodes indicate an increase in CPU usage after the subscriber was connected.

SET

set-notify pprof-style


DEL

del-notify pprof-style

The call graphs show that the performance bottleneck is in `Server::PublishMessage`. The underlying Pub/Sub delivery path is part of Kvrocks’ existing implementation, and this PR invokes it only when notifications are enabled and a notification event is actually generated. These results therefore indicate that the PR does not affect performance when notifications are disabled or when unrelated commands are executed. For the hot path that actually generates notifications, the observed performance cost has a clear and expected source.

jihuayu:
@Aetherance Wow, you’re incredibly thorough, and the diagrams are excellent. I really like them. I’m currently away on a business trip and have been quite busy, but I’ll make time to take a look. Thank you for your patience.

Aetherance:
> @Aetherance Wow, you’re incredibly thorough, and the diagrams are excellent. I really like them. I’m currently away on a business trip and have been quite busy, but I’ll make time to take a look. Thank you for your patience.

@jihuayu  Thank you! I hope the comments and diagrams make this large PR easier to review—it’s my responsibility as the author. Feel free to reach out with any questions.

src/commands/cmd_key.cc
```diff
@@ -372,9 +374,17 @@ class CommandDel : public Commander {
...
+    std::vector<rocksdb::Slice> deleted_keys;
+    if (notify_del) deleted_keys.reserve(keys.size());
+
+    auto s = redis.MDel(ctx, keys, &cnt, notify_del ? &deleted_keys : nullptr);
     if (!s.ok()) return {Status::RedisExecErr, s.ToString()};
 
+    for (const auto &key : deleted_keys) {
+      conn->AddKeyspaceEvent(kNotifyGeneric, "del", std::string_view(key.data(), key.size()));
+    }
+
```
jihuayu: cc @git-hulk @PragmaTwice 
Do you think this is a good design? My concern is that, with this approach, all the affected keys would need to be propagated from the storage layer up to the command layer, and every storage-layer method would need an additional field for them.

Would it be better to emit the event immediately after the deletion is performed in the storage layer, rather than propagating the information all the way up to the command layer?

Aetherance: > Would it be better to emit the event immediately after the deletion is performed in the storage layer, rather than propagating the information all the way up to the command layer?

The initial version of this PR used a similar approach, but a reviewer pointed out that it was not very extensible:

> Yes, I fully agree with this concern. We have an entry point to execute commands, so it should be possible to catch the changed events/data in one place instead of doing this in each command.

Therefore, I changed it to the current approach, where events are emitted at the unified command execution entry point.

Aetherance: According to my first commit: https://github.com/apache/kvrocks/pull/3541/changes/44d332b9aa35d10d245893efd100229b7e3ef02b#diff-4680a40ab45d8476a5a37bb17d76163432115086c0fa2dd613ea80e5aee1cdfbR393

jihuayu: What I mean is, could we emit the event here instead of propagating it up to the command layer?

https://github.com/apache/kvrocks/pull/3541/changes#diff-df4860055937f1bb78946c49a63a6c4120812a1ec6229a5ded3aa3536220c253R210

Please don’t rush to make changes. I’d like to hear what others think first.

Aetherance: Thank you for the patient explanation.

I considered emitting the notification directly here when I first started implementing this PR, since this is indeed the cleanest place to access the complete semantics of the operation. I ultimately decided against it for the following reasons, which I hope help explain the decision:

The main reason is that this layer has no access to `Server` or `Connection`, so it cannot call `PublishMessage`. I also don't think the storage layer should be responsible for publishing messages.

jihuayu: @Aetherance  It looks like they don’t have time to look into this issue. I’ve come up with a possible solution—do you think it would be better?
code: https://github.com/jihuayu/kvrocks/pull/6

Please note that this is only an idea, and I cannot guarantee that it is 100% correct. I’d like to hear your thoughts. You can view the diff here.https://github.com/jihuayu/kvrocks/compare/376592a1337312749e52cb504cd965d2d711bdf3...75ca860f01c8828f82e4af62a778807c4b70e9bb

jihuayu:
@Aetherance I think this PR would be a great fit for [stacked PRs](https://docs.github.com/en/pull-requests/get-started/stacked-prs-quickstart).

This is not a requirement. If you are interested in stacked PRs and would like to try them out, you could split this PR into a stack of smaller PRs. If that feels like too much trouble, you do not have enough time, or you are simply not interested, please feel free to ignore this message.

This is only an invitation in case the idea interests you.

src/commands/cmd_key.cc
Aetherance: @jihuayu  Thank you! I’ll take a careful look at the implementation and get back to you as soon as possible.

Aetherance:
> @Aetherance I think this PR would be a great fit for [stacked PRs](https://docs.github.com/en/pull-requests/get-started/stacked-prs-quickstart).
> 
> This is not a requirement. If you are interested in stacked PRs and would like to try them out, you could split this PR into a stack of smaller PRs. If that feels like too much trouble, you do not have enough time, or you are simply not interested, please feel free to ignore this message.
> 
> This is only an invitation in case the idea interests you.

Hi @jihuayu . Thanks for the suggestion! I’d be happy to try this approach.

However, since this PR is being submitted from a fork and I don’t have permission to push branches directly to `apache/kvrocks`, I’m still not sure how the native stacked PR workflow would work in this case. I noticed that one of the prerequisites may prevent me from using this feature:

> A GitHub repository you can push to.

From:
https://docs.github.com/en/pull-requests/get-started/stacked-prs-quickstart#prerequisites

If the native stacked PR workflow is indeed not applicable here, I’d also be happy to split the current PR into several regular PRs.

src/commands/cmd_key.cc
Aetherance: Hi @jihuayu , I took a closer look at this approach, and passing key events through the context seems like a good fit. The context’s lifetime naturally spans the entire command execution flow, each command has its own independent context, and the context itself is designed to support passing information across components.

The only remaining concern may be performance. At the moment, it does not appear likely to introduce any significant overhead, but we may still need to benchmark it after the implementation is complete before drawing a final conclusion.

For now, I think passing key events through the context is the best solution.

Aetherance: Also, if this PR should be split into several smaller PRs, should I first refactor the current PR to use the context-based approach you suggested, and then split it? Or should I implement this approach only after the PR has been split?

Aetherance: Hi @jihuayu,

I’ve implemented this refactoring locally and rerun the benchmarks. The performance issue I was previously concerned about does not appear to exist; in fact, this version performs slightly better than my previous implementation.

I also found that, with this design, simply adjusting where `notify` is triggered can significantly reduce the overall complexity and eliminate a considerable amount of unnecessary code, such as the guards for re-entrant execution in Lua scripts or `EXEC`.

However, I’m still not sure whether this approach would be acceptable, so I’d like to hear your thoughts before updating this PR. For now, I’ve opened a demonstration PR in my own repository with the local implementation:

https://github.com/Aetherance/kvrocks/pull/2

Do you think this approach would be acceptable? I’d like to get your feedback before proceeding further.

Aetherance: You can take a quick look at this implementation. If the approach looks good, we can continue with a more detailed review in this PR.

jihuayu: Let me have a see

Aetherance: I’ve added some comments to explain the refactored implementation. I hope they’re helpful.

jihuayu:
@Aetherance I think your new implementation looks good. You can merge the code in, and then we can have the other reviewers take a look as well.

Regarding stacked PRs, I checked that they’re indeed not supported yet. The official response says it’s on the roadmap github/gh-stack#46, so we can leave it aside for now.

Aetherance:
> @Aetherance I think your new implementation looks good. You can merge the code in, and then we can have the other reviewers take a look as well.
> 
> Regarding stacked PRs, I checked that they’re indeed not supported yet. The official response says it’s on the roadmap [github/gh-stack#46](https://github.com/github/gh-stack/issues/46), so we can leave it aside for now.

Got it! Thank you for your patience.

Aetherance:
Hi @jihuayu, sorry to bother you again. It looks like the CI needs maintainer approval before it can run. Could you please help run the CI when you have a chance?

jihuayu:
Hi @git-hulk. I think the new abstraction looks good. Could you take a look?

Aetherance:
Some CI checks failed, but judging from the error logs, the failures don't appear to have been introduced by this PR.

Aetherance:
Hi @git-hulk  @PragmaTwice , we’d love to hear your thoughts on this PR. Would you have time to take a look?

Aetherance:
Hi @git-hulk @PragmaTwice, gentle ping on this PR. Please let me know if any further changes are needed. Thanks!

git-hulk:
@Aetherance, I will have a look today. Thanks for your patience.

src/commands/cmd_string.cc
```diff
@@ -24,6 +24,7 @@
 
 #include "commander.h"
 #include "commands/command_parser.h"
+#include "common/string_util.h"
```
git-hulk: revert this line

src/server/server.cc
git-hulk: Should only do this while the db mapping is enabled.

git-hulk:
Overall looks correct. I will take another pass tomorrow. See if @jihuayu @PragmaTwice has more comments.

src/storage/storage.h
```diff
@@ -492,6 +503,35 @@ struct Context {
     return snapshot_;
   }
 
+  void EnableKeyspaceEventCollection(const std::string &ns, int notify_flags) {
+    if (!ShouldNotifyKeyspaceEvent(notify_flags, kNotifyAll)) return;
+    keyspace_event_ns_ = ns;
+    keyspace_event_notify_flags_ = notify_flags;
+  }
+
+  bool IsKeyspaceEventEnabled(int type_flag) const {
```
PragmaTwice: should `type_flag` be typed `int` or `NotifyKeyspaceEventFlag`?

src/storage/storage.h
```diff
@@ -492,6 +503,35 @@ struct Context {
...
+    return ShouldNotifyKeyspaceEvent(keyspace_event_notify_flags_, type_flag);
+  }
+
+  void AddKeyspaceEvent(int type_flag, std::string_view event, std::string_view key) {
+    if (!IsKeyspaceEventEnabled(type_flag)) return;
+    const int channel_flags = keyspace_event_notify_flags_ & (kNotifyKeyspace | kNotifyKeyevent);
+    keyspace_events_.emplace_back(KeyspaceEvent{
```
```suggestion
    keyspace_events_.emplace_back(
```

src/storage/storage.h
```diff
@@ -492,6 +503,35 @@ struct Context {
     return snapshot_;
   }
 
+  void EnableKeyspaceEventCollection(const std::string &ns, int notify_flags) {
+    if (!ShouldNotifyKeyspaceEvent(notify_flags, kNotifyAll)) return;
+    keyspace_event_ns_ = ns;
```
PragmaTwice: I'm not sure that we should put `keyspace_event_ns_` into `Context`. maybe it can just be a param of `AddKeyspaceEvent`.

src/storage/redis_db.cc
```diff
@@ -196,15 +205,23 @@ rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &k
...
   }
 
   if (*deleted_cnt == 0) return rocksdb::Status::OK();
 
-  return storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+  s = storage_->Write(ctx, storage_->DefaultWriteOptions(), batch->GetWriteBatch());
+  if (!s.ok()) return s;
+
+  for (const auto index : deleted_key_indexes) {
+    ctx.AddKeyspaceEvent(kNotifyGeneric, "del", std::string_view(keys[index].data(), keys[index].size()));
```
PragmaTwice:
```suggestion
    ctx.AddKeyspaceEvent(kNotifyGeneric, "del", keys[index].ToStringView()));
```

src/commands/cmd_string.cc
Aetherance: done

src/storage/storage.h
Aetherance: Thanks! It should be NotifyKeyspaceEventFlag

src/storage/storage.h
Aetherance: done

src/storage/redis_db.cc
Aetherance: done

src/storage/storage.h
Aetherance: Thanks! It would indeed be better to pass this as a parameter to AddKeyspaceEvent.

src/server/server.cc
Aetherance: This function formats a string with dual ns/DB semantics into either an ns or DB result, and the logic for choosing between ns and DB is already included in the function(It is formatted as a DB only when redis-databases is enabled.). The issue here was probably that the function name was not accurate enough. I’ve renamed it to FormatKeyspaceNotificationScope, which seems to better reflect the intended semantics.

src/common/keyspace_events.h
```diff
@@ -0,0 +1,53 @@
...
+
+struct KeyspaceEvent {
+  int channel_flags;
+  std::string event;
+  std::string ns;
+  std::string key;
+};
+
+// Parses notify-keyspace-events flags.
+Status ParseNotifyKeyspaceEventsFlags(const std::string &input, int *flags);
```
PragmaTwice:
```suggestion
StatusOr<int> ParseNotifyKeyspaceEventsFlags(std::string_view input);
```

src/common/keyspace_events.cc
```diff
@@ -0,0 +1,84 @@
...
+        return {Status::NotOK, fmt::format("unsupported notify-keyspace-events flag: '{}'", c)};
+    }
+  }
+
+  *flags = result;
+  return Status::OK();
+}
+
+namespace {
+std::string PercentEncode(const std::string &input) {
```
PragmaTwice: Is the encoding necessary? I wonder what we get/prevent from the encoding.

src/common/keyspace_events.cc
```diff
@@ -0,0 +1,84 @@
...
+}  // namespace
+
+std::string FormatKeyspaceNotificationScope(const std::string &ns, int redis_databases) {
+  if (ns == kDefaultNamespace) {
+    return "0";
+  }
+  if (redis_databases > 0 && ns.rfind(kDatabaseNamespacePrefix, 0) == 0) {
+    return ns.substr(strlen(kDatabaseNamespacePrefix));
+  }
+  return "ns:" + PercentEncode(ns);
```
PragmaTwice: maybe a good chance to discuss if `ns:` the best prefix here. cc @git-hulk @jihuayu

src/common/keyspace_events.cc
Aetherance: just a format, actually, we could also just use the original `ns`.

Aetherance: this is only to align with Redis.

src/common/keyspace_events.h
```diff
@@ -0,0 +1,53 @@
...
+
+// Flags for notify-keyspace-events, separate from RedisType.
+enum NotifyKeyspaceEventFlag {
+  kNotifyKeyspace = 1 << 0,  // K, keyspace channels
+  kNotifyKeyevent = 1 << 1,  // E, keyevent channels
+  kNotifyGeneric = 1 << 2,   // g, emits del
+  kNotifyString = 1 << 3,    // $, emits set
+  // A, supported data classes without K or E.
+  kNotifyAll = kNotifyGeneric | kNotifyString,
+};
```
PragmaTwice: I'm thinking if it's a good idea to split NotifyKeyspaceEventFlag into two enums, one for keyevent/keyspace and another for these notify types. Any suggestions? cc @git-hulk @jihuayu

src/common/keyspace_events.cc
PragmaTwice: > this is only to align with Redis.

Redis also has such %-prefixed escaping? I didn't check. though i think redis doesn't have namespace so here we can do what ever we want except breaking protocol / unable to parse.

src/common/keyspace_events.cc
Aetherance: This is essentially a choice between:

Not adding a prefix for namespaces, but adding one for DBs.
Not adding a prefix for DBs, but adding one for namespaces.

The latter requires encoding and some awkward logic, but makes keyspace notifications look the same as Redis. The former avoids this extra logic, but when DB mapping is enabled, it would display something like `__keyspace@db1__:<key>` instead of Redis’s `__keyspace@1__:<key>`.

src/common/keyspace_events.h
git-hulk: Agree, it would be better.

src/server/server.cc
```diff
@@ -478,6 +479,17 @@ int Server::PublishMessage(const std::string &channel, const std::string &msg) {
   return cnt;
 }
 
+void Server::NotifyKeyspaceEvent(int flags, const std::string &event, const std::string &ns, const std::string &key) {
```
git-hulk: Can we pass `KeyspaceEvent` as its parameter instead of flattening into four? Also, the same for AddKeyspaceEvent.

src/common/keyspace_events.cc
Aetherance: Redis does not use this kind of `%`-prefixed escaping. This is special handling for Kvrocks’s native `ns`.

Aetherance: I think the simplest approach here is to use `ns` directly, which would eliminate the need for formatting functions such as `PercentEncode`. The trade-off is that the resulting format would differ slightly from Redis. https://github.com/apache/kvrocks/pull/3541#discussion_r3829554559

src/common/keyspace_events.cc
PragmaTwice: ya for redis databases we should be compatible to redis. e.g. `__keyspace@0__:<key>` or `__keyspace@1__:<key>`. for namespaces (except default) now it's like `__keyspace@ns:my_namespace1__:<key>`. hmm it's fine to me. i was wondering if two `:` can cause some parsing problems.

src/common/keyspace_events.h
jihuayu: Yes, I agree

Aetherance: That makes sense. I will fix it soon.

src/server/server.cc
Aetherance: Thanks! It indeed looks better.

Aetherance: done

src/common/keyspace_events.h
Aetherance: done

src/common/keyspace_events.h
Aetherance: done

src/common/keyspace_events.cc
Aetherance: I’m wondering in what scenario this could happen. Within Kvrocks itself, it doesn’t seem like this issue can occur at the moment.

PragmaTwice: ya i'm wonder if some third party clients will do it. think about it again and I think maybe we can just remove the prefix. e.g. `__keyspace@0__:<key>` or `__keyspace@my_namespace1__:<key>`

src/storage/storage.h
```diff
@@ -492,6 +501,29 @@ struct Context {
...
+
+  bool IsKeyspaceEventEnabled(NotifyKeyspaceEventTypeFlag type_flag) const {
+    return ShouldNotifyKeyspaceEvent(keyspace_event_notify_flags_, type_flag);
+  }
+
+  void AddKeyspaceEvent(KeyspaceEvent event) {
+    if (!IsKeyspaceEventEnabled(event.type_flag)) return;
+    event.channel_flags = keyspace_event_notify_flags_ & (kNotifyKeyspace | kNotifyKeyevent);
+    keyspace_events_.emplace_back(std::move(event));
+  }
```
```suggestion
  template <typename ...Args>
  void AddKeyspaceEvent(Args && ...args) {
    if (!IsKeyspaceEventEnabled(event.type_flag)) return;
    event.channel_flags = keyspace_event_notify_flags_ & (kNotifyKeyspace | kNotifyKeyevent);
    keyspace_events_.emplace_back(std::forward<Args>(args)...);
  }
```

src/common/keyspace_events.cc
Aetherance: > we can just remove the prefix. e.g. __keyspace@0__:<key> or __keyspace@my_namespace1__:<key>

The problem is that if `ns` is a num, it could be confused with a `db`. Also, could this cause parsing issues, for example, an purely numeric `ns` being incorrectly parsed as a `db`?

git-hulk:
Generally looks good to me.

src/common/keyspace_events.cc
PragmaTwice: no i think there's no ambiguity. a kvrocks instance can only be either redis-db compatible or namespace enabled. there's no intermediate state. if you want to change it you need to restart the instance.

Aetherance: Got it, so shall we go with this approach? Maybe we should also get opinions from other reviewers? cc @git-hulk @jihuayu

src/storage/storage.h
Aetherance: I think this may be worth reconsidering: AddKeyspaceEvent needs to check event.type_flag to determine whether a keyspace notification should be triggered. If we use a variadic template, it seems that we won’t be able to access event.type_flag here.

src/common/keyspace_events.cc
jihuayu: No problem. I think this is OK

src/storage/storage.h
PragmaTwice: You can use a template like `Flag type_flag, Args && ...args`.

src/common/keyspace_events.h
```diff
@@ -0,0 +1,61 @@
...
+bool ShouldNotifyKeyspaceEvent(int notify_flags, NotifyKeyspaceEventTypeFlag type_flag);
+
+struct KeyspaceEvent {
+  KeyspaceEvent(NotifyKeyspaceEventTypeFlag type_flag, std::string_view event, std::string_view ns,
+                std::string_view key)
+      : type_flag(type_flag), event(event), ns(ns), key(key) {}
+
+  NotifyKeyspaceEventTypeFlag type_flag;
+  int channel_flags = 0;
+  std::string event;
```
jihuayu, line 51: It looks like we can infer the event type from the event itself. Why do we need to store both `event` and `eventType` here? 
Also, could `event` be an enum instead?

src/common/keyspace_events.h
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+#pragma once
+
+#include <string>
+#include <string_view>
+
+#include "status.h"
+
+enum NotifyKeyspaceEventChannelFlag {
```
PragmaTwice:
```suggestion
enum KeyspaceEventChannel {
```

src/common/keyspace_events.h
```diff
@@ -0,0 +1,61 @@
...
+  kNotifyKeyspace = 1 << 0,  // K, keyspace channels
+  kNotifyKeyevent = 1 << 1,  // E, keyevent channels
+};
+
+// Event type flags for notify-keyspace-events, separate from RedisType.
+enum NotifyKeyspaceEventTypeFlag {
```
PragmaTwice:
```suggestion
enum KeyspaceEventType {
```

src/common/keyspace_events.cc
Aetherance: done

src/common/keyspace_events.h
Aetherance: good idea.

src/storage/storage.h
Aetherance: I’m wondering whether this approach introduces two sources of truth for `flag`: the `flag` passed as the first argument and the one stored in the `Event` object. The former is used for the check, while the latter is what actually takes effect. Could this inconsistency cause any issues?

src/common/keyspace_events.h
jihuayu: It feels like the logic has become more complicated after the changes.

I checked the Redis source code, and they also keep the event type and name separate, which is consistent with your original design. Keeping the implementation aligned with Redis seems reasonable as well.

If we still want to keep the new strongly typed design, I think an abstraction like the following would be better:

```C++
struct KeyspaceEventName {
  std::string_view value;
  KeyspaceEventType event_type;

  constexpr bool ShouldNotify(int flags) const {
    return flags & event_type;
  }

};

namespace KeyspaceEventNames {

inline constexpr KeyspaceEventName Set{
    "set",
    KeyspaceEventClass::kString,
};

inline constexpr KeyspaceEventName Del{
    "del",
    KeyspaceEventClass::kGeneric,
};
}
```

We used `ctx.AddKeyspaceEventIfEnabled(KeyspaceEventNames::Set, ns, key);` and `event_name.value`.

This way, we won't need to update three different places whenever we add a new event name.


cc @PragmaTwice If you think we don't need this layer of strong typing, removing it would also be fine, since the Redis source code itself doesn't have such an abstraction.

Aetherance: This type seems a bit too heavyweight? For keyspace notifications, I wonder whether a lighter-weight type might be sufficient here.

jihuayu: Yes, this change feels a bit too heavy. Let’s revert it. Thank you!

Aetherance: done

src/storage/storage.h
Aetherance: Are there any drawbacks to passing the arguments directly here? It seems simpler while still allowing us to defer the construction of the `Event` object until it is actually needed. cc @git-hulk @jihuayu

jihuayu (dismissed):
Thanks!

Aetherance:
@jihuayu Thank you 🙏 I’d still appreciate hearing your thoughts on this, though. https://github.com/apache/kvrocks/pull/3541#discussion_r3850986012

src/storage/storage.h
Aetherance: To clarify, by “passing the arguments directly,” I mean an API like this:

```cpp
void AddKeyspaceEvent(
    KeyspaceEventType type_flag,
    std::string_view event,
    std::string_view ns,
    std::string_view key);
```

This is also the approach I’ve kept in the code for the time being.

src/common/keyspace_events.h
```diff
@@ -0,0 +1,61 @@
...
+  kNotifyKeyevent = 1 << 1,  // E, keyevent channels
+};
+
+// Event type flags for notify-keyspace-events, separate from RedisType.
+enum KeyspaceEventType {
+  kNotifyGeneric = 1 << 2,  // g, emits del
+  kNotifyString = 1 << 3,   // $, emits set
+  // A, supported data classes without K or E.
+  kNotifyAll = kNotifyGeneric | kNotifyString,
+};
```
PragmaTwice, line 42:
```suggestion
enum KeyspaceEventType {
  kNotifyGeneric = 1 << 0,  // g, emits del
  kNotifyString = 1 << 1,   // $, emits set
  // A, supported data classes without K or E.
  kNotifyAll = kNotifyGeneric | kNotifyString,
};
```

src/common/keyspace_events.cc
```diff
@@ -0,0 +1,67 @@
...
+#include <cstring>
+
+#include "config/config.h"
+#include "fmt/format.h"
+
+bool ShouldNotifyKeyspaceEvent(int notify_flags, KeyspaceEventType type_flag) {
+  return (notify_flags & type_flag) != 0 && (notify_flags & (kNotifyKeyspace | kNotifyKeyevent)) != 0;
+}
+
+StatusOr<int> ParseNotifyKeyspaceEventsFlags(std::string_view input) {
```
PragmaTwice:
```suggestion
StatusOr<std::pair<int, int>> ParseNotifyKeyspaceEventsFlags(std::string_view input) {
```

I think here we should split the flag into two part, one for KeyspaceEventChannel and another for KeyspaceEventType.

src/common/keyspace_events.cc
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+#include "keyspace_events.h"
+
+#include <cstring>
+
+#include "config/config.h"
+#include "fmt/format.h"
+
+bool ShouldNotifyKeyspaceEvent(int notify_flags, KeyspaceEventType type_flag) {
```
PragmaTwice: Same as above.

src/common/keyspace_events.h
```diff
@@ -0,0 +1,61 @@
...
+
+bool ShouldNotifyKeyspaceEvent(int notify_flags, KeyspaceEventType type_flag);
+
+struct KeyspaceEvent {
+  KeyspaceEvent(KeyspaceEventType type_flag, std::string_view event, KeyspaceEventChannel channel_flags,
+                std::string_view ns, std::string_view key)
+      : type_flag(type_flag), channel_flags(channel_flags), event(event), ns(ns), key(key) {}
+
+  KeyspaceEventType type_flag;
+  KeyspaceEventChannel channel_flags;
```
PragmaTwice, line 50: Just wondering if `channel_flags` is necessary. Could we just check the channel flags when publishing instead of storing them here?

src/config/config.h
```diff
@@ -219,6 +219,9 @@ struct Config {
   // Enable transactional mode in engine::Context
   bool txn_context_enabled = false;
 
+  // Parsed notify-keyspace-events flags.
+  int notify_keyspace_events = 0;
```
PragmaTwice: Same. I think we should store two integers here.

git-hulk (dismissed):

src/storage/redis_db.cc
```diff
@@ -186,6 +188,13 @@ rocksdb::Status Database::MDel(engine::Context &ctx, const std::vector<Slice> &k
   storage_->MultiGet(ctx, ctx.DefaultMultiGetOptions(), metadata_cf_handle_, slice_keys.size(), slice_keys.data(),
                      pin_values.data(), statuses.data());
 
+  const bool collect_del_events = ctx.IsKeyspaceEventEnabled(kNotifyGeneric);
+  std::vector<size_t> deleted_key_indexes;
+  if (collect_del_events) deleted_key_indexes.reserve(keys.size());
+
+  std::unordered_set<std::string_view> deleted_ns_keys;
+  const bool deduplicate_keys = keys.size() > 1;
+  if (deduplicate_keys) deleted_ns_keys.reserve(keys.size());
```
PragmaTwice, line 197: Looks a good optimization!

src/common/keyspace_events.h
Aetherance: This is mainly to preserve Redis transaction semantics: event notifications need to be deferred, while the decision of whether to emit them must be based on the configuration in effect when the events occur.

src/common/keyspace_events.cc
Aetherance: Thanks. And should we use an enum here instead of an int?

PragmaTwice: I don't have strong opinion here. Either is fine to me.

Aetherance: Thanks!

jihuayu (dismissed):

src/types/redis_string.cc
```diff
@@ -345,7 +345,13 @@ rocksdb::Status String::Set(engine::Context &ctx, const std::string &user_key, c
   metadata.expire = expire;
   metadata.Encode(&new_raw_value);
   new_raw_value.append(value);
-  return updateRawValue(ctx, ns_key, new_raw_value);
+  auto s = updateRawValue(ctx, ns_key, new_raw_value);
+  if (!s.ok()) return s;
+
+  if (ctx.IsKeyspaceEventEnabled(kNotifyString)) {
+    ctx.AddKeyspaceEvent(kNotifyString, "set", namespace_, user_key);
+  }
```
```suggestion
  ctx.AddKeyspaceEvent(kNotifyString, "set", namespace_, user_key);
```

Aetherance: In that case, we also rename `AddKeyspaceEvent` to something like `AddKeyspaceEventIfEnabled`?

Aetherance:
Hi @PragmaTwice , are there any further suggestions?

PragmaTwice (approved):

jihuayu (approved):

git-hulk (approved):

Aetherance:
Thanks for all reviewers' time and patience 🙏

jihuayu:
@Aetherance Thanks! If you'd like to continue working on the synchronization part, you can take a look at `WriteBatchLogData` in the WAL.

Aetherance:
> @Aetherance Thanks! If you'd like to continue working on the synchronization part, you can take a look at `WriteBatchLogData` in the WAL.

Thanks! For the follow-up PRs, I’ll open a tracking issue to track their progress.

PR #3549, opened 2026-07-08
feat(info): support FORMAT (TXT | JSON) for the INFO command

### LINKED ISSUE #2302: Add option to INFO to return a json payload
### Search before asking

### Motivation

try to monitor some specs from my service and would be better to be able to consume them in a format that's easier to consume rather than parsing strings like the kvrocks_exporter does.

### Solution

would be nice to expose somenthing like:
``` INFO `section` JSON/TXT```

### PR DESCRIPTION
**Pull request by nhancdt2602**
## Summary

Add an optional `FORMAT (TXT | JSON)` option to the `INFO` command:

```
INFO [<section> ...] [FORMAT (TXT | JSON)]
```

- `FORMAT TXT` (the default) keeps the existing Redis-compatible text output.
- `FORMAT JSON` emits the same data as a JSON object keyed by section.

```
127.0.0.1:6666> INFO server FORMAT JSON
{"Server":{"version":"unstable","tcp_port":6666,"process_id":123,...}}
```

Closes #2302.

## Changes

- `CommandInfo`: parse the optional `FORMAT` keyword, separating it from section names.
- `InfoEntry`: store each value as a typed `std::variant`, with `InfoEntry::ToString()` for the text form; `GetInfo` serializes to text or JSON.
- Add `TestInfoFormat` integration tests.

## Implementation Note

Each `INFO` value carries its original type in the `InfoEntry` variant `std::variant<std::string, int64_t, double, bool>`, captured in the constructors, so the collectors stay unchanged and each format renders the value at serialization time:

- **Text** (`InfoEntry::ToString`): booleans as `0/1`, numbers via `std::to_string`, strings verbatim — identical to the previous output.
- **JSON**: native types — numbers unquoted, booleans as `true/false`, strings quoted.

The `FORMAT` keyword is case-insensitive; an unknown or missing value returns `syntax error`.

### CODE REVIEW

**Comment by nhancdt2602** [comment]
@PragmaTwice Could you have a look on this PR? I'm looking for your review.

File: src/server/server.cc [hunk 1]
```diff
@@ -1548,25 +1548,64 @@ std::string Server::GetInfo(const std::string &ns, const std::vector<std::string
   };
 
   std::string info_str;
+  jsoncons::ojson json_obj;
```
PragmaTwice: Better to use `::json` instead.

src/server/server.cc
```diff
@@ -1548,25 +1548,64 @@ std::string Server::GetInfo(const std::string &ns, const std::vector<std::string
...
+                if constexpr (std::is_same_v<T, double>) {
+                  // Serialize via the same %f text form used above so the JSON number stays consistent
+                  // with the text output (and free of float-to-double widening noise).
+                  sec_obj[entry.name] = std::stod(std::to_string(v));
+                } else {
+                  // string -> JSON string, int64/uint64 -> JSON number, bool -> JSON true/false.
+                  sec_obj[entry.name] = v;
+                }
+              },
+              entry.val);
```
PragmaTwice: I don't think it's useful and it made the logic more complicated. Please consider to remove it.

src/server/server.cc
```diff
@@ -1548,25 +1548,64 @@ std::string Server::GetInfo(const std::string &ns, const std::vector<std::string
...
+                using T = std::decay_t<decltype(v)>;
+                if constexpr (std::is_same_v<T, std::string>) {
+                  return v;
+                } else if constexpr (std::is_same_v<T, bool>) {
+                  return v ? "1" : "0";
+                } else {
+                  return std::to_string(v);
+                }
+              },
+              entry.val);
```
PragmaTwice: This should be a method of `InfoEntry`.

src/server/server.h
```diff
@@ -263,15 +264,30 @@ class Server {
   int64_t GetLastBgsaveTime();
   std::string GetRoleInfo();
 
+  // An INFO entry holds its value with its original type in a variant, so each output format can
+  // render it appropriately: GetInfo emits the Redis-compatible text (e.g. a bool as 0/1, numbers via
+  // std::to_string) while FORMAT JSON emits the native JSON type (a bool as true/false, numbers
+  // unquoted). The type is captured here at construction, where it is statically known.
   struct InfoEntry {
+    using Value = std::variant<std::string, int64_t, uint64_t, double, bool>;
```
PragmaTwice: Is `uint64_t` useful here?

PragmaTwice (commented):
Thank you for your contribution.

Please read https://kvrocks.apache.org/community/contributing/#guidelines-for-ai-assisted-contributions to understand our AI policy.

src/server/server.cc
nhancdt2602: If we remove this logic, values of float type will widen because we are using float to compute `used_cpu_sys`, `used_cpu_user`,... Float value like `0.026695` will be printed as `0.026695000007748604` in json double format.

Is it an expected behavior? If so, I will remove the logic.

src/server/server.cc
nhancdt2602: Agree, I will add it.

src/server/server.h
nhancdt2602: Just for redundancy. Btw, to keep it simple, I will remove `uint64_t`

src/server/server.cc
nhancdt2602: Make sense.

src/server/server.cc
PragmaTwice: I'm fine with this. I think these JSON is usually parsed and processed by machines than human. And these two float numbers are exactly identical, just that in the former more precision is omitted.

nhancdt2602: Thank for clearing it up. I'll keep double value as is in the json format.

src/server/server.h
nhancdt2602: Removed it.

src/server/server.cc
nhancdt2602: Moved the logic into `ToString` method

src/server/server.cc
nhancdt2602: Resolved.

src/server/server.cc
nhancdt2602: Switched to `::json`

src/server/server.h
```diff
@@ -263,15 +264,40 @@ class Server {
...
+    InfoEntry(std::string name, const char *val) : name(std::move(name)), val(std::string(val)) {}
+    InfoEntry(std::string name, bool v) : name(std::move(name)), val(v) {}
+    // Floating-point values (incl. float, which widens to double) are stored as double.
+    InfoEntry(std::string name, double v) : name(std::move(name)), val(v) {}
+    // Integers (bool handled above) are stored as int64_t.
+    template <typename T, std::enable_if_t<std::is_integral_v<T> && !std::is_same_v<T, bool>, int> = 0>
+    InfoEntry(std::string name, T v) : name(std::move(name)), val(static_cast<int64_t>(v)) {}
+
+    // Redis-compatible text form: strings verbatim, booleans as 0/1, numbers via std::to_string.
+    std::string ToString() const {
```
```suggestion
    std::string ValueToString() const {
```

src/server/server.cc
```diff
@@ -1548,25 +1548,38 @@ std::string Server::GetInfo(const std::string &ns, const std::vector<std::string
...
+        else
+          info_str.append("\r\n");
 
-      info_str.append("# " + sec + "\r\n");
+        info_str.append("# " + sec + "\r\n");
 
-      for (const auto &entry : fn(this)) {
-        info_str.append(fmt::format("{}:{}\r\n", entry.name, entry.val));
+        for (const auto &entry : entries) {
+          info_str.append(fmt::format("{}:{}\r\n", entry.name, entry.ToString()));
```
```suggestion
          info_str.append(fmt::format("{}:{}\r\n", entry.name, entry.ValueToString()));
```

PragmaTwice (approved):

PR #3550, opened 2026-07-09
fix(rdb): reject over-large declared lengths when loading RDB payloads

git-hulk:
The RDB load path sized buffers from an attacker-declared 64-bit length before checking the stream actually holds that many bytes. A write-level client could send a `RESTORE` payload declaring a huge length and make the server allocate it up front and crash (OOM) — via both the plain string path (`read_string.resize(len)`) and the LZF path (`out_buf(len, 0)`).

Add `RdbStream::EnsureRemainingBytes()` and validate every length that drives a read or allocation against the bytes still remaining in the stream, so a value can never be declared larger than the input that carries it. The LZF decompressed size may legitimately exceed the remaining stream, so it is additionally capped against `proto-max-bulk-len`.

Add gocase regression tests for both the plain-string and LZF-encoded `RESTORE` payloads: the server now returns an error and stays up.

### DIFF
--- a/src/common/rdb_stream.cc
+++ b/src/common/rdb_stream.cc
@@ -20,6 +20,9 @@
 
 #include "rdb_stream.h"
 
+#include <filesystem>
+#include <system_error>
+
 #include "fmt/format.h"
 #include "vendor/crc64.h"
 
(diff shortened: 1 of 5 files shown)


src/common/rdb_stream.cc
```diff
@@ -52,6 +52,14 @@ Status RdbFileStream::Open() {
     return {Status::NotOK, fmt::format("failed to open rdb file: '{}': {}", file_name_, strerror(errno))};
   }
 
+  ifs_.seekg(0, std::ifstream::end);
+  auto file_size = ifs_.tellg();
+  if (file_size < 0) {
+    return {Status::NotOK, fmt::format("failed to get the size of rdb file: '{}': {}", file_name_, strerror(errno))};
+  }
+  ifs_.seekg(0, std::ifstream::beg);
+  file_size_ = static_cast<uint64_t>(file_size);
```
PragmaTwice: I think we can use `fs::file_size()` or `fstat()` to get the file size instead of `seekg`/`tellg`?

git-hulk: Yes, good catch. Sorry for totally missing this.

PragmaTwice (approved):

Issue #3436, opened 2026-04-09
Tracking: Hash Field Expiration (HFE) implementation


PragmaTwice:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Proposal: #3432

This issue is to track the implementation of HFE.

Task list:
- [x] #3444 
- [ ] Update existing hash read and write commands to correctly process and filter out expired fields
- [ ] Implement the `HLEN` fast-path algorithms and the `SCAN_AND_REPAIR` lazy deletion mechanism
- [ ] Modify the RocksDB compaction filter to safely drop expired subkeys without modifying metadata
- [ ] Introduce new field-level TTL commands including `HEXPIRE`, `HPERSIST`, `HPTTL`, and `HTTL`
- [ ] Implement a background thread and a dedicated index column family to actively clean up cold expired fields (optional)

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

nagisa-kunhah:
@PragmaTwice Hi, I'd like to help with the HFE work. I'm interested in taking the HLEN fast-path and SCAN_AND_REPAIR part of this task. Could you assign it to me?

PR #3554, opened 2026-07-11
feat(hash): add HSETEX and HGETEX support

PragmaTwice:
Adds Redis-compatible HSETEX and HGETEX commands on top of the hash field expiration encoding. Implements batched field/value and metadata transitions for persistent, expiring, expired, missing, and ghost fields, including FXX/FNX, KEEPTTL, PERSIST, and legacy-encoding rejection. Adds C++ unit and Go integration coverage for metadata invariants, command parsing, replication, restart, and concurrency.

Proposal: https://github.com/apache/kvrocks/discussions/3432
Tracking issue: https://github.com/apache/kvrocks/issues/3436

Assisted by GPT 5.5 Sol.

### DIFF
--- a/src/commands/cmd_hash.cc
+++ b/src/commands/cmd_hash.cc
@@ -257,23 +257,6 @@ Status HashCommandStatus(const rocksdb::Status &status) {
   return {Status::RedisExecErr, status.ToString()};
 }
 
-Status ProbeHashTypeForHGetEx(engine::Context &ctx, Server *srv, Connection *conn, const Slice &key, uint64_t now_ms) {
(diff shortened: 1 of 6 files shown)


jihuayu (approved):
Just a reminder: we need to clearly document this limitation.  https://github.com/apache/kvrocks/discussions/3432#discussioncomment-16972102

git-hulk (approved):
LGTM.

We now have many places need to maintain the metadata with the field state. Perphas we could export some dedicated functions to allow maintaining those behaviors, so that we don't need to take care of the state machine in every place.

PR #3556, opened 2026-07-12
fix(txn): discard partial writes from failed EXEC commands

### LINKED ISSUE #3555: MULTI/EXEC can commit partial writes from a command that returns a runtime error
### Search before asking

### Version

Unstable `567fcf6c26fc38367c1bda81ca0541c6ab861086` on Linux, using the default storage encoding. The reproduction only changes `rocksdb.write_options.write_batch_max_bytes` at runtime.

### Minimal reproduce step

The following was reproduced four times on the current unstable build. The value `180` makes the first internal field write fit in the transaction batch while the next one reaches the batch limit. It may need a small adjustment if the storage encoding or namespace configuration differs.

```text
CONFIG SET rocksdb.write_options.write_batch_max_bytes 0
DEL txhash
HSET txhash f1 old1 f2 old2 f3 old3
CONFIG SET rocksdb.write_options.write_batch_max_bytes 180

# Standalone control: this fails atomically and leaves every field unchanged.
HSET txhash f1 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA f2 BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB f3 CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
HMGET txhash f1 f2 f3

# The same HSET inside MULTI leaks one of its writes into the shared batch.
MULTI
HSET txhash f1 AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA f2 BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB f3 CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
SET after ok
HMGET txhash f1 f2 f3
EXEC

CONFIG SET rocksdb.write_options.write_batch_max_bytes 0
HMGET txhash f1 f2 f3
GET after
```

The standalone HSET returns:

```text
ERR Operation aborted: Memory limit reached
```

and its following HMGET returns `old1`, `old2`, `old3`.

Runtime errors should not abort the remaining queued commands or roll back earlier successful commands. An ordinary atomic command such as HSET must not leave only some of its field updates behind. EXEC should therefore return the HSET error, commit `SET after ok`, and both HMGET calls should return `old1`, `old2`, `old3`.

### What did you see instead?

EXEC returns:

```text
1) ERR Operation aborted: Memory limit reached
2) OK
3) old1, old2, CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
```

After EXEC, `GET after` returns `ok`, but HMGET still returns `old1`, `old2`, and the new `f3` value. The failed HSET processes fields in reverse order: its first Put succeeds, the next Put returns `MemoryLimit`, and the successful Put remains in the transaction-wide batch.

### Anything Else?

This is a transaction-wide issue rather than an HSET-specific one:

- [`Storage::BeginTxn`](https://github.com/apache/kvrocks/blob/567fcf6c26fc38367c1bda81ca0541c6ab861086/src/storage/storage.cc#L972-L984) creates one shared `WriteBatchWithIndex` for the whole EXEC.
- [`Storage::GetWriteBatchBase`](https://github.com/apache/kvrocks/blob/567fcf6c26fc38367c1bda81ca0541c6ab861086/src/storage/storage.cc#L1002-L1008) returns an observer to that shared batch in transaction mode, and [`Storage::Write`](https://github.com/apache/kvrocks/blob/567fcf6c26fc38367c1bda81ca0541c6ab861086/src/storage/storage.cc#L714-L720) returns OK without flushing it.
- A RocksDB batch operation that exceeds `max_bytes` rolls back only that individual operation. Earlier successful operations from the same Redis command remain in the shared batch.
- [`CommandExec`](https://github.com/apache/kvrocks/blob/567fcf6c26fc38367c1bda81ca0541c6ab861086/src/commands/cmd_txn.cc#L83-L90) correctly continues after runtime command errors, but it then commits those leftover operations because there is no per-command rollback boundary.

Many multi-record write paths are affected, including MSET/MSETEX/MSETNX, SADD/SREM, ZADD and range removals, list push/pop/trim, XADD/XTRIM, hash mutations, bitmap mutations, and other types. Depending on where the failure occurs, this can leave metadata inconsistent with subkeys, break ZSet's two-column-family index, or commit deletes from a failed trim/pop operation. `write_batch_max_bytes` is a deterministic trigger, but RocksDB read/write errors or decode/validation failures after earlier batch mutations have the same risk.

This is distinct from #2992, which handled errors from the final DB commit, and from #2554, which tracks transaction concurrency. The transaction-wide shared batch introduced for #1281 provides EXEC-level atomic commit, but it currently has no command-level savepoints.

A general fix likely needs savepoints at ordinary atomic command execution boundaries while the shared transaction batch is active: establish a savepoint before the command, `PopSavePoint()` after full success, and `RollbackToSavePoint()` on error. The boundary should cover `PutLogData` and associated index updates while preserving writes from previously successful commands. If rollback itself fails, the whole transaction batch should be discarded rather than committed.

This must not become an unconditional rollback around every outer command. Redis deliberately retains successful writes performed by an EVAL/Function before a later script runtime error. Nested `redis.call` commands still need their own atomic boundaries, but an outer script/function error must not roll back its already successful nested writes. For example, `MULTI; EVAL "redis.call('SET',KEYS[1],'v'); error('boom')" 1 scriptkey; EXEC` should return a script error while leaving `scriptkey=v`.

Regression tests should force a mid-command `MemoryLimit`, verify that earlier and later successful commands still commit, and verify that no value, metadata, secondary-index, or replication-log record from the failed atomic command survives. They should also protect the EVAL/Function behavior above so a command-level fix does not accidentally add Redis-incompatible script rollback.

advisedy:
## Summary

Fixes #3555.

Add command-level RocksDB savepoints during `EXEC` so a failed write command
cannot leave partial ops in the shared transaction batch.

## Changes

- Add `Storage::{Set,Pop,Rollback}TxnSavePoint` around the transaction write batch
- Apply savepoints in `Connection::ExecuteCommands` for `kCmdWrite` during `EXEC`
- Apply the same boundary for nested script write commands in `scripting.cc`
- Mark `EVAL` / `FCALL` with `kCmdSkipTxnSavepoint` so outer script errors do not
  roll back already successful nested writes
- On fatal savepoint errors, `EXEC` aborts the shared batch and returns errors for
  the queued command list

## Tests

Go cases in `tests/gocase/unit/multi/multi_test.go`:

1. Issue repro: a failed `HSET` under `write_batch_max_bytes` leaves no partial
   hash fields; the following `SET` still commits
2. An `EVAL` runtime error keeps a previously successful nested `SET`
3. A failed multi-field write via `redis.pcall` rolls back only that command; a
   later nested `SET` still commits

### DIFF
--- a/src/commands/cmd_function.cc
+++ b/src/commands/cmd_function.cc
@@ -121,8 +121,9 @@ uint64_t GenerateFunctionFlags(uint64_t flags, const std::vector<std::string> &a
 }
 
 uint64_t GenerateFCallFlags(uint64_t flags, const std::vector<std::string> &, const Config &config) {
+  flags |= kCmdSkipTxnSavepoint;
   if (!config.lua_strict_key_accessing) {
-    return flags | kCmdExclusive;
+    flags |= kCmdExclusive;
   }
 
   return flags;
--- a/src/commands/cmd_script.cc
+++ b/src/commands/cmd_script.cc
@@ -126,8 +126,9 @@ uint64_t GenerateScriptFlags(uint64_t flags, const std::vector<std::string> &arg
 }
 
 uint64_t GenerateEvalFlags(uint64_t flags, const std::vector<std::string> &, const Config &config) {
+  flags |= kCmdSkipTxnSavepoint;
   if (!config.lua_strict_key_accessing) {
-    return flags | kCmdExclusive;
+    flags |= kCmdExclusive;
   }
 
   return flags;
--- a/src/commands/cmd_txn.cc
+++ b/src/commands/cmd_txn.cc
@@ -78,16 +78,25 @@ class CommandExec : public Commander {
     }
 
     auto storage = srv->storage;
+    // If EXEC stops early, some commands may not have replies yet.
+    // Use the queued command count to keep the error array length correct.
+    auto command_count = conn->GetMultiExecCommands()->size();
     // Execute multi-exec commands
     conn->SetInExec();
     auto s = storage->BeginTxn();
     if (s.IsOK()) {
-      conn->ExecuteCommands(conn->GetMultiExecCommands());
+      s = conn->ExecuteCommands(conn->GetMultiExecCommands());
       // In Redis, errors happening after EXEC instead are not handled in a special way:
       // all the other commands will be executed even if some command fails during
       // the transaction.
       // So, if conn->IsMultiError(), the transaction should still be committed.
-      s = storage->CommitTxn();
+      // Only fatal internal errors make ExecuteCommands return a non-OK status
+      // (e.g. savepoint failures) and abort the shared batch.
+      if (s.IsOK()) {
+        s = storage->CommitTxn();
+      } else if (auto abort_s = storage->AbortTxn(); !abort_s.IsOK()) {
+        s = abort_s;
+      }
     }
 
     conn->ResetMultiExec();
@@ -96,7 +105,7 @@ class CommandExec : public Commander {
     if (s) {
       conn->Reply(Array(conn->GetQueuedReplies()));
     } else {
-      conn->Reply(Array(std::vector<std::string>(conn->GetQueuedReplies().size(), redis::Error(s))));
+      conn->Reply(Array(std::vector<std::string>(command_count, redis::Error(s))));
     }
 
     conn->ClearQueuedReplies();
--- a/src/commands/commander.h
+++ b/src/commands/commander.h
@@ -89,6 +89,8 @@ enum CommandFlags : uint64_t {
   kCmdAdmin = 1ULL << 11,
   // "skip-monitor" flag, for commands that should skip monitor feed
   kCmdSkipMonitor = 1ULL << 12,
+  // "skip-txn-savepoint" flag, for commands that should skip savepoint
+  kCmdSkipTxnSavepoint = 1ULL << 13,
 };
 
 enum class CommandCategory : uint8_t {
--- a/src/server/redis_connection.cc
+++ b/src/server/redis_connection.cc
@@ -113,7 +113,11 @@ void Connection::OnRead([[maybe_unused]] struct bufferevent *bev) {
     return;
   }
 
-  ExecuteCommands(req_.GetCommands());
+  s = ExecuteCommands(req_.GetCommands());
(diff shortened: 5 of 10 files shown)


advisedy:
Hi @PragmaTwice, this PR tries to fix #3555 with command-level savepoints during EXEC so a failed write does not leave partial ops in the shared batch.

Would you mind taking a look when you have time?
Thanks!

advisedy:
@PragmaTwice, Thanks for triggering CI.

The only failure is :

```
--- FAIL: TestSlaveLostMaster (15.59s)
    client.go:99: forwarding tcp stream stopped
    client.go:147: forward tcp stream failed, err: read tcp 127.0.0.1:38281->127.0.0.1:35050: read: connection reset by peer
```

This matches the known flake tracked in #3515 and is unrelated to this PR. 

The new test in `unit/multi` passed:

    ok  github.com/apache/kvrocks/tests/gocase/unit/multi  0.505s

Could you please re-run the failed jobs when you have a chance?

advisedy:
Hi @mapleFU , quick question — all CI checks pass, and I've added
Go tests for the new behavior (in tests/gocase/unit/multi/multi_test.go). 

Does that mean I can ignore the SonarCloud coverage failure here? Thanks!

mapleFU:
I'm quite busy these days, will review before 7.16 night

PragmaTwice:
> Does that mean I can ignore the SonarCloud coverage failure here? Thanks!

SonarCloud failure is not a blocker. Reviewer will decide if it's good to merge.

PR #3557, opened 2026-07-13
feat(info): report command statistics per namespace

### LINKED ISSUE #755: Metrics for namespace  level
### Search before asking

### Motivation

Ref: https://github.com/apache/incubator-kvrocks/discussions/726
We are using the namespace to separate data between users and it works well.
But the metrics from the info command stats didn't separate by namespace as well.
As a user, we want to know the speed of commands for each namespace.

### Solution

_No response_

nhancdt2602:
## Summary

This change tracks command calls, latency, and histograms per namespace. A namespaced connection sees its own numbers in INFO (Stats, CommandStats) and LATENCY HISTOGRAM; the admin/default namespace sees the aggregate across all namespaces.

Behavior:
- In namespace ns1, call `INFO commandstats` -> returns stats scoped by ns1 only
- In namespace admin, call `INFO commandstats` -> returns stats aggregated from all namespaces

Closes #755.

## Changes

- Per-namespace Stats kept in `ns_stats_` field in the Server class (guarded by a shared_mutex), created lazily via `GetOrCreateNamespaceStats`.
- Command hot path records into the connection's cached per-namespace Stats (no lock needed).
- `GetStatsInfo`, `GetCommandsStatsInfo`, `LATENCY HISTOGRAM` resolve their source by caller namespace, or `AggregateNamespaceStats()` for the admin/default view.
- Go integration test TestNamespaceStats.

## Implementation notes

- Stats such as command counts, ops-per-sec, commandstats, hist I/O bytes and keyspace hits/misses stay global for now. They're counted where the namespace isn't in scope (request parsing / storage layer). Open for follow-ups.

## Test
- Passed regression tests under `unit/namespace` module
- Added and passed tests in `unit/info` module

## AI disclosure

- Human: outlined the behaviors, interfaces scaffolding and finalized approaches.
- AI: implementation and writing tests

---

<a href="https://app.devin.ai/review/apache/kvrocks/pull/3557" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -104,6 +104,7 @@ class CommandNamespace : public Commander {
       WARN("New namespace: {} with token: {}, addr: {}, result: {}", args_[2], args_[3], conn->GetAddr(), s.Msg());
     } else if (args_.size() == 3 && sub_command == "del") {
       Status s = srv->GetNamespace()->Del(args_[2]);
+      if (s.IsOK()) srv->ClearNamespaceStats(args_[2]);
       *output = s.IsOK() ? redis::RESP_OK : redis::Error(s);
       WARN("Deleted namespace: {}, addr: {}, result: {}", args_[2], conn->GetAddr(), s.Msg());
     } else if (args_.size() == 2 && sub_command == "current") {
@@ -1722,16 +1723,23 @@ class CommandLatency : public Commander {
       return Status::OK();
     }
 
+    // Histograms are per-namespace: use the caller's namespace stats, or the aggregate for the
+    // admin/default namespace. Hold the shared_ptr for the duration of the response build.
+    auto stats_holder = conn->GetNamespace() == kDefaultNamespace
+                            ? srv->AggregateNamespaceStats()
+                            : srv->GetOrCreateNamespaceStats(conn->GetNamespace());
+    const Stats &cmd_stats = *stats_holder;
+
     std::vector<const std::pair<const std::string, CommandHistogram> *> target_histograms;
     if (args_.size() > 2) {
       for (size_t i = 2; i < args_.size(); i++) {
-        auto it = srv->stats.commands_histogram.find(util::ToLower(args_[i]));
-        if (it != srv->stats.commands_histogram.end() && it->second.calls > 0) {
+        auto it = cmd_stats.commands_histogram.find(util::ToLower(args_[i]));
+        if (it != cmd_stats.commands_histogram.end() && it->second.calls > 0) {
           target_histograms.push_back(&(*it));
         }
       }
     } else {
-      for (const auto &iter : srv->stats.commands_histogram) {
+      for (const auto &iter : cmd_stats.commands_histogram) {
         if (iter.second.calls > 0) {
           target_histograms.push_back(&iter);
         }
@@ -1750,8 +1758,8 @@ class CommandLatency : public Commander {
         if (cumulative == 0) continue;
 
         int64_t boundary = 0;
-        if (i < srv->stats.bucket_boundaries.size()) {
-          boundary = static_cast<int64_t>(srv->stats.bucket_boundaries[i]);
+        if (i < cmd_stats.bucket_boundaries.size()) {
+          boundary = static_cast<int64_t>(cmd_stats.bucket_boundaries[i]);
         } else {
           boundary = -1;
         }
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -1723,8 +1723,7 @@ class CommandLatency : public Commander {
       return Status::OK();
     }
 
-    // Histograms are per-namespace: use the caller's namespace stats, or the aggregate for the
-    // admin/default namespace. Hold the shared_ptr for the duration of the response build.
+    // Report the caller's namespace histogram; the admin/default namespace sees the aggregate.
     auto stats_holder = conn->GetNamespace() == kDefaultNamespace
                             ? srv->AggregateNamespaceStats()
                             : srv->GetOrCreateNamespaceStats(conn->GetNamespace());
@@ -1832,5 +1831,5 @@ REDIS_REGISTER_COMMANDS(
     MakeCmdAttr<CommandSST>("sst", -3, "write exclusive admin", 1, 1, 1),
     MakeCmdAttr<CommandFlushMemTable>("flushmemtable", -1, "exclusive write", NO_KEY),
     MakeCmdAttr<CommandFlushBlockCache>("flushblockcache", 1, "exclusive write", NO_KEY),
-    MakeCmdAttr<CommandLatency>("latency", -2, "read-only admin", NO_KEY), )
+    MakeCmdAttr<CommandLatency>("latency", -2, "read-only", NO_KEY), )
 }  // namespace redis
--- a/src/commands/cmd_server.cc
+++ b/src/commands/cmd_server.cc
@@ -104,7 +104,6 @@ class CommandNamespace : public Commander {
       WARN("New namespace: {} with token: {}, addr: {}, result: {}", args_[2], args_[3], conn->GetAddr(), s.Msg());
     } else if (args_.size() == 3 && sub_command == "del") {
       Status s = srv->GetNamespace()->Del(args_[2]);
-      if (s.IsOK()) srv->ClearNamespaceStats(args_[2]);
       *output = s.IsOK() ? redis::RESP_OK : redis::Error(s);
       WARN("Deleted namespace: {}, addr: {}, result: {}", args_[2], conn->GetAddr(), s.Msg());
     } else if (args_.size() == 2 && sub_command == "current") {
--- a/src/server/server.cc
+++ b/src/server/server.cc
@@ -66,21 +66,7 @@ Server::Server(engine::Storage *storage, Config *config)
       config_(config),
       namespace_(storage) {
   // init commands stats here to prevent concurrent insert, and cause core
-  auto commands = redis::CommandTable::GetOriginal();
-
-  for (const auto &iter : *commands) {
-    stats.commands_stats[iter.first].calls = 0;
-    stats.commands_stats[iter.first].latency = 0;
-
-    if (stats.bucket_boundaries.size() > 0) {
-      // NB: Extra index for the last bucket (Inf)
-      for (std::size_t i{0}; i <= stats.bucket_boundaries.size(); ++i) {
-        stats.commands_histogram[iter.first].buckets.push_back(std::make_unique<std::atomic<uint64_t>>(0));
-      }
-      stats.commands_histogram[iter.first].calls = 0;
-      stats.commands_histogram[iter.first].sum = 0;
-    }
-  }
+  initCommandStats(&stats);
 
   // init cursor_dict_
   cursor_dict_ = std::make_unique<CursorDictType>();
@@ -872,7 +858,18 @@ uint64_t Server::GetClientID() { return client_id_.fetch_add(1, std::memory_orde
 
 void Server::recordInstantaneousMetrics() {
   auto rocksdb_stats = storage->GetDB()->GetDBOptions().statistics;
-  stats.TrackInstantaneousMetric(STATS_METRIC_COMMAND, stats.total_calls);
+  // Sample each namespace's command metric, and feed the sum into the global metric so the
+  // admin/default view reports aggregate ops/sec without keeping a global command counter on the hot path.
+  uint64_t total_calls = 0;
+  {
+    std::shared_lock<std::shared_mutex> lock(ns_stats_mu_);
+    for (const auto &[ns, ns_stats] : ns_stats_) {
+      auto calls = ns_stats->total_calls.load();
+      ns_stats->TrackInstantaneousMetric(STATS_METRIC_COMMAND, calls);
+      total_calls += calls;
+    }
+  }
+  stats.TrackInstantaneousMetric(STATS_METRIC_COMMAND, total_calls);
   stats.TrackInstantaneousMetric(STATS_METRIC_NET_INPUT, stats.in_bytes);
   stats.TrackInstantaneousMetric(STATS_METRIC_NET_OUTPUT, stats.out_bytes);
   stats.TrackInstantaneousMetric(STATS_METRIC_ROCKSDB_PUT,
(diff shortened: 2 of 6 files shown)


nhancdt2602:
@PragmaTwice I'd love to hear your thought about this change. Especially, some stats still stay global in this PR for the sack of simplicity.

src/server/server.h
```diff
@@ -444,6 +450,13 @@ class Server {
 
   std::map<std::string, DBScanInfo> db_scan_infos_;
 
+  // Per-namespace command statistics (keyed by namespace name), guarded by ns_stats_mu_. The global
+  // `stats` keeps the non-namespaced counters (net bytes, replication) and the sampled aggregate ops/sec.
+  std::unordered_map<std::string, std::shared_ptr<Stats>> ns_stats_;
```
PragmaTwice, line 452: `shared_ptr` is only good for when dynamic lifetime is needed. But here is a counterexample.

nhancdt2602: The use of shared pointer here is intentional. As the stats object is held in the connection cache field, the on going connection might have referred to the stats object when it is removed from the map, i.e: deleted by remove namespace operation. The shared pointer simplify the cleanup process.

nhancdt2602:
@jihuayu Thanks for running the CI pipeline. The only failing test is:
```
--- FAIL: TestSlaveLostMaster (10.74s)
    client.go:99: forwarding tcp stream stopped
    client.go:147: forward tcp stream failed, err: read tcp 127.0.0.1:49515->127.0.0.1:49516: read: connection reset by peer
FAIL
exit status 1
```

This seems unrelated to the PR change. (as it was reported as flaky test in #3515). 

Could you kindly re-run the pipeline again?

src/server/server.h
PragmaTwice: Ahh got it.

src/server/server.cc
```diff
@@ -1392,11 +1389,82 @@ int64_t Server::GetLastBgsaveTime() {
...
+  }
+
+  std::unique_lock<std::shared_mutex> lock(ns_stats_mu_);
+  if (auto it = ns_stats_.find(ns); it != ns_stats_.end()) {
+    return it->second;
+  }
+  auto ns_stats = std::make_shared<Stats>(config_->histogram_bucket_boundaries);
+  initCommandStats(ns_stats.get());
+  ns_stats_[ns] = ns_stats;
+  return ns_stats;
```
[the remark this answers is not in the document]
nhancdt2602: `histogram_bucket_boundaries` is a read-only config, so it can't be changed at runtime.

src/server/server.cc
```diff
@@ -1392,11 +1389,82 @@ int64_t Server::GetLastBgsaveTime() {
...
+    }
+    for (const auto &[cmd, hist] : ns_stats->commands_histogram) {
+      auto &agg_hist = agg->commands_histogram[cmd];
+      agg_hist.calls.fetch_add(hist.calls.load(), std::memory_order_relaxed);
+      agg_hist.sum.fetch_add(hist.sum.load(), std::memory_order_relaxed);
+      for (std::size_t i = 0; i < hist.buckets.size(); ++i) {
+        agg_hist.buckets[i]->fetch_add(hist.buckets[i]->load(), std::memory_order_relaxed);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
nhancdt2602: `histogram_bucket_boundaries` is a read-only config, so it can't be changed at runtime.

src/server/server.cc
```diff
@@ -872,7 +858,18 @@ uint64_t Server::GetClientID() { return client_id_.fetch_add(1, std::memory_orde
...
+  uint64_t total_calls = 0;
+  {
+    std::shared_lock<std::shared_mutex> lock(ns_stats_mu_);
+    for (const auto &[ns, ns_stats] : ns_stats_) {
+      auto calls = ns_stats->total_calls.load();
+      ns_stats->TrackInstantaneousMetric(STATS_METRIC_COMMAND, calls);
+      total_calls += calls;
+    }
+  }
+  stats.TrackInstantaneousMetric(STATS_METRIC_COMMAND, total_calls);
```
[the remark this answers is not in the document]
nhancdt2602: As this method is not called too frequently, I believe the overhead is acceptable.

src/server/server.cc
```diff
@@ -1392,11 +1389,82 @@ int64_t Server::GetLastBgsaveTime() {
...
   Server::InfoEntries entries;
   entries.emplace_back("total_connections_received", total_clients_.load());
-  entries.emplace_back("total_commands_processed", stats.total_calls.load());
-  entries.emplace_back("instantaneous_ops_per_sec", stats.GetInstantaneousMetric(STATS_METRIC_COMMAND));
+  entries.emplace_back("total_commands_processed", cmd_stats.total_calls.load());
+  // Per-namespace ops/sec comes from the namespace's own sampled metric; the admin/default view uses
+  // the global metric, which the sampler feeds with the sum across all namespaces.
+  auto ops_per_sec = ns == kDefaultNamespace ? stats.GetInstantaneousMetric(STATS_METRIC_COMMAND)
+                                             : cmd_stats.GetInstantaneousMetric(STATS_METRIC_COMMAND);
+  entries.emplace_back("instantaneous_ops_per_sec", ops_per_sec);
```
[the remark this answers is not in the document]
nhancdt2602: As the aggregation logic is fast, I believe we can ignore the concern here.

jihuayu (commented):
Hello @nhancdt2602.Thank you for your contribution. I think this PR still has a few issues.

1. latency command is admin only. It will make the namespace user could not get it's latency. https://github.com/apache/kvrocks/blob/634ed15dfd9249cfb57a52541dddba05b377df0a/src/commands/cmd_server.cc#L1827
2. Deleting a namespace does not disconnect existing connections. This causes the old connections’ `ns_stats_` references to diverge.

```text
After authentication:
ns_stats_["ns1"] -> Stats A
Active connection -> Stats A

After deleting `ns1`:
ns_stats_ no longer contains ns1
Active connection -> Stats A  // kept alive by shared_ptr

After the old connection executes `INFO`:
ns_stats_["ns1"] -> newly created Stats B
Active connection -> Stats A

```
3. Deleting a namespace may cause an underflow here.
https://github.com/apache/kvrocks/blob/8c26549809823b342532acc0d52f4871ecba49e6/src/stats/stats.cc#L107-L117

nhancdt2602:
@jihuayu Thanks for your reply. To fix points 2 and 3, can I keep the stats entry in the namespace map even after the namespace is deleted? This would be consistent with the current behavior of `db_scan_infos_`

nhancdt2602:
@jihuayu Regarding point 1, I’m still not sure I understand your concern. Are you suggesting that we should not expose latency and histogram metrics for namespace-scoped connections?

jihuayu:
@nhancdt2602 Sorry, I’m late.

> @jihuayu Thanks for your reply. To fix points 2 and 3, can I keep the stats entry in the namespace map even after the namespace is deleted? This would be consistent with the current behavior of db_scan_infos_

I have no objection to that, since deleting a namespace is a low-frequency operation. 


What I meant is that this command should not be restricted to ADMIN users. Namespace users are not ADMIN users, so they would be unable to retrieve latency information.

nhancdt2602:
> @nhancdt2602 Sorry, I’m late.
> 
> > @jihuayu Thanks for your reply. To fix points 2 and 3, can I keep the stats entry in the namespace map even after the namespace is deleted? This would be consistent with the current behavior of db_scan_infos_
> 
> I have no objection to that, since deleting a namespace is a low-frequency operation.
> 
> What I meant is that this command should not be restricted to ADMIN users. Namespace users are not ADMIN users, so they would be unable to retrieve latency information.

@jihuayu Regarding point 1, LATENCY command already is scoped to ADMIN before this PR. If I'm not mistaken, you want to relax it by removing admin flag in LATENCY command config in this PR?

nhancdt2602:
@jihuayu I've addressed all three points from your review. Looking for your feedback when you I have time.

1. `latency command is admin only` -> remove admin tag from LATENCY command
2. `Deleting a namespace does not disconnect existing connections. This causes the old connections’ ns_stats_ references to diverge.` -> keep the stats entry in the map without deleting it.
3. `Deleting a namespace may cause an underflow here` -> solved by not deleting stats from the map.

nhancdt2602:
@jihuayu Could you re-run the flaky test in CI pipeline again. All other tests pass.

git-hulk:
@nhancdt2602 It seems only the calls/latency are related to the namespace, so I'm wondering if it's better to extend the existing functions in `stats`:

```C++
void IncrCalls(const string &ns, const string &cmd) {
  stats->total_calls++;
  if (ns != kDefaultNamespace) {
     stats->ns_command_calls[ns]++;
  }
}
```

so that we don't need to intercepting the stats and namespace in serval places. cc @jihuayu @PragmaTwice

nhancdt2602:
> @nhancdt2602 It seems only the calls/latency are related to the namespace, so I'm wondering if it's better to extend the existing functions in `stats`:
> 
> ```c++
> void IncrCalls(const string &ns, const string &cmd) {
>   stats->total_calls++;
>   if (ns != kDefaultNamespace) {
>      stats->ns_command_calls[ns]++;
>   }
> }
> ```
> 
> so that we don't need to intercepting the stats and namespace in serval places. cc @jihuayu @PragmaTwice

Thanks @git-hulk for your review. I saw your point about moving the namespace resolution logic into stats fields/methods like `IncrCall`, `IncrLatency` or `commands_histogram`.  The reason I resolve the namespace at the caller layer rather than inside `IncrCalls/IncrLatency` is to keep the command hot path lock-free.

I want to keep the overhead of adding namespace telemetry information as minimal as possible. By keeping the namespace stats pointer when a connection is established, we are free of lock when accessing the namespace stat entry.

That said, if you and project members prefer consolidating into Stats, accepting a per-command lock. I'm happy to refactor that way; just let me know the preferred trade-off.

I'd love to hear your opinion about this ^^

git-hulk:
> > @nhancdt2602 It seems only the calls/latency are related to the namespace, so I'm wondering if it's better to extend the existing functions in `stats`:
> > ```c++
> > void IncrCalls(const string &ns, const string &cmd) {
> >   stats->total_calls++;
> >   if (ns != kDefaultNamespace) {
> >      stats->ns_command_calls[ns]++;
> >   }
> > }
> > ```
> >    
> >   
> > so that we don't need to intercepting the stats and namespace in serval places. cc @jihuayu @PragmaTwice
> 
> Thanks @git-hulk for your review. I saw your point about moving the namespace resolution logic into stats fields/methods like `IncrCall`, `IncrLatency` or `commands_histogram`. The reason I resolve the namespace at the caller layer rather than inside `IncrCalls/IncrLatency` is to keep the command hot path lock-free.
> 
> I want to keep the overhead of adding namespace telemetry information as minimal as possible. By keeping the namespace stats pointer when a connection is established, we are free of lock when accessing the namespace stat entry.
> 
> That said, if you and project members prefer consolidating into Stats, accepting a per-command lock. I'm happy to refactor that way; just let me know the preferred trade-off.
> 
> I'd love to hear your opinion about this ^^

@nhancdt2602 I'm not sure how much performance would be impacted. Consolidating into one stats would be more consistent from my personal perspective since only a few metrics are related to the namespace.

nhancdt2602:
@git-hulk I can relate your preference. From my side, I lean toward keeping the stat object scoped to namespace level and move the stat entry resolution to caller layer. 

However, I'm happy to apply your suggestion if it's the direction of the project. Shall I consolidate them or wait for others opinion ^^

git-hulk:
> @git-hulk I can relate your preference. From my side, I lean toward keeping the stat object scoped to namespace level and move the stat entry resolution to caller layer.
> 
> However, I'm happy to apply your suggestion if it's the direction of the project. Shall I consolidate them or wait for others opinion ^^

We can wait for a while to see if other maintainers have any comments.

jihuayu:
> That said, if you and project members prefer consolidating into Stats, accepting a per-command lock. I'm happy to refactor that way; just let me know the preferred trade-off.


@nhancdt2602 Do you mean the lock in this piece of code?
image

https://github.com/apache/kvrocks/pull/3557/changes#diff-0782ead7a805bf491f44cf0bbbab901ba0ae3f18bc57181dd6e58a8b5c594171R1409-R1424

nhancdt2602:
> > That said, if you and project members prefer consolidating into Stats, accepting a per-command lock. I'm happy to refactor that way; just let me know the preferred trade-off.
> 
> 
> 
> 
> 
> @nhancdt2602 Do you mean the lock in this piece of code?
> 
> image
> 
> 
> 
> https://github.com/apache/kvrocks/pull/3557/changes#diff-0782ead7a805bf491f44cf0bbbab901ba0ae3f18bc57181dd6e58a8b5c594171R1409-R1424

@jihuayu Exactly. Or do you have any better ideas?

nhancdt2602:
@jihuayu just a friendly ping What's your opinion about the above concern? And shall I refactor in the way git-hulk suggested or keep it as is?

jihuayu:
@nhancdt2602 Sorry, I’ve been on a business trip recently and have been quite busy. I’ll take some time to look into this issue when I get a chance.

src/server/server.cc
```diff
@@ -1392,11 +1389,75 @@ int64_t Server::GetLastBgsaveTime() {
...
+      }
+    }
+  }
+  return agg;
+}
+
+Server::InfoEntries Server::GetStatsInfo(const std::string &ns) {
+  // Command stats are per namespace; the admin/default namespace sees the aggregate across all of them.
+  auto cmd_stats_ptr = ns == kDefaultNamespace ? AggregateNamespaceStats() : GetOrCreateNamespaceStats(ns);
+  const Stats &cmd_stats = *cmd_stats_ptr;
```
jihuayu, line 1453: 🟡 **Command statistics become per-database when Redis SELECT compatibility mode is enabled**

Command counters are looked up by the connection's namespace (`GetStatsInfo(ns)`/`GetCommandsStatsInfo(ns)` at `src/server/server.cc:1450-1487`), and in SELECT-compatibility mode each database is its own namespace, so a client that selected database 1 or higher only sees the commands issued on that database instead of the whole server.
Impact: Monitoring tools that select a non-zero database get partial, Redis-incompatible command statistics and latency histograms.

<details>
<summary>How database indexes become namespaces and split the stats</summary>

When `redis-databases > 0`, `CommandSelect` maps db 0 to `kDefaultNamespace` and db N>0 to `"db" + N` and calls `conn->SetNamespace(ns)` (`src/commands/cmd_server.cc:227-232`). `Server::GetInfo` passes `conn->GetNamespace()` (`src/commands/cmd_server.cc:296`) down to `GetStatsInfo`/`GetCommandsStatsInfo`, which now resolve a per-namespace `Stats` unless the namespace is the default one. `CommandLatency::getHistogram` (`src/commands/cmd_server.cc:1725-1729`) does the same. Consequently, after `SELECT 3`, `INFO commandstats`, `total_commands_processed`, `instantaneous_ops_per_sec` and `LATENCY HISTOGRAM` all report only db3's activity, whereas real Redis reports these server-wide regardless of the selected database.

A possible fix is to treat all `kDatabaseNamespacePrefix` namespaces (i.e. when `redis_databases > 0`) as the aggregate/default view.

</details>

<details>
<summary>Prompt for agents</summary>

```
In SELECT-compatibility mode (config redis_databases > 0), Connection namespaces are synthesized per database ("db1", "db2", ... with db0 mapped to kDefaultNamespace, see CommandSelect in src/commands/cmd_server.cc). Because Server::GetStatsInfo, Server::GetCommandsStatsInfo and CommandLatency::getHistogram now select the stats source by namespace, a client that issued SELECT 1 sees only that database's command counters, latency and histograms, while real Redis reports them server-wide irrespective of the selected DB. Consider resolving the aggregate view when config_->redis_databases > 0 (or when the namespace has the kDatabaseNamespacePrefix form), so SELECT-mode clients keep the server-wide view.
```

</details>


<a href="https://app.devin.ai/review/apache/kvrocks/pull/3557" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

src/server/server.cc
```diff
@@ -1392,11 +1389,75 @@ int64_t Server::GetLastBgsaveTime() {
...
+      auto &agg_hist = agg->commands_histogram[cmd];
+      agg_hist.calls.fetch_add(hist.calls.load(), std::memory_order_relaxed);
+      agg_hist.sum.fetch_add(hist.sum.load(), std::memory_order_relaxed);
+      for (std::size_t i = 0; i < hist.buckets.size(); ++i) {
+        agg_hist.buckets[i]->fetch_add(hist.buckets[i]->load(), std::memory_order_relaxed);
+      }
+    }
+  }
+  return agg;
+}
```
jihuayu, line 1448: 🟡 **Every INFO or latency query from an admin connection rebuilds a full statistics snapshot from scratch**

A complete statistics object for every registered command is allocated and zero-initialized (`AggregateNamespaceStats()` at `src/server/server.cc:1427-1448`) on each admin-scoped INFO or latency request, so routine monitoring polls do thousands of small allocations per call.
Impact: Frequent monitoring of the server does noticeably more CPU and allocation work than before, and requesting several sections at once repeats the whole computation.

<details>
<summary>Cost breakdown and duplicated work</summary>

`AggregateNamespaceStats()` calls `initCommandStats()` (`src/server/server.cc:1392-1407`), which inserts an entry for every command in `CommandTable::GetOriginal()` into two `std::map`s, and — when `histogram-bucket-boundaries` is configured — heap-allocates `boundaries+1` `std::atomic<uint64_t>` per command (≈250 commands × (N+1) `make_unique` calls). It then sums every namespace's maps.

This runs once for `INFO stats` and again for `INFO commandstats` (`src/server/server.cc:1452` and `src/server/server.cc:1485`), so `INFO all` builds the aggregate twice, and `LATENCY HISTOGRAM` (`src/commands/cmd_server.cc:1726-1728`) builds a third one. Caching the aggregate per request, or summing directly into the output without materializing a full `Stats`, would avoid this.

</details>


<a href="https://app.devin.ai/review/apache/kvrocks/pull/3557" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

src/commands/cmd_server.cc
```diff
@@ -1824,5 +1830,5 @@ REDIS_REGISTER_COMMANDS(
     MakeCmdAttr<CommandSST>("sst", -3, "write exclusive admin", 1, 1, 1),
     MakeCmdAttr<CommandFlushMemTable>("flushmemtable", -1, "exclusive write", NO_KEY),
     MakeCmdAttr<CommandFlushBlockCache>("flushblockcache", 1, "exclusive write", NO_KEY),
-    MakeCmdAttr<CommandLatency>("latency", -2, "read-only admin", NO_KEY), )
+    MakeCmdAttr<CommandLatency>("latency", -2, "read-only", NO_KEY), )
```
jihuayu, line 1833: 🟨 **LATENCY command is no longer admin-only, exposing per-namespace latency data to any authenticated user**

The `admin` flag was removed from the LATENCY command registration (`src/commands/cmd_server.cc:1833`), so any authenticated namespace user can now run `LATENCY HISTOGRAM` and `LATENCY RESET`. The histogram is scoped to the caller's namespace (`src/commands/cmd_server.cc:1725-1729`), so cross-namespace data is not leaked to non-default users, but this is still a widening of the command's access control surface that was previously restricted to administrators.


<a href="https://app.devin.ai/review/apache/kvrocks/pull/3557" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>


---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

jihuayu (commented):
Don't look this

nhancdt2602:
@jihuayu just a quick reminder, any updates on this?

PR #3560, opened 2026-07-14
feat(server): support the client-output-buffer-limit configuration

git-hulk:
Add the client-output-buffer-limit configuration to disconnect clients that are not reading data from the server fast enough, following the same semantics as Redis: a client is disconnected immediately once the hard limit is reached, or when it stays over the soft limit for more than the configured seconds continuously.

Unlike Redis, all the classes are configured in a single line, and only the specified classes are changed. The limits are disabled by default for all classes of clients, e.g. to protect against slow Pub/Sub subscribers with the same limits as Redis:

```
client-output-buffer-limit pubsub 32m 8m 60
```

The limit takes effect on the normal and pubsub classes since the replication stream is written to the socket directly instead of going through the connection output buffer. The slave class is accepted for compatibility but takes no effect: slow replicas are still handled by max-replication-lag and replication-send-timeout-ms.

The check runs every time data is appended to the connection output buffer. A client over the limit is closed asynchronously by manually triggering the deferred write callback, since the write callback of a stuck client may never fire on its own; the callback then runs in the owner worker's event loop where freeing the connection is safe.

Also expose the client_output_buffer_limit_disconnections counter in INFO stats, and classify clients subscribed to only shard channels as pubsub clients so that they are covered by the pubsub class as well.

Part of #2284

Assistant-By Claude Fable 5

### DIFF
--- a/src/config/config.cc
+++ b/src/config/config.cc
@@ -180,6 +180,8 @@ Config::Config() {
       {"timeout", false, new IntField(&timeout, 0, 0, INT_MAX)},
       {"tcp-backlog", true, new IntField(&backlog, 511, 0, INT_MAX)},
       {"maxclients", false, new IntField(&maxclients, 10240, 0, INT_MAX)},
+      {"client-output-buffer-limit", false,
+       new StringField(&client_output_buffer_limit_str_, "normal 0 0 0 slave 0 0 0 pubsub 0 0 0")},
       {"max-backup-to-keep", false, new IntField(&max_backup_to_keep, 1, 0, 1)},
       {"max-backup-keep-hours", false, new IntField(&max_backup_keep_hours, 0, 0, INT_MAX)},
       {"master-use-repl-port", false, new YesNoField(&master_use_repl_port, false)},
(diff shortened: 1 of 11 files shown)


src/config/config.cc
```diff
@@ -444,6 +446,47 @@ void Config::initFieldValidator() {
...
+    if (util::EqualICase(args[i], "normal")) {
+      kind = ClientKind::kNormal;
+    } else if (util::EqualICase(args[i], "slave") || util::EqualICase(args[i], "replica")) {
+      kind = ClientKind::kSlave;
+    } else if (util::EqualICase(args[i], "pubsub")) {
+      kind = ClientKind::kPubsub;
+    } else {
+      return {Status::NotOK, fmt::format("unknown client kind '{}'", args[i])};
+    }
+    auto hard = GET_OR_RET(ParseSizeAndUnit(args[i + 1]).Prefixed("invalid hard limit"));
```
jihuayu, line 476: It looks a little different from Redis: https://github.com/redis/redis/blob/e5e1eaa97ad3020305b84442197011abc8ce40bc/redis.conf#L8-L18

src/server/redis_connection.cc
```diff
@@ -93,11 +93,60 @@ std::string Connection::ToString() {
...
+  uint64_t hard_limit_bytes = limit.hard_limit_bytes.load(std::memory_order_relaxed);
+  uint64_t soft_limit_bytes = limit.soft_limit_bytes.load(std::memory_order_relaxed);
+  if (hard_limit_bytes == 0 && soft_limit_bytes == 0) return false;
+
+  uint64_t used_bytes = evbuffer_get_length(Output());
+  if (hard_limit_bytes != 0 && used_bytes >= hard_limit_bytes) return true;
+
+  if (soft_limit_bytes != 0) {
+    if (used_bytes >= soft_limit_bytes) {
+      auto soft_limit_seconds = static_cast<int64_t>(limit.soft_limit_seconds.load(std::memory_order_relaxed));
```
jihuayu: Can we use the same type for limit.soft_limit_seconds and soft_limit_seconds to avoid overflow?

jihuayu (commented):
The async close is not guaranteed to invoke `Connection::OnWrite`, because blocking commands replace the bufferevent callbacks.

For example, with:

    client-output-buffer-limit pubsub 1m 0 0

a RESP3 client can run:

    HELLO 3
    SUBSCRIBE ch
    BLPOP never-exists 0

and another client can publish an 8 MiB message to `ch`.

I reproduced this on this PR. `BlockingCommander::StartBlocking()` had replaced the write callback, so `bufferevent_trigger(..., EV_WRITE, ...)` invoked `BlockingCommander::OnWrite()` instead of `Connection::OnWrite()`. Since the list was still empty, it disabled `EV_WRITE` and returned without checking `kCloseAsync`.

After 12 seconds the client was still present in `CLIENT LIST` with:

    cmd=blpop obuf=8388658

and `client_output_buffer_limit_disconnections` was already 1. The connection was only removed after the test client closed its socket. A non-blocked subscriber was closed normally in the control test.

src/config/config.cc
git-hulk: Yes, but we don't handle this in this PR.

git-hulk:
> The async close is not guaranteed to invoke `Connection::OnWrite`, because blocking commands replace the bufferevent callbacks.
> 
> For example, with:
> 
> ```
> client-output-buffer-limit pubsub 1m 0 0
> ```
> 
> a RESP3 client can run:
> 
> ```
> HELLO 3
> SUBSCRIBE ch
> BLPOP never-exists 0
> ```
> 
> and another client can publish an 8 MiB message to `ch`.
> 
> I reproduced this on this PR. `BlockingCommander::StartBlocking()` had replaced the write callback, so `bufferevent_trigger(..., EV_WRITE, ...)` invoked `BlockingCommander::OnWrite()` instead of `Connection::OnWrite()`. Since the list was still empty, it disabled `EV_WRITE` and returned without checking `kCloseAsync`.
> 
> After 12 seconds the client was still present in `CLIENT LIST` with:
> 
> ```
> cmd=blpop obuf=8388658
> ```
> 
> and `client_output_buffer_limit_disconnections` was already 1. The connection was only removed after the test client closed its socket. A non-blocked subscriber was closed normally in the control test.

@jihuayu Yes, good catch.

jihuayu (commented):
@git-hulk There are still several issues:

1. `WAIT` also replaces the bufferevent callback, but this case was missed in the current changes.
2. The soft limit is not checked proactively after it expires. We seem to be missing a cron task that actively closes the connection; instead, the connection is closed lazily. Is this intentional by design?
3. The return value of the `PUBLISH` command that triggers the connection closure is incorrect. The current PR returns `0`, while Redis 8.6.3 returns `1`. The message has already been added to the subscriber’s buffer, and the connection is only scheduled to be closed afterward, so the subscriber should still be counted as a receiver.
4. Redis does not require the soft limit to be lower than the hard limit. For example, Redis accepts `CONFIG SET client-output-buffer-limit "normal 1mb 2mb 10"`, while this PR rejects it.

git-hulk:
@jihuayu I think we don't need to fix everything that pops up from AI reviews.


> 1. `WAIT` also replaces the bufferevent callback, but this case was missed in the current changes.

It's just a wait for the replication, and should be reasonable to keep it without checking the output buffer on every OnWrite.

> 2. The soft limit is not checked proactively after it expires. We seem to be missing a cron task that actively closes the connection; instead, the connection is closed lazily. Is this intentional by design?

Yes, the lazy behavior is intentional behavior.

> 4. Redis does not require the soft limit to be lower than the hard limit. For example, Redis accepts CONFIG SET client-output-buffer-limit "normal 1mb 2mb 10", while this PR rejects it.

I prefer to keep rejecting the config explicitly instead of silence-and-ignore in Redis.

jihuayu (approved):
All right! LGTM

PragmaTwice (approved):

Issue #3527, opened 2026-06-16
test(go): optimize replication test execution time


Tangruilin:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

replication test cost about 120s now

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Issue #3524, opened 2026-06-16
Track Issue: Optimize the execution time of go test cases


Tangruilin:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

In current GitHub Action workflows, go test execution now often takes more than 30 minutes, which may affect our responsiveness and PR processing efficiency.

We need to analyze which test cases take too long and optimize their execution time without affecting the test results.

This PR is track for：
- #2642 

Include: 
- #3507 
- #3526
- #3527
- #3528
- #3529

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #3561, opened 2026-07-18
perf(test): poll is_compacting instead of a fixed sleep in replication test

ishita-0301:
### What this does

`TestReplicationShareCheckpoint` runs `COMPACT` and then blindly
`time.Sleep(time.Second)` before starting the replicas. That is slower
than necessary and potentially flaky — on a loaded CI runner one second
may not be enough for the manual compaction to finish, so the test can
proceed while compaction is still running.

This replaces the fixed sleep with the same `is_compacting` poll already
used elsewhere in the same file (`TestReplicationWithHole`), returning as
soon as compaction completes:

```go
    return util.FindInfoEntry(masterClient, "is_compacting") == "no"
}, 10*time.Second, 100*time.Millisecond)
```

### Related

Part of #3524, addresses #3527.

### Note on testing

I was not able to run the full Go integration suite locally for this
change; it mirrors an existing, already-tested pattern in the same file
rather than introducing new behavior. Happy to adjust if maintainers
prefer a different approach.

### DIFF
--- a/tests/gocase/integration/replication/replication_test.go
+++ b/tests/gocase/integration/replication/replication_test.go
@@ -389,7 +389,9 @@ func TestReplicationShareCheckpoint(t *testing.T) {
 	ctx := context.Background()
 	require.NoError(t, masterClient.Set(ctx, "a", "b", 0).Err())
 	require.NoError(t, masterClient.Do(ctx, "compact").Err())
-	time.Sleep(time.Second)
+	require.Eventually(t, func() bool {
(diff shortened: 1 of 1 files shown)


jihuayu:
Hi @ishita-0301. If you can't run the tests locally, you can use the GitHub Actions workflow in your repository: [https://github.com/ishita-0301/kvrocks/actions](https://github.com/ishita-0301/kvrocks/actions) to make sure all tests are passing.

jihuayu (approved):

git-hulk (approved):

PR #3562, opened 2026-07-20
perf(search): share a thread-local RNG for HNSW and drop seed (#2398)

### LINKED ISSUE #2398: Avoid `HnswIndex` heavy construction because of `mt19937`
### Search before asking

### Motivation

As mentioned in PR #2368 in this [comment](https://github.com/apache/kvrocks/pull/2368#discussion_r1670432215), 
`HnswIndex` is a relatively heavy class since it includes `mt19937, sizeof(mt19937) ~= 5000)`, we can try to optimize its construction. 

Thanks @PragmaTwice for pointing it out! :)

### Solution

One way I could think of to improve it is to make all instances of HnswIndex share a single `mt19937` instance. 

We can implement a `RandomGenerator` using Singleton Pattern as below:

```
class RandomGenerator {
public:
    static RandomGenerator& getInstance() {
        static RandomGenerator instance;
        return instance;
    }

    std::mt19937& getGenerator() {
        return generator_;
    }

private:
    RandomGenerator() {
        std::random_device rd;
        generator_ = std::mt19937(rd());
    }
    // delete copy & move constructor
    std::mt19937 generator_;
};

struct HnswIndex {
    HnswIndex(...) {
        generator_ = &RandomGenerator::getInstance().getGenerator();
    }
};
```
One callout is that `mt19937` is not thread-safe, so I'm not quite sure if sharing it among all `HnswIndex` instances will cause any side effects, will further investigate if we finally use the solution. 

Welcome any other ideas and thoughts! 

advisedy:
Fixes #2398.

### Problem

`HnswIndex` currently owns a `std::mt19937` generator directly, so every
`HnswIndex` construction also constructs the random generator.

However, the generator is only needed by `RandomizeLayer()` when inserting
vector entries. `HnswIndex` can also be constructed on query paths such as
KNN/RANGE search, where random layer generation is not used.

### Fix

- Replace the eager `std::mt19937` member with a lazily initialized
  `std::unique_ptr<std::mt19937>`.
- Initialize the generator only on the first `RandomizeLayer()` call.

### Tests

new C++ gtests added (hnsw_index_test.cc):

- `RandomGeneratorIsInitializedLazily`

### AI-assisted contribution

I used an LLM to help draft the new GTest case for verifying lazy generator
initialization.

### DIFF
--- a/src/search/hnsw_indexer.cc
+++ b/src/search/hnsw_indexer.cc
@@ -182,7 +182,8 @@ HnswIndex::HnswIndex(const SearchKey& search_key, HnswVectorFieldMetadata* vecto
 
 uint16_t HnswIndex::RandomizeLayer() {
   if (!generator) {
-    generator = std::make_unique<std::mt19937>(seed.value_or(std::random_device()()));
+    const auto actual_seed = seed ? *seed : std::random_device()();
+    generator = std::make_unique<std::mt19937>(actual_seed);
   }
   std::uniform_real_distribution<double> level_dist(0.0, 1.0);
   double r = level_dist(*generator);
--- a/src/search/hnsw_indexer.cc
+++ b/src/search/hnsw_indexer.cc
@@ -171,16 +171,15 @@ StatusOr<double> ComputeSimilarity(const VectorItem& left, const VectorItem& rig
   }
 }
 
-HnswIndex::HnswIndex(const SearchKey& search_key, HnswVectorFieldMetadata* vector, engine::Storage* storage,
-                     std::optional<std::random_device::result_type> seed)
+thread_local std::mt19937 HnswIndex::generator{std::random_device()()};
+
+HnswIndex::HnswIndex(const SearchKey& search_key, HnswVectorFieldMetadata* vector, engine::Storage* storage)
     : search_key(search_key),
       metadata(vector),
       storage(storage),
-      seed(seed),
       m_level_normalization_factor(1.0 / std::log(metadata->m)) {}
 
 uint16_t HnswIndex::RandomizeLayer() {
-  static thread_local std::mt19937 generator = [this] { return std::mt19937(seed.value_or(std::random_device()())); }();
   std::uniform_real_distribution<double> level_dist(0.0, 1.0);
   double r = level_dist(generator);
   double log_val = -std::log(r);
--- a/src/search/hnsw_indexer.cc
+++ b/src/search/hnsw_indexer.cc
@@ -173,16 +173,19 @@ StatusOr<double> ComputeSimilarity(const VectorItem& left, const VectorItem& rig
 }
 
 HnswIndex::HnswIndex(const SearchKey& search_key, HnswVectorFieldMetadata* vector, engine::Storage* storage,
-                     std::random_device::result_type seed)
+                     std::optional<std::random_device::result_type> seed)
     : search_key(search_key),
       metadata(vector),
       storage(storage),
-      generator(std::mt19937(seed)),
+      seed(seed),
       m_level_normalization_factor(1.0 / std::log(metadata->m)) {}
 
 uint16_t HnswIndex::RandomizeLayer() {
+  if (!generator) {
+    generator = std::make_unique<std::mt19937>(seed.value_or(std::random_device()()));
+  }
   std::uniform_real_distribution<double> level_dist(0.0, 1.0);
-  double r = level_dist(generator);
+  double r = level_dist(*generator);
   double log_val = -std::log(r);
   double layer_val = log_val * m_level_normalization_factor;
   return static_cast<uint16_t>(std::floor(layer_val));
--- a/src/search/hnsw_indexer.cc
+++ b/src/search/hnsw_indexer.cc
@@ -179,7 +179,7 @@ HnswIndex::HnswIndex(const SearchKey& search_key, HnswVectorFieldMetadata* vecto
       storage(storage),
       m_level_normalization_factor(1.0 / std::log(metadata->m)) {}
 
-uint16_t HnswIndex::RandomizeLayer() {
+uint16_t HnswIndex::RandomizeLayer() const {
   std::uniform_real_distribution<double> level_dist(0.0, 1.0);
   double r = level_dist(generator);
   double log_val = -std::log(r);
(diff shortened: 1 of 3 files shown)


jihuayu (dismissed):
Thanks! LGTM

advisedy:
Hi, @git-hulk , @PragmaTwice ,just friendly ping on this one too, whenever you have some time. Thanks!

src/search/hnsw_indexer.cc
```diff
@@ -173,16 +173,20 @@ StatusOr<double> ComputeSimilarity(const VectorItem& left, const VectorItem& rig
...
       metadata(vector),
       storage(storage),
-      generator(std::mt19937(seed)),
+      seed(seed),
       m_level_normalization_factor(1.0 / std::log(metadata->m)) {}
 
 uint16_t HnswIndex::RandomizeLayer() {
+  if (!generator) {
+    const auto actual_seed = seed ? *seed : std::random_device()();
+    generator = std::make_unique<std::mt19937>(actual_seed);
```
git-hulk: From the original issue, it should expect to keep only one random generator for all HNSW indexes? @PragmaTwice

PragmaTwice: Yup. maybe thread local to avoid race.

PragmaTwice (changes requested):
Refer to the issue, I think it's better to make the RNG static.

src/search/hnsw_indexer.cc
advisedy: switched to a single `static thread_local` generator  as you suggested. please take a look.

advisedy:
I was being silly and misunderstood，Now fixed: switched to a single `static thread_local` generator  as you suggested. Updated in the latest push, please take a look @PragmaTwice .

src/search/hnsw_indexer.cc
```diff
@@ -173,14 +172,15 @@ StatusOr<double> ComputeSimilarity(const VectorItem& left, const VectorItem& rig
...
+                     std::optional<std::random_device::result_type> seed)
     : search_key(search_key),
       metadata(vector),
       storage(storage),
-      generator(std::mt19937(seed)),
+      seed(seed),
       m_level_normalization_factor(1.0 / std::log(metadata->m)) {}
 
 uint16_t HnswIndex::RandomizeLayer() {
+  static thread_local std::mt19937 generator = [this] { return std::mt19937(seed.value_or(std::random_device()())); }();
```
PragmaTwice: 1. Please don't make it a local static variable;
2. The semantics of `seed` is very weird here.

advisedy: You're right — the semantics of `seed` is indeed weird here. Since the
  generator is now shared, only the first instance's seed would take
  effect, which is misleading.

  I see two possible directions:

  1. Share the generator only among indexes with the same seed, and let
     indexes with different seeds keep their own. I haven't fully figured
     out the implementation yet, but maybe a `static thread_local` map
     could work? It seems like that might be doable.

  2. Drop `seed` entirely, seed the generator with `std::random_device`,
     and make it a class-level `static thread_local std::mt19937` member
     (no longer a function-local static), so its initialization no longer
     depends on any specific instance.

  I'd like to know which direction you think we should take. If you
  prefer option 1, I'll think about how to implement it in detail. If
  it's option 2, I'll go ahead and do it now.

PragmaTwice: Could you stop copy-pasting to me what your agent replies to you? You can do your own decision. I have enough token that I don't need an agent proxy.

advisedy: Sorry, my english is not good, so I tell AI what I was thinking and let it translate. But The decision is mine. please.

advisedy: Here is my origin words: 确实感觉seed有点奇怪在这里，但是generator得根据seed去生成出来，不如说这里seed，generator，和我们期望共享的逻辑有点混乱，我的理解有两种思路，第一种思路就是我们考虑对于seed一样的进行共享，不一样的就是各自持有，这种方法用static + thread_local + map 应该能实现出来？ 但我感觉有点麻烦。第二种思路就是或许可以考虑直接把seed删掉，使用std::random_device，这样的话，static thread_local std::mt19937 generator; 放到类里面，不再是局部了，这样初始化也不会依赖某个实例。 想知道一下，你觉得我们应该怎么做会比较好呢？如果你想要第一种的思路，我再去思考具体该怎么实现会比较好。如果是第二种的话，我可以现在就去实现出来代码。

advisedy: I’m very sorry that I relied too much on AI when replying, but my genuine intention was to contribute. I’m truly sorry.

PragmaTwice: 1. I think we can just remove `seed` here.
2. After the change, don't forget to rewrite the PR title and description to match the current implementation.

jihuayu:
Hi @advisedy.

There is nothing wrong with using AI to translate your writing into English, but do not let it turn a simple point into something unnecessarily long and bloated.

No one wants to feel as though they are talking to an AI agent. Copying and pasting an AI-generated response without editing it shows little respect for other people’s time. You need to think through what the AI gives you, incorporate your own understanding, and express it in your own words.

We are not short on tokens, but we do not need robots either. We are not looking for contributors who are only here to meet a KPI. We want contributors to grow with the community and remain part of it over the long term. That requires you to spend real time and effort thinking about and understanding the project, rather than relying entirely on AI to complete your tasks.

It is okay if you do not do everything well at the beginning, so there is no need to worry too much. I hope you can improve over time.

PragmaTwice (commented):
The code is good now. To get merged you need to check [why the lint failed](https://github.com/apache/kvrocks/actions/runs/30027117635/job/89373071612?pr=3562) and fix it (just to add a `const` specifier).

advisedy:
Hi @jihuayu . Thank you for taking the time to leave this comment. I’ll write replies in my own words instead of relying too much on AI. I really appreciate your comment.

advisedy:
Could someone help re-run the failed job? Thanks!

PragmaTwice (approved):

advisedy:
Looks ready to merge. Could someone help merge it? Thansk！

git-hulk (approved):

PR #3564, opened 2026-07-21
fix(kvrocks2redis): translate FLUSHDB/FLUSHALL via DeleteRange (#2177)

### LINKED ISSUE #2177: kvrocks2redis can't sync flush commands
### Search before asking

### Version

unstable 70fac378e9f966f8209c5195cdd2aaa214a87a0e Fri Feb

### Minimal reproduce step

image

master -> slave -> kvrocks2redis -> target

1、master flush
2、aof only write one del command

write flushcommand to aof file

### What did you see instead?

aof only write one del command

### Anything Else?

_No response_

advisedy:
Fixes #2177.

### Problem

`FLUSHDB` / `FLUSHALL` is implemented via `DeleteRange` on the metadata
CF, but the replication path did not propagate it:

- `WriteBatchExtractor::DeleteRangeCF` was a no-op, so the delete was
  silently dropped and never translated into a Redis command on replicas
  or kvrocks2redis.
- `Database::FlushAll` merged all namespaces into a single `DeleteRange`,
  which could not be mapped back to any specific namespace.

### Fix

- `FlushAll` now emits one `DeleteRange` per namespace in the same batch.
- `DeleteRangeCF` translates those ranges into a per-namespace `FLUSHDB`,
  and rejects unrecognized ranges with a warning.

### Tests

- Go integration test `TestReplicationFlushDBAcrossNamespaces`:
  reproduces the issue with `repl-namespace-enabled=yes`.
- C++ gtests in `batch_extractor_test.cc` for single-namespace `FLUSHDB`,
  multi-namespace `FLUSHALL`, and unrecognized ranges.

### AI-assisted contribution

The tests were generated by an LLM. I reviewed all of them to make sure
they are correct and meaningful.

### Scope

Only covers plain `FLUSHDB` / `FLUSHALL`. The `MULTI`/`EXEC` case
mentioned in #2177 is not handled here — I'm not sure what kind of
approach would fit best, and would appreciate reviewer input.

### DIFF
--- a/src/storage/batch_extractor.cc
+++ b/src/storage/batch_extractor.cc
@@ -20,7 +20,11 @@
 
 #include "batch_extractor.h"
 
+#include <cstdint>
+#include <optional>
+
 #include "cluster/redis_slot.h"
+#include "common/string_util.h"
 #include "logging.h"
 #include "parse_util.h"
 #include "server/redis_reply.h"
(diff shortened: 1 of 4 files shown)


advisedy:
Hi, @git-hulk , @PragmaTwice ,just friendly ping on this one, whenever you have some time. Thanks!

Issue #3552, opened 2026-07-10
Tracking Issues: Support Cuckoo Filter data structure and commands


nagisa-kunhah:
### Search before asking

- [x] I had searched in the [issues](https://github.com/apache/kvrocks/issues) and found no similar issues.

### Motivation

Redis Stack supports Cuckoo Filter as a probabilistic data structure, and Kvrocks can support it as part of RedisBloom compatibility.

### Tracking

- [x] CF.RESERVE command: #3481 
- [x] CF.ADD command: #3481 
- [ ] CF.ADDNX command @nhancdt2602 
- [ ] CF.EXISTS/CF.MEXISTS command: #3467 
- [ ] CF.COUNT command @ChaseForward 
- [ ] CF.DEL command : https://github.com/apache/kvrocks/pull/3567
- [ ] CF.INFO command @ChaseForward 
- [ ] CF.INSERT command @geetanshjuneja 
- [ ] CF.INSERTNX command @geetanshjuneja 
- [ ] CF.SCANDUMP command
- [ ] CF.LOADCHUNK command （base on CF.SCANDUMP）

### Solution 
- #3530 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

ChaseForward:
Hi @nagisa-kunhah, I'd like to work on CF.INFO and CF.COUNT
if no one else is already on them. These would be my first
open-source contribution — I'll make sure to read the
contributor guide before starting. Thanks!

nagisa-kunhah:
Hi @ChaseForward , feel free to take them and @ me with any questions.

ChaseForward:
Thanks, @nagisa-kunhah! I'll check the issues and try to contribute. If I run into any problems, I'll ask you. Looking forward to contributing.

nhancdt2602:
@nagisa-kunhah Hello, I'm interested in this feature and would love to work on the `CF.ADDNX command`. Could I take it?

nagisa-kunhah:
> [@nagisa-kunhah](https://github.com/nagisa-kunhah) Hello, I'm interested in this feature and would love to work on the `CF.ADDNX command`. Could I take it?

@nhancdt2602 Sure, just go ahead.

nhancdt2602:
> > [@nagisa-kunhah](https://github.com/nagisa-kunhah) Hello, I'm interested in this feature and would love to work on the `CF.ADDNX command`. Could I take it?
> 
[...]

PR #3567, opened 2026-07-26
feat(CuckooFilter): add cf.del command

nagisa-kunhah:
part of: #3552 

 ## Summary

Add RedisBloom-compatible `CF.DEL key item` support for Cuckoo Filter.

`CF.DEL` deletes one matching fingerprint occurrence and returns `1` when a slot is cleared, or `0` when the key/item
is not found. Duplicate inserts require duplicate deletes, matching RedisBloom behavior.

## Design

The command layer adds `cf.del` as a write command and delegates deletion to `CuckooChain::Delete`.

Deletion loads the Cuckoo Filter metadata, hashes the item, generates the fingerprint, then searches sub-filters from
newest to oldest. The first matching slot in either candidate bucket is cleared, and only one occurrence is removed.

After a successful delete, metadata size is decremented and `num_deleted_items` is incremented. When the chain has
more than one sub-filter and accumulated deletes exceed 10% of the remaining item count, an internal compact pass is
triggered. Compact tries to move fingerprints from newer sub-filters into older ones, removes fully compacted latest
sub-filters, and deletes their persisted page keys to avoid stale data if the chain expands again later.

Page key construction is shared through small Cuckoo page helpers so compact cleanup uses the same encoding as normal
page access.

This pr is This PR was written using codex and GPT-5.5

### DIFF
--- a/src/types/redis_cuckoo_chain.h
+++ b/src/types/redis_cuckoo_chain.h
@@ -20,9 +20,6 @@
 
 #pragma once
 
-#include <memory>
-#include <vector>
-
 #include "cuckoo_filter.h"
 #include "storage/redis_db.h"
 #include "storage/redis_metadata.h"
@@ -54,8 +51,6 @@ class CuckooChain : public Database {
   rocksdb::Status Delete(engine::Context &ctx, const Slice &user_key, const Slice &item, bool *deleted);
 
  private:
-  using CuckooSubFilters = std::vector<std::unique_ptr<CuckooSubFilter>>;
-
   // Loads metadata for a cuckoo filter key.
   rocksdb::Status getCuckooChainMetadata(engine::Context &ctx, const Slice &ns_key, CuckooChainMetadata *metadata);
 
@@ -70,16 +65,8 @@ class CuckooChain : public Database {
                                              bool *inserted);
   rocksdb::Status commitSubFilterAndMetadata(engine::Context &ctx, const Slice &user_key, const std::string &ns_key,
                                              CuckooChainMetadata *metadata, CuckooSubFilter *sub_filter);
-  rocksdb::Status buildSubFilters(engine::Context &ctx, const std::string &ns_key, const CuckooChainMetadata &metadata,
-                                  CuckooSubFilters *sub_filters);
-  rocksdb::Status compactCuckooChain(CuckooChainMetadata *metadata, CuckooSubFilters *sub_filters,
-                                     std::vector<uint16_t> *freed_filter_indexes, bool cont);
-  rocksdb::Status compactSingleSubFilter(uint16_t source_index, CuckooSubFilters *sub_filters, bool *fully_compacted);
-  rocksdb::Status deleteSubFilterPages(rocksdb::WriteBatchBase *batch, const std::string &ns_key,
-                                       const CuckooChainMetadata &metadata, uint16_t filter_index);
   rocksdb::Status commitDelete(engine::Context &ctx, const Slice &user_key, const std::string &ns_key,
-                               CuckooChainMetadata *metadata, CuckooSubFilters *sub_filters,
-                               const std::vector<uint16_t> &freed_filter_indexes);
+                               CuckooChainMetadata *metadata, CuckooSubFilter *sub_filter);
 };
 
 }  // namespace redis
--- a/src/types/redis_cuckoo_chain.h
+++ b/src/types/redis_cuckoo_chain.h
@@ -20,6 +20,9 @@
 
 #pragma once
 
+#include <memory>
+#include <vector>
+
 #include "cuckoo_filter.h"
 #include "storage/redis_db.h"
 #include "storage/redis_metadata.h"
@@ -47,7 +50,12 @@ class CuckooChain : public Database {
   // Duplicate items are allowed, so added is true whenever insertion succeeds.
   rocksdb::Status Add(engine::Context &ctx, const Slice &user_key, const Slice &item, bool *added);
 
+  // Deletes one matching fingerprint from the cuckoo filter.
+  rocksdb::Status Delete(engine::Context &ctx, const Slice &user_key, const Slice &item, bool *deleted);
+
  private:
+  using CuckooSubFilters = std::vector<std::unique_ptr<CuckooSubFilter>>;
+
   // Loads metadata for a cuckoo filter key.
   rocksdb::Status getCuckooChainMetadata(engine::Context &ctx, const Slice &ns_key, CuckooChainMetadata *metadata);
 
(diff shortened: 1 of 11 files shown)


nagisa-kunhah:
Hi @jihuayu, while implementing `CF.DEL`, I found that the automatic compaction path may introduce serious worker-blocking and OOM risks. Currently, it scans all logical buckets of a sub‑filter and loads every page into `CuckooPageCache`, including zero‑filled pages for missing keys. Thus, a large sparse sub‑filter can make a single `CF.DEL` perform work and allocate memory proportional to its logical capacity.

We seek feedback on whether compaction should be part of this PR. Two options:

1. **Land `CF.DEL` without automatic compaction for now.**  
   Only remove fingerprint and update metadata. Semantics are preserved, but deleted slots and extra sub‑filters remain, causing space and lookup amplification.

2. **Move compaction to bounded, incremental background maintenance.**  
   Use Kvrocks’ `TaskRunner` (used by `AsyncCompactDB`, `AsyncScanDBSize`, etc.) with cursor‑based scanning. Process a bounded number of actually stored pages per slice, commit a bounded write batch, release cache and locks, and requeue if more work remains. Requires actual‑page iteration, bounded cache/batch, deduplication, synchronization, and crash‑safe progress tracking.

jihuayu:
@nagisa-kunhah Let's leave this out for now. We can implement background compaction in a follow-up PR. Thanks.

PS: when the key does not exist, RedisBloom returns a `Not found` error, while this PR currently returns success.

nagisa-kunhah:
# CF.DEL Background Compaction

`CF.DEL` must not compact a complete logical sub-filter on the request path. A large sparse filter can have very few RocksDB pages but a huge logical page range, so a full logical scan can block a worker and materialize excessive page-cache memory.

## Compaction

For the current tail `Fn`, compact oldest targets first:

```text
Fn -> F0
remaining Fn -> F1
...
remaining Fn -> F(n - 1)
verify Fn
```

For each non-zero fingerprint in `Fn`, try direct insertion into the two candidate buckets of the current target. If insertion succeeds, write the target fingerprint and clear the source fingerprint in one `WriteBatch`. Do not use Cuckoo kick-out relocation during compaction.

Scan `Fn` through its persisted page-key prefix, so sparse filters do not trigger reads for absent logical pages. Delete a source page key when its page becomes all zero.

One `Fn -> Fi` pass holds the existing logical-key lock. It may flush bounded page batches while keeping that lock. After a pass completes, release the lock, then start the next target pass. If `Fn` is empty after all targets were tried, decrement `n_filters`. If that exposes another tail, continue with it in the same task.

## Task execution and deduplication

A successful `CF.DEL` schedules compact work when the chain has more than one sub-filter and delete debt exceeds the compact threshold. The task carries:

| Field | Description |
|---|---|
| `namespace` and `user_key` | Reconstruct the Cuckoo Filter in the background callback. |
| `ns_key` | Lock key and task-key input. |
| `metadata_version` | Reject a key that was deleted and recreated before the task starts. |

`TaskRunner` should gain a generic keyed API:

```cpp
StatusOr<bool> TryPublishUnique(std::string task_key, Task task);
```

The compact task key is:

```text
cuckoo-compact:<ns_key>:<metadata_version>
```

`TaskRunner` keeps active task keys in memory. The same key is queued only once while it is queued or running. It removes the key when the callback returns, rolls it back if queueing fails, and clears all keys when `Join()` clears the queue.

The callback locks `ns_key`, rereads metadata, and exits if `metadata_version` no longer matches. A restart drops queued work and active task keys. It does not resume compact work. A later qualifying `CF.DEL` can schedule a new best-effort task.

## Trade-offs

- A full `Fn -> Fi` pass can delay foreground writes for that key.
- One tail is scanned once per older target, and target bucket reads can be random.
- `CF.ADD` between target passes can make the current cycle less effective, but cannot make tail removal unsafe because the final empty check is locked.
- `TaskRunner` is shared and single-threaded by default, so long compact tasks can delay other background work.

nagisa-kunhah:
Hi @jihuayu , sorry for the delay. This PR is ready for review. I've outlined some draft ideas on the background compaction above — would really appreciate your advice when you have time. Thanks!

tests/cppunit/types/cuckoo_filter_test.cc
```diff
@@ -743,3 +783,164 @@ TEST_F(RedisCuckooFilterTest, ExpansionWritesNewFilterIndexPage) {
...
+    EXPECT_TRUE(deleted);
+    verifyMetadata(key_, 1000, 4, 500, 2, 2 - i, 1, i + 1);
+  }
+
+  bool deleted = true;
+  auto s = cuckoo_->Delete(*ctx_, key_, "duplicate", &deleted);
+  ASSERT_TRUE(s.ok()) << s.ToString();
+  EXPECT_FALSE(deleted);
+  verifyMetadata(key_, 1000, 4, 500, 2, 0, 1, 3);
+}
```
jihuayu, line 830: I think these case Go tests is OK, so there’s no need to add C++ test cases.

src/types/redis_cuckoo_chain.h
```diff
@@ -62,6 +65,8 @@ class CuckooChain : public Database {
                                              bool *inserted);
   rocksdb::Status commitSubFilterAndMetadata(engine::Context &ctx, const Slice &user_key, const std::string &ns_key,
                                              CuckooChainMetadata *metadata, CuckooSubFilter *sub_filter);
+  rocksdb::Status commitDelete(engine::Context &ctx, const Slice &user_key, const std::string &ns_key,
```
jihuayu, line 68: I feel that `commitDelete` is a bit odd from an architectural perspective, as it duplicates some of the logic in `commitSubFilterAndMetadata`. Could we reuse `commitSubFilterAndMetadata` for the commit-related logic instead?

tests/gocase/unit/type/bloom/cuckoo_filter_test.go
```diff
@@ -107,6 +107,42 @@ func TestCuckooFilter(t *testing.T) {
...
+		require.ErrorContains(t, rdb.Do(ctx, "cf.del", key, "item").Err(), "Not found")
+	})
+
+	t.Run("Del wrong type", func(t *testing.T) {
+		key := "test_cuckoo_filter_del_wrong_type"
+		require.NoError(t, rdb.Set(ctx, key, "value", 0).Err())
+		require.ErrorContains(t, rdb.Do(ctx, "cf.del", key, "item").Err(), "WRONGTYPE")
+	})
+
+	t.Run("Del basic", func(t *testing.T) {
```
jihuayu, line 128: The current tests don't seem to cover cases that span multiple sub-filters. Could you add a test case for that?

jihuayu (commented):
@nagisa-kunhah Overall, it looks good. Thank you!

jihuayu:
Regarding the background task, I think your overall design looks reasonable. I have a few thoughts:

1. I don't think the compaction operation is very urgent. We could add a debounce mechanism, for example, delaying it by 5 minutes, to avoid triggering multiple compactions after several modifications within a short period.
2. I think we should investigate how long the lock will be held during compaction. A short lock is acceptable, but if compaction may hold the lock for a long time, we should consider whether to add some additional fields to the data structure to support the GC process.
3. Regarding the issue of the task being lost after a restart, could we provide an explicit command that allows users to manually trigger the compaction?


PS: We can open a new discussion or issue to discuss this further.

PR #3573, opened 2026-08-03
Support AGGREGATE COUNT for sorted-set union and intersection

### LINKED ISSUE #3568: Support Redis 8.8 AGGREGATE COUNT for sorted-set union and intersection commands
### Search before asking

### Motivation

Redis 8.8 added the `COUNT` aggregation method to the following sorted-set commands:

- `ZUNION`
- `ZINTER`
- `ZUNIONSTORE`
- `ZINTERSTORE`

The relevant syntax is:

```text
ZUNION numkeys key [key ...]
  [WEIGHTS weight [weight ...]]
  [AGGREGATE <SUM | MIN | MAX | COUNT>]
  [WITHSCORES]

ZINTER numkeys key [key ...]
  [WEIGHTS weight [weight ...]]
  [AGGREGATE <SUM | MIN | MAX | COUNT>]
  [WITHSCORES]
```

The store variants support the same aggregation methods while writing the result to a destination key.

Kvrocks currently recognizes `SUM`, `MIN`, and `MAX`, but rejects `COUNT`. Supporting this option would improve compatibility with Redis 8.8.

#### Redis `COUNT` semantics

`COUNT` is an aggregation method; it does not limit the number of returned members.

For every member in the result:

```text
COUNT score(member) =
    sum of the weights of every input-key occurrence containing that member
```

If `WEIGHTS` is not specified, every input has an implicit weight of `1`. More formally:

```text
contribution(i, member) =
    weight[i], if member exists in input key i
    0,         otherwise
```

The original member scores are ignored. This is different from the existing aggregation methods:

```text
SUM / MIN / MAX contribution = original member score * input weight
COUNT contribution           = input weight only
```

The membership operation itself remains unchanged:

- `ZUNION` includes a member if it exists in at least one input.
- `ZINTER` includes a member only if it exists in every input.
- Consequently, without explicit weights, every member returned by `ZINTER ... AGGREGATE COUNT` has a score equal to `numkeys`.
- With weights, every member returned by the intersection has a score equal to the sum of all input weights.

Each input-key argument contributes independently. If the same key is supplied more than once, each occurrence contributes separately using the weight at that input position.

#### Example

```text
ZADD count:zset1 10 alice 20 bob
ZADD count:zset2 100 alice 200 carol
```

Although the original scores differ, they are ignored by `COUNT`:

```text
ZUNION 2 count:zset1 count:zset2 AGGREGATE COUNT WITHSCORES
```

The resulting member-to-score mapping is:

| Member | Score | Reason |
| --- | ---: | --- |
| `alice` | 2 | Present in both inputs |
| `bob` | 1 | Present only in `count:zset1` |
| `carol` | 1 | Present only in `count:zset2` |

With non-uniform weights:

```text
ZUNION 2 count:zset1 count:zset2
  WEIGHTS 2 3
  AGGREGATE COUNT
  WITHSCORES
```

the resulting mapping becomes:

| Member | Score | Reason |
| --- | ---: | --- |
| `alice` | 5 | Contributions `2 + 3` |
| `bob` | 2 | Contribution from the first input |
| `carol` | 3 | Contribution from the second input |

For the corresponding intersection, only `alice` is returned, with score `2` without weights and score `5` with `WEIGHTS 2 3`.

This aggregation method can be useful for membership frequency, weighted voting, consensus scoring, recommendation aggregation, and similar ranking or analytics workloads.

### Solution

I would like to propose extending the existing sorted-set union/intersection implementation to support `AGGREGATE COUNT` for all four commands.

Based on the current shared implementation, one possible approach is:

1. Add `COUNT` to the internal aggregation-method representation.
2. Extend the parser shared by `ZUNION`/`ZINTER` and the parser shared by `ZUNIONSTORE`/`ZINTERSTORE` to recognize `COUNT`.
3. For `COUNT`, use the input weight as the member's contribution instead of multiplying the original member score by the weight.
4. Use that contribution both when a member is first inserted into the intermediate result and when later occurrences are accumulated.
5. Reuse the existing union/intersection membership selection, result construction, and store paths.

The fourth point is important because the first occurrence is initialized before subsequent aggregation. Handling `COUNT` only in the later accumulation branch would incorrectly retain the first input's original member score.

This approach should keep the existing asymptotic complexity. No storage-format or command-registration changes are expected. I am also happy to adjust the implementation approach based on maintainer feedback.

#### Proposed test strategy

I propose adding both focused C++ unit tests and Go integration tests so that the internal calculation and external command behavior are verified independently.

##### C++ unit tests

Add focused GoogleTest cases in `tests/cppunit/types/zset_test.cc` that directly exercise `ZSet::Union` and `ZSet::Inter` with `COUNT`.

The C++ tests would cover:

- unweighted union and intersection;
- non-uniform weighted union and intersection;
- original source scores being ignored;
- correct initialization from the first input and accumulation from later inputs;
- members occurring in one, some, or all inputs;
- duplicate input-key arguments contributing once per input position.

##### Go integration tests

Extend `tests/gocase/unit/type/zset/zset_test.go` to verify the command layer and stored results.

The Go tests would cover:

- `ZUNION`, `ZINTER`, `ZUNIONSTORE`, and `ZINTERSTORE`;
- `COUNT` with and without `WEIGHTS`;
- `WITHSCORES` for the non-store commands;
- cardinality and stored scores for the store commands;
- case-insensitive parsing of `COUNT`;
- missing input keys and wrong-type inputs;
- rejection of an invalid aggregation method;
- regression coverage confirming that the default `SUM` behavior and the existing `SUM`, `MIN`, and `MAX` methods remain unchanged.

The added tests should exercise every new `COUNT`-specific branch and changed line. The existing C++ and Go test suites should continue to pass.

#### Scope

This proposal is limited to adding the new aggregation method to the existing four command paths. It does not propose a broader refactor of sorted-set union/intersection or any behavior change to `SUM`, `MIN`, or `MAX`.

#### References

- [Redis 8.8 release notes](https://redis.io/docs/latest/develop/whats-new/8-8/)
- [Redis 8.8 announcement](https://redis.io/blog/announcing-redis-8-8/#sorted-sets-union-and-intersection---count-aggregator)
- [`ZUNION` documentation](https://redis.io/docs/latest/commands/zunion/)
- [`ZINTER` documentation](https://redis.io/docs/latest/commands/zinter/)
- [`ZUNIONSTORE` documentation](https://redis.io/docs/latest/commands/zunionstore/)
- [`ZINTERSTORE` documentation](https://redis.io/docs/latest/commands/zinterstore/)

If the maintainers agree that this scope is appropriate, I would be happy to implement it and add both focused C++ unit tests and Go integration tests. I would also appreciate any feedback on the proposed approach. Thank you!

lintanghui:
Redis 8.8 added `COUNT` as an aggregation method for the sorted-set union/intersection commands, but kvrocks currently rejects `AGGREGATE COUNT` with a parse error.

This adds `kAggregateCount` to `AggregateMethod`, accepts `count` in the `ZUNION`/`ZUNIONSTORE`/`ZINTER`/`ZINTERSTORE` parsers, and implements the semantics in `ZSet::Union`/`ZSet::Inter`: each occurrence of a member contributes the key's weight (instead of the member score), summed across inputs — matching Redis behavior with the default weight of 1 acting as an occurrence counter.

Includes gocase coverage for union and intersection with `AGGREGATE COUNT`, including the weighted case.

Fixes #3568

### DIFF
--- a/src/commands/cmd_zset.cc
+++ b/src/commands/cmd_zset.cc
@@ -1138,6 +1138,8 @@ class CommandZUnion : public Commander {
           aggregate_method_ = kAggregateMin;
         } else if (util::ToLower(aggregate_value) == "max") {
           aggregate_method_ = kAggregateMax;
+        } else if (util::ToLower(aggregate_value) == "count") {
+          aggregate_method_ = kAggregateCount;
         } else {
           return {Status::RedisParseErr, "aggregate param error"};
         }
@@ -1223,6 +1225,8 @@ class CommandZUnionStore : public Commander {
           aggregate_method_ = kAggregateMin;
         } else if (util::ToLower(args[i + 1]) == "max") {
           aggregate_method_ = kAggregateMax;
+        } else if (util::ToLower(args[i + 1]) == "count") {
+          aggregate_method_ = kAggregateCount;
         } else {
           return {Status::RedisParseErr, "aggregate param error"};
         }
--- a/src/types/redis_zset.cc
+++ b/src/types/redis_zset.cc
@@ -658,7 +658,7 @@ rocksdb::Status ZSet::Inter(engine::Context &ctx, const std::vector<KeyWeight> &
   if (!s.ok() || target_mscores.empty()) return s;
 
   for (const auto &ms : target_mscores) {
-    double score = ms.score * keys_weights[0].weight;
+    double score = aggregate_method == kAggregateCount ? keys_weights[0].weight : ms.score * keys_weights[0].weight;
     if (std::isnan(score)) score = 0;
     dst_zset[ms.member] = score;
     member_counters[ms.member] = 1;
@@ -671,10 +671,11 @@ rocksdb::Status ZSet::Inter(engine::Context &ctx, const std::vector<KeyWeight> &
     for (const auto &ms : target_mscores) {
       if (dst_zset.find(ms.member) == dst_zset.end()) continue;
       member_counters[ms.member]++;
-      double score = ms.score * keys_weights[i].weight;
+      double score = aggregate_method == kAggregateCount ? keys_weights[i].weight : ms.score * keys_weights[i].weight;
       if (std::isnan(score)) score = 0;
       switch (aggregate_method) {
         case kAggregateSum:
+        case kAggregateCount:
           dst_zset[ms.member] += score;
           if (std::isnan(dst_zset[ms.member])) {
             dst_zset[ms.member] = 0;
@@ -766,13 +767,14 @@ rocksdb::Status ZSet::Union(engine::Context &ctx, const std::vector<KeyWeight> &
     auto s = RangeByScore(ctx, key_weight.key, spec, &target_mscores, &target_size);
     if (!s.ok() && !s.IsNotFound()) return s;
     for (const auto &ms : target_mscores) {
-      double score = ms.score * key_weight.weight;
+      double score = aggregate_method == kAggregateCount ? key_weight.weight : ms.score * key_weight.weight;
       if (std::isnan(score)) score = 0;
       if (dst_zset.find(ms.member) == dst_zset.end()) {
         dst_zset[ms.member] = score;
       } else {
         switch (aggregate_method) {
           case kAggregateSum:
+          case kAggregateCount:
             dst_zset[ms.member] += score;
             if (std::isnan(dst_zset[ms.member])) dst_zset[ms.member] = 0;
             break;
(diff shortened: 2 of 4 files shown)


git-hulk:
Please check [Guidelines for AI-assisted Contributions](https://kvrocks.apache.org/community/contributing#guidelines-for-ai-assisted-contributions).

jihuayu:
Closing because this was automatically submitted by AI.

PR #3578, opened 2026-08-04
fix(stream): decrement pending_number when XAUTOCLAIM removes deleted PEL entries

wengsht:
The group's and the owning consumer's `pending_number` can be serialized as a RESP integer larger than `INT64_MAX`, which breaks any client that decodes integer replies as signed 64-bit — the reply fails to parse and the connection is dropped. Hit in production.

## XAUTOCLAIM `pending_number` drift

When XAUTOCLAIM meets a PEL entry whose stream entry was trimmed or XDEL'd, it deletes the PEL entry but never decrements the group's or the owning consumer's `pending_number`. The counts drift upward on every such sweep and, being unsigned, eventually underflow-wrap on a later decrement — after which `XINFO GROUPS` and `XPENDING` report garbage.

The fix mirrors `DeletePelEntries` (XACK): when a deleted entry is removed from the PEL, decrement the owning consumer and the group. A claim only moves ownership within the group (net-zero for the group count), so only deleted entries reduce it. Subtractions are saturating so a counter that has already drifted can't wrap.

Test: read entries into a consumer, XDEL them, XAUTOCLAIM, then check the group and consumer pending counts return to 0. The existing XDEL autoclaim tests only asserted the returned deleted-id list.

> This PR previously also carried an `XINFO GROUPS` lag-underflow fix. Per review it is a separate issue (SRP + clean cherry-pick), so I've dropped it here and will open a follow-up PR.

---
AI assistance: diagnosis and drafting were done with AI help; I've reviewed the changes and tests and understand the behavior.

### DIFF
--- a/src/types/redis_stream.cc
+++ b/src/types/redis_stream.cc
@@ -643,19 +643,25 @@ rocksdb::Status Stream::AutoClaim(engine::Context &ctx, const Slice &stream_name
   // A claim keeps the group total (entry moves between consumers); a deleted dangling
   // entry leaves the PEL, so decrement both its consumer and the group. Saturating.
   const uint64_t deleted_count = deleted_entries.size();
+  // The current consumer never appears in claimed_consumer_entity_count (we only claim from other
+  // consumers), so keep its own deleted-entry decrement in a scalar and let the map hold only the
+  // other consumers, avoiding a find-and-erase on the map.
+  uint64_t current_consumer_decrement = 0;
   std::map<std::string, uint64_t> consumer_pending_decrements = claimed_consumer_entity_count;
   for (const auto &[consumer, cnt] : deleted_consumer_count) {
-    consumer_pending_decrements[consumer] += cnt;
+    if (consumer == consumer_name) {
+      current_consumer_decrement += cnt;
+    } else {
+      consumer_pending_decrements[consumer] += cnt;
+    }
   }
 
   if (total_claimed_count > 0 || deleted_count > 0) {
     current_consumer_metadata.pending_number += total_claimed_count;
-    if (auto it = consumer_pending_decrements.find(consumer_name); it != consumer_pending_decrements.end()) {
-      current_consumer_metadata.pending_number = current_consumer_metadata.pending_number >= it->second
-                                                     ? current_consumer_metadata.pending_number - it->second
-                                                     : 0;
-      consumer_pending_decrements.erase(it);
-    }
+    current_consumer_metadata.pending_number =
+        current_consumer_metadata.pending_number >= current_consumer_decrement
+            ? current_consumer_metadata.pending_number - current_consumer_decrement
+            : 0;
     current_consumer_metadata.last_attempted_interaction_ms = now_ms;
 
     s = batch->Put(stream_cf_handle_, consumer_key, encodeStreamConsumerMetadataValue(current_consumer_metadata));
--- a/src/types/redis_stream.cc
+++ b/src/types/redis_stream.cc
@@ -582,6 +582,7 @@ rocksdb::Status Stream::AutoClaim(engine::Context &ctx, const Slice &stream_name
 
   auto iter = util::UniqueIterator(ctx, read_options, stream_cf_handle_);
   uint64_t total_claimed_count = 0;
+  std::map<std::string, uint64_t> deleted_consumer_count;
   for (iter->SeekToFirst(); iter->Valid() && count > 0 && attempts > 0; iter->Next()) {
     std::string tmp_group_name;
     StreamEntryID entry_id = groupAndEntryIdFromPelInternalKey(iter->key(), tmp_group_name);
@@ -605,6 +606,9 @@ rocksdb::Status Stream::AutoClaim(engine::Context &ctx, const Slice &stream_name
         deleted_entries.push_back(entry_id);
         s = batch->Delete(stream_cf_handle_, iter->key());
         if (!s.ok()) return s;
+        // The referenced entry was trimmed/XDEL'd; dropping this PEL record must
+        // decrement pending_number below, otherwise the counter drifts up and wraps.
+        deleted_consumer_count[penl_entry.consumer_name] += 1;
         --count;
         continue;
       }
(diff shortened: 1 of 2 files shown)


wengsht:
@LindaSummer  could you help approve the CI run? This fixes pending_number drifting upward when XAUTOCLAIM sweeps PEL entries whose stream body was already trimmed/XDEL'd — the counter is an unclamped uint64, so in production it eventually wraps past INT64_MAX and breaks any client that decodes XINFO GROUPS pending as a signed int. The decrement mirrors the existing logic in DeletePelEntries. Repro added as a gocase test.

LindaSummer:
Hi @wengsht ,

Thanks for your effort in this problem.
I will review this patch in recent days.

If you want to run a full CI by yourself, you could create a PR in your forked repo and enable its github action.
This will accelerate your development without pending by upstream CI gate.

wengsht:
> Hi @wengsht ,
> 
> Thanks for your effort in this problem. I will review this patch in recent days.
> 
> If you want to run a full CI by yourself, you could create a PR in your forked repo and enable its github action. This will accelerate your development without pending by upstream CI gate.

Thanks for the tip!

Ran CI:
https://github.com/wengsht/kvrocks/pull/1/checks

src/types/redis_stream.cc
```diff
@@ -1308,12 +1339,17 @@ static void CheckLagValid(const StreamMetadata &stream_metadata, StreamConsumerG
...
     group_metadata.lag = stream_metadata.entries_added - group_metadata.entries_read;
     valid = true;
   } else {
     int64_t entries_read = StreamEstimateDistanceFromFirstEverEntry(stream_metadata, group_metadata.last_delivered_id);
-    if (entries_read != -1) {
+    // entries_read can exceed entries_added when the group's read counter has been
+    // corrupted (e.g. XSETID lowering entries_added below it). lag is an unsigned
+    // field, so the subtraction would wrap to a near-2^64 value that overflows the
+    // i64 the RESP integer reply is decoded as. Treat that as an invalid lag instead.
+    if (entries_read != -1 && static_cast<uint64_t>(entries_read) <= stream_metadata.entries_added) {
```
LindaSummer: This behavior seems not aligning with redis.

Here is the redis version 8.8.0 docker  result.
```txt
redis-cli -p 6379
127.0.0.1:6379> DEL repro-lag
(integer) 1
127.0.0.1:6379> XADD repro-lag 1-0 f v
"1-0"
127.0.0.1:6379> XADD repro-lag 2-0 f v
"2-0"
127.0.0.1:6379> XADD repro-lag 3-0 f v
"3-0"
127.0.0.1:6379> XGROUP CREATE repro-lag grp 0 ENTRIESREAD 1000000
OK
127.0.0.1:6379> XINFO GROUPS repro-lag
1)  1) "name"
    2) "grp"
    3) "consumers"
    4) (integer) 0
    5) "pending"
    6) (integer) 0
    7) "last-delivered-id"
    8) "0-0"
    9) "entries-read"
   10) (integer) 3
   11) "lag"
   12) (integer) 3
```

Here is this patch's result.
```txt
redis-cli -p 6666
127.0.0.1:6666> DEL repro-lag
(integer) 1
127.0.0.1:6666> XADD repro-lag 1-0 f v
 repro-lag grp 0"1-0"
127.0.0.1:6666> XADD repro-lag 2-0 f v
"2-0"
127.0.0.1:6666> XADD repro-lag 3-0 f v
"3-0"
127.0.0.1:6666> XGROUP CREATE repro-lag grp 0 ENTRIESREAD 1000000
OK
127.0.0.1:6666> XINFO GROUPS repro-lag
1)  1) "name"
    2) "grp"
    3) "consumers"
    4) (integer) 0
    5) "pending"
    6) (integer) 0
    7) "last-delivered-id"
    8) "0-0"
    9) "entries-read"
   10) (integer) 1000000
   11) "lag"
   12) (integer) 3
```

The `entries-read` should be the `entries_added` if `entries_read` is greater than `entries_added`.

Here is redis implemetation.

https://github.com/redis/redis/blob/8f365337d48c8a264bc9f3381c7b8cc12ffb51d2/src/t_stream.c#L3663-L3665

LindaSummer (commented):
Hi @wengsht ,

Thanks very much for your effort for these two issues.

I have several suggestions for this patch.

1. It appears that we solved two independent issues in one patch. This is not good practice since we need to cherry-pick in the release process, and it violates SRP (Single Responsibility Principle). We'd better create two patches for these two issues. 
2. The fix for the `entries-read` has a case not aligning with Redis's behavior.

For the `XAUTOCLAIM` fix, it generally looks good to me.

Have a wonderful day!

src/types/redis_stream.cc
wengsht: Thanks for your careful review! @LindaSummer 

I have stripped the second change from this PR and also the XAUTOCLAIM fix remains. Can you help take another look?

On the entries-read misalignment — you're right, my version was wrong. Redis clamps entries_read down to entries_added at write time. I will draft second PR for that.

wengsht: For the PR #3581 

 # kvrocks with PR #3581 (:6666)
  127.0.0.1:6666> DEL repro-lag
  (integer) 1
  127.0.0.1:6666> XADD repro-lag 1-0 f v
  "1-0"
  127.0.0.1:6666> XADD repro-lag 2-0 f v
  "2-0"
  127.0.0.1:6666> XADD repro-lag 3-0 f v
  "3-0"
  127.0.0.1:6666> XGROUP CREATE repro-lag grp 0 ENTRIESREAD 1000000
  OK
  127.0.0.1:6666> XINFO GROUPS repro-lag
  1)  1) "name"
      2) "grp"
      3) "consumers"
      4) (integer) 0
      5) "pending"
      6) (integer) 0
      7) "last-delivered-id"
      8) "0-0"
      9) "entries-read"
     10) (integer) 3
     11) "lag"
     12) (integer) 3

src/types/redis_stream.cc
```diff
@@ -636,25 +640,52 @@ rocksdb::Status Stream::AutoClaim(engine::Context &ctx, const Slice &stream_name
...
+    consumer_pending_decrements[consumer] += cnt;
+  }
+
+  if (total_claimed_count > 0 || deleted_count > 0) {
     current_consumer_metadata.pending_number += total_claimed_count;
+    if (auto it = consumer_pending_decrements.find(consumer_name); it != consumer_pending_decrements.end()) {
+      current_consumer_metadata.pending_number = current_consumer_metadata.pending_number >= it->second
+                                                     ? current_consumer_metadata.pending_number - it->second
+                                                     : 0;
+      consumer_pending_decrements.erase(it);
```
LindaSummer: Maybe we could make current consumer data stored outside of the `consumer_pending_decrements` to reduce erase of a map.

Just a nitpick, non-mandatory.

LindaSummer (commented):
Hi @wengsht ,

Thanks very much for your dig!

Left one more comment.

Generally LGTM.

src/types/redis_stream.cc
wengsht: thanks for the review again!

updated.

LindaSummer (approved):
🚀 LGTM

jihuayu (approved):

PR #3591, opened 2026-08-13
feat(CuckooFilter): add cf.info and cf.count command

ChaseForward:
part of: #3552

## Summary
Add RedisBloom-compatible CF.INFO and CF.COUNT support for Cuckoo Filter.

## Design

CF.INFO returns the filter metadata (total slot size, number of buckets,
number of sub-filters, number of items inserted/deleted, bucket size,
expansion rate, max iterations), matching RedisBloom's field order.

CF.INFO key items returns a nested key-value pair for inserted and deleted
counts; CF.INFO key returns all fields as a flat array.

CF.COUNT returns the total number of fingerprint matches across all
sub-filters for the given item, counting each occurrence separately
(duplicates inserted via CF.ADD are all counted), without modifying any
slot.

This PR was written using TRAE.

### DIFF
--- a/src/commands/cmd_cuckoo_filter.cc
+++ b/src/commands/cmd_cuckoo_filter.cc
@@ -131,8 +131,142 @@ class CommandCFAdd : public Commander {
   }
 };
 
-// Register the CF.RESERVE and CF.ADD commands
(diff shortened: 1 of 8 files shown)


jihuayu:
Hi @ChaseForward Thank you for your contribution. Unless there is a necessary reason to do otherwise, please limit each PR to adding a single command.

I’m going to close this PR. You can split it into two separate PRs and resubmit them.

PR #3594, opened 2026-08-18
feat(rocksdb): add optional direct I/O for reads and flush/compaction

bocharov:
### Summary

Add two opt-in RocksDB options, both **off by default**:

- `rocksdb.use_direct_reads`
- `rocksdb.use_direct_io_for_flush_and_compaction`

They map to the corresponding RocksDB `DBOptions` and let an operator use `O_DIRECT` to bypass the OS page cache.

### Motivation

With buffered reads, the OS page cache grows with the working set. Under a hard memory limit (for example a cgroup memory limit), that page cache is accounted against the limit and can lead to the process being OOM-killed even when the block cache and heap are well within budget. Enabling direct reads keeps the process's resident memory bounded to the RocksDB block cache.

The trade-off is that reads no longer benefit from the OS page cache, so `rocksdb.block_cache_size` should be sized to the working set when enabling this.

### Notes

- Both default to `no`, so existing behavior is unchanged.
- They are applied when the DB is opened, so they are read-only at runtime (a `CONFIG SET` is rejected rather than silently ignored).

### Tests

- cppunit `Config.DirectIO`: defaults off, parsed from the config file, and read-only at runtime.
- gocase `TestConfigDirectIO`: through a running server, defaults `no` and `CONFIG SET` rejected.


jihuayu:
Hello @bocharov. Please disclose in the PR which AI tools and models you used. Thank you.

git-hulk:
@bocharov Want to know how would you use those configurations?

bocharov:
@git-hulk Honestly: the driver was bounding cgroup-accounted page cache on memory-capped kvrocks pods — but after deploying, our OOM turned out to be output buffers, which O_DIRECT doesn't touch, and that one workload has since moved off kvrocks. We still run kvrocks for other use cases, so the option remains relevant to us where RSS is page-cache-dominated — but I no longer have a proven production win for it. Happy to close unless you'd still like it in.

@jihuayu Built with Claude Code (Anthropic Claude), reviewed and verified by me.

PR #3598, opened 2026-08-24
fix(server): add O_TRUNC when creating PID file to avoid stale content

weim0000:
What problem does this change solve?

CreatePidFile opens the PID file with O_RDWR | O_CREAT but without O_TRUNC. If a stale PID file already exists from a previous run and the old PID string is longer than the new one, open() + write() will only overwrite the first N bytes of the file, leaving the trailing bytes from the old content intact. This results in a corrupted PID file.

Example:
Step                                                                File content    Note
Previous run (PID=12345)                            12345             correct
New run (PID=678), without O_TRUNC      67845             corrupted — trailing 45 is stale
New run (PID=678), with O_TRUNC            678                 correct

A corrupted PID file can cause process management tools (e.g., kill $(cat pidfile), systemd, init scripts) to target the wrong process or fail to stop/restart the service.

How does this change fix the problem?

Add O_TRUNC to the open() flags in CreatePidFile so the file is truncated to zero length before writing the new PID, ensuring no stale bytes remain.

Changes

- src/cli/pid_util.h: Add O_TRUNC flag to the open() call in CreatePidFile.
- tests/cppunit/pid_util_test.cc: Add unit tests covering:
- - Truncation on existing file: Write a longer fake PID first, then call CreatePidFile and verify the content matches exactly the current PID with no leftover bytes.
- - Creation from scratch: Verify CreatePidFile works correctly when no file exists.
- - RemovePidFile: Verify the PID file is properly removed.

### DIFF
--- a/tests/cppunit/pid_util_test.cc
+++ b/tests/cppunit/pid_util_test.cc
@@ -21,13 +21,12 @@
 #include "cli/pid_util.h"
 
 #include <fcntl.h>
+#include <gtest/gtest.h>
 #include <unistd.h>
 
 #include <cstdio>
 #include <fstream>
 #include <string>
-
-#include <gtest/gtest.h>
 
 TEST(PidUtil, CreatePidFileTruncatesExistingContent) {
   // Simulate a stale PID file whose content is longer than the current PID.
--- a/tests/cppunit/pid_util_test.cc
+++ b/tests/cppunit/pid_util_test.cc
@@ -59,29 +59,3 @@ TEST(PidUtil, CreatePidFileTruncatesExistingContent) {
 
   RemovePidFile(path);
 }
-
-TEST(PidUtil, CreatePidFileFromScratch) {
-  const std::string path = "/tmp/kvrocks_pid_util_test_fresh.pid";
-  std::remove(path.data());
-
-  auto status = CreatePidFile(path);
-  ASSERT_TRUE(status.IsOK());
-
-  std::ifstream ifs(path);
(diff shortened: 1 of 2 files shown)


tests/cppunit/pid_util_test.cc
```diff
@@ -0,0 +1,87 @@
...
+
+TEST(PidUtil, RemovePidFile) {
+  const std::string path = "/tmp/kvrocks_pid_util_test_remove.pid";
+
+  ASSERT_TRUE(CreatePidFile(path).IsOK());
+  ASSERT_EQ(access(path.data(), F_OK), 0);
+
+  RemovePidFile(path);
+  ASSERT_NE(access(path.data(), F_OK), 0);
+}
```
git-hulk: @weim0000 Could we remove those two tests? because they are not related to the current fix.

git-hulk (commented):
Looks good. One comment to remove unrelated tests.

tests/cppunit/pid_util_test.cc
PragmaTwice: Yah agree with you.

weim0000: ok, thanks, I will remove those two tests

git-hulk (dismissed):

PragmaTwice (dismissed):

git-hulk:
@weim0000 could you use `./x.py format` to format your code to pass the CI, see: https://github.com/apache/kvrocks/actions/runs/33581512347/job/100096813700?pr=3598

git-hulk (approved):

PR #3599, opened 2026-08-25
fix(comptaction): use int64_t for parsing total_keys and deleted_keys

DanFaudemer:
# Summary 

When the number of total keys is above 2^31 the following error is present in the logs 

```
[E][compaction_checker.cc:87] [compaction checker] Parse total_keys error: out of range of integer type
``` 

Parsing as `int64_t` match `total_keys` and `deleted_keys` type.

### DIFF
--- a/src/storage/compaction_checker.cc
+++ b/src/storage/compaction_checker.cc
@@ -82,7 +82,7 @@ void CompactionChecker::PickCompactionFilesForCf(const engine::ColumnFamilyConfi
 
     for (const auto &property_iter : iter.second->user_collected_properties) {
       if (property_iter.first == "total_keys") {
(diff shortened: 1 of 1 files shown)


PragmaTwice (approved):
Looks good. With this patch does the compaction procedure work well for > 2^31 keys?

PragmaTwice:
@git-hulk @jihuayu  Maybe we should have some stress testing go cases, like this one: with > 2^31 keys.

git-hulk (approved):

git-hulk:
@DanFaudemer, I guess it might have had a compaction performance issue while the number of keys was too large. cc @PragmaTwice

jihuayu (approved):

PR #3600, opened 2026-08-29
feat(keyspace): add keyspace notifications for string commands

Aetherance:
Following #3541, which introduced the keyspace notification mechanism, this PR adds keyspace notification support for the remaining `$` string command paths.

The following commands are covered:

* SET
* SETNX
* SETEX
* PSETEX
* GETSET
* MSET
* MSETNX
* MSETEX
* APPEND
* SETRANGE
* INCR
* DECR
* INCRBY
* DECRBY
* INCRBYFLOAT
* CAS

assisted by codex gpt 5.6

### DIFF
--- a/kvrocks.conf
+++ b/kvrocks.conf
@@ -617,7 +617,7 @@ lua-strict-key-accessing no
 #   K  keyspace channels
 #   E  keyevent channels
 #   g  generic events, currently del
-#   $  string events, currently set
+#   $  string events: set, append, setrange, incrby, incrbyfloat
 #   A  same as g$, without K or E
 #
 # Default namespace uses db 0. Redis database namespaces use db indexes.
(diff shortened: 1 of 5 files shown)


Aetherance:
Hi @git-hulk, @jihuayu, and @PragmaTwice, this PR is one of the follow-ups to [https://github.com/apache/kvrocks/pull/3541](https://github.com/apache/kvrocks/pull/3541). If you have time, I’d be happy to hear your thoughts.

Aetherance:
After some consideration, I think it would be better to support replicas first, so I’ve marked this PR as a draft for now.

jihuayu:
> After some consideration, I think it would be better to support replicas first,

Yes! I agree with you!

PS: If you'd like, you can turn these small, specific tasks into issues so others can take a look and pick them up.

Aetherance:
> > After some consideration, I think it would be better to support replicas first,
> 
> Yes! I agree with you!
> 
> PS: If you'd like, you can turn these small, specific tasks into issues so others can take a look and pick them up.

Sure! But that will need to wait until replica reads are implemented, since the replica-read PR may change the existing `AddKeySpace` API.

PragmaTwice:
You can open a tracking issue like this: https://github.com/apache/kvrocks/issues/3436.

Aetherance:
> You can open a tracking issue like this: #3436.

Thanks!

PR #3609, opened 2026-09-06
refactor(cuckoo-filter): move page cache ownership to chain

nagisa-kunhah:
- Move cuckoo filter page-cache ownership from individual subfilters to the chain operation.
- Share cached pages across subfilters and commit dirty pages with metadata at the chain layer.
- Add slot-level undo for failed kick-out attempts while preserving previously staged page mutations.
- Separate the insert, kick-out, and expansion stages from persistence so future multi-item insert operations can reuse the same metadata and page cache before committing once.

## Testing

- `./x.py check format`
- `cmake --build build --target kvrocks -j2`
- `cmake --build build --target unittest -j2`
- `./build/unittest --gtest_filter='*Cuckoo*'`
- `./build/unittest`
This pr is This PR was written using codex and GPT-5.6 sol


src/types/cuckoo_filter_page.h
```diff
@@ -34,9 +34,22 @@ namespace redis {
 
 class CuckooPageCache {
  public:
+  struct SlotMutation {
+    uint8_t OldFingerprint() const { return old_fingerprint_; }
+
+    uint16_t filter_index_ = 0;
+    uint32_t num_buckets_ = 0;
+    uint32_t bucket_index_ = 0;
+    uint32_t slot_idx_ = 0;
```
geetanshjuneja: Is it better to store page key in mutation instead of bucket and slot idx?

Issue #1372, opened 2023-04-05
Add test cases for the data encoding compatibility


git-hulk:
### Search before asking

- [X] I had searched in the [issues](https://github.com/apache/incubator-kvrocks/issues) and found no similar issues.

### Motivation

As @PragmaTwice mentioned in an offline topic(@DenizPiri also mentioned in https://github.com/apache/incubator-kvrocks/issues/414#issuecomment-1432883367.), we didn't have any test cases to ensure the data encoding is compatible with the old version, which is very important for a storage service.

### Solution

We can create a minimized dataset for all types per release(starting from 2.3.0), then load and check the dataset is ok to load.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

HolyLow:
@git-hulk I see this issue in the roadmap, and as it seems that no one is working on it (because this issue is troublesome and dirty...?), could you assign it to me?

Currently I only have some unclear thoughts on this topic. I think to check the data encoding compatibility, we are supposed to:

1. Write the raw data in different version's encoding (use the RocksDB api directly and bypass the kvrocks api);
2. Then read the data with the current version's kvrocks api to check if the data is readable and the data semantic is correct.

But still, I am not sure of the proper workflow, or how far we should go. 
Maybe you could kindly refer me to some existing projects' "data encoding checking" mechanism/code to shed me some light, if you happen to know. 
I would try to find some material myself as well.

git-hulk:
> @git-hulk I see this issue in the roadmap, and as it seems that no one is working on it (because this issue is troublesome and dirty...?), could you assign it to me?
> 
> Currently I only have some unclear thoughts on this topic. I think to check the data encoding compatibility, we are supposed to:
> 
> 1. Write the raw data in different version's encoding (use the RocksDB api directly and bypass the kvrocks api);
> 2. Then read the data with the current version's kvrocks api to check if the data is readable and the data semantic is correct.
> 
> But still, I am not sure of the proper workflow, or how far we should go. Maybe you could kindly refer me to some existing projects' "data encoding checking" mechanism/code to shed me some light, if you happen to know. I would try to find some material myself as well.

Hi @HolyLow Assigned, thanks for your interest.

Let me think a bit about this, I will ping you back soon.

HolyLow:
@git-hulk Hello, do you have any thoughts yet?

git-hulk:
[...]

PR #3610, opened 2026-09-08
test: Add compatibility tests across versions of `kvrocks`

yatriks:
## What

Adds integration tests validating that kvrocks data files written by older releases load and serve correctly on the current code, using [testcontainers-go](https://github.com/testcontainers/testcontainers-go) (#1372). Old versions run in containers against a shared volume, populate real data, then a freshly built `current` image starts on the same volume and verifies every key/value survived intact.

## Compatibility matrix

| Old version | Types exercised | Type first supported in |
|-------------|-----------------|-------------------------|
| v2.3.0 | string, hash, list, set, zset, stream, bitmap | — |
| v2.6.0 | + bloom filter | v2.6.0 |
| v2.7.0 | + json | v2.7.0 |
| v2.10.0 | + hyperloglog | v2.10.0 |
| v2.14.0 | + tdigest | v2.14.0 |
| v2.15.0 | + timeseries | v2.15.0 (disabled in v2.14.0) |

Each old version is populated only with the types it supports (version-aware capability map), then all are verified on the new side.

## CI

New `compatibility.yml` workflow runs the 6 pairs in parallel: builds the PR's own image once (`current`), pre-pulls the published old-version images, then each matrix job validates one pair. Runs on PRs touching storage/types/ commands/config, on pushes to `unstable`, and on releases.

Tests are gated behind a `compat` build tag so the default gocase suite in the main CI stays unaffected.

Local verification: full matrix passes (`KVROCKS_OLD_VERSION=v2.x go test -tags compat -count=1 -run TestCompatibilityAllTypes ./integration/compatibility/`).
Verified with GHA on fork: https://github.com/yatriks/kvrocks/pull/1

## Feedback notes
- Can the CI live in a better place? Downstream of other builds perhaps? (Note currently we build the `current` image and pull the rest from DockerHub)
- Does a release checklist need to be updated to add new images to this files with release/something we should automate?
- Please double check the compatibility matrix for correctness, as I have inferred from release notes.


PragmaTwice (commented):
Sorry for the late review.

It seems that this PR doesn't trigger the newly-added CI job so I think maybe we need to add `.github/workflows/compatibility.yml` to `paths`.

yatriks:
@PragmaTwice fixed 🙏

.github/workflows/compatibility.yml
```diff
@@ -0,0 +1,112 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v6
+
+      # The old-version images are pulled from Docker Hub by the test; only
+      # the "new" side is built here, once, and handed to the parallel test
+      # jobs. Uses the same buildx method as the nightly image build; push
+      # defaults to false, so the image stays local.
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd
```
PragmaTwice:
```suggestion
        uses: docker/setup-buildx-action@f87e5991a6d7451dcb8d9637bfbc97413f497069
```

.github/workflows/compatibility.yml
```diff
@@ -0,0 +1,112 @@
...
+  push:
+    branches: [unstable]
+  pull_request:
+    paths:
+      - '.github/workflows/compatibility.yml'
+      - 'src/storage/**'
+      - 'src/types/**'
+      - 'src/commands/**'
+      - 'src/config/**'
+      - 'tests/gocase/**'
```
PragmaTwice:
```suggestion
      - 'src/**'
      - 'tests/**'
```

PragmaTwice (dismissed):

PragmaTwice (approved):

PR #3615, opened 2026-09-10
fix(cluster): accept Redis-style combined address in CLUSTERX SETNODES

### LINKED ISSUE #3495: CLUSTERX SETNODES returns "Invalid cluster node port" when the bus port (default = client + 10000) is not reachable
## Summary

In container/Mesos/Kubernetes-style environments only a single TCP port
is exposed per data node; the implicit "cluster bus port = client + 10000"
that Redis Cluster uses is not reserved by the scheduler and is not
listening. `CLUSTERX SETNODES` then refuses any well-formed nodes string:

```
> CLUSTERX SETNODES "<id> <host>:<port>@<bus> master - 0-5460" 100
-ERR Invalid cluster node port
```

Stripping `@<bus>` (i.e. passing only `<host>:<port>`) fails the same way.

This blocks operators from using `SETNODES` as a manual recovery path
when the controller-managed propagation falls behind (see related
`apache/kvrocks-controller` issue).

## Expected

When `cluster-announce-bus-port` is 0 (or absent from the nodes string),
skip the bus-port reachability validation, or fall back to the client port.
The bus port should not be a strict requirement when nodes are managed
by an external controller that drives them solely over the client port.

## Versions

- Apache Kvrocks 2.15.0

## Environment context

- Aurora/Mesos dynamic-port scheduler — each task is given **one** port
  (the `kvrocks` port). Adding `+10000` for the bus port lands on a port
  the kernel has not bound and that the firewall has not allowed.
- Same shape applies to Kubernetes Services that expose only one port,
  or Docker containers run with `-p <client>:<client>` only.

linsen458-spec:
Fixes #3495

`CLUSTERX SETNODES` currently requires the address to be split into two fields (`<host> <port>`), so node lines written in the Redis-style combined notation fail with `ERR Invalid cluster node port`:

```
> CLUSTERX SETNODES "<id> 127.0.0.1:7000@17000 master - 0-5460" 1
-ERR Invalid cluster node port
```

That combined notation isn't exotic: it's exactly what `CLUSTER NODES` prints (kvrocks itself emits `host:port@bus` there), and it's what you get when feeding node lines copied from a Redis cluster or from an external cluster manager — which is what happened in #3495, where only one port exists in the deployment.

This PR normalizes the address field before the existing field checks: `parseClusterNodes` now also accepts `host:port[@bus_port]` (bracketed IPv6 hosts too, e.g. `[::1]:7000`), and the bus port is parsed but ignored since kvrocks has no separate bus port. The old `<host> <port>` layout and all existing error messages are unchanged.

Tests:
- `ClusterTest.ClusterSetNodesWithRedisStyleAddress` (cppunit): master with/without bus port, slave with bus port, bracketed IPv6, and a malformed address keeping the old error — passes locally (`./build/unittest --gtest_filter='ClusterTest.*'`, 9/9 green)
- gocase `TestClusterNodes/accept_Redis-style_combined_address`: end-to-end SETNODES with `host:port@bus` and `host:port` (CI will cover this; I don't have a local Go toolchain yet)


linsen458-spec:
The first CI run failed in the gocase cluster test: my new subtest was inserted before `command_line_simulation_with_missing_newlines` and bumped the cluster version to 4, so the later subtest's SETNODES with version 2 correctly failed with `Invalid cluster version`. Moved the subtest to the end of the test with versions 3 and 4 — pushed.

src/cluster/cluster.cc
```diff
@@ -745,6 +782,20 @@ Status Cluster::parseClusterNodes(const std::string &nodes_str, ClusterNodes *no
   // Parse all nodes
   for (const auto &node_str : nodes_info) {
     std::vector<std::string> fields = util::Split(node_str, " ");
+
+    // Also accept the Redis-style combined address "<host>:<port>[@<bus_port>]"
+    // (the same format GetClusterNodes prints). Normalize it to the plain
+    // "<host> <port>" layout so the parser below stays unchanged.
+    if (fields.size() >= 2 && fields[1].find(':') != std::string::npos) {
```
git-hulk, line 789: This might break the old format if it contains an IPv6 address. I think we can recognize it by checking whether field[2] is one of `master`|`slave`|`replica`.

linsen458-spec: Thanks for catching this, you're right — checking for ':' would indeed break old-format lines with an IPv6 host.

Fixed as you suggested: we now detect the combined form by checking whether fields[2] is one of master/slave/replica. In the old layout that position holds the port number (digits only), and in the combined form it holds the role flags, so there's no ambiguity, and lines like `::1 30002` won't be misdetected anymore.

Also added tests for both: one making sure an old-format line with an IPv6 host isn't mistaken for the combined form, and one for the bracketed IPv6 notation. Unit tests and integration tests all pass locally.

git-hulk (approved):

PR #3617, opened 2026-09-11
feat(commands): add WALGET to inspect write batches by sequence number

weim0000:
## Summary

Add `WALGET <seq> [DETAIL]` to inspect the WAL write batch containing a given sequence number. This helps troubleshoot stored data and replication issues without manually decoding binary write batches.

The command returns the batch's start and end sequence numbers, followed by its records. It rejects requests when the requested sequence is not contained in the available batch.

With `DETAIL`, the command:
- Decodes namespace, optional slot ID, user key, version, and subkey fields for supported column families.
- Decodes metadata values and sorted-set index scores and members.
- Displays binary key fields as escaped text and hexadecimal strings.
- Decodes both boundaries of range deletions and identifies truncated key fields.

Supported records include Put, Merge, Delete, SingleDelete, DeleteRange, and LogData. The command is read-only, restricted to administrators, and unavailable in transactions and scripts.

## Usage

    WALGET 12345
    WALGET 12345 DETAIL

The command inspects the entire containing batch, including when the requested sequence is in the middle of that batch. It requires the corresponding WAL data to remain available.

## Testing

Added C++ and Go coverage for batch selection, argument validation, summary/detail output, binary key fields, slot encoding, range deletions, sorted-set index keys, and malformed data.

Validation completed:
- Formatting and whitespace checks.
- Clang syntax checks for the modified implementation and C++ tests.
- Focused clang-tidy checks with static analyzer checks disabled.

Validation limitations:
- Full build and runtime tests remain unverified because GCC 14 fails to compile GoogleTest against the local macOS SDK.
- Go tests could not run because required dependencies were unavailable.
- Repository-wide lint checks did not pass.


git-hulk:
@weim0000 We aleady have `POLLUPDATES` command for this propose, could you please have a look at if that fits what you want?

weim0000:
> @weim0000 We aleady have `POLLUPDATES` command for this propose, could you please have a look at if that fits what you want?

Thanks  for pointing this out, @git-hulk! I did take a close look at `POLLUPDATES`, and you're right that the two overlap quite a bit — both read write batches from the WAL via `GetUpdatesSince`, so I completely understand the concern about duplication.

From what I can tell, they're aimed at slightly different scenarios, though:

- `POLLUPDATES` seems designed for **incremental / streaming consumption** — it walks forward from a sequence, reads up to `MAX` batches, and returns a `next_sequence` cursor so a consumer (e.g. a replication/CDC-like client) can keep resuming. Its `RESP` format reconstructs logical Redis commands, and `RAW` returns the hex of the whole batch.

- `WALGET` is intended more as a **point-inspection / debugging tool**: given a single sequence number, it locates the one write batch that contains it, verifies that the sequence is actually present in the available WAL (rather than silently returning a later batch), and — with `detail` — decodes each record field by field: column family name, operation type, the raw key/value bytes plus hex, and the internal metadata fields (expire, version, size, etc.), keeping binary or malformed data visible.

The main gap I was trying to fill is that last part. `POLLUPDATES`'s `RESP` output is a command-level view (and is intentionally selective — e.g. it skips the secondary-subkey CF, relies on `LogData` to infer types, and some types like bloom filters are still TODO), while `RAW` is just the whole-batch hex. Neither really exposes the **storage-level encoding of each key/value field**, which is exactly what I found myself needing when debugging encoding issues.

That said, I'm definitely open to your thoughts here — if you feel this could be folded into `POLLUPDATES` (e.g. as an additional `FORMAT`/`DETAIL` option) instead of a separate command, I'd be happy to go that route. Just wanted to share the reasoning and hear what you think. Thanks again for the review!

weim0000:
Thanks again for the pointer, @git-hulk — I ran both commands against the exact same write so the difference is concrete rather than abstract.

**Scenario:** a single `HSET user:1000 name Alice age 30` (one write batch, spanning seq 9–11).

---

**`POLLUPDATES 8 MAX 1 FORMAT RESP`** — reconstructs the logical command (ideal for a consumer):

    updates -> "default":
      1) HSET user:1000 age 30
      2) HSET user:1000 name Alice
    next_sequence = 11, latest_sequence = 11

**`POLLUPDATES 8 MAX 1 FORMAT RAW`** — the whole batch as one undifferentiated hex blob:

    09000000000000000300000003013201240B5F5F6E616D6573706163650000000975736572
    3A3130303032D99CC23C440FFA61676502333001250B5F5F6E616D65737061636500000009
    757365723A3130303032D99CC23C440FFA6E616D6505416C6963650501150B5F5F6E616D65
    7370616365757365723A313030301982000000000000000032D99CC23C440FFA0000000000
    000002

---

**`WALGET 9`** (summary) — one line per physical record, with column family and sizes:

    start_seq=9
    end_seq=11
    type=LOGDATA,raw_value_size=1
    type=PUT,cf=default,raw_key_size=36,raw_value_size=2
    type=PUT,cf=default,raw_key_size=37,raw_value_size=5
    type=PUT,cf=metadata,raw_key_size=21,raw_value_size=25

**`WALGET 9 detail`** — decodes every record's key/value into its storage-level fields (namespace / user_key / version / sub_key, each with raw bytes + hex, plus the metadata internals):

    start_seq=9
    end_seq=11
    type=LOGDATA,raw_value_size=1,raw_value=2,raw_value_hex=32,redis_type=2
    type=PUT,cf=default,raw_key_size=36,raw_key=\x0B__namespace\x00\x00\x00\x09user:10002\xD9\x9C\xC2<D\x0F\xFAage,raw_key_hex=0B5F5F6E616D65737061636500000009757365723A3130303032D99CC23C440FFA616765,ns_size=11,ns=__namespace,ns_hex=5F5F6E616D657370616365,user_key_size=9,user_key=user:1000,user_key_hex=757365723A31303030,version=3664132129891291130,sub_key_size=3,sub_key=age,sub_key_hex=616765,raw_value_size=2,raw_value=30,raw_value_hex=3330
    type=PUT,cf=default,raw_key_size=37,...,user_key=user:1000,version=3664132129891291130,sub_key=name,sub_key_hex=6E616D65,raw_value=Alice,raw_value_hex=416C696365
    type=PUT,cf=metadata,raw_key_size=21,...,user_key=user:1000,raw_value_hex=82000000000000000032D99CC23C440FFA0000000000000002,type=hash,expire_at_ms=0,version=3664132129891291130,size=2

---

The piece I couldn't get from `POLLUPDATES` is that last view. `RESP` gives me the reconstructed command and `RAW` gives me a single opaque hex string, but neither shows **how each record is actually encoded on disk** — e.g. that a subkey record carries `version=3664132129891291130` / `sub_key=age`, the exact `raw_key_hex`, or that the metadata value decodes to `type=hash, expire_at_ms=0, size=2`. That field-level breakdown (which also stays faithful for binary/malformed data and for records `POLLUPDATES` intentionally skips, such as the secondary-subkey CF) is exactly what I reach for when debugging an encoding issue.

That said, I'm happy to avoid a new command if you'd prefer — I could fold this into `POLLUPDATES` as an extra output mode (e.g. `FORMAT DETAIL`) instead. Just let me know which direction you'd like and I'll adjust. Thanks!

git-hulk:
@weim0000 Perhaps we can extend the format for `POLLUPDATES` command? cc @apache/kvrocks-committers

weim0000:
Thanks @git-hulk — extending `POLLUPDATES` sounds good, I'll rework the PR in
that direction. Laying out a concrete plan below so the committers can weigh in
on the specifics.

**Proposed syntax** (purely additive; existing `RAW` / `RESP` behavior
unchanged):

    POLLUPDATES <sequence> [MAX <count>] [STRICT] [FORMAT <RAW|RESP|DETAIL>]

The single-batch point-inspection case that `WALGET` covered maps directly to
`POLLUPDATES <seq-1> MAX 1 STRICT FORMAT DETAIL`. Nice side effect: the
existing `STRICT` flag already provides the "sequence must actually be inside
this batch, not silently skipped forward" guarantee that `WALGET` gave, so no
new option is needed for that.

**Proposed `FORMAT DETAIL` output**, kept inside the existing
`{latest_sequence, updates, next_sequence}` envelope:

- `updates` becomes an array of batches; each batch is a Map with `start_seq`,
  `end_seq`, and `records`.
- Each entry in `records` is a Map with the storage-level fields: `type`, `cf`,
  `raw_key` / `raw_key_hex`, decoded `namespace` / `user_key` / `version` /
  `sub_key`, `raw_value` / `raw_value_hex`, and for the metadata CF
  additionally `redis_type` / `expire_at_ms` / `size`.
- Records that `RESP` intentionally skips (secondary-subkey CF, `LogData`,
  etc.) are still surfaced — that's exactly the debugging case this format is
  meant for.

**Two small decisions I'd like input on before pushing:**

1. **FORMAT name** — I lean toward `DETAIL` (matches the wording in my earlier
   examples), but `STORAGE` or `DEBUG` are equally fine if the committers
   prefer something more descriptive of the perspective.
2. **Output granularity** — is the nested Map-per-record above OK, or would you
   rather keep it closer to the flat one-line-per-record shape shown earlier?

Once those settle I'll rename the PR, fold the implementation into
`CommandPollUpdates`, and port the cppunit / gocase tests over.

PR #3618, opened 2026-09-11
fix(timeseries): avoid deleting empty key on first write to an empty series

weim0000:
## Summary

On the first write to an empty time series (created via `TS.CREATE`, no chunk
persisted yet), `upsertCommonInBatch` emits a spurious `Delete("")`, leaving an
empty-key tombstone in the WAL.

## Root Cause

When the source series has no persisted chunk, `latest_chunk_key` stays `""`,
but the rename guard only checked `new_key != latest_chunk_key`. Since `new_key`
always carries the key prefix, the condition was unconditionally true and ran
`batch->Delete("")`. The empty `latest_chunk_key` is a valid "source empty"
state — the function already relies on it a few lines below
(`was_source_empty = latest_chunk_key.empty()`); the delete branch just missed
the same guard.

## Fix

Add the empty-key precondition:

```cpp
if (i == 0 && !latest_chunk_key.empty() && new_key != latest_chunk_key) {

### CODE REVIEW

File: tests/cppunit/types/timeseries_test.cc [hunk 1]
```diff
@@ -1319,3 +1321,112 @@ TEST_F(TimeSeriesTest, DelComprehensive) {
...
+  void SetUp() override {
+    const char *path = "test_ts_wal.conf";
+    unlink(path);
+    std::ofstream output_file(path, std::ios::out);
+    output_file << "";
+
+    auto s = config_.Load(CLIOptions(path));
+    config_.db_dir = "testdb_ts_wal";
+    config_.rocks_db.compression = rocksdb::CompressionType::kNoCompression;
+    config_.rocks_db.write_buffer_size = 64 * 1024 * 1024;
```
yezhizi:
```suggestion
    config_.rocks_db.write_buffer_size = 64;
```
`write_buffer_size` is in MiB, not in byte.

yezhizi (changes requested):

tests/cppunit/types/timeseries_test.cc
weim0000: OK，thanks

yezhizi (approved):

git-hulk (approved):

PR #3623, opened 2026-09-15
feat(replication): send replica host as TLS server name

### LINKED ISSUE #3622: Support SNI on TLS replication connections
### Search before asking

### Motivation

Kvrocks supports encrypted replication using:
```
tls-replication yes
replicaof kvrocks.example.com 6666
```

However, the replica's TLS client does not send the Server Name Indication (SNI), even when replicaof contains a DNS hostname.

This prevents replication through TLS-aware Layer 4 routing implementations that use SNI to select a backend, such as Kubernetes Gateway API TLSRoute.

### Solution

Use the host from `replicaof` as the SNI server name for every TLS replication connection.

Before starting the TLS handshake, use `inet_pton()` to determine whether the configured host is a literal IPv4 or IPv6 address. If it is not an IP address, treat it as a DNS hostname and set it as the SNI server name.

M4t7e:
Send the configured `replicaof` hostname as SNI for all TLS replication connections. SNI is omitted for literal IP addresses.

Closes #3622


src/cluster/replication.cc
```diff
@@ -57,6 +57,26 @@
...
+  if (inet_pton(AF_INET, host.c_str(), &ipv4) == 1 || inet_pton(AF_INET6, host.c_str(), &ipv6) == 1) {
+    return Status::OK();
+  }
+
+  if (SSL_set_tlsext_host_name(ssl, host.c_str()) != 1) {
+    return {Status::NotOK, fmt::format("Failed to set TLS server name: {}", fmt::streamed(SSLErrors{}))};
+  }
+
+  return Status::OK();
+}
```
PragmaTwice, line 75: Could this be put into `server/tls_util.h`?

PragmaTwice (commented):
One comment. Other parts look good : )

PR #3624, opened 2026-09-15
fix(cluster): make migration response timeout configurable

wengsht:
## Summary

- add mutable `migrate-response-timeout-ms` configuration with a 5000 ms default and a 1000-300000 ms range
- apply the configured receive timeout before migration response parsing and before every `APPLYBATCH`
- add focused coverage for live config updates and responses that arrive after the previous one-second deadline

This is the timeout-only split from #3613. It intentionally excludes the subkey-iterator optimization.

## Motivation

Slot migration currently sets `SO_RCVTIMEO` to one second while checking destination responses, then reuses that socket for `APPLYBATCH`. A production canary failed after about 1.14 seconds with `Resource temporarily unavailable` even though the destination was lightly loaded. This indicates an `EAGAIN` receive timeout rather than destination saturation.

An explicit bounded setting lets operators test a longer response deadline without introducing low-level retries. Reapplying it before every batch also makes `CONFIG SET migrate-response-timeout-ms ...` effective during an active migration.

## Validation

- `./x.py check format --clang-format-path clang-format-18`
- built the `kvrocks` and `unittest` targets
- `./build/unittest --gtest_filter="IOUtil.SockReadLineWaitsForConfiguredReceiveTimeout:Config.GetAndSet"`: 2 tests passed
- `git diff --check`

🤖 Generated with [Codex](https://developers.openai.com/codex) (gpt-5.6-sol)


src/config/config.cc
```diff
@@ -243,6 +243,7 @@ Config::Config() {
       {"migrate-sequence-gap", false, new IntField(&sequence_gap, 10000, 1, INT_MAX)},
       {"migrate-batch-size-kb", false, new IntField(&migrate_batch_size_kb, 16, 1, INT_MAX)},
       {"migrate-batch-rate-limit-mb", false, new IntField(&migrate_batch_rate_limit_mb, 16, 1, INT_MAX)},
+      {"migrate-response-timeout-ms", false, new IntField(&migrate_response_timeout_ms, 5000, 1000, 300000)},
```
git-hulk, line 246: Can we keep the default configuration to `1s` to prevent from changing the existing behavior.

PR #3625, opened 2026-09-16
perf(cluster): reuse subkey iterator during migration

wengsht:
## Summary

- reuse a compatible RocksDB subkey iterator across adjacent keys during slot migration
- avoid redundant iterator construction and seeks while retaining the seek fallback
- add focused unit coverage for contiguous prefixes, gaps, and stream/non-stream column-family transitions

## Motivation

Raw key-value migration currently creates and seeks a new subkey iterator for every complex key. Metadata and subkey records use compatible ordering, so after one key is exhausted the iterator can often continue directly at the next key prefix. Reusing it removes repeated iterator setup without changing the migration protocol or slot-ownership semantics.

## Validation

- `git diff --check`
- Added focused C++ iterator tests
- The project formatter was not run locally because `clang-format-18` is unavailable on this host; CI will run formatting and the test suite.

🤖 Generated with [Codex](https://developers.openai.com/codex) (gpt-5.6-sol)


wengsht:
@LindaSummer can you help review this patch?

LindaSummer:
> @LindaSummer can you help review this patch?

I'm not so familiar with this module.

Hi @jihuayu ,
Could you help take a look when you have a chance? 😊

PR #3634, opened 2026-09-19
feat(hash): support HFE in WAL-to-RESP conversion

PragmaTwice:
Adds hash encoding context to WAL and preserves field expiration when converting hash writes and COPY/RENAME to RESP.

Legacy hash LogData remains `2`; HFE hash writes now use `2 13 1` (`kRedisHash`, `kRedisCmdHSet`, `kFieldExpiration`) to follow the existing `<type> <operation> <args...>` convention and record the hash's actual encoding even when a batch contains no metadata update.

Refs #3436; migration batching and kvrocks2redis full export are left for follow-up PRs.

Assisted-by: Codex / GPT6


src/storage/batch_extractor.cc
```diff
@@ -133,9 +133,23 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
...
+        uint64_t expire = 0;
+        auto s = metadata.DecodeSubkeyValue(&field_value, &expire);
+        if (!s.ok()) return s;
+
+        resp_commands_[ns].emplace_back(redis::ArrayOfBulkStrings({"HSET", user_key, sub_key, field_value.ToString()}));
+        // Replay expired writes too: the destination may still contain an older field value.
+        if (expire > 0) {
+          resp_commands_[ns].emplace_back(
+              redis::ArrayOfBulkStrings({"HPEXPIREAT", user_key, std::to_string(expire), "FIELDS", "1", sub_key}));
+        }
```
PragmaTwice, line 150: HSET && HPEXPIREAT: requires Redis >= 7.4
HSETEX: requires Redis >= 8.0

git-hulk (approved):

src/storage/batch_extractor.cc
```diff
@@ -133,9 +133,23 @@ rocksdb::Status WriteBatchExtractor::PutCF(uint32_t column_family_id, const Slic
     ns = ikey.GetNamespace().ToString();
 
     switch (log_data_.GetRedisType()) {
-      case kRedisHash:
-        command_args = {"HSET", user_key, sub_key, value.ToString()};
-        break;
+      case kRedisHash: {
```
jihuayu, line 136: Can we make sure the new version will never read WAL from an older version? The current code cannot tell old WAL from new WAL. If it reads an old HFE WAL, it may silently write wrong data. Is this too risky?

PragmaTwice: HFE has not been released yet. So there's no old HFE WAL.

jihuayu: Ohh，I mean old Hash Set WAL. It also used `kRedisHash` RedisType. @PragmaTwice

PragmaTwice: Yah it is considered. You can see we have `log_data_.GetHashSubkeyEncodingMode()`

jihuayu (approved):

src/storage/batch_extractor.cc
jihuayu: OK!

